Issue #765, opened 2016-05-19
Are we ready for automatic TS push?


bgruening:
@galaxyproject/core are we ready to push conda-tested tools automatic to the MTS?
I think it worked well for TTS or was there any problems?

What about packages? Do we want to upload them as well?

nsoranzo:
:+1: but at the same time we should document that on the README and mention that you can skip testing/uploading with this trick: https://docs.travis-ci.com/user/customizing-the-build/#Skipping-a-build

bgruening:
Oh yes, documentation should be updated :)

mvdbeek:
For tools that are conda-tested a definite 👍 
For packages I would say yes as well, given that we have no formal checks in place anyway.
Ideally we would test the package installation from the testtoolshed first before merging. Could we perhaps let jenkins do that?

martenson:
this is splendid, thank you fellas!

PR #772, opened 2016-05-30
Travis MTS push

bgruening:
xref: https://github.com/galaxyproject/tools-iuc/issues/765

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,4 +37,5 @@ after_success:
     if [ "$TRAVIS_PULL_REQUEST" == "false" -a "$TRAVIS_BRANCH" == "master" -a "$PLANEMO_TEST_RUN" == "True" ]
     then 
       for DIR in `cat changed_repositories.list`; do planemo shed_update --shed_target testtoolshed --shed_email $SHED_EMAIL --shed_password $SHED_PASSWORD --force_repository_creation $DIR || exit 1; done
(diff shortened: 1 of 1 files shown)


blankenberg:
+1 

Really nice to have this automated.

Although it might be better if we were using API keys, e.g. `--shed_key $TEST_SHED_KEY` and  `--shed_key $SHED_KEY` or `--shed_key_from_env TEST_SHED_KEY`... instead of having usernames and passwords in the environment. (But this is the same methodology as the TTS pushing.)

bgruening:
@blankenberg what is the advantage of API-key over user/passwd? Let me know if I should change this or feel free to merge :)

bgruening:
Thanks, let's try it!

martenson:
@bgruening Among others you can force change the key, expire it and do other fancy stuff; also you help eliminate the risk of exposing other services with the same password if you lose it.

blankenberg:
Biggest issue tends to be leaking shared passwords. In practicality, it probably doesn't matter much key vs pw, especially if there is never a leak.

One thing to consider, I guess, is whether you can change the password using the key, if not thats a bit better to prevent a leak causing an account lockout (although we can manually fix that in our case).

martenson:
@blankenberg I believe we do not allow changing passwords with just API key in `user.py`, good point

bgruening:
@martenson all points are equally true for passwords, especially if we use them correctly :)
I don't mind changing it to API-key. I find it more verbose with user/password especially if we consider this repo as blueprint for users that want to understand what we are doing.

So happy we have tests and automatic pushs now! Exciting!

nsoranzo:
As mentioned in https://github.com/galaxyproject/tools-iuc/issues/765#issuecomment-220469244 we should update the README.

Issue #777, opened 2016-06-02
hisat2 bug: toolshed dependency warning for hisat version 2.0.3


lparsons:
The main toolshed shows this warning:

```
The settings for name, version and type from a contained tool configuration file's requirement tag does not match the information for the following tool dependency definitions in the tool_dependencies.xml file, so these tool dependencies have no relationship with any tools within this repository.
* name: hisat, type: package, version: 2.0.3
```

PR #778, opened 2016-06-02
Fix hisat dependency

bgruening:
Should fix: https://github.com/galaxyproject/tools-iuc/issues/777


nsoranzo:
Thanks @bgruening!

PR #782, opened 2016-06-03
Added "../" back to patch command

Slugger70:
Last changeset removed the ../ in front of the dojo.js.patch file. This made it not work. I replaced it and it works for me now. See issue: #781

### DIFF
--- a/packages/package_jbrowse_1_12_0/tool_dependencies.xml
+++ b/packages/package_jbrowse_1_12_0/tool_dependencies.xml
@@ -11,7 +11,7 @@
                 <action type="download_by_url" sha256sum="a8bb2a56f1ca0470df2d072811db1f90fd6756feed92d8656eba1e6f67f438d3">http://jbrowse.org/wordpress/wp-content/plugins/download-monitor/download.php?id=101&amp;fmt=.zip</action>
                 <action type="download_file" sha256sum="84dfdc81c8cae61ee61e03fe6be1b9d76813d933c268609a2c24dc85349b5840">https://gist.github.com/erasche/9cf020138d787eb47891/raw/e6ac8852be51c4869bd10d1977d2f68c731ffbc1/dojo.js.patch</action>
                 <action type="shell_command">
-                    patch -p1 &lt; dojo.js.patch
+                    patch -p1 &lt; ../dojo.js.patch
                 </action>
                 <action type="change_directory">..</action>
                 <action type="move_directory_files">


nsoranzo:
Closing for the reasons explained in https://github.com/galaxyproject/tools-iuc/issues/781#issuecomment-223549240 , thanks for spotting the problem!

erasche:
@nsoranzo when you synchronized the packages, that did not include this patch.

>  the package on the ToolShed was messed up for some reason, I've synchronised it with the GitHub version now. Can you try to update the package and reinstall the jbrowse dependency?

https://toolshed.g2.bx.psu.edu/repos/iuc/package_jbrowse_1_12_0/rev/0af6238ca362 was the change you pushed, it's just a formatting change.

I've re-opened this PR, I'm :+1: on it, and will merge after you've had time to respond.

nsoranzo:
@erasche No, it's not a formatting change at all, it changes the `<action>` type from "download_by_url" to "download_file", which downloads the file in the same folder, so the '..' it is not needed. Please revert your TS upload, MTS should not be used for testing.

erasche:
@nsoranzo ah. I see you changed the action now, I missed that in the diff view. Why did you change to download_file? It was previously functional as download_by_url + `../`

Additionally I'm not seeing your change in git anywhere.

nsoranzo:
https://github.com/galaxyproject/tools-iuc/commit/bcff6c88385e589611292af11bdfa94fd858ad26#diff-f58cef68c0b88896d167eaddbf5e645c

The reason is simply because `download_file` is the appropriate action to download a single file, while `download_by_url` is better for packages because it automatically deflates an archive.

Slugger70:
I thought that when you a tarball/zip/package with download_by_url, it unarchives the files and changes to that directory.. But when you download the patch file it lands in the same directory that the archive did (i.e. one level above the current dir.) That's why the patch needs the ../ in front of it. Or am I wrong? Anyway, the only way I could get it to work was to add the ../ back in.

erasche:
@Slugger70 https://wiki.galaxyproject.org/ToolDependenciesTagSets?highlight=%28download_file%29

erasche:
@nsoranzo I think we have the same goals here (best possible packages for galaxy, most well trained and well equipped developers writing them) so I'd like to talk through this

I tested on MTS, this was not good, but sometimes the scream test works. Someone screamed, you reverted. (Thank you.)

You changed the implementation of a package. As the author of the package I feel entitled to my implementation decisions, unless they're non-functional. The change that was made, was buried in an unrelated PR, without developer education. (I looked through the "Remove whitespace" pr, that looked like the *only* time an implementation was changed (but my eyes continue to fail me in detecting small changes, so I may be incorrect))

Since I believe we both have the same end goals, I'd like to propose the following:

- @erasche will keep their testing on the TTS in the future, so end users don't hit issues like this.
- It would be extremely helpful for dev education if such changes could come in a separate PR so the author (@erasche) could be learn that there's another function that might behave slightly more correctly (judging by docs), and that produces a possibly cleaner `tool_dependencies.xml`.
- If anyone is making wide sweeping changes which touch every package, it would be helpful if the one place where something unrelated to the title of the PR, it was mentioned in the PR (#617).

(Closing the issue as it would introduce a bug solved by either of our changesets. Not as a statement of finality.)

nsoranzo:
@erasche Completely agree, I apologise for introducing the unrelated change in #617, I guess I was checking that `download_by_url` was the first action in the `<actions>` group and forgot to mention in the PR description (like I did instead about the changes for `packages/package_bcftools_1_2/tool_dependencies.xml` . Sorry about that!

martenson:
:koala:

erasche:
@nsoranzo ok, thank you. I've merged in the your, more correct, download_file action to my testing branch. https://github.com/galaxyproject/tools-iuc/commit/3bbca939ca8a3298a3a2d6450abb04a96851e1ed

PR #784, opened 2016-06-07
Lapack: build shared lib too

### LINKED ISSUE #780: package_trinity_2_1_1 fails with /usr/bin/ld: cannot find -lblas
We tried adding LD_LIBRARY_PATH variable to lapack dependency but the issue still persists, any ideas.

abretaud:
Hi,
This should fix #780: the lapack was only built in static mode, but shared libs are needed. I now compile both as many packages seem to depend on lapack, and I'm not sure if they expect static builds
Anthony (back to business!)

### DIFF
--- a/packages/package_lapack_3_4/tool_dependencies.xml
+++ b/packages/package_lapack_3_4/tool_dependencies.xml
@@ -14,7 +14,7 @@
                     cd build &&
                     cmake .. -DCMAKE_INSTALL_PREFIX=$INSTALL_DIR/lapack -DCMAKE_Fortran_FLAGS='-O2 -fPIC' &&
                     make &&
-                    make install
+                    make install &&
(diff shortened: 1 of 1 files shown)


packages/package_lapack_3_4/tool_dependencies.xml
```diff
@@ -8,21 +8,20 @@
...
-                    cmake .. -DCMAKE_INSTALL_PREFIX=$INSTALL_DIR/lapack -DCMAKE_Fortran_FLAGS='-O2 -fPIC'
-                </action>
-
+                <!-- compile both static and shared library -->
                 <action type="shell_command"><![CDATA[
                     mkdir build &&
                     cd build &&
                     cmake .. -DCMAKE_INSTALL_PREFIX=$INSTALL_DIR/lapack -DCMAKE_Fortran_FLAGS='-O2 -fPIC' &&
                     make &&
                     make install
```
bgruening: && is missing here, isn't it?

nsoranzo:
LGTM

nsoranzo:
I've pushed `package_lapack_3_4` to MTS.

PR #785, opened 2016-06-07
WIP: Mothur toolshed dependencies

shiltemann:
adds toolshed dependencies definition for mothur as well as legacy blast version 2.2.22 package (to match the version used in conda)


tools/mothur/macros.xml
```diff
@@ -1,7 +1,7 @@
 <macros>
     <xml name="requirements">
         <requirements>
-            <requirement type="package" version="1.36.1">mothur</requirement>
+            <requirement type="package" version="1.36.1" owner="iuc">mothur</requirement>
```
bgruening: You don't need an owner here.

shiltemann: hmm ok, I took a guess, when I tested this in testtoolshed it tried to reinstall the mothur dependency for the second of the separate mothur tools I tried to install ..maybe I've got something else misconfigured?

shiltemann: or do I need to make it a package if we're going to have multiple separate tools?

bgruening: This would be great otherwise we will install mothur 125 times :)

shiltemann: lol, yeah that would be bad, okay, back to the drawing board :)

shiltemann:
@bgruening ok, I created a separate package for mothur, tested this in testtoolshed and it seems to work, I can't tell what travis is failing on though?

bgruening:
LGTM. Travis is failing mainly because of the missing/wrong datatypes that are missing in the 16.4 release I guess.

PR #791, opened 2016-06-10
Add Entrez Direct tools. efetch, einfo, esearch, so far.

blankenberg:
Partially based off of tools-iuc/tools/ncbi_entrez_eutils


erasche:
@blankenberg you're duplicating the functionality of ncbi_entrez_eutils, except using entrez direct? Why? They should have absolutely identical functionality. The last time I looked the command line `direct` tools were pretty dumb wrappers around entrez's API, with no more features than biopython's.

If the tools add new functionality, could we consider adding them ncbi_entrez_eutils? I think it would be more functional for users to install a single set of entrez API tools than to have the option of two sets with extremely overlapping functionality and small differences underlying them.

blankenberg:
I could not get the ncbi_entrez_eutils to work properly. It expects JSON in lots of places, but makes xml instead. For example, the simplest case of passing the output of esearch to efetch, couldn't get it to work.

Using this set of tools, along with other existing tools, I can create a workflow that will create a dataset collection containing all of the SRRs from an SRA project ID. This works currently with this set of tools.

erasche:
@blankenberg could we please fix your issues in ncbi_entrez_eutils rather than wrapping a completely secondary copy?

> I could not get the ncbi_entrez_eutils to work properly.

I'm sorry to hear that. I only use a subset of them, so there may be workflows I didn't consider.

> It expects JSON in lots of places, but makes xml instead. 

The only places that json is expected is passing around a WebEnv, and it looks like a couple of outputs are as json instead of xml.

> For example, the simplest case of passing the output of esearch to efetch, couldn't get it to work.

This should not be the case. How are you doing these searches? Are you using a WebEnv to store your ESearch results before fetching them?

blankenberg:
@erasche Using WebEnv on eutils results in this error:
```
esearch.py: error: unrecognized arguments: --use_history
```

However, eutils exists only on the test toolshed, and that is the version that I am using. I did not check that it matches the latest in tools-iuc GH though. Perhaps it has been updated in GH and not toolshed (although quick look at efetch.py shows that there is no `--use_history` parameter, although there is `--history_file`. 

A very simple command-line example that can be reconstituted with these tools:
```esearch -db sra -query "SRP000007" | efetch -format runinfo```
results in
```
Run,ReleaseDate,LoadDate,spots,bases,spots_with_mates,avgLength,size_MB,AssemblyName,download_path,Experiment,LibraryName,LibraryStrategy,LibrarySelection,LibrarySource,LibraryLayout,InsertSize,InsertDev,Platform,Model,SRAStudy,BioProject,Study_Pubmed_id,ProjectID,Sample,BioSample,SampleType,TaxID,ScientificName,SampleName,g1k_pop_code,source,g1k_analysis_group,Subject_ID,Sex,Disease,Tumor,Affection_Status,Analyte_Type,Histological_Type,Body_Site,CenterName,Submission,dbgap_study_accession,Consent,RunHash,ReadHash
SRR000080,2008-04-04,2008-04-04,158320,42230342,0,266,98,,http://sra-download.ncbi.nlm.nih.gov/srapub/SRR000080,SRX000013,1970180360,WGS,RANDOM,GENOMIC,SINGLE,0,0,LS454,454 GS FLX,SRP000007,PRJNA19663,,19663,SRS000007,SAMN00000007,simple,445974,Erysipelatoclostridium ramosum DSM 1402,19663,,,,,,,no,,,,,WUGSC,SRA000146,,public,1C20B1B2D8895EB26B96264F12F718FA,72A8026356704ADF9294E5E1140F98B4
SRR000081,2008-04-04,2008-04-04,180220,48201615,0,267,111,,http://sra-download.ncbi.nlm.nih.gov/srapub/SRR000081,SRX000013,1970180360,WGS,RANDOM,GENOMIC,SINGLE,0,0,LS454,454 GS FLX,SRP000007,PRJNA19663,,19663,SRS000007,SAMN00000007,simple,445974,Erysipelatoclostridium ramosum DSM 1402,19663,,,,,,,no,,,,,WUGSC,SRA000146,,public,E2B8898B2078C6AAC047A44770D0539F,38243F6991989DA4301ED895902B0EB6
```
which i can then do eg (with equivalents all in Galaxy):
```
esearch -db sra -query "SRP000007" | efetch -format runinfo | cut -d ',' -f 1 | grep 'SRR'
```
giving:
```
SRR000080
SRR000081
```
which makes great input to eg. the sra fastq dump tool.

With the eutils search tool, I ask for eg list of plain-text UIDs and it dumps out a bunch of XML.

Entrez direct is a standard set of utilities that continues to be well-maintained by NCBI, with several updates just this year.

**Comment by erasche** [comment]
@blankenberg 

> esearch.py: error: unrecognized arguments: --use_history

ok! that's a bug we can fix!

> However, eutils exists only on the test toolshed, and that is the version that I am using. I did not check that it matches the latest in tools-iuc GH though. Perhaps it has been updated in GH and not toolshed (although quick look at efetch.py shows that there is no --use_history parameter, although there is --history_file.

No, it hasn't been updated. I can fix this bug for you and push an update (and to MTS). 

> With the eutils search tool, I ask for eg list of plain-text UIDs and it dumps out a bunch of XML.

There is an xpath tool which permits XML queries. That said, that sounds like something the existing suite of tools should be able to do! This would have been a nice feature request, that I would be more than happy to add.

> Entrez direct is a standard set of utilities that continues to be well-maintained by NCBI, with several updates just this year.

I'm not arguing that direct are the official command line tool suite provided by them, but we use the API which *should* be equivalent.

**Comment by erasche** [comment]
Ok, here's what happened with the existing set of tools:

- $use_history was a left-over artifact of earlier design decisions. Removing it allows for proper functionality.
  - This should have been caught by tests, but we don't run those and mocking Entrez for them would be extremely painful.
- runinfo was not listed as a possible type for SRA.
    - We can update the types more regularly.

I strongly believe XPath queries are the right way to go here. The direct tools make decisions for you about what data is tabular, whereas with XML + XPath you can make arbitrary queries that are much more powerful (and more clear. I'd take a tool which says `//Run` over one which does `cut -d',' -f 1` any day.)

**Comment by blankenberg** [comment]
Well, I can get the runinfo with this set of tools as xml by setting the mode e.g. 
```esearch -db sra -query "SRP000007" | efetch -format runinfo -mode xml```

**Comment by erasche** [comment]
Yep! I saw that in your command an am adding all of your updated formats to the existing tool interfaces.

With that said, do you think the existing tool set (+bug fixes, +updated format list) + xpath will meet your needs, or will you still require other things that are only available in entrez direct?

**Comment by blankenberg** [comment]
I'd also recommend allowing manual entry of optional db, format, and mode to address not-yet-known or unneeded options for passing. For example, if I am using a WebEnvHistory from esearch, I shouldn't necessarily have to pass a db to efetch, and I may only want to pass a format, but not any mode, etc. 

Also need to make sure that formats are being set correctly on the outputs for eg. tabular, json, etc.

We should be able to get any type of formatted output from NCBI that NCBI provides.

**Comment by erasche** [comment]
> I'd also recommend allowing manual entry of optional db, format, and mode to address not-yet-known or unneeded options for passing. 

Sure, that makes sense.

> For example, if I am using a WebEnvHistory from esearch, I shouldn't necessarily have to pass a db to efetch

> I may only want to pass a format, but not any mode, etc.

I tried to handle this option nicely by providing the mode + format as a list of all possible combinations. Easier to see exactly what you can do.

> Also need to make sure that formats are being set correctly on the outputs for eg. tabular, json, etc.

OK, if you have any examples where it isn't happening correctly, please let me know.

> We should be able to get any type of formatted output from NCBI that NCBI provides.

Indeed.

**Comment by erasche** [comment]
`Fetch-db+history` may not be trivial to add, but it's a very small overhead. The API requires the specification of a DB. I wonder how the entrez direct tools get around this.

**Comment by blankenberg** [comment]
> `Fetch-db+history` may not be trivial to add, but it's a very small overhead. The API requires the specification of a DB. I wonder how the entrez direct tools get around this.

Its part of the entrez direct xml, so they probably pull it from there when not provided:
```
<ENTREZ_DIRECT>
  <Db>nucleotide</Db>
  <WebEnv>NCID_1_25024504_130.14.22.215_9001_1465565885_224033649_0MetA0_S_MegaStore_F_1</WebEnv>
  <QueryKey>1</QueryKey>
  <Count>1</Count>
  <Step>1</Step>
</ENTREZ_DIRECT>
```

**Comment by erasche** [comment]
```
+	<ERROR>NCBI C++ Exception:

+    Error: TXCLIENT(CException::eUnknown) "/pubmed_gen/rbuild/version/20150724/entrez/2.11/src/internal/txclient/TxClient.cpp", line 1025: ncbi::CTxRawClientImpl::readAll() --- Read failed: EOF (the other side has unexpectedly closed connection), peer: 130.14.18.45:8064
```

sure love that NCBI API. Otherwise passing.

erasche:
@blankenberg I'm reverting my original stance on this PR. NCBI broke the external API wrappers and I don't want to deal with fixing them. I'd rather replace the existing wrappers with the official suite under the hood with the hope that it's better supported.

I'll PR your suite here sometime in the next couple of months here and merge your wrappers in, if that's OK with you.

blankenberg:
Should consider adding NCBI API KEY support: https://ncbiinsights.ncbi.nlm.nih.gov/2017/11/02/new-api-keys-for-the-e-utilities/

bernt-matthias:
Lets retest and finish this :) .. I did not settle my mind on the question if one should replace the (bio)python based implementation (silently) or by adding this one and deprecation the original one. 

The latter seems more appealing to me.

bernt-matthias:
@blankenberg some fixes for your PR are here: https://github.com/blankenberg/tools-iuc/pull/2 .. last thing might be the 404 url.

bernt-matthias:
> Should consider adding NCBI API KEY support

Good idea I guess. Question is how to implement this. If left optional then it does not help much on larger Galaxy instances. If mandatory it might be annoying for users on smaller instances.

bernt-matthias:
https://github.com/galaxyproject/tools-iuc/pull/4462

PR #107, opened 2015-04-14
Initial version of XPath tool

erasche:
Allows querying XML files.

PR #792, opened 2016-06-10
Existing Entrez tool updates

erasche:
Should fix bugs raised in #791 

- [x] esearch.py: error: unrecognized arguments: --use_history
- [x] format list updated with new formats.
- [x] xpath (#107) was not on MTS
- [ ] Entrez to MTS
 
@blankenberg would you be able to test the update to see if it allows you to accomplish the queries you want? I cannot add the "auto database" feature on such short notice I'm afraid.


bgruening:
@erasche you get a +1 from me if you remove this line https://github.com/galaxyproject/tools-iuc/blob/master/.tt_blacklist#L18 and it stays green :)

blankenberg:
I'll give it a shot, thanks.

blankenberg:
Using Esearch for sra on SRP000007: "ESearch Results for SRP000007" (output 1), the output is marked as json, but it is an xml file.

Using Efetch on the webenv (output2) from SRP000007 against SRA creating a SraRunInfo (xml), filetype is set to "text" (it does contain xml), so I can't use it in xpath, without manually changing the datatype to xml.

Once I change the datatype, I can use XPATH on the Efetch Data, with `//Run/text()` to extract the flat list of IDs I need, however. 

When getting fasta files from nucleotide using 'NC_001552', they are also txt instead of fasta. So possibly after all of the formats are set correctly, most of the functionality is there. However, being able to manually set `format` and `mode` and having the output datatypes match in efetch would be very useful.

bgruening:
@erasche will it help if you annotate the tests with ftpye? This should catch a few of @blankenberg errors.

erasche:
@bgruening sure, I can do that.

bgruening:
@erasche we need a rebase here.
@blankenberg can you help to get the datatypes correct?

erasche:
@bgruening yep, can do.

erasche:
I'll patch datatypes too

erasche:
@bgruening rebased, squashed some since I'd left a mess (and in the process of rebasing anyway). Added ftypes. We'll see if it passes.

bgruening:
Thanks! These are nice changes and I hope it will fix Dan's problems.

erasche:
Hope so too, but the datatypes may still be weird. Those are hard to track down issues unfortunately :(

.tt_blacklist
```diff
@@ -1,21 +1,30 @@
...
 tools/gatk2
 tools/gemini
 tools/gff3_rebase
+tools/hmmer3
+tools/htseq
 tools/jbrowse
 tools/macs2
 tools/meme
-tools/ncbi_entrez_eutils
+tools/mothur
```
bgruening: @erasche this seems to be a wrong merge. All these tools should be under testing already.

erasche: ah, bother. I did `git checkout --theirs`, looked ok. Didn't check the diff against master, will do next time.

erasche:
:green_heart:

bgruening:
Great!

erasche:
@blankenberg, when you get a chance, would you mind testing the updates? Please let me know if any of your use cases are not well covered... I am very interested in having this be the best entrez suite possible.

PR #793, opened 2016-06-10
mothur: symlink inputs to working dir

shiltemann:
mothur creates output files at the location of its input files, so we need to symlink those to the working directory to avoid clashes and clutter.

additionally, because mothur doesn't like spaces or dashes in filepaths, *all* input datasets should be symlinked (not just the ones used as basename for outputs), so if you see any I've missed please let me know

..also encountered one small issue with the datatypes, will PR a fix shortly


tools/mothur/align.seqs.xml
```diff
@@ -33,13 +35,12 @@
...
-        | mothur &&
+        | ./mothur &&
 
         ## move output files to correct destination
-        prefix="$candidate" &&
         mv mothur.*.logfile "$logfile" &&
-        mv \${prefix%.dat}.align "$out_file" &&
-        mv \${prefix%.dat}.align.report "$report"
+        mv candidate.align "$out_file" &&
+        mv candidate.align.report "$report"
```
bgruening: We will not change this now, but maybe in a later version we can just use `from_work_dir` in the output section of Galaxy and save this move step.

tools/mothur/chimera.slayer.xml
```diff
@@ -6,13 +6,21 @@
     <expand macro="requirements"/>
     <expand macro="version_command"/>
     <command detect_errors="aggressive"><![CDATA[
-        ## cd to mothur directory otherwise it cannot find
-        ## relative blast location on execution :/
-        loc=`which mothur`;
-        loc=\${loc%/mothur};
-        cd \$loc;
+        ## enable mothur to find blast
+        loc=`which mothur` && ln -s "\${loc%/mothur}"/* . &&
```
bgruening: This hack is only needed because of tool_dependencies? blast should be on PATH in conda, isn't it?

tools/mothur/align.seqs.xml
shiltemann: ah yes, of course, much nicer :)

tools/mothur/chimera.slayer.xml
shiltemann: yep, if only mothur looked on the PATH..

shiltemann: actually it works without this for tool_dependencies but not conda, even though they both have the executables linked in the same location relative to mothur executable.. but there's something about       conda that makes mothur look for the blast executables in a different location..

tools/mothur/align.seqs.xml
shiltemann: I may just do this tomorrow, not that much work and just going to be bothering me now :P

bgruening: Sorry Saskia!
Often you have a * in the filename than you need to leave it by `mv` but otherwise `from_work_dir` should be fine.

shiltemann: ah yes, mothur often names its output with some of the settings used, so most of the time you need that wildcard, so most of them probably can't be changed indeed

tools/mothur/make.group.xml
```diff
@@ -6,17 +6,22 @@
     <expand macro="requirements"/>
     <expand macro="version_command"/>
     <command detect_errors="aggressive"><![CDATA[
+        ## create symlinks to input datasets
+        ln -s "$fasta" fasta.dat &&
+        #for i in $inputs:
+            ln -s "$i.fasta" fasta${inputs.index($i)}.dat &&
```
bgruening: Do you need the real index value or do you just want to create a unique filename? If the latter, you can simply do:
```
for $counter, $i in $inputs:
    fasta_$counter.dat
end for
```
An other thing, the `i` should be `$i` isn't it? Or is this a secret cheetah feature?

tools/mothur/merge.files.xml
```diff
@@ -6,12 +6,17 @@
     <expand macro="requirements"/>
     <expand macro="version_command"/>
     <command detect_errors="aggressive"><![CDATA[
-        #set infiles=str($merge.inputs).replace(',','-').replace(' ','')
+        ## create symlinks to input datasets
+        #for i in $merge.inputs:
```
bgruening: `$i`?

tools/mothur/seq.error.xml
```diff
@@ -6,65 +6,74 @@
...
-            && mv \${prefix%.dat}*.error.count "$count_out"
+            && mv fasta_in*.error.count "$count_out"
         #end if
         #if 'matrix' in str($output_sel).split(','):
-            && mv \${prefix%.dat}*.error.matrix "$matrix_out"
+            && mv fasta_in*.error.matrix "$matrix_out"
         #end if
         #if 'ref_query' in str($output_sel).split(','):
-            && mv \${prefix%.dat}*.error.ref "$ref_query_out"
+            && mv fasta_in*.error.ref "$ref_query_out"
```
bgruening: @jmchilton this seems to be like a common pattern in many tools. Do you think we could enhance the `from_work_dir` to use a patter that should only fine one file?

tools/mothur/make.group.xml
shiltemann: hah, guess cheetah is smart enough to fix my typos :P ..will update this, just need a unique filename indeed :)

bgruening:
As usual thanks Saskia!

PR #808, opened 2016-06-15
Fix bam.h not found during stacks compilation

abretaud:
Another install problem with stacks package


abretaud:
MTS is updated (and the compilation worked on our server)

PR #809, opened 2016-06-15
Propogate the Perl environment to the perl_xml_parser_expat recipe.

gregvonkuster:
This will allow dependent packages to locate the Perl environment at compile time and / or runtime.


gregvonkuster:
Thanks @bgruening This has been pushed to both the TTS and the MTS.

PR #810, opened 2016-06-15
Upgrade Fimo version 4.11.0.3 and fix functional tests

gregvonkuster:
- All functional test should now pass for both Fimo and Meme (if not using conda).
- Fimo is enhance with the following new options:
  a. Output a collection consisting of a separate dataset for each motif in the input
  b. When "Check each sequence header for UCSC style genomic coordinates" is set to yes, allow the user to choose to "Remove duplicate entries as defined by the unique GFF coordinates".


tools/meme/fimo.xml
```diff
@@ -99,10 +104,23 @@
...
                 <param name="no_qvalue" label="Do not compute a q-value for each p-value" type="boolean" truevalue="--no_qvalue" falsevalue="" checked="True" help="The q-value calculation is that of Benjamini and Hochberg (1995)."/>
                 <param name="norc" label="Do not score the reverse complement DNA strand" type="boolean" truevalue="--norc" falsevalue="" checked="False" />
-                <param name="parse_genomic_coord" label="Check each sequence header for UCSC style genomic coordinates" type="boolean" truevalue="--parse_genomic_coord" falsevalue="" checked="False" />
+                <conditional name="parse_genomic_coord_cond">
+                    <param name="parse_genomic_coord" label="Check each sequence header for UCSC style genomic coordinates" type="boolean" truevalue="--parse_genomic_coord" falsevalue="" checked="False" />
+                    <when value="--parse_genomic_coord">
+                        <param name="remove_duplicate_coords" type="select" label="Remove duplicate entries in unique GFF coordinates?" help="Remove duplicate entries as defined by the unique GFF coordinates">
```
bgruening: This is a boolean, isn't it? One boolean in a conditional can be modelled as one boolean. Why have you changed this?

bgruening:
I would rather see conda tests ;)

bgruening:
Besides that boolean thing it looks good.

tools/meme/fimo.xml
gregvonkuster: @bgruening I'm not quite sure I understand your question, but I'll try to clarify the enhancement here.  The original "parse_genomic_coord" boolean, when set to true would"Check each sequence header for UCSC style genomic coordinates".  I have enhanced this feature so also allow the user to remove duplicate coordinates when this value is set to True.  This enhancement does not change the original behavior if the user elects to not remove duplicate coordinates.  Does this make sense?

gregvonkuster:
@bgruening I spent some time trying to get conda tests to pass, but was not able to get the required packages (i.e., meme and its dependencies) installed with conda. :(

gregvonkuster:
Regrind the boolean, is there a different approach you can provide?  I'm happy to change it to a better approach, but I'm not seeing it.

tools/meme/fimo.xml
bgruening: Ok, I see makes sense. The only thing that does not fit our standards is the conditional boolean. http://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html?highlight=conditionals#booleans

I would actually make the conditional a select and the remove_duplicate_coords a boolean.

gregvonkuster: Ah, ok. I'll make this change.

bgruening:
What happens when you use this:
```
    <requirement type="package" version="1.3.23">graphicsmagick</requirement>
    <requirement type="package" version="4.11.1">meme</requirement>
```

gregvonkuster:
That worked - thanks!!

gregvonkuster:
Fixes and changes should all be included here: https://github.com/galaxyproject/tools-iuc/pull/810/commits/e4cc2c9b3eaba35b0219708f52b901a04abcecba  Thanks!

nsoranzo:
@gregvonkuster Can you also remove `tools/meme` from `.tt_blacklist` file to enable conda testing on Travis, please?

gregvonkuster:
Done.  So what is the process we are using for handling these differences between the packages available in the Tool Shed verses the packages available in conda?  With the changes made to these tools so that conduit tests will pass, there will need to be these new packages (i.e., meme 4.11.1 and all underlying requirements and graphichsmagick 1.3.23) added to the Tool Shed right?

tools/meme/fimo.xml
```diff
@@ -1,11 +1,12 @@
-<tool id="meme_fimo" name="FIMO" version="4.11.0.2">
+<tool id="meme_fimo" name="FIMO" version="4.11.1.0">
     <description>- Scan a set of sequences for motifs.</description>
     <requirements>
-        <requirement type="package" version="6.9.3">imagemagick</requirement>
-        <requirement type="package" version="4.11.0">meme</requirement>
+        <requirement type="package" version="1.3.23">graphicsmagick</requirement>
+        <requirement type="package" version="4.11.1">meme</requirement>
```
nsoranzo: graphicsmagick 1.3.20 is available both in conda and TS, I would suggest to use that version. Remember to also update `tool_dependencies.xml` to depend on `package_graphicsmagick_1_3_20` repository.

Not sure what is best to do for meme.

gregvonkuster: I'll wait for feedback from @bgruening on this.  Hopefully it is easier to add support for meme 4.11.0 to conda than it is to add support for meme 4.11.1 to the Tool Shed.

bgruening: It should be pretty easy to add it to conda. Just wondering if it would not be better to add a newer version to the ToolShed? Is it more than changing the URL in a new tool_dependencies.xml file?

bgruening:
> @nsoranzo @bgruening I think the IUC needs to consider this issue if it has not yet done so. Clearly moving to conda recipes is an improvement over the tool_dependencies.xml recipes still required by the Tool Shed. However, conda recipes should be created to replace those that are currently in the Tool Shed even if they are for older versions of the packages. I feel it is counter-productive to be forced to add new, but soon deprecated recipes to the Tool Shed because conda does not support the versions that are available in the Tool Shed and we want to have conda tests passing. My choice for this specific case would be to revert the tools back to using the versions of the tool_dependencies.xml recipes already in the Tool Shed and add recipes to conda to support them.

Feel free to do so @gregvonkuster. You are also part of the IUC and I just offered my opinion that I would like to see conda testing rather than tool_dependency testing that I can not easily do.
This PR is/was fine, it would be better with conda support. I will hope that we can agree at the GCC on the way forward here. I agree that the current situation is confusing and that I may have pushed too hard for conda during the last month. Sorry for this.

tools/meme/fimo.xml
gregvonkuster: The best-case scenario for adding the new recipe to the Tool Shed is that it requires just the URL change in the meme recipe.  However, this requires testing with the underlying packages, adding a new entry to cargo port, getting everything reviewed and approved, etc etc.  So although it is probably reasonable to take this approach, it is at least a few hours of work.  I would rather have those few hours invested in the conda end than the Tool Shed end if conda recipes will replace tool_dependencies.xml recipes starting shortly after the GCC.

Even if the tool_dependencies.xml recipes are used for the foreseeable future, I think the optimal approach is to not be forced to upgrade recipes in the Tool Shed just so that we can get conda tests to pass for tool enhancements.

I'll revert the changes to this PR back to use the packages that are currently available in the Tool Shed.

nsoranzo: Any particular reason not to use graphicsmagick 1.3.20, as I suggested above? It's already a step forward.

gregvonkuster: Just that ImageMagick has been used for the tool until now.  Is ImageMagick deprecated?

nsoranzo: I don't think ImageMagick is deprecated, GraphicsMagick was forked from it many years ago and they are still developed in parallel. Only reason to switch would be bioconda support, but creating a bioconda recipe for ImageMagick would be even better, if/when you would like to support conda testing.

gregvonkuster: I certainly want to support conda testing - I'm continuing to invest significant time into migrating the ImageJ tools here https://github.com/bgruening/galaxytools/tree/master/tools/image_processing/imagej2 to conda.  However, my time is currently stretched every thin across many components of the ChIP-exo Galaxy environment that I am setting up, so now is not a good time for me to invest in making these changes to the limo tool.  It is difficult for me to set aside the time to even install and test the GraphicsMagick package with meme right now.

I would like to merge this PR as is since I need these enhancements asap in order to get the ChIP-exo workflows completed, and the changes in this PR do not introduce any new issues into the conda migration work.  As soon as I can, I will submit another PR for meme and limo to accommodate conda if that is ok.

nsoranzo: Sure, no rush @gregvonkuster!

gregvonkuster: Thanks -  am going to merge this then, and I promise to come back to it with a separate PR as soon as I can!  ;)

gregvonkuster:
These enhancements and test data fixes have been uploaded to both the TTS and the MTS.

PR #814, opened 2016-06-16
Load cmake env before compiling lapack

### LINKED ISSUE #812: Bug: package_lapack_3_4 unable to find cmake dependency
When I attempt to install the latest [trinity package](https://toolshed.g2.bx.psu.edu/view/iuc/trinity/f7564c3f9dc0) (revision 6:f7564c3f9dc0) from the main toolshed, I've run into an error with the lapack dependency. `package_lapack_3_4` installs, missing the `lapack` dependency due to an error during installation: 

    /bin/sh: line 3: cmake: command not found

The `package_cmake_3_2_3` installs correctly, but it seems that the `PATH` is not being set properly during installation.

This is using `release_16.04` of Galaxy, on CentOS 6.

abretaud:
This should fix #812

### DIFF
--- a/packages/package_lapack_3_4/tool_dependencies.xml
+++ b/packages/package_lapack_3_4/tool_dependencies.xml
@@ -9,7 +9,7 @@
                 <action type="download_by_url" sha256sum="60a65daaf16ec315034675942618a2230521ea7adf85eea788ee54841072faf0">https://depot.galaxyproject.org/software/lapack/lapack_3.4.2_src_all.tar.gz</action>
 
                 <action type="set_environment_for_install">
-                    <repository name="package_cmake_3_2_3" owner="iuc" prior_installation_required="True">
+                    <repository name="package_cmake_3_2_3" owner="iuc">
                         <package name="cmake" version="3.2.3" />
                     </repository>
                 </action>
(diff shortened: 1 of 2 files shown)


packages/package_lapack_3_4/tool_dependencies.xml
```diff
@@ -8,6 +8,12 @@
             <actions>
                 <action type="download_by_url" sha256sum="60a65daaf16ec315034675942618a2230521ea7adf85eea788ee54841072faf0">https://depot.galaxyproject.org/software/lapack/lapack_3.4.2_src_all.tar.gz</action>
 
+                <action type="set_environment_for_install">
+                    <repository name="package_cmake_3_2_3" owner="iuc" prior_installation_required="True">
```
bgruening: You don't need `prior_installation_required="True"`

abretaud:
Oh, and I've pushed a tiny fix for #815 too

bgruening:
Thanks @abretaud

lparsons:
Awesome, thanks guys... Let me know when the main toolshed is updated and I'll retest.

abretaud:
Toolsheds are updated now

Issue #815, opened 2016-06-16
Bug: package_atlas_3_10 fails to compile during install


lparsons:
When I attempt to install the latest [trinity package](https://toolshed.g2.bx.psu.edu/view/iuc/trinity/f7564c3f9dc0) (revision 6:f7564c3f9dc0) from the main toolshed, I've run into an error with the atlas dependency. `package_atlas_3_10` installs, missing the `atlas` dependency due to an error during installation: 

    mkdir: cannot create directory `ATLAS/build': No such file or directory

This is using `release_16.04` of Galaxy, on CentOS 6.

EDITED: Changed name to clarify, this issue is to reference the problems during install when compiling. Opened issue #831 to track issues during install from precompiled linux_x64 package.

abretaud:
Thanks for being the trinity beta tester!
Are you installing on a 64 or 32 bit version of centos?
Looking at the atlas package, on a x64 linux it should install precompiled binaries
Anyway, I think it should be fixed with #814

lparsons:
@abretaud You're welcome!
I'm installing on 64bit CentOS 6.

lparsons:
@abretaud I'm testing the fix now, however, looking at the logs more carefully, you are correct. It is attempting to install the precompiled version, but aparently another bug is preventing that and causing it to fall back to compiling...

```
tool_shed.galaxy_install.tool_dependencies.recipe.step_handler DEBUG 2016-06-16 16:34:50,023 Attempting to download via url: https://depot.galaxyproject.org/software/atlas/atlas_3.10.2_linux_x64.tar.gz
tool_shed.galaxy_install.tool_dependencies.recipe.step_handler DEBUG 2016-06-16 16:35:11,984 Successfully downloaded from url: https://depot.galaxyproject.org/software/atlas/atlas_3.10.2_linux_x64.tar.gz
tool_shed.galaxy_install.tool_dependencies.recipe.install_environment:atlas DEBUG 2016-06-16 16:35:11,991 +
                        command -v gfortran || return 0
                        BUNDLED_LGF_CANON=/opt/galaxy/galaxy-app/tool-dependencies/atlas/3.10.2/iuc/package_atlas_3_10/f27d2f8761ae/lib/libgfortran.so.3.0.0 &&
[...]

PR #818, opened 2016-06-17
Fix for atlas compilation

abretaud:
This shold fix part of #815 issue

### DIFF
--- a/packages/package_atlas_3_10/tool_dependencies.xml
+++ b/packages/package_atlas_3_10/tool_dependencies.xml
@@ -48,9 +48,9 @@
                     <action type="shell_command"><![CDATA[
                         cd ATLAS &&
                         mkdir build &&
-                        patch -p1 < ../static_full_blas_lapack.diff &&
-                        patch -p1 < ../shared_libraries.diff &&
-                        patch -p1 < ../cpu-throttling-check.diff &&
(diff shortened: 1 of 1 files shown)


packages/package_atlas_3_10/tool_dependencies.xml
```diff
@@ -48,9 +48,9 @@
                     <action type="shell_command"><![CDATA[
                         cd ATLAS &&
                         mkdir build &&
-                        patch -p1 </host/static_full_blas_lapack.diff &&
-                        patch -p1 </host/shared_libraries.diff &&
-                        patch -p1 </host/cpu-throttling-check.diff &&
+                        patch -p1 < ../static_full_blas_lapack.diff &&
+                        patch -p1 < ../shared_libraries.diff &&
+                        patch -p1 < ../cpu-throttling-check.diff &&
```
nsoranzo: I think also the file names are wrong, shouldn't they be the ones downloaded in lines 43-45?

nsoranzo:
Not sure this fixes everything, but a good step forward, thanks @abretaud!

abretaud:
Yep, I have no clue for the problem with precompiled binaries...

nsoranzo:
Pushed to MTS.

PR #820, opened 2016-06-17
Trinity: add quotes to TRINITY_MEM_OPTIONS

abretaud:
We need quotes here or env.sh will fail to set TRINITY_MEM_OPTIONS variable (and trinity will use as much memory as it wants)


nsoranzo:
I'd use single quotes, but +1 anyway.

PR #824, opened 2016-06-18
Fix travis testing and include initial parallel builds

bgruening:


.travis.yml
```diff
@@ -26,11 +35,17 @@ script:
   - |
     if [ -s changed_repositories.list ]
     then
-      for DIR in `cat changed_repositories.list`; do planemo conda_install --conda_ensure_channels iuc,bioconda,r --conda_prefix $CONDA_PREFIX $DIR ; done
-      for DIR in `cat changed_repositories.list`; do planemo test --conda_dependency_resolution --conda_prefix $CONDA_PREFIX --galaxy_branch $GALAXY_RELEASE  --galaxy_source $GALAXY_REPO --skip_venv $DIR || exit 1; done
+      split -n l/4 changed_repositories.list
       export PLANEMO_TEST_RUN="True"
```
nsoranzo: PLANEMO_TEST_RUN should be set below, after checking that CHUNK is not empty.

.travis.yml
```diff
@@ -26,11 +35,17 @@ script:
...
     fi
-    
+  - echo $CHUNK
+  - |
+    if [ ${CHUNK} ] && [ -s ${CHUNK} ] 
+    then
+      cat $CHUNK
```
nsoranzo: export PLANEMO_TEST_RUN="True"

should be moved here.

.travis.yml
```diff
@@ -26,11 +35,17 @@ script:
...
+      for DIR in `cat $CHUNK`; do planemo conda_install --conda_ensure_channels bioconda,r,iuc --conda_prefix $CONDA_PREFIX $DIR ; done
+      for DIR in `cat $CHUNK`; do planemo test --conda_dependency_resolution --conda_prefix $CONDA_PREFIX --galaxy_branch $GALAXY_RELEASE  --galaxy_source $GALAXY_REPO $DIR || exit 1; done
+    fi
 
 after_success:
   - |
```
nsoranzo: s/changed_repositories.list/$CHUNK/ in lines 54-55 below, otherwise multiple updates are tried.

PR #828, opened 2016-06-20
Stacks: various fixes

abretaud:
Several Stacks tools can print important things to the log, so we want to keep it too

### DIFF
--- a/packages/package_samtools_0_1_19/tool_dependencies.xml
+++ b/packages/package_samtools_0_1_19/tool_dependencies.xml
@@ -9,12 +9,18 @@
                         <source_directory>.</source_directory>
                         <destination_directory>$INSTALL_DIR</destination_directory>
                     </action>
+                    <action type="set_environment">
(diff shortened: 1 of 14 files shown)


nsoranzo:
Can you increase the version number of the modified tools?

bgruening:
@abretaud are these informations really worth a new history element? Can we use information from these datasets in other tools? Can we hide the extra output by default?

abretaud:
I just discussed it with @yvanlebras and it turns out the logs contain important metrics, so it's hard to do a complete stacks analysis without looking at it. And having it in the history makes it much simpler for users to have a look at it.
On the other hand it's right that we cannot use it in other tools...
Maybe we could add it as a dataset for now (and the gcc tutorial), and see a little later if some of theses logs can be hidden?

Anyway, I just pushed to this PR 2 little bug fixes

abretaud:
I've also pushed here fixes for several bugs we found while testing this on real data. There might be others tomorrow...

abretaud:
I think we're done with these bugfixes, is anyone available to review this?

abretaud:
Thanks! TSs updated

PR #834, opened 2016-06-21
Fix linking bam and bai files for bam_to_scidx tool

gregvonkuster:
@bgruening pointed out a flaw in how I was linking bam and bai files in this tool https://github.com/gregvonkuster/tools-iuc/tree/master/tools/pe_histogram, which I based on this bam_to_scidx tool.  So I made this fix as soon as I could.  Perhaps this will have a positive affect on this issue https://github.com/galaxyproject/galaxy/issues/2460 for this tool running on Galaxy main, but I'm not sure.

### DIFF
--- a/tools/bam_to_scidx/bam_to_scidx.xml
+++ b/tools/bam_to_scidx/bam_to_scidx.xml
@@ -13,8 +13,9 @@
     </stdio>
     <command>
         <![CDATA[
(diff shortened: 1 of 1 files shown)


gregvonkuster:
This fix has been uploaded to both the TTS and the MTS.

PR #849, opened 2016-06-25
JBrowse: Added option to use locally cached genomes as reference fasta

Slugger70:


tools/jbrowse/jbrowse.xml
```diff
@@ -216,13 +220,41 @@ $trackxml;
...
+        <when value="history">
+            <param 
+              format="fasta"
+              label="Select the reference genome"
+              name="genomes"
+              type="data"
+              multiple="True" />
+        </when>
+    </conditional>
+    <!-- <param label="Fasta Sequence(s)"
```
erasche: delete these lines

tools/jbrowse/jbrowse.xml
```diff
@@ -216,13 +220,41 @@ $trackxml;
...
+        <when value="indexed">
+            <param 
+              help="If your genome of interest is not listed, contact the Galaxy team"
+              label="Select a reference genome"
+              name="genomes"
+              type="select"
+              >
+                <options from_data_table="all_fasta">
+                    <filter column="2" type="sort_by" />
+                    <validator message="No genomes are available for the selected input dataset" type="no_options" />
```
erasche: Is it possible for this to be a multi-select? Is there value in that?

Slugger70: No, type=select is not iterable

bgruening:
@erasche  Ready to merge?

yhoogstrate:
I think some semicolons have to be replaced with `&&` according to our guidelines. Also some paths have to be escaped with quotes.

erasche:
@yhoogstrate paths here are all used inside XML so escaping shouldn't be necessary.

As for `;` -> `&` I'll get those in another commit, no reason for us to ask @Slugger70 do it.

erasche:
@yhoogstrate https://github.com/galaxyproject/tools-iuc/pull/864

bgruening:
Thanks a lot @Slugger70!

PR #858, opened 2016-06-28
Minor fixes for fimo version 4.11.0.3

gregvonkuster:
1. Change remove_duplicate_coords param from a boolean to a select list - works better in workflows
2. Add dest attribute to argparse arguments


gregvonkuster:
This fix is available in the TTS and MTS.

PR #865, opened 2016-07-01
Clarify filtering methods for fimo options

gregvonkuster:
This does not seem to affect the tool's behavior, but its a more ideal filter.


bgruening:
version bumb?

gregvonkuster:
This actually does not change the behavior of the tool, at least in all of the cases I've used it.  But I may be missing something obvious - I will bump the version if you feel I should.

bgruening:
Just double checking :)
You are more familiar than me with it :)
Thanks Greg!

PR #866, opened 2016-07-01
Enable meme 4.11.1 to find graphicsmagick at compile time

gregvonkuster:
Some meme features require it to be compiled using either ImageMagick or GraphicsMagick.  This version uses GraphicsMagick to be compatible with conda.


bgruening:
Thanks!

PR #877, opened 2016-07-12
Gemini morecols

TorHou:
More options to select columns from the variants table. I needed this to pick things like
`gt_ref_depths.sample1, gt_alt_depths.sample1`
which you can't add to the list of choices statically.

### DIFF
--- a/tools/gemini/gemini_comp_hets.xml
+++ b/tools/gemini/gemini_comp_hets.xml
@@ -12,7 +12,11 @@
         gemini @BINARY@
 
             #if $report.report_selector != 'all':
-                --columns "${report.columns}"
+                --columns "${report.columns}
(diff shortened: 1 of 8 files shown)


bgruening:
Can you activate testing and set $HOME to `.` as part of your tool? Does this work?

tools/gemini/gemini_comp_hets.xml
```diff
@@ -12,7 +12,11 @@
         gemini @BINARY@
 
             #if $report.report_selector != 'all':
-                --columns "${report.columns}"
+                --columns "${report.columns}
+                #if str($report.extra_cols).strip()
```
bgruening: This should probably be a macro.

bgruening:
Thanks Torsten!

PR #889, opened 2016-07-19
nerd_weather_tool

bimbam23:
simple nerd weather tool


bgruening:
@bimbam23 can you add such a file: https://github.com/galaxyproject/tools-iuc/blob/master/tools/art/.shed.yml

This is needed to trigger tests and to make an automatic upload to the TS possible.
Cou can simple commit and push to your branch it will automatically end up in this PR.

tools/weather_app/nerd_weather.xml
```diff
@@ -0,0 +1,80 @@
+<tool id="nerd_weather" name="NerdWeather" version="0.1.0">
+    <requirements>
```
bgruening: Can we rename this tool to sound a little bit more serious ;)

tools/weather_app/nerd_weather.xml
```diff
@@ -0,0 +1,80 @@
...
+        <requirement type="package" version="7.45.0">curl</requirement>
+        <requirement type="package" version="4.1.3">gawk</requirement>
+    </requirements>
+    <stdio>
+        <exit_code range="1:" />
+    </stdio>
+    <command><![CDATA[
+        curl http://wttr.in/$place$units 2>/dev/null  | sh $__tool_directory__/ansi2html.sh >> $out_file1;
```
bgruening: All inputs and outputs needs to be quoted. '$out_file1'.

bgruening:
This project is a result of the GCC Hackathon 2016 and I think it is a nice, small and fun tool for training and demonstration purpose.

tools/weather_app/nerd_weather.xml
```diff
@@ -0,0 +1,80 @@
...
+::
+ 
+    moon                    # Moon phase (add ,+US or ,+France for these cities)
+    moon@1999-Oct-02        # Moon phase on a particular date
+ 
+Units:
+------
+::
+ 
+    ?m                      # metric (SI) (used by default everywhere except US)
```
erasche: probably not necessary to expose the implementation detail to users.

tools/weather_app/nerd_weather.xml
bimbam23: Sure this was just a working title.

tools/weather_app/nerd_weather.xml
bimbam23: I added the .shed.yml. I hope its correct

tools/weather_app/.shed.yml
```diff
@@ -0,0 +1,11 @@
+# repository published to https://toolshed.g2.bx.psu.edu/repos/iuc/simple_weather
+owner: iuc
+name: SimpleWeather
+description: provides simple weather in ASCII code format
+long_description: |
+    Weather tool using
+    https://wttr.in
+categories:
+    - Weather
```
bgruening: Only categories from the TS are allowed: https://toolshed.g2.bx.psu.edu/
What about:
 * Visualization	
or
 * Web Services

bimbam23: we need to add a weather section... there will be more future tool ;-)

bgruening: lol ;)

bgruening:
`simple_weather[0]: passed` !!!

yhoogstrate:
I want to warn that such a tool has a slight security risc. It does a curl/wget and then some bash parsing to generate a html file within galaxy (originating from 3rd party). The wget occurs in every invocation of the tool. Currently the website looks reliable (except expired SSL certificate) but this is not something we can measure and we can't foresee the future. I know there's html sanitation but when the website gets compromised or changes owner it might start hosting advertisements/wares or work it's way around sanitation.

Please don't see this as a No because the implementation itself looks good but this is generating HTML files that originate from 3rd party HTML.

tools/weather_app/.shed.yml
```diff
@@ -0,0 +1,12 @@
+# repository published to https://toolshed.g2.bx.psu.edu/repos/iuc/simple_weather
+owner: iuc
+name: SimpleWeather
+description: provides simple weather in text format
+long_description: |
+    Weather tool using
+    https://wttr.in
```
bgruening: @bimbam23 maybe you can add @yhoogstrate concern here in the descriptions. That this tool should be treated carefully because it is using an external webservice.

erasche:
This really is a great example of the inverse relationship between PR size and volume of discussion ;) xref https://github.com/galaxyproject/galaxy/issues/522 which would be nice here but that's a galaxy missing feature.

bgruening:
I take this comment from @erasche as: `Merge it dude!` and so I will.

Thanks @bimbam23 and colleagues for this nice tool! Hope to see you here now more often :)

bimbam23:
Thanks,  it was a pleasure doing business with you!

PR #893, opened 2016-07-21
Set allow_duplicate_entries="False" for Tool Data Tables.

blankenberg:
Prevents duplicate entries from appearing in data table when they exist in the files. Prevents cases where a duplicate entry occurs and is then passed as a comma separated list onto the command line (which breaks most tools).

I just did all of the data tables here, if folks could review and check if any of these tables actually should allow duplicate entries, that would be a big help.

### DIFF
--- a/data_managers/data_manager_bowtie_index_builder/tool_data_table_conf.xml.sample
+++ b/data_managers/data_manager_bowtie_index_builder/tool_data_table_conf.xml.sample
@@ -1,16 +1,16 @@
 <tables>
     <!-- Locations of all fasta files under genome directory -->
-    <table name="all_fasta" comment_char="#">
+    <table name="all_fasta" comment_char="#" allow_duplicate_entries="False">
         <columns>value, dbkey, name, path</columns>
         <file path="tool-data/all_fasta.loc" />
     </table>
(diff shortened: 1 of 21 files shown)


bgruening:
```
snpEff[0]: failed
snpEff_databases[0]: failed
snpEff_download[0]: failed
```
I guess they should be added to the tool-testing blacklist or fixed.

yhoogstrate:
SnpEff is failing because it has issues independent of this PR. @blankenberg I don't know how duplicate entries are checked exactly, is this done on `dbkey` or on the entire entry? In case of the second, let's merge :+1:

mvdbeek:
@yhoogstrate It should be on all fields: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/data/__init__.py#L642

:+1:, already had that issue.
@blankenberg 
Just for my understanding: What's the point of allowing duplicate data table entries in galaxy?

blankenberg:
It checks across all fields. Restricting the check to optionally only check specific field(s) is a future enhancement.

Keeping duplicates is a backwards compatibility issue, there maybe a datatable that expects it, although I do not know of one off the top of my head.

mvdbeek:
Alright, thanks @blankenberg!

PR #894, opened 2016-07-25
Trinity 2.2.0

abretaud:
A new Trinity version + some other changes:
- a new abundance estimation method: salmon
- macros and tool suite to be more IUC-friendly
- one or two options that I needed to add
- 2 new tools for trinity scripts that I needed

I have added a tool dependency xml for this new 2.2.0 version, but I didn't create a new one for salmon, if anyone needs it tell me.

I *think* there may still be a failing test, but I'm not sure if it comes from my config or not, we'll see

PR #898, opened 2016-07-27
Proper names for trinity tools

abretaud:
Just corrections for tool names from #894

### DIFF
--- a/tools/trinity/.shed.yml
+++ b/tools/trinity/.shed.yml
@@ -10,7 +10,7 @@ owner: iuc
 remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/master/tools/trinity
 type: unrestricted
 auto_tool_repositories:
-  name_template: "{{ tool_id }}"
+  name_template: "trinity_{{ tool_id }}"
   description_template: "{{ tool_name }} (from the Trinity tool suite)"
 suite:
   name: "suite_trinity"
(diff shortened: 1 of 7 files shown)


tools/trinity/.shed.yml
```diff
@@ -10,7 +10,7 @@ owner: iuc
 remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/master/tools/trinity
 type: unrestricted
 auto_tool_repositories:
-  name_template: "{{ tool_id }}"
+  name_template: "trinity_{{ tool_id }}"
```
bgruening: If we change this than we will create a new repository in the TS and this will break the tool lineage.

martenson: yep, we cannot change those :/

abretaud: Oh yep, at least for trinity_trinity that would be created beside the trinity repo...
For the others, as it was not a suite before, they're new repos (though they were included in trinity repo before).

So what would be best?
- suite with prefix except for trinity repo
- suite with no prefix, just rename run_DE_analysis to run_de_analysis
- forget about the suite and keep all tools in the trinity repo as before
- something else?

nsoranzo: Apart from the `trinity` repo, the others have been created only on the TestTS because the Travis upload failed before starting the upload on MTS.

martenson: I am actually fine with option 1 given that these are not on MTS. How will you address the trinity tool that is already there though?

abretaud:
Ok, I changed to option 1, what do you think?

nsoranzo:
:+1:

bgruening:
This should work! Thanks!

nsoranzo:
Manually pushed to MTS because TTS was unreachable at some point during upload.

abretaud:
Thank you!

PR #913, opened 2016-08-19
Add UCSC twoBitToFa tool

bebatut:
Hi, 

twoBitToFa is a tool to convert all or part of .2bit file to fasta

Bérénice


tools/ucsc_tools/twobittofa/twobittofa.xml
```diff
@@ -0,0 +1,85 @@
...
+      <token name="@WRAPPER_VERSION@">332</token>
+    </macros>
+
+    <description>Convert all or part of .2bit file to fasta</description>
+
+    <requirements>
+      <requirement type="package" version="@WRAPPER_VERSION@">ucsc-twobittofa</requirement>
+    </requirements>
+
+    <version_command>@WRAPPER_VERSION@</version_command>
```
shiltemann: I think this would just try to run the command `332` and as a result probably generate an error string as version in the job info. There doesn't seem to be a version command in this tool so just leave it empty I guess

tools/ucsc_tools/twobittofa/twobittofa.xml
```diff
@@ -0,0 +1,85 @@
...
+
+            $noMask
+]]>
+    </command>
+
+    <inputs>
+        <param name="twobit_input" type="data" format="twobit"
+          label="Input TwoBit file" help=""/>
+
+        <param argument="-seq" type="text" label="Restrict this to just one sequence
```
shiltemann: you forgot to specify the `name` attribute in these (and following) params?

bgruening: Argument is kind of a new feature. It will create the name attribute from the argument attribute (stripping the dashes) and it will attach ('-seq') automatically to the help text. So less typing for us and better annotation of our parameters.

shiltemann: awesome! good to know :)

bgruening:
:+1: from me given the tests passing.
@shiltemann what do you think?

shiltemann:
yep looks good to me, thanks Bérénice :)

PR #914, opened 2016-08-25
Fix wrong option for trinity filter script

abretaud:
Just a little bugfix for the trinity filter script


tools/trinity/filter_low_expr_transcripts.xml
```diff
@@ -11,7 +11,7 @@
             --transcripts "$assembly"
 
             #if $additional_params.gene_map.has_gene_map == "no":
-                --gene_trans_map $additional_params.gene_map.gene_trans_map
+                --gene_to_trans_map $additional_params.gene_map.gene_trans_map
```
yhoogstrate: @abretaud would it make sense to add quotes to the path in case the path contains spaces? `--gene_to_trans_map "$additional_params.gene_map.gene_trans_map"`

yhoogstrate:
Hey Anthony, I have restarted Travis and now it seems to work. The fix seems appropriate so please merge and shed_update :+1: , thanks!

abretaud:
Thank you!

PR #917, opened 2016-08-26
Add dependency on expat version 2.1.0

gregvonkuster:
Attempt to correct this issue: https://github.com/galaxyproject/galaxy/issues/1913#issuecomment-242829812 - ping @natefoo

### DIFF
--- a/packages/package_perl_xml_parser_expat_2_41/tool_dependencies.xml
+++ b/packages/package_perl_xml_parser_expat_2_41/tool_dependencies.xml
@@ -1,5 +1,8 @@
 <?xml version="1.0"?>
 <tool_dependency>
+    <package name="expat" version="2.1.0">
+        <repository name="package_expat_2_1" owner="iuc" prior_installation_required="True" />
+    </package>
     <package name="perl" version="5.18.1">
         <repository name="package_perl_5_18" owner="iuc" prior_installation_required="True" />
     </package>
@@ -7,6 +10,9 @@
         <install version="1.0">
             <actions>
                 <action type="setup_perl_environment">
+                    <repository name="package_expat_2_1" owner="iuc">
+                        <package name="expat" version="2.1.0" />
+                    </repository>
                     <repository name="package_perl_5_18" owner="iuc">
                         <package name="perl" version="5.18.1" />
                     </repository>


jennaj:
(commenting so I'll get direct notifications on how this turns out @jennaj)

gregvonkuster:
It looks to me lime this failed test is unrelated to the change in this PR, so merging.  Hope this is ok.

gregvonkuster:
I've uploaded this change to the TTS, and it installs ok on OS X, but it fails on Linux with the following error.
```
Expat must be installed prior to building XML::Parser and I can't find
it in the standard library directories. Install 'expat-devel' package with your
OS package manager. See 'README'.

Or you can download expat from:

http://sourceforge.net/projects/expat/

If expat is installed, but in a non-standard directory, then use the
following options to Makefile.PL:

    EXPATLIBPATH=...  To set the directory in which to find libexpat

    EXPATINCPATH=...  To set the directory in which to find expat.h

For example:

    perl Makefile.PL EXPATLIBPATH=/home/me/lib EXPATINCPATH=/home/me/include

Note that if you build against a shareable library in a non-standard location
you may (on some platforms) also have to set your LD_LIBRARY_PATH environment
variable at run time for perl to find the library.

/usr/bin/ld: cannot find -lexpat
collect2: error: ld returned 1 exit status
make[1]: *** [../blib/arch/auto/XML/Parser/Expat/Expat.so] Error 1
make: *** [subdirs] Error 2
```
I've not had the opportunity to mess around with perl packages very much (and I am definitely not a perl expert), so I'm hoping for some help here.

The code in ~/lib/tool_shed/galaxy_install/tool_dependencies/recipe/step_handler.py for the setup_perl_environment action does not seem to take command line arguments like EXPATLIBPATH, but seems to restrict the command line arguments to just INSTALL_BASE:
```
if os.path.exists( os.path.join( tmp_work_dir, 'Makefile.PL' ) ):
    cmd += '''perl Makefile.PL INSTALL_BASE=$INSTALL_DIR && make && make install'''
```
What is the best way to correct this issue?

natefoo:
I don't know if there's a way to do this... any ideas @bgruening? Looking at XML::Parser::Expat's Makefile.PL it doesn't attempt to read these from the environment, only as command line arguments. Unless there is some facility to making Devel::CheckLib check alternate paths, but I don't see anything in its documentation.

Issue #916, opened 2016-08-26
stacks 1.40 fails to install


tiagoantao:
stacks fails to compile because include directories do not add -I/path/to/samtools

```
In file included from src/pstacks.h:64:0,
                 from src/pstacks.cc:25:
src/BamI.h:32:17: fatal error: bam.h: No such file or directory
 #include "bam.h"
                 ^
compilation terminated.
make[1]: *** [src/pstacks-pstacks.o] Error 1
make: *** [all] Error 2
```

I sorted this in a very dirty way by `apt-get install libbam-dev` and copying include files around.

But stacks needs to be configure to include .h from the samtools package

tiagoantao:
ping @abretaud @bgruening

yhoogstrate:
I think this line https://github.com/galaxyproject/tools-iuc/blob/master/packages/package_stacks_1_40/tool_dependencies.xml#L17 has to be changed to: `./configure --prefix=$INSTALL_DIR --enable-bam --includedir=$BAM_ROOT/include --libdir=${BAM_LIB_PATH}`

abretaud:
That's strange, looking at the code, the --with-bam-include-path option should be enough to avoid this error, did you test it or not?

yhoogstrate:
I did this because in 1.42 you will get: `configure: WARNING: unrecognized options: --with-bam-include-path, --with-bam-lib-path` and includedir and libdir also cover it. I think the actual problem was `${BAM_INCLUDE}` which is not part of the samtools 0.1.19 wrapper and should  be replaced with `$BAM_ROOT/include`. I did not test it as Galaxy tool, only via command line and it manages to configure both 1.40 and 1.42.

abretaud:
Ah ok, I hadn't seen the change of the include path
I'm going to make a PR, thanks!

yhoogstrate:
Thanks for reporting @tiagoantao :+1: ! And thanks for the fix @abretaud :+1: . Could you please confirm whether it works @tiagoantao and close the issue if it has been solved for you?

tiagoantao:
Sorry for the extreme delay. But I am teaching this week. I suggest closing this. And if I have the problem (I will revisit this in the future), I will get back here

PR #920, opened 2016-08-31
Fix for stacks compilation

abretaud:
A little fix for #916 thanks to @yhoogstrate 
At some point we should update to 1.42 too, but I don't have the time right now


yhoogstrate:
I don't know why conda is failing but it looks good to me and I don't see how this PR can make a test fail. I'll rerun the CI tests until we get green light and then we can merge.

PR #924, opened 2016-09-02
The twobitreader dependency was missing in the deepTools package

dpryan79:
Apparently we never updated the package dependencies to add twobitreader.


bgruening:
Thanks!

PR #926, opened 2016-09-05
SnpEff: solved conda incompatibilities and test-case

yhoogstrate:
This solves SnpEff in situations where conda is used - the wrapper depends on a shell variable that is tool dependencies specific. With some (ugly) bash scripting I was able to mimic this behavior and solve the issue.

Also the tests have been slightly changed as I couldn't reproduce the test result from the earlier XML file


yhoogstrate:
`realpath` is not available on travis VMs

PR #927, opened 2016-09-06
Use planemo to split repository list and find repos.

jmchilton:
The old method was very clever and very unix-y - the payoff isn't large for this change (doesn't eliminate much if any YAML) and I didn't dislike the previous approach but I feel it would be better to move to planemo so we can unify things (best practices) across repos more easily. This should eliminate the need for copying the script around, centralizes logic for determining what a shed repo is into planemo, and hopefully appears more natural for the novice tool developer (xaa seems more like magic).

Ultimately, I'd like to move toward a new file - maybe a .planemo_ci.yml file and hopefully Travis CI can be as simple as:

    planemo ci

... for these repos. This would hopefully let us be more CI independent and would hopefully enable testing the CI process outside of the CI server. It would also hopefully simplify the process of setting up new repos.


peterjc:
I like the motivation for this change, moving the Unix script magic into planemo :)

yhoogstrate:
Very elegant, thanks @jmchilton. Ready to merge according to me

bgruening:
I guess this one wait for me as I invented this bogus with xaa? Good for me! Thanks @jmchilton!

PR #929, opened 2016-09-07
update snpeff.py

nlapalu:
I am writing a tool to compile my own snpeffdb. The supported datatype is snpeffdbv4, so I need to get the version of snpeff in metadata for further analyses with snpeff in galaxy.

### DIFF
--- a/datatypes/snpeff_datatypes/snpeff.py
+++ b/datatypes/snpeff_datatypes/snpeff.py
@@ -57,6 +57,9 @@ class SnpEffDb( Text ):
                         # if snpEffectPredictor.bin download succeeded
                         genome_version = os.path.basename(root)
                         dataset.metadata.genome_version = genome_version
+                        snpeff_version = self.getSnpeffVersionFromFile('{}/{}/snpEffectPredictor.bin'.format(data_dir,genome_version))
+                        if snpeff_version:
+                            dataset.metadata.snpeff_version = snpeff_version
                     else:
                         m = re.match(regulation_pattern, fname)
                         if m:


yhoogstrate:
Dear @nlapalu, I think we can't merge this until the snpeffdbv4 format is defined in here: https://github.com/galaxyproject/galaxy/blob/dev/config/datatypes_conf.xml.sample

Is there an open PR for this?

nlapalu:
Dear @yhoogstrate, the format is defined in its own "datatype" repository, here:
https://github.com/galaxyproject/tools-iuc/blob/master/datatypes/snpeff_datatypes/datatypes_conf.xml

An update of the datatypes_conf.xml is required ?

yhoogstrate:
blame/ping @bgruening :)

bgruening:
I get always blamed ;-)

Youri is right here the best way is to put it into the galaxy main datatypes_conf.xml file. The datatypes stored here are just for historical reasons here and should be included upstream already.

Thanks for this contribution!

jj-umn:
Can you use the existing snpeffdb datatype from the galaxy distribution: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/datatypes/text.py   
It already has the version metadata.  It relies on the first line in snpEffectPredictor.bin having the SnpEff version:

$ gzcat tool-data/snpEff/v4_1/data/GRCm38.78/snpEffectPredictor.bin | head -n 3
SnpEff  4.1
CHROMOSOME      2       1       0       171031748       MG4214_PATCH    false
CHROMOSOME      3       1       0       42056   GL456367.1      false

yhoogstrate:
Hey @nlapalu,

according to the comment of @jj-umn (thanks for that!) you should proceed with something like this:

```
<command>
	#if $snpeff_db.metadata.snpeff_version == 'SnpEff4.0':
		echo "4.0!"
	#else
		print "not 4.0?"
		exit(1)
	#end if
</command>
...
<param format="snpeffdb" name="snpeff_db" type="data" />
...

```

If you want to test your tool, please update galaxy to at least this revision by @jj-umn:
https://github.com/galaxyproject/galaxy/commit/fb30e2b06103f3fbfa3dd1238c1f4e707d04b633
from (9 dec 2015). Also, try not to use a custom `snpeffdb4` format but the `snpeffdb` as provided by galaxy.

I am closing this pull request because I think the issue can be resolved without changes to IUC. I also think the snpeff data type definition(s) has to be removed from the IUC repo to avoid confusion. Thanks for getting here and taking action on this :+1: . If you have any more questions regarding this don't hezitate to reopen the issue.

all best,

Youri

PR #930, opened 2016-09-07
Update snpEff.xml

nlapalu:


yhoogstrate:
Dear @nlapalu,

could you try to get it working like this (and please report back whether you succeeded):
```
<param name="inp" type="data" format="snpeff">
    <options>
        <filter type="data_meta" ref="snpeff_version" key="SnpEff4.0"/>
    </options>
</param>
```
I did not test this myself but I think this might be the ideal way to solve this.

kind regards,

Youri

PR #932, opened 2016-09-07
Quast

Delphine-L:
New wrapper for Quast tool (Quality Assessment Tool for genome assemblies). A wrapper exists in the tool shed but this new one includes more options for the quality metrics including the possibility to add a genome reference file and annotation files. The wrapper also allows to get additional outputs like html report and contig size visualisation. 
Future development should be made to include more graphical output and additional features like gene discovery and variant detection.


tools/quast/.shed.yml
```diff
@@ -0,0 +1,13 @@
+categories: [Assembly]
+description: Quast
+long_description: QUAST (Quality ASsessment Tool) evaluates genome assemblies.
```
bgruening: can we move this to description and remove the long description here, it is at the bottom as well

tools/quast/quast.xml
```diff
@@ -0,0 +1,178 @@
+<tool id="quast" name="Quast" version="2.0.0" >
+	<description>QUAST (Quality ASsessment Tool) evaluates genome assemblies.</description>
```
bgruening: Can you please improve the code style a little bit.
 * 4 spaces indent 
 * less empty lines

tools/quast/quast.xml
```diff
@@ -0,0 +1,178 @@
...
+
+	<requirements>
+		<requirement type="package" version="4.1">quast</requirement>
```
bgruening: Quast already depends on matplotlib, so this is not required if you only target conda packages: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/quast/meta.yaml

tools/quast/quast.xml
```diff
@@ -0,0 +1,178 @@
...
+
+
+		#for $k in $files:
+			#if $k.type_file == "scaffold" :
+				-s
+			#end if
+      $k.input
+		#end for
+		;
+		cp outputdir/report.txt $quast_out;
```
bgruening: all paths and text inputs needs to be quoted, e.g. '$quast_out'

tools/quast/quast.xml
```diff
@@ -0,0 +1,178 @@
...
+		$names
+
+
+		#for $k in $files:
+			#if $k.type_file == "scaffold" :
+				-s
+			#end if
+      $k.input
+		#end for
+		;
```
bgruening: please use && to separate commands, otherwise the second command will run even if the first crashes

tools/quast/quast.xml
```diff
@@ -0,0 +1,178 @@
...
+            description="Something went wrong. Check the following errors : No labels..." />
+	</stdio>
+
+	<command>
+
+<![CDATA[	
+	 	quast
+		 -o outputdir
+
+		#if $gene_selection == "eukaryote" :
```
bgruening: space after eukaryote is not needed

tools/quast/quast.xml
```diff
@@ -0,0 +1,178 @@
...
+	</command>
+
+	<inputs>
+		<repeat name="files" title="Input assembly files" min="1">
+			<param type="data" format="fasta" name="input" label="Contigs/scaffolds output file"/>
+			<param name="type_file" type="select" label="Type of data">
+				<option value="contig">Contig</option>
+				<option value="scaffold">Scaffold</option>
+			</param>
+			<param name="label" type="text" label="Label the Dataset" value="Assembly File Label"/>
```
bgruening: How important is this? You could also use the history name of the dataset if you want.

tools/quast/quast.xml
```diff
@@ -0,0 +1,178 @@
...
+		</repeat>
+
+
+		<param name="input_size" type="integer" label="Size of reference genome(Optional)" optional="True" help="Estimated reference genome size (in bp) for computing NGx statistics. This value will be used only if a reference genome file is not specified "/>
```
bgruening: long lines can be broken into 2 or 3, so that the help in in the next line

tools/quast/quast.xml
```diff
@@ -0,0 +1,178 @@
...
+
+
+
+      <param name="input_ref" type="data" format="fasta"  label="Reference File(Optional)" optional="True" help="Many metrics can't be evaluated without a reference. If this is omitted, QUAST will only report the metrics that can be evaluated without a reference."/>
```
bgruening: Please either include the original parameter name of the tool in the help text, e.g. `help="foo bla (--argument)"`
Or use `param argument="--argument"` to specify the name of the parameter.

tools/quast/quast.xml
```diff
@@ -0,0 +1,178 @@
...
+      <param name="annot" type="data" format="txt" label="Gene Annotations (Optional)" optional="True" help="File with gene positions in the reference genome. "/>
```
bgruening: You don't need to label it as Optional, this should be obvious from the Galaxy UI

tools/quast/quast.xml
```diff
@@ -0,0 +1,178 @@
+<tool id="quast" name="Quast" version="2.0.0" >
```
bgruening: You could use the tool version 4.1.0 to reflect the underlying dependency.

tools/quast/quast.xml
```diff
@@ -0,0 +1,178 @@
...
+	+ An assessment summary in plain text format
+	+ An assessment summary in tabulation separated values format
+	+ An assessment summary in LateX format
+	+ An assessment summary in HTML format
+	+ An HTML view of contig sizes wit Icarus
+
+
+-----
+
+**Tool Information**
```
bgruening: I think this is too detailed for the target audience and the version should be indicated by the tool version, but this is just my opinion.

tools/quast/quast.xml
```diff
@@ -0,0 +1,178 @@
...
+
+- QUAST Tool v4.1:
+	+ Release Date: May 2016
+	+ URL: http://bioinf.spbau.ru/quast
+	+ Source URL: https://downloads.sourceforge.net/project/quast/quast-4.1.tar.gz
+
+-----
+
+**QUAST Galaxy Wrapper Author and Contact Information**
+
```
bgruening: Can we move this in the citation section or in a readme?

bgruening:
This is a great first version. Nice work @Delphine-L! I did a small review with some comments, but overall it looks good.

bgruening:
Also the test html files are quite large could we use the md5 checksum to compare them, or the `contains` test method?

tools/quast/.shed.yml
```diff
@@ -0,0 +1,13 @@
...
+remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/master/tools/quast
+name: quast
+owner: iuc
+long_description: QUAST evaluates genome assemblies by computing various metrics, including 
+    - N50, length for which the collection of all contigs of that length or longer covers at least 50% of assembly length,
+    - NG50, where length of the reference genome is being covered,
+    - NA50 and NGA50, where aligned blocks instead of contigs are taken,
+    - misassemblies, misassembled and unaligned contigs or contigs bases,
+    - genes and operons covered.
+    http://quast.bioinf.spbau.ru/
```
martenson: please use `homepage_url: http://quast.bioinf.spbau.ru/` instead of including it in the description

tools/quast/quast.xml
martenson: @bgruening What if in future quast won't depend on matplotlib?

bgruening: A future version will have a new version number and no dependency on matplotlib ... so I think this is save.

Delphine-L:
Thank you @bgruening  and @martenson for your useful comments ! I tried to change the wrapper accordingly, I hope the new version will be more suitable.

Delphine-L:
We rebased the branch in order to suppress the heavy html files from the repository

martenson:
👍

tools/quast/README.txt
```diff
@@ -0,0 +1,22 @@
+**Tool Information**
```
bgruening: Should this be a README.md file?

tools/quast/quast.xml
```diff
@@ -0,0 +1,129 @@
...
+        --min-contig $min_contig
+        --contig-thresholds $threshold_contig
+        #for $k in $files:
+            #if $k.type_file == "scaffold":
+                -s
+            #end if
+        $k.input
+        #end for
+        &&
+        cp outputdir/report.txt '$quast_out';
```
bgruening: please also use `&&` here and below.

bgruening: please also use `mv`. If the `tmp` dir is on the same partition as the Galaxy store, `mv` should be much faster.

bgruening:
@Delphine-L very nice! Just these two small changes and we are ready for a merge!

tools/quast/quast.xml
```diff
@@ -39,11 +39,11 @@
...
-        cp outputdir/report.txt '$quast_out';
-        cp outputdir/report.tsv '$quast_tsv';
-        cp outputdir/report.tex '$quast_tex';
-        cp outputdir/icarus_viewers/contig_size_viewer.html '$icarus';
-        cp outputdir/report.html '$report_html';
+        mv outputdir/report.txt '$quast_out' &&
+        mv outputdir/report.tsv '$quast_tsv' &&
+        mv outputdir/report.tex '$quast_tex' &&
+        mv outputdir/icarus_viewers/contig_size_viewer.html '$icarus' &&
+        mv outputdir/report.html '$report_html' &&
```
bgruening: The last one should be removed. A command should not end with a `&&` or `;`

bgruening:
And tests pass again! Thanks so much!

PR #935, opened 2016-09-14
Add bwameth and its requirements

dpryan79:
This adds "bwa-meth" from @brentp as a tool. This is a bisulfite aligner similar to bismark, but is generally a bit faster (and at least used to produce better quality output). I have a copy of this in the testtoolshed that seems to be working properly when I install it in a clean dockerized Galaxy environment. Feedback would be quite welcome on this (also, let me know if you want this in the IUC repo at all).


bgruening:
Great! This is a great addition. More detailed review tonight. Can you change the namespaces in the packages? I'm not sure about the toolshed name :-) Is it also called toolshed in conda? 
Thanks Devon!

dpryan79:
I can change whatever you like. The "toolshed" package has the same name already in the bioconda repo.

mvdbeek:
Also the test-data is pretty big, and could we have a bwameth package? Looks like bwameth is pip-installable, so I would even suggest dropping the toolshed package and instead installing bwameth including the toolshed dependency.

dpryan79:
@mvdbeek: Sure, not a problem.

dpryan79:
I'll make a new PR once I have all of that done.

PR #936, opened 2016-09-14
Datamanager bowtie update

mvdbeek:
This uses bowtie 1.1.2, which is also available in conda.


bgruening:
@mvdbeek the database is compatible with the older versions? Aka we can leave the old data_table entry?

mvdbeek:
That is my understanding, but I will check this.

mvdbeek:
Yes, all is well. Indexes generated with bowtie-1.1.2 continue to work with 0.12.7 (both colorspace and regular fastqsanger), and 0.12.7 indexes work with bowtie-1.1.2.

mvdbeek:
@bgruening ready?

bgruening:
Oh yes, thanks for checking!

PR #938, opened 2016-09-14
Upload packages on-merge

bgruening:
I would like to upload packages after merging PR to master. This needs changes to planemo and the packages needs to be one the list of to-be-uploaded-repos. But this can cause troubles. First packages needs to be uploaded before tools and ultimately the packages needs to be sorted according to a DAG of there dependencies.

Anyone a more pragmatic idea to solve this?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ install:
   - echo $TRAVIS_COMMIT_RANGE
   - |
       planemo ci_find_repos --exclude_from .tt_blacklist \
-                            --exclude packages --exclude data_managers \
+                            --exclude data_managers \
                             --changed_in_commit_range $TRAVIS_COMMIT_RANGE \
                             --chunk_count 4 --chunk ${CHUNK} \
                             --output changed_repositories_chunk.list


mvdbeek:
> Anyone a more pragmatic idea to solve this?

Let's kill 'em already :).
In all seriousness though, how about we have 2 labels, "Uploaded to TS" and "Uploaded to MTS", and have a bot comment if those labels are not set. The merger is then responsible to upload the packages and set the labels. It's not like we have a huge traffic of package updates.

bgruening:
@mvdbeek you lost me :) How can labels help to figure out the correct order of uploading packages/tools?

mvdbeek:
It can't, it'll just make sure we don't forget uploading. I was thinking it's probably not worth adding new, intricate functionality to planemo for uploading packages, which we want to phase out. But maybe this is easier than I think.

bgruening:
I agree it is probably more work than it's worth and as soon as travis fails during merge we upload packages manually.

nsoranzo:
1. What about data managers?

2. `planemo conda_install` and `planemo test` should not be run for packages (and data managers)

bgruening:
Ok, closing and hope that this will be less of a problem in the future.
For data_managers I would really like to be able to test these with planemo. @jmchilton Is this possible?

jmchilton:
Data managers - someday.

PR #933, opened 2016-09-08
Standing PR with fixes detected by XSD.

jmchilton:

PR #939, opened 2016-09-15
fixes detected by XSD - based on John PR

bgruening:
This is a replacement for https://github.com/galaxyproject/tools-iuc/pull/933


bgruening:
Ready for review!

tools/art/macros.xml
```diff
@@ -63,7 +63,7 @@
 
     <xml name="frag_len_sd">
         <param label="the average DNA fragment size"
-            name="fragment_size" type="integer" value="200" minimum="101" />
+            name="fragment_size" type="integer" value="200" />
```
mvdbeek: We do have a min="" setting for type="integer", right?

tools/bcftools/bcftools_view.xml
```diff
@@ -116,9 +116,9 @@ ${section.header}
                    help="Trim alternate alleles not seen in the subset" />
         </section>
         <section name="sec_filter" expanded="false" title="Filter Options">
-            <param name="min_ac" type="integer" label="Min Ac" default="nref" optional="True" 
+            <param name="min_ac" type="integer" label="Min Ac" optional="True"
```
mvdbeek: I'm afraid I don't understand the parameter, you set one integer for all of nref/alt1/minor etc .. ?

tools/biom_format/biom_convert.xml
```diff
@@ -69,7 +69,7 @@
             <when value="tsv">
                 <param name="header_key" type="text" value="" label="observation metadata to include" help="Leave empty to ignore" argument="--header-key"/>
                 <param name="output_metadata_id" type="text" value="" label="name to be given to the observation metadata column" help="Leave empty to ignore" argument="--output-metadata-id"/>
-                <param name="tsv_metadata_formatter" type="select" label="Method for formatting the observation metadata" argumet="--tsv-metadata-formatter">
+                <param name="tsv_metadata_formatter" type="select" label="Method for formatting the observation metadata" argument="--tsv-metadata-formatter">
```
mvdbeek: This changes the result on re-run, bump version ?

tools/mothur/pairwise.seqs.xml
```diff
@@ -50,7 +50,7 @@
         <param name="cutoff" type="float" value="0.0" min="0" label="cutoff - Distance Cutoff threshold - ignored if not > 0" help="Do not save any distances larger than this, a common value would be 0.10"/>
         <conditional name="scoring">
             <param name="setby" type="select" label="Scoring match, mismatch, gapopen, and gapextend" help="">
-                <option value="default" selecte="true">Use default settings</option>
+                <option value="default" selected="true">Use default settings</option>
```
mvdbeek: Bump version?

tools/vsearch/clustering.xml
```diff
@@ -70,7 +70,7 @@
         <param name="infile" type="data" format="fasta" label="Select your input FASTA file" help="" />
         <conditional name="clustering_mode">
             <param name="clustering_mode_select" type="select" label="Choose sorting method to use before clustering" help="">
-                <option value="cluster_fast" default="True">Cluster sequences after sorting by length (--cluster-fast)</option>
+                <option value="cluster_fast" selected="True">Cluster sequences after sorting by length (--cluster-fast)</option>
```
mvdbeek: Bump version?

tools/art/macros.xml
bgruening: if you set min, than the test does not pass and I have not found any reference for a real min in art.

mvdbeek: That's OK for me then.

tools/biom_format/biom_convert.xml
bgruening: It should only change the help text afaik.

tools/mothur/pairwise.seqs.xml
bgruening: It's the first one in the select, so was always selected.

tools/vsearch/clustering.xml
bgruening: It is the first one in select so was selected by default.

tools/bcftools/bcftools_view.xml
bgruening: Not sure what this is. If I look at the readme I think a text field would fit?
```
-C, --max-ac INT[:nref|:alt1|:minor|:'major'|:'nonmajor']
maximum allele count (INFO/AC) of sites to be printed. Specifying the type of allele is optional and can be set to non-reference (nref, the default), 1st alternate (alt1), the least frequent (minor), the most frequent (major) or sum of all but the most frequent (nonmajor) alleles.
```

bgruening:
So mothur is timing out more or less. I once tried to split the testing of large repositories into different chunks but this yields to other problems, like if one chuck is finished successfully it but another is failing the upload will be triggered. Any idea here?

I hope mothur is the only repository with that many tests and tools.

TorHou:
Hi Björn,

wir könnten uns nachher mal zusammen setzen und du könntest mir sagen, 
wie Du das Diagramm lieber hättest. Dann kann ich das in Inkscape 
reinhacken und fertig ist die Laube.

Grüße,
Torsten


Björn Grüning wrote:

tools/bcftools/bcftools_view.xml
mvdbeek: So I could do "-C 2:nref 3:alt1"? Or "-C 2:nref" "-C 3:alt1". Or I can only set a global maximum applied to one of nref, alt1 etc?
Based on this I'm still confused :/.
I guess the initial intentent was to simply use <INT:nref> ?

mvdbeek:
> Any idea here?

Is it this https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received ?
In that case we could prefix with `travis_wait`. Else move the mothur fixes in a separate PR and use `no_ci` in the commit message (assuming this tests fine locally)?

bgruening:
It is the too much output generated. Planemo is way to verbose for this setup.

mvdbeek:
> It is the too much output generated. Planemo is way to verbose for this setup.

But that's not really a problem. After all, the push PR passed fine. I would guess that it just doesn't *always* work. As far as I can tell, I saw a segfault in the bcftools in https://travis-ci.org/galaxyproject/tools-iuc/jobs/160104674.

peterjc:
My changes were done with sed - which as a side effect put in any missing trailing newline at the end of some files, which I committed separately.

jmchilton:
@peterjc Isn't the default for optional different based on the parameter type? Does your commit account for that?

jmchilton:
For instance, multi-select boxes are optional by default by single select boxes aren't. I'd argue for being explicit given this.

peterjc:
Interesting point @jmchilton - that means https://wiki.galaxyproject.org/Admin/Tools/ToolConfigSyntax#A.3Cparam.3E_tag_set is wrong about the ``optional=`` default, and I may need to review the similar changes I've made in my tools.

In any case, am I right that currently ``required=...`` is non-functional?

Should I rewrite branch history to drop that commit, and replace with ``optional="true"`` with an explicit ``required="false"`` instead?

nsoranzo:
Also for types `data`, `float` and `integer`, `optional="true"` is not redundant.
Never heard of a `required` attribute.

jmchilton:
I reviewed basic.py and parser/xml.py and I can't find any reference to a ``required`` parameter - I think it is non-functional.

peterjc:
I believe all cases of ``required=...`` are human error (some mine), and ought to have been written using the inverse ``optional=..``.

I believe Galaxy currently ignores the ``<param>`` tags' ``required=...`` entries so removing them is fine and will have no effect.

The question is how many of these ought to be using ``optional=...`` instead (given its has complex default rules).

jmchilton:
https://github.com/galaxyproject/tools-iuc/pull/939/commits/68bd66f7e9491071e9837bd47229cc0f2eaaa8d1 doesn't seem to replace any required options. Are you referring to other commits in this PR or commits on your own branches? I guess it should be fine to just swap the "required" with an explicit optional of the opposite boolean, because the author intention was that and I think all of the parameters acception ``optional``.

peterjc:
I'm confusing myself with required/optional. Will remove that commit...

nsoranzo:
@mvdbeek I think this is ready for merge, ok for you too?

mvdbeek:
Yes, it's OK !

jmchilton:
Awesome sauce - thanks all!

bgruening:
This is getting better and better! Thanks a lot everyone!

PR #941, opened 2016-09-15
Bwameth datamanager

dpryan79:
…es data table to include dbkey

@bgruening: I took a stab at making a data manager for bwameth. I have a copy in the testtoolshed that seems to work, but it's unclear to me what's "best practice" for these things. For whatever it's worth, I largely copied the hisat2 data manager.

Note that this also adds a `dbkey` column to the bwameth_indexes table. That seems to be more conformant with what hisat2 and other tools are doing.


bgruening:
@blankenberg can you please look at this? 
Dan Is there any way we can test data managers with planemo?

bgruening:
@dpryan79 your yaml file is not properly formatted. It the description is not correctly intended.

dpryan79:
@bgruening Thanks, that's fixed now.

bgruening:
@galaxyproject/iuc anyone? I think this one is ready to go.

bgruening:
Thanks a lot @dpryan79!!!

PR #953, opened 2016-09-21
HTSeq-Count: Added more quotes, support for meta-data and fixes test

yhoogstrate:
@lparsons I think we got rid of the FIX-ME part


tools/htseq_count/htseq-count.xml
```diff
@@ -279,6 +289,7 @@ Public License v3. Part of the 'HTSeq' framework.
     <citations>
         <citation type="bibtex">
```
nsoranzo: Why not just use:
```xml
<citation type="doi">10.1093/bioinformatics/btu638</citation>
```

PR #958, opened 2016-09-22
DrawRNAjs wrapper

yhoogstrate:
Hey everyone,

This PR is a bit different than usual. DrawRNAjs by @bene200 is a Javascript visualization for RNA 2D structure. Therefore I have the dependencies (css and js files) not managed by conda but dropped them in the static folder of the tool wrapper. Please review this carefully as this is a bit different way of doing it and might initiate a new direction for other contributions sa well.
The reason I built it as a Galaxy tool rather is because I really like the type of visualization and having it in a tool shed gives it much more visibility than only having it in @bgruening visualizations directory at github.

A screenshot:
image

Thanks for looking into this,

Youri


bgruening:
This is a really nice hack! I like the functionality - I'm a little bit scared how this works, but well done!
@guerler what do you think about this? I tend to merge it, as I think getting Galaxy VIS into workflows will take a few month/years right? So this would be a nice in the meantime ...

tools/drawrnajs/drawrnajs.xml
```diff
@@ -0,0 +1,108 @@
+<tool id="drawrnajs" name="DrawRNAjs" version="1.2.6-0">
+    <description>Draw RNA 2D Structures</description>
+
+    <command><![CDATA[
+        cp '$html_template' '$output' ;
+        sed -i.bak 's@__UID_PLACEHOLDER__@${__app__.security.encode_id($input.id)}@g' '$output' ;
+        
+        mkdir -p '${output.files_path}' ;
+            
+        ln -s '$__tool_directory__/static/bootstrap.css' '${output.files_path}/' ;
```
mvdbeek: Wouldn't it be better to copy this over?. If the tool is uninstalled/updated, or the dependency folder moves the visualization will stop working, right?

tools/drawrnajs/drawrnajs.xml
```diff
@@ -0,0 +1,108 @@
...
+        ln -s '$__tool_directory__/static/spectrum.css' '${output.files_path}/' ;
+        ln -s '$__tool_directory__/static/style.css' '${output.files_path}/' ;
+        ln -s '$__tool_directory__/static/drawrnajs@1.2.6' '${output.files_path}/drawrnajs.js' ;
+        ln -s '$__tool_directory__/static/galaxy_handshake_utils.js' '${output.files_path}/galaxy_handshake_utils.js' ;
```
jmchilton: Can you copy these instead? There is no guarantee these directories will remain on the same locations back on the Galaxy web server. Unless Galaxy resolves the symbolic link for you - but I don't think it does.

jmchilton:
This isn't best practice - it shouldn't be considered a best practice. I'm fine with you developing tools this way - but can you move it to a different repository? This is supposed to be examples of best practices.

There is so much static JS there that needs to live with the tool forever - if there are security problems down the road it is a mess for admins to deal with that - and in fact you have no way of conveying it to the the admin.

bgruening:
Oh damn it, I thought this PR was against `galaxytools`. Sorry I need to correct me, I would be fine to merge this into `galaxytools` not here. @jmchilton is completely correct this should not be part of IUC.

guerler:
This is purely js-based right? I dont want to hold this PR up, but I think it would be a great candidate to be added as charts plugin.

bgruening:
@guerler thought so. I will start with all the plugin in `galaxytools`. But the use-case here is that it would be nice to run these tools in the end of a workflow. Is this planned or already possible?

guerler:
Not yet, but its something I can work on if there is enough interest/priority.

PR #962, opened 2016-09-23
Ensure have new line at end of XML files

peterjc:
Whitespace change only, something sed does anyway if making blanket changes to multiple files.


peterjc:
The TravisCI build on my account failed https://travis-ci.org/peterjc/tools-iuc/builds/162233013 because ``$TRAVIS_COMMIT_RANGE`` was empty - might only exist for pull requests?

In any case, seems safer to quote all the environment variables.

nsoranzo:
Thanks @peterjc!

PR #963, opened 2016-09-26
Adopting package_gmap_2013_05_09

peterjc:
As per email discussion with @jj-umn, see https://lists.galaxyproject.org/pipermail/galaxy-dev/2016-September/024963.html

This pull request ports the original commit https://github.com/jj-umn/galaxytools/commit/fff24235f2fcb299330e34d04c03e1d931606f8d for https://toolshed.g2.bx.psu.edu/view/jjohnson/package_gmap_2013_05_09/953f5eb53593 (the IUC account should have write access to this now)

### DIFF
--- a/packages/package_gmap_2013_05_09/tool_dependencies.xml
+++ b/packages/package_gmap_2013_05_09/tool_dependencies.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0"?>
+<tool_dependency>
+    <package name="gmap" version="2013-05-09">
+        <install version="1.0">
+            <actions>
+                <action type="download_by_url" target_filename="gmap-2013-05-09.tar.gz">http://research-pub.gene.com/gmap/src/gmap-gsnap-2013-05-09.tar.gz</action>
+                <action type="shell_command"> ./configure --prefix=$INSTALL_DIR --with-gmapdb=$INSTALL_DIR/gmapdb</action>
+                <action type="shell_command">make</action>
+                <action type="shell_command">make install</action>
+                <action type="set_environment">
+                    <environment_variable name="PATH" action="prepend_to">$INSTALL_DIR/bin</environment_variable>
+                </action>
+            </actions>
+        </install>
+        <readme>
+        </readme>
+    </package>
+</tool_dependency>
--- a/packages/package_gmap_2013_05_09/tool_dependencies.xml
+++ b/packages/package_gmap_2013_05_09/tool_dependencies.xml
@@ -3,7 +3,7 @@
     <package name="gmap" version="2013-05-09">
         <install version="1.0">
             <actions>
-                <action type="download_by_url" target_filename="gmap-2013-05-09.tar.gz">http://research-pub.gene.com/gmap/src/gmap-gsnap-2013-05-09.tar.gz</action>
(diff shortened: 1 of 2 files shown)


yhoogstrate:
The version is quite out of date and does not match with the bio-conda version of gmap (2015.09.10). I can also imagine certain tools depend on older versions of gmap/gsnap (defuse?). Do you have any ideas or suggestions about this @jj-umn? If there is no reason to hold back, what about updating to a more up to date version?

bgruening:
:+1: for updating it before inclusion.

nsoranzo:
Or just go conda-only?

yhoogstrate:
I also see that ` --with-gmapdb=$INSTALL_DIR/gmapdb` is not used in the bio-conda build. This would translate to: ` --with-gmapdb=$INSTALL_DIR/share`. Similarly I see in the bio-conda build `env MAX_READLENGTH=500 `. These settings have to be identical if we want to support both. I think @nsoranzo's suggestion is the most appropriate way to deal with this.

peterjc:
I want to preserve this old release (2013/05/09) as a first step, and look at packaging a more recent version later - such as the version already in bio-conda.

I'll be preparing a second pull request for the tool wrapper next.

peterjc:
Thinking about this, since the current wrapper on the Tool Shed has been broken for some time (following what looks like an accidental bad upload, see IUC mailing list thread), I'm OK with going straight to a bio-conda packaged version.

packages/package_gmap_2013_05_09/tool_dependencies.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0"?>
+<tool_dependency>
+    <package name="gmap" version="2013-05-09">
+        <install version="1.0">
+            <actions>
+                <action type="download_by_url" target_filename="gmap-2013-05-09.tar.gz">http://research-pub.gene.com/gmap/src/gmap-gsnap-2013-05-09.tar.gz</action>
```
yhoogstrate: checksum

peterjc:
Is there any reason to convert ``package_gmap_2013_05_09`` to use conda?

(I agree we should try that for packaging the latest GMAP)

nsoranzo:
Since this is perfect moment to create a clean new version, my preference would be to just go with the latest GMAP version and Conda-only.
@peterjc Do you know if your revised wrapper work with latest GMAP?
If yes, we could:
- drop this TS package (no reason to distribute a 3-years old software)
- update the Bioconda package to latest release (now it is 2015.09.10). Either @bgruening or I can do this
- update the requirement on the wrappers.

peterjc:
I don't know if the current wrapper will work as is with the latest release or the latest conda release (2015.09.10), but JJ did note when he last worked on this that the command line API was changing quite often.

I want to first recreate what JJ was working on, and get it working and installable from the Test Tool Shed (and ideally checked into ``tools-iuc``).

yhoogstrate:
@peterjc If you decide to continue with a tool_dependencies version of this package, would you mind to merge it with the large gmap PR?

peterjc:
OK, I'll do that - see PR #964

PR #964, opened 2016-09-26
Adopting JJ's gmap wrappers

peterjc:
As per email discussion with @jj-umn, see https://lists.galaxyproject.org/pipermail/galaxy-dev/2016-September/024963.html

As far as I could see, this is not included under https://github.com/jj-umn/galaxytools therefore this captures the last proper upload to the ToolShed as a back-dated commit under JJ's GitHub email address, https://toolshed.g2.bx.psu.edu/view/jjohnson/gmap/93911bac43da _Modifications for ToolShed proprietary data types_ from 2012-01-15.

See also #963 Adopting package_gmap_2013_05_09


peterjc:
TravisCI is failing, first with flake8 on the datatype:

```
... flake8 --exclude=.git .
./tools/gmap/lib/galaxy/datatypes/gmap.py:8:1: F401 'galaxy.datatypes.data' imported but unused
./tools/gmap/lib/galaxy/datatypes/gmap.py:270:28: F821 undefined name 'data'
./tools/gmap/lib/galaxy/datatypes/gmap.py:274:33: F821 undefined name 'data'
```

Is this meant to be:

``` python
from galaxy.datatypes import data
```

tools/gmap/lib/galaxy/datatypes/gmap.py
```diff
@@ -0,0 +1,487 @@
...
+        dataset.metadata.data_lines = data_lines
+        dataset.metadata.annotations = annotations
+
+    def set_peek(self, dataset, is_multi_byte=False):
+        if not dataset.dataset.purged:
+            dataset.peek = data.get_file_peek(dataset.file_name, is_multi_byte=is_multi_byte)
+            if dataset.metadata.annotations:
+                dataset.blurb = "%s annotations" % util.commaify(str(dataset.metadata.annotations))
+            else:
+                dataset.blurb = data.nice_size(dataset.get_size())
```
nsoranzo: `nice_size` is now defined in `galaxy.util` .

peterjc: If you want to fix this now, please go ahead - I'm going home in a moment and won't touch this until tomorrow at the earliest.

nsoranzo:
> Is this meant to be:
>
> ``` python
> from galaxy.datatypes import data
> ```

@peterjc Yes, but see my other comment about `nice_size()`. Anyway, I really think the datatype should be added to Galaxy repository, not here.

peterjc:
If we can get it working again, sure, we can submit the GMAP datatypes to the Galaxy core.

peterjc:
The next problem from the GMAP tools will be tool linting with XSD validation, multiple cases of:

```
Element 'param', attribute 'metadata_name': The attribute 'metadata_name' is not allowed.
```

This might be a planemo/XSD bug as there look to be examples of ``metadata_name`` in https://docs.galaxyproject.org/en/latest/dev/schema.html although it is only described as an attribute of the ``<validator>`` element?

CC @martenson & @jmchilton

jmchilton:
That metadata name thing should not be on ``param``s, I double checked this as part of the IUC clean up. It doesn't do anything - it is a viral piece of non-coding XML that has propagated through copy-paste I believe. There is an example of this in an XSD doc example, but it is a code fragment that wouldn't validate.

martenson:
>  There is an example of this in an XSD doc example, but it is a code fragment that wouldn't validate.

I am not sure I understand, should we remove it then?

jmchilton:
@martenson Yes - we should remove it. https://github.com/galaxyproject/galaxy/issues/2974 does exactly that.

peterjc:
OK, the next big issue from ``planemo shed_lint --tools --xsd ...`` is the lack of any tests.

And there no explicit dependency declaration (it depends on ``gmap``, ``gsnap``, ``iit_store`` and ``snpindex`` being on ``$PATH``), so I am only assuming that it was used with https://toolshed.g2.bx.psu.edu/view/jjohnson/package_gmap_2013_05_09/ - the ``README`` file says it was consistent with gmap version 2011-11-30.

If I have to put together some test data, perhaps I might as well try this with the latest version of gmap in bioconda?

Also there are no ``<stdio>`` tags.

peterjc:
I've just realised there are two uploads by JJ to the Test Tool Shed, both later than the final good version on the main Tool Shed - so I will try to incorporate that work next:

 https://testtoolshed.g2.bx.psu.edu/view/jjohnson/gmap

This work includes providing ``tool-data/datatypes_conf.xml`` (which I just recreated myself) and ``tool-data/gmap_indices.loc.sample`` (which I was planning to create).

peterjc:
@nsoranzo @frederikcoppens @yhoogstrate etc - Would you object to me updating JJ's existing *Test* Tool Shed entry for GMAP with this work? This would facilitate my testing locally. i.e. https://testtoolshed.g2.bx.psu.edu/view/jjohnson/gmap

(I'd update the version numbers and README file first)

nsoranzo:
> I've just realised there are two uploads by JJ to the Test Tool Shed, both later than the final good version on the main Tool Shed

@peterjc In my first reply on the "GMAP tool_dependencies.xml package for ToolShed" thread on the `galaxy-iuc` mailing list, I attached a tar.gz archive with the latest code developed by JJ. I'm not sure if this is different from the one on the TTS, but take a look if you have time.

bgruening:
@peterjc test is for testing, please use it as you like :)

peterjc:
@nsoranzo thanks - I'd missed that attachment.

The latest version from the (Test) Tool Shed was wrappers at version 2.0.0 and 2.0.1,

```
$ cd  ~/Downloads/gmap-74391fc6e3f2/
$ grep "version=" *.xml
gmap.xml:<tool id="gmap" name="GMAP" version="2.0.1">
gmap.xml:    <requirement type="package" version="2011-11-30">gmap</requirement>
gmap_build.xml:<tool id="gmap_build" name="GMAP Build" version="2.0.0">
gmap_build.xml:      <requirement type="package" version="2011-11-30">gmap</requirement>
gsnap.xml:<tool id="gsnap" name="GSNAP" version="2.0.1">
gsnap.xml:      <requirement type="package" version="2011-11-30">gmap</requirement>
iit_store.xml:<tool id="gmap_iit_store" name="GMAP IIT" version="2.0.0">
iit_store.xml:      <requirement type="package" version="2011-11-30">gmap</requirement>
snpindex.xml:<tool id="gmap_snpindex" name="GMAP SNP Index" version="2.0.0">
snpindex.xml:      <requirement type="package" version="2011-11-30">gmap</requirement>
tool_dependencies.xml:<?xml version="1.0"?>
tool_dependencies.xml:    <package name="gmap" version="2011-11-30">
tool_dependencies.xml:        <install version="1.0">
```

The version you emailed from JJ has wrapper versions 3.0.0 instead:

```
$ cd ~/Downloads/gmap
$ grep "version=" *.xml
datatypes_conf.xml:<?xml version="1.0"?>
gmap.xml:<tool id="gmap" name="GMAP" version="3.0.0">
gmap.xml:    <requirement type="package" version="2013-05-09">gmap</requirement>
gmap_build.xml:<tool id="gmap_build" name="GMAP Build" version="3.0.0">
gmap_build.xml:      <requirement type="package" version="2013-05-09">gmap</requirement>
gsnap.xml:<tool id="gsnap" name="GSNAP" version="3.0.0">
gsnap.xml:      <requirement type="package" version="2013-05-09">gmap</requirement>
iit_store.xml:<tool id="gmap_iit_store" name="GMAP IIT" version="3.0.0">
iit_store.xml:      <requirement type="package" version="2013-05-09">gmap</requirement>
snpindex.xml:<tool id="gmap_snpindex" name="GMAP SNP Index" version="3.0.0">
snpindex.xml:      <requirement type="package" version="2013-05-09">gmap</requirement>
tool_dependencies.xml:<?xml version="1.0"?>
tool_dependencies.xml:    <package name="gmap" version="2013-05-09">
```

This  also explains why ``package_gmap_2013_05_09`` exists, see #963.

I'll update the branch again...

peterjc:
OK, thanks to Nicola for pointing me at v3.0.0 of the wrappers received from JJ by email, that's included in the revised history.

Of note this adds two test files, ``GRCh37.69.cdna.19.fa`` (2.5Kb file, two cDNA sequences ENST00000269812 and ENST00000434325) and ``Homo_sapiens.GRCh37.71.dna.chromosome.19.fa`` (0.5Mb, perhaps a bit large), but no actual test cases.

I was able to rebase some of my tidying work easily, other steps like moving the ``<configfiles>`` or PEP8 work I simply repeated.

yhoogstrate:
I think 10.000+ lines of test data is way too much. This will frustratie everybody who clones this repo. If you are able to find a Smaller test and cna erase it From the Git history  I would be really Happy. Apart From that This is outstanding work...

peterjc:
Good - I agree that test file is too large - and I was expecting to rewrite this branch history anyway so that's not a problem.

peterjc:
Uploaded to the Test Tool Shed  https://testtoolshed.g2.bx.psu.edu/view/jjohnson/gmap

I had to do this manually via ``planemo shed_upload --tar_only ...`` and the website as currently ``planemo shed_upload ...`` is broken, see https://github.com/galaxyproject/planemo/issues/575

peterjc:
The current TravisCI failures in linting are false positives,

```
.. ERROR: URL Error <urlopen error unknown url type: doi> accessing doi:10.1093/bioinformatics/bti310
.. ERROR: URL Error <urlopen error unknown url type: doi> accessing doi:10.1093/bioinformatics/bti310
.. ERROR: HTTP Error 403 accessing https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2844994/?tool=pubmed
.. ERROR: URL Error <urlopen error unknown url type: doi> accessing doi:10.1093/bioinformatics/bti310
.. ERROR: URL Error <urlopen error unknown url type: doi> accessing doi:10.1093/bioinformatics/bti310
```

This is due to https://github.com/galaxyproject/planemo/issues/573 (false positives in URL detection) which will be fixed in the next planemo release, and https://github.com/galaxyproject/planemo/issues/578 (false positive from NCBI blocking some pages depending on the UserAgent).

peterjc:
Is it acceptable to merge this and #963 now, to get the current state of the GMAP wrapper and package under IUC version control - even though there are not yet up to best practise standards?

yhoogstrate:
I would not recommend to merge this until there is a conda dependency. I don't know how others think about this.

bgruening:
If no-one has stepped up to create this old version for conda I will do this over the weekend.
Here is the newer one: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/gmap/meta.yaml

tools/gmap/gmap.xml
```diff
@@ -0,0 +1,514 @@
...
+      2> $gmap_stderr > $output
+    #end if
+  </command>
+  <inputs>
+    <!-- Input data -->
+    <param name="input" type="data" format="fasta,fastqsanger,fastqillumina" label="&lt;H2&gt;Input Sequences&lt;/H2&gt;Select an mRNA or EST dataset to map" />
+    <repeat name="inputs" title="addtional mRNA or EST dataset to map">
+      <param name="added_input" type="data" format="fasta,fastqsanger,fastqillumina" label=""/>
+    </repeat>
+    <param name="quality_protocol" type="select" label="Protocol for input quality scores">
```
yhoogstrate: Can this be removed by using the filetype of $input?

tools/gmap/gsnap.xml
```diff
@@ -0,0 +1,888 @@
...
+               help="If more than maximum number of paths are found, then nothing is printed."/>
+        <param name="show_refdiff" type="boolean" checked="false" truevalue="--show-refdiff" falsevalue="" label="Show SNP-tolerant alignment"
+               help="For GSNAP output in SNP-tolerant alignment, shows all differences relative to the reference genome as lower case (otherwise, it shows all differences relative to both the reference and alternate genome)"/>
+        <param name="clip_overlap" type="boolean" checked="false" truevalue="--clip-overlap" falsevalue="" label="Clip Overlap"
+               help="For paired-end reads whose alignments overlap, clip the overlapping region."/>
+      </when>
+    </conditional>
+    <conditional name="result">
+      <param name="format" type="select" label="Select the output format" help="">
+        <option value="sam">SAM</option>
```
yhoogstrate: I think we should convert to BAM using samtools

tools/gmap/gmap.xml
peterjc: In theory, yes.

tools/gmap/gsnap.xml
peterjc: Are there good examples of other tools doing this?

yhoogstrate: https://github.com/galaxyproject/tools-iuc/blob/master/tools/rgrnastar/rg_rnaStar.xml#L136

peterjc: Something like that would make sense as an enhancement :)

peterjc:
Good news - the new versions of this and ``package_gmap_2013_05_09`` (#963) which I uploaded to the Test Tool Shed have installed nicely onto our (old) Galaxy server. This ought to make developing some tests more practical...

tools/gmap/gmap.xml
peterjc: Looking at this, I would like to remove ``added_input`` by adding ``multiple="true"`` to the main ``input`` parameter - but that doesn't enforce all the inputs using the same quality scores. Unless there's a trick to require they all be the same file format?

Perhaps we should turn ``quality_protocol`` into a conditional which then lets the user pick multiple FASTA files, multiple Sanger FASTQ file, or multiple (legacy) Illumina scores?

tools/gmap/lib/galaxy/datatypes/gmap.py
```diff
@@ -0,0 +1,621 @@
...
+        flist = os.listdir(efp)
+        for i, fname in enumerate(flist):
+            m = re.match(pat, fname)
+            if m:
+                assert len(m.groups()) == 6
+                dataset.metadata.db_name = m.groups()[0]
+                dataset.metadata.snps_name = m.groups()[-1]
+
+
+class IntervalIndexTree(Text):
```
yhoogstrate: Should this inherit from Text or from Binary?

tools/gmap/lib/galaxy/datatypes/gmap.py
```diff
@@ -0,0 +1,621 @@
...
+        """
+        Determines whether the file is a gmap splice site annotation file
+        """
+        try:
+            # >total chr:position..position
+            pat = '^>(\d+)\s((\S+):(\d+)\.\.(\d+))$'
+            pat2 = '^[GATCN]\d.*$'  # BaseCountDeatails
+            fh = open(filename)
+            count = 0
+            while True and count < 10:
```
yhoogstrate: `while count < 10`

tools/gmap/lib/galaxy/datatypes/gmap.py
```diff
@@ -0,0 +1,621 @@
...
+        C1 T1   0.888,0.9,0.888,0.9,0.913,0.9,0.911,0.888,0.9,0.913     -2.66,-2.78,-2.66,-2.78,-2.91,-2.78,-2.89,-2.66,-2.78,-2.91
+    """
+
+    def sniff(self, filename):  # TODO
+        """
+        Determines whether the file is a gmap splice site annotation file
+        """
+        try:
+            # >total chr:position..position
+            pat = '^>(\d+)\s((\S+):(\d+)\.\.(\d+))$'
```
yhoogstrate:
```
pat = re.compile('^>(\d+)\s((\S+):(\d+)\.\.(\d+))$|^[GATCN]\d.*$')
...
if not pat.match(line):
```

tools/gmap/lib/galaxy/datatypes/gmap.py
```diff
@@ -0,0 +1,621 @@
...
+                    break  # EOF
+                line = line.strip()
+                if line:  # first non-empty line
+                    count += 1
+                    # Failed to match
+                    if re.match(pat, line) is None and re.match(pat2, line) is None:
+                        return False
+        finally:
+            fh.close()
+        return False
```
yhoogstrate: Can this function actually return `True`?

peterjc: It doesn't look like it, does it? We need some sample data for this file format...

tools/gmap/lib/galaxy/datatypes/gmap.py
peterjc: Seeing as this (and the other sniffers) reuse the regular expression, I agree, compiling it would be better.

tools/gmap/lib/galaxy/datatypes/gmap.py
peterjc: Agreed - it looks like JJ was debugging something when the snapshot was taken.

tools/gmap/lib/galaxy/datatypes/gmap.py
peterjc: I've not seen any of the sample data yet, so wouldn't like to guess.

tools/gmap/iit_store.xml
```diff
@@ -0,0 +1,183 @@
...
+  <description>Create a map store for known genes or SNPs</description>
+  <requirements>
+      <requirement type="package" version="2013-05-09">gmap</requirement>
+  </requirements>
+  <version_command>iit_store --version</version_command>
+  <command interpreter="command"> /bin/bash $shscript 2> $log </command>
+  <configfiles>
+    <configfile name="shscript">
+#!/bin/bash
+#set $catcmd = 'gzcat -f'
```
yhoogstrate: `zsh: command not found: gzcat` this requires a dependency

peterjc:
The GMAP source code has some basic self tests, which should be useful for making wrapper tests. However, starting with ``gmap.xml`` I ran into https://github.com/galaxyproject/galaxy/issues/2987

``` xml
    <test>
      <!-- 
      mimic first test from GMAP source code, mapping Human ERBB2 onto fragment of chr17
      $ gmap -A -g ss.chr17test ss.her2
      -->
      <param name="input" value="ss.her2.fasta" ftype="fasta"/>
      <param name="refGenomeSource.genomeSource" value="history"/>
      <param name="refGenomeSource.ownFile" value="ss.chr17.fasta" ftype="fasta"/>
      <param name="result.format" value="align"/>
      <output name="output" file="ss.her2.chr17.txt" ftype="txt"/>
    </test>
```

tools/gmap/iit_store.xml
peterjc: It looks like an unfinished feature, the next line is just ``#set $catcmd = 'cat'`` so we could even replace all of the ``$catcmd`` usage with plain ``cat``?

tools/gmap/lib/galaxy/datatypes/gmap.py
peterjc: Fixed in later commit, thanks.

yhoogstrate:
I I think that's the same issue as I once reported in planemo: https://github.com/galaxyproject/planemo/issues/490 

A possible workaround could be to update the .travis.yml such that if there is a loc file, it will copy the `^(.+.loc).sample$` file into `~/galaxy/tool-data/$1`

peterjc:
Thanks for the ``.loc`` tip @yhoogstrate - that could very well be important. I'm currently testing locally (not on TravisCI), and hoped to setup some test files via ``.loc`` later on. I'll give that a try next.

peterjc:
Progress:

* gmap_build runs locally on our old Galaxy
* gmap seems to be running with a FASTA reference (sample reference likely too big, it is still running)

However, trying to run gmap on our old Galaxy using the gmap_build index is failing with the ``kmer`` setting:

screenshot 2016-10-06 12 34 09

```
An invalid option was selected for kmer, u'15', please verify
```

(As an aside, the help text needs tweaking - it says it default to the largest available kmer but it seems to default to the smallest)

Adding more debugging to ``lib/galaxy/tools/parameters/basic.py`` makes me think the problem is ``value`` is a (unicode) string ``u'15'`` but ``legal_values = [12, 13, 14, 15]`` with integers. Sprinkling the code with ``str(...)`` seems to "fix" this, but it needs testing with the latest Galaxy code.

peterjc:
This is getting frustrating, without having a ``.loc`` file:

```
$ planemo test --install_galaxy gmap.xml
...
galaxy.web.framework.decorators: ERROR: Uncaught exception in exposed API method:
  File "/private/tmp/tmp3JRNlC/galaxy-dev/lib/galaxy/web/framework/decorators.py", line 257, in decorator
    rval = func( self, trans, *args, **kwargs)
  File "/private/tmp/tmp3JRNlC/galaxy-dev/lib/galaxy/webapps/galaxy/api/tools.py", line 262, in create
    vars = tool.handle_input( trans, incoming, history=target_history )
  File "/private/tmp/tmp3JRNlC/galaxy-dev/lib/galaxy/tools/__init__.py", line 1112, in handle_input
    self.populate_state( request_context, self.inputs, expanded_incoming, params, errors )
  File "/private/tmp/tmp3JRNlC/galaxy-dev/lib/galaxy/tools/__init__.py", line 1708, in populate_state
    state[ input.name ] = input.get_initial_value( request_context, context )
  File "/private/tmp/tmp3JRNlC/galaxy-dev/lib/galaxy/tools/parameters/grouping.py", line 579, in get_initial_value
    rval[ child_input.name ] = child_input.get_initial_value( trans, child_context )
  File "/private/tmp/tmp3JRNlC/galaxy-dev/lib/galaxy/tools/parameters/basic.py", line 971, in get_initial_value
    value = [ optval for _, optval, selected in options if selected ]
ValueError: too many values to unpack
...
gmap[0]: failed
```

Until Planemo fixes this problem - see https://github.com/galaxyproject/planemo/issues/490 and https://github.com/galaxyproject/planemo/issues/530 - we can avoid it by using a prepared copy of Galaxy and creating an empty ``.loc`` file:

```
$ touch ~/repositories/galaxy/tool-data/gmap_indices.loc && planemo test --galaxy_root ~/repositories/galaxy gmap.xml
...
galaxy.util.object_wrapper: WARNING: Unable to create dynamic subclass for <type 'instance'>, None: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
galaxy.jobs.runners: ERROR: (3) Failure preparing job
  File "/Volumes/Hitachi-3TB-mirrored/repositories/galaxy/lib/galaxy/jobs/runners/__init__.py", line 170, in prepare_job
    job_wrapper.prepare()
  File "/Volumes/Hitachi-3TB-mirrored/repositories/galaxy/lib/galaxy/jobs/__init__.py", line 905, in prepare
  File "/Volumes/Hitachi-3TB-mirrored/repositories/galaxy/lib/galaxy/tools/evaluation.py", line 445, in build
    raise e
ValueError: invalid literal for int() with base 10: ''
...
gmap[0]: failed
```

This problem looks like https://github.com/galaxyproject/galaxy/issues/998 - and I found earlier that cutting out most of the ``<command>`` tag I could get the test to pass so hopefully there is a similar workaround.

peterjc:
The test failure with ``Exception: invalid literal for int() with base 10: '''`` is related to galaxyproject/galaxy#998 although specifically from the boolean parameter ``split_output``

tools/gmap/gmap.xml
```diff
@@ -378,7 +378,7 @@
       </when>
     </conditional> <!-- name="result" -->
 
-    <param name="split_output" type="boolean" truevalue="--split-output=gmap_out" falsevalue="" checked="false" label="Separate outputs for nomapping, uniq, mult, and chimera" help="(chimera only when chimera-margin is selected)"/>
+    <param name="split_output" type="boolean" truevalue="True" falsevalue="False" checked="false" label="Separate outputs for nomapping, uniq, mult, and chimera" help="(chimera only when chimera-margin is selected)"/>
```
nsoranzo: I think you can also do something like:
```
#if $split_output
  --split-output=gmap_out
#end if
```
...
```
<param name="split_output" type="boolean" label="Separate outputs for nomapping, uniq, mult, and chimera" help="(chimera only when chimera-margin is selected)"/>
```

peterjc: Thanks! Direct boolean testing is a much nicer solution

peterjc:
This alternative fix for ``split_output`` is based on @nsoranzo's idea, but preserving the original true/false values for anyone wanting to re-run old jobs.

peterjc:
Rebased to include package_gmap_2013_05_09 originally on separate PR #963 as per @yhoogstrate's request.

peterjc:
Progress with testing, e.g. https://travis-ci.org/peterjc/tools-iuc/jobs/168674176 - although failing:

```
...
galaxy.web.framework.decorators ERROR 2016-10-18 16:48:05,923 Uncaught exception in exposed API method:
  File "/tmp/tmpgWgb3o/galaxy-dev/lib/galaxy/web/framework/decorators.py", line 261, in decorator
    rval = func( self, trans, *args, **kwargs)
  File "/tmp/tmpgWgb3o/galaxy-dev/lib/galaxy/webapps/galaxy/api/tools.py", line 289, in create
    vars = tool.handle_input( trans, incoming, history=target_history )
  File "/tmp/tmpgWgb3o/galaxy-dev/lib/galaxy/tools/__init__.py", line 1138, in handle_input
    self.populate_state( request_context, self.inputs, expanded_incoming, params, errors )
  File "/tmp/tmpgWgb3o/galaxy-dev/lib/galaxy/tools/__init__.py", line 1761, in populate_state
    state[ input.name ] = input.get_initial_value( request_context, context )
  File "/tmp/tmpgWgb3o/galaxy-dev/lib/galaxy/tools/parameters/grouping.py", line 586, in get_initial_value
    rval[ child_input.name ] = child_input.get_initial_value( trans, child_context )
  File "/tmp/tmpgWgb3o/galaxy-dev/lib/galaxy/tools/parameters/basic.py", line 974, in get_initial_value
    value = [ optval for _, optval, selected in options if selected ]
ValueError: too many values to unpack
requests.packages.urllib3.connectionpool DEBUG 2016-10-18 16:48:05,926 "POST /api/tools HTTP/1.1" 500 None
ERROR
======================================================================
ERROR: GMAP ( gmap ) > Test-1
----------------------------------------------------------------------
  File "/tmp/tmpgWgb3o/galaxy-dev/test/functional/test_toolbox.py", line 299, in test_tool
    self.do_it( td )
  File "/tmp/tmpgWgb3o/galaxy-dev/test/functional/test_toolbox.py", line 60, in do_it
    raise e
RunToolException: Error creating a job for these tool inputs - Uncaught exception in exposed API method:
-------------------- >> begin captured logging << --------------------
...
```

I think the exception is a bug in Galaxy, see https://github.com/galaxyproject/galaxy/issues/2987 - but can likely be avoided with explicit handling of the test ``*.loc`` files in the TravisCI setup.

peterjc:
After some pain (see https://github.com/galaxyproject/planemo/issues/588 for some of it), I have a potential GMAP test which currently fails on the ``tools-iuc`` TravisCI because I ask for gmap version 2013-05-09 (which has a package https://toolshed.g2.bx.psu.edu/view/jjohnson/package_gmap_2013_05_09 on the ToolShed), which is not in BioConda.

How would I change this to request gmap 2016.09.23 which is in BioConda according to https://github.com/bioconda/bioconda-recipes/blob/master/recipes/gmap/meta.yaml ?

CC @bgruening

nsoranzo:
@peterjc You need to add a second requirement for the Conda version:
```
<requirement type="package" version="2016.09.23">gmap</requirement> <!-- Conda dependency -->
```

Or just go Conda-only :wink:

tools/gmap/gmap.xml
```diff
@@ -0,0 +1,532 @@
+<tool id="gmap" name="GMAP" version="3.0.2">
+  <description>Genomic Mapping and Alignment Program for mRNA and EST sequences</description>
+  <requirements>
+    <requirement type="package" version="2016-09-23">gmap</requirement>
```
nsoranzo: Conda package version is `2016.09.23` (with dots), because the version cannot contain dash ('-') characters, see http://conda.pydata.org/docs/building/meta-yaml.html#package-version

nsoranzo:
Should we remove `packages/package_gmap_2013_05_09/` directory now?

peterjc:
No, we'd keep ``packages/package_gmap_2013_05_09/`` as the associated (legacy) Tool Shed entry will still exist.

peterjc:
Thanks @nsoranzo - I finally have a TravisCI run with GMAP installed from BioConda: https://travis-ci.org/peterjc/tools-iuc/jobs/170449796

It still fails due to Galaxy issue galaxyproject/galaxy#2987 - this can likely be avoided with explicit handling of the test ``*.loc`` files in the TravisCI setup, but right now the IUC lets planemo create a disposable Galaxy instance so we can't easily modify it prior to ``planemo test``.

bgruening:
@peterjc can you add this tool to the .tt_blacklist with a comment that these 2 issues are blocking?
Thanks!

peterjc:
I'm not likely to get back to this for several days @bgruening - if you want to do that now, please go ahead - I think GitHub will let you update this branch?

peterjc:
I suspect the recent change to try to use ``from_data_table`` for the ``.loc`` file didn't work, still need to pin that down with some manual testing before asking to merge this.

tools/gmap/tool-data/datatypes_conf.xml~
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
```
nsoranzo: This file should be removed.

peterjc:
Is it possible to retrospectively allow you guys to edit this branch? That's a relatively recent GitHub feature... otherwise I could push the branch to the main ``tools-iuc`` repo (rather than a branch on my fork)?

nsoranzo:
@peterjc I think we can already push (I see the message `Add more commits by pushing to the gmap branch on peterjc/tools-iuc.`). I can try to push it forward, if that's OK with you.

peterjc:
@nsoranzo Please go ahead, thanks!

tools/gmap/gmap.xml
```diff
@@ -0,0 +1,532 @@
...
+                                   of this filter
+    -->
+
+
+
+  </inputs>
+  <outputs>
+    <data format="txt" name="gmap_stderr" label="${tool.name} on ${on_string}: stderr"/>
+    <data format="txt" name="output" label="${tool.name} on ${on_string} ${result.format}" >
+      <filter>(split_output == False)</filter>
```
nsoranzo: s/(split_output == False)/not split_output/

tools/gmap/gmap.xml
```diff
@@ -0,0 +1,532 @@
...
+        <when input="result['format']" value="gff3_match_est" format="gff3"/>
+        <when input="result['format']" value="sam" format="sam"/>
+        <when input="result['format']" value="splicesites" format="gmap_splicesites"/>
+        <when input="result['format']" value="introns" format="gmap_introns"/>
+        <when input="result['format']" value="map_genes" format="gmap_annotation"/>
+        <when input="result['format']" value="map_exons" format="gmap_annotation"/>
+      </change_format>
+    </data>
+    <data format="txt" name="uniq" label="${tool.name} on ${on_string} uniq.${result.format}"  from_work_dir="gmap_out.uniq">
+      <filter>(split_output == True)</filter>
```
nsoranzo: s/(split_output == True)/split_output/

tools/gmap/gmap.xml
```diff
@@ -0,0 +1,532 @@
...
+        <when input="result['format']" value="gff3_match_est" format="gff3"/>
+        <when input="result['format']" value="sam" format="sam"/>
+        <when input="result['format']" value="splicesites" format="gmap_splicesites"/>
+        <when input="result['format']" value="introns" format="gmap_introns"/>
+        <when input="result['format']" value="map_genes" format="gmap_annotation"/>
+        <when input="result['format']" value="map_exons" format="gmap_annotation"/>
+      </change_format>
+    </data>
+    <data format="txt" name="transloc" label="${tool.name} on ${on_string} transloc.${result.format}"  from_work_dir="gmap_out.transloc">
+      <filter>(split_output == True)</filter>
```
nsoranzo: s/(split_output == True)/split_output/

tools/gmap/gmap.xml
```diff
@@ -0,0 +1,532 @@
...
+        <when input="result['format']" value="gff3_match_est" format="gff3"/>
+        <when input="result['format']" value="sam" format="sam"/>
+        <when input="result['format']" value="splicesites" format="gmap_splicesites"/>
+        <when input="result['format']" value="introns" format="gmap_introns"/>
+        <when input="result['format']" value="map_genes" format="gmap_annotation"/>
+        <when input="result['format']" value="map_exons" format="gmap_annotation"/>
+      </change_format>
+    </data>
+    <data format="txt" name="nomapping" label="${tool.name} on ${on_string} nomapping.${result.format}"  from_work_dir="gmap_out.nomapping">
+      <filter>(split_output == True)</filter>
```
nsoranzo: s/(split_output == True)/split_output/

tools/gmap/gmap.xml
```diff
@@ -0,0 +1,532 @@
...
+        <when input="result['format']" value="gff3_match_est" format="gff3"/>
+        <when input="result['format']" value="sam" format="sam"/>
+        <when input="result['format']" value="splicesites" format="gmap_splicesites"/>
+        <when input="result['format']" value="introns" format="gmap_introns"/>
+        <when input="result['format']" value="map_genes" format="gmap_annotation"/>
+        <when input="result['format']" value="map_exons" format="gmap_annotation"/>
+      </change_format>
+    </data>
+    <data format="txt" name="mult" label="${tool.name} on ${on_string} mult.${result.format}"  from_work_dir="gmap_out.mult">
+      <filter>(split_output == True)</filter>
```
nsoranzo: s/(split_output == True)/split_output/

PR #965, opened 2016-09-26
Enable VCF to be used for annotating GEMINI database.

jgoecks:
Currently only BED files can be used for annotating a GEMINI database. This commit enables VCF files to be used for annotation as well.

### DIFF
--- a/.tt_blacklist
+++ b/.tt_blacklist
@@ -2,7 +2,6 @@ tools/blastxml_to_gapped_gff3
 tools/dexseq
 tools/differential_count_models
 tools/gatk2
-tools/gemini
 tools/htseq
 tools/gff3_rebase
 tools/jbrowse
--- a/.tt_blacklist
+++ b/.tt_blacklist
@@ -11,6 +11,7 @@ tools/progressivemauve
 tools/rglasso
 tools/rgrnastar
 tools/taxonomy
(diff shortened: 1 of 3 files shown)


yhoogstrate:
(`tabix` is not yet in bioconda and gemini is in the .tt_blacklist)

jgoecks:
But GEMINI suite is on Galaxy main, and it would be nice to get this fix out there.

bgruening:
`tabix` is actually in bioconda. It is now part of htslib. `gemini` is on this blacklist because I think we do not have tests for all tools? Because the db was to big, or something similar?

yhoogstrate:
It would only require to replace `<requirement type="package" version="0.2.6">tabix</requirement>` with `<requirement type="package" version="1.3.1">htslib</requirement>` in gemini_macros.xml

bgruening:
Do we need to bump the version number?

nekrut:
pls, push this onto main asap

jgoecks:
@bgruening Thanks for picking this up and following through, great work!

bgruening:
Will get this in today ... see you later.

bgruening:
@yhoogstrate let's work on these wrappers in a separate PR and bring them finally under testing!

PR #967, opened 2016-09-27
Mothur fixes

shiltemann:
adding fixes here as I/others encounter them, more will undoubtedly be added, we can leave this open until an update in (test)toolshed is desired

TODO:
- [x] fix broken links
- [x] add tool for get.communitytype command
- [x] add tool for lefse command
- [x] add tool for classify.rf command

Edit: 
Some additional TODOs based on feedback from @arbernard (thanks!)
- [x] align.seqs: output accnos file if flip=true
- [x] single input param for name/count file in cluster.split, create.database, get.oturep
- [x] make.contigs: allow choice between collection input or two regular forward and reverse files


tools/mothur/rarefaction.single.xml
```diff
@@ -30,6 +30,11 @@
         )'
         | sed 's/ //g'  ## mothur trips over whitespace
         | mothur
+
+        ## rename output from sobs calculator to be consistent with others
+        #if 'sobs' in str($calc).split(','):
```
yhoogstrate: `#if str($calc).find('sobs') > -1:`

blankenberg: `#if str($calc).find('sobs') > -1:` is a much less stringent test, it would also match on 'newsobs', 'sobs2', etc, not that they are actually valid options at this time.

shiltemann:
this will probably be it for changes for at least a little while from me, so if anybody has some time for a review? :)

blankenberg:
Was trying this locally using conda, but looks like the dependencies might be broken due to lib boost version mismatch (using `make.group`):
```
mothur: error while loading shared libraries: libboost_iostreams.so.1.57.0: cannot open shared object file: No such file or directory
mothur: error while loading shared libraries: libboost_iostreams.so.1.57.0: cannot open shared object file: No such file or directory
sed: couldn't flush stdout: Broken pipe
```

```
The following NEW packages will be INSTALLED:

    boost:      1.62.0-py35_0   
    certifi:    2016.9.26-py35_0
    icu:        56.1-4          
    mothur:     1.36.1-0        
    ncurses:    5.9-10          
    openssl:    1.0.2j-0        
    pip:        9.0.1-py35_1    
    python:     3.5.2-2         
    readline:   6.2-2           
    setuptools: 29.0.1-py35_1   
    sqlite:     3.13.0-1        
    tk:         8.5.19-1        
    wheel:      0.29.0-py35_0   
    xz:         5.2.2-0         
    zlib:       1.2.8-3     
```

Will see if I can find some time to dig further.

(Also ended up with Green Dataset, instead of error, will take a look at that as well.)

bgruening:
@blankenberg boost needs to be pinned to an exact version.

So I guess this line is wrong: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/mothur/meta.yaml#L24

We have one global boost version defined here:
https://github.com/bioconda/bioconda-recipes/blob/master/scripts/env_matrix.yml#L5

Can you fix this line increase the build number and I think the new version should work as expected.

shiltemann:
thanks for looking at this @blankenberg :) 

@bgruening huh, I thought this /was/ how you pinned things? (we ran into the exact same thing before and then it was fixed by pinning it this way right?) what should this line be changed to now? 

also, looks like it is getting build zero right? `mothur:     1.36.1-0` ..while the latest build was 2 before it got overwritten by a recipe for mothur 1.38 (and build set back to 0) in bioconda/bioconda-recipes/pull/3077 ..is this causing problems somehow? ..should I set the build version too in the requirements of the galaxy tool? 

And as the recipe got overwritten to version 1.38 but we still need version 1.36, should I make a folder in there for version 1.36 (like there is for 1.25) in case we ever want to update version 1.36 (like maybe now)?

bgruening:
> @bgruening huh, I thought this /was/ how you pinned things? (we ran into the exact same thing before and then it was fixed by pinning it this way right?) what should this line be changed to now? 

I guess its needs to look like in the run-section.

> also, looks like it is getting build zero right? `mothur:     1.36.1-0` ..while the latest build was 2 before it got overwritten by a recipe for mothur 1.38 (and build set back to 0) in bioconda/bioconda-recipes/pull/3077 ..is this causing problems somehow? ..should I set the build version too in the requirements of the galaxy tool? 

That the more likely error which Dan has seen. I guess this is due to the old conda version we are using.

> And as the recipe got overwritten to version 1.38 but we still need version 1.36, should I make a folder in there for version 1.36 (like there is for 1.25) in case we ever want to update version 1.36 (like maybe now)?

Yes, a subfolder would work.

shiltemann:
@bgruening ah yes, I think you're right, tested this today with miniconda3 and didn't run into this issue

blankenberg:
Thanks for looking into the dependency issue.

For the Green OK Dataset on tool error. It looks like this has to do with piping mothur's output through `tee`, since we lose previous commands' exit codes when using pipes and we are not catching "error" on stderr. Good discussion here: http://stackoverflow.com/questions/1221833/bash-pipe-output-and-capture-exit-status

shiltemann:
thanks @blankenberg, and thanks for the link, will add a fix soon :)

shiltemann:
I think we should probably split this repo up, tests aren't even coming close to completing within the max build time (3,5 hours locally), plus hitting install on the entire suite in the (test)toolshed is still leading to an internal server error, presumably due to number of tools ..maybe a `mothur_core` directory with all the tools needed for the SOP and one (or two) with the rest? any thoughts?

shiltemann:
cool, I will rebase in a bit :)

bgruening:
@shiltemann http://www.bioinformatics-toolkit.org is gone. Should we remove the link.

bgruening:
Is this here green? Looks like it - I guess 18.5 is the official Galaxy Metagenomics day! :)
Thanks a bunch @shiltemann!

shiltemann:
thanks @bgruening !

bgruening:
I'm currently uploading manually, as the upload was exceeding the travis limit.

PR #969, opened 2016-09-28
Fix matplotlib installation in some environments

davebx:
In the specific case where I encountered this issue, matplotlib was trying to link to /usr/local/lib/libpython2.7.a, when it should have been using $INSTALL_DIR/lib/python2.7/config/libpython2.7.a

This was fixed by adding LDFLAGS as in the tool dependency definition.

### DIFF
--- a/packages/package_python_2_7_10/tool_dependencies.xml
+++ b/packages/package_python_2_7_10/tool_dependencies.xml
@@ -67,6 +67,7 @@
             $INSTALL_DIR/bin/python setup.py install --prefix=$INSTALL_DIR
         </action>
         <action type="set_environment">
+            <environment_variable action="set_to" name="LDFLAGS">-L$INSTALL_DIR/lib/python2.7 -L$INSTALL_DIR/lib/python2.7/config</environment_variable>
             <environment_variable action="prepend_to" name="PATH">$INSTALL_DIR/bin</environment_variable>
             <environment_variable action="prepend_to" name="PYTHONPATH">$INSTALL_DIR</environment_variable>
             <environment_variable action="prepend_to" name="PYTHONPATH">$INSTALL_DIR/lib</environment_variable>


packages/package_python_2_7_10/tool_dependencies.xml
```diff
@@ -67,6 +67,7 @@
             $INSTALL_DIR/bin/python setup.py install --prefix=$INSTALL_DIR
         </action>
         <action type="set_environment">
+            <environment_variable action="set_to" name="LDFLAGS">-L$INSTALL_DIR/lib/python2.7 -L$INSTALL_DIR/lib/python2.7/config</environment_variable>
```
nsoranzo: `prepend_to` instead of `set_to` maybe?

davebx: `prepend_to` uses a colon as separator.

gregvonkuster:
I know this fix worked for my specific environment, so LGTM.  However, I am not an expert here, so please take my approval with a grain of salt or two...  ;)

packages/package_python_2_7_10/tool_dependencies.xml
nsoranzo: Ouch! By looking at other packages, it seems the preferred solution would be to modify matplotlib recipe by adding an:
```
export LDFLAGS="-L$PYTHON_ROOT_DIR/lib/python2.7 -L$PYTHON_ROOT_DIR/lib/python2.7/config" &amp;&amp;
```

On the other hand, this package has a bug below:
```
<environment_variable action="prepend_to" name="PYTHON_ROOT_DIR">$INSTALL_DIR</environment_variable>
```
should be:
```
<environment_variable action="set_to" name="PYTHON_ROOT_DIR">$INSTALL_DIR</environment_variable>
```

bgruening:
@davebx @nsoranzo how do we proceed here?

bgruening:
@davebx please remember this needs to be uploaded manually.

PR #972, opened 2016-09-29
Increase maximum download filesize to 200G and use new sra-tools version

mvdbeek:
Default was 20G apparently ... .
May solve https://github.com/galaxyproject/galaxy/issues/2533#issuecomment-250280494 ?
ping @natefoo


natefoo:
:+1: thanks! Not sure if I am allowed to merge in tools-iuc. =)

Issue #975, opened 2016-10-02
DESeq2 option to export Rdata object 


AMaheras7091:
I have completed an RNA-seq workflow to find differential gene expression and it is paramount to my paper that I am able to visualize my results in R accordingly. Thus, this is a feature request for an option to export the RData object.

Thank you.

PR #714, opened 2016-04-05
upgrade deseq2 and include tximport to use sailfish results directly 

bgruening:
This one is called DESeq2-TNG I should have named it RNA-seq TNG ;)
I'm very excited about this PR and I'm seeking for help in testing. A little bit of background: until now we need to map the reads and use htseq-count to count reads in a region to feed this into DESeq2.
On the other hand Sailfish is a "rapid Alignment-free Quantification of Isoform Abundance" tool ... but we can not use the output natively with DESeq2. This has changed now!
@mikelove has developed a nice tool called tximport that can read sailfish and salmon outputs directly and does the DESEq2 magic on top of it.

This PR updated our DESEq2 wrapper to make use of this new functionality. The workflow should be much simpler and much faster, as it is alignment free. I have created a new DESEq2 conda package to ease testing. So please activate the "bgruening" conda channel and run planemo on it if you want to provide feedback!

@mikelove and me would love to have as much feedback as possible until the new bioc release in two weeks and the new DESeq2 release. After this there is a freeze for the next 6 month ... more or less.

Let us know what you think we both are pretty excited about this :)
Cheers,
Michael and Bjoern

xref: https://github.com/galaxyproject/tools-iuc/pull/699

PR #978, opened 2016-10-03
Optional RData output file for DESeq2

yhoogstrate:
solves https://github.com/galaxyproject/tools-iuc/issues/975


bgruening:
@yhoogstrate thanks for the enhancement. Do you know if a RData object is persistent between different R version? Which promises are we doing by offering this option?

yhoogstrate:
I saw an issue on the Galaxy biostar where someone requested the possibility to get *publishable figures* (https://biostar.usegalaxy.org/p/19742/#19752). The figures currently produced are detailed but I can imagine someone like to customize them e.g. by changing the axes or naming the samples in the PCA plot. I don't know to what extend this is a problem especially as PDFs are vector based figures, but having an R object would give enless options. You are also right that there is backwards incompatibility as far as I read in the manual:

*Although the default version has not changed since R 1.4.0, this does not mean that saved files are necessarily backwards compatible. You will be able to load a saved image into an earlier version of R unless use is made of later additions (for example, raw vectors, external pointers and some S4 objects).*
*One such ‘later addition’ was long vectors, introduced in R 3.0.0 and loadable only on 64-bit platforms.*

I don't know what the best choice would be, and thanks for reviewing this @bgruening

bgruening:
Ok, let's do the following. Can we create a macro which we can share between all R tools that contains an advanced section with this option, hided by default and with an explicit statement that the backward compat is not guaranteed between different R versions?

This way we can add this feature to all our R tools in a consistent way.

bgruening:
@yhoogstrate what do you think about the macro idea? 
Can you also please look at https://github.com/galaxyproject/tools-iuc/pull/714 - new generation of DESeq2

yhoogstrate:
Thanks @bgruening, I like the macro idea. I will look at your PR in the beginning of next week.

bgruening:
So this needs to be rebased on the new DESeq2 wrapper :)

PR #982, opened 2016-10-06
Update pilon requirement to v 1.20

pvanheus:
* Cosmetic fix to label of reference genome
* Note new 'circles' fix, not implemented yet


nsoranzo:
Thanks @pvanheus!

tools/pilon/pilon.xml
```diff
@@ -1,7 +1,7 @@
-<tool id="pilon" name="pilon" version="0.1">
+<tool id="pilon" name="pilon" version="0.1.2">
```
nsoranzo: Not needed, but you may track the `pilon` version here using e.g. `1.20.0` where the last 0 is for the wrapper version.

PR #983, opened 2016-10-07
Added Torsten Seemann's MLST tool

madiflannery:


bgruening:
The failing test is not due to the wrapper. It seems that list::moreutils is missing: `can't locate List/MoreUtils.pm in @INC (you may need to install the List::MoreUtils module)`.

But it is included in the conda recipe: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/mlst/meta.yaml#L24

Any perl expert here?

madiflannery:
Hmmm how weird! The conda package of MLST installed on my mac (i.e. outside of galaxy) works fine. Unfortunately I'm not a perl (or conda) expert, but pinging @Slugger70 @tseemann just in case they have any ideas.

madiflannery:
Hi @bgruening :) Just a thought - is it possible that the order of the dependencies in my mlst.xml wrapper is causing this? The order is MLST, then perl, then blast+ - Assuming it resolves the dependencies in order, is it possible that it grabs MLST from conda then when it deals with the perl 5.18.1 dependency it resets the perl environment or something?

bgruening:
@madiflannery good idea. Can you remove them? Btw. the Conda resolver don't need the perl and blast one.

madiflannery:
OK I can do that - should I move them into the tool_dependencies.xml inside package_mlst_2_6? So then they still get installed if conda isn't used for the dependency management?

bgruening:
@madiflannery mh, forget tool_dependencies.xml at the moment. Let try if we can get Conda to work by removing them from the `>requirements>` tag. tool_dependencies.xml are not strictly needed anymore.

madiflannery:
That seems to have fixed the perl error yay :) But one of my tests failed and I'm not 100% sure why- looks like test 4 errored with `JobOutputsError: No JSON object could be decoded` - any hints on what could be causing that? Planemo test passes when I run it on my computer.

bgruening:
Don't worry this was a random failure with db locking. I restarted Travis and I'm confident it will work. Keep the tools coming :-)

bgruening:
@madiflannery but now the tool_dependencies.xml does not work anymore. It's up to you if you want to support this as well. You could also simply remove this xml file.
Thanks,
Bjoern

madiflannery:
Hmmm OK - I'll investigate tomorrow and get back to you :)

bgruening:
@madiflannery anything we can help?

madiflannery:
Hey @bgruening - so sorry for the delay I have been crazy busy! But this is on my to-do list for today & tomorrow :)

madiflannery:
Hey @bgruening - it looks like the SSL certs for the toolshed and test toolshed have expired :( Planemo is having a whinge about it - getting this error when I try and update the tool in the test toolshed `[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)`

jmchilton:
Ping @natefoo

bgruening:
@madiflannery no worry, can you test again the TS should work again. Thanks @natefoo!

madiflannery:
Thanks guys :)

I've re-added the blast dependency into the MLST package tool_dependencies.xml. Tested it on a non-conda galaxy instance (i.e. so it doesn't use conda as dependency resolver) and seems to be working :)

bgruening:
Thanks @madiflannery! I have two last requests:
* your test data is large, is there any way you can reduce your fasta file, ideally below 1000 lines? 
* Can you add more data to the shed.yml files, like repository_url and a long description.

Thanks again. Let me know if you can not work on this and I will try to help.

peterjc:
Does this specifically need BLAST+ 2.2.31 which is being pulled in via ``package_blast_plus_2_2_31`` (which will not work on BioConda due to a naming mismatch)

See https://github.com/peterjc/galaxy_blast/issues/88 and https://github.com/peterjc/galaxy_blast/pull/89 which will create ``package_blast_plus_2_5_0`` named to match BioConda

madiflannery:
@bgruening no worries- I can do that :)
@peterjc I think it should be ok to update to `package_blast_plus_2_5_0` but I'll chase it up to double check :) Thanks for the heads up!

bgruening:
Great work @madiflannery! Thanks a lot and sorry for being so nitpicking :)

PR #987, opened 2016-10-10
Stacks: update to 1.46 + summary HTML report

abretaud:
Hi all,
Added calls to stacks_summary (by @mariabernard) to produce beautiful HTML report from stacks output.
@mariabernard can you look at this and/or test it?
I didn't add the commands for genotypes and populations as we don't currently save the log file. Is it required to produce good reports?


mariabernard:
the log file is only required when cstacks is launched, so for cstacks, denovo_map and ref_map.

abretaud:
Ok, I finished updating the wrappers to 1.46, added missing bits of stacks_summary, fixed a few bugs and made a bit of cleanup.
Tests are (almost) all green when testing with a local install of stacks.

There are 2 problems left:
-a segfault in stacks_populations test (reported there: https://groups.google.com/forum/#!topic/stacks-users/uqm1peUEod4)
-the stacks bioconda recipe is not yet available as it requires on a newer version of gcc (https://github.com/bioconda/bioconda-recipes/pull/3987)

So it cannot be merged yet, but it can be reviewed/tested!
ping @lecorguille (in case you have not seen this PR)

abretaud:
We now have a workaround for the segfault :)

bgruening:
@abretaud where is it? :)

abretaud:
There: https://github.com/galaxyproject/tools-iuc/pull/987/commits/b6439930dcba993ee32f1f17be9aef848adc0634#diff-cc5f39592bae6848b3049fdbbb110555L367
I removed the (not needed) option that was provoking the segfault, and the segfault itself should be fixed in next stacks version

bgruening:
@abretaud great! Can they also relax the requirement for a newer GCC ;)

abretaud:
I haven't asked (yet!)
No news on the possible newer gcc in conda?

bgruening:
Not yet, no :(

abretaud:
The conda package is ok now :)
But I get [this error](https://travis-ci.org/galaxyproject/tools-iuc/jobs/223082420#L481) now :(
Any idea?

bgruening:
There seems to be an error in your shed.yaml file.

abretaud:
Holy cow! It's green!

bgruening:
Yeah!!!!

tools/stacks/stacks_denovomap.xml
```diff
@@ -6,54 +6,47 @@
...
-                    #set $data_path = $data_path + ".fa"
+
+        @CLEAN_EXT@
+
+        #if str( $options_usage.rad_analysis_type ) == "genetic"
+            #for $input_parent in $options_usage.parent_sequences
+                #if $input_parent.is_of_type('fastqsanger')
+                    #set $data_path = $clean_ext($input_parent.element_identifier) + ".fq"
+                #else if $input_parent.is_of_type('fastqsanger.gz')
+                    #set $data_path = $clean_ext($input_parent.element_identifier) + ".fq.gz"
```
bgruening: Is there support for bz2 files?

tools/stacks/stacks_ustacks.xml
```diff
@@ -75,28 +74,38 @@
 
             -o stacks_outputs
 
-             > ustacks.log 2>&1
+            2>&1 | tee ustacks.log
+
+            ## If input is in gz format, stacks will output gzipped files (no option to control this)
```
bgruening: Can you not write something like `if is_of_type(foobar.gz): gunzip bar/*`

bgruening:
This is really a nice work! I added two small comments.

tools/stacks/stacks_denovomap.xml
abretaud: No unfortunately, only gz for now

tools/stacks/stacks_ustacks.xml
abretaud: Done for ustacks! For denovo_map it's a little trickier, we feed him multiple files and I think it can be a mix of fq and fq.gz. I don't know what stacks will do in this case...

abretaud: Self-answering: it will produce a mix of zipped/non-zipped tsv files

abretaud:
Hey,
@bgruening is it ok for you now?
Anyone else willing to review it or do we merge it?

bgruening:
Sure! Thanks!

abretaud:
Wunderbar! Thank you!

lecorguille:
Thanks a bunch, like they use to say!

PR #992, opened 2016-10-11
Allow gff/gff3/vcf in sortBed, single_quote paths and upgrade to bedt…

mvdbeek:
…ools 2.26.0, drop TS dependency, reset  in wrapper version.


tools/bedtools/macros.xml
```diff
@@ -1,11 +1,11 @@
 <macros>
     <xml name="requirements">
         <requirements>
-            <requirement type="package" version="2.24">bedtools</requirement>
+            <requirement type="package" version="2.26.0">bedtools</requirement>
             <yield/>
         </requirements>
     </xml>
-    <token name="@WRAPPER_VERSION@">2.24</token>
+    <token name="@WRAPPER_VERSION@">2.26</token>
```
nsoranzo: Should we also use 2.26.0 here?

nsoranzo:
Some tests failed, it looks like that upgrading to 2.26.0 creates differences in some test output.

mvdbeek:
Yes, there are some off by one fixes mentioned in the 2.26.0 release notes, though it should only affect vcf/gff test-data. I am running planemo locally with --update_test_data to see what those differences are.

tools/bedtools/macros.xml
mvdbeek: I think so, good catch.

bgruening:
mh, `bedtools_shufflebed` failed :(

mvdbeek:
yeah, it works repeatedly on my local OSX, I'm guessing the new way of shuffling may result in a different outcome based on the platform.

mvdbeek:
Alright, tests pass,  but I have some more commits on my machine in the lab for quoting everything. I'll update this tomorrow.

mvdbeek:
Alright, this is ready to go.

bgruening:
Thanks Marius!

mvdbeek:
Seems no tests were run: https://travis-ci.org/galaxyproject/tools-iuc/builds/168227744, and subsequently no upload :(

nsoranzo:
@mvdbeek That's my fault, should be fixed in https://github.com/galaxyproject/tools-iuc/pull/998 .

PR #994, opened 2016-10-12
Jbrowse: new tool + contextual menus

abretaud:
A new tool to export a jbrowse data dir to an external jbrowse instance, to be used with erasche/docker-jbrowse/pull/3
Also added support for adding contextual menu in annotation track/blast result


erasche:
Will review tomorrow

abretaud:
I moved the generate-names.pl call to the end to also index the tracks (to enable searching by gene names). I'm not sure if this should be optional, as we might not want to index some tracks (repeats for example)

erasche:
> I moved the generate-names.pl call to the end to also index the tracks (to enable searching by gene names). I'm not sure if this should be optional, as we might not want to index some tracks (repeats for example)

I'd be tempted to say we should make it a) optional per-track, and b) optional at a global level until I can get the stupid "Error reading from name store" messages to go away in production (hashBits doesn't fix it)

abretaud:
Hej!
I think I found a solution for the stupid (indeed!) message: GMOD/jbrowse/pull/820
And also a small change that could be of interest: GMOD/jbrowse/pull/821
I'll work on the optional indexing

tools/jbrowse/jbrowse.py
```diff
@@ -388,6 +419,9 @@ def add_blastxml(self, data, trackData, blastOpts, **kwargs):
         self.subprocess_check_call(cmd)
         os.unlink(gff3)
 
+        if blastOpts.get('index', 'false') == 'true':
+            self.tracksToIndex.append("'%s'" %s trackData['label'])
```
mvdbeek: I think `self.tracksToIndex.append("'%s'" %s trackData['label'])` should be `self.tracksToIndex.append("'%s'" % trackData['label'])`

tools/jbrowse/jbrowse.py
```diff
@@ -486,6 +520,9 @@ def add_features(self, data, format, trackData, gffOpts, **kwargs):
 
         self.subprocess_check_call(cmd)
 
+        if gffOpts.get('index', 'false') == 'true':
+            self.tracksToIndex.append("'%s'" %s trackData['label'])
```
mvdbeek: same as above

erasche:
(Fyi @abretaud, I'm mid-review on this, sorry for the delays, I've got a dozen orso comments, work keeps interrupting. Will have stuff for you soon)

abretaud:
Can't wait for the "stuff for me" !
No worry for the delay, it's busy for me too, I still haven't find the time to test my last commits...

tools/jbrowse/jbrowse_to_existing.xml
```diff
@@ -0,0 +1,42 @@
...
+        #set dataset_name = $unique_id
+    #else:
+        #set dataset_name = $jbrowse.id
+    #end if
+
+    cp -R '${jbrowse.extra_files_path}/data/' "@DATA_DIR@/${dataset_name}"
+
+    &&
+
+    echo -e "[datasets.${dataset_name}]\nurl  = ?data=data/${dataset_name}\nname = ${name}" >> "@DATA_DIR@/datasets.conf"
```
erasche: Based on my understanding of what's going on here... you've deployed jbrowse somewhere on the server, and set `GALAXY_JBROWSE_SHARED_DIR=$JBROWSE_ROOT/data/`, correct?

I have success (i.e. a list of organisms) if I name it "tracks.conf", but not if I name it datasets.conf.

This took a bit of figuring out, maybe need notes for admin using / installing this tool

tools/jbrowse/jbrowse_to_existing.xml
```diff
@@ -0,0 +1,42 @@
...
+
+    &&
+
+    echo -e "[general]\ndataset_id  = ${dataset_name}" >> "@DATA_DIR@/${dataset_name}/tracks.conf"
+  ]]></command>
+  <inputs>
+    <param name="jbrowse" type="data" format="html" label="JBrowse HTML Output" />
+    <param name="name" type="text" label="Display name" optional="False" />
+    <param name="unique_id" type="text" label="Unique ID" help="will be used in HTTP links (default=auto generated id)" optional="True">
```
erasche: Is there an overwhelming reason to expose this? I know the ID is visible in the URL, but I have to think it'll cause headaches.

If the user can configure it, they can re-use previously used names/numbers. This will lead to undesirable outcomes:

```
$ ls 5 5/data
5:
data  names  seq  trackList.json  tracks  tracks.conf

5/data:
names  seq  trackList.json  tracks  tracks.conf
```

Even though I know biologists really love being able to customize things and think that URLs are hugely important and must have the name in them because random IDs are bad (or so my PI yells at me for on a regular basis), I think this field gives them more ways to screw up that they won't understand compared to just forcing it to use `$jbrowse.id`

tools/jbrowse/jbrowse_to_existing.xml
```diff
@@ -0,0 +1,42 @@
...
+        <validator type="regex" message="Value must be a not empty string composed by alphanumeric characters and underscores">^\w+$</validator>
+    </param>
+  </inputs>
+  <outputs>
+    <data format="json" name="output"/>
```
erasche: no stdout, json seems illogical here. If the admin is expected to set `GALAXY_JBROWSE_SHARED_DIR` for their galaxy, maybe they should also set `GALAXY_JBROWSE_SHARED_URL`, and this tool should output an HTML / iframe that points to that URL (or something similar that gets them to be able to see their output datasets in production)

abretaud:
Thanks for the review, gonna fix things in the coming days!

For the jbrowse_to_existing.xml tool, I agree this is a little awkward, and maybe it would be better outside of IUC (with the apollo tools maybe?).

It's really intended to be used mainly/only for the genome annotation stuff, with a jbrowse container running and a shared folder between it and the galaxy container.

That's the solution I ended up with, because :
 - I need to access a jbrowse instance at a fixed address (cool if it's readable, but not essential) as I will make some links to it from other places (tripal, groene-thee one day?, ...). And if I update my jbrowse (new tracks, ...) it should be at the same location (to avoid hours of support for bookmarks problems + update of all links pointing to it everywhere)
- I need to have a unique jbrowse with multiple organisms (for variants of the same species)

I think sharing a galaxy dataset link would not allow the first item (but maybe I'm wrong?).
And currently I don't think it's possible to do the second point with the galaxy tools (jbrowse.xml) without making them even more complicated.

I have in mind to be able to deploy something like aphidbase.org (=jbrowse, apollo, tripal/groene-thee, & co) with just a docker-compose.yml and a galaxy workflow to populate everything, so I thought this small tool would resolve my constraints happily.

*But* I'm totally open to any other suggestion, and your opinion on that is really appreciated!

abretaud:
Trying to write a conda recipe for jbrowse (to ease testing).
Is [this patch](https://gist.githubusercontent.com/erasche/9cf020138d787eb47891/raw/e6ac8852be51c4869bd10d1977d2f68c731ffbc1/dojo.js.patch) mandatory? This is the only thing specific to galaxy

erasche:
@abretaud the patch is not mandatory.

JBrowse does semi-reasonable tracking in order to collect usage statistics for their grant, that patch is in reference to https://github.com/GMOD/jbrowse/issues/674.

tools/jbrowse/jbrowse_to_existing.xml
erasche: Ahhhh xref https://github.com/erasche/docker-jbrowse/pull/3

erasche:
> For the jbrowse_to_existing.xml tool, I agree this is a little awkward, and maybe it would be better outside of IUC (with the apollo tools maybe?).

Always welcome there :) Given that IUC tools are supposed to be "best practices" (someone else's tool got rejected on those grounds recently) and there are no ways the IUC handles this sort of outside-of-galaxy, data local to the server... maybe it would be more appropriate there. (Other IUC people are welcome to chime in with their thoughts!)

> It's really intended to be used mainly/only for the genome annotation stuff, with a jbrowse container running and a shared folder between it and the galaxy container. 

Aha! Now the "datasets.conf" that I complained about makes sense xref https://github.com/erasche/docker-jbrowse/pull/3 of course. I was wondering why you specified a non-default-loading file.

> I need to access a jbrowse instance at a fixed address (cool if it's readable, but not essential) as I will make some links to it from other places (tripal, groene-thee one day?, ...).

Sure, totally reasonable, a single jbrowse per site makes a lot more sense than N instances that can't be cached the same way in galaxy.

>And if I update my jbrowse (new tracks, ...) it should be at the same location (to avoid hours of support for bookmarks problems + update of all links pointing to it everywhere)

Yep, yep.

> I need to have a unique jbrowse with multiple organisms (for variants of the same species)
>
> I think sharing a galaxy dataset link would not allow the first item (but maybe I'm wrong?).
And currently I don't think it's possible to do the second point with the galaxy tools (jbrowse.xml) without making them even more complicated.

Yes, agreed, this is the correct solution for your problem.

Hmm, you could add it [to the action list](https://github.com/galaxyproject/tools-iuc/blob/master/tools/jbrowse/jbrowse.xml#L275) I suppose, but keeping a separate tool makes more sense to me (everyone wants jbrowse, not everyone is doing this particular action, it makes it easier for us to keep the non-reproducible stuff in a separate tool)

> I have in mind to be able to deploy something like aphidbase.org (=jbrowse, apollo, tripal/groene-thee, & co) with just a docker-compose.yml and a galaxy workflow to populate everything, so I thought this small tool would resolve my constraints happily.

Yep, that's the idea here too ;) I'll be shipping a docker-compose.yml + rancher-compose.yml in the future, for rancher folks (since we just transitioned to that here).

abretaud:
Oh yes, it's completely dependent on erasche/docker-jbrowse#3 I should have added the xref
I'm polishing the jbrowse.xml and I'll see later where I can push the new tool, maybe a new https://github.com/galaxy-genome-annotation/galaxy-tools repo?

erasche:
new repo - sounds good. I have internal changes to the tools that make use of code files to provide an organism selection dialog, I need to push those soon as well.

abretaud:
Before leaving, I just committed the conda testing stuff, and removed jbrowse from .tt_blacklist
It should become green(er) once bioconda/bioconda-recipes#2722 is merged
I'll continue working on this when I'm back

abretaud:
Hi,
I modified the jbrowse.xml tool, I think it's ready for a new review now, don't hesitate if I forgot something! The tests pass on my machine, it should get green here once bioconda/bioconda-recipes/pull/2860 is pushed to the bioconda channel.

For the moment I left untouched the jbrowse_to_existing.xml, I'll remove it from here after pushing it to the new repo. If we name it https://github.com/galaxy-genome-annotation/galaxy-tools, can you create it, I don't have the rights for it?

abretaud:
Green \o/
I didn't update the version numbers, do you prefer a 0.6 or a 0.5.1? (there's also @WRAPPER_VERSION@ token, but it's not used apparently)

bgruening:
Using @wrapper_version@ would be cool :)

erasche:
Hmm, let's go for 0.5.1 for now since there's a roadmap of features I
wanted for the minor version release.

Will review shortly, there were distractions here. Did I really never use
wrapper_version? Sigh @ me.

Den ons. 9. nov. 2016, 20.15 skrev Björn Grüning <notifications@github.com>:

tools/jbrowse/jbrowse.xml
```diff
@@ -475,84 +471,265 @@ $trackxml &&
...
-      <param name="track_groups_3|data_tracks_0|match_part|match_part_select" value="false"/>
-      <param name="track_groups_3|data_tracks_0|jbcolor_scale|color_score|color_score_select" value="none"/>
-      <param name="track_groups_3|data_tracks_0|jbcolor_scale|color|color_select" value="automatic"/>
-      <param name="track_groups_3|data_tracks_1|data_format" value="gene_calls"/>
-      <param name="track_groups_3|data_tracks_1|annotation" value="gff3/2.gff"/>
-      <param name="track_groups_3|data_tracks_1|match_part|match_part_select" value="true"/>
-      <param name="track_groups_3|data_tracks_1|match_part|name" value="cDNA_match"/>
-      <param name="track_groups_3|data_tracks_1|jbcolor_scale|color_score|color_score_select" value="none"/>
-      <param name="track_groups_3|data_tracks_1|jbcolor_scale|color|color_select" value="automatic"/>
+      <repeat name="track_groups">
```
erasche: Oooh thanks! I never figured out how to do this, so thanks for the working example. These tests are much nicer to read!!

abretaud:
Great! The changelog is updated now
So I guess we can merge if noone else have remarks on this?

bgruening:
One more tool under testing, how awesome!
Thanks all!

PR #997, opened 2016-10-13
correct name on input parameter

elenimijalis:


erasche:
Thanks @elenimijalis!

tools/seqtk/seqtk_subseq.xml
```diff
@@ -32,7 +32,7 @@ $in_file
         <param name="in_bed" type="data" format="bed" label="Input BED file"/>
       </when>
       <when value="name">
-        <param name="name_list" type="data" format="txt" label="Input fasta file"/>
+        <param name="name_list" type="data" format="txt,tabular" label="Input ID list"/>
```
nsoranzo: I'm not sure adding `tabular` here is needed/appropriate:
- documentation of `seqtk subseq` says "one sequence name per line"
- `Tabular` is a subclass of `Text`, so all tabular files should match the filter anyway.

elenimijalis: removed

PR #1009, opened 2016-10-25
Add py2bit-0.2.0, which will be needed by the next deepTools release

dpryan79:
py2bit will be required by the next deepTools release, so please consider this PR so that it's already available once that comes out.

As an aside, this is a convenient package for reading 2bit files that works with python 2 or 3, if anyone happens to need to do such a thing.


bgruening:
Thanks Devon!

dpryan79:
Thanks Björn!

PR #1018, opened 2016-11-01
Quast improvement : Dataset names

Delphine-L:
Modification of Quast wrapper to improve user experience :  instead of displaying the files names as created by galaxy, the columns reports use the dataset names as seen in histories.


bgruening:
Can you please increase the version number of the tool. This is needed as soon as you touch the parameters and the result can potentially change.

bgruening:
Thanks this looks really nice!

tools/quast/quast.xml
```diff
@@ -31,6 +31,9 @@
             --est-ref-size $input_size
         #end if
         --min-contig $min_contig
+        -l
+        #set names = ','.join( ['"'+str( $x.input.element_identifier)+'"' for $x in $files ])
+        $names
```
jmchilton: I think these need to be escaped to prevent shell attacks using malicious element identifiers.

``['"'+ re.sub('[^\w\-_]', '_', str( $x.input.element_identifier))+'"' for $x in $files ])``

bgruening: Should this not be handled at a framework level?

jmchilton: It isn't as far as I am aware. If you wish to have a longer conversation we can move this to Galaxy.

bgruening: `longer conversation` sounds scary ;)
Just wanted to point out that it will be hard to tell people this hack and that it would make things easier if such attacks can be prohibited on the framework level.

@Delphine-L I think for now you need to include it, sorry for the too fast merge.

Issue #1016, opened 2016-10-31
featureCounts with built-in annotations?


dpryan79:
It would be nice if the IUC featureCounts wrapper could use built-in annotations (presumably from the `gene_sels.loc` table). I'd be happy to try to put together a PR to add this functionality if others also found it useful (it's in the original wrapper from @yhoogstrate, but that doesn't handle dataset collections, while this one does). Thoughts?

lparsons:
Annotations in featureCounts would be very nice (also in DeSeq2). I didn't know about the `gene_sels.loc` table, what tool(s) use it?

bgruening:
:+1:

dpryan79:
@lparsons That should have been `gene_sets.loc`, sorry about the typo. I know the original featureCounts wrapper used the gene_sets table, but aside from that I haven't a clue if anything else does.

yhoogstrate:
I Made the tablet name and structure up myself So it's not according to any standard afaik and probably unique. I did not include it in the featurecounts PR because I thought Nobody else would like to use tables for gtf files (as they update so frequently) but apparently i was wrong. Great initiative @dpryan79

PR #1026, opened 2016-11-09
datatable support in featureCounts

dpryan79:
This implements my request in #1016, namely that featureCounts support datatables, so people needn't keep uploading the same GTF file.

### DIFF
--- a/tools/featurecounts/featurecounts.xml
+++ b/tools/featurecounts/featurecounts.xml
@@ -8,7 +8,12 @@
     <command><![CDATA[
         ## Check whether all alignments are from the same type (bam || sam)
         featureCounts
-            -a "$reference_gene_sets"
+            #if $source.ref_source=="history":
+                -a "$source.reference_gene_sets"
+            #else:
+                -a "$source.reference_gene_sets_builtin.fields.path"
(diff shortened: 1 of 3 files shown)


bgruening:
Looks great to me.
ping @yhoogstrate

bgruening:
I guess it would make sense to add the same patch to sailfish and salmon and deseq2 - opinions?

dpryan79:
What is the DESeq2 wrapper doing with an annotation?

bgruening:
More tximport before DESeq2: https://github.com/galaxyproject/tools-iuc/blob/master/tools/deseq2/deseq2.xml#L114

dpryan79:
Ah tximport, that makes sense.

@yhoogstrate: I thought a bit about filtering by dbkey, but isn't it the case that dataset collections don't have one? I recall that being the main issue we're having with the non-IUC wrapper when using collections (we get an internal server error).

yhoogstrate:
In that case we leave it as it is. Thanks @dpryan79 @bgruening ! If it turns out to be convenient to use `gene_sets.loc` you could make a PR to the galaxy core code as well.

PR #1027, opened 2016-11-10
Update/fix KOBAS wrapper

blankclemens:
The input field only allowed fasta files - fixed to support tabular and text files as well.


bgruening:
Thanks!

PR #1039, opened 2016-11-16
Trinity analyze diff expr

loraine-gueguen:
Add script analyze_diff_expr.pl in the Trinity suite. With tests.
(And I have slightly modified the tool name in run_de_analysis.xml from "Differential expression analysis" to "Run differential expression analysis" in order to distinguish the 2 tools)


cganote:
I hope this goes through. I need the --max_genes_clust param and gkumar's toolshed repo doesn't have it.

abretaud:
That's a very cool addition!
For the tool names, I fear having both 'Analyze differential expression' and 'Run differential expression analysis' tools would be confusing. What about renaming the new tool something like 'Extract and cluster differentially expressed transcripts', like [here](https://github.com/trinityrnaseq/trinityrnaseq/wiki/Trinity-Differential-Expression#extracting-and-clustering-differentially-expressed-transcripts)?

tools/trinity/analyze_diff_expr.xml
```diff
@@ -0,0 +1,179 @@
...
+    </macros>
+    <expand macro="requirements">
+    </expand>
+    <expand macro="stdio"/>
+    <command><![CDATA[
+
+	## DE results input files must be in the working directory and have suffix .DE_results
+	#for $input in $DE_results
+
+            ln -s "${input}" "${input.element_identifier}.DE_results"
```
abretaud: Needs to be escaped: https://github.com/galaxyproject/tools-iuc/pull/1040#issuecomment-261020210

tools/trinity/analyze_diff_expr.xml
```diff
@@ -0,0 +1,179 @@
...
+ConditionB  CondB_replicate2
+----------- ----------------
+ConditionC  CondC_replicate1
+----------- ----------------
+ConditionC  CondC_replicate2
+----------- ----------------
+ConditionC  CondC_replicate3
+=========== ================
+
+It will probably be the same file as used in the tool 'RNASeq samples quality check for transcript quantification' or in the tool 'Run differential expression analysis'.
```
abretaud: As there will be a new tool in #1040 to generate this file, I'd change a little the doc here as in
https://github.com/galaxyproject/tools-iuc/pull/1040/files#diff-3018259e54236f14dff9d812b196b366

loraine-gueguen:
@abretaud Thanks! I have made the changes.
My tests were fine on Planemo, but no more on Travis. I try solve that.
And a test on align_and_estimate_abundance.xml now fails too (as in your PR #1040, `salmon: error while loading shared libraries: libboost_iostreams.so.1.60.0: cannot open shared object file: No such file or directory`)

tools/trinity/analyze_diff_expr.xml
```diff
@@ -0,0 +1,180 @@
+<tool id="trinity_analyze_diff_expr" name="Extracting and clustering differentially expressed transcripts" version="@WRAPPER_VERSION@.0">
```
abretaud: I'd write "Extract and cluster differentially expressed transcripts" here

tools/trinity/analyze_diff_expr.xml
```diff
@@ -0,0 +1,180 @@
...
+    <description>using a Trinity assembly</description>
```
abretaud: "from a Trinity assembly"

abretaud:
Yep, there is a problem with salmon, and updating to 0.7.2 didn't fixed it.
I suspect it's a problem with conda symlinks, but not sure

tools/trinity/analyze_diff_expr.xml
loraine-gueguen: That's not good for the tool, because is looking for input files with a pattern contaning dots (`$result_file =~ /matrix\.(\S+)_vs_(\S+)\.[^\.]+\.DE_results$/`)
May I change the escape pattern to `${re.sub('[^\w\-_.]', '_', input.element_identifier)}` ?

loraine-gueguen:
ok, it is better with those requirements (one test passed). But I do not understand why my tests where fine on my machine with planemo test without those requirements.

nsoranzo:
> But I do not understand why my tests where fine on my machine with planemo test without those requirements.

@loraine-gueguen You probably had these R packages installed on the system R library path.

loraine-gueguen:
@nsoranzo  I use arg `--conda_dependency_resolution` in planemo test so it should not use the system env, should it ?

nsoranzo:
@loraine-gueguen This will try to resolve the dependencies with Conda, but it does not isolate from the system env, you need to test under docker to do that.

loraine-gueguen:
@nsoranzo ok

loraine-gueguen:
My tests passed!

bgruening:
Is the salmon error resolved? Or do I need to fix something?

Nice work!!!

loraine-gueguen:
For the salmon error, @abretaud has disabled the test that fails on travis in the PR #1040, which was merged today. So everything is fine for me now.

abretaud:
For the salmon problem it really looks like a conda/symlink problem. The test run fine on my clean machine and it used to pass when it was merged...
I can have a look at it, but after the training session :)

tools/trinity/analyze_diff_expr.xml
```diff
@@ -0,0 +1,185 @@
...
+
+        --matrix "${matrix}"
+
+        --samples "${samples}"
+
+	-P ${p}
+
+	-C ${c}
+
+       	#if $additional_params.max_DE_genes_per_comparison:
```
nsoranzo: Since `$additional_params.max_DE_genes_per_comparison` is an optional integer, it should be converted to string before testing:
```
       	#if str($additional_params.max_DE_genes_per_comparison)
```

tools/trinity/analyze_diff_expr.xml
```diff
@@ -0,0 +1,185 @@
...
+            <param name="order_columns_by_samples_file" argument="--order_columns_by_samples_file" type="boolean" checked="false" truevalue="--order_columns_by_samples_file" falsevalue="" label="Order columns by samples file" help="Instead of clustering samples or replicates hierarchically based on gene expression patterns, order columns according to order in the --samples file."/>
+            <param name="max_genes_clust" argument="--max_genes_clust" type="integer" value="10000" label="Maximum genes in cluster" help="If more than 10000, heatmaps are not generated, since too time consuming"/>
+            <conditional name="GO_enrichment">
+                <param type="boolean" name="examine_GO_enrichment" argument="--examine_GO_enrichment" checked="false" truevalue="--examine_GO_enrichment" falsevalue="" label="Run GO enrichment analysis" help="To examine GO enrichment, you must first run Trinotate and then extract all GO assignments for each gene feature, with the Trinotate script extract_GO_assignments_from_Trinotate_xls.pl"/>
```
nsoranzo: Boolean are discouraged for conditionals, because a checkbox is not expected to change the content of the mask from a user point of view, see https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#booleans

Can you change `examine_GO_enrichment` to be a select param?

tools/trinity/analyze_diff_expr.xml
```diff
@@ -0,0 +1,185 @@
...
+----------- ----------------
+ConditionC  CondC_replicate1
+----------- ----------------
+ConditionC  CondC_replicate2
+----------- ----------------
+ConditionC  CondC_replicate3
+=========== ================
+
+This file can be generated with the 'Describe samples and replicates' tool.
+It will probably be the same file as used in the tool 'RNASeq samples quality check for transcript quantification' or in the tool 'Run differential expression analysis'.
```
nsoranzo: s/Run differential expression analysis/Differential expression analysis/

tools/trinity/analyze_diff_expr.xml
```diff
@@ -0,0 +1,185 @@
...
+The names in column 2 must match the names given in the tool 'Build expression matrix for a de novo assembly of RNA-Seq data by Trinity'.
+
+You must also provide as a data collection the files resulting from the differential expression analysis (outputs of tool 'Run differential expression analysis').
```
nsoranzo: s/Run differential expression analysis/Differential expression analysis/

tools/trinity/analyze_diff_expr.xml
loraine-gueguen: ok, done

tools/trinity/analyze_diff_expr.xml
loraine-gueguen: Done. I have forgotten to change back the name of the tool.

tools/trinity/analyze_diff_expr.xml
loraine-gueguen: done

loraine-gueguen:
`The job exceeded the maximum time limit for jobs, and has been terminated.`
Is it possible to launch the test job again ? Thanks

bgruening:
Restarted.

bgruening:
@abretaud @nsoranzo ready for the merge?

abretaud:
:+1: for me!

nsoranzo:
Thanks a lot for the contribution @loraine-gueguen!

PR #1019, opened 2016-11-02
Quast : escape identifiers to prevent malicious attacks

Delphine-L:

PR #1040, opened 2016-11-16
Trinity: data collection everywhere + new describe_samples tool

abretaud:
Hi all!
A quick PR before a small training session next week...
Now abundance_estimates_to_matrix is able to work with data collection produced by previous tools in the trinity workflow (run_de_analysis)
Also added a small tool to generate the tabular file needed to describe the samples (ping @cganote)


jmchilton:
I'd checkout https://github.com/galaxyproject/tools-iuc/pull/1019 and http://planemo.readthedocs.io/en/latest/writing_advanced.html#processing-identifiers. If you are going to use element_identifier on the command-line I think it needs to be escaped.

tools/trinity/abundance_estimates_to_matrix.xml
```diff
@@ -19,16 +20,11 @@
...
         #end for
     ]]></command>
     <inputs>
-        <repeat name="samples" title="Abundance estimates for samples">
-            <param name="file" label="Add file" type="data" format="tabular"/>
-            <param name="sample_name" label="Sample name" type="text">
-                <validator type="regex" message="Value must be a not empty string composed by alphanumeric characters and underscores">^\w+$</validator>
-            </param>
-        </repeat>
+        <param name="samples" label="Abundance estimates" type="data" multiple="true" format="tabular" help="output(s) from 'Align reads and estimate abundance tool'" />
```
nsoranzo: s/'Align reads and estimate abundance tool'/'Align reads and estimate abundance' tool/

abretaud:
Cool, I didn't know this, I should do it in stacks wrappers too

By the way, if in my history I have a (list) collection, is there a way to launch a tool on a single element of the collection (and/or extract it to a normal dataset)? I couldn't find it

tools/trinity/describe_samples.xml
```diff
@@ -0,0 +1,68 @@
...
+<tool id="describe_samples" name="Describe samples" version="@WRAPPER_VERSION@.0">
+    <description>and replicates</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <command detect_errors="aggressive"><![CDATA[
+        cp $samples $output
+      ]]></command>
+    <configfiles>
+        <configfile name="samples"><![CDATA[#for $sample in $samples:
```
nsoranzo: I think that using the name `samples` for both the `repeat` and the `configfile` is a bit confusing.

nsoranzo:
Failing on Travis with:
```
salmon: error while loading shared libraries: libboost_iostreams.so.1.60.0: cannot open shared object file: No such file or directory
```

abretaud:
Yes, it's strange because the lib is in the salmon package...
Could it be that `conda install` uses symlinks in travis jobs? I can reproduce the error here with symlinks

abretaud:
Ok, disabled the test that fails on travis, I guess this should be fixed in conda at some point...
Any objections to merge this anyone?

jmchilton:
> By the way, if in my history I have a (list) collection, is there a way to launch a tool on a single element of the collection (and/or extract it to a normal dataset)? I couldn't find it

Not currently unfortunately :(.

PR #1043, opened 2016-11-21
use single quotes in travis.yml

shiltemann:
Ran into this when using this testing in our own repo and our randomly generated password had some special characters in it. 

Any reason to not just use single quotes everywhere?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,19 +12,19 @@ env:
 before_install:
   - export GALAXY_REPO=https://github.com/galaxyproject/galaxy
   - export GALAXY_RELEASE=release_16.07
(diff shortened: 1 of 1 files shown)


peterjc:
Assuming TravisCI is happy, +1 from me.

Using single quotes was recently adopted as IUC best practise for the ``<command>`` tag for much the same reason.

nsoranzo:
@shiltemann In bash you cannot use single quotes if you want the env variables to be interpreted, which is the case here.

peterjc:
Do you know which special character was triggering this? That would help for testing.

shiltemann:
@nsoranzo of course, I am an idiot ;P guess thinking too much of cheetah. So guess only solution is make sure password contains characters that play nice?

PR #1055, opened 2016-12-01
Jbrowse: python3 + new options

abretaud:
A tiny fix for a problem with the jbrowse script and python3...

### DIFF
--- a/tools/jbrowse/jbrowse.py
+++ b/tools/jbrowse/jbrowse.py
@@ -264,11 +264,11 @@ def etree_to_dict(t):
     if children:
         dd = defaultdict(list)
         for dc in map(etree_to_dict, children):
-            for k, v in dc.iteritems():
+            for k, v in dc.items():
                 dd[k].append(v)
-        d = {t.tag: {k: v[0] if len(v) == 1 else v for k, v in dd.iteritems()}}
+        d = {t.tag: {k: v[0] if len(v) == 1 else v for k, v in dd.items()}}
     if t.attrib:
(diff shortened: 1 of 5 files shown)


abretaud:
Oh, hang on before merging, there are other python3 problems. Looks like there were some changes in the conda deps, I'll fix this tomorrow

tools/jbrowse/jbrowse.py
```diff
@@ -370,7 +370,7 @@ def _add_json(self, json_data):
         self.subprocess_check_call(cmd)
 
     def _add_track_json(self, json_data):
-        if len(json_data.keys()) == 0:
+        if len(list(json_data.keys())) == 0:
```
nsoranzo: That's not needed: https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects

The simplest code is `len(json_data)` anyway.

abretaud: Oh yes, I blindly trusted 2to3 on this

abretaud:
I think it's ready for merge now, it worked fine with some real data tonight

abretaud:
As I needed some additional options, I just added them to this PR.
They are described there: http://gmod.org/wiki/JBrowse_FAQ#My_CanvasFeatures_don.27t_show_up_with_subfeatures.2C_why_not.3F

nsoranzo:
Python3 support looks good, I'll leave the rest to @erasche.

bgruening:
Let's leave @erasche rest a little bit and merge it directly. Looks great to me!

PR #1060, opened 2016-12-02
migrate unconflicting devteam packages

martenson:
the rest is in namespace conflict:

```
drwxr-xr-x   4 staff   136B Nov  2  2015 package_bx_python_0_7/
drwxr-xr-x   4 staff   136B Nov  2  2015 package_ccat_3_0/
drwxr-xr-x   4 staff   136B Nov  2  2015 package_libxslt_1_1_28/
drwxr-xr-x   4 staff   136B Nov  2  2015 package_numpy_1_7/
drwxr-xr-x   4 staff   136B Nov  2  2015 package_readline_6_2/
```


bgruening:
@martenson have you checked if the rest is identical?

martenson:
@bgruening they are not always identical, some just have namespace collisions :/

packages/package_bamtools_2_3_0_2d7685d2ae/.shed.yml
```diff
@@ -0,0 +1,14 @@
...
+long_description: |
+    Binary files in this package are compiled from source code version2d7685d2ae
+    This is package dependency for tools relying on bamtools toolkit developed by
+    Derek Barnett. This package is distributed as x86_64 binaries only. 
+    These binaries should work on any of our stated supported linux platforms 
+    other than RHEL/CentOS 5.
+name: package_bamtools_2_3_0_2d7685d2ae
+owner: devteam
+homepage_url: https://github.com/pezmaster31/bamtools
+remote_repository_url: https://github.com/galaxyproject/tools-devteam/tree/master/packages/package_bamtools_2_3_0_2d7685d2ae
```
martenson: good call

packages/package_picard_1_126_0/.shed.yml
```diff
@@ -0,0 +1,8 @@
+categories:
```
mvdbeek: Should we perhaps skip packages for tools that are conda-only by now?

nsoranzo:
@martenson Is this connected to https://github.com/galaxyproject/tools-devteam/issues/406 ? I'm not sure IUC wants to maintain even more packages (we already have 367) since we decided to concentrate on Conda.

martenson:
@nsoranzo Partially. It is a selfish move too - IUC does a way better job at maintaining than devteam and the repo's automatic test setup and maintenance is also superior. I am perfectly ok with killing this PR if there is resistance though.

bgruening:
As part of to migration from devteam to here, I like it - but they should be removed from devteam and a placeholder link should be established pointing to this repo.

martenson:
I will do all this if IUC approves of the general idea.

mvdbeek:
I am -0.5 on moving any package at all because 
  - changes of package ownership is notoriously bad for repository-repository dependency relationships in galaxy
  - I don't think we want to maintain packages
  - the IUC infrastructure will not benefit package quality at all

I am :+1: to maintain devteam tools that have conda dependencies (or no dependencies at all).

peterjc:
Overall I'm also slightly against this change.

If the IUC github has better TravisCI configuration and Tool Shed integration for automated uploads, we (the IUC) can surely help bring that to the Galaxy devteam?

I appreciate there is overlap between the two teams, but I think it would be clearer if in general things on the Tool Shed under devteam lived in https://github.com/galaxyproject/tools-devteam - and things on the Tool Shed under iuc lived here in https://github.com/galaxyproject/tools-iuc

On the other hand, the IUC has adopted a number of tools and packages which remain on the Tool Shed under the original creator's username. That is a precent we could follow with devteam tools too?

bgruening:
I think tools are a different story and we should probably move all tools that are still supported and maintained into IUC - this makes it a lot simpler for us.

> If the IUC github has better TravisCI configuration and Tool Shed integration for automated uploads, we (the IUC) can surely help bring that to the Galaxy devteam?

We try this and we update devteam, but it constantly slippes under (at least) my radar.

For the packages I'm not sure, there were a bunch of bulk-update to IUC packages which never went into devteam so it would make sense to have them here as well. I think the question boils down to which degree we want to maintain these packages in the future and if we expect such bulk-updates and maintenance.

peterjc:
OK, having the IUC adopt many/most/all of the devteam owned *package* definitions is more reasonable. I withdraw my (slight) objection.

yhoogstrate:
For some rather commonly used tools (https://github.com/galaxyproject/tools-devteam/issues/406#issuecomment-257566124 thanks for mentioning them @martenson) I think it could be beneficial to have them (also?) under the radar here, in particular to get fixes or PRs quicker reviewed. For the deprecated tool_dependencies packages I don't see a clear advantage.

bgruening:
@martenson let's keep the packages were they are and hopefully we don't need to deal with them to much in the future. For tools this is a different story and we should migrate them. Would be great if you can create similar PR for devteam tools.

PR #1061, opened 2016-12-05
Add deepTools-2.4.1

dpryan79:
This adds deepTools-2.4.1, which is also now available via conda.

@bgruening Can you upload the wrappers to the toolshed too? There are some people at EMBL waiting on this version to handle some issues they're experiencing.

### DIFF
--- a/packages/package_python_2_7_deeptools_2_4_1/tool_dependencies.xml
+++ b/packages/package_python_2_7_deeptools_2_4_1/tool_dependencies.xml
@@ -21,7 +21,7 @@
     <package name="python" version="2.7.10">
         <repository name="package_python_2_7_10" owner="iuc" prior_installation_required="True" />
     </package>
-    <package name="deepTools" version="2.4.1">
+    <package name="deeptools" version="2.4.1">
         <install version="1.0">
             <actions>
                 <action type="setup_python_environment">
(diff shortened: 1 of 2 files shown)


packages/package_python_2_7_deeptools_2_4_1/tool_dependencies.xml
```diff
@@ -0,0 +1,71 @@
...
+    <package name="scipy" version="0.14">
+        <repository name="package_python_2_7_scipy_0_14" owner="iuc" prior_installation_required="True" />
+    </package>
+    <package name="pysam" version="0.8.3">
+        <repository name="package_python_2_7_pysam_0_8_3" owner="iuc" prior_installation_required="True" />
+    </package>
+    <package name="python" version="2.7.10">
+        <repository name="package_python_2_7_10" owner="iuc" prior_installation_required="True" />
+    </package>
+    <package name="deepTools" version="2.4.1">
```
dpryan79: Good catch! I bet we've been doing this for a while...

nsoranzo: Conda package names are not case-sensitive, for what I know.

dpryan79: For what it's worth, case seems to be ignored at some point since all of the planemo tests that use this pass, but you're correct that we should fix this now before this inconsistency ends up biting us. I'll add a WIP tag until this is all changed.

bgruening:
This is not WIP anymore, correct?

dpryan79:
@bgruening: Correct, I've also modified the branches on the deepTools repository, so the case will be consistent in the wrappers (I just merged that change into all the branches).

bgruening:
Thanks!

bgruening:
Everything updated!

dpryan79:
Thanks, updating all of our servers now!

PR #1062, opened 2016-12-06
Add wrapper for rnaSPAdes.

davebx:
Based on Lionel Guy's SPAdes wrapper.


tools/rnaspades/rnaspades.xml
```diff
@@ -0,0 +1,195 @@
+<tool id="rnaspades" name="rnaSPAdes" version="1.0.0">
```
bgruening: version 3.9.0?

tools/rnaspades/rnaspades.xml
```diff
@@ -0,0 +1,195 @@
...
+        <exit_code range="1:" />
+    </stdio>
+    <command>
+        <![CDATA[
+        rnaspades.py -o .
+        ## Forces unzipped output, faster
+        --disable-gzip-output
+        $draft $onlyassembler -t \${GALAXY_SLOTS:-4} $iontorrent
+        ## Sequence files, libraries
+        #for $i, $library in enumerate( $libraries ):
```
bgruening: enumerate can take a start parameter: https://docs.python.org/2/library/functions.html#enumerate

tools/rnaspades/rnaspades.xml
```diff
@@ -0,0 +1,195 @@
...
+                    --$prefix$num-2 fastq:$file.file_type.fastq_collection.reverse
+                #end if
+            #end for
+        #end for
+        ## PacBio reads
+        #for $i, $pacbiolib in enumerate( $pacbio )
+            --pacbio fastq:$pacbiolib.pacbio_reads
+        #end for
+        ## Nanopore
+        #for $i, $nanoporelib in enumerate( $nanopore )
```
bgruening: $i is not used?

tools/rnaspades/rnaspades.xml
```diff
@@ -0,0 +1,195 @@
...
+                    --$prefix$num-s fastq:$file.file_type.unpaired_reads
+                #elif $file.file_type.type == "paired-collection":
+                    --$prefix$num-1 fastq:$file.file_type.fastq_collection.forward
+                    --$prefix$num-2 fastq:$file.file_type.fastq_collection.reverse
+                #end if
+            #end for
+        #end for
+        ## PacBio reads
+        #for $i, $pacbiolib in enumerate( $pacbio )
+            --pacbio fastq:$pacbiolib.pacbio_reads
```
bgruening: $i is not used?

tools/rnaspades/rnaspades.xml
```diff
@@ -0,0 +1,195 @@
...
+        #end for
+        ## Nanopore
+        #for $i, $nanoporelib in enumerate( $nanopore )
+            --nanopore fastq:$nanoporelib.nanopore_reads
+        #end for
+        ## Sanger
+        #for $i, $sangerlib in enumerate( $sanger )
+            --sanger $sangerlib.file_type.type:$sangerlib.file_type.sanger_reads
```
bgruening: $i is not used?

tools/rnaspades/rnaspades.xml
```diff
@@ -0,0 +1,195 @@
...
+        #end for
+        ## Contigs
+        #for $i, $trustedcontigs in enumerate( $trustedcontigs )
+            --trusted-contigs $trustedcontigs.file_type.type:$trustedcontigs.file_type.trusted_contigs
+        #end for
+        #for $i, $untrustedcontigs in enumerate( $untrustedcontigs )
+            --untrusted-contigs $untrustedcontigs.file_type.type:$untrustedcontigs.file_type.untrusted_contigs
+        #end for
+        ;
+        if [ -f transcripts.fasta ] ; then mv transcripts.fasta $output_transcripts ; fi ;
```
bgruening: Can we remove this and use `from_work_dir` instead?

tools/rnaspades/rnaspades.xml
```diff
@@ -0,0 +1,195 @@
...
+            --sanger $sangerlib.file_type.type:$sangerlib.file_type.sanger_reads
+        #end for
+        ## Contigs
+        #for $i, $trustedcontigs in enumerate( $trustedcontigs )
+            --trusted-contigs $trustedcontigs.file_type.type:$trustedcontigs.file_type.trusted_contigs
+        #end for
+        #for $i, $untrustedcontigs in enumerate( $untrustedcontigs )
+            --untrusted-contigs $untrustedcontigs.file_type.type:$untrustedcontigs.file_type.untrusted_contigs
+        #end for
+        ;
```
bgruening: Please use && for concatenating commands.

tools/rnaspades/rnaspades.xml
```diff
@@ -0,0 +1,195 @@
...
+        if [ -f transcripts.fasta ] ; then mv transcripts.fasta $output_transcripts ; fi ;
+        if [ -f spades.log ] ; then mv spades.log $logfile ; fi ;
+        ]]>
+    </command>
+    <inputs>
+        <param checked="False" falsevalue="" label="Draft assembly. Faster, but more error-prone" name="draft" truevalue="--draft-assembly" type="boolean" />
```
bgruening: any reason not to use argument=""?

tools/rnaspades/rnaspades.xml
```diff
@@ -0,0 +1,195 @@
...
+        </repeat>
+        <repeat name="pacbio" title="PacBio CLR reads">
+            <param format="fastq" help="FASTQ format. For PacBio pre-corrected or CCS reads, use single reads above." label="PacBio CLR reads." name="pacbio_reads" type="data" />
+        </repeat>
+        <repeat name="nanopore" title="Nanopore reads">
+            <param format="fastq" help="FASTQ format. For Nanopore, use single reads above." label="Nanopore reads." name="nanopore_reads" type="data" />
+        </repeat>
+        <repeat name="sanger" title="Sanger reads">
+            <conditional name="file_type">
+                <param help="No read correction is done on Sanger reads, no need to provide quality information." label="Select file format" name="type" type="select">
```
bgruening: This can be a macro. I'm wondering actually if it is needed at all and can not be replaces by `<param format="fasta,fastq" help="FASTA/Q format" label="Sanger reads" name="sanger_reads" type="data" />` and ask for $sanger_reads.ext for the filetype.

tools/rnaspades/rnaspades.xml
davebx: There's a very ~good~ bad reason I didn't use `from_work_dir`:

I couldn't remember the syntax.

tools/rnaspades/rnaspades.xml
davebx: I should be able to rework this into something with `multiple="True"`. Watch this space for updates.

tools/rnaspades/rnaspades.xml
davebx: Fixed.

tools/rnaspades/rnaspades.xml
davebx: Fixed.

tools/rnaspades/rnaspades.xml
davebx: The RNA iteration doesn't even support pacbio.

tools/rnaspades/rnaspades.xml
davebx: Or nanopore.

tools/rnaspades/rnaspades.xml
davebx: Or sanger.

tools/rnaspades/rnaspades.xml
davebx: Fixed.

tools/rnaspades/rnaspades.xml
davebx: Fixed.

tools/rnaspades/rnaspades.xml
davebx: Fixed.

tools/rnaspades/rnaspades.xml
davebx: New revision pushed.

tools/rnaspades/rnaspades.xml
```diff
@@ -0,0 +1,135 @@
...
+            </param>
+            <param label="Orientation" name="orientation" type="select">
+                <option selected="true" value="fr">-&gt; &lt;- (fr)</option>
+                <option value="rf">&lt;- -&gt; (rf)</option>
+                <option value="ff">-&gt; -&gt; (ff)</option>
+            </param>
+            <repeat min="1" name="files" title="Files">
+                <conditional name="file_type">
+                    <param label="Select file format" name="type" type="select">
+                        <option value="separate">Separate input files</option>
```
bgruening: What happens if you remove the repeat and the conditional and offer all files as multiple=True and optional=True ... Not good design?

How likely is it that you have multiple fastq files, one separate and one interleaved etc ... I feel multple=True is more usefule to include, so maybe just removing the repeat?

tools/rnaspades/rnaspades.xml
```diff
@@ -0,0 +1,135 @@
...
+        <repeat help="It is not possible to specify only mate-pair libraries. Scaffolds are not produced if neither a paired-end nor a mate-pair library is provided." min="1" name="libraries" title="Libraries">
+            <param label="Library type" name="lib_type" type="select">
+                <option value="paired_end">Paired-end / Single reads</option>
+                <option value="mate_paired">Mate pairs</option>
+                <option value="high_mate_paired">High Quality Mate pairs</option>
+                <option value="nxmate_paired">Lucigen NxMate pairs</option>
+            </param>
+            <param label="Orientation" name="orientation" type="select">
+                <option selected="true" value="fr">-&gt; &lt;- (fr)</option>
+                <option value="rf">&lt;- -&gt; (rf)</option>
```
bgruening: CDATA?

tools/rnaspades/rnaspades.xml
```diff
@@ -0,0 +1,135 @@
...
+                    </when>
+                </conditional>
+            </repeat>
+        </repeat>
+        <param optional="true" format="fasta,fastq" label="Trusted contigs" multiple="true" name="trusted_contigs" type="data" />
+        <param optional="true" format="fasta,fastq" label="Untrusted contigs" multiple="true" name="untrusted_contigs" type="data" />
+    </inputs>
+    <outputs>
+        <data format="fasta" label="rnaSPAdes transcripts" name="output_transcripts" from_work_dir="transcripts.fasta" />
+        <data format="txt" label="rnaSPAdes log" name="logfile" from_work_dir="spades.log" />
```
bgruening: Can we make this optional, or is the log really needed by every run?

tools/rnaspades/rnaspades.xml
```diff
@@ -0,0 +1,135 @@
...
+    </outputs>
+    <tests>
+        <test>
+            <param name="lib_type" value="paired_end" />
+            <param name="type" value="separate" />
+            <param name="fwd_reads" value="ecoli_1K_1.fq" ftype="fastq" />
```
bgruening: Can we make the test inputs smaller?

tools/rnaspades/rnaspades.xml
peterjc: You mean three optional multiple file inputs: Paired, Interleaved, Singles?

I think that would work BUT the Galaxy UI will auto-fill these with the most recent file of a suitable format, meaning the user must notice this and explicitly select no files. This is an unusual thing to have to do, and sounds error prone.

(I had to think about this for wrapping MIRA 4 and CLC Bio, but they have explicit read groups which work well for a repeat structure.)

bgruening: @peterjc it does not fill the select box if it is optional=True, isn't it?
But having all inputs optional is also error prone.

peterjc: @bgruening from memory it does pre-fill, but I have not checked on a recent Galaxy.

tools/rnaspades/rnaspades.xml
davebx: I'm not sure the log is needed at all.

bgruening:
Thanks @davebx!

lparsons:
Awesome, thanks @davebx! Are there plans for wrappers for other versions of SPADES? ("normal", metagenomics, etc.)? I'm willing to help with those, but don't want to duplicate effort.

nsoranzo:
@lparsons There are already wrappers on the MTS for "normal", metaspades, plasmidspades and rnaspades with owner nml (which should be @takadonet). It would be nice to have under IUC if there is interest.

lparsons:
There is certainly interest from me. @Takadonet, would you be amenable to moving these tools under the IUC?

Takadonet:
Sorry for the late reply. Yes I am amenable to moving all the spades repos over. The latest version is https://github.com/phac-nml/galaxy_tools .

bgruening:
Cool @Takadonet!

davebx:
SPAdes pull request is at #1088

PR #1066, opened 2016-12-09
Concept of beaconnetwork

joachimwolff:
This is a in development version of a wrapper for the beaconnetwork.


bgruening:
Thanks Jo!

bgruening:
We would like to get some initial ideas about the implementation and if you have any ideas how ew can represent the results to the user - essentially we get a very unstructured json back.

bgruening:
pin @kmurat1

PR #1067, opened 2016-12-10
Travis update to 16.10

bgruening:
As part of the next release we should switch to this next release and test tools against latest stable.
This PR should also test Galaxy 16.10.

@galaxyproject/iuc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,10 @@ python: 2.7
 env:
   - CHUNK=0
   - CHUNK=1
-  - CHUNK=2
-  - CHUNK=3
 
 before_install:
   - export GALAXY_REPO=https://github.com/galaxyproject/galaxy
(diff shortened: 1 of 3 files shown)


bgruening:
I have one :+1: so I gonna break our setup now :smile:

abretaud:
That's cool, with 16.10 and the new `use_cached_dependency_manager` + `--conda_copy_dependencies` in planemo, I think we could have a solution for the problems of symlinks during conda creation (RPATH stuff and timeout for trinity tests).

.travis.yml
```diff
@@ -6,12 +6,10 @@ python: 2.7
 env:
   - CHUNK=0
   - CHUNK=1
-  - CHUNK=2
-  - CHUNK=3
```
nsoranzo: This indeed broke Travis setup because `planemo ci_find_repos` still uses `--chunk_count 4` !

PR #1069, opened 2016-12-11
Cwpair2 conda

bgruening:
This should not depend on anaconda, which is one GB in size.

### DIFF
--- a/tools/cwpair2/cwpair2.xml
+++ b/tools/cwpair2/cwpair2.xml
@@ -5,7 +5,7 @@
         <import>cwpair2_macros.xml</import>
     </macros>
     <expand macro="requirements" />
-    <command>
+    <command detect_errors="aggressive">
         <![CDATA[
             python $__tool_directory__/cwpair2.py
             #for $i in $input:
--- a/tools/cwpair2/cwpair2_macros.xml
+++ b/tools/cwpair2/cwpair2_macros.xml
@@ -1,9 +1,9 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <macros>
-    <token name="@WRAPPER_VERSION@">1.0</token>
+    <token name="@WRAPPER_VERSION@">1.1</token>
     <xml name="requirements">
         <requirements>
-            <requirement type="package" version="2.3.0">anaconda</requirement>
+            <requirement type="package" version="1.5.3">matplotlib</requirement>
         </requirements>
     </xml>
     <xml name="stdio">
--- a/tools/cwpair2/tool_dependencies.xml
+++ b/tools/cwpair2/tool_dependencies.xml
@@ -1,5 +1,0 @@
-<tool_dependency>
-    <package name="anaconda" version="2.3.0">
-        <repository name="package_anaconda_2_3_0" owner="iuc" />
-    </package>
-</tool_dependency>


gregvonkuster:
Yes, this recipe and likely some others for which I'm responsible pre-date conda dependencies - I'll take a look asap.

gregvonkuster:
@bgruening what is the standard for handling the java dependency here?  Does it require a recipe that this tool would depend on?

bgruening:
What do you mean by recipe? You can simply add `java` as a conda dependency?

gregvonkuster:
BY recipe, I mean a cwpair bioconda-recipe, which would allow for an entry like this:
```
java-jdk >=7
```
If I do not use a separate recipe like this, am I forced to used a specific version in the &lt;requirements&gt;?
```
<requirement type="package" version="8.0.92">java-jdk</requirement>
```
I'm not sure if the Galaxy side allows for versionless requirements.

bgruening:
A I see, I'm always in favor of having a dedicated conda package for it :)

gregvonkuster:
Trying to figure out why this one is failing, I assume this is the error:
```
--------------------------------------
| Dataset Job Standard Error:

|  /tmp/tmpsUt84W/job_working_directory/000/4/conda-env/lib/python2.7/site-packages/matplotlib/font_manager.py:273: UserWarning: Matplotlib is building the font cache using fc-list. This may take a moment.
```
But I'm not able to get planemo to install conda dependencies if I don't have conda in my PATH:
```
planemo test --conda_dependency_resolution --conda_prefix /home/greg/_conda ./cwpair2.xml
...
functional test error:

  File "/home/greg/work/tools-iuc/cwpair2_conda/tools-iuc/tools/cwpair2/cwpair2.py", line 17, in <module>
    import cwpair2_util
  File "/home/greg/work/tools-iuc/cwpair2_conda/tools-iuc/tools/cwpair2/cwpair2_util.py", line 7, in <module>
    import matplotlib
ImportError: No module named matplotlib
```
If I use miniconda2 in my environment:
```
$ which conda
~/miniconda2/bin/conda
[greg@bigsky cwpair2]$ ~/miniconda2/bin/conda --version
conda 4.2.13
```
I get this exception:
```
galaxy.jobs.output_checker: DEBUG: Tool produced standard error failing job - [Traceback (most recent call last):
  File "/tmp/tmp11rMyk/galaxy-dev/tools/data_source/upload.py", line 17, in <module>
    from galaxy import util
  File "/tmp/tmp11rMyk/galaxy-dev/lib/galaxy/util/__init__.py", line 30, in <module>
    from six impo]
galaxy.jobs: DEBUG: (2) setting dataset 2 state to ERROR
```
Does planemo no longer support miniconda2?

bgruening:
@gregvonkuster planemo supports both miniconda2 and minconda3 ... only conda needs to be available.
The error looks like the matplotlib warining: https://travis-ci.org/galaxyproject/tools-iuc/jobs/183052613#L2875

Is treaded as error by planemo/Galaxy. So if use stdio or the detect_error attribute in the command tag, this should probably solve this problem.

gregvonkuster:
Yeah, that what I was going to try to test, but I cannot get planemo to work for me locally using conda, and I haven't been able to for some time.  :(  I'll add the change though and see if tests will pass.

gregvonkuster:
I see you've done it already.   Thanks.

bgruening:
@gregvonkuster what was the problem with Java? I have not defined it but tests are passing?

gregvonkuster:
@bgruening, no problem with Java on this one - sorry for the confusion.  But I think there are other recipes I've built that require anaconda and Java, so I was asking how I should be handling them.  ;)

bgruening:
Thanks  @gregvonkuster!

PR #1076, opened 2016-12-13
Jbrowse: finish python3 conversion

abretaud:
Some code that needed to be ported to python3 (just converted as explained in http://stackoverflow.com/questions/4296249/how-do-i-convert-a-hex-triplet-to-an-rgb-tuple-and-back)

### DIFF
--- a/tools/jbrowse/jbrowse.py
+++ b/tools/jbrowse/jbrowse.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 import argparse
+import codecs
 import copy
 import hashlib
 import json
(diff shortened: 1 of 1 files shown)


bgruening:
One test is failing.
@abretaud maybe you should pin a specific python version to be safe?

tools/jbrowse/jbrowse.py
```diff
@@ -127,7 +127,7 @@ def __init__(self):
 
     def rgb_from_hex(self, hexstr):
         # http://stackoverflow.com/questions/4296249/how-do-i-convert-a-hex-triplet-to-an-rgb-tuple-and-back
-        return struct.unpack('BBB', hexstr.decode('hex'))
+        return struct.unpack('BBB', bytes.fromhex(hexstr))
```
nsoranzo: I think that
```python
        return struct.unpack('BBB', codecs.decode(rgbstr, 'hex'))
```
should work for both Python 2 and 3.

Obviously you need to `import codecs` at the top.

abretaud:
Thanks @nsoranzo! it should be ok now

Issue #350, opened 2015-10-25
SpeedSeq pipeline for Galaxy


bgruening:
Would be nice to have the [SpeedSeq](https://github.com/hall-lab/speedseq) pipeline in Galaxy as a workflow.

Many of the used tools are already provided by the Galaxy community.
A few others are missing and needs to be integrated. Anyone up for this task? 
The IUC will support you wherever we can!

 * [ ] https://github.com/GregoryFaust/samblaster
 * [ ] https://github.com/arq5x/lumpy-sv
 * [ ] https://github.com/cc2qe/vawk
 * [ ] [Gemini](https://github.com/galaxyproject/tools-iuc/tree/master/tools/gemini) update (Bjoern/Ann)
 * [ ] building the workflow

bgruening:
Freiburg will update the Gemini wrappers to a recent version.

bgruening:
And here is my first version of the new multi-core freebayes wrapper: https://github.com/galaxyproject/tools-devteam/pull/265

souravsingh:
I would like to pick up the task

bgruening:
Awesome! Let me know what you want to do. 
Feebayes needs some testing: https://github.com/galaxyproject/tools-devteam/pull/265
Especially in the merging part of the vcf files!

PR #1082, opened 2016-12-18
Samblaster

mvdbeek:
This is a nice tool that can mark duplicate reads, output discordant read pairs, and most importantly, clipped reads according to various criteria. This is a prerequisite for [TEPID](https://github.com/ListerLab/TEPID), for which I am currently writing a wrapper and which I would also like to include in the iuc when it's done.


bgruening:
@mvdbeek see here: https://github.com/bgruening/galaxytools/blob/master/tools/samblaster/samblaster.xml

I guess I'm going to deprecate ours then.

mvdbeek:
:( ... too bad this wasn't on any of the toolsheds, I think that would have saved us some work.

bgruening:
Jupp, it was part of this effort: https://github.com/galaxyproject/tools-iuc/issues/350

tools/samblaster/samblaster.xml
```diff
@@ -0,0 +1,210 @@
+<tool id="samblaster" name="samblaster" version="0.1.24.2">
```
bgruening: why the leading .2?

tools/samblaster/samblaster.xml
```diff
@@ -0,0 +1,210 @@
...
+    <description>marks duplicates, outputs split reads, discordant read pairs and unmapped reads</description>
+    <requirements>
+        <requirement type="package" version="0.1.24">samblaster</requirement>
+        <requirement type="package" version="0.6.5">sambamba</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        #if $input.is_of_type('sam'):
+        #set stream="<(sambamba view -S -f bam -t ${GALAXY_SLOTS:-4} -h '%s')" % $input
```
bgruening: quote of input

tools/samblaster/samblaster.xml
```diff
@@ -0,0 +1,210 @@
...
+        #else:
+            #set stream="'%s'" % $input
```
bgruening: quote

tools/samblaster/samblaster.xml
```diff
@@ -0,0 +1,210 @@
...
+        #end if
+        sambamba view -t \${GALAXY_SLOTS:-4} -h <(sambamba sort -t \${GALAXY_SLOTS:-4} -n $stream -o /dev/stdout) |
+        samblaster
+        $output
```
bgruening: quote

tools/samblaster/samblaster.xml
```diff
@@ -0,0 +1,210 @@
...
+            && sambamba sort -o splitter.bam -l 6 -t \${GALAXY_SLOTS:-4} <(sambamba view -S -f bam splitter.sam)
+        #end if
+    ]]></command>
+    <inputs>
+        <param name="input" argument="--input" type="data" format="bam,sam"/>
+        <param name="output" argument="--ouput" label="Output bam file for all input alignments" type="boolean" checked="true" truevalue="-o output.sam" falsevalue="-o /dev/null"/>
+        <param name="discordantFile" argument="--discordantFile" label="Output discordant read pairs?" type="boolean" truevalue="-d discordant.sam" falsevalue=""/>
+        <param name="splitterFile" argument="--splitterFile" label="Output split reads?" type="boolean" truevalue="-s splitter.sam" falsevalue=""/>
+        <param name="unmappedFile" argument="--unmappedFile" label="Output unmapped/clipped reads as FASTQ?" type="boolean" truevalue="-u unmapped.fastq" falsevalue=""/>
+        <param name="acceptDupMarks" argument="--acceptDupMarks" label="Accept duplicate marks already in input file instead of looking for duplicates in the input?" type="boolean" truevalue="-a" falsevalue=""/>
```
bgruening: the `name` can be dropped

tools/samblaster/samblaster.xml
mvdbeek: local testing, I was comparing samtools against sambamba for the conversions.
I'll reset this.

tools/samblaster/samblaster.xml
mvdbeek: This is already quoted on the commandline because of the string formatting into '%s'. Quoting again shouldn't be necessary, no? If I look at the generated commandline this is quoted ...

tools/samblaster/samblaster.xml
mvdbeek: This shouldn't be necessary either, since $output will invariably expand to `-o output.sam` or `''`.
If I do quote, the commandline would be `sambamba ... '-o output.sam'`, which won't work.
I can rework this to a more verbose cheetah statement though.

bgruening: Passt! Ich brauche nur schlaf.

mvdbeek: No, problem, there's no rush here!

bgruening:
Thanks @mvdbeek!

PR #1083, opened 2016-12-19
Tripal data loaders

abretaud:
Hello,
Finally, here are the galaxy wrappers for [python-tripal](https://github.com/abretaud/python-tripal).

It is intended to be used mainly in the [galaxy genome annotation project](https://github.com/galaxy-genome-annotation/), but it can be used on any tripal instance with corresponding modules installed.

The tests are quite basic, it just checks that it cannot find the server as expected. But python-tripal checks the input params before throwing the connection error, so it's better than nothing.

These tools could be a little too exotic to be in IUC, if you think it would be better in [GGA-galaxy-tools](https://github.com/galaxy-genome-annotation/galaxy-tools).

(next should be [python-chado](https://github.com/abretaud/python-chado) I think)

### DIFF
--- a/tools/tripal/.shed.yml
+++ b/tools/tripal/.shed.yml
@@ -1,0 +1,19 @@
+categories:
+- Web Services
+long_description: |
+  Galaxy tools allowing to load data into a distant Tripal server.
+
+  Tripal is a toolkit for construction of online biological (genetics, genomics, breeding, etc), community database,
+  and is a member of the GMOD family of tools. Tripal provides by default integration with the GMOD Chado database schema and Drupal, a popular Content Management Systems (CMS).
+
+  https://github.com/abretaud/python-tripal
+homepage_url: https://github.com/abretaud/python-tripal
+owner: iuc
+remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/master/tools/tripal
+type: unrestricted
+auto_tool_repositories:
+  name_template: "{{ tool_id }}"
+  description_template: "{{ tool_name }} (from the Tripal tool suite)"
+suite:
+  name: "suite_tripal"
+  description: Galaxy tools allowing to load data into a distant Tripal server.
--- a/tools/tripal/.shed.yml
+++ b/tools/tripal/.shed.yml
@@ -1,7 +1,7 @@
 categories:
 - Web Services
 long_description: |
-  Galaxy tools allowing to load data into a distant Tripal server.
+  Galaxy tools allowing to load data into a remote Tripal server.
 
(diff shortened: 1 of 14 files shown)


bgruening:
@abretaud any pressure with this one? 
This looks really great and I think IUC is happy to maintain this with you :)

tools/tripal/.shed.yml
```diff
@@ -0,0 +1,19 @@
+categories:
+- Web Services
+long_description: |
+  Galaxy tools allowing to load data into a distant Tripal server.
```
erasche: `s/distant/remote`

tools/tripal/macros.xml
```diff
@@ -0,0 +1,161 @@
...
+            <param name="url"
+                   type="text"
+                   label="Server URL"
+                   help="The root URL of the Tripal instance" />
+            <param name="user"
+                   type="text"
+                   label="Username" />
+            <param name="password"
+                   type="text"
+                   label="Password" />
```
erasche: This is the reason I haven't PR'd any apollo tools, unfortunately. If there's no pressure, we might wait until @bgruening / @katbeaulieu have progress on the user preferences stuff.

abretaud:
> Some of these look like "admin" tools. I'm not really sure how tripal is used in the wild, and what the spectrum is like from "I want to analyse data from tripal in galaxy" to "I'm contributing data back to my local tripal site" to "I administer the tripal instance and am allowed to possibly issue disk/db expensive sync operations."

Well, these tools are for "I administer the tripal instance and am allowed to possibly issue disk/db expensive sync operations." :)
 
> Surprised to not see any "fetch data from tripal" tools.

For this I would personally prefer to use some REST stuff or python-chado rather than using tripal (though this could change with the api in tripal 3)

> Hmm, I may have different views of data management with these remote resources. It is causing me to view new tools that are useful to others too narrowly. I view from perspective of "what tools should my end users be able to access", not considering other people's end users may have more permissions than I give mine.

In fact on my side I have 2 kind of users:
- end users who will jut go to the read-only tripal instance and find the data we loaded for them
- admin users (4~5 here currently, including me) who want to load data in tripal but also in jbrowse/apollo/etc They have access to a small dedicated galaxy server with all these tools to ease their life. So this PR is for them

tools/tripal/macros.xml
abretaud: hehe no worry, comments are always welcome!

abretaud:
Let's close this for now, I pushed tripal tools to https://github.com/galaxy-genome-annotation/galaxy-tools, we'll see if it can come to IUC one day

PR #1086, opened 2016-12-20
Remove deprecated `metadata_name` attribute from bwameth and add cate…

mvdbeek:
…gory for mothur.

I'll admit that I don't know what the `metadata_name` tag was supposed to do, so I just removed it.
Or was this just to make sure a build had been specified and should be replaced with
```
 <validator type="unspecified_build" />
```
?

### DIFF
--- a/tools/bwameth/bwameth.xml
+++ b/tools/bwameth/bwameth.xml
@@ -41,7 +41,7 @@
                 <option value="indexed">Use a built-in index</option>
             </param>
             <when value="history">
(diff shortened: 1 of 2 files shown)


mvdbeek:
Urgh, it seems the URL linting is failing, though this works if I run it locally.
So ... merge anyway and upload manually?

yhoogstrate:
The webserver of mothur probably blocks it because it gets overloaded. The PR seems perfect though. Thanks @mvdbeek, merging is Fine with me. If nobody objects within a few days just merge it. I think with the latest version of planemo it is now impossible to get a positive lint on mothur...

Issue #1081, opened 2016-12-16
sra-tools fails when forward has no output from fastq-dump


Takadonet:
So tool will finish to completion but the collection will be in finished or running state with no data. Issue is that discover output will fail because we see a SRA_2.fastq file but no corresponding _1 fastq file. 

Reason being is that the SRA_1.fastq does not exist because fastq-dump did not create one becuase the forward read was empty.

Example of SRA would be https://trace.ncbi.nlm.nih.gov/Traces/sra/?run=SRR1993644

No idea how to solve this.

yhoogstrate:
Hi @Takeadonet,

I don't really see this as a issue but more as a (valid) error message.

I am aware of FASTQ tools that depend on the order in the _R1 and _R2 of the files to link mate pairs.
Given that this is not the case in your example, i.e. the number of lines does not correspond, we're dealing with FASTQ files that are likely to cause unexpected behaviour and I would prefer to see an error.

I think this problem has to be solved upstream and such dataset should be set to single-end.

I am happy to hear your thoughts on this.


Youri

Takadonet:
I agree that it has to be solved upstream and I am thinking more on the lines that SRA should be reject outright or have a 'discarded' collection. Perhaps allow the user to select which one to use.

blankenberg:
I'd lean towards the side that there is a bug with the way that these tools handle the collection discovery when some pairs might be empty. Upstream fixes are always great, but we definitely don't want to have the case where we end up with a 'forever running' collection. 

Also, there an issue where, when we have empty collections that had job errors there is no way for the user to see the error.

 xref https://github.com/galaxyproject/tools-iuc/issues/970

Takadonet:
Fixed in #1212

PR #1087, opened 2016-12-20
sra-tools: solves hanging behaviour for missing *_1 files for PE data

yhoogstrate:
I think this solves #1081. It does not give an error if something like this happens. Do we want this or would it be desired to return an error in the end as well?

### DIFF
--- a/tools/sra-tools/fastq_dump.xml
+++ b/tools/sra-tools/fastq_dump.xml
@@ -94,19 +94,41 @@
 
 
     #if str( $outputformat ) == "fasta":
-
-        for f in *_2.fasta ; do   mv "\$f" "`basename \$f _2.fasta`_reverse.fasta" ;  mv "`basename \$f _2.fasta`_1.fasta" "`basename \$f _2.fasta`_forward.fasta"  ; done &&
+        ls -alsh;
+        
+        for f in *_2.fasta ; do
+            basename=`basename \$f _2.fasta` ;
+            mv "\$f" "\$basename"_reverse.fasta ;
+            
+            ## in exceptional cases this file may be absent
+            if [ "\$basename"_1.fasta does not exist ];
+            then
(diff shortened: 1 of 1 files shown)


bgruening:
ping @galaxyproject/iuc

mvdbeek:
I would prefer to error out, or to have an option like `fail when forward read of paired collection not present`. For larger scale analysis I think it's easier to react to failed datasets than to datasets that seem OK. But anything is better than it is now, so 👍

tools/sra-tools/fastq_dump.xml
```diff
@@ -94,21 +94,37 @@
 
 
     #if str( $outputformat ) == "fasta":
-
-        for f in *_2.fasta ; do   mv "\$f" "`basename \$f _2.fasta`_reverse.fasta" ;  mv "`basename \$f _2.fasta`_1.fasta" "`basename \$f _2.fasta`_forward.fasta"  ; done &&
-        for f in *_1.fasta; do mv "\$f" "`basename \$f _1.fasta`__single.fasta"; done
-
+        for f in *_2.fasta ; do
+            basename=`basename \$f _2.fasta` &&
```
nsoranzo: Should `\$f` be double-quoted?

tools/sra-tools/fastq_dump.xml
```diff
@@ -94,21 +94,37 @@
...
+            mv "\$f" "\$basename"_reverse.fasta &&
+            
+            ## in exceptional cases this file may be absent
+            if [ ! -e "\$basename"_1.fasta ];
+            then
+               touch "\$basename"_1.fasta &&
+            fi ;
+            
+            mv "\$basename"_1.fasta "\$basename"_forward.fasta &&
+          done &&
```
nsoranzo: Indentation

PR #1088, opened 2016-12-21
Add wrapper for SPAdes.

davebx:
Based on @Takadonet's SPAdes wrapper, with some changes:

- Modified the XML to conform to the syntax I prefer.
- Removed tool_dependencies.xml, since 3.9.0 is available in bioconda
- Removed wrapper script. If the stats are important, I can add a sed/awk postprocessing command to generate them.

I wasn't able to make the tests pass with smaller datasets, if anyone has <= 100kB fastq files that run correctly in spades, I'm happy to use them.


tools/spades/.shed.yml
```diff
@@ -0,0 +1,9 @@
+categories:
+- Assembly
+description: St. Petersburg genome assembler.
+long_description: |
+   SPAdes is intended for both standard isolates and single-cell MDA bacteria assemblies.
+name: spades
+owner: iuc
```
nsoranzo: Should we keep using `nml`? @Takadonet What do you prefer?

nsoranzo:
Fantastic, thanks @davebx and @Takadonet!

tools/spades/.shed.yml
Takadonet: I rather keep using NML user just because I do not have to change half dozen workflows we already have in 'production' on our galaxy instance.

nsoranzo: Cool, can you grant authority to make changes to `iuc` in the main repository page and add "Intergalactic Utilities Commission" to the admin groups in "Manager repository administrators"?

Takadonet: I added write permission to all the repos. What do you mean by admin groups?

davebx: I'll make the owner nml here, then.

bgruening:
Nice! Thanks this is a great addition!

nsoranzo:
@Takadonet Can you please create the `spades` repository as user `nml` in the Test Tool Shed (and give iuc write and admin rights)? Our automatic planemo update fails without it.

Takadonet:
Will do tomorrow. I am away from any real workstation

Takadonet:
Done. Please confirm all is well.

nsoranzo:
Looks good, thanks a bunch @Takadonet!

blankenberg:
@Takadonet could you do the same for `metaspades` on the test toolshed, please?

Takadonet:
Done.

blankenberg:
Awesome, thanks :)

PR #1093, opened 2017-01-02
First version of fermi2 / fermikit variant caller

mvdbeek:
Fermikit can be used to call variants from high coverage short read sequencing
data.


bgruening:
`tools/fermikit/.fermi2.xml.swo` should be removed, isn't it?

tools/fermikit/fermi2.xml
```diff
@@ -0,0 +1,65 @@
+<tool id="fermi2" name="fermi2" version="0.14.dev1">
+    <description>assembles Illumina reads into unitigs</description>
+    <requirements>
+		<requirement type="package" version="r193">fermi2</requirement>
```
bgruening: indentaion

tools/fermikit/fermi2.xml
```diff
@@ -0,0 +1,65 @@
+<tool id="fermi2" name="fermi2" version="0.14.dev1">
```
bgruening: is the version number correct? `dev` seems wired to me.

bgruening: Oh I see the versions are different between the tools: 
```
+        <requirement type="package" version="0.14.dev1">fermikit</requirement>
```

So fermikit is a bunch of tools and fermi2 is a workflow/makefile thingy?

mvdbeek: There is a fermi2 binary, that `focuses on the exploration of FMD-index as a graph. It relies on fermi or ropebwt2 for index construction.` fermi2.pl is a wrapper with which you can then do an assembly, and fermikit includes fermi2, bwa, trimadap etc so that you can go from raw reads to variant calls based on the assembly done by fermi2. I split the fermikit functionality into the assembly (using the `fermi2` wrapper), mapping (using the standard bwa-mem wrapper with `intractg` option) and variant calling (with the `fermikit_variants` wrapper). I suppose we could rely on multiple exact tools instead of the fermikit package, if you prefer that.

mvdbeek: and I chose version 0.14dev1 because the dependencies in the fermikit conda package match the current master (see https://github.com/lh3/fermikit/blob/master/NEWS.md)

bgruening: Ok, makes somehow sense :)

bgruening:
Thanks!

mvdbeek:
Thanks Bjoern!

PR #1096, opened 2017-01-03
Add metaSPAdes wrapper.

davebx:


tools/metaspades/metaspades.xml
```diff
@@ -0,0 +1,118 @@
...
+                </param>
+                <when value="separate">
+                    <param name="fwd_reads" type="data" format="fastq" label="Forward reads" help="FASTQ format" />
+                    <param name="rev_reads" type="data" format="fastq" label="Reverse reads" help="FASTQ format" />
+                </when>
+                <when value="interleaved">
+                    <param name="interleaved_reads" type="data" format="fastq" label="Interleaved paired reads" help="FASTQ format" />
+                </when>
+                <when value="paired-collection">
+                    <param name="fastq_collection" type="data_collection" label="Paired-end reads collection" optional="false" format="fastq" collection_type="paired" help="FASTQ format" />
```
bgruening: optional="false" is redundant ... isn't it?
@davebx why you don't need to have the repeat blog in your test section? Is this a Galaxy bug?

tools/metaspades/metaspades.xml
```diff
@@ -0,0 +1,118 @@
...
+    </tests>
+    <help>
+<![CDATA[
+**What it does**
+
+SPAdes – St. Petersburg genome assembler – is intended for both standard isolates and single-cell MDA bacteria assemblies. See http://bioinf.spbau.ru/en/spades for more details on SPAdes.
+
+This wrapper runs SPAdes 3.9, collects the output, and throws away all the temporary files. It also produces a tab file with contig names, length and coverage.
+
+**License**
```
bgruening: Can we move all of this into a README file. It feels wired to have so many things in the tool help.

tools/metaspades/metaspades.xml
davebx: It's not required if you're only testing with one instance in the repeat.

bgruening: Interesting ... didn't know that! Thanks!

tools/metaspades/metaspades.xml
davebx: That specific line doesn't really need to be there, I think. And the rest can be moved to a README.md file, you're right.

bgruening:
@davebx thanks a lot and happy new year!

Issue #1072, opened 2016-12-11
deprecate redundant htseq wrapper?


bgruening:
Should this tool be deprecated in favor of https://github.com/galaxyproject/tools-iuc/blob/master/tools/htseq_count/htseq-count.xml

https://github.com/galaxyproject/tools-iuc/tree/master/tools/htseq

ping @yhoogstrate

yhoogstrate:
Yes please, these packages areredundant. Good catch, thanks @bgruening. The `htseq-count` fits the IUC standards a bit bitter than `htseq`, so I would go for `htseq-count` too.

bgruening:
@yhoogstrate do you have time to do this?

yhoogstrate:
I will take care of this. Do I just need to move it to `/deprecated` and disable it via the webinterface or does this need to be done via the `.shed.yml` these days?

bgruening:
Via the webinterface is enough I think.

bgruening:
Thanks @yhoogstrate! Sorry in a meeting right now :(

yhoogstrate:
@bgruening I don't think IUC is able to deprecate it, only fubar seems to be able to do this.

PR #1102, opened 2017-01-05
cleanups of htseq-count

bgruening:


bgruening:
This explains why this tool was never really tested and fixes https://github.com/galaxyproject/tools-iuc/issues/1072
ping @lparsons

lparsons:
Thanks for the detective work @bgruening, but something's fishy here. The conda env can't be built due to conflicting samtools versions, yet these tests seem to pass...

lparsons:
I'm not sure, but the version of htseq-count should probably be bumped to `0.6.1galaxy3`.

nsoranzo:
@bgruening Why move only the `.shed.yml` to `deprecated/`? I'd say move everything there or just remove everything.

nsoranzo:
Superseded by #1126.

PR #1107, opened 2017-01-10
Fixed an error when moving html report (Validate mapping file)

ryotas:
Can we get the name of input file? Then, p26 can be simply:
mv '$html_report.files_path/<file name of $mapping_fp>.html' '$html_report'


ryotas:
Oh, you have fixed already. I will try it.

PR #1109, opened 2017-01-11
migrate in emboss_5, fastqc, freebayes, picard from devteam

martenson:
part of https://github.com/galaxyproject/tools-devteam/issues/406
xref: https://github.com/galaxyproject/tools-iuc/pull/1060

ping @galaxyproject/iuc 

this is an effort to address part of https://github.com/galaxyproject/tools-devteam/issues/416


martenson:
thanks to @erasche for helping with figuring out how to treat the missing docs

martenson:
freebayes fails linting because of `<conversion>` not being allowed in schema https://docs.galaxyproject.org/en/latest/dev/schema.html

```
<param name="input_variant_vcf" type="data" format="vcf_bgzip" label="Use variants reported in VCF file as input to the algorithm">
    <conversion name="Tabixized_input" type="tabix" />
</param>
```

nsoranzo:
@martenson I've pushed 2 commits to your branch and I plan to do some more work tomorrow, hope that's OK.

yhoogstrate:
Same for me (Just one pr)

tools/picard/read_group_macros.xml
```diff
@@ -0,0 +1 @@
+../../macros/read_group_macros.xml
```
nsoranzo: I've added the file in commit 10954c0 .

nsoranzo:
I made a massive clean-up of the EMBOSS wrappers in commit cafb44d . More can be done post-merge, e.g.:
- fix datatype assignment to output datasets
- change many `type="select"` params to `boolean`.

nsoranzo:
> freebayes fails linting because of `<conversion>` not being allowed in schema

@martenson @jmchilton I've created https://github.com/galaxyproject/galaxy/issues/3453 to track this.

nsoranzo:
@martenson I've added the changes requested in my review of https://github.com/martenson/tools-iuc/pull/1 .
@jmchilton If you can update Planemo with the new XSD changes, then the tests should pass.

tools/emboss_5/tool_dependencies.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0"?>
```
yhoogstrate: @nsoranzo Can this be removed?

tools/fastqc/rgFastQC.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="fastqc" name="FastQC" version="0.67">
+<tool id="fastqc" name="FastQC" version="0.11.5.0">
```
martenson: You effectively downgraded the version here.

tools/emboss_5/tool_dependencies.xml
nsoranzo: It is not in Bioconda, but it is in `iuc`:
```
$ conda search emboss
Fetching package metadata ...............
emboss                       5.0.0                         0  iuc             
                             6.5.7                         0  bioconda        
                             6.5.7                         1  bioconda        
                             6.5.7                         2  bioconda        
```
Anyway, I have no strong opinion, maybe we can remove it when we upgrade to a newer EMBOSS version.

martenson:
All tests are failing now, I don't know why.

nsoranzo:
The problem was that release_16.10 has not been update to require six 1.10, but now that we have a `twill` wheel and a new Planemo release (thanks to @jmchilton), the linting passes and only the Java tools fail for same reason. @bgruening ideas?

tools/fastqc/rgFastQC.xml
```diff
@@ -0,0 +1,140 @@
...
+        <requirement type="package" version="0.11.5">fastqc</requirement>
+    </requirements>
+    <stdio>
+        <exit_code range="1:" />
+        <exit_code range=":-1" />
+        <regex match="Error:" />
+        <regex match="Exception:" />
+    </stdio>
+    <command><![CDATA[
+        python '${__tool_directory__}/rgFastQC.py'
```
bgruening: Nicola this looks wired, isn't it?

nsoranzo: Why? rgFastQC.py is part of the wrapper.

bgruening: Isn't it a string substitution: https://github.com/galaxyproject/galaxy/blob/29cdd6007fa333e8e9caa85a1fc58ae389405cf2/lib/galaxy/jobs/__init__.py#L983
and should be `$__tool_directory__`

nsoranzo: The curly braces should be fine, see https://docs.python.org/2/library/string.html#template-strings

nsoranzo: I think I've fixed the FastQC issue with commit d38e1e7, let's see.

bgruening: Oh ok! Good to know. I thought thats the reason for the strange error message. I still think we should change the variable name and keep is consistent with other places.
Thanks NIcola!

martenson:
fastqc now passes, yay! thanks @nsoranzo

nsoranzo:
So FastQC is finally fixed, but all Picard tools still fail because of this error: Exception in thread "main" java.lang.UnsupportedClassVersionError: picard/cmdline/PicardCommandLine : Unsupported major.minor version 52.0

This indicates that Picard tools are run with Java <8 even if the correct java-jdk Bioconda package is installed. Maybe we need to use Galaxy 17.01? @bgruening

nsoranzo:
The JAVA_HOME change was not applied during the Travis build, trying to close/reopen.

bgruening:
Thanks @nsoranzo. Thats seems to fix it.
Ready for the merge?

nsoranzo:
Green at last! Thanks all!

martenson:
Splendid. Thanks @nsoranzo @bgruening and @yhoogstrate for help!

PR #1111, opened 2017-01-12
add busco tool

abretaud:
Hi,
Here's a wrapper for busco, updated/adapted from the older wrapper I wrote there: https://github.com/genouest/galaxy-tools/tree/master/tools/busco
I guess I should write a data manager too, but time lacks for now...


abretaud:
I think they are all needed yes, the whole arthropoda dir is the reference base used by busco. I already removed all the data I could from this dir, but I think if I remove more busco will just fail running

bgruening:
Alright! Thanks this is a great quality!

PR #1116, opened 2017-01-12
Conda-fy pe_histogram

gregvonkuster:


bgruening:
Do you think we need to bump the version number?

gregvonkuster:
Hmmm...I don't think any of the "conda-fy" changes I've made require a version bump.  I don't think Java version upgrades will change the results of the underlying Java code.  I feel this is also true with the changes to eliminate the anaconda_2_3_0 package in some of the enhancements.

bgruening:
You are as IUC then I'm  :)
Let's wait for someone else to chime in.

gregvonkuster:
What I'm wondering now is if not updating the version will be a problem when these are uploaded to the tool sheds.  IIRC, since requirements have changed, a new install-able revision will be created, so 2 revisions of the same tool version can be installed.  I don't quite remember how all of this works in the TS, but perhaps a tool installed with the earlier revision cannot be upgraded to the new install-able revision.  But perhaps I'm wrong.  Does anyone remember how this works without having to do a test?

bgruening:
So better bump it all of these :)
Thanks @gregvonkuster for doing this getting them into conda!

gregvonkuster:
Ok, I'll bump the versions on all of these.  Thanks!

mvdbeek:
> Does anyone remember how this works without having to do a test?

Just for reference that works without problems (limitation is that you don't get an automatic toolbox reload on multihandler galaxy setups).

PR #1125, opened 2017-01-17
Collect standard error and standard output

manabuishii:
When featureCoutns output Running result to stderr.

Running result is not collect and goes to stderr.
This featureCounts behavior is depended on
 featureCounts compile options I think.

### DIFF
--- a/tools/featurecounts/featurecounts.xml
+++ b/tools/featurecounts/featurecounts.xml
@@ -57,6 +57,7 @@
                 $pe_parameters.exclude_chimerics
 
         "${alignment}"
+        2&gt;&amp;1 | grep .
 
         ## Removal of comment and column-header line
         && grep -v "^#" "output" | tail -n+2 > body.txt


manabuishii:
Sorry , this PR collect real error message as stdout and always success.

yhoogstrate:
Hi @manabuishii,

Do you have a use-case where it is desired to have all stderr explicitly written to stdout?
If I am not mistaken you can read both in your history items.

A small technical remark. When we use `<![CDATA[ ... ]]>` we have to unescape XML and we can use `>&1` instead.

thanks,

Youri

manabuishii:
@yhoogstrate Thanks

This patch is not good.
I don't want to every stderr goes to stdout.

I'll try to think about my problem and featureCounts tool.

yhoogstrate:
@manabuishii are you in Strasbourg? Please let me know we can discuss this?

manabuishii:
@yhoogstrate Sorry I'm in Japan .

subread version 1.4.6p5

My investigation.
My colleague inform me.
"Result looks well, but it appear in stderr"

So I grep "featureCounts setting" in source code.

```
$ grep -r "featureCounts setting"      
src/readSummary.c:	print_in_box(80,1,1,"featureCounts setting");
```

I look into `print_in_box` in `core.c`

Inside the function calls `sublog_printf` in `core.c`

In `sublog_printf`
There are following code and similar code.

```c
		#if defined(MAKE_STANDALONE) || defined(RUNNING_ENV)
		fputs(vsbuf,stderr);
		fputs("\n", stderr);
		#else
		Rprintf("%s\n",vsbuf);
		#endif
```

```RUNNING_ENV``` is not declare at least `Makefile.linux`

`MAKE_STANDALONE` is appeared in `Makefile.Linux`

and it is in `LDFLAGS`

and `featureCounts` is defined following.

```
featureCounts: readSummary.c subread.h  ${ALL_OBJECTS}
	${CC} -o featureCounts readSummary.c ${ALL_OBJECTS} ${LDFLAGS}
```

`MAKE_STANDALONE` is defined by default.

and it used in `readSummary.c` for fuction `int main` .

3437-3442
```
#ifdef MAKE_STANDALONE
int main(int argc, char ** argv)
#else
int feature_count_main(int argc, char ** argv)
#endif
{
```

I think `MAKE_STANDALONE` has (at least) two purpose.
(Sorry I don't know much about history of subread and featureCounts)

One is to make standalone executable(s)
Another one is to control output destination of `print_in_box`

I don't know about higher version of subread 1.5 or etc ...

At least 1.4.6p5 has some kind of patch for Galaxy tool I think.

yhoogstrate:
@manabuishii I suppose this is about the runtime statistics and not the final output which are written to disk. The travis tests have passed succesfully so I think the results of this wrapper are appropriate, but please comment if you believe otherwise. Writing output to stderr instead of stdout is indeed not in concordance with conventions but the author seems to have chosen for this.

manabuishii:
@yhoogstrate Yes, final output are written to disk. It's OKAY and No problem.

My colleagues wants to fix if it is possible.
So I want to know wrapper's test code for travis .
Could you tell me URL of wrapper's test code for travis ?

Currently Runtime statistics goes to stderr , history turns red .
Hopefully I want to fix if featureCounts works well.

yhoogstrate:
`Currently Runtime statistics goes to stderr , history turns red .`

Now it becomes clear to me. It's the way Galaxy handles errors and this has changed over time.

The commands are given in here:
https://github.com/galaxyproject/tools-iuc/blob/master/.travis.yml

manabuishii:
Thanks I'll check that URL!

yhoogstrate:
Travis runs a `planemo test` on the repositories in which XML files have changed.

So, for featureCounts it wil run planemo test (https://github.com/galaxyproject/tools-iuc/blob/master/.travis.yml#L44) that automatically downloads the version of subread as defined in the tool wrapper (`1.4.6.p5`), and run the tests defined in the tool wrapper.

This should come down to something like `planemo test --conda_dependency_resolution --galaxy_branch "$GALAXY_RELEASE" --galaxy_source "$GALAXY_REPO" ./tools/featurecounts/featurecounts.xml`. Make sure the version of Galaxy is up to date - or, for deugging,  equal to the system you experience the error.

The featureCounts wrapper has no `<command detect_errors="..." />`-tag set, which before galaxy 16.04 would result in capturing stderr messages (behaviour you describe) and from 16.04 and above would be equal to `exit_code`, returning red boxes on non zero exit codes, which is the expected behaviour.

I suspect you are using a Galaxy version before 16.04 or you have not the conda-build of featureCounts/subread. If you do have Galaxy 16.04 and the conda build of featureCounts/subread I think you are experiencing a real bug and I would like you to file it so we can try o fix this.

Thanks for reporting,

Youri

PR #1127, opened 2017-01-17
Dexseq

dyusuf:
@bgruening check out the version that relies on the conda pkgs. :)


bgruening:
Awesome! Thanks a bunch, I removed it from the blacklist as well to trigger testing.

tools/dexseq/dexseq_count.xml
```diff
@@ -16,12 +16,12 @@
     <command>
     <![CDATA[
         #if $mode.mode_select == "prepare":
-            python \$DEXSEQ_ROOT/python_scripts/dexseq_prepare_annotation.py
+            dexseq_prepare_annotation.py
                 -r $mode.aggregate
                 $mode.gtffile
```
yhoogstrate: does this need to be escaped?

tools/dexseq/dexseq_count.xml
```diff
@@ -16,12 +16,12 @@
     <command>
     <![CDATA[
         #if $mode.mode_select == "prepare":
-            python \$DEXSEQ_ROOT/python_scripts/dexseq_prepare_annotation.py
+            dexseq_prepare_annotation.py
```
yhoogstrate: does this need to be escaped?

tools/dexseq/dexseq_count.xml
dyusuf: this is not an environmental variable but a cheetah variable, so no need to escape. or, do I misunderstand your question?

bgruening: @dyusuf all text and data parameters needs to be escaped to prevent errors if you are running galaxy on a path with whitespaces or other jokes.

dyusuf: @bgruening I intended to make the update that allows the conda integration

I never thought to enhance the original in terms of other aspects... :)

you mean escape by adding quotes, right?

yhoogstrate: Yes, single quotes 👍

tools/dexseq/dexseq_count.xml
```diff
@@ -1,9 +1,9 @@
-<tool id="dexseq_count" name="DEXSeq-Count" version="1.0">
+<tool id="dexseq_count" name="DEXSeq-Count" version="1.20.1">
     <description>Prepare and count exon abundancies from RNA-seq data</description>
     <requirements>
-        <requirement type="package" version="1.14">dexseq</requirement>
-        <requirement type="package" version="0.6.1">htseq</requirement>
-        <requirement type="package" version="0.7.7">pysam</requirement>
+        <requirement type="package" version="1.20.1=py27r3.3.1_0">bioconductor-dexseq</requirement>
```
bgruening: @dyusuf is this strict pinning really needed or can we remove everything behind the =?

dyusuf:
@bgruening I would prefer to keep them because 'version+build' provides the necessary specificity that may avoid the potential conflicts between the dependencies.  dexseq and htseq are both py2k based pkgs. Well, pysam was built with both py2k and py3k so it is not sure which build (py2k vs py3k) conda would choose to install via galaxy.

Similarly, for dexseq.xml, there is one dependency with both of R 3.3.1 and R 3.3.2. if the builds are not clear,  it may cause conflict with other dependencies where only R 3.3.1 is available.

tools/dexseq/dexseq_count.xml
dyusuf: @bgruening I would prefer to keep them because 'version+build' provides the necessary specificity that may avoid the potential conflicts between the dependencies. dexseq and htseq are both py2k based pkgs. Well, pysam was built with both py2k and py3k so it is not sure which build (py2k vs py3k) conda would choose to install via galaxy.

Similarly, for dexseq.xml, there is one dependency with both of R 3.3.1 and R 3.3.2. if the builds are not clear, it may cause conflict with other dependencies where only R 3.3.1 is available.

bgruening: @galaxyproject/iuc what do you think?

nsoranzo: Seems a bit too Conda-specific for a problem which is still only "potential".

jmchilton: Yeah I agree with @nsoranzo - I'd try to avoid this. 17.01 will ship with https://github.com/galaxyproject/galaxy/pull/3391 which should resolve all dependencies together for every Conda tool and therefore allow Conda to figure out these issues (unless I misunderstand Conda's capabilities).

dyusuf: not sure why stricter specificity is not preferred here. to resolve dependency during package management, it might be good to be more specific than less. 

the old resolver ensures the specificity even at the level of source code. 

by conda, simply with a version number, you are not even sure, over the time, whether the binary during the installation is always the one during the testing. it depends on how you order the channels or possibly even on whether you use conda 2 or 3. adding the build spec does not fully solve the problem while limiting the room of uncertainty.

moreover, there is no stability guarantee of conda itself and of the channels, at least until recently. i witnessed this type of issues when building the pkgs in bioconda.

well, I may not fully comprehend the context here.

any way, I can remove the build spec and let's see whether it is 'potential' or not. if it passes the test, I am fine with removing the build specs.

bgruening:
Thanks @dyusuf and @yhoogstrate!

yhoogstrate:
Nice @dyusuf & @bgruening ! I am curcious how you solved the unicode problem. Was it this change: `export GALAXY_RELEASE=release_17.01`?

PR #1134, opened 2017-01-19
MAnorm

zfrenchee:
Add MAnormFast from the Shao Lab.
Project page: http://139.224.227.24:2013/
Github repo: https://github.com/semal/MAnormFast
My fork: https://github.com/zfrenchee/MAnormFast
On Pypi: https://pypi.python.org/pypi/MAnormFast/0.1.0
Anaconda package: https://anaconda.org/zfrenchee/manormfast


bgruening:
Alex do you want a review soon, or should we wait until to are ready?

zfrenchee:
@bgruening Let me test it on my local galaxy a little bit first. I'll let you know once it's ready. I would definitely appreciate a quick review now though, especially about whether I've done the dependencies right using conda. 

I also forget how to do optional numeric parameters, e.g.:
```
<param name="n"  type="integer" optional="true" default="5" label="..." />
```
Isn't quite right I don't think.

nsoranzo:
> I also forget how to do optional numeric parameters, e.g.:

@zfrenchee The default value is specified with value="5", the rest looks good.

tools/manorm/MAnorm.xml
```diff
@@ -0,0 +1,32 @@
+<tool id="manorm" name="MAnorm" version="0.1.2">
+  <description>Quantitative comparison of ChIP-Seq data sets describing transcription factor binding sites and epigenetic modifications. </description>
+  <parallelism method="basic"></parallelism>
```
bgruening: I would remove this, its not working reliable :(

tools/manorm/MAnorm.xml
```diff
@@ -0,0 +1,32 @@
+<tool id="manorm" name="MAnorm" version="0.1.2">
+  <description>Quantitative comparison of ChIP-Seq data sets describing transcription factor binding sites and epigenetic modifications. </description>
```
bgruening: This should go to help and the desc should be way smaller!

tools/manorm/MAnorm.xml
```diff
@@ -0,0 +1,32 @@
...
+  <command interpreter="python">
+    MAnormFast --p1 $p1 --p2 $p2 --r1 $r1 --r2 $r2 -o $out --s1 $s1 --s2 $s2 -n $n -e $e -d $d -p $p -m $m -u $u $s $v
+  </command>
+  <inputs>
+    <param name="p1" type="data" optional="false" label="peak file of sample 1" />
+    <param name="p2" type="data" optional="false" label="peak file of sample 2" />
+    <param name="r1" type="data" optional="false" label="reads file of sample 1" />
+    <param name="r2" type="data" optional="false" label="reads file of sample 2" />
+    <!-- optional params -->
+    <param name="s1" type="integer" optional="true" default="100" label="read shift size of sample 1, which should be set as the average size of DNA fragments." />
```
bgruening: There is no default, it should be value ... also please use `planemo lint` to check for XML errors.

tools/manorm/MAnorm.xml
```diff
@@ -0,0 +1,32 @@
...
+    <!-- output params -->
+    <param name="s" type="boolean" truevalue="-s" falsevalue="" checked="false" label="By default, MAnorm will first separate both sets of input peaks into common and unique peaks, by checking whether they have overlap with any peak in the other sample, and then merge the 2 sets of common peaks into 1 group of non-overlapping ones. But if this option is used, MAnorm would not merge the common peaks and the peaks in output files will be exactly the same as those from input." />
+    <param name="v" type="boolean" truevalue="-v" falsevalue="" checked="false" label="if this option is used, MAnorm will choose biased peaks only from unique peaks, and choose unbiased peaks only from common peaks. But if this option is not used, MAnorm will choose biased and unbiased peaks just based on the M-value and P-value cutoffs, without checking whether they are common or unique peaks." />
+  </inputs>
+  <outputs>
```
bgruening: tests are missing

zfrenchee:
@nsoranzo @bgruening My question about optional parameters is subtle I think. Here's the command: 
```
MAnormFast --p1 $p1 ...  --s1 $s1 ...
```
In the command, s1 is the first of multiple optional numeric parameters. However, I don't want the command to be run with "... --s1 ..." if no value is entered. The default value I supply in the xml is what the python uses if no other value is provided. It would be silly to use all the optional flags and pass in what the python already uses as a default, right? 

I may need to change the command section of the file, because I want either neither or both `--s1` and some value for `$s1` (never just `--s1` and then blank)

zfrenchee:
2 more questions: 
1. MAnorm expects an "output directory" after the `--out` flag. What should I pass? 
2. the default value for one of the optional fields is `0.01`. Is it still `type="integer"` or is there a `type="double"`?

bgruening:
@zfrenchee please follow this tutorial, it will make you a Galaxy Tool Dev in 3h :)
https://planemo.readthedocs.io/en/latest/writing.html

zfrenchee:
@bgruening That solves my second question but not my first question, I think. What do you do when the script expects an output path to a directory, not a filename?

mvdbeek:
> What do you do when the script expects an output path to a directory, not a filename?

You can give it any random folder, and then you use `from_work_dir` in the output parameter to pick the file(s) up.

zfrenchee:
@mvdbeek Hey Marius! Good to hear from you again! 

I'm sorry, but I'm not sure I understand. Would you mind linking some documentation or explaining just a little further. I'm particularly nervous about what kind of "random" folder is allowed. I'd rather not do something random (which seems messy)...

mvdbeek:
OK, random was a bit exaggerated. When you run a job, galaxy provides you a cleaning working directory. This is where all your temporary files that your job produces are located. So if you know that your tool writes a file `result.tab` to the folder called `'output_dir'` you can add `from_work_dir="output_dir/result.tab"`. In your case you would change the output to something like this
```
<data name="out"  from_work_dir="output_dir/result.tab"/>
```
[Here's](https://planemo.readthedocs.io/en/latest/_writing_scripts.html?highlight=from_work_dir) an example from planemos documentation.
You could also copy the output to the parameter, i.e `cp output_dir/result.tab '$out'`, but I prefer to let galaxy handle this for me.

zfrenchee:
@mvdbeek so you're suggesting something like this? https://github.com/galaxyproject/tools-iuc/pull/1134/commits/d11dba4c27632a91e8d67180f02ef0bbc260c22f 

I just need to find out exactly what the output files are now, but this is the general gist?

mvdbeek:
yep, looks good.

zfrenchee:
@mvdbeek  Do I still use `<command interpreter="python">` if I'm using conda?

nsoranzo:
No, the `interpreter` attribute is deprecated.

tools/manorm/MAnorm.xml
```diff
@@ -0,0 +1,86 @@
...
+journal="Genome Biology",
+year="2012",
+volume="13",
+number="3",
+pages="R16",
+abstract="ChIP-Seq is widely used to characterize genome-wide binding patterns of transcription factors and other chromatin-associated proteins. Although comparison of ChIP-Seq data sets is critical for understanding cell type-dependent and cell state-specific binding, and thus the study of cell-specific gene regulation, few quantitative approaches have been developed. Here, we present a simple and effective method, MAnorm, for quantitative comparison of ChIP-Seq data sets describing transcription factor binding sites and epigenetic modifications. The quantitative binding differences inferred by MAnorm showed strong correlation with both the changes in expression of target genes and the binding of cell type-specific regulators.",
+issn="1474-760X",
+doi="10.1186/gb-2012-13-3-r16",
+url="http://dx.doi.org/10.1186/gb-2012-13-3-r16"
+}</citation>
```
nsoranzo: You can replace this with the simpler:
```xml
<citation type="doi">10.1186/gb-2012-13-3-r16</citation>
```

nsoranzo:
It'd be good to have a `<tests>` section and a `.shed.yml` file so Travis can run the tests.

zfrenchee:
@nsoranzo would you send me an example of a test. Since this tool takes files and converts them into other files, it seems to me like it would be impossible to run a test without attaching some multi-GB file that MAnom takes as input. What is it that you mean as a test, I guess?

zfrenchee:
Oh, and also how do you say an input can be multiple formats. For our case, the input can be a .xls or .bed. 

```
<param name="p1" type="data" format="....?
```

peterjc:
@zfrenchee use comma separated list of galaxy's datatype names, e.g. ``type="excel,bed"`` should work here.

nsoranzo:
> Since this tool takes files and converts them into other files, it seems to me like it would be impossible to run a test without attaching some multi-GB file that MAnom takes as input.

I don't know MAnorm, does it work only on files of this size? Test data should be small (on the order of KB).

zfrenchee:
@nsoranzo I double checked and the inputs/outputs are usually in the 10s to 100s of MB. Still not worth it I don't think. Since I'm not the maintainer of the actual tool's codebase, and this tool is built from a conda package which is built from a pypi package. I think checking the functionality of the algorithm this far downstream doesn't make sense. Two more factors: the codebase I used to build the pypi package is widely used (which is a proxy for error-free, imo) and I only intend to make one version of this package -- and tests are really for making sure software updates work out okay. 

Basically, I think it would be too much hassle to have tool-functionality based tests, but I'd be interested in adding other tests which "test" just this xml file. What do you think?

nsoranzo:
> Basically, I think it would be too much hassle to have tool-functionality based tests, but I'd be interested in adding other tests which "test" just this xml file. What do you think?

I agree that it's not our goal to test the functionalities of the wrapped tool, but to test the Galaxy wrapper you still need some basic input to run the tool on and verify it produces the expected output Galaxy dataset, no?

yhoogstrate:
@nsoranzo is right, tests are part of the best practices (http://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tests) and required for IUC tools. I think for this tool it is not really complicated to shorten the data... these are just tabular files. The data files provided by the author are less then one Mb, but the BED files are missing so I can't get it up and running.

zfrenchee:
I believe this can be closed, because the original MAnorm author has uploaded a tool to the toolshed, without opening a PR against this repo. I will be testing their tool, and offering suggestions to improve it.

PR #1139, opened 2017-01-20
is_type_of for seqtk and sickle

mvdbeek:


tools/sickle/sickle.xml
```diff
@@ -14,9 +14,9 @@
 
             #if $readtype.input_single.ext in ("fastq", "fastqsanger"):
```
yhoogstrate: I suppose it is not possible to overload the `is_of_type()`-function with multiple types?

mvdbeek: actually it is https://github.com/galaxyproject/galaxy/commit/9d521f38c9033d7d83697130cf3cfcad0f23010b
Thanks for thinking about this!

bgruening:
Thanks @mvdbeek!

tools/sickle/sickle.xml
```diff
@@ -12,11 +12,11 @@
         #if str($readtype.single_or_paired) == "se":
             se -f "${readtype.input_single}" -o "$output_single"
 
-            #if $readtype.input_single.ext in ("fastq", "fastqsanger"):
+            #if $readtype.is_of_type("fastq", "fastqsanger"):
```
nsoranzo: I'm not sure these are equivalent. I think that is_of_type works as Python isinstance, so this change probably breaks the logic for fastqillumina and solexa because the first is_of_type evaluates to True for all fastq subtypes (I'm on the phone, so not tested).

mvdbeek: The code is 
```
    def is_of_type( self, *exts ):
        datatypes = [ self.datatypes_registry.get_datatype_by_extension( e ) for e in exts ]
        return self.dataset.datatype.matches_any( datatypes )
```
And the datatype registry has a dictionary with extension as key and datatype as value, so I think this should be OK

mvdbeek: damnit, and then matches_any is using the isinstance mechanism. I guess we have to revert this.

mvdbeek: Do you think we can just move the `"fastq", "fastqsanger"` check as the last item in the checks,
or should we revert to comparing the extension?

nsoranzo: I'd move it last and change it to be simply an else.

Issue #1041, opened 2016-11-17
hisat2 output fastq files of unaligned reads?


dpryan79:
This request is inspired by [this post](https://biostar.usegalaxy.org/p/20551/) on the biostars site (and the fact that it's only a matter of time before one of our post-docs asks for this). It'd be nice if the hisat2 wrapper supported the `--un-conc` and `--un` options. In experiments like ribosomal profiling, you end up with a LOT of rRNA and small non-coding RNAs that are basically just nuisance species. At the command line, we get rid of these by making a fasta file with rRNA and the other things in it, align everything against that, and then just continue with the reads that don't align. I presume that the "remove rRNA using bowtie" on RiboGalaxy is working in a similar way, but of course it'd be nice to not need a separate tool that's not in the main toolshed for this.

As always, since I'm making the request, I'm happy to implement it if the IUC supports the idea.

nsoranzo:
You can probably copy/paste from the devteam's bowtie2 Galaxy wrapper.

yhoogstrate:
@dpryan79 This looks like a great addition to me and I would be happy to review it.

dpryan79:
@yhoogstrate: I totally forgot about this :P I'll see if I can put this together tomorrow and then tag you in the PR :)

yhoogstrate:
🍻

yhoogstrate:
Thanks @dpryan79

PR #1141, opened 2017-01-20
Implement #1041 (add fastq output to hisat2)

dpryan79:
ping @yhoogstrate 

This enables the use of the `--un`/`--un-conc` and `--al`/`--al-conc` options in the hisat2 wrapper, as mentioned in #1041 . The changes are modelled on the devteam bowtie2 wrapper. I've tested single/paired/paired-collection with galaxy-docker-stable, though I haven't added any new tests for these features. Let me know if that should be done.

### DIFF
--- a/tools/hisat2/hisat2.xml
+++ b/tools/hisat2/hisat2.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<tool id="hisat2" name="HISAT" version="2.0.3.1">
+<tool id="hisat2" name="HISAT2" version="2.0.3.2">
     <description>A fast and sensitive alignment program</description>
     <macros>
         <import>hisat2_macros.xml</import>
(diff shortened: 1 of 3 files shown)


tools/hisat2/hisat2.xml
```diff
@@ -84,6 +90,22 @@
...
+            #if $output_unaligned_reads_l and $output_unaligned_reads_r:
+                #set left = str($output_unaligned_reads_l).replace(".dat", ".1.dat")
+                #set right = str($output_unaligned_reads_l).replace(".dat", ".2.dat")
+                && mv '${left}' '${output_unaligned_reads_l}'
+                && mv '${right}' '${output_unaligned_reads_r}'
+            #end if
+            #if $output_aligned_reads_l and $output_aligned_reads_r:
+                #set left = str($output_aligned_reads_l).replace(".dat", ".1.dat")
+                #set right = str($output_aligned_reads_l).replace(".dat", ".2.dat")
+                && mv '${left}' '${output_aligned_reads_l}'
```
mvdbeek: I think I would prefer `mv '${left}' left_unaligned.fastq` and then in the output parameter from_work_dir="left_unaligned.fastq". While I don't think we do anything special currently, this would allow galaxy to decide whether to use mv/cp/pipe and this is probably preferable for some object store implementations.

tools/hisat2/hisat2.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<tool id="hisat2" name="HISAT" version="2.0.3.1">
+<tool id="hisat2" name="HISAT2" version="2.0.3.2">
```
bgruening: Do we really want to change this name? HISAT1 was never really used people use the term HISAT more then HISAT2 from my perspective.

tools/hisat2/hisat2.xml
bgruening: If this is preferable for some object store implementations, this this should be fixed on the framework level $output_aligned_reads_l could just be a tmp location as any other and Galaxy will know about this and can trigger a mv/cp whatsoever ...
I fail to see a compelling reason why we should add complexity and code here, without a real technical reason.

tools/hisat2/hisat2.xml
mvdbeek: I'm indifferent to it, since the tool_id stays the same. It kind of makes sense should there ever be a hisat3 ...

bgruening: ping @jgoecks 
I recall a discussion about TopHat vs. TopHat2, I think you had some arguments for TopHat at this time.
But this should not hold up the PR, let's merge it.

mvdbeek:
thanks @dpryan79 and  @yhoogstrate, that's a useful option.

dpryan79:
Awesome!
@yhoogstrate: Thanks for tidying up the PR!

PR #1142, opened 2017-01-21
Use sanger option if not solexa or illumina datatype

mvdbeek:
Followup to https://github.com/galaxyproject/tools-iuc/pull/1139
ping @nsoranzo


yhoogstrate:
This makes sense. Thank you also for this one @mvdbeek  :+1:

PR #1144, opened 2017-01-22
QIIME: alpha rarefaction wrapper

bebatut:
Hi,

I cleaned `alpha_rarefaction` wrapper and add test data

Bérénice


tools/qiime/qiime_core/alpha_rarefaction.xml
```diff
@@ -19,58 +19,132 @@
                 --parameter_fp '$parameter_fp'
             #end if
             --num_steps '$num_steps'
-            '$parallel'
-            -O "\${GALAXY_SLOTS:-4}"
```
bgruening: was the -O parameter removed?

PR #1145, opened 2017-01-24
QIIME: fix flake8

bebatut:
Hi,

I fixed flake8 linting for 2 python scripts.

Bérénice


bgruening:
:) Sorry for being nitpicking.

bebatut:
No it is good for me :)

nsoranzo:
I think it would be better to use functions from the [shutil](https://docs.python.org/2/library/shutil.html#directory-and-files-operations) Python standard library module instead of `cp`.

bgruening:
This and subprocess when needed are on the todo list.

PR #1146, opened 2017-01-24
Update dir2html.py

bgruening:
Python3 porting.

### DIFF
--- a/tools/macs2/dir2html.py
+++ b/tools/macs2/dir2html.py
@@ -1,18 +1,19 @@
 #!/usr/bin/env python
+
 import os
 import sys
 from xml.sax.saxutils import escape
 
 
-def make_table( directory ):
+def make_table(directory):
     ret = ['<table class="fileList">\n']
-    for file in os.listdir( directory ):
-        ret.append('<tr><td class="file"><a href="%s">%s</a></td></tr>\n' % ( file, escape(file).replace( 'MACS2_', '' ) ))
+    for file in os.listdir(directory):
+        ret.append('<tr><td class="file"><a href="%s">%s</a></td></tr>\n' % (file, escape(file).replace('MACS2_', '')))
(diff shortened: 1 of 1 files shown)


tools/macs2/dir2html.py
```diff
@@ -1,18 +1,19 @@
 #!/usr/bin/env python
```
nsoranzo: If you want to maintain Python2 compatibility you can add here:
```python
from __future__ import print_function
```

tools/macs2/dir2html.py
```diff
@@ -1,18 +1,19 @@
...
 import sys
 from xml.sax.saxutils import escape
 
 
-def make_table( directory ):
+def make_table(directory):
     ret = ['<table class="fileList">\n']
-    for file in os.listdir( directory ):
-        ret.append('<tr><td class="file"><a href="%s">%s</a></td></tr>\n' % ( file, escape(file).replace( 'MACS2_', '' ) ))
+    for file in os.listdir(directory):
```
nsoranzo: `file` is a Python reserved word, can you change it to something else?

tools/macs2/dir2html.py
bgruening: Python2.7 does support the brackets, isn't it? And we do only support Python2.7. I added it nevertheless.

nsoranzo: `print(foo)` in Py2.7 should be read as `print (foo)`, i.e. `print` is not a function and the parentheses are interpreted as defining the order of operations. This breaks if e.g. a comma is added inside the parentheses.

Python3:
```
In [1]: print ('foo', 'bar')
foo bar
```
Python2:
```
In [1]: print ('foo', 'bar')
('foo', 'bar')
```

So I suggest to always add the `print_function` import.

PR #1147, opened 2017-01-24
Bugfix for Busco

abretaud:
Just a little bugfix for the --species of busco. It turns out "generic" is not a good default value (returns 0 results), it's better to specify a species only if we know exactly which one.

### DIFF
--- a/tools/busco/busco.xml
+++ b/tools/busco/busco.xml
@@ -15,7 +15,9 @@
         --evalue ${adv.evalue}
         ${adv.long}
         --limit ${adv.limit}
-        --species ${adv.species}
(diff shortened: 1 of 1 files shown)


tools/busco/busco.xml
```diff
@@ -15,7 +15,9 @@
         --evalue ${adv.evalue}
         ${adv.long}
         --limit ${adv.limit}
-        --species ${adv.species}
+        #if $adv.species:
+            --species ${adv.species}
```
nsoranzo: Please add single quotes, it's a `text` param.

abretaud:
Thanks @nsoranzo !

PR #1153, opened 2017-01-28
bcftools_call fix novel_rate param

jj-umn:
The optional —novel-rate can take 1 to 3 floats (comma separated) as
arguments.

### DIFF
--- a/tools/bcftools/bcftools_call.xml
+++ b/tools/bcftools/bcftools_call.xml
@@ -248,6 +248,8 @@ SNP/indel variant calling from VCF/BCF. To be used in conjunction with samtools 
   - Some of the original functionality has been temporarily lost in the process of transition to htslib, but will be added back on popular demand. 
   - The original calling model can be invoked with the -c option.
 
+The novel-rate option can be set to modify the likelihood of novel mutation for constrained -C trio calling. The trio genotype calling maximizes likelihood of a particular combination of genotypes for father, mother and the child P(F=i,M=j,C=k) = P(unconstrained) * Pn + P(constrained) * (1-Pn). By providing three values, the mutation rate Pn is set explicitly for SNPs, deletions and insertions, respectively. If two values are given, the first is interpreted as the mutation rate of SNPs and the second is used to calculate the mutation rate of indels according to their length as Pn=float*exp(-a-b*len), where a=22.8689, b=0.2994 for insertions and a=21.9313, b=0.2856 for deletions [pubmed:23975140]. If only one value is given, the same mutation rate Pn is used for SNPs and indels. 
+
 @REGIONS_HELP@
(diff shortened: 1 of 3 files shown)


jj-umn:
@bgruening   I removed man page from the param help and remove taxi requirements as bgzip and taxi commands are installed from htslib conda package.

bgruening:
@jj-umn thanks!

jj-umn:
@bgruening   test timed out.  It is testing all bcftools.   The only tool that failed is bcftools_plugin_counts.  That is working in my environment.  It uses sed so maybe that is an issue.   Do you have any thoughts on that?

bgruening:
Mh ...

```
    raise Exception( "Test output defines nothing to check (e.g. must have a 'file' check against, assertions to check, metadata or checksum tests, etc...)")
Exception: Test output defines nothing to check (e.g. must have a 'file' check against, assertions to check, metadata or checksum tests, etc...)
```
I'm not sure if it comes from annotate. Not sure why it is timing out.

jj-umn:
@bgruening   I switched the configuration generated python script to use the print function rather than statement.  bcftools_plugin_count now passed (it was the only tool that had a file before), but testing again ran out of time.   Is it possible to extend the time limit?

yhoogstrate:
One thing related to this could be the behavior of `planemo shed_lint --tools --recursive tools/bcftools`. This seems to run a `n * n` number of lints (every XML file is linted 32 times).

tools/bcftools/bcftools_call.xml
```diff
@@ -1,9 +1,32 @@
...
+                   help="likelihood of novel mutation for constrained trio calling, see man page for details" />
+            <param name="novel_rate_del" type="float" label="Novel Rate Deletions" value="" optional="true" 
+                   help="likelihood of novel mutation for constrained trio calling, see man page for details" />
+            <param name="novel_rate_ins" type="float" label="Novel Rate Insertions" value="" optional="true" 
+                   help="likelihood of novel mutation for constrained trio calling, see man page for details" />
+        </xml>
+        <token name="@NOVEL_RATE@">
+            #set $novel_rate = []
+            #if $section.genotypes.novel_rate_snp:
+              $novel_rate.append(str($section.genotypes.novel_rate_snp))
```
nsoranzo: It's not required because `append()` returns nothing, but an initial [`#silent`](https://pythonhosted.org/Cheetah/users_guide/output.html#executing-expressions-without-output-silent) would be more explicit, e.g.:
```
#silent $novel_rate.append(str($section.genotypes.novel_rate_snp))
```

jj-umn:
@nsoranzo Thanks for the suggestions.  I updated the code as recommended.

nsoranzo:
Thanks @jj-umn!

PR #431, opened 2015-11-30
Add Qiime wrappers

bgruening:
_edit Saskia:_
..and the GCC2016 hackathon.

Resources:
Qiime script documentation: http://qiime.org/scripts/index.html
Qiime github repo: https://github.com/biocore/qiime
Qiime test data in github repo (folder qiime_test_data)

- [x] add datatypes (biom1 is now in Galxy core)

Update wrappers:
- [x] align_seqs (missing dependencies)
- [x] alpha_diversity
- [x] alpha_rarefaction
- [x] assign_taxonomy
- [x] beta_diversity
- [x] beta_diversity_through_plots
- [x] compare_categories
- [x] core_diversity_analyses
- [x] count_seqs
- [x] filter_alignment
- [x] filter_fasta
- [x] filter_otus_from_otu_table
- [x] filter_samples_from_otu_table
- [x] filter_taxa_from_otu_table
- [x] jackknifed_beta_diversity
- [x] make_otu_heatmap
- [x] make_phylogeny
- [x] pick_open_reference_otus
- [x] pick_otus
- [x] pick_rep_set
- [x] plot_taxa_summary
- [x] split_libraries_fastq
- [x] split_libraries
- [x] summarize_taxa_through_plots
- [x] summarize_taxa
- [x] upgma_cluster
- [x] multiple_join_paired_ends
- [x] multiple_split_libraries_fastq
- [x] pick_closed_reference_otus
- [x] validate_mapping_file

Add wrappers (if useful):
- [ ] add_alpha_to_mapping_file
- [ ] add_qiime_labels
- [ ] adjust_seq_orientation
- [ ] alpha_diversity
- [ ] ampliconnoise
- [ ] beta_significance
- [ ] blast_wrapper
- [ ] categorized_dist_scatterplot
- [ ] clean_raxml_parsimony_tree
- [ ] cluster_quality
- [x] collapse_samples
- [ ] collate_alpha
- [ ] compare_alpha_diversity
- [ ] compare_distance_matrices
- [ ] compare_taxa_summaries
- [ ] compare_trajectories
- [ ] compute_core_microbiome
- [ ] compute_taxonomy_ratios
- [ ] conditional_uncovered_probability
- [ ] consensus_tree
- [ ] convert_fastaqual_fastq
- [ ] demultiplex_fasta
- [ ] denoise_wrapper
- [ ] denoiser
- [ ] denoiser_preprocess
- [ ] denoiser_worker
- [ ] detrend
- [ ] differential_abundance
[...]

PR #1154, opened 2017-01-30
QIIME: assign_taxonomy

bebatut:
Hi,

I cleaned `assign_taxonomy` wrapper and add some tests.
But I have some issues with some dependencies: I commented some parameter options to avoid using these dependencies.

Bérénice


bebatut:
I will merge this PR today if nobody complain :)
We can still review this tool in the main QIIME PR #431

PR #1155, opened 2017-01-30
Remove --foldenrichment option from macs2

dpryan79:
This option hasn't existed in a few years. Macs2 uses a q-value cutoff of 0.05 by default, so neither `--pvalue` nor `--qvalue` need to be explicitly specified.


bgruening:
This tool is still not under testing, gnarf. xref: https://github.com/galaxyproject/tools-iuc/pull/710
I'm +1 in merging but would like to have a second opinion, ping @galaxyproject/iuc

bgruening:
Thanks Youri! Can you please push this repo to the TS. As this is not tested it will not be pushed.

PR #1156, opened 2017-01-30
Add plant_tribes_assembly_post_processor tool

gregvonkuster:


gregvonkuster:
Is there some way to simulate this test environment locally so that I can see what might be causing the tests to fail?

bgruening:
Look at the root folder, simulate_travis.py is your friend. And we have some nice docs about it at the new website.

gregvonkuster:
You mean in the bioconda-recipes repo here https://github.com/bioconda/bioconda-recipes/blob/master/simulate-travis.py?  Will it work for the tools in the tools-iuc repo?

bgruening:
Oh, wrong repo! lol 

-- reset --

All what you need to test this is written here on words.
https://github.com/galaxyproject/tools-iuc/blob/master/.travis.yml#L11 

It assumes a kind of clean environment.

gregvonkuster:
Yes, I use the configuration settings in https://github.com/galaxyproject/tools-iuc/blob/master/.travis.yml when testing my new tools in my dev environment.  Tools work just fine in my local Galaxy environment, but the same tool execution using planemo test always fails.  Hmmm...

gregvonkuster:
I found that this works for me, so I'll stop trying to use the configuration in the .travis.yml file.  
```
$ planemo test --galaxy_root ~/galaxy --job_output_files ~/output --no_cleanup --conda_dependency_resolution --conda_auto_install --conda_ensure_channels iuc,bioconda,r,defaults,conda-forge --conda_prefix ~/_conda --conda_exec ~/_conda/bin/conda --conda_debug --test_data ~/test-data .
```
From what I've seen, it looks like a data problem, perhaps due to the smaller files I'm trying to use in the tests along with the compare="contains" approach.  I'll try to get a fix in, but trying to keep small files.  This is difficult because this tool produces very large outputs.  I'll open a separate PR.

Issue #1157, opened 2017-01-30
fastqc erroneously reads bz2 files


martenson:
Found by @jmchilton here: https://github.com/galaxyproject/galaxy/pull/3510

> Update: This line https://github.com/galaxyproject/tools-iuc/blob/master/tools/fastqc/rgFastQC.py#L63 seems wrong according to https://docs.python.org/2/library/bz2.html.

PR #1158, opened 2017-01-31
Fix compressed fastq support for fastqc

mvdbeek:
Fix https://github.com/galaxyproject/tools-iuc/issues/1157


martenson:
Thanka a lot @mvdbeek !

PR #1161, opened 2017-02-03
Jbrowse fixes

abretaud:
3 little fixes for jbrowse:

- parse menus in xml file only when applicable (fixes tool error with bam track for example)

- add another placeholder in the generated menuTemplate: this is needed because apollo have 4 default menu items by default instead of 3 for jbrowse. (on jbrowse there will be an empty line in the menu, but couldn't find a better way)

- I also changed the urlTemplate for bam/vcf/wig tracks: it generated urls like http://jbrowse/data/../data/ that worked well most of the time, but failed with https://github.com/galaxy-genome-annotation/galaxy-tools/blob/master/tools/jbrowse/jbrowse_to_container.xml

ping @erasche I think :)

### DIFF
--- a/tools/jbrowse/jbrowse.py
+++ b/tools/jbrowse/jbrowse.py
@@ -435,8 +435,9 @@ class JbrowseConnector(object):
         cmd = ['ln', data, dest]
         self.subprocess_check_call(cmd)
 
+        url = os.path.join('raw', trackData['label'] + '.bw')
         trackData.update({
-            "urlTemplate": os.path.join('..', dest),
+            "urlTemplate": url,
             "storeClass": "JBrowse/Store/SeqFeature/BigWig",
             "type": "JBrowse/View/Track/Wiggle/Density",
         })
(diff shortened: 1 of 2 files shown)


erasche:
Looks good to me! :+1: 

(But the travis build failed? Despite exiting 0?)

yhoogstrate:
It looks like travis did not 'see' any differences comparing to master, and did no testing after all. The following commands both give no results:
```
cd '/home/travis/build/galaxyproject/tools-iuc' && git diff --name-only '9df9b52baf62b70fbcfc3fbe965d7197d4e8738e...e0a73d5e5ba07c18585129bbf758bf3599fab523' --
install.12
0.01s$ cat changed_repositories_chunk.list
```
@abretaud could you bump the version of the XML file? I think that will trigger a test.

abretaud:
Aaaaand... green! thanks!

yhoogstrate:
This is fine with me! @erasche You may press the button :+1: . Thanks @abretaud @erasche

erasche:
Ah, so that's what happened. Thanks @yhoogstrate for the explanation!

PR #1163, opened 2017-02-05
modified to include freetype fonts in build, needed for Circos

rdmorin:
This is a small change that is necessary for the package to be compatible with Circos, which requires freetype fonts.

### DIFF
--- a/packages/package_libfreetype_2_6_5/tool_dependencies.xml
+++ b/packages/package_libfreetype_2_6_5/tool_dependencies.xml
@@ -1,0 +1,22 @@
+<?xml version="1.0"?>
+<tool_dependency>
+  <package name="libfreetype" version="2.6.5">
+    <install version="1.0">
+      <actions>
+        <action type="download_by_url"> https://sourceforge.net/projects/freetype/files/freetype2/2.6.5/freetype-2.6.5.tar.gz </action>
+        <action type="autoconf"/>
+        <action type="set_environment">
+          <environment_variable action="prepend_to" name="PATH">$INSTALL_DIR/bin</environment_variable>
+          <environment_variable action="set_to" name="LIBFREETYPE_ROOT">$INSTALL_DIR</environment_variable>
+          <environment_variable action="set_to" name="LIBFREETYPE_LIB_PATH">$INSTALL_DIR/lib</environment_variable>
+          <environment_variable action="set_to" name="LIBFREETYPE_INCLUDE_PATH">$INSTALL_DIR/include</environment_variable>
+          <environment_variable action="prepend_to" name="LD_LIBRARY_PATH">$INSTALL_DIR/lib</environment_variable>
+          <environment_variable action="prepend_to" name="PKG_CONFIG_PATH">$INSTALL_DIR/lib/pkgconfig</environment_variable>
+        </action>
+      </actions>
+    </install>
+    <readme>
+      Installs freetype fonts needed for libgd
+    </readme >
+  </package>
+</tool_dependency>
--- a/packages/package_libgd_2_1/tool_dependencies.xml
+++ b/packages/package_libgd_2_1/tool_dependencies.xml
@@ -4,14 +4,14 @@
         <repository name="package_libpng_1_6_7" owner="iuc" prior_installation_required="True" />
     </package>
     <package name="libfreetype" version="2.6.5">
-        <repository name="package_libfreetype_2_6_5" owner="iuc" prior_installation_required="True" />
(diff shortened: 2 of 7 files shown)


packages/package_libgd_2_1/tool_dependencies.xml
```diff
@@ -1,29 +1,34 @@
...
     </package>
     <package name="libgd" version="2.1.0">
         <install version="1.0">
             <actions>
-                <action type="download_by_url" sha256sum="3ceef69d5454a392e8793ae90b5f0d632dd3e20879c12856aa1d1d3d063a51c8">https://depot.galaxyproject.org/software/libgd/libgd_2.1_src_all.tar.gz</action>
-
+	        <action type="download_by_url" sha256sum="3ceef69d5454a392e8793ae90b5f0d632dd3e20879c12856aa1d1d3d063a51c8">https://depot.galaxyproject.org/software/libgd/libgd_2.1_src_all.tar.gz</action>
                 <action type="set_environment_for_install">
-                    <repository name="package_libpng_1_6_7" owner="devteam">
+                    <repository name="package_libpng_1_6_7" owner="morinlab">
```
mvdbeek: can you change this to `owner=iuc` ?

packages/package_libgd_2_1/tool_dependencies.xml
```diff
@@ -1,29 +1,34 @@
...
                         <package name="libpng" version="1.6.7" />
                     </repository>
+                    <repository name="package_libfreetype_2_6_5" owner="morinlab">
```
mvdbeek: same here

packages/package_libfreetype_2_6_5/tool_dependencies.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0"?>
```
mvdbeek: You'll have to create a .shed.yml file for this package.

rdmorin:
The only reason is that I already have a working tool for circos that uses the dependencies this way and I'd rather not break it. Currently it sits in the test tool shed using this slightly modified dependency package. I thought that if the IUC package could be modified to allow freetype fonts to be installed, at least that redundancy could be removed.

packages/package_libgd_2_1/tool_dependencies.xml
```diff
@@ -1,29 +1,34 @@
 <?xml version="1.0"?>
 <tool_dependency>
     <package name="libpng" version="1.6.7">
-        <repository name="package_libpng_1_6_7" owner="devteam" prior_installation_required="True" />
+        <repository name="package_libpng_1_6_7" owner="iuc" prior_installation_required="True" />
+    </package>
+    <package name="libfreetype" version="2.6.5">
+        <repository name="package_libfreetype_2_6_5" owner="morinlab" prior_installation_required="True" />
```
bgruening: This still seems to be wrong?

rdmorin: Should I be adhering to 
s/morinlab/iuc/g

?
I wasn't sure if the new package (libbfreetype) would immediately become owned by iuc.

bgruening: I guess you should put everything under IUC, this is the easiest way. This repo will push automatically to the Tool Shed, but we can only do so if we have permissions to the repo.

rdmorin: Which repo do I need to give you permissions to such that tools will be pushed automatically?

packages/package_libfreetype_2_6_5/tool_dependencies.xml
rdmorin: Regarding the pypy package just added, I doubt that you will want to adopt the associated sequenza package but I thought that perhaps others may want to use pypy in place of python as a compiler. We've found it runs at least twice as fast when we use it with Sequenza.

bgruening: If you like to store these packages here, feel free to do so. I don't think the IUC will spend much time in creating new packages, but we are certainly interested in maintaining old oncs or migrating them to Conda packages.

packages/package_libgd_2_1/tool_dependencies.xml
bgruening: more or less all repos that we should push to, so the package_* repos in this case. It would be easier if you can just change all owners to `iuc`. This way everything will work automatically.

bgruening:
@rdmorin sorry to be not able to merge this. But we really don't want to support the `packages` any further. Please let me know if you need any help with Conda packages and I will help you!

PR #1170, opened 2017-02-09
JBrowse: Add --type option for htmlfeature

abretaud:
As discussed in https://github.com/GMOD/Apollo/issues/1483, added the --type option when selecting HTMLFeature track type


abretaud:
Just pushed another fix for the className option: it is just ignored by flatfile-to-json.pl when passed through the --clientConfig. The guilty line of code is there: https://github.com/GMOD/jbrowse/blob/40aab138058772dd7e061bebd1aa14c48221868b/src/perl5/Bio/JBrowse/Cmd/FlatFileToJson.pm#L104

erasche:
LGTM, thanks @abretaud

PR #1175, opened 2017-02-15
HiSat2: remove tool_dependencies.xml

yhoogstrate:


nsoranzo:
Should the version number be increased? There's a change in the requirements.

yhoogstrate:
@nsoranzo is this a question or a request for change?

nsoranzo:
We never formalised which changes require a version increase, I think modifications to requirements should be one of them. So I'm requesting a change but you can convince us (IUC) that's not necessary :)

PR #1177, opened 2017-02-19
Drop TS dependencies for Trinity

mvdbeek:
Remove trinity from testing blacklist, update README and use
detect_errors="aggressive" throughout.

PR #1178, opened 2017-02-21
Trinity contig exn50 statistic

loraine-gueguen:
Add wrapper for Trinity script contig_ExN50_statistic.pl.


tools/trinity/contig_exn50_statistic.xml
```diff
@@ -0,0 +1,46 @@
...
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements">
+    </expand>
+    <expand macro="stdio"/>
+    <command><![CDATA[
+
+    contig_ExN50_statistic.pl
+        "${matrix}"
```
bebatut: Can you use simple quotes here? Thanks

bebatut: You can remove also the `{` :)

tools/trinity/contig_exn50_statistic.xml
```diff
@@ -0,0 +1,46 @@
...
+        "${transcripts}"
+        | tee "${output}"
```
nsoranzo: You can just redirect with
```
> '$output'
```

tools/trinity/contig_exn50_statistic.xml
```diff
@@ -0,0 +1,46 @@
...
+
+    ]]></command>
+    <inputs>
+        <param format="tabular" name="matrix" argument="" type="data" label="Expression matrix" help="Matrix of normalized expression values produced by 'Build expression matrix for a de novo assembly of RNA-Seq data by Trinity' tool"/>
+        <param format="fasta" name="transcripts" argument="" type="data" label="Transcripts" help="De novo assembly of RNA-Seq data"/>
```
nsoranzo: I'd remove the empty `argument` in both `<param>`

tools/trinity/contig_exn50_statistic.xml
```diff
@@ -0,0 +1,46 @@
...
+            <param name="transcripts" value="count/contig_exn50_statistic/Trinity.fasta"/>
+            <output name="output" file="count/contig_exn50_statistic/ExN50.stats"/>
+        </test>
+    </tests>
+    <help>
+<![CDATA[
+Trinity_ assembles transcript sequences from Illumina RNA-Seq data.
+This tool computes the N50 statistic limited to the top most highly expressed transcripts that represent x% of the total normalized expression data. This requires that you have first performed transcript abundance estimation with 'Align reads and estimate abundance for a de novo assembly of RNA-Seq data by Trinity' tool and that you have built the expression matrix with 'Build expression matrix for a de novo assembly of RNA-Seq data by Trinity' tool. More details on this page:
+
+.. _Trinity manual: https://github.com/trinityrnaseq/trinityrnaseq/wiki/Transcriptome-Contig-Nx-and-ExN50-stats
```
nsoranzo: This link doesn't seem to be used.

tools/trinity/contig_exn50_statistic.xml
loraine-gueguen: done

tools/trinity/contig_exn50_statistic.xml
loraine-gueguen: done

tools/trinity/contig_exn50_statistic.xml
loraine-gueguen: done

tools/trinity/contig_exn50_statistic.xml
loraine-gueguen: removed

loraine-gueguen:
Thanks for the review!
I have seen that Trinity is in blacklist for Travis test because of timeout problems. Can I remove it from the list to test ?

abretaud:
@mvdbeek is currently working at removing it from the blacklist in #1177

abretaud:
Apart from that the tool looks good to me, thanks!
Can we wait #1177 before merging or is it urgent?

loraine-gueguen:
ok, no problem, we can wait.

loraine-gueguen:
@mvdbeek Do you mind if I go on by removing trinity from blacklist ? (I intend to try to suppress timeout by adding  bioconductor-qvalue requirement directly in the trinity.xml)

mvdbeek:
Sure, no problem. I think it's gonna be quite some work figuring out a compatible set of trinity dependencies unfortunately. They are still requiring samtools 0.1.19, which makes it difficult to go with the latest dependencies of the other packages :(

loraine-gueguen:
ok, thanks. But I am not going to change trinity version (I keep 2.2.0), so no need to change the dependencies. I only need to resolve the timeout due to bioconductor-qvalue in tool analyze_diff_expr.

loraine-gueguen:
No timeout in Travis. "All 23 test(s) executed passed".
I removed deseq2 requirement in tool run_de_analysis because the conda package "deseq2" doesn't exist and there was an error in Travis. Is that ok ?
Ready to merge ?

nsoranzo:
@mvdbeek If you look at the Travis log, there's something "funny". Conda fails to create the mulled environments because it choose an old `r-base` package. On the other end, the following `conda create`, and `conda installs` succeed because the merged envs all contain R 3.3.1 .

mvdbeek:
I think the mulled env building should happen when you do `planemo conda_install` ... I can't find that in the planemo output. I thought the latest planemo was doing this. Though that has nothing to do with this PR, we should merge it.

mvdbeek:
`conda create -n test_trinity bioconductor-edger=3.14.0 bioconductor-limma=3.28.10 bioconductor-deseq2=1.10.0` will also get r-base=3.2.2-0 ... we'll have to find a compatible set of dependencies in #1177, if I find some time.

jmchilton:
> I think the mulled env building should happen when you do planemo conda_install ... I can't find that in the planemo output. 

This isn't quite happening yet... I was waiting for Galaxy's ``master`` branch to be updated to 17.01. That has happened now so I need to do some work.

mvdbeek:
Oh right, thanks @jmchilton. No rush, that wouldn't solve the mulled-env building in this PR, but I still think this may help other PRs.

mvdbeek:
thanks @loraine-gueguen !

nsoranzo:
@mvdbeek I think I've found the source of the PaddingError issue for `abundance_estimates_to_matrix` and `align_and_estimate_abundance` (but don't know how to solve it): `salmon` requires `icu 56.*` while `r-base 3.3.1` requires `icu 54.*`. Since the older `r-base 3.2.2` does not have a requirement on `icu`, it gets selected when creating the mulled env and causes the PaddingError.

Ping also @bgruening for suggestions.

PR #1192, opened 2017-03-03
Add a wrapper for HUMAnN2

bebatut:
Hi,

I would like to migrate the wrapper for HUMAnN from [here](https://github.com/asaim/galaxytools).

The tests are passing locally but HUMAnN2 execution seems killed on Travis :(
Any clue?

Bérénice


bgruening:
Mh, the error code indicates that the process was killed with SIGKIL 9 ... and the log indicates that the job was started with 32 threads? Maybe this is what causes travis to kill your process?

bebatut:
Thanks @bgruening for having fixed Travis
@galaxyproject/iuc can anyone review this, please? Thanks a lot

tools/humann2/humann2_split_table.xml
```diff
@@ -0,0 +1,68 @@
...
+            #if $taxonomy_index
+                --taxonomy_index '$taxonomy_index'
+            #end if
+            #if $taxonomy_level
+                --taxonomy_level '$taxonomy_level'
+            #end if
+    ]]></command>
+
+    <inputs>
+        <param argument="--input" type="data" format="tsv,biom" label="Gene table to split"/>
```
shiltemann: s/biom/biom1/

tools/humann2/humann2.xml
```diff
@@ -0,0 +1,382 @@
...
+            --gap-fill '$gap_fill'
+            --output-format '$output_format'
+            --output-max-decimals '$output_max_dec'
+            --output-basename 'humann2'
+            $remove_statified_output
+            $remove_column_description_output
+            --pathways '$pathways'            
+    ]]></command>
+    <inputs>
+        <param argument="--input" type="data" format="fastq,fasta,sam,bam,biom" label="Input sequence file" help=""/>
```
shiltemann: datatype name for biom files is `biom1` (https://github.com/galaxyproject/galaxy/blob/dev/config/datatypes_conf.xml.sample#L497)

tools/humann2/humann2.xml
```diff
@@ -0,0 +1,382 @@
+<tool id="humann2" name="HUMAnN2" version="@WRAPPER_VERSION@.0">
```
shiltemann: this tool has a lot of parameters, you could consider putting some in a section called advanced if they're often left to default settings anyway by users?

tools/humann2/humann2.xml
```diff
@@ -0,0 +1,382 @@
...
+        <param name='pathways' type="select" label="Database to use for pathway computations" help="(--pathways)">
+            <option value="metacyc" selected="true">MetaCyc</option>
+            <option value="unipathway">UniPathway</option>
+        </param>
+    </inputs>
+    <outputs>
+        <data format="tsv" name="gene_families_tsv" from_work_dir="output/humann2_genefamilies.tsv" label="${tool.name} on ${on_string}: Gene families and their abundance" >
+            <filter>output_format == "tsv"</filter>
+        </data>
+        <data format="biom" name="gene_families_biom" from_work_dir="output/humann2_genefamilies.biom" label="${tool.name} on ${on_string}: Gene families and their abundance" >
```
shiltemann: biom1

tools/humann2/humann2_reduce_table.xml
```diff
@@ -0,0 +1,65 @@
...
+    <command><![CDATA[
+        humann2_reduce_table
+            --input '$input'
+            -o '$output_table'
+            --function '$function'
+            --sort-by '$sort' 
+    ]]></command>
+
+    <inputs>
+        <param argument="--input" type="data" format="tsv" label="Gene/pathway table"/>
```
shiltemann: add `tabular` to format here? upon upload the test dataset was sniffed as tabular and not given as an option here (or maybe even just use tabular instead of tsv everywhere as I think tsv is subclass of tabular anyway?)

tools/humann2/humann2_join_tables.xml
```diff
@@ -0,0 +1,64 @@
...
+]]>
+    </version_command>
+
+    <command><![CDATA[
+        mkdir tmp_dir
+
+        && 
+
+        #for $table in $input_tables
+            cp '$table.input_table' 'tmp_dir' &&
```
shiltemann: if you select multipe files here the `cp` command will fail as `$table.input_table` will be a comma separated list of files, think you need another for loop here :)

shiltemann: or do you even need the repeat here? would a simple multiple file select not do?

tools/humann2/humann2_join_tables.xml
```diff
@@ -0,0 +1,64 @@
...
+    <macros>
+        <import>humann2_macros.xml</import>
+    </macros>
+
+    <expand macro="requirements"/>
+    <expand macro="stdio"/>
+
+    <version_command>
+<![CDATA[
+    humann2_join_tables --version
```
shiltemann: `--version` doesn't appear to be a valid parameter? I get as tool version now:
`usage: humann2_join_tables [-h] [-v] -i INPUT -o OUTPUT [--file_name FILE_NAME] [-s] humann2_join_tables: error: argument -i/--input is required ` ;)

shiltemann: (multiple of the tools have problems with version command btw)

tools/humann2/humann2_macros.xml
```diff
@@ -0,0 +1,22 @@
...
+        <requirements>
+            <requirement type="package" version="@WRAPPER_VERSION@">humann2</requirement>
+            <yield/>
+        </requirements>
+    </xml>
+    <xml name="stdio">
+        <stdio>
+            <exit_code range="1:" />
+        </stdio>
+    </xml>
```
bebatut: Changed!

tools/humann2/humann2_join_tables.xml
bebatut: I changed it with 
```
<xml name="version">
        <version_command>@WRAPPER_VERSION@</version_command>
</xml>
```

tools/humann2/humann2_split_table.xml
bebatut: Corrected!

tools/humann2/humann2_split_table.xml
```diff
@@ -0,0 +1,68 @@
...
+                <element name="demo_Coverage" file="demo_Coverage.tsv" />
+                <element name="demo_Abundance" file="demo_Abundance.tsv" />
+            </output_collection>
+        </test>
+    </tests>
+
+    <help><![CDATA[
+**What it does**
+
+HUMAnN is a pipeline for efficiently and accuretly profiling the presence/absence and abundance of microbial pathways in a community from metagenomic or metatranscriptomic sequencing data. `Read more about the tool <http://huttenhower.sph.harvard.edu/humann2/manual>`_.
```
bebatut: I corrected and moved the header to macros.xml

tools/humann2/humann2_renorm_table.xml
```diff
@@ -0,0 +1,70 @@
...
+            --input '$input'
+            -o '$output_table'
+            --units '$units'
+            --mode '$mode'
+            --special '$special'
+    ]]></command>
+
+    <inputs>
+        <param argument="--input" type="data" format="tsv" label="Gene/pathway table"/>
+        <param argument="--units" type="select" label="Normalization scheme" help="(--norm)">
```
bebatut: Good point. Corrected

tools/humann2/humann2_renorm_table.xml
```diff
@@ -0,0 +1,70 @@
...
+            <option value="cpm" selected="true">Copies per million</option>
+            <option value="relab">Relative abundance</option>
+        </param>
+        <param argument="--mode" type="select" label="Normalization level" help="(--norm)">
```
bebatut: Good point. Corrected

data_managers/data_manager_humann2_database_downloader/data_manager/data_manager_humann2_download.xml
```diff
@@ -0,0 +1,56 @@
...
+    <outputs>
+           <data name="out_file" format="data_manager_json" label="${tool.name}"/>
+    </outputs>
+    <tests>
+    </tests>
+    <help>
+
+This tool downloads the HUMAnN2 databases.
+
+For details about this tool, please go to http://gemini.readthedocs.org
```
bebatut: Good point. I forgot to change that... :(

tools/humann2/humann2_reduce_table.xml
bebatut: I added tabular for every input with tsv

tools/humann2/humann2_reduce_table.xml
```diff
@@ -0,0 +1,65 @@
...
+    </tests>
+
+    <help><![CDATA[
+**What it does**
+
+HUMAnN is a pipeline for efficiently and accuretly profiling the presence/absence and abundance of microbial pathways in a community from metagenomic or metatranscriptomic sequencing data. `Read more about the tool <http://huttenhower.sph.harvard.edu/humann2/manual>`_.
+
+Reduce HUMAnN2 table is a tool to reduce the table given a function (max, sum, mean or min).
+
+HUMAnN is a pipeline for efficiently and accuretly profiling the presence/absence and abundance of microbial pathways in a community from metagenomic or metatranscriptomic sequencing data. `Read more about the tool <http://huttenhower.sph.harvard.edu/humann2/manual>`_.
```
bebatut: Removed. Thanks!

tools/humann2/humann2_rename_table.xml
```diff
@@ -0,0 +1,77 @@
...
+            <option value="kegg-module">KEGG Module</option>
+            <option value="ec">Enzyme commision (EC)</option>
+            <option value="go">Gene Ontology (GO)</option>
+            <option value="pfam">PFAM</option>
+            <option value="eggnog">eggNOG</option>
+            <option value="uniref50">UniRef50</option>
+            <option value="kegg-pathway">KEGG Pathway</option>
+            <option value="kegg-orthology">KEGG Orthology</option>
+        </param>
+        <param argument='--simplify' type='boolean' checked="false" truevalue='--simplify' falsevalue='' label="Remove non-alphanumeric characters from names?"/>
```
bebatut: You are right. I changed that :)

tools/humann2/humann2.xml
bebatut: I mixed every time. Thanks

tools/humann2/humann2.xml
bebatut: Corrected

tools/humann2/humann2_join_tables.xml
bebatut: I removed the repeat. It does not make any sense here

bebatut:
I remove some test-data too big. Some of them I need to keep them: it will need more time (and a lot of more work) to make them smaller

tools/humann2/README.rst
```diff
@@ -0,0 +1,12 @@
+Galaxy wrappers for HUMAnN2
+===========================
+
+Galaxy should be able to automatically install the dependencies, i.e. the
+HUMAnN2 binaries and its dependencies.
+
+After installation, you must tell Galaxy about the defaut databases for MetaPhlAn2 and HUMAnN2.
```
bebatut: Corrected

bebatut:
Thanks @shiltemann for all these comments!

tools/humann2/humann2.xml
bebatut: I need to think a bit more about that...

bgruening: Using <section> maybe?

tools/humann2/humann2_join_tables.xml
shiltemann: it looks like you could also just do `humann2 --version`? ..that way it will always be the version reported by the tool itself rather than the galaxy wrapper?

tools/humann2/humann2.xml
bebatut: I used `<section>`

tools/humann2/humann2_join_tables.xml
bebatut: Ok I changed it :)

shiltemann:
looks good @bebatut ..you can remove the `<expand macro="stdio"/>` lines from the tools now that you've added the `detect_errors="exit_code"` right?

travis is complaining about long log lengths, hmm, it doesn't seem to get that big locally..

bebatut:
I removed the stdio...
I do not understand also why Travis is complaining. It was at least running the tests yesterday and not anymore now :(

mvdbeek:
This is the progress bar output of the conda installation that is causing trouble.
See the text output https://s3.amazonaws.com/archive.travis-ci.org/jobs/208137727/log.txt
@jmchilton I think we should use the `--quiet` option, at least in planemo.

bebatut:
@galaxyproject/iuc is it ok to merge with this patch on the travis.yml until the quiet option is added to planemo?

bgruening:
Thanks @bebatut and @shiltemann!

bgruening:
Any idea what is going on here? https://travis-ci.org/galaxyproject/tools-iuc/jobs/210306329
I have uploaded the datamanagers manually, but I'm still not able to upload the tools. Is there a typo I do not see somewhere?

nsoranzo:
No clues, @martenson can you take a look?

gregvonkuster:
I've taken a look at this as well, and I think the problem lies with combining the use of conda recipes (where there is no tool_dependencies.xml file) and defining a "complex repository dependency" which assumes the use of the tool_dependencies.xml file.  I've not fully investigated the cause, but the the error message is likely generated from here: https://github.com/galaxyproject/galaxy/blob/dev/lib/tool_shed/dependencies/attribute_handlers.py#L173.

A complex repository dependency relationship like this was never intended to build a relationship with other tools (i.e. the data manager tools defined in this recipe), but to repositories that provide a recipe for installing a package (i.e. a tool dependency).  Complex repository dependencies allow many tool shed repository installations into a single Galaxy instance where any number of the contained tools across installed repositories can all reference a single installation of a specific tool dependency package.

I'm wondering about the plan for defining complex repository dependencies like this.  With the introduction of conda, the use of tool_dependencies.xml has been deprecated, and fully eliminated for many recipes.  I think that in most (likely all) cases, the use of complex repository dependencies using the repository_dependencies.xml file should be deprecated as well.  At the time these features were implemented, the complex repository dependencies assumed the use of the tool_dependencies.xml file, and the general purpose was to make sure 3rd party dependencies (packages) needed by upstream tools were all available.

I think the likely fix here is to simple eliminate the repository_dependencies.xml file and add information to the tools' help section regarding the need to have the 2 associated data manager tools installed into the same Galaxy instance.

nsoranzo:
I'm not sure if this is the same problem @gregvonkuster describes, but `data_manager_humann2_database_downloader` is not installable:
```
This revision can be installed: False
```

nsoranzo:
The HUMAnN2 data manager should be fixed by https://github.com/galaxyproject/tools-iuc/pull/1204 .

bebatut:
Thanks a lot!

nsoranzo:
I've successfully upload `data_manager_humann2_database_downloader` and all `humann2` tools to both TTS and MTS, so the issue is fully resolved.

@gregvonkuster Luckily there is no problem with `repository_dependencies.xml` files, which can continue to be used for suites.

mvdbeek:
> @gregvonkuster Luckily there is no problem with repository_dependencies.xml files, which can continue to be used for suites.

This does require that we always install the repository dependencies though. Currently that means you have to install TS and conda dependencies for suites. That should be cleaned up at some point.

bebatut:
Thanks a lot to all of you!

nsoranzo:
@mvdbeek I don't get it, if I try to install `suite_humann2` for example, there is no TS package as dependency.

mvdbeek:
screen shot 2017-03-13 at 6 07 20 pm

If there are any TS dependencies in there you would get the packages in status `new`.

mvdbeek:
In this case for example I don't think there should be a checkbox. It should just list the repository dependencies.

lparsons:
> Currently that means you have to install TS and conda dependencies for suites. That should be cleaned up at some point.

This has been an issue for me for some time (e.g. Trinity suite). I'd prefer to only install conda deps, but that's not possible with suites (atm).

PR #1196, opened 2017-03-06
Add MultiGPS version 0.73.0

gregvonkuster:


gregvonkuster:
Hmm..no idea why tests fail with history differences as planemo tests pass locally for me:
```
$ planemo --version
planemo, version 0.38.1
$ planemo test --galaxy_root /home/greg/work/git_workspace/galaxy --conda_dependency_resolution --conda_auto_install --conda_ensure_channels iuc,bioconda,r,defaults,conda-forge --conda_prefix /home/greg/_conda --conda_exec /home/greg/_conda/bin/conda --conda_debug --test_data /home/greg/work/tools-iuc/multigps/tools-iuc/tools/multigps/test-data .
Testing using galaxy_root /home/greg/work/git_workspace/galaxy
Testing tools with command [cd /home/greg/work/git_workspace/galaxy; if [ -d .venv ] || [ -f dist-eggs.ini ]; then GALAXY_VIRTUAL_ENV=.venv;  else GALAXY_VIRTUAL_ENV=/home/greg/.planemo/gx_venv; fi; export GALAXY_VIRTUAL_ENV; if [ ! -e $GALAXY_VIRTUAL_ENV ]; then /usr/bin/virtualenv -p /usr/bin/python2.7 $GALAXY_VIRTUAL_ENV; fi; [ -e $GALAXY_VIRTUAL_ENV ] && . $GALAXY_VIRTUAL_ENV/bin/activate; sh run_tests.sh $COMMON_STARTUP_ARGS --report_file /home/greg/work/tools-iuc/multigps/tools-iuc/tools/multigps/tool_test_output.html --xunit_report_file /tmp/tmpE7J3Tn/xunit.xml --structured_data_report_file /home/greg/work/tools-iuc/multigps/tools-iuc/tools/multigps/tool_test_output.json functional.test_toolbox]
Activating virtualenv at .venv
...stuff eliminated here...
2017-03-06 13:09:09,088 DEBUG [galaxy.tools.deps] Using dependency samtools version None of type conda
2017-03-06 13:09:09,088 DEBUG [galaxy.tools.deps] Using dependency samtools version None of type conda
2017-03-06 13:09:12,384 INFO  [galaxy.tools.actions] Handled output named replicates_counts for tool multigps (48.342 ms)
2017-03-06 13:09:12,407 INFO  [galaxy.tools.actions] Handled output named all_events_table for tool multigps (23.146 ms)
2017-03-06 13:09:12,421 INFO  [galaxy.tools.actions] Handled output named experiment_events for tool multigps (13.871 ms)
2017-03-06 13:09:12,435 INFO  [galaxy.tools.actions] Handled output named output_html for tool multigps (13.563 ms)
2017-03-06 13:09:12,445 INFO  [galaxy.tools.actions] Added output datasets to history (10.769 ms)
2017-03-06 13:09:12,474 INFO  [galaxy.tools.actions] Verified access to datasets for Job[unflushed,tool_id=multigps] (15.005 ms)
2017-03-06 13:09:12,474 INFO  [galaxy.tools.actions] Setup for job Job[unflushed,tool_id=multigps] complete, ready to flush (28.620 ms)
2017-03-06 13:09:12,491 INFO  [galaxy.tools.actions] Flushed transaction for job Job[id=5,tool_id=multigps] (17.042 ms)
2017-03-06 13:09:12,876 INFO  [galaxy.jobs.handler] (5) Job dispatched
2017-03-06 13:09:13,290 DEBUG [galaxy.tools.deps] Using dependency multigps version 0.73 of type conda
2017-03-06 13:09:13,291 DEBUG [galaxy.tools.deps] Using dependency multigps version 0.73 of type conda
2017-03-06 13:09:13,319 INFO  [galaxy.jobs.command_factory] Built script [/tmp/tmpE7J3Tn/job_working_directory/000/5/tool_script.sh] for tool command [[ "$CONDA_DEFAULT_ENV" = "/home/greg/_conda/envs/__multigps@0.73" ] || . /home/greg/_conda/bin/activate '/home/greg/_conda/envs/__multigps@0.73' > conda_activate.log 2>&1 ; mkdir -p /tmp/tmpE7J3Tn/job_working_directory/000/5/dataset_11_files && multigps --eventsaretxt --meme1proc --expt '/tmp/tmpE7J3Tn/files/000/dataset_6.dat' --format scidx --ctrl '/tmp/tmpE7J3Tn/files/000/dataset_7.dat' --threads ${GALAXY_SLOTS:-4} --geninfo '/home/greg/work/git_workspace/galaxy/tool-data/shared/ucsc/chrom/hg19.len' 2>&1 --out '/tmp/tmpE7J3Tn/job_working_directory/000/5/dataset_11_files' && cp /tmp/tmpE7J3Tn/job_working_directory/000/5/dataset_11_files/*.events.txt '/tmp/tmpE7J3Tn/files/000/dataset_10.dat' || true && mv /tmp/tmpE7J3Tn/job_working_directory/000/5/dataset_11_files/*.html '/tmp/tmpE7J3Tn/files/000/dataset_11.dat' || true && mv /tmp/tmpE7J3Tn/job_working_directory/000/5/dataset_11_files/*.table.txt '/tmp/tmpE7J3Tn/files/000/dataset_9.dat' || true && mv /tmp/tmpE7J3Tn/job_working_directory/000/5/dataset_11_files/*.counts '/tmp/tmpE7J3Tn/files/000/dataset_8.dat' || true]
2017-03-06 13:09:13,774 DEBUG [galaxy.tools.deps] Using dependency samtools version None of type conda
2017-03-06 13:09:13,775 DEBUG [galaxy.tools.deps] Using dependency samtools version None of type conda
ok

----------------------------------------------------------------------
XML: /tmp/tmpE7J3Tn/xunit.xml
----------------------------------------------------------------------
Ran 2 tests in 110.536s

OK
2017-03-06 13:10:00,697 INFO  [test_driver] Shutting down
2017-03-06 13:10:00,697 INFO  [test_driver] Shutting down embedded galaxy web server
2017-03-06 13:10:00,700 INFO  [test_driver] Embedded web server galaxy stopped
2017-03-06 13:10:00,700 INFO  [test_driver] Stopping application galaxy
2017-03-06 13:10:00,700 INFO  [galaxy.jobs.handler] sending stop signal to worker thread
2017-03-06 13:10:00,700 INFO  [galaxy.jobs.handler] job handler queue stopped
2017-03-06 13:10:00,700 INFO  [galaxy.jobs.runners] TaskRunner: Sending stop signal to 2 worker threads
2017-03-06 13:10:00,700 INFO  [galaxy.jobs.runners] LocalRunner: Sending stop signal to 4 worker threads
2017-03-06 13:10:00,701 INFO  [galaxy.jobs.handler] sending stop signal to worker thread
2017-03-06 13:10:00,701 INFO  [galaxy.jobs.handler] job handler stop queue stopped
2017-03-06 13:10:00,701 INFO  [test_driver] Application galaxy stopped.
2017-03-06 13:10:01,556 INFO  [galaxy.jobs.handler] sending stop signal to worker thread
2017-03-06 13:10:01,557 INFO  [galaxy.jobs.handler] job handler queue stopped
2017-03-06 13:10:01,557 INFO  [galaxy.jobs.runners] TaskRunner: Sending stop signal to 2 worker threads
2017-03-06 13:10:01,557 INFO  [galaxy.jobs.runners] LocalRunner: Sending stop signal to 4 worker threads
2017-03-06 13:10:01,557 INFO  [galaxy.jobs.handler] sending stop signal to worker thread
2017-03-06 13:10:01,557 INFO  [galaxy.jobs.handler] job handler stop queue stopped
Testing complete. HTML report is in "/home/greg/work/tools-iuc/multigps/tools-iuc/tools/multigps/tool_test_output.html".
All 2 test(s) executed passed.
multigps[0]: passed
multigps[1]: passed
```

jmchilton:
``hg19`` is probably not available in a Planemo created environment but is available in your ``--galaxy_root``.

Does anyone know if the IUC has any other tools that specify a ``dbkey`` on a test input the way this tool does?

nsoranzo:
@jmchilton meme

gregvonkuster:
@jmchilton My datasets are simply named hg19 because that is the data I am using, but don't need to be.  The tool allows for, but does not require a reference genome, and the tests I have do not use the option that requires the use of the reference genome.  So I don't think that the problem is because I am using hg19 test data, unless I am mistaken.  Am I missing something on this?

jmchilton:
I just assumed based on output in Travis:

```
---------------------- >> begin tool stdout << -----------------------
MultiGPS version 0.73
Invalid genome info file name
----------------------- >> end tool stdout << ------------------------
---------------------- >> begin tool stderr << -----------------------
mv: cannot stat `/tmp/tmpuCg8Md/job_working_directory/000/2/dataset_5_files/*.html': No such file or directory
mv: cannot stat `/tmp/tmpuCg8Md/job_working_directory/000/2/dataset_5_files/*.table.txt': No such file or directory
mv: cannot stat `/tmp/tmpuCg8Md/job_working_directory/000/2/dataset_5_files/*.counts': No such file or directory
```

gregvonkuster:
Ah, ok - thanks!

gregvonkuster:
Should I add this tool to the blacklist since the test won't pass with travis?  This shouldn't keep the tool from being merged, right?

tools/multigps/.shed.yml
```diff
@@ -0,0 +1,13 @@
+name: multigps
+owner: iuc
+description: Contains a tool that analyzes collections of multi-condition ChIP-seq data.
+homepage_url: http://mahonylab.org/software/multigps/
+long_description: |
+  Contains a tool that runs MultiGPS, a framework for analyzing collections of multi-condition ChIP-seq
+  datasets and characterizing differential binding events between conditions.  MultiGPS encourages consistency
+  in the reported binding event locations across conditions and provides accurate estimation of ChIP enrichment
+  levels at each event.  MultiGPS loads all data to memory, so you will need a lot of available memory if you
```
yhoogstrate: Hi @gregvonkuster, do you have any idea what is 'a lot'? 1gb? 10gb? 100gb? Please write this in a README file, that will be automatically shown to the Galaxy admin during installation.

tools/multigps/.shed.yml
```diff
@@ -0,0 +1,13 @@
+name: multigps
+owner: iuc
+description: Contains a tool that analyzes collections of multi-condition ChIP-seq data.
```
yhoogstrate: `description: Analyzes collections of multi-condition ChIP-seq data.`

tools/multigps/multigps.xml
```diff
@@ -0,0 +1,457 @@
...
+                        --nodifftests
+                    #end if
+                    --edgerod $rbec.edgerod
+                    --diffp $rbec.diffp
+                #end if
+            #end if
+            2>&1
+            --out '$output_html.files_path'
+            && cp $output_dir/*.events.txt '$experiment_events' || true
+            && mv $output_dir/*.html '$output_html' || true
```
yhoogstrate: ` || true` is a hack to bypass '&&'. This needs appropriate if statements

tools/multigps/multigps.xml
```diff
@@ -0,0 +1,457 @@
...
+                    --q $rbec.minqvalue
+                    --minfold $rbec.minfold
+                    #if str($rbec.nodifftests) == 'no':
+                        --nodifftests
+                    #end if
+                    --edgerod $rbec.edgerod
+                    --diffp $rbec.diffp
+                #end if
+            #end if
+            2>&1
```
yhoogstrate: This is not necesary if <stdio> or <command> (https://docs.galaxyproject.org/en/latest/dev/schema.html#detect-errors) is set correctly

tools/multigps/multigps.xml
```diff
@@ -0,0 +1,457 @@
...
+                #set bmsc = $aoc.binding_model_smoothing_cond
+                #set gmsc = $aoc.gauss_model_smoothing_cond
+                #set rbec = $aoc.report_binding_events_cond
+                #set rloc = $aoc.reads_limits_options_cond
+                #set sdc = $aoc.scale_data_cond
+                #set umc = $aoc.use_motif_cond
+                #if str($umc.use_motif) == 'yes':
+                    #set rgc = $umc.reference_genome_cond
+                    #if str($rgc.reference_genome_source) == 'cached':
+                        #set seq = $rgc.reference_genome.fields.path
```
yhoogstrate: `$seq` is used at just one place. I think this would be fine too:

```
                      --seq
 +                    #if str($rgc.reference_genome_source) == 'cached':
 +                       '${rgc.reference_genome.fields.path}'
 +                    #else:
 +                        '${rgc.reference_genome}'
 +                    #end if
```

tools/multigps/.shed.yml
gregvonkuster: @yhoogstrate This is a pretty difficult measurement to pin down since the amount of memory is dependent upon data content (conditions) and size.  Setting the memory allocation in Galaxy is handled using the ```<env id="_JAVA_OPTIONS">``` tag for a specific runner in the job_conf.xml file.

I don't think you could make a general statement like "if your input data is X in size, you'll need to set your minimum memory limits for this tool to Y" since the equation is much more complex than that.  I think the important thing here is to alert a user that the tool loads everything into memory so that if execution results in  a memory error, the admin can increase the minimum.  These settings are likely not going to be the same across Galaxy instances.

I am certainly willing to add additional details here, but I would need some advice on what to state that would be useful to all that may install this tool.  Also, I am not aware of how the README works during installation.   Is there a tool that uses this feature that provides a good example?

tools/multigps/.shed.yml
gregvonkuster: ok.

tools/multigps/multigps.xml
gregvonkuster: ```|| true``` handles the case where the tool produces no output, so the ```cp``` or ```mv``` would result in a non-zero status, causing Galaxy to set the job state to error.  Since empty datasets are valid results, the dataset state should be ok rather than error.  I've not found a cleaner way to handle this than ```|| true```, but I will change this if you have a better solution.

tools/multigps/multigps.xml
gregvonkuster: I've used this to capture the significant logging that Java does to stdout.  I felt this was the cleanest way to do this within the cheetah syntax, but there certainly may be a better way.  If this tools included a Python script, I could have done the same thing I did here: https://github.com/bgruening/galaxytools/blob/master/tools/image_processing/imagej2/imagej2_base_utils.py#L97, but this is not possible in cheetah.  If the significant Java stdout logging is not contained, it is displayed in the history item as part of the blurb.  Please let me know if you have a better way of eliminating the Java stdout logging and I'll use your approach here.

tools/multigps/multigps.xml
gregvonkuster: ok

gregvonkuster:
@yhoogstrate Thanks for the review!

tools/multigps/.shed.yml
yhoogstrate: https://github.com/galaxyproject/tools-iuc/blob/master/tools/featurecounts/README.rst
:arrow_down: 
image

gregvonkuster: Thanks @yhoogstrate Does my last commit bring this up to par?

gregvonkuster:
Are any additional changes needed for this in order to merge?  Should the tool be added to the blacklist or should it be allowed to err out since the problem lies with planemo?

tools/multigps/multigps.xml
yhoogstrate: I think this could be captured with an `if`-statement in bash (http://stackoverflow.com/questions/638975/how-do-i-tell-if-a-regular-file-does-not-exist-in-bash) and a `for`-loop if multiple files may need to be moved.

tools/multigps/multigps.xml
yhoogstrate: I was just thinking of leaving it to stderr, that's all. There are no best practice on this so am I fine with it :+1: .

tools/multigps/.shed.yml
yhoogstrate: absolutely

gregvonkuster:
Hmm...ok, sorry for merging the PR.  :(  I am not clear on the rules for merging in tools-iuc, so I won't do this again.  I asked if there was any additional changes needed and didn't get a response for 4 days, so figured nothing further was necessary.  Is there a document that provides clear details of the process for merging a PR in tools-iuc?

tools/multigps/multigps.xml
gregvonkuster: Can you let me know your concerns for using the ```|| true``` approach?  I don't consider is a hack to bypass &&, but I very well could be missing something here.

yhoogstrate: `|| true` suppresses all possible errors that `cp` can give (full disk, wrong permissions) while you want those errors to be captured and reported

gregvonkuster: Ok, thanks - this makes sense.  I've eliminated using these in https://github.com/galaxyproject/tools-iuc/pull/1206 - looking at other tools that mv outputs like this don't use it, so I'll follow the standard here.

gregvonkuster:
How should I now deal with the failing test?  Should I add the tool to the blacklist?

yhoogstrate:
I think skipping all tests is more dangerous than excluding the dbkey feature. Or do you need it?

I would then prefer to comment `<validator type="unspecified_build" />` and `<filter type="data_meta" key="dbkey" ref="expt" column="1"/>` out until this gets fixed in planemo.

gregvonkuster:
Commenting out the validator and filter for the dynamic select lists will make the tool unusable.  I'm not sure on what is missing in planemo that causes the tests to fail.  @jmchilton would changing this tag:

```<param name="expt" value="expt_hg19.scidx" ftype="scidx" dbkey="hg19" />```

to be this instead:

```<param name="expt" value="expt_hg19.scidx" ftype="scidx" />```

allow the tests to pass?

**Comment by yhoogstrate** [comment]
In what way will that make it unusable? As far as I can see it's only being used to make sure there reference fasta file is on the same dbkey, or is there something I don't see?

**Comment by gregvonkuster** [comment]
Hmm...maybe I've forgotten how the framework processes these.  If it is ok to eliminate the combination of these 2 items...
```
<validator type="unspecified_build" />
<filter type="data_meta" key="dbkey" ref="expt" column="1"/>
```
...and the tool will work as expected, then why are the combination of these used in so many other tools?  My understanding is that the tool framework will dynamically populate the "reference genome" select list based on the filter, which itself is defined for a specified genome associated with the selected input.  This feature restricts the user from being able to attempt to associate a genome for which an entry does not exist in the data tables registry for the Galaxy instance.

Does eliminating these items properly populate the dynamic select list for the reference genome with all possible options available in the data tables, allowing the user to choose any of them?  If so, does everything work as expected throughout the tool's execution?

Do the planemo tests fail for all of the other tools that use this feature?

gregvonkuster:
I'm also wondering if eliminating this ```<validator type="unspecified_build" />``` on this tool's input would adversely affect workflows that include tools that use the dbkey in various ways.  Would eliminating the dbeky for the input on this tool eliminate the dbkey on all downstream tools in a workflow?

PR #1200, opened 2017-03-12
fixes for broken bwameth datamager 

bwlang:
before this patch the data manager fails with a key not found dict error

### DIFF
--- a/data_managers/data_manager_bwameth_index_builder/data_manager/bwameth_index_builder.py
+++ b/data_managers/data_manager_bwameth_index_builder/data_manager/bwameth_index_builder.py
@@ -24,7 +24,7 @@ def build_bwameth_index(data_manager_dict, params, args):
     if return_code:
         print >> sys.stderr, "Error building index."
         sys.exit( return_code )
-    data_table_entry = dict(value=args.dbkey, dbkey=args.dbkey, name=args.name, path=args.dbkey)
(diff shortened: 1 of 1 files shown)


data_managers/data_manager_bwameth_index_builder/data_manager/bwameth_index_builder.py
```diff
@@ -24,7 +24,7 @@ def build_bwameth_index(data_manager_dict, params, args):
     if return_code:
         print >> sys.stderr, "Error building index."
         sys.exit( return_code )
-    data_table_entry = dict(value=args.dbKey, dbkey=args.dbkey, name=args.name, path=args.dbKey)
+    data_table_entry = dict(value=args.dbkey, dbkey=args.dbkey, name=args.name, path=fasta_base_name)
```
mvdbeek: I think this should be `sym_linked_fasta_filename`, which points to a symlink in the target directory, and I suppose that doubles as the basename for the bwa indices. Does that look right to you, @dpryan79 ?

dpryan79: @mvdbeek I believe you're correct.

bgruening:
ping @mvdbeek @dpryan79

mvdbeek:
Excellent stuff, thanks everyone!

Issue #1130, opened 2017-01-18
GEMINI - Later version


xirasasa:
Do you plan to integrate the latest version of GEMINI (0.19.1)?

bgruening:
@xirasasa this is all a time issue, we would love to get help from the community to update the tools.
A first step would be to add tests to all tools and then update them.

gregvonkuster:
This work is started with this PR: https://github.com/galaxyproject/tools-iuc/pull/1202

xirasasa:
Thank you, we appreciate it!
Sandhya

PR #1202, opened 2017-03-13
Condafy data_manager_gemini_download tool

gregvonkuster:
And update the gemini version to 0.19.1.  This is the first step for this issue: https://github.com/galaxyproject/tools-iuc/issues/1130


bgruening:
Thanks @gregvonkuster for working on this. Are you planning to put the tool in here as well? I don't think it makes sense to update the data manager without the tools.

gregvonkuster:
Yes, I am taking look.  Regarding the tabix requirement for the current tool, I'm wondering if I should create a conda recipe based on this: http://wiki.wubrowse.org/How_to_install_tabix (this is my current approach).  However, it is also a component of htslib (https://github.com/samtools/htslib) but that package includes a lot more stuff.  Let me know if you'd rather have me use it though.

bgruening:
Depending on htslib is ok, I think.

yhoogstrate:
This should be pending on: https://github.com/galaxyproject/galaxy/pull/3686

bgruening:
Thanks @gregvonkuster!

PR #1206, opened 2017-03-15
Eliminate use of || true when moving outputs

gregvonkuster:


bgruening:
@gregvonkuster what was this problem here?

mvdbeek:
@bgruening https://github.com/galaxyproject/tools-iuc/pull/1196#discussion_r105896374

bgruening:
@mvdbeek this I know. But I don't know why this PR is failing. Anything we can do here?

mvdbeek:
it was failing before. It's probably the dbkey validator that doesn't work in testing.

gregvonkuster:
@bgruening @mvdbeek Yes, the tests for this tools pass just fine using my local Galaxy instance as shown in the above comments.  Per @jmchilton comments above, it seems that planemo is missing some feature to handle these tests, but I'm not sure just what it is.  I don't believe it is this...
```
<validator type="no_options" message="A built-in reference genome is not available for the build associated with the selected input file"/>
```
...since many tools include this same validator:
```
~/tools-iuc/bcftools
~/tools-iuc/busco
~/tools-iuc/bwameth
many, many more...
```
I wonder if it is the use of dbkey in this test tag...
```
<param name="expt" value="expt_hg19.scidx" ftype="scidx" dbkey="hg19" />
```
...but again, many other tool tests include this attribute:
```
tools-iuc/extract_genomic_dna
tools-iuc/meme
many tools in tools-devteam repo
```
I will add this tool to the blacklist if you feel this would be best.

blankenberg:
The dbkey is being set correctly, but the testing environment most likely doesn't have `.len` files available, as they aren't distributed with the distribution, but can be pulled down via a script. 

If you look at the tool command-line it contains:
```
--geninfo '/tmp/tmplBJzc1/galaxy-dev/tool-data/shared/ucsc/chrom/hg19.len'
```

blankenberg:
For many GOPs tools, when a `.len` file isn't available, it will assume defaults for chromosome lengths, usually something like 255Mbases (IIRC).

gregvonkuster:
@blankenberg Thanks for the insight on this!  Now that you mentioned it, I remember this issue from my past life!  ;)

So what is the best way to handle this?  To me, the optimal solution is for planemo to install these ```.len``` files for testing tools that require them.  It seems that setting max value for a chromosome length just so that planemo tests will pass is not ideal (doesn't it introduce potential errors if the chromosome length is not precise?).  If the consensus is to enhance planemo to do this, perhaps the IUC can decide on the approach we should take to handle this category of tools (i.e., where the current test components do not include features needed for testing the tool).  I have several tools in development (close to being finalized) that fall into this category.  They require the use of very large scaffolds datasets that are installed via a Data Manager tool.

For this particular tool, I don't own the Java code, so I'll have to attempt to get the author to agree to set a max default value for the chromosome length if the ```.len``` file does not exist within the Galaxy environment for the selected genome.  If the consensus is to use this approach, I'll do it asap.

jmchilton:
This feels less like a Planemo issue and more like a Galaxy issue - we could hack something into Planemo but I would rather fix Galaxy.

Two entirely reasonable and not too difficult approaches would be to either:
- Add a galaxy.ini config option that Planemo could enable (say ``auto_fetch_chrom_lens``) to just fetch these as needed. Right now ``$chormInfo`` is populated as a string - but instead it could be an object that when evaluated would check for the existence of the target file and use the ``cron/build_chrom_db.py`` to populate the target and then return that string.
- We could just pick a few databases to support for testing and pre-populate test values - the way we do for tools and datatypes and such.

mvdbeek:
Can't we use (this)[http://planemo.readthedocs.io/en/latest/writing_how_do_i.html#test-index-loc-data] apparoach for len files?

gregvonkuster:
I just spoke with @nekrut and he wants this tool on Galaxy main, so merging.  We'll need to handle the failing tests with changes to Galaxy - I agree with @jmchilton on this.  Should I open an issue for having Galaxy automatically install ```.len``` files for popular chromosomes?

PR #1207, opened 2017-03-15
Updates to STAR

yhoogstrate:
This patch updates STAR to 2.5.2 and now supports:
 - [x] fastqgz / fastqsangergz, thanks to @hrhotz
 - [x] data manager (indices are not explicitly built with GTF files because these can be added during tool runtime)
 - [x] gene counting (as built-in star option)
 - [x] a few additional arguments
 - [x] bioconda dependencies (DM as well as tool)


data_managers/data_manager_star_index_builder/data_manager/rna_star_index_builder.xml
```diff
@@ -0,0 +1,107 @@
+<tool id="rna_star_index_builder_data_manager" name="RNA STAR index" tool_type="manage_data" version="0.0.4" profile="17.01">
+    <description>builder</description>
+    
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    
+    <expand macro="requirements" />
+    
+    <command interpreter="python"><![CDATA[
```
bgruening: Can we remove the interpreter?

data_managers/data_manager_star_index_builder/data_manager/rna_star_index_builder.xml
```diff
@@ -0,0 +1,107 @@
...
+        #set target_directory = $params[ 'output_data' ][0]['extra_files_path'].encode('ascii', 'replace')
+        #set subdir = os.path.basename(target_directory)
+
+        mkdir -p '${target_directory}/${subdir}' &&
+
+        STAR
+            --runMode genomeGenerate
+            --genomeFastaFiles '${all_fasta_source.fields.path}'
+            --genomeDir '${target_directory}/${subdir}'
+            --runThreadN 2  &&
```
bgruening: GALAXY_SLOTS?

tools/rgrnastar/rg_rnaStar.xml
```diff
@@ -1,417 +1,456 @@
...
+    
     <stdio>
         <regex match="EXITING: FATAL INPUT ERROR:" source="both" level="fatal"/>
         <regex match="EXITING: fatal error trying to allocate genome arrays, exception thrown: std::bad_alloc" source="both" level="fatal"/>
         <regex match="\[sam_read1\] missing header\? Abort!" source="both" level="fatal"/>
         <regex match=".*" source="both" level="warning" description="Some stderr/stdout text"/>
     </stdio>
 
+<!--
+    important quote (https://groups.google.com/forum/#!topic/rna-star/q4zGzlPgwXY):
```
bgruening: Nice one! Good to know. I still recall the paper were they discouraged users to do this.

tools/rgrnastar/rg_rnaStar.xml
```diff
@@ -1,417 +1,456 @@
...
+-->
     <command><![CDATA[
     ## Create temporary index for custom reference
     #if str($refGenomeSource.genomeSource) == 'history':
         mkdir -p tempstargenomedir &&
         STAR
             --runMode genomeGenerate
             --genomeDir "tempstargenomedir"
-            --genomeFastaFiles "$refGenomeSource.ownFile"
+            --genomeFastaFiles "$refGenomeSource.genomeFastaFiles"
```
bgruening: single quotes

tools/rgrnastar/rg_rnaStar.xml
```diff
@@ -1,417 +1,456 @@
...
-        "$singlePaired.input.forward" "$singlePaired.input.reverse"
-    #else
-        "$singlePaired.input1"
-        #if str($singlePaired.sPaired) == "paired"
-            "$singlePaired.input2"
+        --runThreadN \${GALAXY_SLOTS:-4}
+        --genomeLoad NoSharedMemory
+        --genomeDir
+        #if str($refGenomeSource.genomeSource) == 'history':
+            "tempstargenomedir"
```
bgruening: single quotes

bgruening:
@jmchilton this is an interesting one. It seems that the sqlite lock error is reproducible on travis at least.

data_managers/data_manager_star_index_builder/.shed.yml
```diff
@@ -0,0 +1,11 @@
+categories:
+- Data Managers
+description: Data Manager for building RNA STAR indexes
+long_description: |
+  STAR is an ultrafast universal RNA-seq aligner.
+  WARNING! Indexing will consume mucm  RAM (9*size of fasta file, 30G
```
nsoranzo: mucm?

data_managers/data_manager_star_index_builder/data_manager_conf.xml
```diff
@@ -1,22 +1,23 @@
 <?xml version="1.0"?>
 <data_managers>
-    
-    <data_manager tool_file="data_manager/rnastar_index_builder.xml" id="rnastar_index_builder" version="0.0.1">
-        <data_table name="rnastar_index">
+    <data_manager tool_file="data_manager/rna_star_index_builder.xml" id="rna_star_index_builder" version="0.0.3">
+        <data_table name="RNA_STAR_index">
```
nsoranzo: Not sure using uppercase in table names or filenames is a good idea.

data_managers/data_manager_star_index_builder/.shed.yml
yhoogstrate: Descriptions and README's definitely require some extra work

dpryan79:
@yhoogstrate What (other than making TravisCI happy) still needs to be done for this? I'd quite like this and would be happy to make a PR on your fork to help expedite things :)

yhoogstrate:
Wow awesome to get this forward @dpryan79  :)

I am a bit stuck on how to proceed with custom GTF files or include them in the build. I asked the author and he replied the following to me:

> building the genome without annotations, and supplying them at the run time with a specific --sjdbOverhang is a good idea.
> However, I see a few disadvantages:
> 1. Including junctions into the genome will add few minutes (3-5) to each run.
> 2. You will need to calculate the read length before the mapping to supply to --sjdbOverhang
> 3. If you have reads of varied length choosing the "ideal" --sjdbOverhang is not possible.
> 4. You cannot use shared memory options which allow running multiple mapping jobs to the same genome.
> 
> In general, for reads longer than 50b there is very little mapping accuracy improvement when selecting ideal --sjdbOverhang.
> 
> So I would recommend the following -
> 1. Generate two genome indexes: (i) without annotations and (ii) with default --sjdbOverhang
> 2. For reads shorter that 50b, use the (i) index with the annotations and --sjdbOverhang=ReadLength-1, and for >=50b use the (ii) index.
> For <50b reads with varying length find (or estimate) the max length to use as ReadLength.

So both with and without may be important use-cases.

I am thinking of a new way of doing this but I would also appreciate some feedback on this. This would be to add a column to the reference loc file <gtf_file_provided> that is either set to 1 or 0, or to 'with-gtf' and 'without-gtf' or something to distinguish between them. Then for the client part there needs to be selected for either a reference WITH or reference WITHOUT custom GTF. This would look something like this:

```
<conditional>
	<param type="select">
		<option value="with-gtf">use genome reference with builtin gene-model</option>
		<option value="witout-gtf">use genome reference without builtin gene-model</option>
	</param>
	<when value="with-gtf">
		<param name="index" type="select" label="Select reference genome" help="If your genome of interest is not listed, contact the Galaxy team">
			<options from_data_table="rna_star_indexes">
				<filter type="static_value" value="1"/>
			</options>
		</param>
	</when>
	<when value="without-gtf">
		<param name="index" type="select" label="Select reference genome" help="If your genome of interest is not listed, contact the Galaxy team">
			<options from_data_table="rna_star_indexes">
				<filter type="static_value" value="0"/>
			</options>
		</param>
		
		<conditional to add GTF file>
			<param ... gtf file, optional/>
			<param sjdbOverhang />
		</conditional>
	</when>
</conditional>
```
I just don't know what is the best way to do this

dpryan79:
@yhoogstrate Ah yes, Alex does enjoy his complexity and optimality.

The strategy we've taken is to never create indices with GTF files, they get updated to quickly and the STAR indices are...large. Anyway, for how this should be handled in Galaxy I like your proposal of allowing both. Perhaps the data manager should warn people of the space repercussions of creating a new index for every GTF and read length combination they're likely to use...

yhoogstrate:
We can also decide to already add the column and not implement the feature (yet?). If the column is present in the loc file, the feature can always be added later on without breaking loc file hierarchy.

dpryan79:
As the Germans would say, "Wenn schon denn schon", so if we're going to do it we might as well go all the way with it. I'm happy to give that a whirl after the holidays (so starting Tuesday).

dpryan79:
@yhoogstrate: I seem to have things working in my branch now, though it needs further testing for confirmation. I have a questions regarding data table naming that's related to this, though. The current STAR wrapper uses a data table called `rnastar_index`, which sources `rnastar_index.loc`. The current version of this wrapper changes the data table name to `RNA_STAR_index`, but still sources `rnastar_index.loc`. I would think that if the data table name is changes then the file on the disk should also be changed.

Since we need to change the format of the table anyway, which will require manual intervention on the part of admins, I would propose either (A) keeping all names the same or (B) modifying both `rnastar_index` and `rnastar_index.loc` to something else. I suspect that the latter would be the proper route, since otherwise older versions of STAR will break. Maybe just naming things `rnastar_index2` and `rnastar_index2.loc` would suffice (I have no actual preference on the naming).

mvdbeek:
> Maybe just naming things rnastar_index2 and rnastar_index2.loc would suffice (I have no actual preference on the naming).

that sounds good to me.

yhoogstrate:
@dpryan79 as pointed out by @nsoranzo, the uppercase should be reverted back to lower case. I don't know what's the right choice here, so I'm fine following @mvdbeek's suggestion.

dpryan79:
@yhoogstrate You'll find that implemented in my PR on your fork :)

bgruening:
@yhoogstrate do you feel like removing the WIP label, so we can merge it? :)

yhoogstrate:
I will try to find time to test the DM by hand. If I don't manage to get this done in a reasonable time, feel free to merge (or review).

yhoogstrate:
I can't approve your work @dpryan79 because it's my own PR, but I tested creating a genome build with and without GTF and they're both selectable in the tool wrapper. Thanks for this awesome piece of work and pushing it forward! :+1: from my side

dpryan79:
@yhoogstrate Happy for the confirmation that it works :) I'm always a bit nervous about working on the data managers, I find them easy to muck up.

bgruening:
Works also for me! Let's merge this!
Thanks a ton!

PR #1209, opened 2017-03-17
Prokka version bump to release 1.12

Slugger70:


bgruening:
I think only the updated test-data is needed to get it running! :)

.travis.yml
```diff
@@ -15,7 +15,7 @@ before_install:
   - unset JAVA_HOME
 
 install:
-  - pip install flake8 flake8-import-order planemo
+  - pip install flake8 flake8-import-order planemo==0.39
```
nsoranzo: What's wrong with 0.40.0 ? Any GitHub issue?

Slugger70: Travis kept stopping half way through testing. Running planemo 0.40 locally wouldn't clone release_17.01 or 16.10 and if I didn't specify a release it failed.

bgruening: @nsoranzo I do not know what is not working, but have seen this now 3 times. It does not happen evertime. Currently in Rom on a ELIXIR meeting, why are you not here? :)

bgruening:
Thanks @Slugger70!

.travis.yml
nsoranzo: I've tried installing locally Planemo 0.40.0 with pip in a virtualenv, it downloaded Galaxy release_17.01 and ran tests for a tool without issues. @Slugger70 if you can still reproduce it, can you open an issue on https://github.com/galaxyproject/planemo/issues with the traceback/error, please?

Also, ping @jmchilton

bgruening: It just hangs, there is no traceback :(
It seems that `planemo test` is not executed or returns nothing ...?

Issue #1080, opened 2016-12-15
sra-tools fail download from weird 3 reads per spot issue


Takadonet:
It seems that some sequencing centre have uploaded some SRAs that are paired end libraries but for some reason, they decided to indicate there is 3 reads instead of 2.  Example would be  https://trace.ncbi.nlm.nih.gov/Traces/sra/?run=ERR027434

The first and THIRD reads are the forward and reverse respectively but the second read (_2) is a no length read between forward and reverse.

So fastq-dump will dump out SRA_1 and SRA_**3** instead of SRA_1 and SRA_**2** . Hence the discover datasets in <output> section will fail.  fastq-dump actually indicate that is ignoring _2 because all reads have zero length.

Takadonet:
Fixed in #1212

Issue #970, opened 2016-09-28
SRA-tools FASTQ dump fails when using a file list


blankenberg:
and selecting `apply left and right clips` or `Dump only biological reads` for certain identifiers.

Example, have a file that contains this identifier `SRR042817`. These are single ended reads, and the error occurs whether or not I set `split spot by read pairs` to true or false

results in this error / traceback
```
galaxy.tools.parameters.output_collect DEBUG 2016-09-28 12:32:03,946 (1697) Created dynamic collection dataset for path [/galaxy/database/jobs_directory/001/1697/working/SRR042817_4.fastq] with element identifier [SRR042817:4] for output [list_paired] (1695.361 ms)
galaxy.tools.parameters.output_collect DEBUG 2016-09-28 12:32:04,339 (1697) Add dynamic collection datsets to history for output [list_paired] (393.093 ms)
galaxy.tools.parameters.output_collect ERROR 2016-09-28 12:32:04,369 Problem gathering output collection.
Traceback (most recent call last):
  File "/galaxy/lib/galaxy/tools/parameters/output_collect.py", line 65, in collect_dynamic_collections
    collection_builder.populate()
  File "/galaxy/lib/galaxy/dataset_collections/builder.py", line 87, in populate
    elements = self.build_elements()
  File "/galaxy/lib/galaxy/dataset_collections/builder.py", line 58, in build_elements
    new_elements[identifier] = element.build()
  File "/galaxy/lib/galaxy/dataset_collections/builder.py", line 64, in build
    collection = build_collection( type_plugin, self.build_elements() )
  File "/galaxy/lib/galaxy/dataset_collections/builder.py", line 12, in build_collection
    set_collection_elements( dataset_collection, type, dataset_instances )
  File "/galaxy/lib/galaxy/dataset_collections/builder.py", line 19, in set_collection_elements
    for element in type.generate_elements( dataset_instances ):
  File "/galaxy/lib/galaxy/dataset_collections/types/paired.py", line 24, in generate_elements
    self._validation_failed( INVALID_IDENTIFIERS_MESSAGE )
  File "/galaxy/lib/galaxy/dataset_collections/types/__init__.py", line 23, in _validation_failed
    raise exceptions.ObjectAttributeInvalidException( message )
ObjectAttributeInvalidException: Paired instance must define 'forward' and 'reverse' datasets .
```

[...]

PR #1212, opened 2017-03-20
Address some issues with weird spots from SRA archive

Takadonet:
Changes are to address changes from https://github.com/galaxyproject/tools-iuc/issues/1080 and probably a few others.


Takadonet:
Also will solve #1081

Takadonet:
And also #970  . Try it out @blankenberg .  It should work when toggle on "Dump only biological reads" option. If you don't both collection will be empty. No idea how to handle that without reading each fast(a|q) coming out of sra-tools

mvdbeek:
That looks great to me, thanks a lot @Takadonet

PR #1229, opened 2017-04-03
consider packages and data-managers as well

bgruening:


bgruening:
ping @nsoranzo @mvdbeek

nsoranzo:
Should we skip `planemo test` for packages and data managers?

bgruening:
This would at least save a few seconds I guess.

yhoogstrate:
Is this just for linting or is `planemo test` now capable of testing data managers?

nsoranzo:
As I've mentioned above, `planemo test` does not work for data managers yet and will never work on packages, so it should probably be skipped. This is for linting and especially for automatic upload.

bgruening:
@nsoranzo can we merge this then or try to skip the planemo test before?

bgruening:
@nsoranzo I have added the skip, the failing test is due to and ssl certification error - I hope they will fix it. Can we merge?

tools/fastqc/rgFastQC.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="fastqc" name="FastQC" version="0.68">
+<tool id="fastqc" name="FastQC" version="0.67">
```
nsoranzo: That should be reverted.

bgruening: Done.

nsoranzo:
> the failing test is due to an SSL certification error - I hope they will fix it.

That's strange, the certificate validates fine with both Firefox and Planemo on my laptop, is it a Travis problem?

bgruening:
It can be a travis problem they updated the image last week or so.

bgruening:
:) Thanks.

PR #1232, opened 2017-04-04
Add metagenomeSeq normalization tool: Cumulative Sum Scaling.

blankenberg:


tools/metagenomeseq/metagenomeseq_normalization.xml
```diff
@@ -0,0 +1,148 @@
+<tool id="metagenomeseq_normalizaton" name="metagenomeSeq Normalization" version="1.16.0-0.0.1">
+    <description>Cumulative sum scaling</description>
+    <requirements>
+        <requirement type="package" version="1.16.0">bioconductor-metagenomeseq</requirement>
+        <requirement type="package" version="1.2.0">bioconductor-biomformat</requirement>
+    </requirements>
+    <stdio>
```
bgruening: Can be replaced by `<command detect_errors="exit_code">`

tools/metagenomeseq/metagenomeseq_normalization.xml
```diff
@@ -0,0 +1,148 @@
...
+        </when>
+    </conditional>
+
+    <param name="log2" type="boolean" truevalue="TRUE" falsevalue="FALSE" checked="False" label="Log2 transform scaling" help=""/>
+    <param name="sl" type="integer" value="1000" label="The value to scale by." help="(default=1000)"/>
+
+    <param name="include_outputs" type="select" multiple="True" label="Datasets to create">
+        <option value="output_tabular" selected="True">Normalized Tabular Matrix</option>
+        <!-- <option value="output_biom" selected="True">Normalized BIOM</option> -->
+        <option value="output_stats" selected="True">Simple Statistics</option>
```
bgruening: `true` can also be written lower case, like the `false` below.

tools/metagenomeseq/metagenomeseq_normalization.xml
```diff
@@ -0,0 +1,148 @@
...
+::
+  
+  Requires a BIOM formatted file for input. 
+
+Outputs
+-------
+
+::
+  
+  Creates a normalized and scaled output abundance matrix in Tabular or BIOM (broken: https://github.com/joey711/biomformat/issues/4) format. Additionally can create simple statistics and the RScript.
```
bgruening: Broken BIOM format? This seems to be confusing for the user? What is the consequence of this bug?

bgruening: BIOM is also not an output anymore, it is commented out.

blankenberg: Will update help, thanks for catching. 

Unfortunately, the bioconductor biomformat package writes out something that almost looks like BIOM, but isn't BIOM (see https://github.com/joey711/biomformat/issues/4). Wonder how many other R packages are unfortunately relying on this broken behavior. Reading BIOM does seem to work, however.

bgruening:
Alle guten Dinge sind 3!

PR #1233, opened 2017-04-04
Add Sort Column Order by heading tool.

blankenberg:


tools/column_order_header_sort/column_order_header_sort.xml
```diff
@@ -0,0 +1,38 @@
...
+    <requirements>
+        <requirement type="package" version="3.6.1">python</requirement>
+        <requirement type="package" version="4.1.3">gawk</requirement>
+    </requirements>
+    <stdio>
+        <exit_code range="1:" />
+        <exit_code range=":-1" />
+    </stdio>
+    <command><![CDATA[
+        python '$__tool_directory__/column_order_header_sort.py' '${input_tabular}' '${output_tabular}' '${input_tabular.unsanitized.metadata.delimiter}' '${key_column}'
```
bgruening: Can we have this with line breaks between the parameters?

tools/column_order_header_sort/column_order_header_sort.py
```diff
@@ -0,0 +1,35 @@
...
+key_column = sys.argv[4]
+
+try:
+    key_column = int( key_column ) - 1
+except Exception:
+    key_column = None
+
+header = None
+with open( input_filename, 'r' ) as fh:
+    header = fh.readline().strip( '\r\n' )
```
bgruening: Is there any reason to strip `\r\n` and not `\n`?

tools/column_order_header_sort/column_order_header_sort.xml
```diff
@@ -0,0 +1,38 @@
...
+    ]]>
+    </command>
+    <inputs>
+        <param name="input_tabular" type="data" format="tabular" multiple="False" optional="False" label="Tabular file"/>
+        <param name="key_column" type="data_column" data_ref="input_tabular" value="0" optional="True" label="Identifier column"/>
```
bgruening: Can we add a help here and in the tool-help section what this parameter is doing?

blankenberg:
> @blankenberg looks good, a few comments inline.
> Btw. do you know this tool: https://github.com/bgruening/galaxytools/blob/master/tools/text_processing/column_arrange_by_header/columnArrange.xml
> Will this also solve your use-case?

This would require manually entering, and knowing, all of the possible column names. This doesn't scale and won't work when e.g. pushing a file that contains sample names for column headers through a workflow.

tools/column_order_header_sort/column_order_header_sort.xml
blankenberg: Changed.

tools/column_order_header_sort/column_order_header_sort.xml
blankenberg: Definitely a great idea, thanks, updated help.

tools/column_order_header_sort/column_order_header_sort.py
blankenberg: It is more complete and does not assume POSIX line endings.

peterjc: If anyone was confused, this will strip any and all ``\r`` and ``\n`` (not just the Windows style combination of ``\r\n``.

e.g.

```
>>> "abcde".strip("edf")
'abc'
```

So +1 on stripping both line endings, unless you want to open the file in universal new lines mode instead?

bgruening:
Thanks again!

PR #1239, opened 2017-04-07
Fix stacks container requirement.

jmchilton:
Without this I'm getting an unauthorized request error when trying to run the tool.


bgruening:
Thanks John!

PR #1242, opened 2017-04-10
Fix picard index builder

mvdbeek:
The wrapper was still adjusted to the tool shed package.
Fixes https://github.com/galaxyproject/galaxy/issues/3907


bgruening:
Thanks!

PR #1245, opened 2017-04-10
Fastq.gz support in hisat2

dpryan79:
This PR finishes part of #1128, adding support for gz and bz2 input into hisat2. Additionally, if one selects the "Write unaligned reads..." or "Write aligned reads..." options, then the output type will match the inputs (i.e., fastqsanger.gz input leads to fastqsanger.gz output).

I added a test for gz and bz2 input. I'm hoping to use this sort of structure for the other tools mentioned in #1128, so feedback regarding that would be appreciated.

PR #1246, opened 2017-04-10
Fixes, courtesy of @nsoranzo

dpryan79:
This is a follow up to #1245 that includes fixes that were mentioned by @nsoranzo.


bgruening:
I fixed the older version manually and uploaded it and then merged this one. So ideally both versions should work.

Thanks for fixing so fast.

Issue #1128, opened 2017-01-17
add fastq.gz as supported filetype


bgruening:
Am I correct, that we need to update tools here? 
https://github.com/galaxyproject/galaxy/pull/3145

nsoranzo:
Ping @mvdbeek.

mvdbeek:
If you want to avoid invoking the fastq.gz <> fastq conversion, yes, that was one of my intentions behind the PR.

Some tools that I'm currently using and that support fastq.gz as input
 - bwa
 - bowtie2
 - trimmomatic
 - hisat2
 - sickle
 - fastqc
 - star-fusion

For trimmomatic, sickle and the sra tools I would say an option to output fastq.gz would make sense.

dpryan79:
Has there been any work on this? I would love this to all be natively handled rather than needing to do hacky local solutions. If there's not been any work on it, then I'd be happy to make PRs for everything except star-fusion (we don't use it locally).

mvdbeek:
BWA, trimmomatic and fastqc are already supporting fastq.gz. If you want to tackle the others that would be fantastic!

dpryan79:
OK, my current list, then is:

 - [ ] bowtie2
 - [ ] hisat2
 - [ ] star (I'd like an updated version too, so I'll make it a single PR)
 - [ ] sickle

bgruening:
@dpryan79 great! Please also look at: https://github.com/galaxyproject/tools-iuc/pull/1207

dpryan79:
@bgruening Thanks for alerting me to that, I'll not duplicate the effort!

bgruening:
@yhoogstrate is currently writing up his thesis so he probably welcomes and help :)

dpryan79:
Pfft, silly priorities (good to know, I'll see what I can do to chip in).

dpryan79:
FYI, FastQC only partially supports compressed input. It relies on the item name in the history ending with .gz/.bz2 rather than the data type actually assigned. This stops working in a workflow. For example, if I start with a `fastqsanger.gz` file and put it through a trimmer in a workflow then the name becomes something like `Trim Galore! on data 3: trimmed reads`, which the python wrapper then incorrectly classifies as `fastq`. I'll make a PR so FastQC can handle gzipped files in all cases. Allegedly the wrapper supports bzipped input, but I'm a bit doubtful that that actually works, since that information is never actually given to `fastqc`.

dpryan79:
[...]

PR #1247, opened 2017-04-11
Compressed input for bwameth

dpryan79:
For #1128, this adds support for fastqsanger.gz and fastqsanger.bz2 input into bwameth.


bgruening:
Thanks @dpryan79!

PR #1249, opened 2017-04-11
Fix macs2 typo causing failures.

dannon:


dannon:
@bgruening Looks like this broke with a7a30c6b74c549a3014a407dc4279c1710a47b66

PR #1251, opened 2017-04-12
WIP Compressed input in sickle

dpryan79:
Xref #1128 

This PR adds support for gzipped input/output in Sickle. If the input files are gzipped, then the output files will be too. Bzip2 isn't supported by sickle, so I haven't added it.

I also took the liberty of fixing the tests, which were a bit weird before since the files were incorrectly labeled as fastqsanger (phred+33) when they were actually phred+64.


tools/sickle/sickle.xml
```diff
@@ -1,20 +1,94 @@
...
         <requirement type="package" version="1.33">sickle</requirement>
         <!-- conda dependency -->
         <requirement type="package" version="1.33">sickle-trim</requirement>
     </requirements>
     <version_command>sickle --version | head -n 1</version_command>
-    <command>
+    <command><![CDATA[
+        ## Link in the input files, which also determines the type of the output
+        #if str($readtype.single_or_paired) == "se":
+            #if $readtype.input_single.is_of_type('fastqillumina.gz', 'fastqsanger.gz', 'fastqsolexa.gz'):
```
mvdbeek: I think we can compact this to `#if $readtype.input_single.is_of_type(fastq.gz)`

dpryan79: Excellent point, that would seriously shorten things :)

dpryan79:
Oh, travis is failing, I'll add a WIP to this.

dpryan79:
Actually, I'll redo this, since I see a shorter way to do all of it now.

PR #1253, opened 2017-04-13
Compressed input for sickle

dpryan79:
This is a reworking of #1251.

Now, input files are linked in (so sickle can figure out if they're compressed), but output files no longer are. That output files were linked in was a mistake on my part, since sickle uses the `-g` option to determine whether to compress output or not.

Note, gzipped I/O is supported, bz2 I/O is not.

Xref: https://github.com/galaxyproject/tools-iuc/issues/1128


mvdbeek:
This looks good to me, and I have no objection merging, but what do you think about exposing the compressed option as a boolean like so `<param name="compressed" type="boolean" truevalue="-g" falsevalue="" value="True" label="Output compressed fastq?"` ? I think this does give the user a bit more flexibility.

dpryan79:
@mvdbeek Uncompressed fastq files are a crime against nature. They're bigger and slower to process. I look at this as preventing people from blowing up their quotas :)

mvdbeek:
I agree with you wholeheartedly ;). I guess we should have a fastq to fastq.gz converter in Galaxy for those who, for whatever reason, start with uncompressed fastq.

tools/sickle/sickle.xml
```diff
@@ -88,21 +135,21 @@
             </param>
 
             <when value="se">
-                <param format="fastq" name="input_single" type="data" label="Single-end FASTQ reads" help="(-f)" />
+                <param format="fastq,fastq.gz" name="input_single" type="data" label="Single-end FASTQ reads" help="(-f)" />
             </when>
 
             <when value="pe_combo">
-                <param format="fastq" name="input_combo" type="data" label="Paired-end interleaved FASTQ reads" help="(-c)" />
+                <param format="fastq.fastq.gz" name="input_combo" type="data" label="Paired-end interleaved FASTQ reads" help="(-c)" />
```
nsoranzo: s/fastq.fastq.gz/fastq,fastq.gz/

tools/sickle/sickle.xml
```diff
@@ -1,20 +1,94 @@
...
         <requirement type="package" version="1.33">sickle</requirement>
         <!-- conda dependency -->
         <requirement type="package" version="1.33">sickle-trim</requirement>
     </requirements>
     <version_command>sickle --version | head -n 1</version_command>
-    <command>
+    <command><![CDATA[
+        ## Link in the input files, which also determines the type of the output
+        #if str($readtype.single_or_paired) == "se":
+            #if $readtype.input_single.is_of_type('fastqillumina.gz', 'fastqsanger.gz', 'fastqsolexa.gz'):
```
nsoranzo: Add also `'fastq.gz'` ? (also below, of course)

tools/sickle/sickle.xml
dpryan79: Gah, good catch!

tools/sickle/sickle.xml
```diff
@@ -88,21 +135,21 @@
             </param>
 
             <when value="se">
-                <param format="fastq" name="input_single" type="data" label="Single-end FASTQ reads" help="(-f)" />
+                <param format="fastq,fastq.gz" name="input_single" type="data" label="Single-end FASTQ reads" help="(-f)" />
             </when>
 
             <when value="pe_combo">
-                <param format="fastq" name="input_combo" type="data" label="Paired-end interleaved FASTQ reads" help="(-c)" />
+                <param format="fastq.gz" name="input_combo" type="data" label="Paired-end interleaved FASTQ reads" help="(-c)" />
```
nsoranzo: @dpryan79 You removed "fastq", should be `fastq,fastq.gz` as above, no?

dpryan79: Correct, the previous "fastq.fastq.gz" should have been "fastq,fastq.gz", I just used the wrong punctuation and didn't realize that when I changed it. It's now updated, thanks!

bgruening:
Thanks everyone!

PR #1257, opened 2017-04-17
Pr 1252  suggested changes for SnpEff v4.3k

jj-umn:
Sorry,  I couldn't figure out how to add these changes to PR 1252.  
Updated annotation options for 4.3k.
Removed extra annotation options. Previous version snpEff versions required these to be explicitly included.  Now they are automatically used unless explicitly excluded by -noMotif etc.
Put regulation options in a collapsed section since the list can be long and not frequently used.


tool_collections/snpeff/snpEff.xml
```diff
@@ -131,41 +128,32 @@
                            <filter type="unique_value" column="2" />
                     </options>
                 </param>
-                <param name="extra_annotations" type="select" display="checkboxes" multiple="true" label="Additional annotations">
-                       <help>These are available for only a few genomes</help>
-                       <options from_data_table="snpeffv_annotations">
-                           <filter type="param_value" ref="genomeVersion" key="genome" column="2" />
-                           <filter type="unique_value" column="3" />
-                       </options>
-                </param>
```
nsoranzo: @jj-umn Should we also remove `snpeffv_annotations` from `tool_collections/snpeff/tool_data_table_conf.xml.sample` file and from `data_manager_snpeff`? (I can do that if you agree)

nsoranzo:
Thanks @jj-umn for testing and for the new commits, I've pulled them into #1252. You should be able to push to https://github.com/nsoranzo/tools-iuc `snpeff4.3k` branch, if you have more changes.

Issue #1262, opened 2017-04-19
Declaring packages legacy and marking them as unmaintained


mvdbeek:
Dear @galaxyproject/iuc 

I would like to move all packages on the IUC's github repo into the deprecated folder, so that it is clear that the way forward is to abandon these packages. The infrastructure for using Conda is in place in galaxy since 16.01, and I would consider this mature since 17.01.

This may be a problem for administrators running older galaxy versions, but the likelihood of successfully installing things like R or deeptools with the complete dependency chain from scratch today is low (outdated download links, newer/older gcc versions, etc. pp.).
I personally consider time spent on toolshed packages in 2017 a waste of time (on top of it being a frustrating experience).

Consequently, I think we should also lobby for toolshed package dependency installations to not be the default, starting as soon as possible.

nsoranzo:
:100:

yhoogstrate:
I like this idea too. Do you think this requires to deprecate the tools listed in https://github.com/galaxyproject/tools-iuc/blob/master/.tt_blacklist ?

mvdbeek:
That's a good point. So right now these are:
- [ ] tools/differential_count_models
- [ ] tools/gatk2
- [ ] tools/gff3_rebase
- [ ] tools/progressivemauve
- [ ] tools/rglasso
- [ ] tools/rgrnastar
- [ ] tools/taxonomy
- [ ] tools/gemini
- [ ] tools/taxonomy_krona_chart
- [ ] tools/tool_factory_2
- [ ] tools/transtermhp

tool_factory_2 just needs ghostscript or graphicsmagick, that should be doable.
differential_count_models is kind of nice because of the comparison between deseq2/edgeR/limma and the nice plots it produces by default. But all these packages are outdated, and I think it is effectively already unmaintained.

The other tools I haven't used, so I'm not sure about their status.

peterjc:
I also agree in principle, but think we need to continue to support the legacy packaging for a little longer - while as you say pushing people towards the bioconda infrastructure.

Separately I've emailed the IUC list about strategies for ensuring things are available on both during this transition period. Is it acceptable/sensible to add old versions of packages already on bioconda?

mvdbeek:
> Is it acceptable/sensible to add old versions of packages already on bioconda?

I think so. IIRC this was the reason we didn't invest more energy in homebrew.

peterjc:
[...]

PR #1266, opened 2017-04-20
Condafy gff3_rebase

abretaud:
This one was simple!
see #1262


tools/gff3_rebase/macros.xml
```diff
@@ -2,18 +2,8 @@
 <macros>
   <xml name="requirements">
     <requirements>
-      <requirement type="package" version="2.7">python</requirement>
-      <requirement type="package" version="1.65">biopython</requirement>
```
peterjc: Why are you dropping the Biopython dependency? The python script explicit uses it via ``from Bio.SeqFeature import FeatureLocation``

Are you assuming the bcbiogff dependency will pull in an appropriate version of Biopython? I don't know what is considered IUC best practice here.

bgruening: @peterjc bcbiogff has a dependency on biopython https://github.com/bioconda/bioconda-recipes/blob/master/recipes/bcbiogff/meta.yaml#L20
So this is not needed anymore, assuming that bcbiogff pulls in all dependencies on it's own.

bgruening:
Maybe we can cite biopython here?
Tests are running and passing! Awesome: https://travis-ci.org/galaxyproject/tools-iuc/jobs/223834501#L881

Thanks @abretaud!

tools/gff3_rebase/macros.xml
peterjc: Is this documented, perhaps as a best practise? I'm not sure this is the best approach. In this case the precise version of Biopython is unlikely to matter -but  it looks like bcbiogff recipe will pull in the latest which may not always work. Would it not be safer to be explicit about which version of Biopython this was written and tested with?

mvdbeek: If that happens, the bioconda package build for bcbiogff should fail. The idea in general is that we only ask for the top-level package, assuming that the top-level package is not broken. Similarly, if we need a r-library, we just ask for that library, not for r.

abretaud: i'm not sure this point is documented, but in this case bcbio-gff itself isnot usable without biopython, and I feel that it we need to pin the biopython version at some point, it should be done in the bioconda recipe rather than here

peterjc: OK, understood. Thanks for clarifying.

mvdbeek:
A version bump would be good, since this tool won't work anymore with toolshed packages, and revision updates (given you had a previosu toolshed version of this tool installed) won't trigger a conda install.

bgruening:
Great, thanks all!

PR #1269, opened 2017-04-20
Fix compressed input into FastQC

dpryan79:
Xref: #1128 

Compressed input into the current FastQC wrapper only works if the history item's name ends in `.gz`. That will generally work for uploaded or linked in files (assuming people don't change the names) but won't if you run FastQC on trimmed data. I've now modified the python script to pay closer attention to whether the input datatype is `fastq.gz` (or `fastq.bz2`). This fixes things locally for me at least.

### DIFF
--- a/tools/fastqc/rgFastQC.py
+++ b/tools/fastqc/rgFastQC.py
@@ -52,19 +52,19 @@ class FastQCRunner(object):
         trimext = False
         # decompression at upload currently does NOT remove this now bogus ending - fastqc will barf
         # patched may 29 2013 until this is fixed properly
-        type = mimetypes.guess_type(self.opts.input)
-        if linf.endswith('.gz') or linf.endswith('.gzip') or type[-1] == "gzip" or informat.endswith('.gz'):
+        ftype = mimetypes.guess_type(self.opts.input)
+        if linf.endswith('.gz') or linf.endswith('.gzip') or ftype[-1] == "gzip" or informat.endswith('.gz'):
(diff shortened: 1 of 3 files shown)


tools/fastqc/rgFastQC.py
```diff
@@ -48,20 +48,23 @@ def prepare_command_line(self):
...
         trimext = False
         # decompression at upload currently does NOT remove this now bogus ending - fastqc will barf
         # patched may 29 2013 until this is fixed properly
         type = mimetypes.guess_type(self.opts.input)
-        if linf.endswith('.gz') or linf.endswith('.gzip') or type[-1] == "gzip":
+        if linf.endswith('.gz') or linf.endswith('.gzip') or type[-1] == "gzip" or informat.endswith('.gz'):
             f = gzip.open(self.opts.input)
             try:
                 f.readline()
+                type = ['gzip']
```
bgruening: We should use `ftype` or something similar as `type` is a reserved keyword on python.
https://docs.python.org/2/library/functions.html#type

dpryan79: Fine by me, I was just trying to be consistent with the previous code :)

bgruening: I know, thanks for fixing! Hope to see you today - I guess I own you a juice for all this work!

dpryan79: Or I owe you one for merging my PRs :)

PR #1270, opened 2017-04-20
Make.contigs - Combo mode

arbernard:
This Pull Request concern an update of the make.contigs wrapper (as discussed previously with @shiltemann)

The wrapper has been modified so that the end user can use make.contigs in combo mode (i.e. make it process more than one pair of fastq files at a time).

The end user can now select a list:paired collection as input.

If such a collection is selected, the wrapper will create a valid combo file "on the fly" and use it in the make.contigs command line through the "file" option.

My goal here was to avoid forcing the end user to run many make.contigs in batch mode and then use several tools (merge.files, make.group) to merge all the result files (since make.contigs is already able to do it by itself).

The new test added in the <tests> section use md5 checksums to avoid having more test files in the test-data folder.

The wrapper syntax has been checked with "planemo lint".

The correct execution of the tool has been tested with "planemo test" and "planemo serve".

### DIFF
--- a/tools/mothur/make.contigs.xml
+++ b/tools/mothur/make.contigs.xml
@@ -9,14 +9,19 @@
     <command><![CDATA[
         @SHELL_OPTIONS@
 
-        ## create symlinks to input datasets
-        #if $input_type.type == 'collection'
-            ln -s '$input_type.fastq_pair.forward' ffastq.dat &&
-            ln -s '$input_type.fastq_pair.reverse' rfastq.dat &&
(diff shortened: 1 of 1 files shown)


shiltemann:
I will test this tomorrow, but this looks great @arbernard! thanks for your contribution!

shiltemann:
works like a charm, thanks @arbernard!!

arbernard:
You're welcome ! Glad to be of help :)

PR #1287, opened 2017-05-02
Mothur suite reorganisation

shiltemann:
the mothur tools had two problems:

1. tests were not finishing within travis' maximum build time
2. installing the full suite from the toolshed gives an internal server error (before even hitting install, just to view the repo)

To fix this I've:

1. split the tools into 3 subdirectories, now each gets assigned to a different chunk and run in parallel by travis and all tests can now complete in time (and I'm using `include` in shed.yml so they all still share things like test data and macros)

2. made each of those 3 subdirectories a separate "regular" multi-tool repo as well. Ideally this shouldn't be necessary but maybe it's an acceptable workaround until the full suite install works from toolshed again? (it's really no fun hitting install on 130 tools individually if you want all of them)

what to you think? this ok? better ways to do it? (ping @bgruening) 

P.S. shed_linting is failing on the urls, but they seem to be fine and locally shed_lint passes so not entirely sure what is going on there, but scrolling down on chunks1-3 shows that all tests passed, (chunk0 is trying to run all tests because this last commit changed something in the superdirectory)


tools/mothur/mothur1/.shed.yml
```diff
@@ -0,0 +1,23 @@
...
+homepage_url: http://www.mothur.org
+type: unrestricted
+include:
+- strip_components: 1
+  source:
+  - ../macros.xml
+  - ../test-data/
+  - ../tool-data/
+  - ../tool_data_table_conf.xml.sample
+  - ../tool_dependencies.xml
```
bgruening: Is the tool_dependencies.xml needed?

bgruening:
To be honest, I see no harm in 1). If we (let's say) get two IUC persons to run the tests locally and confirm that they pass, I would be happy to merge them. I can even ask for longer build time if you like. We do have more than 45min right? Ideally we should enhance planemo to split repositories with multiple tools into multiple travis jobs - this would profit many tools. I'm rally scared about Qiime as well ;)

2) is a bummer and I'm not sure what to do here. Something in me says this should be fixed in Galaxy and we should create good tools and don't over complicated things on the tool level. But I also know that it is a pain to install them. Would it be ok for you and our users to include a readme file (Oh I see this will not be displayed even, right?) that gives hints how to install it via the API?

If we agree to split the tools into sub-repos, can we split them into logical groups? Clustering, filter, get?

Sorry, to be not super helpful :(

tools/mothur/mothur1/.shed.yml
bgruening: I'm wondering if the tool dependencies also make loading the TS view more complicated.

shiltemann:
thanks @bgruening 

yeah, neither feels really great to me either, thanks for sharing your views

more time would be great, we have 90 minutes now I think, but locally it takes over 3,5 hours to run all the tests, not sure how easy it is to get more time? adjusting planemo to split a single directory into multiple chunks if it is a large directory would also solve it, not sure how easy this is to do? ..on the other hand, just this change without item 2 does not affect the resulting repos in toolshed (still one repo per tool and the repository suite) but I guess it complicates things a bit here in the github repo..  
but we can undo that if/when the toolshed issue is fixed down the road..

yep, installing via API is an option, but I fear this is just makes it unnecesarily complicated, especially for less experienced admins.. as soon as the toolshed install of the full suite works again we can change things back and mark those extra repos as deprecated of course..

yep, I thought about a better splitting of tools into groups but since it has to be at least 3 groups I could not think of anything obvious that would still have each of the groups a good size to finish in time (but suggestions welcome) ..if we decide that is the way we want to go I can give it some more thought though

yeah, I had removed the tool_dependencies.xml at first, then remembered a group using the mothur tools that were not able to switch to conda yet on their system (but happy to remove this again if you think enough time has passed that we can go conda-only)

mostly I would just like to see the mothur tools get into the main toolshed, the mothur-fixes branch has some nice improvements. If I add the one tool fix from this PR there we could also merge that first and think about this a bit longer if you prefer?

bgruening:
> more time would be great, we have 90 minutes now, but locally it takes over 3,5 hours to run all the tests, not sure how easy it is to get more time? adjusting planemo to split a single directory into multiple chunks if it is large would also solve it, not sure how easy this is to do?

I think this is fairly easy. There is a function in planemo that returns a list of all tools that needs to be tested. This is currently a list of directories. If this function returns a list of single XML files all X (currently 4) travis jobs will work on this in parallel.

> yeah, I had removed the tool_dependencies.xml at first, then remembered a group using the mothur tools that were not able to switch to conda yet on their system (but happy to remove this again if you think enough time has passed that we can go conda-only)

This is at least the current policy, we have removed a lot of them during the last month.

I also want to see mothur in MTS as soon as possible! :)

ping @galaxyproject/iuc

shiltemann:
ok great, then why don't I close this, as it sounds like the travis testing issue will get resolved at some point in the future anyway (and with bit of luck same will be true for the toolshed issue). 

As for the suite install issue, I have added instructions for installation via the API in the mothur-fixes branch, which we could point people to (as you say they won't be able to see this in the toolshed as it fails before then already, but they will at least be able to see it in the individual tool repos and here on github)

bgruening:
Let's wait for others to chime in. I'm not sure how long all this will take.

shiltemann:
I will close this for now, the travis changes from the qiime branch will fix point 1, and we can always revisit point 2 if users strongly prefer it over API solution.

PR #1289, opened 2017-05-04
Freebayes regionfix

bwlang:
old method always examined all entries in bam file, 
this method respect the user's entry to target or region fields

### DIFF
--- a/tools/freebayes/freebayes.xml
+++ b/tools/freebayes/freebayes.xml
@@ -40,7 +40,7 @@
     #if str( $target_limit_type.target_limit_type_selector ) == "limit_by_target_file":
         cp '${target_limit_type.input_target_bed}' regions_all.bed &&
     #elif str( $target_limit_type.target_limit_type_selector ) == "limit_by_region":
-        echo '${target_limit_type.region_chromosome}:${target_limit_type.region_start}..${target_limit_type.region_end}' > regions_all.bed &&
+        echo -e '${target_limit_type.region_chromosome}\t${target_limit_type.region_start}\t${target_limit_type.region_end}' > regions_all.bed &&
     #else
         ##divide up the regions in the bam file for efficient processing
         #for $bam_count, $input_bam in enumerate( $reference_source.input_bams ):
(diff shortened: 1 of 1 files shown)


bwlang:
@bgruening  this is that freebayes issue we chatted about earlier today

tools/freebayes/freebayes.xml
```diff
@@ -35,14 +35,25 @@
...
-        grep "^@SQ" |
-        cut -f 2- |
-        awk '{ gsub("^SN:","",$1);
-        gsub("^LN:","",$2);
-        print $1"\t0\t"$2; }' >> regions_all.bed &&
-    #end for
+    ##if user has specified a region or target file, just use instead of calculating a set of unique regions
+
+    #if str( $target_limit_type.target_limit_type_selector ) == "limit_by_target_file":
+        cp '${target_limit_type.input_target_bed}' regions_all.bed &&
```
bgruening: Symlink here?

bgruening:
Thanks @bwlang. LGTM. I just added one small comment.

tools/freebayes/freebayes.xml
```diff
@@ -35,14 +35,25 @@
...
+    #elif str( $target_limit_type.target_limit_type_selector ) == "limit_by_region":
+        echo -e '${target_limit_type.region_chromosome}\t${target_limit_type.region_start}\t${target_limit_type.region_end}' > regions_all.bed &&
```
yhoogstrate: `-e` is not POSIX (http://pubs.opengroup.org/onlinepubs/009695399/utilities/echo.html), how do you thing about using `printf` over `echo -e`?

tools/freebayes/freebayes.xml
```diff
@@ -35,14 +35,25 @@
...
+    #else
+        ##divide up the regions in the bam file for efficient processing
+        #for $bam_count, $input_bam in enumerate( $reference_source.input_bams ):
+            samtools view -H b_${bam_count}.bam |
```
yhoogstrate: Do you think this is simpler and equivalent: `samtools idxstats b_${bam_count}.bam | grep -v '^*' | cut -f 1,2 | sed 's@\t@\t0\t@'` ?

bwlang: certainly simpler code... but it's a more invasive change and I'm not certain we can count on these bams to be indexed 100% of the time.  I'm leaning toward leaving it as is unless there is a strong opinion against that.

yhoogstrate:
Awesome, thanks @bwlang @bgruening and @nsoranzo :+1:

PR #1290, opened 2017-05-04
Fix for running headless krona

davebx:
Main produces:

tput: No value for $TERM and no -T specified
tput: No value for $TERM and no -T specified

### DIFF
--- a/tools/taxonomy_krona_chart/taxonomy_krona_chart.xml
+++ b/tools/taxonomy_krona_chart/taxonomy_krona_chart.xml
@@ -1,10 +1,11 @@
-<tool id="taxonomy_krona_chart" name="Krona pie chart" version="2.6.1">
+<tool id="taxonomy_krona_chart" name="Krona pie chart" version="2.6.2">
   <description>from taxonomic profile</description>
   <requirements>
     <requirement type="package" version="2.6.1">krona</requirement>
   </requirements>
   <version_command>ktImportText | sed -n 's/.*\(KronaTools [[:digit:].]\{1,\}\).*/\1/p'</version_command>
   <command><![CDATA[
+    if [ -z "\$TERM" ] ; then export TERM=xterm ; fi &&
     #if $type_of_data.input[0].is_of_type('taxonomy'):
         ktImportGalaxy
         -d '${type_of_data.max_rank}'


tools/taxonomy_krona_chart/taxonomy_krona_chart.xml
```diff
@@ -1,10 +1,11 @@
-<tool id="taxonomy_krona_chart" name="Krona pie chart" version="2.6.1">
+<tool id="taxonomy_krona_chart" name="Krona pie chart" version="2.6.2">
   <description>from taxonomic profile</description>
   <requirements>
     <requirement type="package" version="2.6.1">krona</requirement>
   </requirements>
   <version_command>ktImportText | sed -n 's/.*\(KronaTools [[:digit:].]\{1,\}\).*/\1/p'</version_command>
   <command><![CDATA[
+    if [ -z "\$TERM" ] ; then export TERM=xterm ; fi &&
```
nsoranzo: That's not adding a dependency on `xterm`, right?

bgruening:
Does this needs to be handles by tools? The framework should take care of it IMHO.

tools/taxonomy_krona_chart/taxonomy_krona_chart.xml
davebx: No, it's just lying to krona.

nsoranzo: Way to go!

natefoo:
I am not sure what a good value for `$TERM` would be when there is no controlling terminal. A tool really should not assume a terminal is present.

davebx:
It doesn't seem to require a specific value at all, I tried with `TERM=headless` and `TERM=none` and krona still produced valid output.

natefoo:
My opinions:

Fixing it in conda doesn't seem right either - the conda version would be used on the command line so its use of tput is okay.

"Most correct" solution: Krona shouldn't assume a terminal, and instead should not call `tput`, using a default value if `$TERM` is unset.

"Second best" solution: Work around it in the wrapper by setting `$TERM` a la this PR.

I don't like working around this using an `<env>` tag in `job_conf.xml` because that would affect other tools. Tools should not be assuming they have a controlling terminal, that breaks the UNIX paradigm.

bgruening:
With @natefoo blessing I'm ok with this as well :)
@davebx can you add a comment!

davebx:
I should instead use exit code detection, because that message does not result in a non-zero exit code.

PR #1293, opened 2017-05-05
Better fix for running headless krona

davebx:
The `tput: No value for $TERM and no -T specified` message is sent to stderr, but the exit code remains 0.

### DIFF
--- a/tools/taxonomy_krona_chart/taxonomy_krona_chart.xml
+++ b/tools/taxonomy_krona_chart/taxonomy_krona_chart.xml
@@ -3,11 +3,8 @@
   <requirements>
     <requirement type="package" version="2.6.1">krona</requirement>
   </requirements>
-  <stdio>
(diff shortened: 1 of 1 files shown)


nsoranzo:
Now the recommended way to do that is adding the attribute `detect_errors="exit_code"` to `<command>`, but this should work too.

davebx:
Interestingly, this line is returning exit code 1: `&& [ -d '${output}.files' ] && ( mkdir '${output.extra_files_path}' && mv '${output}.files' '${output.extra_files_path}/' )`

nsoranzo:
Maybe now krona put its files in some other directory instead of `${output}.files`?

Also I think `(` and `)` can be removed there.

davebx:
Turns out krona only generates `dataset.files/` under certain conditions. Updating the pull request in a moment.

nsoranzo:
Thanks @davebx!

PR #1294, opened 2017-05-05
change owner of bam_to_cram and cram_to_cam

martenson:
these were never uploaded to TTS or MTS so we can change safely


nsoranzo:
Thanks @martenson!

PR #1296, opened 2017-05-06
QIIME - Add data manager

bebatut:
Hi,

I added a data manager to download the databases for QIIME.
I tested it locally on an empty Docker image, but each test is quite long. So sometimes it failed probably because of internet connection but I got also `This job failed for reasons that could not be determined.`
If someone could test it and help me, it would be really great!

Thanks,

Bérénice


bgruening:
TTS bebatut account? Is this the latest version?

bebatut:
Yes the lastest version on TTS is the `data_manager_qiime_database_downloader` with `bebatut` as author

bgruening:
Works like a charm! Great work @bebatut!

bgruening:
Can you rebase this one against master. I think we can merge it already.

PR #1297, opened 2017-05-08
Update and fixes to sra-tools

nekrut:
- if accession contained leading and training spaces - tool failed. this is now fixed
- upgraded to 2.8.1 version of sra-tools
- completely rewrote user help. It was not accessible to an average (like myself) user. Now it clearly states what the tool does
- did not do much with pileup tool. I think is has limited utility for most users.


tools/sra-tools/fastq_dump.xml
```diff
@@ -9,14 +9,21 @@
...
-      do
+    
+        for acc in `cat $input.file_list` ;
+        do
+    
     #elif $input.input_select=="accession_number":
-      acc="$input.accession" &&
+
+        ## Stripping leading and trailing spaces in case user typed them in 
+        acc="${ $input.accession.strip() }" &&
```
mvdbeek: We should probably just disallow these in the `accession` parameter using:
```
<sanitizer>
        <remove value=" "/>
    </valid>
</sanitizer>
```

tools/sra-tools/fastq_dump.xml
```diff
@@ -160,7 +160,7 @@
         <discover_datasets pattern="(?P&lt;designation&gt;.+)_\d+.fastq__single\.fastq" directory="." ext='fastqsanger'/>
         <discover_datasets pattern="(?P&lt;designation&gt;.+)_\d+.fasta__single\.fasta" directory="." ext='fasta'/>
       </collection>
-      <data format="fastqsanger" name="output_accession" >
+      <data format="fastqsanger" name="output_accession" label="${$input.accession.strip()}.${outputformat}">
```
mvdbeek: Does this really work? The `${$` looks suspicious to me. In any case I'd prefer to add a sanitizer to $input.accession over using cheetah in the label. Also I would say adding the output format to the name doesn't really help, we got that in the datatype. I'd prefer a label like `fastq-dump on accession $input.accession`.

tools/sra-tools/sra_macros.xml
```diff
@@ -24,15 +24,14 @@
...
-            <label>aligned or unaligned reads</label>
+        <param name="alignments" type="select" value="both" label="Output aligned or unaligned reads" help="Output reads according to their alignment status. --aligned and --unaligned">
             <option value="both">both</option>
             <option value="aligned">aligned only</option>
             <option value="unaligned">unaligned only</option>
         </param>
     </macro>
     <macro name="minMapq">
-        <param name="minMapq" type="integer" min="0" max="42" label="minimum mapping quality" optional="true"/>
+        <param name="minMapq" type="integer" min="0" max="42" label="Minimum mapping quality" optional="true" help="minimum mapping quality an alignment has to have, to be dumped. --min-mapq"/>
```
mvdbeek: The standard way to add `--min-mapq` at the end of the help is to use `argument="min-mapq"`. Capitalize `minimum`.

tools/sra-tools/fastq_dump.xml
```diff
@@ -83,29 +90,25 @@
     #elif $input.input_select=="file_list":
         "\$acc"
     #else:
-         --stdout
-        "\$acc" > "$output_accession" )
+        --stdout
```
mvdbeek: I think you can remove the
```
     #elif $input.input_select=="file_list":
          "\$acc"
```
and change the `#else` into `#elif $input.input_select=="accession":`

tools/sra-tools/fastq_dump.xml
```diff
@@ -115,43 +118,40 @@
...
-            <param name="minID" type="integer" label="minimum spot ID" optional="true"/>
-            <param name="maxID" type="integer" label="maximum spot ID" optional="true"/>
-            <param name="minlen" type="integer" label="minimum read length" optional="true"/>
-            <param name="split" type="boolean" checked="true" truevalue="--split-spot" falsevalue="">
-                <label>split spot by read pairs</label>
-            </param>
+            <param name="minID" type="integer" label="Minimum spot ID" optional="true" help="Minimum spot id to be dumped. --minSpotId"/>
+            <param name="maxID" type="integer" label="Maximum spot ID" optional="true" help="Maximum spot id to be dumped. --maxSpotId"/>
+            <param name="minlen" type="integer" label="Minimum read length" optional="true" help="Filter by sequence length. Will dump only reads longer or equal to this value. --minReadLen"/>
+            <param name="split" type="boolean" checked="true" truevalue="--split-spot" falsevalue="" label="split spot by read pairs" help="Split spots into individual reads. --split-spot"/>
```
mvdbeek: Capitalize `split` in the label.

tools/sra-tools/sra_macros.xml
nekrut: Tx. I'll fix all of these shortly

tools/sra-tools/fastq_dump.xml
nekrut: This appears to fill spaces with `X` symbols ;(

nekrut: Should be:

```
<sanitizer>
                            <remove value=" "/>
                        </valid>
                        <mapping initial="none">
                            <add source=" " target=""/>
                        </mapping>
                    </sanitizer>
```

tools/sra-tools/fastq_dump.xml
nekrut: `#elif $input.input_select=="accession_number":`

done

tools/sra-tools/fastq_dump.xml
nekrut: done

tools/sra-tools/fastq_dump.xml
nekrut: sanitized took care of this.  I did change output labels as well

tools/sra-tools/sra_macros.xml
nekrut: good to know about `argument=` I had no idea

nekrut:
don't merge yet - adding gzip and bzip2 support

nekrut:
Ready to merge

tools/sra-tools/fastq_dump.xml
```diff
@@ -115,129 +119,239 @@
...
-        </change_format>
-      </data>
-      <data format="fastqsanger" name="output_file" label="${input.file.name}.${outputformat}">
-        <filter>input['input_select'] == "file"</filter>
-        <change_format>
-          <when input="outputformat" value="fasta" format="fasta"/>
-        </change_format>
-      </data>
+        
+            <discover_datasets pattern="(?P&lt;identifier_0&gt;[^_]+)_\d+.fastq_(?P&lt;identifier_1&gt;[^_]+)\.fastqsanger" ext="fastqsanger" visible="false" />
```
mvdbeek: `visible="false"` should not be doing anything here, in a collection context the underlying HDAs should be automatically hidden. If that's not the case it should be fixed in galaxy IMO.

nekrut: indeed. fixed

mvdbeek:
Thanks a lot @nekrut, this is great. I think we should at some point also drop support for sra files, I don't think people are really using them, and it would streamline the logic a bit more.

nekrut:
same goes for generating pileup, I can hardly imagine this being useful

PR #1302, opened 2017-05-11
Trinity analyze_diff_expr: fix input file names

loraine-gueguen:
This PR fixes several points: 
- Ask for DE_matrices as input only when using option "Run GO enrichment analysis".
- Improve handling of input file names for DE_results and DE_matrix files. Adjust file names in tests consequently. Add test for DE_results input files with non-standard names. (See below for more explanation)
- Keep '.subset' in the name generated for extracted_DE_genes output files.
- Remove unnecessary bioconductor-go.db requirement. This conda package is already installed with bioconductor-goseq.

The wrapper does not handle correctly DE_results files generated by previous Trinity tool (run_de_analysis.pl) and produces an error when using option "Run GO enrichment analysis" (unless removing the standard terminal ".DE_results" for each file). Indeed, the wrapper adds a ".DE_results" extension systematically to input DE_results file, even if it is already there and, consequently, there is no more link through file name between DE_results files and DE_matrices files, which is essential for the tool. 

I hope I am clear enough!


bgruening:
@loraine-gueguen what happens if a file is called `foo_bar_foo.DE_results.DE_results` is it still working?

abretaud:
Appart from @bgruening remark, it looks ok to me

loraine-gueguen:
Yes. It will work. What is important  is to have a terminal '.DE_results' (which is ensured by the for loop on DE_results files in the command line) because the tool analyze_diff_expr.pl is looking for these files this way:
`my @DE_result_files = <*.DE_results>;` 
No matter if the files have 2 or 3 times '.DE_results'.

The difficulty is when using option 'go_enrichment', which needs supplementary input files 'DE_matrices'. The file names of 'DE_matrices' files must be exactly identical to the file names of differential expression results, with extension '.count_matrix' instead of '.DE_results'. The tool analyze_diff_expr.pl is looking for the 'count_matrices' files this way:
`$background_file =~ s/\.DE_results$/\.count_matrix/`
In the wrapper, in the for loop on 'DE_matrices' files, I addressed the general case where all the input files are generated by the run_de_analysis.pl tool of Trinity. And I have added a warning on file names in the 'help' section of  input 'DE_matrices' files.

bgruening:
Wah, this is crazy :)! Thanks @loraine-gueguen for the fix!

PR #1306, opened 2017-05-15
QIIME - Add data manager (from master)

bebatut:
To replace #1296


bgruening:
Fantastic, it works for me, at least 4 out of the 1000 different options :)
Thanks B!

PR #1307, opened 2017-05-15
Update prinseq.xml

lecorguille:
To conserve the `fastqsanger` format through the tool


nsoranzo:
It would be nice to also add `ftype` at line 694 to check that the correct datatype is assigned to the output dataset.

lecorguille:
Like that?

nsoranzo:
Exactly, thanks @lecorguille!

PR #1311, opened 2017-05-18
activate pep8 testing again

bgruening:
Qiime and mothur are merged. Now we can cleanup python files if needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,20 +39,23 @@ install:
   - | 
     if [ -s changed_repositories_chunk.list ]; then  
         planemo ci_find_tools --exclude data_managers \
-                            --exclude packages \
-                            --chunk_count 4 --chunk "${CHUNK}" \
-                            --output changed_tools_chunk.list \
-                            `cat changed_repositories_chunk.list`
+                              --exclude packages \
+                              --chunk_count 4 --chunk "${CHUNK}" \
+                              --output changed_tools_chunk.list \
+                              $(cat changed_repositories_chunk.list)
+        cat changed_tools_chunk.list
     fi
   - cat changed_repositories_chunk.list
-  - cat changed_tools_chunk.list | true  # this file might not exist
-
+  
 script:
   - set -e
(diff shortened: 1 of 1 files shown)


peterjc:
Everything broke - do you need to include another ``cd``command?

bgruening:
Mh, `cat changed_repositories_chunk.list` is empty, which results in testing all tools. We should guard this beind a IF. If no one beats me to it, I will do it tomorrow. Betime here!

nsoranzo:
It's this line in the modified `.travis.yml` :

https://github.com/galaxyproject/tools-iuc/blob/master/.travis.yml#L44

When `changed_repositories_chunk.list` is empty, all tools in the repo are tested, but no Conda dependency is installed.

nsoranzo:
@bgruening Indeed! Anyway 

> The command "cd "$TRAVIS_BUILD_DIR" && flake8 --exclude=.git ." exited with 0.

bgruening:
Master is also failing, because TS upload takes so long ;)
Really time for bed! :)
I will upload them tomorrow manually.

.travis.yml
```diff
@@ -39,20 +39,23 @@ install:
...
         planemo ci_find_tools --exclude data_managers \
-                            --exclude packages \
-                            --chunk_count 4 --chunk "${CHUNK}" \
-                            --output changed_tools_chunk.list \
-                            `cat changed_repositories_chunk.list`
+                              --exclude packages \
+                              --chunk_count 4 --chunk "${CHUNK}" \
+                              --output changed_tools_chunk.list \
+                              $(cat changed_repositories_chunk.list)
+        cat changed_tools_chunk.list
```
bgruening: I kind of likes the display in travis, that you have a separate item which you can open and see, this is now mixed/hidden in an other idem. I don't feel strongly about it.

nsoranzo: The previous version was wrong because you were using a pipe `|` instead of an or `||`, so I thought it would be cleaner to have the `cat` inside the `if`, so we also don't have the "File not found" error message.
I'll merge as it is, we can move it out if it turns out to be useful.

PR #1313, opened 2017-05-19
QIIME: Add dependency to the data manager

bebatut:
Hi,

I added the `repository_dependencies.xml` and the dependency to the data manager

Bérénice


bgruening:
I like this, but I know @blankenberg had some arguments against it.
Dan what do you think?

blankenberg:
IMHO, it makes sense to have the suite contain the datamanager, but not have a requirement of the individual tools on the data manager. The general thought is that you if you want to install a single tool, but not install the DM, then you should be able to do that easily - you may have the reference data configured manually, for example.

bgruening:
@blankenberg that makes indeed more sense. The problem is planemo creates the suite and we have currently no way to include some additional files. Maybe we can add to the shed.yml some hints about data-managers?

bgruening:
@jmchilton can we annotate additional repo in the shed.yml file that are automatically included during the de-multiplex step? Any other idea?

PR #1314, opened 2017-05-19
QIIME: re-enable alpha diversity test

shiltemann:
this test can be re-enabled again now that we have the extended log output timeout limit in travis.


bgruening:
Great thanks!

PR #1317, opened 2017-05-19
bcftools concat - add naive option

jj-umn:
naive option needed for merging vcf files generated per chromosome:
HISAT2 -> bamtools split -reference => bcftools mpileup => bcftools
call => bcftools concat (naive) ->

### DIFF
--- a/tools/bcftools/bcftools_concat.xml
+++ b/tools/bcftools/bcftools_concat.xml
@@ -43,7 +43,7 @@ ${section.compact_PS}
 ## Primary Input/Outputs
 @INPUT_FILES@ 
 #if $sec_default.mode.naive == "yes" and $output_type == 'v':
-| bcftools view -O v -o '$output_file' -.gz
+> output.gz && bcftools index output.gz && bcftools view -O v -o '$output_file' output.gz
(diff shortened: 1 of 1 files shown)


tools/bcftools/bcftools_concat.xml
```diff
@@ -47,31 +55,43 @@ ${section.compact_PS}
...
-                        ]]></help>
-                        <option value="snps">snps - SNP records</option>
-                        <option value="indels">indels - indel records</option>
-                        <option value="both">both - both SNP and indel records</option>
-                        <option value="all">all - records</option>
-                        <option value="none">none - output multiple records instead</option>
-                    </param>
+                <when value="yes"/>
+                <when value="no">
+                    <conditional name="overlaps">
```
bgruening: I guess we can use sections for these kind of conditionals.

jj-umn: @bgruening   I used conditionals since bcftools exits with error if options are not in correct combinations.

jj-umn:
@bgruening Test with naive option passed.  Anything else I should consider?

PR #1326, opened 2017-05-24
add limma_voom tool

mblue9:
Modified the voom_rnaseq tool from the main toolshed:
- Changed to use bioconda packages for limma, edgeR, statmod and scales
- Made the statmod and scales packages requirements to ensure they're installed to avoid errors
- Added in a test to check the tsv output
- Added a version_command block to report the versions of R, limma and edgeR used 
- Changed the tool description to be shorter as it was quite long in the tool panel


nsoranzo:
@mblue9 It seems that `tools/limma_voom/.shed.yml` is missing from the pull request. Can you add it? Otherwise tests are not run on the tool.

mblue9:
Sorry @nsoranzo I've added it now, or I think I have, please let me know if that's not the right way to do it!

nsoranzo:
@mblue9 You added it correctly, tests ran and it's already green, good job!
There a couple warnings from `planemo lint` that would be nice to fix:
```
Applying linter inputs... WARNING
.. WARNING: No <when /> block found for select option 'no' inside conditional 'anno'
.. WARNING: No <when /> block found for select option 'no' inside conditional 'weightCond'
.. WARNING: No <when /> block found for select option 'yes' inside conditional 'weightCond'
Applying linter citations... WARNING
.. WARNING: No citations found, consider adding citations to your tool.
```

mblue9:
Thanks @nsoranzo :) I think I've fixed those warnings now and have made a few other small changes.

tools/limma_voom/limma_voom.xml
```diff
@@ -397,4 +383,14 @@ Please report problems or suggestions to: su.s@wehi.edu.au
 .. _limma: http://www.bioconductor.org/packages/release/bioc/html/limma.html
 ]]>
 </help>
+    <citations>
+        <citation type="doi">10.1007/0-387-29362-0_23</citation>
+        <citation type="doi">10.1186/gb-2014-15-2-r29</citation>
+        <citation type="doi">10.1093/nar/gkv412</citation>
```
bgruening: @mblue9 can you only add these references here, which are needed to cite if I use this tool?

mblue9: ok done :)

tools/limma_voom/limma_voom.xml
```diff
@@ -0,0 +1,389 @@
...
+      <option value="none">None (Don't normalise)</option>
+      
+    </param>
+
+    <param name="report" type="boolean" truevalue="True" falsevalue="False" checked="true"
+            label="Output a HTML report?"
+            help="Generate a HTML report that includes mean-variance trend, MDS and MA plots." >
+    </param>
+
+    <param name="rdaOption" type="boolean" truevalue="yes" falsevalue="no" checked="false"
```
shiltemann: this Rdata output is only possible if user chose `yes` for the html report right? it makes a link in the report? Maybe you could put this parameter inside a conditional so that it is only shown if user chose to output the html report.

tools/limma_voom/limma_voom.xml
```diff
@@ -0,0 +1,389 @@
...
+  </outputs>
+
+   <tests>
+        <test>
+            <param name="counts" value="matrix.txt" />
+            <param name="factName" value="Genotype" />
+            <param name="factLevel" value="WT,WT,Mut,Mut,Mut" />
+            <param name="contrast" value="Mut-WT" />
+            <param name="normalisationOption" value="TMM" />
+            <output name="outTab" value="limma-voom_Mut-WT.tsv"  />
```
shiltemann: Maybe we could check the html output here as well?

If the html output is not exactly the same every time, you can also just check for certain words inside the file, something like this:

```
<output name="outFile">
    <assert_contents>
        <has_text text="Limma-voom Analysis Output" />
        <has_text text="Rdata" />
    </assert_contents>
</output>
```

And maybe make two separate tests, one with the Rdata object linked and one without? In the test without Rdata you can check the link is not made with something like `<not_has_text text="Rdata" />`

tools/limma_voom/limma_voom.xml
```diff
@@ -0,0 +1,389 @@
...
+  <![CDATA[
+    Rscript '$__tool_directory__/limma_voom.R' $counts
+              
+              #if $anno.annoOpt=='yes':
+                $geneanno
+              #else:
+                None
+              #end if
+              
+              $outFile
```
shiltemann: Could you put this parameter in single quotes? If somebody is running Galaxy in a path with spaces in it this may cause problems (and any other input/output files or text parameters as well)

tools/limma_voom/limma_voom.xml
shiltemann: and would be nice to have a test with the gene annotations input file as well

mblue9:
Thanks a lot @shiltemann! For both the feedback and for the offer of help! I'm learning how to wrap tools so I'll have a go at making those changes first and get back to you then :)

tools/limma_voom/limma_voom.xml
```diff
@@ -0,0 +1,389 @@
+<tool id="limma_voom" name="limma-voom" version="1.1.1">
+  <description>
```
mblue9: Think I've fixed all these

tools/limma_voom/limma_voom.xml
```diff
@@ -0,0 +1,389 @@
+<tool id="limma_voom" name="limma-voom" version="1.1.1">
+  <description>
+      Perform differential expression analysis using limma-voom
```
mblue9: I've removed that and changed the name to "Differential expression with optional sample weights" just to highlight one reason a user might want to use it in addition to or as an alternative to e.g. deseq2 but am happy to change it if something else would be better

tools/limma_voom/limma_voom.xml
```diff
@@ -0,0 +1,389 @@
...
+  
+  <requirements>
+    <requirement type="package" version="3.16.5">bioconductor-edger</requirement>
+    <requirement type="package" version="3.30.13">bioconductor-limma</requirement>
+    <requirement type="package" version="1.4.29">r-statmod</requirement>
+    <requirement type="package" version="0.4.1">r-scales</requirement>
+  </requirements>
+
+  <stdio>
+    <exit_code range="1:" level="fatal" description="Tool exception" />
```
mblue9: I think that's the case so I've changed it to your recommendation

tools/limma_voom/limma_voom.xml
mblue9: I think I've fixed all these but please let me know if not

tools/limma_voom/limma_voom.xml
```diff
@@ -0,0 +1,389 @@
...
+      </param>
+      
+      <when value="yes">
+        <param name="geneanno" type="data" format="tabular"
+               label="Gene Annotations"/>
+        </when>
+        <when value="no" />
+    </conditional>
+
+    <!--*Code commented until solution for multiple factors is found*
```
mblue9: That's not my comment but multiple factors should be possible and I was going to add it in the next version. I was thinking it might be better to input that factor/group info from a file rather than having people type it in, or is there a recommended way to do that?

tools/limma_voom/limma_voom.xml
mblue9: For the moment I've changed it back to the way the tool authors had it originally - html report is output by default- and so I've made just the rdata output an option. I could make the html report an option though if that's recommended?

tools/limma_voom/limma_voom.xml
```diff
@@ -0,0 +1,389 @@
...
+        
+        <when value="no"/>
+        
+    </conditional>
+
+
+  </inputs>
+  
+  <outputs>
+    <data format="html" name="outFile" label="Limma-voom Report on ${on_string}" />
```
mblue9: See comment above, html report is output by default at the moment

tools/limma_voom/limma_voom.xml
mblue9: Hi @shiltemann I've added in these tests - for Rdata and html output and for gene annotations. 

The tool can handle multiple contrasts so I also included that in a test. I've got it to output the results tsv for each contrast into the history but how do I stop this empty file  being created - see screenshot(history item called"limma-voom on data 1")? 
lv_screenshot

Or should I output these tsvs as a collection?

shiltemann: yes, a collection would be better, that would also allow them to be used in a workflow (with this discovery method Galaxy doesn't know beforehand what outputs will be created so they can't be connected to other tools in the workflow editor)

tools/limma_voom/limma_voom.xml
shiltemann: sure that is fine too :) 
(unless the reports can get very big and are not always interesting for the users orso, then making it optional might be good to save their disk quota, but otherwise this is fine)

tools/limma_voom/limma_voom.xml
shiltemann: I don't know what this factor/group info typically looks like? but if it can be short (like a single line) it would be a nuisance to be forced to make a file for this, but if this is typically a bit larger, a file input would be useful. If it can be either maybe we should offer both options? (e.g. via a conditional) what do you think?

nekrut:
are we ready to merge?

shiltemann:
@nekrut I think the only update still needed here is to output a collection in the case of multiple contrasts rather than using the `discover_datasets` approach (right @mblue9?) (again, let me know if you are too busy for this, I am happy to help as well)

mblue9:
Thanks @shiltemann :) I've updated it to output a collection now and will add multiple factors to the next version, hopefully next week.

shiltemann:
thanks a lot @mblue9 !! :tada:

bgruening:
Cool - thanks @mblue9 and @shiltemann!!

mblue9:
Great! :smile:

PR #1328, opened 2017-05-25
Upgrade and update FreeBayes

davebx:
- Use version 1.1.0
- Remove multiple from bam input. Addresses galaxyproject/tools-devteam#429


mvdbeek:
So here we have a legitimate usecase for multiple="True", which is a reduction. This does break collection input if collection elements are not supposed to be merged. Whether or not the alignments should be reduced should be in a conditional, IMO (sth like "Merge multiple bam files?" ).

tools/freebayes/freebayes.xml
```diff
@@ -283,7 +283,7 @@
                 <option value="history">History</option>
             </param>
             <when value="cached">
-                <param name="input_bams" type="data" format="bam" multiple="True" label="BAM file">
+                <param name="input_bams" type="data" format="bam" label="BAM file">
```
mvdbeek: You'll also have to change https://github.com/galaxyproject/tools-iuc/pull/1328/files#diff-35704a222d51e215b3082ed72846e724R79. But it would be best to wrap the input in a conditional.

davebx:
@mvdbeek fair enough. I'll make that modification.

davebx:
@mvdbeek done, thanks for pointing that out.

bgruening:
Looks great, thanks @davebx! And thanks for the great review @mvdbeek and @nsoranzo!

nsoranzo:
Crap, just tried to install the updated version from ToolShed and noticed that `leftalign.xml` does not use the macro for the requirement, so now the TS repo requires 2 different version of freebayes. Opening a PR in a bit.

nsoranzo:
> Crap, just tried to install the updated version from ToolShed and noticed that leftalign.xml does not use the macro for the requirement, so now the TS repo requires 2 different version of freebayes. Opening a PR in a bit.

Will be fixed by #1344.

PR #1331, opened 2017-05-30
Fixed infinite loop error in data_manager_gatk_picard_index_builder

rhpvorderman:
The sorted_names list was extended ad infinitum. Not only did it append "_random" to all items in the list, but it kept adding it, creating an infinite list of x_random_random_random(etc.) items.
This would eat all memory until the Linux kernel would kill the process, resulting in a failure to run the data manager.

The fix makes sure the extension is evaluated after it is complete, not during completion.


bgruening:
@rhpvorderman thanks for tracking this down and providing a few. Very much appreciated!

nsoranzo:
Funnily, in https://github.com/galaxyproject/tools-iuc/commit/9614122e1a6afa084d81a5ab45fbe5c1985e086f#diff-3be93a53ca7530a51dd095f49986277fR134 I extended to Python2 what was already a bug in Python3. Thanks for the fix @rhpvorderman!

PR #1332, opened 2017-05-31
add join_files_by_id tool

bimbam23:


bimbam23:
I do not understand the error? Is there a problem with tool-dep? Or conda?

nsoranzo:
@bimbam23 It failed because GitHub was temporarily down and Travis could not download Galaxy, I've restarted the failed job.

tools/join_files_by_id/join_datasets.xml
```diff
@@ -0,0 +1,123 @@
...
+            <output name="out_file1" file="df.txt"/>
+        </test>
+    </tests>
+
+    <help>
+<![CDATA[-----
+
+**What it does**
+
+Joins datasets by indentifer colunm
```
nsoranzo: s/colunm/column./

tools/join_files_by_id/join_datasets.xml
```diff
@@ -0,0 +1,123 @@
...
++--------+--------+---------+
+| name   | counts | percent |
++========+========+=========+
+| gene 1 |   13   |   0.5   |
++--------+--------+---------+
+| gene 4 |    4   |   0.4   |
++--------+--------+---------+
+
+
+with Dataset 3:
```
nsoranzo: s/with/and/

tools/join_files_by_id/.shed.yml
```diff
@@ -0,0 +1,10 @@
+# repository published to https://toolshed.g2.bx.psu.edu/repos/iuc/join_by_id
+owner: iuc
+name: Join datasets by identifer column
+description: will join datasets according by a column with identifer
```
nsoranzo: This description should probably be rewritten.

tools/join_files_by_id/.shed.yml
```diff
@@ -0,0 +1,10 @@
...
+long_description: |
+    This tool can be used to join count tables of different libraries by using an identifer column which could be a sequence or GeneID.    
+categories:
+    - Join, Subtract and Group
```
nsoranzo: These should be separated on 2 different lines.

tools/join_files_by_id/join_datasets.xml
```diff
@@ -0,0 +1,123 @@
+<tool id="joindatasetsids" name="Join datasets by identifer column" version="1.0">
```
nsoranzo: Normally, the directory name, the tool XML file and the tool id should be the same, e.g.: `tools/join_files_by_id/join_files_by_id.xml` and `<tool id="join_files_by_id"...`

Doesn't have to be `join_files_by_id`, but choose one and stick with it.

tools/join_files_by_id/.shed.yml
```diff
@@ -0,0 +1,10 @@
+# repository published to https://toolshed.g2.bx.psu.edu/repos/iuc/join_by_id
+owner: iuc
+name: Join datasets by identifer column
```
nsoranzo: This should be equal to the tool id (see comment below).

tools/join_files_by_id/.shed.yml
bimbam23: do you mean the long sentence?

shiltemann: These should be tool shed categories, not sections in the Galaxy toolbox. You can have more than one here if you like, so see which best fit your tool: https://toolshed.g2.bx.psu.edu/ perhaps `Text manipulation` ?

tools/join_files_by_id/join_files_by_id.xml
```diff
@@ -0,0 +1,119 @@
...
++========+========+=========+========+=========+========+=========+
+| gene 1 |   10   |   0.3   |   13   |   0.5   |   10   |   0.3   |
++--------+--------+---------+--------+---------+--------+---------+
+| gene 2 |   20   |   0.2   |    0   |    0    |    0   |    0    |
++--------+--------+---------+--------+---------+--------+---------+
+| gene 3 |   12   |   0.2   |    0   |    0    |   12   |   0.2   |
++--------+--------+---------+--------+---------+--------+---------+
+| gene 4 |    0   |    0    |    4   |   0.4   |    4   |   0.4   |
++--------+--------+---------+--------+---------+--------+---------+
+| gene 5 |    0   |    0    |    0   |    0    |   19   |   0.6   |
```
shiltemann: so this inserts a `0` if no value is present? would it be better to insert `NA` values (or empty fields) here to distinguish between the case of a real value of 0 or a missing value?

Also, doesn't the tool `join two datasets - side by on a specified field` that comes with Galaxy not do almost the same? what was missing from that tool for you?

tools/join_files_by_id/.shed.yml
```diff
@@ -0,0 +1,12 @@
+# repository published to https://toolshed.g2.bx.psu.edu/repos/iuc/join_files_by_id
+owner: iuc
+name: Join datasets by identifer column
+description: 
+    This tool will join datasets according to a column with identifer
```
nsoranzo: s/identifer/identifier/ (also below and in other files)

tools/join_files_by_id/.shed.yml
```diff
@@ -0,0 +1,12 @@
+# repository published to https://toolshed.g2.bx.psu.edu/repos/iuc/join_files_by_id
+owner: iuc
+name: Join datasets by identifer column
```
nsoranzo: `name: join_files_by_id`

tools/join_files_by_id/join_files_by_id.xml
```diff
@@ -0,0 +1,119 @@
...
+        <param name="input2" type="data" format="tabular" label="Select table" help="Single dataset selection only to preserve the order"/>
+        <repeat name="queries" title="Select table">
+            <param name="input" type="data" format="tabular" label="Select table" help="Single dataset selection only to preserve the order"/>
+        </repeat>
+        <param name="field" label="Column used for joining" type="data_column" data_ref="input1" /> 
+    </inputs>
+    <outputs>
+        <data name="out_file1" format="tabular" />
+    </outputs>
+	<tests>
```
nsoranzo: Use 4 spaces instead of tab

tools/join_files_by_id/join_files_by_id.xml
bimbam23: it can insert 0 or "-" so fare. I could also allow NA. So the idea was to use this tool to join count-tables with identifiers and separate counts per library for each file.

join_two datasets can only join two files and it also does not join in a sense that the joined identifier are joint together. My tool also preserves the order of the different files which is needed for a final count-table. Additional I also need a join all verses all that mean table one two three four can have in all files different identifiers and the final joint table has all identifiers with its counts.

shiltemann: ok, was just wondering whether it was worth just updating that tool to handle more than two files

Yes, maybe allow some more mismatch characters, I can see this tool being used for things other than count tables as well, I would say at least add empty field as an option?

tools/join_files_by_id/join_files_by_id.xml
```diff
@@ -0,0 +1,119 @@
...
+            <option value="da">Dash</option>
+            <option value="sp">Space</option>
+        </param>
+        <param name="null_str" type="select" label="Select mismatch character">
+            <option value="0">0</option>
+            <option value="-">-</option>
+        </param>
+        <param name="input1" type="data" format="tabular" label="Select table" help="Single dataset selection only to preserve the order"/>
+        <param name="input2" type="data" format="tabular" label="Select table" help="Single dataset selection only to preserve the order"/>
+        <repeat name="queries" title="Select table">
```
shiltemann: you could also make this a repeat with a minimum of two (`min="2"`) and remove `input1` and `input2` params above

tools/join_files_by_id/.shed.yml
```diff
@@ -0,0 +1,12 @@
+# repository published to https://toolshed.g2.bx.psu.edu/repos/iuc/join_files_by_id
+owner: iuc
+name: Join datasets by identifier column
```
nsoranzo: Again:
`name: join_files_by_id`
Otherwise you get this `planemo shed_lint` warning:
```
Repository name [Join datasets by identifier column] invalid. Repository names must contain only lower-case letters, numbers and underscore.
```

tools/join_files_by_id/join_files_by_id.xml
bimbam23: I agree! I also thought to just change an existing tool but in the end it was to much changes.

tools/join_files_by_id/join_files_by_id.xml
bimbam23: Good idea, but is it possible to catch the type=data_column for my field variable?
<param name="field" label="Column used for joining" type="data_column" data_ref="input1" />

shiltemann: hmm, good point, I am not sure, if not, you can leave `input1` as it is but remove `input2` and put a minimum of 1 on the repeat

bimbam23: I changed it but how to your write a test for a repeat
I tried: 
            <param name="input1" value="df1.txt"/>
            <repeat name="queries">
                <param name="input" value="df2.txt"/>
            </repeat>
            <repeat name="queries">
                <param name="input" value="df3.txt"/>
            </repeat>

bimbam23: I found the bug

bimbam23:
anything more to change? ;-)

shiltemann:
@bimbam23 did you intend to add those latest commits to this PR? (maybe undo those here and start a new branch for your ggplot changes) 

@nsoranzo did you want any more changes to the `join_files_by_id` tool or do you approve now?

bimbam23:
I think I just used the add-tools branch... join files by id is ready to go.

bimbam23:
I will be on holidays from today... I can fix it 2. Aug.

shiltemann:
no problem, enjoy your vacation!

bimbam23:
Thx, in the end the ggplot2 PCA should be another tool of Mo's ggplot2 wrappers

tools/join_files_by_id/join_files.R
```diff
@@ -0,0 +1,103 @@
+# libraries
+library(data.table)
+library(parallel)
+
+# inputs
+args <- commandArgs()
+# print(args)
```
nsoranzo: @bimbam23 Maybe you can remove from this file a bit of the debugging/commented out code.

bimbam23:
hmm this was not what I wanted to do :-)

bgruening:
I can not reopen it, sorry ;)

nsoranzo:
It says 0 commits.

bimbam23:
It look like I have to PR again

nsoranzo:
Can you delete `tools/join_files_by_id/.Rhistory` ?

bgruening:
ping @nsoranzo

bgruening:
Thanks @bimbam23

PR #1333, opened 2017-05-31
Update ebi tools to use EBISearch

bebatut:
Hi,

Since the first version of the ebi tools, I developed a [python package](https://github.com/bebatut/ebisearch) to interact with EBI Search API.
I update then this wrapper to use it.

Bérénice


bgruening:
@bebatut was is necessary to change the tools IDs? These are now complete now tools and not a new version, is this intended?

bebatut:
It is a mistake on my side. I first developed as new wrappers and after I merge together. I will change that

bgruening:
ping @bebatut

PR #1336, opened 2017-06-01
maSigPro wrapper update

blankclemens:
- Fix bug in stepback.R function
- Avoid problems with input file names
- Update maSigPro package dependency to 1.49.3

The main change is in lines 20 to 67 and line 132 in the new `masigpro.R`. However, since I changed the files from CRLF to LF, all lines appear to be new.


bgruening:
Thanks @blankclemens, works for me!

PR #1341, opened 2017-06-05
Use element identifier when building report

mvdbeek:
That should help building a multiqc report.

Ping @nekrut

### DIFF
--- a/tools/fastqc/rgFastQC.xml
+++ b/tools/fastqc/rgFastQC.xml
@@ -1,22 +1,16 @@
-<tool id="fastqc" name="FastQC" version="0.68">
+<tool id="fastqc" name="FastQC" version="0.69">
     <description>Read Quality reports</description>
     <requirements>
         <requirement type="package" version="0.11.5">fastqc</requirement>
     </requirements>
-    <stdio>
-        <exit_code range="1:" />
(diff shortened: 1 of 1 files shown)


tools/fastqc/rgFastQC.xml
```diff
@@ -1,22 +1,16 @@
...
+        python '$__tool_directory__/rgFastQC.py'
             -i '$input_file'
-            -d '${html_file.files_path}'
+            -d '$html_file.files_path'
             -o '$html_file'
             -t '$text_file'
-            -f '${input_file.ext}'
-            -j '${input_file.name}'
+            -f '$input_file.ext'
+            -j '$input_file.element_identifier'
```
jmchilton: Can you escape this 😇 ?

bgruening: @jmchilton should we add a `.element_identifier_escaped` in Galaxy to make this easier, or escape the default one and offer a `_plain` option?

bgruening:
Thanks @mvdbeek!

tools/fastqc/rgFastQC.xml
jmchilton: Yes I agree... we've talked about that before right?

PR #1346, opened 2017-06-07
transtermhp goes conda

bgruening:


tools/transtermhp/macros.xml
```diff
@@ -3,8 +3,9 @@
   <xml name="requirements">
     <requirements>
       <requirement type="package" version="2.09">transtermhp</requirement>
-      <requirement type="package" version="1.65">biopython</requirement>
-      <requirement type="package" version="0.6.2">bcbiogff</requirement>
+      <requirement type="package" version="1.69">biopython</requirement>
+      <requirement type="package" version="0.6.4">bcbiogff</requirement>
+      <requirement type="package" version="2.7">python</requirement>
```
nsoranzo: `transtermhp.py` is Python3 compatible, if that's why you added this last requirement.

tools/transtermhp/transtermhp.xml
```diff
@@ -1,27 +1,28 @@
...
+    <description>finds rho-independent transcription terminators in bacterial genomes</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements"/>
+    <expand macro="stdio"/>
+    <command><![CDATA[
+        #if $reference_genome.source == 'history':
+            ln -s '$reference_genome.genome_fasta' genomeref.fa &&
+        #end if
```
nsoranzo: This absolutely comes down to personal preferences, but I like to start the content of `<command>` with no indentation (as it was before in this case), as we do with the `<help>` content, which is the other big block of CDATA text we have.
I think this makes it easier to properly indent the Cheetah directives. Worth discussing at https://github.com/galaxy-iuc/standards ?

peterjc: Good idea: https://github.com/galaxy-iuc/standards/issues/43

tools/transtermhp/macros.xml
bgruening: Unfortunately, it is not. `bytes` issues all over.

tools/transtermhp/transtermhp.xml
bgruening: I would love to also intent the help section, but rst does not allow it :)

bgruening: I changes it :)

PR #1356, opened 2017-06-09
Condafy bowtie datamanager

mvdbeek:
bowtie 1.1.2 is available in bioconda. I tested this manually, all is working well.


nsoranzo:
Actually I should have done that as part of https://github.com/nsoranzo/tools-iuc/commit/7e16ddc1e27246c6e75b1ce78003572cd58e04b0 where I've updated the requirement to 1.2.0, thanks @mvdbeek!

PR #1358, opened 2017-06-09
Bed tools gff

ieguinoa:
Add option to overlap BED/VCF/GFF file (input A) with built-in GFF files.


tools/bedtools/closestBed.xml
```diff
@@ -24,13 +23,35 @@
             -k $k
         #end if
         -a '$inputA'
-        -b '$inputBs'
+        #if str($overlap_with.source) == "data_table":
+		-b '$overlap_with.table.fields.path'
```
mvdbeek: Can you keep the 4 space indentation here and below?

bgruening:
@ieguinoa any reason why you haven't added it to the other tools as well?

I'm not sure about local gff files in general. Imho they are to fast out of date and we should teach people where to find them. Downloading them is not complicated so personally I don't see much gain in supporting local gff files in tools.

ping @galaxyproject/iuc

ieguinoa:
Hi @bgruening 

I've pushed the changes in closestBed only because I think the most common use case of this tool is to find which new features are close to something already annotated in a reference. That is, one of the inputs would most certainly be the reference annotation, and not any other user's dataset. 
This may be just my point of view from what it is usually executed at our local instance and so, this is the only tool that I've modified so far. But, anyway, I've pushed this mainly as a first step towards changing several tools, and even changing the tool data tables config to include an all_gff by default....or at least to discuss this approach. 

These are the main reasons why we are already handling local versions of reference gff files in our instances:
- There is no convention on what an reference annotation file should contain and many tools expect specific properties from them: feature names, sorting, attributes (names), etc. So, even though in terms of size these files are easy to download and upload from/to reference sources and Galaxy instances,  I've found them to be one of the hardest file types to handle in terms of formatting and finding a standard that can be used in most (all) the cases. This totally compromises the possibility of using a lot of tools, mainly because we cannot expect the user to parse and edit the gff file to fulfil the requirements of each tool.
- Reproducibility: There is not much to say about this, we try to use as much workflows as possible to standardize the analysis and in   so from the very beginning the solution was to handle the annotation files internally. 
- At least in our field (plant genomics) annotation release cycle is pretty much the same as genome versions. 

In any case, as with any other data table, making use of it and actually having local versions of gff files would be just an option. 

Cheers,
Ignacio

bgruening:
> There is no convention on what an reference annotation file should contain and many tools expect specific properties from them: feature names, sorting, attributes (names), etc. So, even though in terms of size these files are easy to download and upload from/to reference sources and Galaxy instances, I've found them to be one of the hardest file types to handle in terms of formatting and finding a standard that can be used in most (all) the cases. This totally compromises the possibility of using a lot of tools, mainly because we cannot expect the user to parse and edit the gff file to fulfil the requirements of each tool.

But this means every tool should have their own gff data table, isn't it? Otherwise you offer a bunch of gff files in Galaxy, all different tools are using them but then you still have different properties.
Please note this is a general remark for gff data-tables. I do think bedtools are general enough.

I think this is ok we just need to fix the failing test :)

.travis.yml
```diff
@@ -58,6 +58,7 @@ before_install:
   - unset JAVA_HOME
 
 install:
+  - apt-get -y install samtools	  #required to convert test output format
```
nsoranzo: I think this is the preferred syntax, which should work for both `sudo: required` and `sudo: false`:

https://docs.travis-ci.com/user/installing-dependencies/#Adding-APT-Packages

nsoranzo:
It's green!

tools/bedtools/macros.xml
```diff
@@ -2,6 +2,7 @@
     <xml name="requirements">
         <requirements>
             <requirement type="package" version="2.26.0gx">bedtools</requirement>
+            <requirement type="package" version="1.2">samtools</requirement>
```
bgruening: I guess this can then be removed again.

tools/bedtools/bamToBed.xml
```diff
@@ -3,9 +3,7 @@
     <macros>
         <import>macros.xml</import>
     </macros>
-    <expand macro="requirements">
-        <requirement type="package" version="1.2">samtools</requirement>
-    </expand>
+    <expand macro="requirements"/>
```
nsoranzo: You need to revert this change as well.

PR #1360, opened 2017-06-10
Fix nobias variable in goseq tool

mvdbeek:
and remove mention of getgo tool that never materialized.


tools/goseq/goseq.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="goseq" name="goseq" version="0.2.2">
+<tool id="goseq" name="goseq" version="0.2.3">
```
nsoranzo: Since this is just a bug fix, it would be fine to not increase the version number.

bgruening:
Thanks @mvdbeek!

PR #1361, opened 2017-06-12
Fix a wrong database name in HUMAnN2 data manager

bebatut:
Hi,

There were a duplication of a database name in the tool data table generated by the HUMAnN2 data manager.
I fixed it  and I increased the version number.

Bérénice


bgruening:
@bebatut if its a real bug, you don't need to increase the version number.

data_managers/data_manager_humann2_database_downloader/data_manager/data_manager_humann2_download.xml
```diff
@@ -1,7 +1,10 @@
-<tool id="data_manager_humann2_download" name="HUMAnN2 download" version="0.9.9" tool_type="manage_data">
+<tool id="data_manager_humann2_download" name="HUMAnN2 download" version="@VERSION@.1" tool_type="manage_data">
     <description>Download HUMAnN2 database</description>
+    <macros>
+        <import>macros.xml</import>
```
nsoranzo: If the macros are used only by one file, you can just defined them here instead of creating `macros.xml`.

bgruening:
Thanks! :)

PR #1364, opened 2017-06-14
Update limma-voom - add support for multiple factors and normalised c…

mblue9:
- added support for multiple factors and a user can either input the info in the tool form or from a file, also added tests for those
- added an option to output the normalised counts, and added a test for that


tools/limma_voom/limma_voom.xml
```diff
@@ -51,15 +51,25 @@
...
+                #end for
+            #end if
+            
             &&
             mkdir ./output_dir
 
             &&
-            mv '$outReport.files_path'/*.tsv output_dir/
-            
+            cp '$outReport.files_path'/*.tsv output_dir/
```
bgruening: mv should be faster than cp

tools/limma_voom/limma_voom.xml
```diff
@@ -153,19 +178,19 @@
   
     <outputs>
         <data format="html" name="outReport" label="${tool.name} on ${on_string}: Report" />
-        <collection name="voom_results" type="list" label="${tool.name} on ${on_string}: DE genes">
+        <collection name="outTables" type="list" label="${tool.name} on ${on_string}: Tables">
```
bgruening: Can you try to use the metadata elements like here: https://github.com/galaxyproject/tools-iuc/blob/master/tools/deseq2/deseq2.xml#L152

This would tell Galaxy magically which column is what in the UI :)

mblue9: Hi @bgruening thanks for reviewing this, I'm on leave now so I'll do this at GCC as I've a couple of questions
-can the metadata columns work with the collections discover datasets output in this tool
-the user has the option of inputting an annotation file where the columns get added to the output, the column headers are not known in advance so how would you include those column headers in the metadata columns

mblue9: Discussed with @bgruening and not going to add the metadata elements at the moment because would be too tricky to add the columns from the optional annotation file

shiltemann:
Your updates look great @mblue9 :) 

I just added some validators to the text parameters so that users cannot run the tool if they leave these fields blank or use any characters other than the ones you specify in the help text. Do you agree with these changes?

mblue9:
Thanks @shiltemann :) Your changes look good to me.

shiltemann:
great, thanks for the nice additions @mblue9 ! :)

PR #1365, opened 2017-06-14
QIIME: Assign_taxonomy wrapper fixes and updates

arbernard:
ping @bebatut 

Hi everyone,

This pull request is about the Qiime Assign_taxonomy wrapper.

During the test I've made on the latest Qiime wrappers I had noticed a few issues with this wrapper so here are some fixes and updates.

The main modifications are:

- Reactivation of the Mothur, Blast and SortMeRNA assignment methods
- Loc file support for the --id_to_taxonomy_fp parameter
- Loc file support for the --reference_seqs_fp parameter
- Addition of the --reference_seqs_fp parameter for the Mothur assignment method since it's a mandatory input for this mode
- No more empty log file returned in Mothur mode
- Addition of tests for the Mothur and Blast assignment methods
- Update of the existing tests (Uclust and SortMeRNA) so that they use md5 checksum comparisons or assert statements instead of full file comparisons
- Update of the test-data/assign_taxonomy folder content
- Update of the generate_test_dat.sh script

A word about the RDP assignment method:

I left the RDP assignment method deactivated and I also commented the RDPtools requirements because the existing conda package for the RDPtools is not compatible with the current Qiime version (more explanation available in the first commit).

The wrapper syntax has been checked with "planemo lint".

The tool execution has been tested with "planemo test".


nsoranzo:
@arbernard One of the tests for `qiime_assign_taxonomy` fails: https://travis-ci.org/galaxyproject/tools-iuc/jobs/242745850

arbernard:
Indeed it seems that the test for the Blast assignment method is failing.

I've checked the log and it's because an environment variable required to use Blast legacy (**BLASTMAT**) is not set properly:

> |  RuntimeError: BLAST cannot run if the BLASTMAT environment variable is not set.
|  Usually, the BLASTMAT environment variable points to the NCBI data directory,
|  which contains matrices like PAM30 and PAM70, etc.
|  Alternatively, you may create a .ncbirc file to define these variables.
|  From help file:
|  2) Create a .ncbirc file. In order for Standalone BLAST to operate, you
|  have will need to have a .ncbirc file that contains the following lines:
|  [NCBI]
|  Data="path/data/"
|  Where "path/data/" is the path to the location of the Standalone BLAST
|  "data" subdirectory. For Example:
|  Data=/root/blast/data
|  The data subdirectory should automatically appear in the directory where
|  the downloaded file was extracted. Please note that in many cases it may
|  be necessary to delimit the entire path including the machine name and
|  or the net work you are located on. Your systems administrator can help
|  you if you do not know the entire path to the data subdirectory.
|  Make sure that your .ncbirc file is either in the directory that you
|  call the Standalone BLAST program from or in your root directory.

The test was running fine on my computer because this environment variable was globally set by an Lmod module, sorry..

I've looked at the [meta.yaml file](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/blast-legacy/2.2.22/meta.yaml) of the bioconda package for blast-legacy 2.2.22 and the Blast installation seems to be based on the pre-compiled binary of Blast and there is no definition of the **BLASTMAT** environment variable.

Speaking of this meta.yaml file, the URL in the source section is outdated. Correct link seems to be: 
ftp://ftp.ncbi.nlm.nih.gov/blast/executables/legacy/2.2.22/blast-2.2.22-x64-linux.tar.gz

In Bioconda, there is no newer blast-legacy package available but there is one on [Biocore](https://github.com/biocore/conda-recipes/tree/master/recipes/blastlegacy/2.2.26). The Blast installation is in this case based on the source and the *BLASTMAT** environment variable is set in the script part of the meta.yaml.

I've redone the planemo tests of the assign_taxonomy wrapper with a mulled env containing the blast-legacy 2.2.26 from Biocore and it seems to be Ok.

However, I don't think packages can be taken from Biocore when the tests are executed by Travis.

Can someone tell me what should be done in this case please ? @bgruening ?

bgruening:
@arbernard the trick is to set such a `$PREFIX/etc/conda/activate.d/` script that export the env variable as shown here.

https://github.com/biocore/conda-recipes/blob/master/recipes/blastlegacy/2.2.26/meta.yaml#L29

We can not use biocore recipes, but you could update the bioconda one if you like with the activate and deactivate stuff.

shiltemann:
@arbernard updated the upstream conda recipe and tests are all green now \o/

bgruening:
@arbernard @shiltemann looks great to me! @arbernard thanks for doing the extra efforts with updating the conda recipe!

bgruening:
@shiltemann ready to merge if you want.

shiltemann:
@bgruening, just checking if the commented test can complete in time with our new extended timeout, will merge once I know :)

shiltemann:
thanks a lot @arbernard !

PR #1366, opened 2017-06-14
Adding a ballgown wrapper (test not working)

### LINKED ISSUE #1183: Create wrapper for Ballgown
A prior dependency package for the tool exists in the MTS. But conda is also available.

Once created, consider reopening this ticket for review (add tool to http://usegalaxy.org):  See: https://github.com/galaxyproject/galaxy/issues/3639

CollardT:
The tests of the tool do not run on either planemo or the official Galaxy docker. I'm open to any critique.

Fix #1183.


tools/ballgown/ballgown.R
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/Rscript
```
bgruening: !#/usr/bin/env Rscript

tools/ballgown/ballgown.xml
```diff
@@ -0,0 +1,294 @@
+<tool id="ballgown" name="Ballgown" version="0.5.0" workflow_compatible="true">
```
bgruening: Shouldn't be the version 2.2.0?

tools/ballgown/ballgown.xml
```diff
@@ -0,0 +1,294 @@
...
+  <description>Flexible, isoform-level differential expression analysis</description>
+  <requirements>
+    <requirement type="package" version="2.2.0">bioconductor-ballgown</requirement>
+    <requirement type="package" version="0.5.0">r-dplyr</requirement>
+    <requirement type="package" version="1.3.2">r-optparse</requirement>
+
+  </requirements>
+  <command interpreter="Rscript" detect_errors="aggressive">
```
bgruening: interpreter is depreacted. Can you use `Rscript $__tool_directory__/script.r` instead.

tools/ballgown/ballgown.xml
```diff
@@ -0,0 +1,294 @@
...
+		## E.g. of result:
+		## mapping = {
+		##     "e2t.ctab"   : "sample1",
+		##     "other.ctab" : "sample2",
+		##     "i2t.ctab"   : "sample1",
+		##     "t_data.ctab": "sample1"
+		##      ...
+		## }
+		##------------------------------------------------------------------------------------
+		#def read_sample_mapping_file(sample_mapping_file):
```
bgruening: This is a lot of magic, could you use configfiles instead? https://github.com/bgruening/galaxytools/blob/master/tools/sklearn/nn_classifier.xml#L15

Or a separate python file?

tools/ballgown/ballgown.xml
```diff
@@ -0,0 +1,294 @@
...
+			#set dir_name = str($output.files_path) + "/" + $key + "/"
+			$os.makedirs($dir_name)
+			#for $file_name, $file_path in $value.iteritems():
+				$os.symlink($file_path, $dir_name + $file_name)
+			#end for
+			#set n_sample = $n_sample + 1
+		#end for
+
+		## Run the R script with the location of the linked files and the name for outpot file
+		ballgown.R --directory $output.files_path --outputtranscript $output --outputgenes $outputgn --texpression $trexpression --phendat $phendata --bgout $bgo
```
bgruening: please single quote all paths and text parameters.

tools/ballgown/ballgown.xml
```diff
@@ -0,0 +1,294 @@
...
+	</command>
+  <inputs>
+    <param name="e_data" type="data_collection" collection_type="list" format="tabular" label="Exon-level expression measurements" help="One row per exon. See below for more details."/>
```
bgruening: indention looks wrong here.

tools/ballgown/ballgown.xml
```diff
@@ -0,0 +1,294 @@
...
+          * gene_name: HUGO gene name for the transcript, if known
+          * cov: per-base coverage for the transcript (available for each sample)
+          * FPKM: Cufflinks-estimated FPKM for the transcript (available for each sample)
+- **e2t**: Tab file or collection of tab files. Table with two columns, e_id and t_id, denoting which exons belong to which transcripts. These ids match the ids in the e_data and t_data tables.
+- **i2t**: Tab file or collection of tab files. Table with two columns, i_id and t_id, denoting which introns belong to which transcripts. These ids match the ids in the i_data and t_data tables.
+- samples_names: (optional) Tab file. Table with five columns, one row per sample. Defines which files from the input belong to each sample in the experiment.
+
+.. class:: infomark
+
+'''TIP''' *Note* Here's an example of a good phenotype data file for your expirement.
```
bgruening: Nice docu!!!

tools/ballgown/ballgown.xml
```diff
@@ -0,0 +1,294 @@
...
+		<param name="e2t" type="data_collection" collection_type="list" format="tabular" label="Exons-transcripts mapping" help="Table with two columns, e_id and t_id, denoting which exons belong to which transcripts. See below for more details."/>
+		<param name="i2t" type="data_collection" collection_type="list" format="tabular" label="Introns-transcripts mapping" help="Table with two columns, i_id and t_id, denoting which introns belong to which transcripts. See below for more details."/>
+		<param name="samples_names" type="data" optional="true" multiple="false" format="tabular" label="File names for samples" help="Optional. Use in case that the names for the analysed samples cannot be extracted from the filenames."/>
+    <param argument="--phendat" name="phendata" type="data" format="csv" label="phenotype data" />
+    <param argument="--texpression" name="trexpression" type="float" value="0.5" label="minimal transcript expression to appear in the results"/>
+  </inputs>
+  <outputs>
+    <data name="bgo" format="rda" label="${tool.name} on ${on_string}: ballgown object (R data file)"/>
```
bgruening: Can we make this optional? Maybe via a select box or a boolean?

tools/ballgown/ballgown.xml
```diff
@@ -0,0 +1,294 @@
...
+          <element name="HBR_Rep1" value="HBR_Rep1/i2t.ctab"/>
+          <element name="HBR_Rep2" value="HBR_Rep2/i2t.ctab"/>
+          <element name="HBR_Rep3" value="HBR_Rep3/i2t.ctab"/>
+          <element name="UHR_Rep1" value="UHR_Rep1/i2t.ctab"/>
+          <element name="UHR_Rep2" value="UHR_Rep2/i2t.ctab"/>
+          <element name="UHR_Rep3" value="UHR_Rep3/i2t.ctab"/>
+        </collection>
+      </param>
+      <param name="phendata" value="pdata.csv"/>
+      <output name="outputgn" file="genes_expression(tabular).csv"/>
```
bgruening: I don't think it is a good idea to have `(` in file names, can you use `_` or just name is genes_expression.tabular?

bgruening: Btw. this file is really a CSV file. I think in the Galaxy context it would be better to output a tsv file. You can post-process this with some shell magic if you like.

tools/ballgown/ballgown.xml
CollardT: What should be optional exactly? The last input or the first output?
Because in the case of the output I cannot do much. When I tried to have an input that is also an output, Galaxy made my tool disappear from my toolshed.

bgruening: The last output. Similar to here: https://github.com/galaxyproject/tools-iuc/blob/6e73e77d186e02d200fc3a61d25311e67864a824/tools/graphlan/graphlan.xml#L62

bgruening:
@CollardT please also add a `.shed.yml` file to the repo. This was the tests are triggered and we can upload it automatically to the TS.
Thanks.

bgruening:
@CollardT I fixed a few linting errors the tests are now running and you have not specified the `e_data` in your tests. Let us know if you need help.

CollardT:
@bgruening Can you add this wrapper to the Tool shed of Galaxy? If it needs any further enhancements, I'm open to it.
Also, I don't see my e_data not being specified in the test.
And finally, you said in one of your previous comment that an external python script can be used to move the huge command block in the XML file. But I don't know how to use the Galaxy API outside of the XML file. Do you have by chance an example or a tutorial explaining how to do so?

bgruening:
@CollardT at the moment we can not merge this as the test are failing. See here for more details: https://travis-ci.org/galaxyproject/tools-iuc/jobs/261304875

Using external python/r/... scripts is not a problem and you don't need to use the Galaxy API here. Have a look at this example for a simple r script: https://github.com/galaxyproject/tools-iuc/tree/master/tools/goseq

Essentially, just write a python script that you can call from your command section if this is really needed.

nsoranzo:
@CollardT I'm not sure that commit https://github.com/galaxyproject/tools-iuc/pull/1366/commits/f8f58a6cfd4f3470c378eac31ee110415118b48f is correct, we generally prefer to use `multiple="true"` for multiple inputs, which accommodates both multiple datasets and a list collection. Am I missing something?

nsoranzo:
I've added some fixes to @CollardT branch, but the way multiple inputs are handled still need to be fixed and I'm not sure which is the best among list collections, `multiple="true"` or repeats for this specific tool.

tools/ballgown/ballgown.xml
```diff
@@ -0,0 +1,331 @@
...
+        <param argument="--phendat" name="phendata" type="data" format="csv" label="phenotype data" />
+        <param argument="--texpression" name="trexpression" type="float" value="0.5" label="minimal transcript expression to appear in the results"/>
+        <conditional name="file_format">
+            <param argument='--format' type="select" label="Output format">
+                <option value="tsv" selected="true">tsv</option>
+                <option value="csv">csv</option>
+            </param>
+            <when value="tsv"/>
+            <when value="csv"/>
+        </conditional>
```
nsoranzo: The wrapper can be simplified a lot if we remove this conditional and just use tsv (i.e. tabular Galaxy datatype). Is csv  format somehow useful?

bimbam23:
Maybe the "$" needs to be escaped? I had a similar problem before. But I do not know much about cheetah code.

Issue #1292, opened 2017-05-04
Review of tools on usegalaxy.org


nekrut:
The following is the list of issues and action items for tools currently accessible from usegalaxy.org

----

## Data Sources

- [ ] EBI Only one dataset can be uploaded at a time, making this datasource useless
- [x] BioMart - does not work. **Remove** (galaxyproject/usegalaxy-playbook#33)
- [x] Gramene - does not work. **Remove** (galaxyproject/usegalaxy-playbook#33)
- [ ] modMine - does not have "Export to Galaxy option". Should contact mine people

----

## Collection operation

- [ ] Help sections contain extremely limited information. Need to add images generated by @jmchilton for SAB slides

----

## Text manipulation

- [ ] Concatenate datasets 1.0.0 should be removed and 0.1.0 kept (there are two of them now) (galaxyproject/usegalaxy-playbook#33)
- [x] tac - fix help section (bgruening/galaxytools#601)
- [ ] retain only one `cut` tool (galaxyproject/usegalaxy-playbook#33)

-----

## Fetch sequences and alignments tools

- [ ] Update cached MAF files for hg38, mm10 etc. (there are currently alignments containing hg38)

-----
 
## NGS: QC and manipulation

- [ ]  Select high quality segments - **Remove** (galaxyproject/usegalaxy-playbook#33)
- [x]  Draw quality score boxplot - **Remove** (galaxyproject/usegalaxy-playbook#33)
- [ ]  Barcode splitter - Images do not display in help section
- [ ] Clip adapters - Images do not display in help section
- [ ] Draw nucleotide distribution chart - Images do not display in help section

-----

## NGS: Mapping

- [ ]  Megablast - update databases
- [ ] Map with BWA for Illumina - **Remove** (galaxyproject/usegalaxy-playbook#33)
- [ ] Map with Bowtie for Illumina - **Remove** (galaxyproject/usegalaxy-playbook#33)

-----

## NGS: SAMTools

- [x]  Upgrade to latest SAMTools version

-----

## NGS: BAMTools

- [ ]  BAMTools incorrectly work with dataset collections. If a collection is fed to, for example, Filter tool, it outputs a single dataset (galaxyproject/tools-iuc#1359)

-----

## NGS: Picard

- [x]  Upgrade to latest Picard version

------

## NGS: VCF manipulation

- [x]  Upgrade to latest vcflib version

-----

## NGS: Peak Calling

- [ ] DiffBind - de-uglify help section
[...]

PR #1367, opened 2017-06-14
Genetrack updates

davebx:
- Do not output a collection, instead output a single GFF file with
peaks for all input files
 - Support mixed ScIdx and GFF input
 - Add explicit dependency on python 2.7, since the csv module behaves differently in python 3


bgruening:
@gregvonkuster can you review this please?

gregvonkuster:
@bgruening I'll take a look at this.  @davebx Can you let me know what motivated these changes, specifically changing the output type from a collection to a single file and eliminating the util code?  This tool is used in the ChIP-exo flavor / workflow workflow, so the changed output will affect that workflow.  Thanks!

davebx:
@gregvonkuster I don't know the motivation, but this is what @nekrut wrote on #1292:

> GeneTrack incorrectly packages outputs in a collection (should be a single output file, but it produces a collection)

gregvonkuster:
@nekrut can you please add your thoughts on this?  The genetrack tool allows for any number of inputs, and produces an output per selected input using a file naming convention defined here: https://github.com/galaxyproject/tools-iuc/blob/master/tools/genetrack/genetrack.py#L35.  This file naming convention is critical for the ChIP-exo analyses defined by the Pugh lab.  I chose to use a dataset collection for the output since it is best for handling these criteria - see https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-outputs-data-discover-datasets.

davebx:
@gregvonkuster would it handle your use case if I added a toggle between map and reduce functions?

gregvonkuster:
I just don't think I understand the justification for the tool having to produce a single output rather than a collection, so let's wait for @nekrut to respond.  There are other issues introduced in your PR.  For example, the underlying tool script allows for only a single input Galaxy datatype (either ```scidx``` or ```gff```, but not a mix of both.  Your PR allows for a mix of both, and this will produce incorrect outputs.

PR #1375, opened 2017-06-21
Roary the pangenome pipeline

Slugger70:
This is the initial commit of a new tool Roary. It is a pangenome pipeline for bacterial sequence analysis.


bgruening:
@Slugger70 means initial commit, ready for review? :)

Slugger70:
Yes it's ready for review.. The author of the tool and I are happy with it.

tools/roary/roary.xml
```diff
@@ -0,0 +1,187 @@
+<tool id="roary" name="Roary" version="0.1.0">
+    <description>Roary the pangenome pipeline - Quickly generate a core gene alignment from gff3 files</description>
+    
+    <requirements>
+      <requirement type="package" version="3.8.2">roary</requirement>
+    </requirements>
+    
+    <stdio>
```
bgruening: please remove this and replace by detect_errors: https://blog.qualys.com/securitylabs/2017/06/19/the-stack-clash

tools/roary/roary.xml
```diff
@@ -0,0 +1,187 @@
+<tool id="roary" name="Roary" version="0.1.0">
+    <description>Roary the pangenome pipeline - Quickly generate a core gene alignment from gff3 files</description>
```
bgruening: Roary can be removed here as it is already the "name".

tools/roary/roary.xml
```diff
@@ -0,0 +1,187 @@
...
+    
+    <stdio>
+        <exit_code range="1:" />
+    </stdio>
+    
+    <command><![CDATA[
+        #set $files=''
+        #for $gff in $gff_input.gffs
+          #if $gff.is_of_type("gff3")
+            #set $file = str($gff)+".gff"
```
bgruening: $gff is always a string isn't it?

tools/roary/roary.xml
```diff
@@ -0,0 +1,187 @@
...
+            cp $gff $file &&
```
bgruening: Can you single quote gff and file, please.
Btw. is copying necessary? I would simply use `ln -s $gff temp${counter}.gff`. This what you also don't need to construct the path above. $counter is `# for $counter, $gff in enumerate($gff_input.gffs)`

tools/roary/roary.xml
```diff
@@ -0,0 +1,187 @@
...
+        </when>
+        <when value='collection'>
+          <param name="gffs" type="data_collection" collection_type="list" format="gff3" label="Dataset collection to submit to Roary" help="A dataset list collection of gff3 files to send to Roary for analysis." />
+        </when>
+      </conditional>
+      <param name="percent_ident" type="integer" label="minimum percentage identity for blastp" help="Sets the minimum percentage identity for protein matches" value="95"/>
+      <param name="core_diff" type="float" label="percentage of isolates a gene must be in to be core" help="The total percentage of the isolates that must have the gene for it to be considered a core gene." value="99.0"/>
+      
+      <param name="outputs" type="select" multiple="true" display="checkboxes" label="Additional outputs" help="Summary statistics, core gene alignments and Gene Presence/Absence tables are always produced.">
+        <option value="abg_fa" selected="False">Accessory binary genes in fasta format</option>
```
bgruening: selected=False can be removed, it is the default.

tools/roary/roary.xml
```diff
@@ -0,0 +1,187 @@
...
+        <option value="coreacctab" selected="False">Core Accessory table in tabular format</option>
+        <option value="genepa_rtab" selected="False">Gene Presence Absence file in Rtab format</option>
+        <option value="numcons_rtab" selected="False">Number of Conserved Genes in Rtab format</option>
+        <option value="numpangene_rtab" selected="False">Number of Genes in Pan Genome in Rtab format</option>
+        <option value="numnew_rtab" selected="False">Number of New Genes in Rtab format</option>
+        <option value="numuniq_rtab" selected="False">Number of Unique Genes in Rtab format</option>
+      </param>
+      
+      <conditional name="advanced">
+        <param name="advanced_selector" type="select" label="Advanced Settings" help="Show or hide advanced settings">
```
bgruening: alternatively you could use a section: https://github.com/galaxyproject/tools-iuc/blob/e1031e4c94b25d1ed535bf221764ab801b710ab2/tools/obitools/obiannotate.xml#L82

tools/roary/roary.xml
```diff
@@ -0,0 +1,187 @@
...
+      </conditional>
+      
+    </inputs>
+    
+    <outputs>
+      <data format="tabular" name="sumstats" label="${tool.name} on ${on_string} Summary statistics" from_work_dir="out/summary_statistics.txt" />
+      <data format="fasta" name="core_gene_aln" label="${tool.name} on ${on_string} Core Gene Alignment" from_work_dir="out/core_gene_alignment.aln" />
+      <data format="csv" name="gene_p_a" label="${tool.name} on ${on_string} Gene Presence Absence" from_work_dir="out/gene_presence_absence.csv" />
+      <data format="fasta" name="acc_bin" label="${tool.name} on ${on_string} Accessory Binary Genes" from_work_dir="out/accessory_binary_genes.fa" >
+        <filter>outputs and 'abg_fa' in outputs</filter>
```
bgruening: I think you can avoid the `outputs and`. outputs should be available at any time.

tools/roary/roary.xml
```diff
@@ -0,0 +1,187 @@
...
+        <param name="gffs" value="ex1.gff,ex2.gff" ftype="gff3"/>
+        <param name="percent_ident" value="50"/>
+        <param name="core_diff" value="50.0"/>
+        <output name="sumstats" file="test2/summary_statistics.txt" ftype="tabular"/>
+      </test>
+    </tests>
+    
+    <help><![CDATA[
+Please cite Roary if you use any of the results it produces:
+Andrew J. Page, Carla A. Cummins, Martin Hunt, Vanessa K. Wong, Sandra Reuter, Matthew T. G. Holden, Maria Fookes, Daniel Falush, Jacqueline A. Keane, Julian Parkhill,
```
bgruening: Is this needed, we have the citation annotated at the end of the tool.

tools/roary/roary.xml
```diff
@@ -0,0 +1,187 @@
...
+"Roary: Rapid large-scale prokaryote pan genome analysis", Bioinformatics, 2015 Nov 15;31(22):3691-3693 
+doi: http://doi.org/10.1093/bioinformatics/btv421
+Pubmed: 26198102
+
```
bgruening: Can we instead include a small description what Rory is doing, how I can combine it with prokka and so on ...

tools/roary/roary.xml
Slugger70: I used your enumerate loop but unfortunately the files have to be copied due to the way Roary works.. Otherwise it renames the files later on with a .dat extension which makes it think it has fasta files instead of gff ones causing all sorts of havoc.

bgruening: Mh, you could smylink the files still and not copy it.

bgruening:
.shed.yml is also missing.

tools/roary/roary.xml
Slugger70: Done

tools/roary/roary.xml
Slugger70: Done

tools/roary/roary.xml
Slugger70: done

tools/roary/roary.xml
Slugger70: done

tools/roary/roary.xml
Slugger70: done

tools/roary/roary.xml
Slugger70: done

tools/roary/roary.xml
Slugger70: Without "outputs and ..." here it outputs every file regardless of choice.

bgruening: Mh, this is strange. I need to check other tools but this should not happen, have a look at: https://github.com/galaxyproject/tools-iuc/blob/643293a896a2ccfac10fe995f48c7f01c1a89a7f/tools/vsearch/clustering.xml#L104

tools/roary/roary.xml
```diff
@@ -0,0 +1,173 @@
...
+    <description>the pangenome pipeline - Quickly generate a core gene alignment from gff3 files</description>
+    
+    <requirements>
+      <requirement type="package" version="3.8.2">roary</requirement>
+    </requirements>
+    
+    <command detect_errors="exit_code"><![CDATA[
+        
+        #for $counter, $gff in enumerate($gff_input.gffs)
+          cp $gff temp${counter}.gff &&
```
bgruening: So `ln '$gff' temp${counter}.gff` does not work?

Slugger70: No. Roary copies the file to a temp directory and the copy is the original file with the .dat extension. Roary assumes anything without a .gff extension is a fasta... And so it breaks.

bgruening: fair enough - maybe you can talk to the devs, this seems bad for performance ...

tools/roary/roary.xml
nsoranzo: That's correct, we had to do the same for the Prokka wrapper. This happens for a select with `multiple="true"` and no `no_options` validator: if no option is selected, `outputs` is set by Galaxy to `None` instead of `[]`, `abg_fa in outputs` raises an Exception (TypeError: argument of type 'NoneType' is not iterable) and the filter passes.

tools/roary/roary.xml
Slugger70: I agree. The author is working on it for the next version.

tools/roary/roary.xml
bgruening: This seems to be a Galaxy bug. Users should not need to deal with this. https://github.com/galaxyproject/galaxy/issues/4221

bgruening:
Thanks @Slugger70! :)

PR #1377, opened 2017-06-22
JQ tool

erasche:
`env -i` is to run jq without an environment. Otherwise you can just do the following which is kinda undesirable.

utvalg_224


erasche:
(No idea why this is failing)

bgruening:
I guess you already tried to rebase against lastest master?

erasche:
yes. (But note that my [other PR](https://github.com/galaxyproject/tools-iuc/pull/1378) is failing with the same message.) Am I doing something wrong?

bgruening:
Thanks Eric!

PR #1379, opened 2017-06-23
Change the STAR index builder DM's sample tool_data_table_conf.xml `withGTF` column to `with-gtf` to match the tool

natefoo:
The column in the [rgrnastar tool_data_table_conf.xml.sample](https://github.com/galaxyproject/tools-iuc/blob/a831282140ce160035a4ce984f48cc20198ed0a1/tools/rgrnastar/tool_data_table_conf.xml.sample) is `with-gtf`. The conflicting names prevents Galaxy from being able to merge the data tables on startup (and it therefore fails to start).

Ping @dpryan79

PR #1381, opened 2017-06-23
Finish withGTF -> with-gtf changes in the STAR datamanager

dpryan79:
Xref: https://github.com/galaxyproject/tools-iuc/pull/1379

I remember now why I preferred `withGTF`, it made creating the dict easier since you can keep the `dict()` function call, which won't allow strings as keywords.


bgruening:
Stupid python dicts :)

PR #43, opened 2015-02-10
differences in OSX/UNIX sed

biomonika:
On OSX, I am getting error when editing my Makefile in tool_dependencies.xml (sed -i 's/O2/O2 -fPIC/' Makefile doesn't work, sed -i -e 's/O2/O2 -fPIC/' Makefile works). When installing this package on OSX, I am getting very similar error. Maybe this small edit will help. 
http://unix.stackexchange.com/questions/13711/differences-between-sed-on-mac-osx-and-other-standard-sed

PR #1392, opened 2017-06-30
Require Python 2 for cwpair

gregvonkuster:
Like numpy, it turns out the default install for matplotlib from conda uses Python 3, but the cwpair2 tool currently supports only Python 2.  This PR also applies recent standards to the command line.

### DIFF
--- a/tools/cwpair2/cwpair2.xml
+++ b/tools/cwpair2/cwpair2.xml
@@ -5,26 +5,24 @@
         <import>cwpair2_macros.xml</import>
     </macros>
     <expand macro="requirements" />
-    <command detect_errors="aggressive">
-        <![CDATA[
-            python $__tool_directory__/cwpair2.py
-            #for $i in $input:
-                 --input "${i}" "${i.hid}"
-            #end for
-            --up_distance $up_distance
-            --down_distance $down_distance
-            --method $method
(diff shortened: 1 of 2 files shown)


peterjc:
Presumably cross referencing #43 on ``CDATA``, pull request to put that into the standards pending.

tools/cwpair2/cwpair2.xml
```diff
@@ -5,26 +5,24 @@
...
+--down_distance $down_distance
+--method $method
+--binsize $binsize
+--threshold_format $threshold_format_cond.threshold_format
+#if str($threshold_format_cond.threshold_format) == "absolute_threshold":
+    --absolute_threshold $threshold_format_cond.absolute_threshold
+#elif str($threshold_format_cond.threshold_format) == "relative_threshold":
+    --relative_threshold $threshold_format_cond.relative_threshold
+#end if
+--output_files $output_files
```
yhoogstrate: quote?

gregvonkuster: In this case, $output_files is a string, not an output dataset: https://github.com/galaxyproject/tools-iuc/blob/master/tools/cwpair2/cwpair2.xml#L51

PR #1393, opened 2017-06-30
Apply recent standards to bam_to_scidx command line

gregvonkuster:


tools/bam_to_scidx/bam_to_scidx.xml
```diff
@@ -3,21 +3,11 @@
...
-        <![CDATA[
-            ln -s "${input_bam}" "localbam.bam" &&
-            ln -f -s "${input_bam.metadata.bam_index}" "localbam.bam.bai" &&
-            java -jar $__tool_directory__/BAMtoscIDX.jar -b localbam.bam -i localbam.bam.bai -p $require_proper_mate_pairing -r $read -m $min_insert_size -M $max_insert_size -o "$output" 1>/dev/null
-        ]]>
-    </command>
+    <command detect_errors="exit_code"><![CDATA[
+ln -s '${input_bam}' localbam.bam &&
+ln -f -s '${input_bam.metadata.bam_index}' localbam.bam.bai &&
+java -jar '$__tool_directory__/BAMtoscIDX.jar' -b localbam.bam -i localbam.bam.bai -p $require_proper_mate_pairing -r $read -m $min_insert_size -M $max_insert_size -o '$output' 1>/dev/null
```
nsoranzo: `min_insert_size` and `max_insert_size` have attribute `optional="True"`, so you need to handle the case where the user leaves it blank with:
```
#if str($min_insert_size):
    -m $min_insert_size
#end if
#if str($max_insert_size):
    -M $max_insert_size
#end if
```
Also this line could be split for better readability.

gregvonkuster:
Great catch @nsoranzo - thanks!

PR #1395, opened 2017-07-01
Set "--defline-qual '+'" option

pvanheus:
The default quality header line from `fastq-dump` produces a problem of [inconsistent identifiers](https://galaxyproject.org/support/ncbi-sra-fastq/). This PR adds a `--defline-qual '+'` option which sets the quality header line to '+', the suggested solution for the identifier problem. 

(The other option would be to harmonise the `--defline-seq` and `--defline-qual` options but that doesn't seem necessary.)


pvanheus:
Closing and replacing with more complete PR (#1397).

PR #1227, opened 2017-04-03
Enhance MultiGPS to handle multiple inputs

gregvonkuster:

PR #1396, opened 2017-07-02
Fix crossmap deploy

erasche:
Deploy of crossMap failed cc @yhoogstrate. I've never really used tool data tables, is this fix correct? Planemo doesn't seem to copy the sample locs in, and I've tried testing with a local instance and the chain files specified, but that isn't working and I'm not convinced it isn't because I've never used locs before.


nsoranzo:
tool_data_table_conf.xml.sample should be in the main tool directory and you also need to add the 2 loc.sample files (in tool-data/ subdir).

peterjc:
Cross reference https://github.com/galaxyproject/planemo/issues/530

If the tests are run with planemo creating its own Galaxy instance, there is no easy workaround.

If the tests are run with planemo an a pre-created Galaxy instance, then the loc files can be setup explicitly e.g. https://github.com/peterjc/pico_galaxy/blob/b3f3e8973e0ab7965f3ba14692433086cdd239a9/.travis.yml#L123

gregvonkuster:
This PR https://github.com/galaxyproject/tools-iuc/pull/1227 exposes a very similar issue to that explained by @peterjc above.  The MultiGPS tool requires a Galaxy instance that has chromosome length files installed, so:

- if the tests are run with planemo creating its own Galaxy instance (which I believe is the case with travis), the tests will not pass
- if the tests are run with planemo configured to use a pre-created Galaxy instance with chromosome length files installed, then the functional tests pass successfully.

It would be great if we could find solutions to these similar kinds of issues that affect functional testing of tools within this repository.

mvdbeek:
@peterjc @gregvonkuster So indeed, there is no way to do this without a data table, but if you do use a data table you should be able to follow http://planemo.readthedocs.io/en/latest/writing_how_do_i.html?highlight=data%20table#test-index-loc-data
Keep in mind that the dbkeys table is called `__dbkeys__`. It would be great if you could try this on your tools.

yhoogstrate:
@erasche I think the order is not correct since it does not match (`<FromSpecies>	<ToSpecies>	<PathToChainFile>`) - https://github.com/galaxyproject/tools-iuc/blob/c9cf18dfac7b6639ccc17cc6c172ca8a7692fc3a/tools/crossmap/tool-data/liftOver.loc.sample#L5:
```
    <table name="liftOver" comment_char="#">
        <columns>value, dbkey, name</columns>
```

I believe it would be better like this:
```
    <table name="liftOver" comment_char="#">
        <columns>dbkey, name, value</columns>
```

gregvonkuster:
@mvdbeek to clarify, the MultiGPS tool here https://github.com/galaxyproject/tools-iuc/pull/1227 is not using a data table - sorry, I just kind of hijacked this discussion to describe an issue that is similar to that described by @peterjc here.  Galaxy installs chromosome length files without the use of a data table - chromosome length file installation pre-dates the use of data tables, and especially Data Manager tools.

mvdbeek:
> chromosome length file installation pre-dates the use of data tables, and especially Data Manager tools.

right, but in principe you can setup chromosome length files using data tables (using the `__dbkeys__`) and they should behave the same way manually setup chromosome length files work.

gregvonkuster:
I've taken a stab at this (attempting to associate chrome length files with data tables) in the past and it just doesn't work.  The general use of ```__dbkeys__``` and data tables is for rendering dynamic select lists on the tool form whereas the the chrome length value is not rendered on the UI, it is just passed as an integer "behind the scenes" on the command line, and it is not reflected in the UI.

mvdbeek:
> The general use of __dbkeys__ and data tables is for rendering dynamic select lists on the tool form whereas the the chrome length value is not rendered on the UI, it is just passed as an integer "behind the scenes" on the command line, and it is not reflected in the UI.

Could I convince you to make a stripped down tool for demonstrating this?

gregvonkuster:
@mvdbeek this should provide a demo:
```
<tool id="MyTool" name="Test" version="1.0.0">
    <description>data tables and chrom length files</description>
    <command>echo $input.metadata.dbkey $reference_genome ${chromInfo} > '$output'</command>
    <inputs>
        <param name="input" format="txt" type="data" label="Input"/>
        <param name="reference_genome" type="select" label="Reference genome">
            <options from_data_table="all_fasta">
                <filter type="data_meta" key="dbkey" ref="input" column="1"/>
            </options>
        </param>
    </inputs>
    <outputs>
        <data name="output" format="txt" />
    </outputs>
    <tests>
    </tests>
    <help>
    </help>
</tool>
```

bgruening:
ping @erasche

yhoogstrate:
@erasche I just realized this wrapper has never been pushed to the tool sheds despite all your great efford. Would it not be wonderfull to press the green button? You have my blessing.

erasche:
Also moved the tool_data table per https://github.com/galaxyproject/tools-iuc/pull/1396#issuecomment-312478264

yhoogstrate:
Ouch, pushing fails for some reason at MTS but works on TTS.

PR #1397, opened 2017-07-03
Fix 'inconsistent identifier' problem

pvanheus:
The default quality header line from fastq-dump produces a problem of inconsistent identifiers, as explained [here](https://galaxyproject.org/support/ncbi-sra-fastq/). This PR adds a --defline-qual '+' option which sets the quality header line to '+', the suggested solution for the identifier problem.

(The other option would be to harmonise the --defline-seq and --defline-qual options but that doesn't seem necessary.)

(This replaces #1395)


bgruening:
Can you bump the version number please. Thanks

pvanheus:
@bgruening version number bumped.

nsoranzo:
Thanks @pvanheus!

PR #1402, opened 2017-07-04
JQ: add shed.yml file

shiltemann:
this tool was not yet pushed to the TS because it was lacking the `.shed.yml` file 

ping @erasche


erasche:
Thanks @shiltemann, looks great.

But for the Travis failure,

- https://travis-ci.org/galaxyproject/tools-iuc/jobs/250115971#L769 here the job fails (769)
- https://travis-ci.org/galaxyproject/tools-iuc/jobs/250115971#L841 here JQ is finally installed (841)

Anyone have ideas why the installation is happening after the job?

erasche:
Upon further inspection, I don't see JQ being tested during any of the https://github.com/galaxyproject/tools-iuc/pull/1377 original test cases. All the three middle steps ran in 1m31s, rather than the expected 3+ minutes.  I don't think that it was ever tested, probably due to missing .shed.yml

shiltemann:
we're clearing the entire environment to run jq, so i'm thinking also the PATH is cleared so JQ itself can no longer be found?

erasche:
The environment clearing *should* only apply to the invoked `jq` process, the binary should still be locatable at the level of the `env` command. But definitely plausible, maybe it isn't happening how I think it is, I didn't test with conda after all.

shiltemann:
aha ok, will poke around a bit see what's going on :)

erasche:
strike my last, you're definitely correct.

```console
$ touch jq2; chmod +x jq2; export PATH=`pwd`:$PATH;
$ env which jq2
/home/hxr/work/tools-iuc/tools/jq/jq2
$ env -i which jq2
$ echo $?
1
```

PR #1129, opened 2017-01-17
Unicycler

nekrut:
A wrapper for unicycler. Manages deps through bioconda package created by @bgruening

PR #1405, opened 2017-07-05
Add Unicycler

davebx:
This should be identical to @nekrut's #1129, but with two issues corrected that were breaking tool tests.

1. The --no_long option had been removed from unicycler, but remained in the wrapper.
2. The unicycler libraries would segfault on some architectures, leading me to believe that it was initially built with architecture-specific code. I made some changes to the [bioconda recipe](https://github.com/bioconda/bioconda-recipes/pull/4901) which should fix that. The downside is that this may potentially result in a slower unicycler, since processor-specific optimizations are no longer being used.


tools/unicycler/unicycler.xml
```diff
@@ -0,0 +1,505 @@
+<tool id="unicycler" name="Create assemblies with Unicycler" version="0.2.0">
+    <requirements>
+         <requirement type="package" version="0.3.0b">unicycler</requirement>
+         <requirement type="package" version="3.5">python</requirement>
```
bgruening: I think we agreed to not specify the python version explicitly? Maybe I'm wrong.

tools/unicycler/unicycler.xml
```diff
@@ -0,0 +1,505 @@
...
+    </requirements>
+    <stdio>
```
bgruening: can you use the detect_errors attribute from command?

tools/unicycler/unicycler.xml
```diff
@@ -0,0 +1,505 @@
...
+            ln -s '${paired_unpaired.fastq_input1}' fq.fastq.gz &&
+        #end if
+
+    #end if
+
+    ## Get location for pilon installation
+
+        pilon=`pilon --jar_dir` &&
+
+    #if $long_reads:
```
bgruening: can we move this blog up

tools/unicycler/unicycler.xml
```diff
@@ -0,0 +1,505 @@
...
+        #elif $long_reads.is_of_type('fastqsanger.gz'):
+            ln -s '${long_reads}' lr.fastq.gz &&
+        #elif $long_reads.is_of_type('fasta'):
+            ln -s '${long_reads}' lr.fasta &&
+        #end if
+    #end if
+
+    ## Running Unicycler
+
+    unicycler -t "\${GALAXY_SLOTS:-8}"
```
bgruening: 8 is a lot as default value, can we use 4 maybe?

tools/unicycler/unicycler.xml
```diff
@@ -0,0 +1,505 @@
...
+            <param name="min_component_size" optional="True" type="integer" value="" label="Unbridged graph components smaller than this size will be removed from the final graph" help="--min_component_size; default = 1000"/>
+            <param name="min_dead_end_size" optional="True" type="integer" value="" label="Graph dead ends smaller than this size will be removed from the final graph" help="--min_dead_end_size; default = 1000"/>
+        </section>
+
+        <section name="lr_align" expanded="False" title="Long read alignment parameters" help="These options control the alignment of long reads to the assembly graph.">
+            <param name="contamination_fasta" optional="True" type="data" format="fasta" label="FASTA file of known contamination in long reads, e.g. lambda, phiXm or puc18 spike-ins." help="--contamination"/>
+            <param name="scores" optional="True" type="text" size="10" value="" label="Comma-delimited string of alignment scores: match, mismatch, gap open, gap extend" help="--scores; default = 3,-6,-5,-2"/>
```
bgruening: size can be removed everywhere

tools/unicycler/unicycler.xml
davebx: Unfortunately, unicycler seems to break with python 2.x, and the conda installations I've seen have defaulted to 2.x. Not sure how to solve that without specifying the version.

bgruening: Oh, should we then skip the py2k builds from conda?

davebx: I did some additional testing, and the tool doesn't appear to need an explicit python dependency at all.

tools/unicycler/unicycler.xml
davebx: How far up?

tools/unicycler/unicycler.xml
```diff
@@ -0,0 +1,483 @@
...
+                <param name="fastq_input1" format="fastqsanger,fastqsanger.gz" type="data_collection" collection_type="paired" label="Select a paired collection" />
+            </when>
+            <when value="single">
+                <param name="fastq_input1" argument="-s" type="data" format="fastqsanger,fastqsanger.gz" label="Select unpaired reads" help="Specify dataset with unpaired reads"/>
+            </when>
+        </conditional>
+        <param name="long_reads" argument="--long" optional="True" type="data" format="fastqsanger,fastqsanger.gz,fasta" label="Select long reads. If there are no long reads, leave this empty"/>
+
+        <section name="uc_opt" expanded ="True" title="Unicycler options">
+             <param name="mode" argument="--mode" type="select" label="Select Bridging mode">
```
shiltemann: you don't need `name` attribute here as it will be determined from `argument` by stripping off all leading dashes :)

davebx: I blame RC. Thanks for the catch.

davebx: ... wait, this is a new thing, isn't it?

tools/unicycler/unicycler.xml
bgruening: Not crucial I was just thinking the symlinks should stick togehter.

bgruening:
@shiltemann if you are ok feel free to merge :)
Thanks all!

shiltemann:
yep, looks good to me, thanks @davebx !

PR #1407, opened 2017-07-06
Fail CI build if shed YAML files are missing required metadata.

jmchilton:
Does this work Travis?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ jobs:
       script:
         - set -e
         - cd "$TRAVIS_BUILD_DIR" && flake8 --exclude=.git .
-        - while read -r DIR; do planemo shed_lint --tools --ensure_metadata --urls --report_level warn --fail_level error --recursive "$DIR"; done < changed_repositories.list
+        - while read -r DIR; do planemo shed_lint --tools --ensure_metadata --urls --report_level warn --fail_level error --recursive "$DIR" && planemo shed_lint --ensure_metadata --report_level warn --fail_level warn --recursive "$DIR"; done < changed_repositories.list
 
     - stage: test
       env: CHUNK=0


.travis.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       script:
         - set -e
         - cd "$TRAVIS_BUILD_DIR" && flake8 --exclude=.git .
-        - while read -r DIR; do planemo shed_lint --tools --ensure_metadata --urls --report_level warn --fail_level error --recursive "$DIR"; done < changed_repositories.list
+        - while read -r DIR; do planemo shed_lint --tools --ensure_metadata --urls --report_level warn --fail_level error --recursive "$DIR" && planemo shed_lint --ensure_metadata --report_level warn --fail_level warn --recursive "$DIR"; done < changed_repositories.list
```
nsoranzo: You may want to remove ` --ensure_metadata` from the first `shed_lint`.

jmchilton: Actually I guess my goal would be to meet in the middle with these - I think my next would be to re-enable ``--tools`` on the second one and then just skip the linters that fail. We could then create a TODO list to enable each of those failing linters. Does that sound reasonable?

I guess the green checkbox below is meaningless though since no repositories changed... hmm...

nsoranzo: Some warnings for tools are fine as warnings, but shouldn't fail the test, e.g. the absence of citations. I think just implementing my suggestion above should be fine.

> I guess the green checkbox below is meaningless though since no repositories changed... hmm...

Indeed.

jmchilton: > Some warnings for tools are fine as warnings, but shouldn't fail the test

There are so many potential problems we are missing at the linting step then :(. It'd be nice if this repository didn't allow warning - or we had some mechanism for producing a list of known and valid warning to ignore.

nsoranzo: Something like the flake8 codes that can be ignored would be cool, but a lot of work in Planemo.

Another example of a warning (from `--ensure_metadata` in this case) that shouldn't fail the linting in my opinion:
```
WARNING: Missing shed metadata field [long_description] for repository
```
Many simple tools already have a satisfactory description and don't need a longer one.

bernt-matthias:
Since we switched to GA this seems not relevant anymore?

PR #1409, opened 2017-07-10
Abricate version bump to 0.5

Slugger70:
Updated wrapper for version 0.5. New abricate can take fasta, gbk or embl inputs.


tools/abricate/abricate.xml
```diff
@@ -1,89 +1,90 @@
...
-        #if $settings.advanced == "advanced"
-            $settings.no_header
-            #if str($settings.min_dna_id)
-                --minid=$settings.min_dna_id
-            #end if
-            #if str($settings.cull_limit)
-                --cull=$settings.cull_limit
-            #end if
+        abricate '$file_input'
+        #if $adv.no_header
```
shiltemann: we don't need the if statement here, it will either put `--noheader` or nothing in place of `$adv.no_header`, either way command should still work

tools/abricate/abricate.xml
```diff
@@ -1,89 +1,90 @@
...
 
     <inputs>
-        <param name="fasta_input" type="data" format="fasta" label="FASTA file" help="To screen for antibiotic resistant genes" />
-        <conditional name="settings">
-            <param name="advanced" type="select" label="Specify advanced parameters">
-                <option value="simple" selected="true">No, use program defaults.</option>
-                <option value="advanced">Yes, see full parameter list.</option>
+        <param name="file_input" type="data" format="fasta,genbank,embl" label="Input file (Fasta, Genbank or EMBL file)" help="To screen for antibiotic resistant genes, can be a fasta file, a genbank file or an EMBL file." />
+        <section name="adv" title="Advanced options" expanded="False">
+            <param name="db" argument="--db" type="select" label="Database to use - default is 'resfinder'" help="Option to switch to other AMR database">
```
shiltemann: `name` attribute will be automatically determined from `argument` by stripping the leading dashes, so we can remove `name="db"` in this case

tools/abricate/abricate.xml
```diff
@@ -1,89 +1,90 @@
...
-                --cull=$settings.cull_limit
-            #end if
+        abricate '$file_input'
+        #if $adv.no_header
+            $adv.no_header
+        #end if
+        #if $adv.min_dna_id
+            --minid=$adv.min_dna_id
+        #end if
+        #if $adv.db
```
shiltemann: also don't need the if statement here as this is not an optional argument, so `$adv.db` will always have a value

Slugger70:
It's difficult to get a gbk test file that's smaller due to the way that abricate works with gbk files, we'd have to artificially construct one which is not trivial apparently... I've asked the author and he said maybe. In the meantime, I'll fix everything else.

shiltemann:
ok, thanks for looking into the file size question, looks good :)

jj-umn:
@shiltemann @Slugger70   Can this be merged now?

shiltemann:
I made a PR to the branch with some smaller test data, would this be ok to use here @slugger70?

bgruening:
Thanks all!

PR #1410, opened 2017-07-10
pe_histogram command line fixes and standardization

gregvonkuster:


tools/pe_histogram/pe_histogram.xml
```diff
@@ -3,31 +3,22 @@
...
-                -l $lower_limit
-            #end if
-            #if str($upper_limit) != '':
-                -u $upper_limit
-            #end if
-            -p "$output1"
-            -t "$output2" 1>/dev/null
-        ]]>
-    </command>
+    <command detect_errors="exit_code"><![CDATA[
```
shiltemann: you could use `detect_errors="aggressive"` if you also want to fail on presence of `Error:` and `Exception:` in stderr/stdout as was the case before?

gregvonkuster: Ah, thanks for the pointer @shiltemann ;)

bgruening:
Thanks @gregvonkuster!

PR #1414, opened 2017-07-10
Add GNU parallel citation

chambm:
I saw the "--will-cite" flag passed to scripts and wondered what it was. 

https://www.gnu.org/software/parallel/man.html
```
If you use --will-cite in scripts to be run by others you are making it harder for others to see the citation notice. The development of GNU parallel is indirectly financed through citations, so if your users do not know they should cite then you are making it harder to finance development. However, if you pay 10000 EUR, you should feel free to use --will-cite in scripts.
```
I assume we didn't pay 10K EUR, so it seems reasonable to add this citation (also to any other tools that use parallel).


peterjc:
Looks good to me in principle.

That journal name looks designed to break as many bibliography tools as possible ;)

tools/freebayes/macros.xml
```diff
@@ -18,6 +18,20 @@
                     url = {http://arxiv.org/abs/1207.3907}
                 }
             </citation>
+            <citation>
```
peterjc: Use:

```xml
<citation type="bibtex">
```

This is why linting failed.

chambm:
I agree the journal name is a bit nuts. Apparently they haven't had enough complaints about it in 40 years (1977) to change it though. :laughing:

peterjc:
The citation change looks good, +1 on merging, but...

I'm not sure why the tool test is now failing, and will leave that to another reviewer to decide.

nsoranzo:
I agree with adding the citation, but anyway GNU Parallel is released under the GNU General Public License v3 or later, so cannot impose any fee for not citing, it's just a joke.

chambm:
It's all fun and games until someone gets a C&D letter! ;)

peterjc:
I interpret the citation thing as reasonable request, but not as @nsoranzo notes, a legal requirement.

Galaxy has the capability to include citation details in the tool XML, so I think we should observe this request.

nsoranzo:
I suppose the error is something about samtools versions, any idea @bgruening ?

bgruening:
Not sure why this breaks. We could try to bump the samtools version to 1.3 and adopt the calls, but this is just a guess :(

nsoranzo:
Even more strangely, the test for this tool passes on my laptop, where samtools 0.1.19 is on the PATH.

bgruening:
Mh, it looks like Galaxy is using samtools 1.3 for the upload: https://travis-ci.org/galaxyproject/tools-iuc/jobs/252091257#L738

yhoogstrate:
The notice only says that you are allowed to skip the parameter if you do pay, it does not say you are not allowed to if you do not pay :wink: 

I am a bit worried about using `parallel` without having it as dependency. It's part of the GNU tools but not of POSIX and on OSX you need to install it via brew if I'm not mistaken (please correct me if I'm wrong, but there is still no ground to assume parallel is present on any system).

nsoranzo:
@yhoogstrate A requirement for `parallel` is added in `freebayes.xml`, this is the macro file. Also, the parallel citation should be added only to that file, not in the macro.

nsoranzo:
> Also, the parallel citation should be added only to that file, not in the macro.

I resolved this and rebased on master, let's see if tests pass (they do on my laptop, as before).

nsoranzo:
Actually after removing the conda mulled environment 2 tests failed also on my laptop. I've updated the test data and they passed again. While I was at it, I also updated the freebayes and parallel requirements, sorry for hijacking your PR @chambm!

bgruening:
Thanks all!

PR #1417, opened 2017-07-13
maSigPro update - option to add cluster information

blankclemens:
This update gives the option to print out the cluster assignment for each gene in the summary.


bgruening:
Thanks!

PR #1420, opened 2017-07-19
Some fixes for QIIME

bebatut:
Hi,

I and @arbernard are testing QIIME wrappers.
I tried to fix some of things we reported

1. Explicit the data manager name and detail the shed description
2. Fix `split_libraries_fastq` and add test for the error
3. Add `extract_barcodes`, a missing wrapper required in the tutorial (not an add-on then)
4. Change the name of the wrappers to fit the name in QIIME documentation
5. Fix a bit of style (to fit more to IUC recommendations)

Bérénice


tools/qiime/qiime_core/alpha_rarefaction.xml
```diff
@@ -1,47 +1,47 @@
-<tool id="qiime_alpha_rarefaction" name="Perform alpha rarefaction" version="@WRAPPER_VERSION@.0">
+<tool id="qiime_alpha_rarefaction" name="alpha_rarefaction" version="@WRAPPER_VERSION@.0">
```
bgruening: I liked the old name more, this looks really nerdy.

bgruening: if the original name is needed, can we maybe add `(alpha_rarefaction)` to the end of description?

bgruening: @shiltemann do you have an opinion here?

shiltemann: I'm sort of in the middle here, how about just "Alpha Rarefaction"? ..especially for the large suites I like being able to find a tool by alphabet so dislike the ones starting with "Perform ..", "Calculate .." etc, but sticking with the underscore/all lowercase etc seems perhaps overdoing it a bit imho

bebatut: The idea was also to use the exact same name and description as in the [QIIME documentation](http://qiime.org/scripts/) to make it easier to find the tools (a comment from @arbernard) 
But I agree that make it difficult to read. I do not have a good solution there :(

bgruening: What is the disadvantage of putting it into the description at the end like we to it with the text_processing tools?

shiltemann: guess it just comes down to personal preference, but I have no problem with that solution @bgruening

arbernard: I have the same opinion than @shiltemann. What my colleagues that are used to use Qiime in command line don't like is when the correspondance between the original tool name and the wrapper name is not direct/obvious (Like "Perform taxonomy" instead of "Summarize taxa through plots" for example). I guess they would have no issue with using uppercase and removing underscores in the names as long as the correspondance is more obvious though.

bebatut: I moved the script name in the description and tried to have explicit names.
@arbernard Can you check if it could fit?
Thanks @bgruening, @shiltemann and @arbernard for this discussion

bgruening:
Thanks B!

PR #1425, opened 2017-08-02
Update picard suite to 2.10.6

mvdbeek:


tools/picard/picard_CollectAlignmentSummaryMetrics.xml
```diff
@@ -1,4 +1,4 @@
-<tool name="Collect Alignment Summary Metrics" id="picard_CASM" version="@TOOL_VERSION@.1">
+<tool name="Collect Alignment Summary Metrics" id="picard_CASM" version="@TOOL_VERSION@.0">
```
bgruening: Do you want to remove the trailing 0 here as well?

PR #1426, opened 2017-08-03
Add ggplot2 pca tool

bimbam23:
add tool ggplot2 PCA


bimbam23:
hmm this does not look right :-(

bgruening:
@bimbam23 what do you mean?
I guess we need to merge at first the other ggplot2 PR and then reuse the shed.yml for your tool here as well and hopefully with many more shared macros. Will try to work tm on it.

bgruening:
@bimbam23 can you have a look at: https://github.com/galaxyproject/tools-iuc/pull/1430

I have tried to use macros wherever I can and make the test pass. The heatmap2 tool could be ported to use ggplot2 instead of gplots and test could be enhanced but I hope this is a first start for many more tools.
Let me know what you think!

bimbam23:
@bgruening sounds good, I will have a look at the heatmap2 tool. Maybe it need to ported in a way that it looks more similar like the other ggplot2 tools.

bimbam23:
@bgruening can you send me a link to the repo?

bgruening:
Which repo @bimbam23? Everything is in https://github.com/galaxyproject/tools-iuc/pull/1430. Heatpamp2 is just using gplots instead of ggplot2. It's not that bad, but if there is interest it could be ported.

bimbam23:
OK, this tool is not wrapped by MO? Where can I find it

bimbam23:
Did not see the link ;-)

bgruening:
https://github.com/galaxyproject/tools-iuc/pull/1430/files#diff-3d09ec650f97f0845b3216ba95b6a2cb :)
Yes from Mo.

bgruening:
@bimbam23 ?

tools/ggplot2/ggplot2_pca.xml
```diff
@@ -1,28 +1,27 @@
...
         <requirement type="package" version="0.4.1">r-scales</requirement>
     </requirements>
     <command detect_errors="exit_code"><![CDATA[
-        Rscript '$__tool_directory__'/ggplot2_pca.R
+Rscript '$__tool_directory__/ggplot2_pca.R'
         --input '$input1'
         ${inputdata.header}
         --rowname_index ${inputdata.row_names_index}
         ${inputdata.sample_name_orientation}
-        --title \'$title\'
```
bimbam23: this is what I meant.

tools/ggplot2/ggplot2_pca.xml
```diff
@@ -1,28 +1,27 @@
...
+        --title '$title'
         --transform '$transform' 
         --background '$theme_selector'
         --scaling '$scaling.Plot_scaling' 
         $legend
         --group_type '$groups.group_type'
         #if str( $groups.group_type ) == "define_groups":
             #for $q in $groups.group_names:
-                --group_name \'$q.groupName\'
```
bimbam23: and here

tools/ggplot2/ggplot2_pca.xml
```diff
@@ -1,28 +1,27 @@
...
+                --group_name '$q.groupName'
                 --group_cols '$q.groupCols'
-                --group_colors \'$q.color_cond.plot_color\'
```
bimbam23: and here

bimbam23:
I also have a ggplot2 barplot wrapper now. I guess I open a new pull request, because it does not fit to PCA?

bgruening:
@bimbam23 can you please look at https://github.com/galaxyproject/tools-iuc/pull/1430 and let me know what you think? This could be a collection of ggplot2 based plotting tools, sharing one common infrastructure and sharing huge portions of the UI.

tools/ggplot2/ggplot2_pca.xml
```diff
@@ -0,0 +1,563 @@
...
+
+-----
+Pictures cominig soon.
+
+
+]]>
+    
+    </help>
+    <citations>
+        <citation type="bibtex">@book{wickham2009ggplot2,
```
bgruening: This is already in the macro, so you can use it instead.

tools/ggplot2/ggplot2_pca.xml
```diff
@@ -0,0 +1,563 @@
...
+          </when>
+        </conditional>
+        <conditional name="svg_bool">
+            <param name="svg_out" type="boolean" checked="false" truevalue="TRUE" falsevalue="FALSE" label="Additional output formats"/>
+            <when value="TRUE">
+                <param name="output_format" type="select" label="Select output format">
+                  <option value="pdf" selected="True">PDF</option>
+                  <option value="svg">SVG</option>
+                </param>
+                <param name="svg_res_width" label="svg width" value="16" type="integer" help="in inch (default: 16)"/>
```
bgruening: Can we have a general height/width here and a select box for formats like here: https://github.com/bgruening/galaxytools/blob/master/tools/antismash/antismash.xml#L180

bgruening: Would be also nice if this can be moved into a macro so the other tools can use it as well.

tools/ggplot2/ggplot2_pca.xml
bimbam23: what exactly do you mean?

shiltemann: @bimbam23 see https://github.com/galaxyproject/tools-iuc/blob/master/tools/ggplot2/macros.xml#L82, so you can use `<expand macro="citations"/>` here :)

bimbam23: ok but I would like to have more than one citation. and the bib format is only shown correctly if it hast a name wickham2009ggplot2,

shiltemann: if you would like to add citations specifically for this component, you can use `<yield/>` in the macro and add the extra citation(s) in your tool. And if it is not displaying correctly please update the macro so it is fixed in all ggplot tools :)

shiltemann: about to leave for home or would find you an example for the yield thing, but I show an example it in my [cheatsheet](https://github.com/shiltemann/cheatsheets/blob/master/galaxy-tools.md#macros)

mvdbeek: Also I think citing the doi would be better `<citation type="doi">10.1007/978-0-387-98141-3</citation`

bimbam23: thanks... I was still using an old pca.xml file. in the newest its fixed ;-)

tools/ggplot2/ggplot2_pca.xml
bimbam23: yes it would be nice, I think in this case its not practical because the selection of heights and widths is quite important for the visualization of the plot. That means it depends on the plot size and elements you plot

bimbam23:
I changed the output options. If everyone likes it I can change it for all ggplot tools. otherwise I will modify the macro in such a way that this output option is only used for some of the tools

bimbam23:
travis is failing because I have not fix the other ggplot tools?

bgruening:
I think this error is not related to you. We need to fix the channel order here in IUC.

bimbam23:
Hej ho, are there any complains about the new outputs? Should I change all ggplots-tools accordingly? Is there anything else to be optimized?

bimbam23:
it looks line travis is not starting.

nsoranzo:
@bimbam23 I saw that on another PR too, I'll try to close and reopen.

bimbam23:
does anyone see a proper error?

tools/ggplot2/ggplot2_pca.xml
```diff
@@ -0,0 +1,411 @@
+<tool id="ggplot2_pcaplot" name="PCA plot w ggplot2" version="@VERSION@">
+    <macros>
+       <import>macros2.xml</import>
```
nsoranzo: Can you use 4-spaces indentation throughout the file, please?

nsoranzo:
The error seems to be that `test-data/ggplot_pca_result1.svg` is different from the output produced when running the tool on Travis. Maybe that's not deterministic?

bimbam23:
Now it works, any other complains?

tools/ggplot2/test-data/ggplot_pca_result1.svg
```diff
@@ -0,0 +1,90 @@
+<?xml version='1.0' encoding='UTF-8' ?>
```
nsoranzo: You can remove this file now.

tools/ggplot2/macros2.xml
```diff
@@ -0,0 +1,251 @@
+<?xml version="1.0"?>
```
nsoranzo: There seems to be a lot of duplication between `macros.xml` and this file, which may be a problem for the long term maintenance of this code, can this be merged back into `macros.xml`?

bimbam23:
That's true. The problem is that heatmap does not use ggplot2 so that I left macros.xml for heatmap... Heatmaps in ggplot2 have not the same nice functionality.

nsoranzo:
@bimbam23 `heatmap2.xml` use only few macros, so I went ahead and merge the 2 files. I pushed the commit with some other small changes to your branch.

bimbam23:
looks good to me.

tools/ggplot2/ggplot2_pca.xml
```diff
@@ -0,0 +1,393 @@
...
+                        help="Use short names, avoid special characters and numbers at the beginning of the name (The names might be changed by the program to make them conform to processing in R)">
+                        <validator type="no_options" message="Please choose a name for this group!"/>
+                    </param>
+                    <param name="groupCols" type="data_column" data_ref="input1" multiple="true" min="1" label="Columns for group"
+                        help="WARNING: please use columns indices after removing the row name column if selected in second section!">
+                        <validator type="no_options" message="Select at least one column for each group!"/>
+                        <!--<filter type="remove_value" meta_ref="groupCols" />-->
+                    </param>
+                    <conditional name="color_cond">
+                        <param name="color_bool" type="boolean" checked="false" truevalue="--colors" falsevalue="--no-colors" label="Select colors"/>
```
nsoranzo: Can you use a `select` instead of a `boolean` type here? See https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#dynamic-options

bimbam23:
I have another idea for the heatmap with ggplot2. I will test if its possible

bimbam23:
I have one issue, how can I add an unusual channel? https://anaconda.org/mcole111/r-egg conda install -c mcole111 r-egg

bimbam23:
Ok I hope that is will be easy because the package is already present in conda

bimbam23:
@bgruening is it possible to get the meta and build files from the conda link? or is there an easy way to put r-egg into conda-forge?

bgruening:
For example this one: https://github.com/bgruening/conda_r_skeleton_helper
Just give me a day and we will have this egg in conda-forge.

bimbam23:
Thanks Björn!

bgruening:
restarted ...

bimbam23:
is this maybe the problem:
```
Loading required package: ggplot2

Attaching package: ‘cowplot’

The following object is masked from ‘package:ggplot2’:

    ggsave

Warning messages:
1: package ‘cowplot’ was built under R version 3.4.3 
2: package ‘ggplot2’ was built under R version 3.4.2 
Loading required package: gridExtra
Warning messages:
1: package ‘egg’ was built under R version 3.4.4 
2: package ‘gridExtra’ was built under R version 3.4.2 

Attaching package: ‘dplyr’

The following object is masked from ‘package:gridExtra’:

    combine

The following objects are masked from ‘package:stats’:

    filter, lag

The following objects are masked from ‘package:base’:

    intersect, setdiff, setequal, union

Warning message:
package ‘dplyr’ was built under R version 3.4.2 
Warning message:
package ‘ggdendro’ was built under R version 3.4.4 
Warning message:
package ‘reshape2’ was built under R version 3.4.3 
```

bimbam23:
Is it a problem to have every package complied under a different R version?
The test at my PC is working...

bimbam23:
With the last changes we could also remove the heatmap.xml that does not use ggplot2?

tools/ggplot2/macros.xml
```diff
@@ -67,31 +144,66 @@ loc <- Sys.setlocale("LC_MESSAGES", "en_US.UTF-8")
         </conditional>
     </xml>
     <xml name="title">
-        <param name="title" type="text" value="plot title" label="Title of plot" />
+	   <param name="title" type="text" label="Plot title" value="">
+            <sanitizer sanitize = "false"/>
```
nsoranzo: Isn't that dangerous? If the user insert a single quote, your R script will probably fail.

tools/ggplot2/macros.xml
```diff
@@ -9,49 +9,126 @@
...
-    gg_scaley = ylim($scaling.yaxismin, $scaling.yaxismax)
-#end if
-    ]]></token>
-
     <xml name="xy_scaling">
         <conditional name="scaling">
-            <param name="plot_scaling" type="select" label="Advanced - Axis scaling">
+            <param name="plot_scaling" type="select" label="Axis scaling">
                 <option value="Automatic" selected="True">Automatic axis scaling</option>
                 <option value="Defined">User deined axis scales</option>
```
nsoranzo: s/User deined/User-defined/

nsoranzo:
> With the last changes we could also remove the heatmap.xml that does not use ggplot2?

@bimbam23 You surely can't have 2 tools with the same tool id `ggplot2_heatmap2` like you have now, you either have to remove the old one or give a different id to the new one.

tools/ggplot2/macros.xml
bimbam23: I thinks this is really not necessary. I only added this to use special characters.

bimbam23:
I think travis failed

nsoranzo:
Restarted the failed job, green now.

tools/ggplot2/ggplot2_heatmap.xml
```diff
@@ -0,0 +1,190 @@
+<tool id="ggplot2_heatmap" name="Heatmap w ggplot" version="@VERSION@">
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <requirements>
+        <requirement type="package" version="7.0">readline</requirement>
```
bgruening: readline should not be needed, can you try without it?

tools/ggplot2/ggplot2_pca.xml
```diff
@@ -0,0 +1,396 @@
+<tool id="ggplot2_pcaplot" name="PCA plot w ggplot2" version="@VERSION@">
```
bgruening: only ggplot2_pca as an ID?

tools/ggplot2/ggplot2_pca.xml
```diff
@@ -0,0 +1,396 @@
...
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <requirements>
+        <requirement type="package" version="7.0">readline</requirement>
```
bgruening: readline should not be needed

tools/ggplot2/ggplot_histogram.xml
```diff
@@ -1,71 +1,72 @@
...
-Rscript '$script'
-    ]]></command>
+        cat '$script' &&
+        Rscript '$script']]>     
+    </command>
     <configfiles>
         <configfile name="script"><![CDATA[
 
-@R_INIT@
+            @R_INIT@
```
bgruening: fyi: you can leave the indentation as it was.

tools/ggplot2/heatmap2.xml
```diff
@@ -1,11 +1,11 @@
-<tool id="ggplot2_heatmap2" name="heatmap2" version="@VERSION@">
+<tool id="ggplot2_heatmap2" name="heatmap2" version="@VERSION@+galaxy0">
```
bgruening: We should give better names to the heatmap tools to help users to choose the proper tools.
Evil question, do we need both?

tools/ggplot2/ggplot2_heatmap.xml
bimbam23: I tried before and it threw an error: libreadline.so.6 ...

tools/ggplot2/heatmap2.xml
bimbam23: at least we should remove the "ggplot2" out of the id name, because it only uses gplot.

bgruening: I'm afraid we can not change the id anymore :(

bimbam23:
`cat '$script' &&` this can also be remove? I think it was debugging stuff from Mo.

bgruening:
> cat '$script'

Will print the actual used script to the stdout, so it might be useful for a few users.

mvdbeek:
> Will print the actual used script to the stdout, so it might be useful for a few users.

In that case we could make this an optional output and not have it be stored in the database

bimbam23:
So should we kick heatmap2? Or should we put it outside ggplot2? Some people prefer plots without ggplot2

bimbam23:
are there more complains or changes needed?

bgruening:
ping @galaxyproject/iuc 
Can someone please help here?

bimbam23:
ok looks like its working... not on my pc here ... But better without readline

bgruening:
Thanks @nsoranzo and @bimbam23!

PR #1427, opened 2017-08-03
Add wrapper for kallisto 0.43.1

davebx:
And data manager to generate kallisto indexes.


tools/kallisto_quant/.shed.yml
```diff
@@ -0,0 +1,14 @@
...
+- Transcriptomics
+description: Quantify abundances of RNA-Seq transcripts
+homepage_url: https://pachterlab.github.io/kallisto/
+long_description: |
+    kallisto is a program for quantifying abundances of transcripts from RNA-Seq
+    data, or more generally of target sequences using high-throughput sequencing
+    reads. It is based on the novel idea of pseudoalignment for rapidly 
+    determining the compatibility of reads with targets, without the need for 
+    alignment.
+name: kallisto_quant
```
bgruening: No suite?

tools/kallisto_quant/kallisto_quant.xml
```diff
@@ -0,0 +1,154 @@
+<?xml version="1.0"?>
+<tool id="kallisto_quant" name="Kallisto" version="@VERSION@.0">
+    <description>quantify abundances of RNA-Seq transcripts</description>
+    <macros>
```
bgruening: These macros only make sense if we put them into a separate file and share them, right?
So maybe we make a suite out of if and this is the first tool?

tools/kallisto_quant/kallisto_quant.xml
```diff
@@ -0,0 +1,154 @@
...
+            -o .
+            #if str($single_paired.single_paired_selector) == 'single':
+            	--single
+                #set $single_reads = "' '".join(str($single_paired.reads).split(','))
+                --fragment-length $single_paired.fragment_length
+                --sd $single_paired.sd
+                '$single_reads'
+            #else:
+            	#if str($single_paired.collection.collection_selector) == 'datasets':
+					#set $forward_reads = str($single_paired.collection.forward).split(',')
```
bgruening: indentation?

tools/kallisto_quant/kallisto_quant.xml
```diff
@@ -0,0 +1,154 @@
...
+                <option value="single" selected="true">Single-end</option>
+                <option value="paired">Paired</option>
+            </param>
+            <when value="single">
+                <param name="reads" type="data" format="fastq" multiple="True" label="Reads in FASTQ format" />
+                <param name="fragment_length" argument="--fragment-length" type="integer" value="200" label="Average fragment length" help="Illumina typically produces reads of 180-200bp" />
+                <param argument="--sd" type="integer" value="20" label="Estimated standard deviation of fragment length" />
+            </when>
+            <when value="paired">
+		        <conditional name="collection">
```
bgruening: indentation?

tools/kallisto_quant/kallisto_quant.xml
```diff
@@ -0,0 +1,154 @@
...
+		                <param name="reverse" type="data" format="fastq" multiple="True" label="Reverse reads" />
+		            </when>
+		            <when value="collection">
+		            	<param name="reads" type="data_collection" format="fastq" collection_type="list:paired" label="Collection of reads" />
+		            </when>
+		        </conditional>
+            </when>
+        </conditional>
+        <param argument="--bias" type="boolean" truevalue="--bias" falsevalue="" label="Perform sequence based bias correction" />
+        <param name="bootstrap_samples" argument="--bootstrap-samples" type="integer" value="0" label="Number of bootstrap samples (default: 0)" />
```
bgruening: can we put argument always at the first position?

bgruening: default should be removed from the label

tools/kallisto_quant/kallisto_quant.xml
```diff
@@ -0,0 +1,154 @@
...
+        <param argument="--seed" type="integer" value="42" label="Seed for the bootstrap sampling (default: 42)" />
```
bgruening: default, please not as label

tools/kallisto_quant/kallisto_quant.xml
```diff
@@ -0,0 +1,154 @@
...
+        <param argument="--fusion" type="boolean" truevalue="--fusion" falsevalue="" label="Search for fusions for Pizzly" />
```
bgruening: for Pizzly, should be moved to help, and we should wrap Pizzly as well :)

tools/kallisto_quant/kallisto_quant.xml
```diff
@@ -0,0 +1,154 @@
...
+        <param argument="--pseudobam" type="boolean" truevalue="--pseudobam" falsevalue="" label="Output pseudoalignments in SAM format to stdout" />
```
bgruening: stdout is confusing for uers, we should just name it - output pseudoalignment file
And we should maybe autoconvert it to BAM

bgruening: This should go into its own file, not stdout, with a filter and default deactivated

tools/kallisto_quant/kallisto_quant.xml
```diff
@@ -0,0 +1,154 @@
...
+    </inputs>
+    <outputs>
+        <data format="h5" name="abundance_h5" from_work_dir="abundance.h5" label="${tool.name} on ${on_string}: Abundances (HDF5)" />
+        <data format="tabular" name="abundance_tab" from_work_dir="abundance.tsv" label="${tool.name} on ${on_string}: Abundances (tabular)" />
+        <data format="json" name="run_info" from_work_dir="run_info.json" label="${tool.name} on ${on_string}: Run info" />
```
bgruening: This should actually go to the stdout imho. It is not useful in a general Galaxy context.

tools/kallisto_quant/kallisto_quant.xml
```diff
@@ -0,0 +1,154 @@
...
+            <param name="reverse" ftype="fastq" value="mm10_chrM-1.r.fq,mm10_chrM-2.r.fq,mm10_chrM-3.r.fq,mm10_chrM-4.r.fq,mm10_chrM-5.r.fq" />
+            <output name="abundance_tab" file="kallisto_quant_out1.tab" ftype="tabular" />
+            <output name="run_info" file="kallisto_quant_out1.json" ftype="json" lines_diff="4" />
+        </test>
+        <test>
+            <param name="reference" ftype="fasta" value="mm10_chrM.fa" />
+            <param name="single_paired_selector" value="paired" />
+            <param name="collection_selector" value="collection" />
+            <param name="reads">
+		        <collection type="list:paired">
```
bgruening: indentation

tools/kallisto_quant/kallisto_quant.xml
nsoranzo: Actually `argument` is second after `name` in https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#coding-style because `name` takes precedence.

tools/kallisto_quant/kallisto_quant.xml
davebx: oh for ... I meant to do tabs to spaces on the file before pushing, but forgot

tools/kallisto_quant/kallisto_quant.xml
davebx: I use the macros in this case so maintainers only have to update the version string in one place, but I have no problem putting this in tool_collections instead, if you think the other kallisto functions are worth wrapping in the near future.

tools/kallisto_quant/kallisto_quant.xml
davebx: You're right. Keep an eye on bioconda-recipes for a pull request in the near future.

davebx:
Tomorrow I'll be creating a data manager that indexes fasta files in the fasta data table and populates the kallisto_indexes data table, because that seems like something useful to have pre-generated. When that's done, I'll update this pull request with the appropriate ... updates.

tools/kallisto_quant/kallisto_quant.xml
bgruening: Not tool_collection, but suite, like: https://github.com/galaxyproject/tools-iuc/blob/master/tools/datamash/.shed.yml

I think we need two tools in the end.

bgruening:
data_manager would be nice, I would still offer the option to use a FASTA file and index on the fly.

tools/kallisto_quant/kallisto_quant.xml
mvdbeek: It may be for summary tools like multiqc: https://github.com/ewels/MultiQC/blob/master/docs/modules/kallisto.md

davebx:
@bgruening oh, of course. Same way hisats and tophats do their alignment, where you can select to align against a history genome or a data table genome.

tools/kallisto_quant/run_info.json
```diff
@@ -0,0 +1,9 @@
+{
+	"n_targets": 1,
```
bgruening: This file can be removed?

davebx: Yeah, that was a remnant of development that was caught in an overzealous `git add .`

tools/kallisto/kallisto_pseudo.xml
```diff
@@ -0,0 +1,246 @@
...
+                        <param name="reads" type="data_collection" format="fastq" collection_type="list:paired" label="Collection of reads" />
+                    </when>
+                </conditional>
+                <param argument="--umi" type="boolean" truevalue="--umi" falsevalue="" label="File with forward reads is a UMI file" />
+            </when>
+        </conditional>
+        <param argument="--pseudobam" type="boolean" truevalue="--pseudobam" falsevalue="" label="Output pseudoalignments in BAM format" />
+    </inputs>
+    <outputs>
+            <data format="tabular" name="sample">
```
bgruening: indentation is wrong here

tools/kallisto/kallisto_pseudo.xml
```diff
@@ -0,0 +1,246 @@
...
+            if [ -f matrix.tsv ] ; then mv matrix.tsv outputs/Matrix.tabular ; fi &&
+            if [ -f matrix.cells ] ; then mv matrix.cells outputs/Matrix.cells ; fi &&
+            if [ -f pseudoalignments.tsv ] ; then mv pseudoalignments.tsv outputs/Pseudoalignments.tabular ; fi &&
+            if [ -f pseudoalignments.ec ] ; then mv pseudoalignments.ec outputs/Pseudoalignments.ec ; fi
+        ]]>
+    </command>
+    <configfiles>
+        <configfile name="batch">
+<![CDATA[
+#if str($single_paired.single_paired_selector) == 'single':
```
bgruening: I think we need to get UMIs in here as well 
https://pachterlab.github.io/kallisto/manual

bgruening:
@gregvonkuster fyi this includes a test for changed reference genome - maybe this is useful.

bgruening:
ping @jxtx 
Thanks a lot @davebx!

lakigigar:
Thanks a lot!

PR #1432, opened 2017-08-07
Remove multiple='true' option for input files in kallisto.

mvdbeek:
The kallisto manual specifies that `Important note: only supply one
sample at a time to kallisto. The multiple FASTQ (pair) option is for
users who have samples that span multiple FASTQ files`. In galaxy the
user should probably concatenate the sample files if that is the case.

Not using multiple=`true` in turn allows us to map over multiple files.
I have also changed `collection_type='list:paired'` to
collection_type='paired'`, so that we can map over collections where
each collection element is one sample.
Finally I have allowed fastq.gz input. Kallisto automatically detects
compression, so no changes in the command secton were required for this.


davebx:
Thanks, @mvdbeek. I missed the .gz part, so that's also a good addition.

PR #1433, opened 2017-08-07
Bump kraken versions to create new installable revisions in toolshed.

davebx:
I updated the version to 1.2.3 so that all of the kraken wrappers will get an update, since one of them was previously already 1.2.2 and the others were 1.2.1

I also granted iuc write access to the four repositories, so that will actually work. 

This will enable toolshed installation of the condafied krakens.


bgruening:
Was the old version here never updated to the MTS?

bgruening:
`condafied krakens` I would love to see a logo for this :)

davebx:
@bgruening IUC did not have write access to devteam/kraken*, which is why it wasn't updated before.

PR #1434, opened 2017-08-08
fix Qiime warnings

bgruening:


tools/qiime/qiime_core/pick_otus.xml
```diff
@@ -134,13 +134,13 @@ pick_otus.py
                 <expand macro="pick_otus_similarity"/>
                 <expand macro="pick_otus_suppress_prefilter_exact_match"/>
             </when>
-            <when value="mothur">
+            <!--when value="mothur">
                 <param argument="--clustering_algorithm" type="select" label="Clustering algorithm">
```
nsoranzo: `--` cannot appear inside XML comments: https://stackoverflow.com/questions/10842131/xml-comments-and

bgruening: Thanks Nicola. I would like to leave the decision to @bebatut if we remove this or mask it in an other way.

bebatut: Why do you comment this argument? I know that some people would like to be able to use the mothur option, I could just not tested it on my computer...

bgruening: @bebatut these options are not enabled in the wrapper and Galaxy returns an error because the `when` exists but it can not be selected.

nsoranzo: Galaxy returns an error because this is not valid XML as I have explained above.

bgruening: Nicola, it is not my decision to remove the when or to comment it out. B! Needs to decide this. I just want to get rid of the warning. @bebatut should I remove both options or do you want it commented out until it is fixed. Currently this option is not included.

PR #1435, opened 2017-08-09
Add UMI-tool extract

blankclemens:
> Extract UMI barcode from a read and add it to the read name, leaving any sample barcode in place. Can deal with paired end reads and UMIs split across the paired ends.

This is only the extract tool from the UMI-tools suite, other parts may follow.


davebx:
Ooh, this is exactly what I need to finish generating proper test data for `kallisto pseudo --umi`

Thanks for the work, @blankclemens

tools/umi_tools/.shed.yml
```diff
@@ -0,0 +1,12 @@
+categories:
+- Sequence Analysis
+- Transcriptomics
+description: UMI-tools extract - Extract UMIs from fastq
+homepage_url: https://github.com/CGATOxford/UMI-tools
+long_description: |
+  Extract UMI barcode from a read and add it to the read name, leaving
+  any sample barcode in place. Can deal with paired end reads and UMIs
+  split across the paired ends
+name: umi-tools_extract
```
bgruening: If the other tools might follow, maybe you want to make a suite definition here already?

tools/umi_tools/umi-tools_extract.xml
```diff
@@ -0,0 +1,306 @@
+<tool id="umi-tools_extract" name="UMI-tools extract" version="0.4.4.0">
+    <description>Extract UMI from fastq(.gz) files</description>
```
bgruening: I would remove the (,gz) here, a user should not care about this detail.

blankclemens: Indeed, thanks!

tools/umi_tools/.shed.yml
blankclemens: Would there be any advantage at the moment compared to making it a suite with the second tool?

tools/umi_tools/umi-tools_extract.xml
```diff
@@ -0,0 +1,306 @@
...
+                #end if
+            #end if
+            #if not $prime3:
+                --3prime
+            #end if
+            #if $quality.quality_selector =='true':
+                --quality-filter-threshold '$quality.quality_filter_threshold'
+                --quality-encoding '$quality.quality_encoding'
+            #end if
+            --log='$out_log'
```
bgruening: Can this be disabled if not used?

tools/umi_tools/umi-tools_extract.xml
```diff
@@ -0,0 +1,306 @@
...
+        </param>
+        <param name="prime3" argument="--3prime" type="boolean" label="Is the barcode at the 5' end?"
+            truevalue="1" falsevalue="0" checked="true"
+            help="By default the barcode is assumed to be on the 5' end of the read, but
+                use this option to sepecify that it is on the 3' end instead." />
+        <param name="print_log" type="boolean" label="Output log?"
+            truevalue="1" falsevalue="0" checked="true"
+            help="Choose if you want to generate a text file containing logging information." />
+        <conditional name="quality">
+            <param name="quality_selector" type="select" label="Enable quality filter?" >
```
bgruening: This looks like a ideal candidate for a `section` :)

tools/umi_tools/.shed.yml
bgruening: the creation of repositories are different ... the names are generated out of ID ... so tool IDs are in sync with the repo names.

blankclemens: Done, thanks!

tools/umi_tools/umi-tools_extract.xml
blankclemens: While I am a big fan of sections, I don't see any advantage here. I like them for big, often used settings that would ruin the tool interface if always displayed. Here we have two options which do not identify as the core utility of this tool.

PR #1438, opened 2017-08-14
Query tabular  issue 1428  

jj-umn:
query_tabular issue 1428 detect column names
    
    1. Allow selection of prefix when including column names in tabular
    output.
    2. Use values in first line of file for column names.  The names will
    be quoted if not a valid SQLite column name:  using double quotes,
    square brackets,  grave accents, or single quotes.  See:
    https://sqlite.org/lang_keywords.html


tools/query_tabular/test-data/psm_report.tsv
```diff
@@ -0,0 +1,10 @@
+
```
bgruening: @jj-umn why is the first line empty?

tools/query_tabular/query_tabular.xml
```diff
@@ -406,6 +463,45 @@ Query Tabular
...
+  Specify Name for Table:
+
+    - By default tables will be named: t1, t2, t3, etc.
+ 
+    - **Specify Name for Table** 
+      You can provide your own name for a database table, the name should begin with a letter and only contain letters, digits, or underscores. 
+      The name should not be a SQLite key word: https://sqlite.org/lang_keywords.html
+
+
+  There are several way to name columns in a table:
```
bgruening: s/way/ways/

tools/query_tabular/test-data/psm_report.tsv
jj-umn: Some versions of PeptideShaker printed a blank line, then the header line in the PSM report output.  That was one of the inspirations for regex line filtering.

tools/query_tabular/query_tabular.xml
jj-umn: Will do.  Thanks.

tools/query_tabular/test-data/psm_report.tsv
bgruening: Ok, thanks for the explanation and thanks again!

bgruening:
Thanks JJ!

PR #1440, opened 2017-08-18
fixing krona warnings

bgruening:
<string>:4: (ERROR/3) Duplicate target name, cannot be used as a unique reference: "krona".


nsoranzo:
LGTM, can you update the output file or increase the lines_diff?

PR #1450, opened 2017-08-25
Picrust edits and metagenome contrib

gavinmdouglas:
3 main edits to the PICRUSt galaxy wrappers:

1) Clarified that other functional predictions besides KEGG orthologs/pathways can be generated.
2) Forced the output to be “biom1” for all scripts so that they can be used as input for the next step.
3) Added in a fourth PICRUSt script: metagenome_contributions.py.


tools/picrust/categorize_by_function.xml
```diff
@@ -1,5 +1,5 @@
-<tool id="picrust_categorize" name="Categorize by function" version="@WRAPPER_VERSION@.0">
```
bgruening: Changing the tool-id of an existing tool is not recommended as this will break the tool-linage and workflow updates are not possible anymore :-(

tools/picrust/categorize_by_function.xml
```diff
@@ -11,36 +11,30 @@
...
             <option value="taxonomy">Using taxonomy metadata</option>
         </param>
         <param argument="--level" type="integer" value="3" min="1" label="Hierarchy Level"
                help="a value of 1 is the highest level, and any higher value nears the leaves of the hierarchy. For instance, if the hierarchy contains 4 levels, specifying 3 would collapse at one level above being fully specified"/>
-        <param argument="--format_tab_delimited" type="boolean" truevalue="-f" falsevalue="" checked="false" label="output the predicted metagenome table in tab-delimited" help="will contain the metadata lost in the classic output"/>
         <expand macro="biom_format_select"/>
     </inputs>
     <outputs>
         <expand macro="biom_output"/>
-        <data name="out_tabdel" format="tabular" from_work_dir="tempbiom" label="${tool.name} on ${on_string}: metagenome table">
```
bgruening: Was this option removed?

tools/picrust/categorize_by_function.xml
```diff
@@ -11,36 +11,30 @@
...
-        </data>
     </outputs>
     <tests>
         <test> <!-- test with Kegg Pathways and json output -->
             <param name="input_fp" value="predicted_metagenomes.L3.biom"/>
             <param name="metadata_category" value="KEGG_Pathways"/>
             <param name="level" value="2"/>
-            <param name="format_tab_delimited" value="-f"/>
             <param name="output_type" value="json"/>
-            <output name="out_biom" ftype="json">
```
bgruening: But the output type chosen by the user, one line above it json?

tools/picrust/macros.xml
```diff
@@ -1,8 +1,8 @@
 <macros>
-    <token name="@WRAPPER_VERSION@">1.0.1</token>
+    <token name="@WRAPPER_VERSION@">1.0.2</token>
```
bgruening: Should this be 1.1.1 as the requirement?

tools/picrust/normalize_by_copy_number.xml
```diff
@@ -53,8 +53,15 @@ This module corrects the abundance of each OTU to better reflect the true organi
 Please ensure that you have properly created your OTU table to be compatible with PICRUSt by following this guide_.
 A sample file can be downloaded here_
 
+The input file should be in biom1 format. You can upload the file from your local computer by going to Get Data -> Upload File. Make sure you choose the "biom1" format when you upload your file.
```
bgruening: I don't think such Info's should be part of the tool. If its unclear we should make it more obvious for the user in Galaxy.

tools/picrust/normalize_by_copy_number.xml
```diff
@@ -53,8 +53,15 @@ This module corrects the abundance of each OTU to better reflect the true organi
...
+
+Make sure that you specify an appropriate 16S database for this command (usually 16S 13_5).
+
+For administrators of this galaxy instance, if you are having trouble getting the precalculated files installed please visit this page_.
+
```
bgruening: Can you add this in a readme file instead of a tool?

bgruening:
ping @shiltemann for a more detailed review about the data types.

tools/picrust/categorize_by_function.xml
gavinmdouglas: Thanks for all of your helpful comments! That's good to know about changing the tool-id. I changed the name back to the original. Will there be a similar problem due to my having renamed an xml file normalize_by_copynumber.xml --> normalize_by_copy_number.xml?

tools/picrust/categorize_by_function.xml
gavinmdouglas: No it's still within the script, but I believe this output option was redundant with the "biom_output" option.

tools/picrust/categorize_by_function.xml
nsoranzo: No, changing filename is fine if you keep the tool id.

tools/picrust/categorize_by_function.xml
gavinmdouglas: There is likely a better way around this issue, but the scripts weren't able to read in the output files of upstream steps in json or hdf5 format since they are expecting "biom1" as the input format. My fix was to set "biom1" as the format when either of these formats were selected. Note that files ending in ".biom" are commonly written in both formats so I don't think this is too confusing.

tools/picrust/macros.xml
gavinmdouglas: Sure I can change that - I wasn't sure how the versioning of the wrappers is usually done.

tools/picrust/normalize_by_copy_number.xml
gavinmdouglas: Sure, I have removed this info.

tools/picrust/normalize_by_copy_number.xml
gavinmdouglas: Yep, I have moved the linked page to be the README.

tools/picrust/categorize_by_function.xml
shiltemann: yes, the `biom1` galaxy format signifies the json-formatted version of the biom file, so I think it is fine to keep this as it is. For hdf5 output however, I would say it may be worth to keep it as `h5` datatype so other tools that work with hdf5 files can still take those files as input too. In that case just add `format="biom1,h5"` to any input datasets that can handle both. 

Ideally we should probably subclass the h5 datatype and look for the format-version and format-url  strings (http://biom-format.org/documentation/format_versions/biom-2.1.html) to make a biom2 datatype to signify the hdf5 version of the biom file (but that can be done at a later time)

tools/picrust/README.md
```diff
@@ -0,0 +1,36 @@
+## PICRUSt Galaxy Wrapper
+
+This page is for anyone interested in installing [PICRUSt](http://picrust.github.io/picrust/) in a [galaxy instance](https://galaxyproject.org/).The below descriptions should clarify the most confusing part of the installation.
+
+The most important steps are that you need to go to the Galaxy IUC tool shed and install the 4 wrapper scripts there along with PICRUSt itself. If you install PICRUSt through this interface the precalculated files will not also be installed, you will need to download them separately (see below).
+
+Once these scripts are installed you'll need to add this section to *galaxy/config/tool_data_conf.xml* (or create the file if it doesn't already exist):
```
shiltemann: galaxy should do this for you if installed via toolshed, was this not the case for you?

tools/picrust/README.md
```diff
@@ -0,0 +1,36 @@
...
+    <table name="picrust_precalculated" comment_char="#">
+        <columns>name, value</columns>
+        <file path="tool-data/picrust_precalculated.loc" />
+    </table>
+</tables>
+```
+
+These xml lines point to a file (*galaxy/tool-data/picrust_precalculated.loc*) which will contain the locations to the PICRUSt precalculated files.
+
+You should download the precalculated files from http://kronos.pharmacology.dal.ca/public_files/picrust/picrust_precalculated_v1.1.1/13_5/.
```
shiltemann: note to self: we could make a data manager to do this

tools/picrust/macros.xml
```diff
@@ -31,29 +31,36 @@
...
+                <when input="output_type" value="hdf5" format="biom1"/>
             </change_format>
         </data>
     </xml>
 
     <token name="@OUTPUT_CONVERSION_COMMANDS@"><![CDATA[
-        #if $output_type == "json":
-            && biom convert -i tempbiom -o '$out_biom' --to-json
+        #if $output_type == "hdf5":
+            && cp tempbiom '$out_biom'
```
shiltemann: `mv` may be better here, save some time if files get big

shiltemann:
Thanks for the update @gavinmdouglas!

tools/picrust/README.md
gavinmdouglas: No it didn't add the section to galaxy/config/tool_data_conf.xml, I had to do that manually.

tools/picrust/macros.xml
gavinmdouglas: Ok good point, I made that change.

gavinmdouglas:
Thanks for your comments @shiltemann!

tools/picrust/README.md
shiltemann: it should be added to the file `config/shed_tool_data_table_conf.xml` which should be loaded by Galaxy alongside the file you mentioned. Is it not there either? (this works for me at least)

gavinmdouglas: Ah ok yes I see that it was added to ```config/shed_tool_data_table_conf.xml```. Does the user need to alter each of the scripts' .loc file individually then? For example, this section was added for normalize_by_copy_number.py:

```
<table comment_char="#" name="picrust_precalculated">
        <columns>name, value</columns>
        <file path="/Users/Gavin/tmp/galaxy/tool-data/toolshed.g2.bx.psu.edu/repos/iuc/picrust_normalize_by_copy_number/6273e2721fa3/picrust_precalculated.loc" />
        <tool_shed_repository>
            <tool_shed>toolshed.g2.bx.psu.edu</tool_shed>
            <repository_name>picrust_normalize_by_copy_number</repository_name>
            <repository_owner>iuc</repository_owner>
            <installed_changeset_revision>6273e2721fa3</installed_changeset_revision>
            </tool_shed_repository>
    </table>
```

However there are different table sections (i.e. different .loc files) specified for the other scripts, which seems overcomplicated.

shiltemann: no, they only need to edit one of the loc files associated with the table name (all are checked). Easiest is to stick to the one placed in the `$GALAXY_ROOT/tool-data/` folder, so I would put that in the README

tools/picrust/categorize_by_function.xml
gavinmdouglas: Ok that's good to know: I've changed the hdf5 biom format to ```h5``` and added this as a possible input format.

tools/picrust/README.md
gavinmdouglas: The issue I'm running into is that the one in ```$GALAXY_ROOT/tool-data/``` isn't pointed to by default. Instead they only seem to be pointing to files in locations like ```/Users/Gavin/tmp/galaxy/tool-data/toolshed.g2.bx.psu.edu/repos/iuc/picrust_normalize_by_copy_number/6273e2721fa3/picrust_precalculated.loc``` above. The user could just edit the path themselves, but it sounds like that shouldn't be necessary.

shiltemann: oh, that's odd, a copy of the sample .loc file is placed there for me, so when you go to the admin panel and look under `data tables registry` this location does not appear in the list there? then perhaps just phrase it to say that any one of the loc files in this list may be edited, but that there is no need to edit more than one of them (they can be reloaded from the admin panel as well to fetch any changed made btw)

gavinmdouglas: Hmm no I don't actually - this is likely something I'm doing wrong though. I've mainly been testing this repo with the planemo virtual box and I haven't figured out how to test the changes I've made to the wrappers through the toolshed GUI itself. 

That's good to hear that the sample .loc file shows up correctly on your end though! I've made your suggested edits to the README.

shiltemann:
thanks a lot for this nice contribution @gavinmdouglas!

PR #1452, opened 2017-08-29
fix heatmap2 output

mblue9:
In the heatmap2 output the rownames of the test data (the car names) are not fully visible at the right of the plot, see https://github.com/galaxyproject/tools-iuc/blob/master/tools/ggplot2/test-data/ggplot_heatmap2_result1.pdf
Made some adjustments to fix.


bgruening:
Thanks @mblue9!

PR #1459, opened 2017-09-03
pizzly - first version

mblue9:
My first pass at wrapping pizzly

Comments welcome 😄 

- pizzly has a parameter --cache that I haven't added yet as was wondering if this is something I should make a data Manager for?
 
"If the cache file exists (if you've run pizzly previously on the same GTF file), pizzly will parse this file instead, which is much faster than parsing the GTF."

- the image (logo.png) I tried to add in the tool form doesn't show up for me with planemo serve (but neither do images in other tools) so not sure if that's working 

- I could add the column names to the table output as metadata, similar to what's done in the featurecounts xml, but will it be possible to download the table with the header (the featurecounts table downloads with no header for me)? as I'd rather keep the header.

- aswell as the flatten_json.py that I've included, pizzly also has another helper script get_fragment_length.py, which examines an abundance.h5 produced by kallisto and finds the 95th percentile of the fragment length distribution. I could maybe add that as an advanced option later.


bgruening:
@mblue9 amazing! Thanks! Saves me a day or so :)

> - pizzly has a parameter --cache that I haven't added yet as was wondering if this is something I should make a data Manager for?
>"If the cache file exists (if you've run pizzly previously on the same GTF file), pizzly will parse this file instead, which is much faster than parsing the GTF."

This has just performance implications so I think it is fine for now and as you said can be extended with a data manager later on.

> - the image (logo.png) I tried to add in the tool form doesn't show up for me with planemo serve (but neither do images in other tools) so not sure if that's working 

I will look into it, but it is normal that the image is only displayed when it comes from the TS.

> - I could add the column names to the table output as metadata, similar to what's done in the featurecounts xml, but will it be possible to download the table with the header (the featurecounts table downloads with no header for me)? as I'd rather keep the header.

We don't have a official recommendation for it. I prefer the metadata one, but you are right we are missing the download option to include the header optionally.

> - aswell as the flatten_json.py that I've included, pizzly also has another helper script get_fragment_length.py, which examines an abundance.h5 produced by kallisto and finds the 95th percentile of the fragment length distribution. I could maybe add that as an advanced option later.

Sounds good!

tools/pizzly/pizzly.xml
```diff
@@ -0,0 +1,160 @@
+<tool id="pizzly" name="pizzly" version="0.0.1">
```
bgruening: 0.37.3 as version?

tools/pizzly/flatten_json.py
```diff
@@ -0,0 +1,49 @@
+import sys
```
bgruening: @mblue9 just a thought ... maybe this is also possible with `jq`? https://stedolan.github.io/jq
jq is in conda so you could just depend on it and an additional command.

tools/pizzly/pizzly.xml
mblue9: Ah I see, I thought that was a version number just for the wrapper. Have changed it to 0.37.3. So the first version of the wrapper should be the same as the version of the tool? Then you increment that version if you modify the wrapper? (e.g. add additional outputs?)

tools/pizzly/flatten_json.py
mblue9: I tried to do this with jq (see command below) but I don't know 

1) how to create the transcripts.list column as a semicolon-separated list. My jq command below has the multiple transcripts tab-separated (as additional columns)   
2) if this command is too ugly - should I stick with the flatten_json.py script? (I guess if the header was defined elsewhere, like in the metadata, I could exclude that bit)

cat out.json | jq -r  '["geneA.name","geneA.id","geneB.name","geneB.id","paircount","splitcount","transcripts.list"], (.genes[] | [.geneA.name,.geneA.id,.geneB.name,.geneB.id,.paircount,.splitcount,.transcripts[].fasta_record]) | @tsv'

mblue9:
Hi @bgruening thanks a lot for the review and replies to my questions!

I've made the changes and removed the image :)

tools/pizzly/pizzly.xml
bgruening: It is the wrapper version, my personal opinion is that these should be in sync as much as possible to not confuse the users and error reports we get.

tools/pizzly/flatten_json.py
bgruening: Let's ping here @erasche, `jq` expert. The script is also fine! Was just wondering ...

bgruening:
@mblue9 tests are still failing :(

mblue9:
Yes @bgruening I can see that :( 
Is it a Python version issue or something specific to that flatten_json script?

nsoranzo:
@mblue9 The failed Travis build was due to a few flake8 linting errors, I've fixed them for you by pushing a commit to your branch.

tools/pizzly/flatten_json.py
shiltemann: @mblue looks like you can use `join()` to make the semicolon-separate list, something like:
 
`jq -r '["geneA.name", "geneB.name", "transcripts"], (.genes[] | [.geneA.name, .geneB.name, ([.transcripts[].fasta_record]| join(";")) ])  | @tsv'`

mblue9:
Thanks a lot @nsoranzo !!

So is it that every python script submitted here (including third-party ones) need to be PEP8 compliant?

tools/pizzly/flatten_json.py
mblue9: Thanks @shiltemann that is great!! That join works! Very cool to know how to do this 😄 )

So this jq command now gives me the same result as the flatten_json python script:

cat out.json | jq -r -c '["geneA.name","geneA.id","geneB.name","geneB.id","paircount","splitcount","transcripts.list"], (.genes[] | [.geneA.name,.geneA.id,.geneB.name,.geneB.id,.paircount,.splitcount,([.transcripts[].fasta_record]| join(";"))]) | @tsv'

But should I just stick with the python script now that it's passed the tests? Would that be 'safer' because it comes with pizzly so if there's any update to the output that will get updated. Rather than add in a separate tool (jq) that's not mentioned on the pizzly site? (would that be confusing to users?) What do you think?

nsoranzo:
> So is it that every python script submitted here (including third-party ones) need to be PEP8 compliant?

Yes, exactly!

mblue9:
Ok thanks @nsoranzo I get it now :smile:

What about R scripts? Are there any rules for those?

tools/pizzly/flatten_json.py
bgruening: Ok, I must admit that I did not got that this script is from pizzly directly.
Please see here: https://github.com/bioconda/bioconda-recipes/pull/5953/files

So the new recipe has both scripts included.
Please note that it is called `pizzly_flatten_json.py`. The second one is also included.

nsoranzo:
We have only 10 R scripts at the moment in the repo (90 Python instead), so there has been no push to add an R linter. If you are interested, there is [lintr](https://github.com/jimhester/lintr).

tools/pizzly/flatten_json.py
mblue9: Great thanks @bgruening!

I guess I have to wait for this new bioconda PR that fixes the pizzly recipe then: https://github.com/bioconda/bioconda-recipes/pull/5959/files

mblue9:
Thanks @nsoranzo I'll check lintr out, as if there are standards for Python then I think it would be good to have for R too, if possible, especially if there'll be more R scripts on the way soon :smile:

tools/pizzly/flatten_json.py
bgruening: Yes.

bgruening:
@mblue9 new pizzly conda package is available afaik with all these scripts.

mblue9:
Thanks @bgruening ! I've tested that now and it works!!

But it didn't work at first, I had to manually remove the pizzly conda package I already had with: 
rm -r /home/maria/miniconda3/envs/__pizzly@0.37.3

Is that what I'm supposed to do when there's a new build of a conda package and I already have an older conda package for the same tool version? Or is there a way to automatically get the package to update with planemo test?

bgruening:
>Is that what I'm supposed to do when there's a new build of a conda package and I already have an older conda package for the same tool version? Or is there a way to automatically get the package to update with planemo test?

Yes this is expected sorry.

bgruening:
Thanks @mblue9 - you rock!

mblue9:
Oh hooray!!! My first Galaxy wrapper from scratch is in! 🎉 💃  ✨ 

Although not really from scratch as I just adapted the kallisto_quant wrapper :smile: so thanks to @davebx for making that one. And thanks to everyone here for all the reviews, comments and help!

bgruening:
@mblue9 congrats!!!!

PR #1461, opened 2017-09-04
Mothur make.contigs: use symlinks in combo mode

shiltemann:
mothur makes intermediate and/or output files at the location of its input files, use symlinks to prevent mothur polluting this location (and prevent permission denied errors when using data library files in locations not writable by Galaxy as @arbernard encountered)


bgruening:
Thanks @shiltemann!

arbernard:
Thanks for this update @shiltemann :)

PR #1463, opened 2017-09-05
Fix kallisto for pizzly

mblue9:
The kallisto Galaxy tool currently has the option to run kallisto --fusion but the additional fusion file created is not output by the wrapper. So this PR

- adds the fusion table as an output
- adds a test for the fusion table output


bgruening:
no strict version bump needed as the functionality does not change, but looks good!

PR #1465, opened 2017-09-06
Improve deseq2 PCA plot

mvdbeek:
First we keep the initial ggplot2 object returned by plotPCA,
and then add our customizations. This allows us the keep the % variance
per principal component. Then I changed the geom_text for
geom_text_repel, so that the label does not overlap with the points.


bgruening:
Nice improvement! Thanks Marius.

bgruening:
I guess we need to bump the version number before merging.

mvdbeek:
Hmm, no clue what's going on with the tests here -- they appear to have passed yesterday.

nsoranzo:
`Error: package 'GenomeInfoDb' could not be loaded` https://travis-ci.org/galaxyproject/tools-iuc/jobs/272817682#L1276

mvdbeek:
Thanks @nsoranzo, I've traced the dependency chain down to deseq2 > genomicranges > GenomeInfoDb.

mvdbeek:
hmm, but the bioconda dependency chain seems fine :(. Maybe bumping the deseq2 build works ?

mvdbeek:
and planemo is installing GenomeInfoDb https://travis-ci.org/galaxyproject/tools-iuc/jobs/272817682#L1589

bgruening:
@mvdbeek the r packages should not come from defaults anymore. Maybe this breaks things. Can you test against latest galaxy dev?

mvdbeek:
Ohh, I think you're on the right track @bgruening:
https://travis-ci.org/galaxyproject/tools-iuc/jobs/272879940#L1627
(It works on dev for me, fwiw)

mvdbeek:
So we may need a new planemo with a new galaxy-lib, I guess. I think @jmchilton was already working on that.

nsoranzo:
@mvdbeek We are already using the latest Planemo/galaxy-lib, but we also need to use the new Galaxy version because the dependencies are installed on-the-fly by Galaxy, not directly by Planemo.

bgruening:
We have no the following error:
```
There were 50 or more warnings (use warnings() to see the first 50)
Import genomic features from the file as a GRanges object ... OK
Prepare the 'metadata' data frame ... OK
Make the TxDb object ... OK
'select()' returned 1:many mapping between keys and columns
reading in files with read.delim (install 'readr' package for speed up)
1 Error: all(c(lengthCol, abundanceCol) %in% names(raw)) is not TRUE
```

mblue9:
That error is from tximport expecting the newer Sailfish/Salmon files that don't start with comment lines and that have 5 columns (the test data only have 4). I fixed it to accept the new Salmon/Sailfish output and accidentally added the changes into the dexseq PR here https://github.com/galaxyproject/tools-iuc/pull/1526. I also bumped the deseq2 to latest version 1.18.1 and made some other little changes. Sorry @mvdbeek I don't know whether I should add your deseq2 changes into my dexseq PR or whether you can add my deseq2 changes to yours (whatever you agree with)?

I also don't know if deseq2 should accept the old Salmon/Sailfish output, but if so then this option below needs to be added to the wrapper (I won't have time to do that). Btw tximport could become it's own tool at some stage as it can also be used with edgeR and limma-voom. 

From https://bioconductor.org/packages/devel/bioc/vignettes/tximport/inst/doc/tximport.html:

>Note: for previous version of Salmon or Sailfish, in which the quant.sf files start with comment lines, it is recommended to specify the  importer argument as a function which reads in the lines beginning with the header. For example:

>txi <- tximport("quant.sf", type = "none", txOut = TRUE, txIdCol = "Name", abundanceCol = "TPM", 
    countsCol = "NumReads", lengthCol = "Length", importer = function(x) read_tsv(x, 
        skip = 8))

mvdbeek:
Hmm, if your diagnosis is correct (I don't doubt it is) this means the ability to run this tools, and therefore the reproducibility has gone without a version bump in the dependencies ... that is scary!!!

mvdbeek:
Which also makes a good point for a separate tximport wrapper (or pinning the tximport version here).

mblue9:
>Hmm, if your diagnosis is correct (I don't doubt it is) this means the ability to run this tools, and therefore the reproducibility has gone without a version bump in the dependencies ... that is scary!!!

Please feel free to doubt it here or anytime (I certainly don't have a track record of being 100% correct!!!) But I totally agree, it unnerved me too to see that an underlying (and unrecorded) dependency resulted in a major change to a tool. So yes, the fact that tximport is not currently captured in the wrapper versioning is a flaw imo. I also think it would just be better to have a separate tool for tximport for the reason I said above, as it is a separate Bioconductor package that can do other things besides working with deseq2 output. I had discussed this a little previously with @bgruening and he is fine with tximport being a separate tool, it just needs someone to work on it. I could do it when I get a chance but I have no idea when that might be, so unless someone else can work on that? 

In the interim, my preference would be to have tximport listed as a requirement of the deseq2 wrapper (rather than as it currently is, kind of hidden as a dependency in the bioconda package) to at least capture that it's a dependency of this tool, along with the version.

bgruening:
Unless we don't specify every single dependency, so more or less an entire environment file, conda will not be 100% reproducible. Even with environment file there are multiple obstacles. At the beginning of our conda usage I tried to pin way more packages in our `<requirements>` section, but we agreed at some point that we should rather communicate that conda gives you some limited level of reproducibility (80%?), if you want more you should use Docker/Singularity.

It is the same for the R/Python version, this can just change without us noticing at the moment. That said @mblue9 I think pinning tximport here is a good idea.

mvdbeek:
Yeah, I know this isn't an easy subject. Pinning too strictly would lead to an explosion of installed package versions (I already had to ask for more storage on our cluster just for conda packages).

bgruening:
@mvdbeek should we increase the version number?

mvdbeek:
Yes, I'm also still fiddling with R to get the element_identifiers in to fix https://github.com/galaxyproject/tools-iuc/issues/1484

mvdbeek:
Alright, this should be ready to go!

bgruening:
Works for me. Thanks @mvdbeek!

lparsons:
Thanks @mvdbeek! However, I'm getting an error trying to run this version, installed using default conda dependency resolution.
```
Error: package or namespace load failed for 'GenomeInfoDb' in loadNamespace(i, c(lib.loc, .libPaths()), versionCheck = vI[[i]]):
 there is no package called 'GenomeInfoDbData'
Error: package 'GenomeInfoDb' could not be loaded
```

mvdbeek:
So the tests ran using conda version 4.3.33 using the channel order iuc,bioconda,conda-forge,defaults, could that be your issue ?

lparsons:
Hmmm... I suppose it's possible. What should I do to check/correct the conda setup in Galaxy?

mvdbeek:
fwiw, [GenomeInfoDb is a dependency of genomicranges](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/bioconductor-genomicranges/meta.yaml#L29) which is a dependency of deseq2.

lparsons:
Yeah, I don't really understand the error, esp since the old version runs fine. Some problem with the new mulled environment I guess. Should I uncomment and set `conda_ensure_channels` in `galaxy.ini` perhaps? And then somehow rebuild  the environment?

mvdbeek:
You can upgrade conda to 4.3.33 by activating the root conda environment and running `conda install -c conda-forge conda=4.3.33`. For the channel order you may need to uncomment https://github.com/galaxyproject/galaxy/blob/release_17.09/config/galaxy.ini.sample#L212 and set it to `iuc,bioconda,conda-forge,defaults`.

lparsons:
Beautiful! Thanks once more @mvdbeek! That solved the issue. Was I supposed to be doing things to keep everything conda working and updated?

mvdbeek:
Great, I hope you're users will be happy too!
Yeah, so given that bioconda is constantly evolving and conda itself is getting fixed/improved all the time every now and then we have to change the defaults (that could be the conda version or the channel order). 
We should communicate this clearly, because that may put you as an admin in the situation where you need to change defaults that you haven't touched. I created https://github.com/galaxyproject/galaxy/issues/5405 for this and I'm hoping we can send out an email to the mailing list every time that happens.

lparsons:
Already got at least one "thanks" from a user, so I'll pass that along to you. I totally understand about the moving pieces of conda, but yeah, putting some into the release notes would be helpful. Maybe an "admin" section of the release notes for things that an admin will need to know and possibly change with an update.

PR #1466, opened 2017-09-07
Small fixes for limma-voom output

mblue9:
This PR has two changes

- in the output html report if only one factor was input the sample check table had extra empty columns, this fixes it to have just the single factor column that it should

-  reverting a suggested change from the previous review, reverting back to using cp instead of the suggested mv as otherwise, on downloading the report or exporting the history, the tsv files are missing, so changing back from mv to: 
`cp '$outReport.files_path'/*.tsv output_dir/ `

### DIFF
--- a/tools/limma_voom/limma_voom.R
+++ b/tools/limma_voom/limma_voom.R
@@ -156,7 +156,7 @@ factPath <- as.character(argv[15])
 # Process factors
 if (as.character(argv[16])=="None") {
     factorData <- read.table(factPath, header=TRUE, sep="\t")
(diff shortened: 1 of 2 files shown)


mvdbeek:
> cp instead of the suggested mv as otherwise, on downloading the report or exporting the history, the tsv files are missing, so changing back from mv to:

Yep, if the intention is to include the tables with the html file and in the collection that's the way to do it.

bgruening:
I think we count this as bugfix and don't need a version bump.

mblue9:
Great! 

Thanks for the merge and the info!

PR #1473, opened 2017-09-13
names/count(/groups) can be used only exclusively in summary.seqs/screen.seqs

bernt-matthias:
names/count can be used only exclusively in summary.seqs
names/counts/groups can be used only exclusively in screen.seqs

Even if the PR will eliminate the error case when names/count(/groups) are input jointly the main question for me is: why doesn't mothur allow to give both (resp all three files). As far as I understand it the names/count(/groups) are updated according to removed sequences/sequence positions. It should be possible to update all three files. Or is there no scenario that requires both/ all three files.


bernt-matthias:
Just found an alternative way to implement this: see Pre.cluster: there is only one input field and the type of the data set is used for changing the command.

shiltemann:
thanks @bernt-matthias! :) 

yes, it isn't always clear which tools can take multiple of those files and update them all and which can't. For `screen.seqs` and `pre.cluster`some of the example commands on the wiki even are inputting both a name and group file, so if that it is not possible it almost seems like a bug. 

Just FYI, I am also [working on updating](https://github.com/shiltemann/tools-iuc/pull/5) the wrappers to version 1.39.5, so perhaps if some of these issues really are bugs they may have been fixed upstream (would have to check that tho)

I will look into the screen.seqs test failure when I get a chance

bernt-matthias:
OK. Then we maybe should leave it as it is: Galaxy expects the "correct" behavior. We can file an issue upstream at mothur's github..?

PR #1475, opened 2017-09-13
version 1.28.0 changed name of bayesian method to wang

bernt-matthias:
also the mothur documentation refers to wang (https://mothur.org/wiki/Classify.seqs)


bgruening:
Thanks @bernt-matthias!

PR #1476, opened 2017-09-14
Update samtools dependency for sam_fasta_index_builder dm

mvdbeek:
I've worked on testing data manager installation and table reloads in galaxy and I would like to include an integration test case that fetches a rhinovirus genome and then uses this datamanager to create an index, but samtools 0.1.19 is broken on OSX.

### DIFF
--- a/data_managers/data_manager_sam_fasta_index_builder/data_manager/data_manager_sam_fasta_index_builder.xml
+++ b/data_managers/data_manager_sam_fasta_index_builder/data_manager/data_manager_sam_fasta_index_builder.xml
@@ -1,7 +1,7 @@
 <tool id="sam_fasta_index_builder" name="SAM FASTA index" tool_type="manage_data" version="0.0.2">
     <description>builder</description>
     <requirements>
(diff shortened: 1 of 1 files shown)


yhoogstrate:
@mvdbeek would it be convenient to have the samtools version in sync with tool_collections/samtools (for example a common dependency macro) ?

mvdbeek:
It would IMO, but is planemo smart enough to handle this ?
In any case we should phase out samtools < 1, there has been so much development since then, including bug fixing and adjustments to newer standards.

yhoogstrate:
@mvdbeek planemo doesn't understand it indeed:

```
    <macros>
        <import>macro_bwa.xml</import>
        <import>macro_samtools.xml</import>
    </macros>
```
\+
```
    <command><![CDATA[
        which samtools > $output ;
        which bwa >> $output
    ]]></command>
```
=
```
--- local_file
+++ history_data
@@ -0,0 +1 @@
+/home/youri/miniconda2/envs/__samtools@1.5/bin/samtools
```

Yes please get rid of the very old samtools series. Thanks for the PR!

PR #1480, opened 2017-09-20
updated humann2 data manager and regroup

bernt-matthias:
- downloads utility function
- enable the use of the utility function again

probably this needs some improvement. since I did not work with data managers I'm not sure if this will work. It would be nice if someone could give me some links where I could start reading. In particular: how to test the data manager and the tool together.

### DIFF
--- a/data_managers/data_manager_humann2_database_downloader/data_manager/data_manager_humann2_download.py
+++ b/data_managers/data_manager_humann2_database_downloader/data_manager/data_manager_humann2_download.py
@@ -142,7 +142,9 @@ if __name__ == "__main__":
     if len(args) != 1:
         sys.stderr.write("Need to supply JSON file name")
         sys.exit(1)
-
+    if options.database not in ["utility_mapping",'humann2_utility_mapping',"chocophlan"]:
+        sys.stderr.write("Invalid data base %s"%options.database)
+        sys.exit(1)
     jsonfile = args[0]
 
     # Read the input JSON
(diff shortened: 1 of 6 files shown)


bernt-matthias:
https://bitbucket.org/biobakery/humann2/src/687004f21924e4cffb215b49d49ba29b5490e41b/humann2/tools/better_regroup_table.py?at=default&fileviewer=file-view-default there also seem to be updated versions of some scripts.

bebatut:
Looks like they are working on it... But there is no new tags or release yet.
For the regroup tool, I disable the options because in the command-line help, they remove them...

To test the data manager, I usually:
- Push it on the TTS with my account there
- Launch a Galaxy Docker container with TTS installation enable
- Install the data manager from the TTS in my instance in container
- Test it and start the loop
    - Fix it
    - Push to the TTS
    - Update on the container

I will test it also

bernt-matthias:
I'm struggling with pushing to the TTS. I have tried to follow http://planemo.readthedocs.io/en/latest/writing_appliance.html#publishing-to-the-tool-shed

- I have registered
- added the key to ~/.planemo.yml
- planemo shed_lint --tools
- changed the repository URL in .shed.yml to https://github.com/bernt-matthias/tools-iuc/tree/master/data_managers/data_manager_humann2_database_downloader
- `planemo shed_create --shed_target testtoolshed` does apparently nothing (no output and I can not find anything in the TTS)

shiltemann:
just to double-check: did you change the owner in the `.shed.yaml` file from `iuc` to your own username? and there is also nothing under "repositories I own" in TTS? (only repos with valid tools will show up in the main tool search) ..as an alternative you could also try using `planemo shed_update` with the `--force_repository_creation` parameter as well

PR #1481, opened 2017-09-21
Add Multiqc to wonderfull IUC

yvanlebras:


bgruening:
Thanks a bunch @yvanlebras!
ping @bebatut

bebatut:
Thanks @yvanlebras !
I will look at it, integrate the new tools supported by MultiQC. And maybe also the collection support :)

yvanlebras:
Thanks @bgruening @bebatut !!!!

mblue9:
This is great!!! 😃 @bebatut will you be integrating HISAT2?

bgruening:
@mblue9 the plan is to support all modules that multiQC supports.

bebatut:
@mblue9 as @bgruening mentioned, I am working to integrate all supported modules

bebatut:
I worked on it: 
- Add support the tools found on the MTS
- Extract the data stats into a collection

yvanlebras:
Great ! Nice job !!

bgruening:
Thanks a lot @bebatut for this herculean/xenalean task.
@galaxyproject/iuc could we get some help with reviewing this, please.

mblue9:
Hi @yvanlebras I don't know if this is the place to report this, but we're now getting an error when trying to install and run the current MultiQC from the toolshed and are wondering if something might have broken during this move to IUC? (This same MultiQC version v1.0.0.0 worked fine for us previously.)  
We get no errors when trying to install MultiQC (either in the logs or admin panel) but when we try to run the tool we get the error "tool_script.sh: line 25: multiqc: command not found". We're wondering if it's to do with this toolshed link that it's looking for (it doesn't exist and gives an error): toolshed.g2.bx.psu.edu/repos/engineson/multiqc/multiqc/1.0.0.0.
Some lines from the log are here: https://gist.github.com/mblue9/e41e80e33a47e83cb41f07285f23da09

tools/multiqc/generate_test_data.sh
```diff
@@ -0,0 +1,204 @@
+#!/usr/bin/env bash
+
+# 1st test
+mkdir multiqc_WDir
+
+mkdir 'multiqc_WDir/cutadapt_0'
+cp 'test-data/cutadapt.txt' 'multiqc_WDir/cutadapt_0/cutadapt.txt'
+sed -i.old 's/You are running/This is/' 'multiqc_WDir/cutadapt_0/cutadapt.txt'
```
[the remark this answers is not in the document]
nsoranzo: I suppose that's for osx compatibility.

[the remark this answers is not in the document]
peterjc: Yes, that looks like a standard form to get sed to work on both Linux and Mac OS (at the cost of leaving a backup file).

yvanlebras:
Hi @mblue9 don't know why this is not working now... I think the best is to wait for the IUC version... On the actual version, we are using an old dependency resolution through a "package_mutliqc_0_6" on the TS and it will be better to use the conda 1.2 package.... I can investigate deeper if you think it does !

tools/multiqc/multiqc.xml
```diff
@@ -0,0 +1,953 @@
...
+                <element name="general_stats" file="aligner_soft_stats.tabular" compare="sim_size"/>
+                <element name="htseq" file="htseq_stats.tabular" compare="sim_size"/>
+                <element name="picard_AlignmentSummaryMetrics" file="picard_AlignmentSummaryMetrics_stats.tabular" compare="sim_size"/>
+                <element name="picard_RnaSeqMetrics" file="picard_RnaSeqMetrics_stats.tabular" compare="sim_size"/>
+                <element name="picard_baseContent" file="picard_baseContent_stats.tabular" compare="sim_size"/>
+                <element name="picard_dups" file="picard_dups_stats.tabular" compare="sim_size"/>
+                <element name="picard_insertSize" file="picard_insertSize_stats.tabular" compare="sim_size"/>
+                <element name="prokka" file="prokka_stats.tabular" compare="sim_size"/>
+                <element name="quast" file="quast_stats.tabular" compare="sim_size"/>
+                <element name="samblaster" file="samblaster_stats.tabular" compare="sim_size"/>
```
shiltemann: is `compare="sim_size"` really needed for all of these? 

not entirely sure what the default delta is, but this file for instance is only 67 bytes to start with, which could already be less than the default delta, maybe exact matches can work for some of these too?

tools/multiqc/multiqc.xml
```diff
@@ -0,0 +1,953 @@
...
+                <when value="featureCounts">
+                    <param name="input" type="data" format="txt,tabular,tsv,csv" multiple="true" label="Output of FeatureCounts"/>
+                </when>
+                <when value="flexbar">
+                    <param name="input" type="data" format="txt,tabular,tsv,csv" multiple="true" label="Output of Flexbar" help="It should contain 'Flexbar - flexible barcode and adapter removal'"/>
+                </when>
+                <when value="gatk">
+                    <repeat name="output" title="GATK output" min="1">
+                        <param name="type" type="select" label="Type of GATK output?">
+                            <option value="varianteval">Variant evol file</option>
```
shiltemann: s/evol/eval/ ?

tools/multiqc/.shed.yml
```diff
@@ -0,0 +1,12 @@
...
+description: MultiQC aggregates results from bioinformatics analyses across many samples into a single repor
+homepage_url: http://multiqc.info/
+long_description: |
+  MultiQC searches a given directory for analysis logs and compiles a HTML report. 
+  It's a general use tool, perfect for summarising the output from numerous bioinformatics tools.
+remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/master/tools/multiqc
+type: unrestricted
+categories:
+- Fastq Manipulation
+- Statistics
```
shiltemann: add "Visualization" category?

tools/multiqc/multiqc.xml
bebatut: The order of the columns changes between the local execution and the one on Galaxy. I am testing if it is changing with a delta = 0

shiltemann: oh hmm, well otherwise just some `assert contents` on column headers and some values orso? ..but you might be testing nothing now

bgruening:
Thanks @yvanlebras, @bebatut and @shiltemann!
ping @lecorguille

bgruening:
@bebatut do you want to tweet this?

lecorguille:
GG!
We should also ping @ewels :)

I will be able to use it in the W4M project :)

ewels:
Wow, amazing work everyone! This is brilliant 😀 

If someone fancies adding a PR for a new section in the [MultiQC documentation](http://multiqc.info/docs/#installing-multiqc) (source [here](https://github.com/ewels/MultiQC/blob/master/docs/installation.md)) that would be fab. I'm a complete n00b when it comes to Galaxy so I don't really feel qualified to write that!

Phil

yvanlebras:
Hi @ewels I can write something to begin. Do you want this on the installation doc ? For example something like:

# Using MultiQC through Galaxy

## On the main Galaxy instance

## On your instance

ewels:
Great! Yes that's pretty much what I was thinking - at the end of the installation docs probably makes most sense.

yvanlebras:
ok, I have PR a first try

tools/multiqc/multiqc.xml
```diff
@@ -1,429 +1,992 @@
...
-                <when value="samtools_idxstats" />
-                <when value="picard" />
-                <when value="bismark" />
-                <when value="bowtie2" />
-                <when value="htseq" />
-                <when value="rnastar_log" />
-                <when value="rnastar_counts" />
-                <when value="custom_content">
-                    <conditional name="cc_cond">
-                        <param name="cc_select" type="select" label="MultiQC configuration" help="The configuration set the graph title and axis labels">
```
lecorguille: @bebatut 
Bouh ... why did you remove this select? 😭 
I allowed that because tools (like mine) can ship the MultiQC configuration within the data like in this [example](https://github.com/galaxyproject/tools-iuc/blob/master/tools/multiqc/test-data/cc_ko15.bpc.embedded.tab#L1). That way the naive user just have to add datasets without being annoyed about all those details.

PR #1489, opened 2017-09-26
More fixes to CrossMap

yhoogstrate:
I can't test so far since my planemo crashes with `CrossMap.py: command not found`

- [x] Make test using loc file
- [ ] See if reverting SF links passes linting

PR #1490, opened 2017-09-27
changed all_fasta.loc to all_trans.loc

mblue9:
- renamed the pizzly transcriptome fasta loc file as all_trans.loc to avoid clashing with the already existing all_fasta.loc used for genomes 
- added a test for using a built-in fasta and gtf
- replaced in_range with min validators with just param min


tools/pizzly/pizzly.xml
```diff
@@ -80,20 +80,14 @@
...
-            <param name="insert_size" argument="--insert-size" type="integer" value="400" label="Maximum insert size of the paired-end library" help="kallisto will estimate this from the reads, default: 400">
-                <validator type="in_range" min="0" />
-            </param>  
-            <param name="align_score" argument="--align-score" type="integer" value="2" label="Number of mismatches allowed" help="default: 2">
-                <validator type="in_range" min="0" />
-            </param> 
-            <param name="kmer_size" argument="--k" type="integer" value="31" label="K-mer size used to create the kallisto index" help="default: 31">
-                <validator type="in_range" min="0" />
-            </param> 
+            <param name="insert_size" argument="--insert-size" type="integer" value="400" label="Maximum insert size of the paired-end library" help="kallisto will estimate this from the reads, default: 400" />
```
bgruening: is the min here missing?

tools/pizzly/pizzly.xml
```diff
@@ -5,16 +5,16 @@
     </requirements>
     <version_command>pizzly --version</version_command>
     <command detect_errors="exit_code"><![CDATA[
-
```
bgruening: We should bump the version number of the tool here.

mvdbeek:
Do we really want to maintain an additional table for non-genome things ?
Personally I am not favorable towards this. The dbkey, name and display name can all indicate whether or not a fasta file belongs to a genome or a transcriptome. If we add another table for disinguising genome and transcriptome (a distinction which is purely "metadata" in my opinion). This would lead to a large expansion of tables to be maintained, since you may also want to have e.g bowtie2/bwa indexes for transcriptomes and so on. I'd rather see an additional optional column that can carry this metadata.

yhoogstrate:
Is the all_gtf not the equivalent of gene_sets.loc (featureCounts) ?

mblue9:
Thanks for the comments! I've made those fixes to the wrapper version and missing min. 
I also got rid of the transcriptome loc table and for the moment have reverted back to using the all_fasta table (where the build id and display name would flag the fasta file as transcriptome) but should I add an extra column like you suggested @mvdbeek?

We were wondering if having both genomes and transcriptomes in the same all_fasta file could result in a long, confusing dropdown in Galaxies that support a lot of species? As there can be multiple transcriptomes for a species e.g. for different Ensembl versions and so we were wondering if it might be better instead for transcriptomes and gtfs to have loc files for specific _versions_ e.g ensembl v90? 

@yhoogstrate yes all_gtf is the equivalent of gene_sets.loc. It's also the equivalent of the bedtools all_gff.loc.  Because there were already 2 loc files for our same set of gtfs anyway I thought it didn't really matter what I called the table 😕 . For us "gene sets" means something different to gtfs, it means lists of genes like those used by MSigDB so I thought all_gtf was a better name. (I have never done any admin of data tables and I didn't know that it's important to use the same name if there's a loc already existing for the data type, and that you shouldn't use a name that's already being used by a tool for a different data type 😕  is there any documentation on that?). I've changed the loc file name to gene_sets now to use the already existing featurecounts loc.

mblue9:
The Travis test looks like it's failing because an old pizzly conda build is being used, one that doesn't have the pizzly_flatten_json.py script. This worked before and passes locally. What should I do?

nsoranzo:
> I also got rid of the transcriptome loc table and for the moment have reverted back to using the all_fasta table (where the build id and display name would flag the fasta file as transcriptome) but should I add an extra column like you suggested @mvdbeek?

I think this may break existing tools/data managers.

> yes all_gtf is the equivalent of gene_sets.loc. It's also the equivalent of the bedtools all_gff.loc.

GFF and GTF are (slightly and irritatingly) different formats.

> I thought all_gtf was a better name.

Agreed, but I suppose that's too late since the `gene_sets` table has been around since November 2016.

> The Travis test looks like it's failing because an old pizzly conda build is being used, one that doesn't have the pizzly_flatten_json.py script. This worked before and passes locally. What should I do?

I think this may be caused by the wrong channel order which Galaxy 17.05 uses. If this is the case, https://github.com/galaxyproject/tools-iuc/pull/1489/commits/14017ddfde1c2450a2d4f0a22a510ea113394fb5 would solve it when the https://github.com/galaxyproject/tools-iuc/pull/1489 is merged (and this PR rebased), but maybe we should fix 17.05 by backporting https://github.com/galaxyproject/galaxy/pull/4409 ?

mblue9:
Thanks also for this review @nsoranzo!

> GFF and GTF are (slightly and irritatingly) different formats.

Yes I know but GTFs and GFFs are used interchangeably by both bedtools and featurecounts so the all_gff.loc and gene_sets.loc are exactly the same afaik just different names.

nsoranzo:
xref. https://github.com/galaxyproject/galaxy/pull/4731

nsoranzo:
@mblue9 @bgruening Even with the correct channel order now available in Galaxy 17.05, the old pizzly build gets installed:
```
$ conda --version
conda 4.3.24
$ more ~/.condarc 
channels: [iuc, bioconda, conda-forge, defaults, r]
$ conda create -n __pizzly@0.37.3 pizzly==0.37.3
Fetching package metadata .................
Solving package specifications: .

Package plan for installation in environment /usr/users/ga002/soranzon/miniconda3/envs/__pizzly@0.37.3:

The following NEW packages will be INSTALLED:

    pizzly: 0.37.3-0 bioconda

Proceed ([y]/n)?
```

instead of `pizzly-0.37.3-py35_2` .

mblue9:
@nsoranzo @bgruening I really don't understand this failure. Why is Travis using the old pizzly build now when it used the new build correctly in the previous pizzly PR here https://travis-ci.org/galaxyproject/tools-iuc/jobs/275214906  😕

bgruening:
Not sure, but I labelled the old ones as broken and now it is green. @nsoranzo ready to be merged?

nsoranzo:
Thanks @mblue9, @bgruening and all!

PR #1497, opened 2017-09-29
fix hisat2 extra outputs

mblue9:
Sorry to be bringing hisat2 back again! but I just discovered that I accidentally left in a bracket that means that it's producing outputs that should be filtered out 😮 I left in the last bracket here: 

`<filter>(library['type'] == 'paired' or library['type'] == 'paired_collection') and (adv['output_options']['output_options_selector'] == "advanced" and adv['output_options']['unaligned_file'] is True) </filter>`

I fixed it here. I also learnt that I can include <test expect_num_outputs="num"> to check that the correct number of outputs are produced so I've added that in.

(a slightly weird thing to me is that nothing complained about the stray bracket, not planemo lint, test or travis. I was wondering if something should or if I really do need to get myself some better 👀!)

### DIFF
--- a/tools/hisat2/hisat2.xml
+++ b/tools/hisat2/hisat2.xml
@@ -290,11 +290,9 @@ hisat2
     --summary-file summary.txt
 #end if
 
-
 ## Convert SAM output to sorted BAM
 
 | samtools sort - -@ \${GALAXY_SLOTS:-1} -l 6 -o '${output_alignments}'
-
 
 ## Rename any output fastq files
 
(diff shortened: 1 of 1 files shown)


tools/hisat2/hisat2.xml
```diff
@@ -567,23 +565,25 @@ hisat2
 
     <!-- Define tests -->
 
-    <tests>
-        <test><!-- Ensure bam output works -->
+    <tests><!-- Ensure bam output works -->
```
nsoranzo:
```xml
    <tests>
        <!-- Ensure bam output works -->
```

nsoranzo:
> I also learnt that I can include to check that the correct number of outputs are produced so I've added that in.

Good idea, thanks!

> (a slightly weird thing to me is that nothing complained about the stray bracket, not planemo lint, test or travis. I was wondering if something should or if I really do need to get myself some better :eyes:!)

Unfortunately the behaviour of output filter is very weird: the expression is evaluated as Python code and if the result is True or an exception is raised, the filter is considered "passed" and the output dataset is added to the history.

mblue9:
Thanks for the review and info @nsoranzo! I've fixed that comment position.

nsoranzo:
Thanks @mblue9!

PR #1509, opened 2017-10-09
mothur pcr.seqs: also allow unaligned fasta input

### LINKED ISSUE #1508: Mothur: Pcr.seqs on non align fasta files
HI,

I have a small problem with the tools pcr.seqs. Actually, it works only on mothur.align input, but this is not necessary for trim by oligos, because simple fasta file could be used for that.

Can you add fasta file as valid format input when oligos trim is choice.
Similarly, the ouput format must be the same as the input: fasta or mothur.align.

Best regards,

Benoit

shiltemann:
fixes #1508

### DIFF
--- a/tools/mothur/pcr.seqs.xml
+++ b/tools/mothur/pcr.seqs.xml
@@ -54,7 +54,7 @@
         | tee mothur.out.log
     ]]></command>
     <inputs>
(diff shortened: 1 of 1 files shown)


bgruening:
Ok, this does not make it worse and fixes an important bug. Nevertheless, we should tackle the failing tests in the other open mothur PRs.

PR #1510, opened 2017-10-09
mothur: make.shared cat logs instead of move

bernt-matthias:
Make.shared seems to produce multiple log files in the working directory: e.g. 
`mothur.1507547483.logfile  mothur.1507547484.logfile  mothur.out.log`. Hence `mv *log outfile` gives an error: 

```
Fatal error: Exit code 1 ()
mv: target `/gpfs1/data/galaxy_server/galaxy/jobs_dir/001/1248/galaxy_dataset_2683.dat' is not a directory
```

Using cat instead works. 

Question is: is make.shared supposed to produce more than one log file? And maybe: are we interested in all?

### DIFF
--- a/tools/mothur/make.shared.xml
+++ b/tools/mothur/make.shared.xml
@@ -38,7 +38,6 @@
         | tee mothur.out.log &&
 
         ## move output files to correct destination
-	cat mothur.*.logfile > "$logfile" &&
         #if $intype.infile == 'otulist' and $intype.groups:
             mv intype_otu*.groups "$groupout"
         #else
--- a/tools/mothur/make.shared.xml
+++ b/tools/mothur/make.shared.xml
@@ -38,7 +38,7 @@
         | tee mothur.out.log &&
 
         ## move output files to correct destination
-        mv mothur.*.logfile "$logfile" &&
+	cat mothur.*.logfile > "$logfile" &&
         #if $intype.infile == 'otulist' and $intype.groups:
             mv intype_otu*.groups "$groupout"
         #else


shiltemann:
Hi @bernt-matthias hmm, yes, I've noticed some mothur components make multiple logfiles under certain parameter settings, often when they internally call another mothur component. I got rid of this `mv` command in all wrappers I thought, but guess I missed this one, will update, thanks for reporting :)

shiltemann:
sorry, didn't notice this was a PR at first and not an issue, could you change it to use `tee` and `from_work_dir` like the other wrappers? that way everything mothur outputs gets put in the logfile, even if this is multiple tool logs.

bernt-matthias:
I investigated this a bit more: 

- `tee` and `from_work_dir` are already used
- mothur creates a logfile (e.g mothur.1507541228.logfile) even if `tee` is used, i.e. nearly equivalent log files `mothur.out.log` and `mothur.1507541228.logfile` are created: 

```
diff mothur.1507624160.logfile mothur.out.log
1c1,3
< Linux version
---
> 

> 
> 
3d4
< Using ReadLine
5d5
< Running 64Bit Version
26d25
< Interactive Mode
28a28
> 
42d41
< quit
```

- also the call to `help()` which is used to get the version creates a log file 

So I would suggest to use only the mothur.out.log, i.e., just remove the `mv` command completely.

shiltemann:
@bernt-matthias yes indeed, that line should have been removed when I switched to using `tee`, thanks for spotting it :) ..I will look into these travis errors, they appear unrelated to your changes

bgruening:
Unrelated travis errors, but we need to look at them.

PR #1512, opened 2017-10-11
Allow STAR to use (some) --limit options

### LINKED ISSUE #1511: STAR with the --limitBAMsortRAM option
It turns out that if you align against a small genome (e.g., dm6), the amount of memory that STAR allocates for BAM sorting might not always suffice. The `--limitBAMsortRAM` parameter can be used to set this, but it's not available in the Galaxy wrapper. It might be nice to add this so (with a default 0 and buried in the advanced settings) for those circumstances when the default needs to be tweaked.

For reference, the error message produced when the default doesn't work is some variant of:

    EXITING because of fatal ERROR: not enough memory for BAM sorting: 
    SOLUTION: re-run STAR with at least --limitBAMsortRAM 4815221017
    Oct 10 14:22:48 ...... FATAL ERROR, exiting

dpryan79:
This fixes #1511 and adds a couple other `--limit*` options that people might occasionally want to tweak. The defaults in the wrapper match those from the command line.

### DIFF
--- a/tools/rgrnastar/rg_rnaStar.xml
+++ b/tools/rgrnastar/rg_rnaStar.xml
@@ -1,4 +1,4 @@
-<tool id="rna_star" name="RNA STAR" version="2.5.2b-0" profile="17.01">
+<tool id="rna_star" name="RNA STAR" version="2.5.2b-1" profile="17.01">
     <description>Gapped-read mapper for RNA-seq data</description>
     <macros>
         <import>macros.xml</import>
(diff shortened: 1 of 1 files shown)


dpryan79:
The error is simply due to database locking and can be ignored.

bgruening:
Can you increase the tool version, please?

dpryan79:
@bgruening Sure, done.

tools/rgrnastar/rg_rnaStar.xml
```diff
@@ -159,6 +159,13 @@
                 --chimScoreJunctionNonGTAG "$params.chim.chimScoreJunctionNonGTAG"
                 --chimJunctionOverhangMin "$params.chim.chimJunctionOverhangMin"
             #end if
+
+            ## Limits
+            --limitBAMsortRAM "$params.limits.limitBAMsortRAM"
```
bgruening: @dpryan79 are the quotes needed? They are all integers right?

dpryan79: They're not needed, I was just going for consistency with all of the other params.

bgruening:
Thanks @dpryan79!

PR #636, opened 2016-03-03
Update intersectBed.xml

pavanvidem:
there is no parameter called "modes"

PR #1513, opened 2017-10-12
Port the DAVID tool to tools-iuc

pvanheus:
This is one of the tools mentioned in  #636 and mentioned in the [RNA Seq tutorial](https://galaxyproject.github.io/training-material/topics/transcriptomics/tutorials/ref-based/tutorial.html). It should be a simple tool to merge with no dependencies.


pvanheus:
Tool is not currently working properly and test didn't pick it up.

mvdbeek:
@pvanheus DAVID is made for microarray data. This should not be used for RNAseq data ... I'd rather see this removed from the tutorial. Valid alternatives are goseq (already wrapped in IUC), GOglm, RNA-Enrich ...

nsoranzo:
xref https://github.com/galaxyproject/training-material/issues/531

Issue #1322, opened 2017-05-23
Stringtie 1.3.3 errors when the option to output Deseq2/EdgeR is used


jennaj:
Switching to no additional output or to Ballgown works fine. All inputs appear to meet specification and this is reproducible across different input datasets/target genomes.

**Error produced:** 

```
Fatal error: Exit code 1 ()
Traceback (most recent call last):
  File "/galaxy/main/deps/_conda/envs/__stringtie@1.3.3/bin/prepDE.py", line 186, in <module>
    t_dict.setdefault(t_id, {})
NameError: name 't_id' is not defined
```

**Example job info**

```
StringTie
Dataset Information
Number:	XX
Name:	StringTie on data X and data X: Assembled transcripts
Created:	Fri 19 May 2017 08:06:59 PM (UTC)
Filesize:	40.5 MB
Dbkey:	mm10
Format:	gtf
Job Information
Galaxy Tool ID:	toolshed.g2.bx.psu.edu/repos/iuc/stringtie/stringtie/1.3.3
Galaxy Tool Version:	1.3.3
Tool Version:	1.3.3
Tool Standard Output:	stdout
Tool Standard Error:	stderr
Tool Exit Code:	1
History Content API ID:	X
Job API ID:	X
History API ID:	X
UUID:	X
Full Path:	/galaxy-repl/main/files/X/X/dataset_X.dat

Tool Parameters
Input Parameter	Value	Note for rerun
Mapped reads to assemble transcripts from	X: SortSam on data X: BAM sorted in coordinate order	
Use GFF file to guide assembly	yes	
Reference annotation to use for guiding the assembly process	X: UCSC Main on Mouse: knownGene (genome)	
Perform abundance estimation only of input transcripts	False	
Output additional files for use in...	deseq2	
Average read length	75	
Whether to cluster genes that overlap with different gene IDs	False	
Options	default	
Job Resource Parameters	no	

Inheritance Chain
StringTie on data X and data X: Assembled transcripts

Command Line
[...]

PR #1517, opened 2017-10-13
fix for #1322

NCEichner:
Thanks @mblue9 for the pdfs. So here is my current version of the updated stringtie.xml that should fix the error described in #1322.
Compared to my last post I've added some 'cosmetic' changes regarding the temp-files. Please test if these changes work for you.

Maybe some last idea:
If someone has a project with several samples the outputs of the stringtie/prepDE combination have to be merged before feeding the table to DESeq. This means a lot of clicks.
prepDE.py is designed to handle .gtf input files from several samples and create one single table in which the samples are represented in the columns and the genes/transcripts are shown in the rows.
In order to use the full power of this tool it should get an own .xml-wrapper as the medium-term consideration.

### DIFF
--- a/tools/stringtie/stringtie.xml
+++ b/tools/stringtie/stringtie.xml
@@ -8,9 +8,6 @@
     <expand macro="version_command" />
     <command><![CDATA[
         mkdir -p ./special_de_output/sample1/ &&
-        #if str($guide.use_guide) == 'yes':
-            ln -s '$guide.guide_gff' ./special_de_output/sample1/guide.gtf &&
-        #end if
 
         #if $input_bam.metadata.ftype == 'sam':
             samtools sort -@ \${GALAXY_SLOTS:-1} '$input_bam' | stringtie
@@ -47,11 +44,13 @@
 
         #if str($guide.use_guide) == 'yes':
             #if $guide.special_outputs.special_outputs_select == 'deseq2':
+                && 
+                ln -s '$output_gtf' ./special_de_output/sample1/output.gtf
                 &&
                 prepDE.py
                     -i ./special_de_output/
-                    -g gene_cout_matrix.tsv
-                    -t transcripts_count_matrix.tsv
+                    -g ./gene_counts.tsv
+                    -t ./transcript_counts.tsv
                     -l $guide.special_outputs.read_length
                     #if str($option_set.options) == 'advanced':
                         -s '$option_set.name_prefix'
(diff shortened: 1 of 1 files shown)


mblue9:
Great that you made these changes @NCEichner! (fixing this output will be handy for the new edgeR 😄 ). I'm not familiar with Stringtie but I tried running your changes and all the tests pass for me with planemo serve except Test No. 6 - the output of transcripts_counts.tsv is now different (CUFF.1.1	181 instead of CUFF.1.1 574). Did that test pass for you as wondering if that output file needs updating?
 
And here Travis is failing on two tests with "tool_script.sh: line 25: prepDE.py: command not found" @nsoranzo @bgruening  is this that conda channel issue again?

tools/stringtie/stringtie.xml
```diff
@@ -47,27 +44,39 @@
...
+                sed -i.bak 's/\r//g' ./transcript_counts.tsv
+                &&
+                mv -T ./transcript_counts.tsv "$transcript_counts"
+                &&
+                sed -i.bak 's/,/\t/g' ./gene_counts.tsv
+                &&
+                sed -i.bak 's/\r//g' ./gene_counts.tsv
                 &&
-                sed -i.bak 's/,/\t/g' gene_cout_matrix.tsv
+                mv -T ./gene_counts.tsv "$gene_counts"
```
yhoogstrate: `-T` is not posix standard (http://pubs.opengroup.org/onlinepubs/9699919799/utilities/mv.html) and I think not included in OSX (https://www.unix.com/man-page/osx/1/mv/)

NCEichner:
Thanks @mblue9 and @yhoogstrate for testing. Because maintaining our local galaxy-server is seen as a 'side project' I didn't find the time to set up an clean planemo testsystem for making these tests myself. At the moment if one user is reporting a problem I only can try to 'hotfix' this by patching the .py or .xml files on the productive server when Galaxy is not in use.
And because I'm not familiar with the planemo test routines I cannot tell anything about Test No. 6 - sorry.

But there is one strange thing with the prepDE output: Depending on wether I run DESeq2 output with or w/o the clustering option the 'gene counts' file appears with an additional third empty column or not. When downloading this file it contains only the expected two columns. And if I use the 'Cut' tool in galaxy keeping the first two columns everything also seems to be fine again. Did anyone see this, too? Maybe this is a prepDE issue?
 
@yhoogstrate 
Thanks for your hint. I wasn't aware that the wrapper is not assembling just simple linux commands. This means maybe more reading the programming guidelines. I'm running Galaxy on a Linux system and somehow my posted version of 'stringtie.xml' didn't break our system... ;-)
In general I wanted to present the files to the galaxy history just AFTER all the replacements from .csv to .tsv are accomplished. That's why I used the 'mv' command. After looking up your referenced docs I dropped the '-T' option in the 'mv' commands. Testing on our productive system worked fine for me.... ;-)

PS.: I also checked the 'sed' command for compatibility. The '-i' option seems to be valid for OSX and Linux but it's not listed in the posix version. Now I kept this in my new commit.

tools/stringtie/stringtie.xml
```diff
@@ -47,27 +44,39 @@
...
+                sed -i.bak 's/\r//g' ./transcript_counts.tsv
+                &&
+                mv ./transcript_counts.tsv "$transcript_counts"
+                &&
+                sed -i.bak 's/,/\t/g' ./gene_counts.tsv
+                &&
+                sed -i.bak 's/\r//g' ./gene_counts.tsv
                 &&
-                sed -i.bak 's/,/\t/g' gene_cout_matrix.tsv
+                mv ./gene_counts.tsv "$gene_counts"
```
nsoranzo: You can remove the `mv` command if you use `from_work_dir="gene_counts.tsv"` in `<data name="gene_counts"` , this is the preferred way if your output has a specific filename.
Similarly for `transcript_counts` and `legend`.

mblue9:
Hi @NCEichner  I took a look at that failed test (test 6) and it looks like the test data did need updating so I updated it and it passes now. I also made some more changes so ended up making a new PR here https://github.com/galaxyproject/tools-iuc/pull/1545 that has your changes and mine. Would be good to know what you think of the new changes (or anyone else)? I couldn't see that mysterious 3rd column with the clustering so if you could check if you still get it with these changes that would be great.

PR #1519, opened 2017-10-16
DexSeq and get_length_and_gc_content update of requirement versions

bernt-matthias:
The version update is necessary to be able to install the DexSeq package at all. 

With the old version I run in several conda / bioconda related problems which make it impossible to install: 

- conda is unable to resolve the dependencies (https://github.com/conda/conda/issues/5536). I aborted it after 3days. With the new version the installation is successful after minutes. 
- some dependencies imply R 3.3.1 which seems impossible to install (https://github.com/conda/conda/issues/6175). I do not understand the explanation in the issue discussion, but the update solves the problem. 

Problems: 
- The DexSeq test yields different results. The question is: is it sufficient to update the test-data?

### DIFF
--- a/tools/dexseq/dexseq.xml
+++ b/tools/dexseq/dexseq.xml
@@ -1,7 +1,7 @@
 <tool id="dexseq" name="DEXSeq" version="1.20.1">
     <description>Determines differential exon usage from count tables</description>
     <requirements>
-        <requirement type="package" version="1.20.1">bioconductor-dexseq</requirement>
+        <requirement type="package" version="1.22.0">bioconductor-dexseq</requirement>
         <requirement type="package" version="1.20.0">r-getopt</requirement>
         <requirement type="package" version="0.2.15">r-rjson</requirement>
     </requirements>
--- a/tools/dexseq/dexseq_count.xml
+++ b/tools/dexseq/dexseq_count.xml
@@ -1,7 +1,7 @@
 <tool id="dexseq_count" name="DEXSeq-Count" version="1.20.1">
     <description>Prepare and count exon abundancies from RNA-seq data</description>
     <requirements>
-        <requirement type="package" version="1.20.1">bioconductor-dexseq</requirement>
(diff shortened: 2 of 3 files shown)


bernt-matthias:
The second problem occurred also for get_length_and_gc_content. All versions except r-optparse (this led to a conflict with r-base) are now the latest. For this tool the tests were successful.

mvdbeek:
> The question is: is it sufficient to update the test-data?

It should be ... I think that's always a possibility when you update DE tools, which may change results slightly (there may be platform specific rounding errors on top of that). As long as the output looks sensible that wold be OK.

bernt-matthias:
One question that came into my mind is reproducibility: Here we have a tool with dependencies, both in specific versions, that is not installable anymore with a current version of the installer. If Galaxy wants real reproducibility should we rather fix the problems with the conda packages (and the installer) such that the current version of the tool is functional? This holds in particular if results change due to the new versions. 

I discussed with @bgruening in another thread about updating conda packages: he hoped that it won't happen to often. But in the long run it will be inevitable -- I guess.

Still, the PR is fine having new software versions (and a small fix) seems good -- I could not find out if there is new functionality in the dexseq package.

mblue9:
@bernt-matthias I had started updating the length and gc content tool for goseq and had updated the requirements the same as you here :smile: I also added in the ability to use the in-built gtfs (gene_sets.loc) but I guess I need to wait to submit a PR with those changes until this one of yours is accepted.

bgruening:
@bernt-matthias sorry for the trouble. We try our best to make Bioconda better, but we depend on other player like conda-forge and defaults. That can at some point not play nicely. However, we wrote a small paper describing the advantages and disadvantages of different systems and the impact on reproducibility here: https://www.biorxiv.org/content/early/2017/10/11/200683

bernt-matthias:
@mblue9 you could incorporate my changes into your PR. Then we could close this one.

mblue9:
ok @bernt-matthias I've added your changes here https://github.com/galaxyproject/tools-iuc/pull/1526

PR #1520, opened 2017-10-16
Fix g2 rsync data manager.

blankenberg:
There are existing newerish Galaxy framework bugs preventing this data manager from working:
 - https://github.com/galaxyproject/galaxy/issues/4810
 - https://github.com/galaxyproject/galaxy/issues/4809


bgruening:
LGTM, but haven't tested it. @blankenberg you have tested it I assume, right?

blankenberg:
Yes, I did some testing with the all_fasta and fasta_indexes tables, which worked well for e.g. sacCer2. It is definitely not *more* broken than it was ;)

bgruening:
In Dan we trust ;)

PR #1522, opened 2017-10-17
FLASH: Add phred offset parameter.

davebx:
Also only allow sanger, illumina, and solexa input fastq files, so that the wrapper will correctly determine the appropriate phred offset.


blankenberg:
Wont this break workflow usage for the now discovered datasets? Why not just change the output format based upon the input format?

davebx:
@blankenberg I wasn't sure if the change format from input format would work with the dataset/collection conditional, with one set being "forward/reverse" and the other being "reads". I can try to find a solution for that, though.

davebx:
Adding WIP tag so this doesn't get merged with a broken test.

tools/flash/flash.xml
```diff
@@ -43,51 +50,82 @@
...
             <output name="hist_in" file="flash_hist_in_out2.tabular" />
             <output name="histogram_in" file="flash_hist_in_out2.txt" />
             <output name="hist_out" file="flash_hist_out_out2.tabular" />
             <output name="histogram_out" file="flash_hist_out_out2.txt" />
         </test>
+        <test>
+            <param name="select_layout" value="collection" />
+            <param name="generate_histogram" value="false" />
+            <param name="reads">
+                <collection type="paired" ftype="fastqillumina">
```
nsoranzo: `ftype` is not allowed here, planemo lint complains.

nsoranzo:
I've removed the WIP from the title following your removal of the tag, hope that's OK.

tools/flash/flash.xml
davebx: Ah, I wasn't sure, because it is possible to specify the datatype when creating a collection.

tools/flash/flash.xml
```diff
@@ -17,23 +17,30 @@
...
         <conditional name="layout">
-            <param name="select_layout" type="select">
+            <param name="select_layout" type="select" label="Input data structure">
                 <option value="individual">Individual datasets</option>
                 <option value="collection">Paired collection</option>
             </param>
             <when value="individual">
-                <param format="fastq" name="forward" type="data" label="Forward reads" />
-                <param format="fastq" name="reverse" type="data" label="Reverse reads" />
+                <param format="fastqsanger,fastqsolexa,fastqillumina" name="forward" type="data" label="Forward reads" />
```
bgruening: should fastq be also in this list

davebx: From @blankenberg's email to me:
> Of course solexa is not Phred scaled, but flash isn't able to specify a non-phred scaled data, and with its low use, it is probably close enough, or we can exclude it. *base fastq should probably not be allowed*
(emphasis mine)

nsoranzo: `fastq` would let select also `fastqcssanger`, which I image is not supported.

davebx: I shudder to think how flash would react to a color-space fastq file.

bgruening:
Besides the fastq question, which I think is from a usability point of view not optimal, this LGTM.
Maybe we can include an option to prevent the output of non-merged reads, but this is not super-urgent.

@nsoranzo ok to merge?

tools/flash/flash.xml
nsoranzo: And fastqsanger is default for uploads in 17.09 anyway.

bgruening: Ok, thats a point.

nsoranzo:
LGTM

bgruening:
I understand that you need this tool today, so lets merge it. Can you put on your todo list to add a flag to not output the none-merged outputs?
Thanks.

davebx:
@bgruening will do, thanks for the merge

PR #1523, opened 2017-10-18
update goseq

mblue9:
This PR contains some suggested updates for goseq, wondering what people think of these? Am especially wondering what people think about adding in the ability to get GO and KEGG categories? I think it would be useful but you have to add the org. package for each species. Here I added 4 to test (human, mouse, fruitfly and zebrafish). We only use those 4 species at the moment so this functionality would work well for us :), but is this ok and what about other species, would adding multiple org packages make this tool package too big?) 

- add ability to get GO and KEGG categories using goseq's getgo function.
- add test for retrieving GO terms for human and a second genome (zebrafish)
- add tests for other outputs: length bias plot, sampling vs wallenius plot, sampling table, hypergeometric table
- update requirement versions (for bioconductor-org.xx.eg.db packages I added the versions available here: https://depot.galaxyproject.org/software/)
- add more help/explanation
- reorder layout of inputs
- increase wrapper version


mvdbeek:
I'd prefer to have a separate tool for getting the go categories (but maybe I'm biased because that's how I thought about this initially). This might also be nicer for reproducibility, since you would have the categories that were used in your history.
Another issue is that the db packages are pretty big IIRC, so people that use non-model organisms would have a lot of useless stuff installed. The rest looks really good!

mblue9:
Thanks a lot for the (very quick!) review @mvdbeek !!

> I'd prefer to have a separate tool for getting the go categories (but maybe I'm biased because that's how I thought about this initially). This might also be nicer for reproducibility, since you would have the categories that were used in your history.

I'm not opposed to having a separate tool but I don't get why that would be nicer for reproducibility, as the categories that are retrieved by the tool are listed in the output file, so they are in the history (and you could cut out just those columns if you wanted), or am I missing what you mean? 

Do you think for the moment I just leave these 4 org packages in then? I'd be happy to create a separate tool if you think that's better but then there'd still be the issue of how many species to add and that, yes they're big packages!

mblue9:
I forgot to add that goseq can also get the gene lengths which could be added in, however, the goseq manual recommends using the output of the upstream tools like featureCounts to get the gene lengths, see quote below from goseq manual pg 3 here: https://www.bioconductor.org/packages/devel/bioc/vignettes/goseq/inst/doc/goseq.pdf

> We encourage all users to manually specify their gene lengths if provided by upstream summarization programs. e.g. featureCounts, as these lengths will be more accurate.

so I was wondering if I should add a recommendation to use featureCounts lengths instead of (or aswell as?) suggesting they can use the Gene length and GC content tool?

mblue9:
Ah @mvdbeek, I think I see what you mean about reproducibility now - in the output the categories are not linked to the gene ids. 
But maybe I could add in an option to output a table with the mappings? It'd be easy to add the output from go_map, something like below (but note that the category classes (BP, CC, MF) aren't present at this stage, just the identifiers, although they could maybe be added by combining with the final output table. 140 and 1100 are KEGG ids)

ENSG0000014045936 | GO:1901615
-- | --
ENSG0000014045937 | GO:1902652
ENSG0000014045938 | 140
ENSG0000014045939 | 1100

mvdbeek:
> Do you think for the moment I just leave these 4 org packages in then?

Yes, I think that's OK, we don't really have a better means of discriminating which packages to use/install. Optional dependencies would be really cool to have, but that would need to be implemented on the Galaxy side.

> so I was wondering if I should add a recommendation to use featureCounts lengths instead of (or aswell as?)

Yes, that would be good, I think.

> But maybe I could add in an option to output a table with the mappings?

That would be great, but if it's too much of a hassle I'm happy to merge without it. After all if you have those packages installed it should be reproducible.

A separate tool would be cool if we implement another gene enrichment tool, but we can get back to this once we start working on it. Let me know if you want to implement some of these things or if I should merge!

mblue9:
@mvdbeek thanks for reviewing. I've made those changes along with a few other small ones. Instead of outputting a mapping file of genes to categories, I decided for the moment to add an option to output the RData file so a user can get all the R objects used. A separate mapping file output could be added later if need be.

mvdbeek:
Wonderful, thanks a lot @mblue9!

Issue #1524, opened 2017-10-18
Add support for bedtools coverage -f and -F options


bwlang:

bwlang:
fixed in #1525

PR #1525, opened 2017-10-18
adds options for -f -F -r and -e

bwlang:
re #1524 

I'd like to get some initial feedback on this before I do more on it... thanks!


bgruening:
@bwlang looks good for me. You need to increase the tool version number and maybe add/extend a test.

bwlang:
I think this is ready for review.

tools/bedtools/coverageBed.xml
```diff
@@ -14,6 +14,14 @@
         $hist
         $split
         $strandedness
+        #if str($overlap_a) != "None" and str($overlap_a):
+          -f $overlap_a
```
bgruening: these are text parameters and should be single quoted
I'm wondering if we should also add some sanitize to the text field.

bwlang:
@bgruening  any further changes needed for this PR?

bgruening:
Yes, looks great. Thanks.

PR #1526, opened 2017-10-19
update length and gc content tool, and dexseq

mblue9:
length and gc content tool:
- add ability to use built-in gtfs (gene_sets.loc) with test
- update requirement versions
- add bit more to help
- increase wrapper version
- add @bernt-matthias's fasta fix

dexseq:
- add @bernt-matthias's requirement version update
- update test data
- increase wrapper version


mblue9:
Forgot to add, there's no citation given for the length and gc content tool, should there be?

tools/dexseq/dexseq.xml
```diff
@@ -1,7 +1,7 @@
-<tool id="dexseq" name="DEXSeq" version="1.20.1">
+<tool id="dexseq" name="DEXSeq" version="1.20.2">
```
nsoranzo: I think should be `1.22.0` to be in sync with the `bioconductor-dexseq` requirement.

tools/dexseq/dexseq_count.xml
```diff
@@ -1,7 +1,7 @@
-<tool id="dexseq_count" name="DEXSeq-Count" version="1.20.1">
+<tool id="dexseq_count" name="DEXSeq-Count" version="1.20.2">
```
nsoranzo: Same here.

mblue9: Thanks @nsoranzo, I've fixed this version and the one above now.

tools/length_and_gc_content/get_length_and_gc_content.xml
```diff
@@ -20,70 +20,136 @@
...
+
+#if $fasta_file.fastaSource == 'indexed':
+    ln -s '$fasta_file.fasta_pre_installed.fields.path' fasta
+#else:
+    ln -s '$fasta_file.fasta_history' fasta
+#end if
+
+&&
+
+Rscript '$__tool_directory__'/get_length_and_gc_content.r
```
bgruening: this should be `'$__tool_directory__/get_length_and_gc_content.r'`

bgruening:
Again the issue with the default channel :(

mblue9:
ok, thanks for the info and for making that correction @bgruening

tools/length_and_gc_content/get_length_and_gc_content.xml
```diff
@@ -20,70 +20,136 @@
                level="fatal"
                description="An undefined error occured, please check your input carefully and contact your administrator." />
     </stdio>
+    <version_command><![CDATA[
+        echo $(R --version | grep version | grep -v GNU)", optparse version" $(R --vanilla --slave -e "library(optparse); cat(sessionInfo()\$otherPkgs\$optparse\$Version)" 2> /dev/null | grep -v -i "WARNING: ")", reshape2 version" $(R --vanilla --slave -e "library(reshape2); cat(sessionInfo()\$otherPkgs\$reshape2\$Version)" 2> /dev/null | grep -v -i "WARNING: ")", rtracklayer version" $(R --vanilla --slave -e "library(rtracklayer); cat(sessionInfo()\$otherPkgs\$rtracklayer\$Version)" 2> /dev/null | grep -v -i "WARNING: ")", data.table version" $(R --vanilla --slave -e "library(data.table); cat(sessionInfo()\$otherPkgs\$data.table\$Version)" 2> /dev/null | grep -v -i "WARNING: ")
```
yhoogstrate: I love this :heart_eyes:

bernt-matthias:
How about making the gc content / length output optional?

nsoranzo:
xref. https://github.com/conda/conda/issues/5536

mblue9:
@bernt-matthias sounds like a good idea to me, I've made the gc content / length output optional now

tools/length_and_gc_content/get_length_and_gc_content.xml
mblue9: :smile: I learnt that one at the May hackathon from @bgruening and the deseq2 wrapper

bernt-matthias:
Just discovered a bug in dexseq (present in the 1.20.1 and 1.22.0): The call to DEXSeqHTML needs the parameter fitExpToVar (see also https://support.bioconductor.org/p/99557/). Should be something like this: 

`    DEXSeqHTML(res, fitExpToVar=primaryFactor, path=opt$reportdir, FDR=opt$fdr, color=c("#B7FEA0", "#FF8F43", "#637EE9", "#FF0000", "#F1E7A1", "#C3EEE7","#CEAEFF", "#EDC3C5", "#AAA8AA"))
`

I'm not entirely sure. Some more expert dexseq user should check 

Also change the test: 
- set report to True. Then this branch of the R code is also tested. 
- and change the name of the first factor in the test from "condition" to something else. Then the bug gets noticed. ("condition" is the default value for fitExpToVar and therefore the test would work)

mblue9:
Finally got a chance to finish this, @bernt-matthias I added your suggestion for the fitExpToVar parameter and test and it looks like it worked, the primary factor no longer needs to be called "Condition". 
Sorry, I also accidentally added into this PR some deseq2 changes I was testing to fix the salmon/sailfish error Jen reported here https://biostar.usegalaxy.org/p/25406/ (they look like they work)

mblue9:
This is failing here because the deseq2 helper scripts (dexseq_count.py and dexseq_prepare_annotation.py) need 2to3 run on them, I'm taking a look to see how to add that to the dexseq bioconda recipe.

mblue9:
Have submitted PR to bioconda to add 2to3 to dexseq here: https://github.com/bioconda/bioconda-recipes/pull/6916

bernt-matthias:
wonderful. thanks.

bernt-matthias:
ping @bgruening maybe you can also merge this? don't know about the conda package...

bgruening:
I'm kinda on this. I fixed the conda package and waiting now that travis is passing. After this, the PR will hopefully pass ... travis has just a huge backlog :(

bgruening:
This one needs to turn green: https://travis-ci.org/bioconda/bioconda-recipes/builds/331319378?utm_source=github_status&utm_medium=notification

bgruening:
Conda package available and test retriggered.

mvdbeek:
This looks promising, I think we just need to update the test output for dexseq

mvdbeek:
Actually it looks like the test output is slightly different every dexseq runs :/. I guess we could use `compare="size"`.

bgruening:
@mvdbeek or @yhoogstrate do you mind having a final look over it?

mvdbeek:
Still looks good to me!

tools/deseq2/deseq2.xml
```diff
@@ -1,8 +1,7 @@
-<tool id="deseq2" name="DESeq2" version="2.11.39">
+<tool id="deseq2" name="DESeq2" version="2.11.40">
     <description>Determines differentially expressed features from count tables</description>
     <requirements>
-        <requirement type="package" version="1.14.1">bioconductor-deseq2</requirement>
-        <requirement type="package" version="1.20.0">r-getopt</requirement>
+        <requirement type="package" version="1.18.1">bioconductor-deseq2</requirement>
```
mvdbeek: We should add tximport here

bgruening: added it

bgruening:
Thanks all!

PR #1532, opened 2017-10-24
Added Shovill - faster Spades assembly from Illumina reads

Slugger70:
Added Torsten Seemann's Shovill assembly pipeline.


tools/shovill/shovill.xml
```diff
@@ -0,0 +1,165 @@
...
+    <description>Faster SPAdes assembly of Illumina reads</description>
+    <requirements>
+        <requirement type="package" version="0.8.0">shovill</requirement>
+    </requirements>
+    <version_command>shovill --version</version_command>
+    <command detect_errors="exit_code"><![CDATA[
+        shovill
+            --outdir 'out'
+            --cpus \${GALAXY_SLOTS:-1}
+            --ram \${SHOVILL_RAM:-4}
```
bgruening: Is this parameter really needed? What happens if we remove it? We had some discussion about including a general variable, similar to GALAXY_SLOTS in https://github.com/galaxyproject/galaxy/issues/4741

tools/shovill/shovill.xml
```diff
@@ -0,0 +1,165 @@
...
+            <param name="asm" argument="--asm" type="select" label="Spades result to correct" help="Which SPAdes output to correct for: before_rr, contigs or scaffolds. (default: 'contigs')">
+                <option value="before_rr">before_rr</option>
+                <option value="contigs" selected="true">contigs</option>
+                <option value="scaffolds">scaffolds</option>
+            </param>
+        </section>
+    </inputs>
+    
+    <outputs>
+        <data name="shovill_std_log" format="txt" label="${tool.name} on ${on_string} Log file" from_work_dir="out/00-shovill.log" />
```
bgruening: can we write this to the Galaxy stdout, or make it at least optional?

bgruening:
Thanks @Slugger70. I added a few comments and made a few small commits.

tools/shovill/shovill.xml
Slugger70: It is needed otherwise stupid java for pilon tries to use too much RAM.. Blame Broad.

Slugger70: It is needed for the meantime otherwise Pilon (java) tries to use too much RAM by default.

Slugger70:
No worries.. I saw the double quotes -> single quotes.. Sorry, that's just the perl programmer in me. hahaha

bgruening:
Thanks @Slugger70!

PR #1534, opened 2017-10-25
HiCExplorer

joachimwolff:
This PR adds galaxy wrappers for HiCExplorer, a software to analyze Hi-C data.
https://github.com/maxplanck-ie/HiCExplorer


bgruening:
I'm running out of ideas here. This test is passing locally for me on two different machines.
The failure is also due to a "warning" which should not fail in this way ... any help welcome.

bgruening:
@joachimwolff I have found this: https://github.com/ContinuumIO/anaconda-issues/issues/6678

bgruening:
So this code snippet might help:
```
import warnings
warnings.filterwarnings("ignore", message="numpy.dtype size changed")
warnings.filterwarnings("ignore", message="numpy.ndarray size changed")
```
A bioconda patch would be nice for this.

bgruening:
For reference: The `warnings` module behaves differently depending on the environment. I assume that travis marks the VM somehow as test environment, which will turn warnings into errors. HiCExplorer now filters the warnings upstream. So tests passes here.
@joachimwolff I did some cosmetic changes, as the stdio tag is not needed in our case.

joachimwolff:
Can imho be merged if there are no more open issues.

tools/hicexplorer/hicCorrelate.xml
```diff
@@ -53,7 +58,7 @@
             </when>
         </conditional>
         <repeat name="chromosomes" title="List of chromosomes to be included in the correlation" min="0">
-            <param name="chromosome" type="text"/>
+            <param name="chromosome" type="text" optional="false"/>
```
bgruening: I think what you want to use is: `<validator type="empty_field" />`

bgruening: e.g. here:https://github.com/galaxyproject/tools-iuc/blob/fbc18d9128669e461e76ed13307ee88dd774afa5/tool_collections/snpsift/snpsift/snpSift_rmInfo.xml

tools/hicexplorer/hicCorrelate.xml
```diff
@@ -18,14 +18,19 @@
...
 
-            #if $use_range.select_use_range == "yes_use_range"
+            #if $use_range.select_use_range == "yes_use_range":
                 --range $range_min:$range_max
             #end if
 
             #if $chromosomes:
-                --chromosomes #echo '" "'.join([ '"%s"' % $chrom for $chrom in $chromosomes ])#
+                
+                #set $chromosome_list =  '" "'.join([ '"%s"' % $chrom.chromosome for $chrom in $chromosomes ])#
```
bgruening: `"' '".join([ "'%s'" % $chrom.chromosome for $chrom in $chromosomes ])#`

Maybe this is better.

bgruening:
Cool thanks @joachimwolff!

PR #1535, opened 2017-10-25
add annotateMyIDs tool

mblue9:
This PR proposes the addition of a new tool called annotateMyIDs that uses the Bioconductor annotation packages to add gene info to ids. 
It can be used to do useful things like 

- add gene names to ids
- map between ids e.g. Entrez, Ensembl, symbols
- add GO and KEGG terms 

(@mvdbeek I was wondering if this tool could maybe be the separate GO/KEGG category getting tool as mentioned in the goseq discussion here https://github.com/galaxyproject/tools-iuc/pull/1523)


mblue9:
Hi guys, thanks for the suggestions here too! I've made those changes now, including changing to use a configfile instead of the Rscript (suggestion from @bgruening). So this commit will:

- change to single quotes for files and text
- remove unneeded licence file
- add configfile and replace rscript (the configfile is only ~30 lines vs > 50 for the Rscript, and with the config there is no longer a need for the optparse requirement so I've removed that)
- add in more options for input ids and output columns, with the ability to select the output columns using checkboxes (I stopped at the options included here but there are other inputs/outputs that could be easily added e.g. Uniprot, PFAM)
- add more help

mblue9:
I added "Annotation" as a new category to the .shed.yml for this tool (I couldn't see another category for annotation tools like this one) is that ok and do I need to create that category somewhere first?

nsoranzo:
@mblue9 The category needs to be added on the TTS and MTS first, otherwise the upload won't work. You can ask on the IRC/gitter channel or open an issue on the galaxy repo.

mblue9:
Ok thanks @nsoranzo, I'll open an issue in the galaxy repo to request adding an Annotation category then and to request one for CRISPR too if possible, as I'll be finished wrapping MAGeCK soon 😄

bgruening:
@martenson should be able to do that.

tools/annotatemyids/annotateMyIDs.xml
```diff
@@ -0,0 +1,222 @@
+<tool id="annotateMyIDs" name="annotateMyIDs" version="0.3">
+    <description>- annotate a generic set of identifiers</description>
```
yhoogstrate: `s/>- />/g`

tools/annotatemyids/annotateMyIDs.xml
```diff
@@ -0,0 +1,222 @@
...
+    ]]></version_command>
+    <command detect_errors="exit_code"><![CDATA[
+        #if $rscriptOpt:
+            cp '${annotatemyids_script}' '${out_rscript}' &&
+        #end if
+        Rscript '${annotatemyids_script}'
+    ]]>
+    </command>
+    <configfiles>
+        <configfile name="annotatemyids_script"><![CDATA[
```
yhoogstrate: @nsoranzo are there conventions about when we use configfiles and when we include scripts as separate files?

nsoranzo: I don't think we have a convention yet, but I'd say that they should be used for config files you pass to a program, not to fill a script with variables you can pass on the command line.
An excellent example of a good use of `<configfile>` is: https://github.com/peterjc/galaxy_mira/blob/e4c56df75150c82d3e9c4ac487c4209108e52412/tools/mira4_0/mira4_mapping.xml

bgruening: I disagree here. In this example and in many more we are ending up with a ton of code, solely for the purpose of passing around arguments. This is error prone in many ways. Don't repeat yourself, especially not if you invent a script with the only purpose of accepting command-line args.

Moreover, this makes it possible to write some really powerful tools which I outlined here:
https://galaxyproject.org/blog/2017-10-10-scipy-galaxy/

I would not put scripts into a configfile, which do have an external purpose and are useful next to Galaxy.

mvdbeek: I understand the advantages that @bgruening mentioned, but I think writing scripts into the configfiles is terrible, because
  - you need to mentally "parse" the galaxy variables and cheetah stuff from the script
  - you can't test your script independently from the galaxy tool
  - you can't step in with a debugger 
  - you can't have code-highlighting / linting etc.

I'm with @nsoranzo, we should use the configfiles section for configfiles.

nsoranzo: > Don't repeat yourself, especially not if you invent a script with the only purpose of accepting command-line args.

We've enforced that for many years, and it's not the point here.

> Moreover, this makes it possible to write some really powerful tools which I outlined here:
https://galaxyproject.org/blog/2017-10-10-scipy-galaxy/

Putting a script in a `<configfile>` makes it impossible to run static checks/linting on it or to run something like `2to3`. Think about mako templates, it's an ugly mess.
I don't think a couple of line of argparse are a big problem, if you really need to have a script.

bgruening: > you need to mentally "parse" the galaxy variables and cheetah stuff from the script

I doubt that this outweighs the disadvantages of writing argarse stuff, deal with UTF-8 errors (as seen in this script etc....).

> you can't test your script independently from the galaxy tool
> you can't step in with a debugger

You can, if you simply output the script into stdout (or extra file) for example, as we usually do if we use configfiles. Which turns out (at least for non-IT people) to be even better because they see in one go what this script is about and don't need to deal with script+variables+commandline.

> you can't have code-highlighting / linting etc.

You can or can not do this with the command-section as well - depending on your editor. 
And please keep in mind I'm not talking about complicated scripts here, I'm talking about a few lines (c) of code that needs (> c x 2) of code for argument parsing.

bgruening: > We've enforced that for many years, and it's not the point here.

It's exactly my point :) I don't see why we need to force people to write a bulk of useless code.

> Putting a script in a <configfile> makes it impossible to run static checks/linting on it or to run something like 2to3. Think about mako templates, it's an ugly mess.
I don't think a couple of line of argparse are a big problem, if you really need to have a script.

It's by far not a couple of lines if you look at the given example in the post. It will become many hundred lines of code, plus an additional dependency etc. Again, I'm talking here about small scripts with a few lines. Nothing which you ever need to port with 2to3.

bgruening: Maybe we should move this discussion to the https://github.com/galaxy-iuc/standards repository.
IUC currently has tools with this trick and it does not violate anything of our standards, afaik. https://github.com/galaxyproject/tools-iuc/tree/master/tools/vegan

mvdbeek: > I doubt that this outweighs the disadvantages of writing argarse stuff, deal with UTF-8 errors (as seen in this script etc....).

I do think it does. That UTF-8 stuff is copy-paste from the deseq2.R script, I have no clue if that is really necessary. I'd say probably not.



> You can, if you simply output the script into stdout (or extra file) for example, as we usually do if we use configfiles.

Right -- with hardcoded variables. Not very easy to debug.



> You can or can not do this with the command-section as well - depending on your editor.

There we don't have a choice.



> It's by far not a couple of lines if you look at the given example in the post.

I'd say even that example is not great. You can write out the galaxy.json and do exactly the same thing ina regular python script. Which could be tested unit-style using a galaxy.json file.

mvdbeek: In any case I'm still fine with the tool as it is. For me this is short enough to be a configfile (but we do have other tools that I find hard to read/maintain), hence my objection here.

mblue9: >That UTF-8 stuff is copy-paste from the deseq2.R script, I have no clue if that is really necessary

Yes it is exactly that! just copy-paste from the deseq2.R script, as I saw that was recommended as the model to follow for R scripts here: https://github.com/nturaga/bioc-galaxy-integration#deseq2-a-model-for-galaxy-tool-integration but if it's not needed please let me know and I won't keep adding it!

bgruening:
@mblue9 thanks a lot and have a great conference!

PR #1536, opened 2017-10-26
Mothur data manager fixes

abretaud:
A few changes to the mothur data manager as I needed it on our instance. I'm not sure of the status of this data manager currently, here's the PR anyway.
As there was no .shed.yml it was not available in toolshed, so I added the file.
I also added newer data to the option + various small fixes


data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
```diff
@@ -1,70 +1,79 @@
+<?xml version="1.0"?>
 <tool id="data_manager_fetch_mothur_reference_data" name="Fetch Mothur toolsuite reference data" version="0.1.2" tool_type="manage_data">
```
bgruening: @abretaud I think we need to increase the version number here.

bgruening:
Thanks @abretaud!

PR #1544, opened 2017-10-27
Add a Manual Data Manager.

blankenberg:
Allows manually building data table entries and optionally providing filesystem content.


bgruening:
@blankenberg any change we could run/test data managers in planemo in the near future? This would be totally awesome and could help in creating a community-maintained CVMFS.

Thanks for this PR and sorry it took that long.

PR #1545, opened 2017-10-29
StringTie modifications

mblue9:
This PR has the changes from this fix https://github.com/galaxyproject/tools-iuc/pull/1517 plus some more suggested modifications for StringTie.

### DIFF
--- a/tools/stringtie/stringtie.xml
+++ b/tools/stringtie/stringtie.xml
@@ -8,9 +8,6 @@
     <expand macro="version_command" />
     <command><![CDATA[
         mkdir -p ./special_de_output/sample1/ &&
-        #if str($guide.use_guide) == 'yes':
-            ln -s '$guide.guide_gff' ./special_de_output/sample1/guide.gtf &&
-        #end if
 
         #if $input_bam.metadata.ftype == 'sam':
             samtools sort -@ \${GALAXY_SLOTS:-1} '$input_bam' | stringtie
@@ -47,11 +44,13 @@
 
         #if str($guide.use_guide) == 'yes':
             #if $guide.special_outputs.special_outputs_select == 'deseq2':
+                && 
+                ln -s '$output_gtf' ./special_de_output/sample1/output.gtf
                 &&
                 prepDE.py
                     -i ./special_de_output/
-                    -g gene_cout_matrix.tsv
-                    -t transcripts_count_matrix.tsv
+                    -g ./gene_counts.tsv
+                    -t ./transcript_counts.tsv
                     -l $guide.special_outputs.read_length
                     #if str($option_set.options) == 'advanced':
                         -s '$option_set.name_prefix'
@@ -59,15 +58,25 @@
                     #if $guide.special_outputs.clustering:
                         -c
                         --legend ./legend.tsv
-
                         &&
                         sed -i.bak 's/,/\t/g' ./legend.tsv
-
+                        &&
+                        sed -i.bak 's/\r//g' ./legend.tsv
+                        &&
+                        mv -T ./legend.tsv "$legend"
                     #end if
                 &&
-                sed -i.bak 's/,/\t/g' transcripts_count_matrix.tsv
+                sed -i.bak 's/,/\t/g' ./transcript_counts.tsv
                 &&
-                sed -i.bak 's/,/\t/g' gene_cout_matrix.tsv
+                sed -i.bak 's/\r//g' ./transcript_counts.tsv
+                &&
+                mv -T ./transcript_counts.tsv "$transcript_counts"
+                &&
+                sed -i.bak 's/,/\t/g' ./gene_counts.tsv
+                &&
+                sed -i.bak 's/\r//g' ./gene_counts.tsv
+                &&
+                mv -T ./gene_counts.tsv "$gene_counts"
             #end if
         #end if
     ]]></command>
@@ -155,16 +164,13 @@
             label="${tool.name} on ${on_string}: intron to transcript mapping">
             <filter>guide['use_guide'] == 'yes' and guide['special_outputs']['special_outputs_select'] == 'ballgown'</filter>
         </data>
-        <data name="gene_counts" format="tabular" from_work_dir="gene_cout_matrix.tsv"
-            label="${tool.name} on ${on_string}: Gene counts">
+        <data name="gene_counts" format="tabular" label="${tool.name} on ${on_string}: Gene counts">
             <filter>guide['use_guide'] == 'yes' and guide['special_outputs']['special_outputs_select'] == 'deseq2'</filter>
         </data>
-        <data name="transcript_counts" format="tabular" from_work_dir="transcripts_count_matrix.tsv"
-            label="${tool.name} on ${on_string}: Transcript counts">
+        <data name="transcript_counts" format="tabular" label="${tool.name} on ${on_string}: Transcript counts">
             <filter>guide['use_guide'] == 'yes' and guide['special_outputs']['special_outputs_select'] == 'deseq2'</filter>
         </data>
-        <data name="legend" format="tabular" from_work_dir="legend.tsv"
-            label="${tool.name} on ${on_string}: legend">
+        <data name="legend" format="tabular" label="${tool.name} on ${on_string}: legend">
             <filter>guide['use_guide'] == 'yes' and guide['special_outputs']['special_outputs_select'] == 'deseq2' and guide['special_outputs']['clustering'] is True</filter>
         </data>
     </outputs>
--- a/tools/stringtie/stringtie.xml
+++ b/tools/stringtie/stringtie.xml
@@ -1,4 +1,4 @@
-<tool id="stringtie" name="StringTie" version="1.3.3">
+<tool id="stringtie" name="StringTie" version="1.3.3.1">
     <description>transcript assembly and quantification</description>
     <macros>
         <import>macros.xml</import>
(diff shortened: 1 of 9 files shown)


mblue9:
All the tests are passing here except test 6, it's still failing here yet passing locally. Am wondering if a chmod  of the prepDE.py needs to be added somewhere to get past the error?: "/tool_script.sh: line 25: prepDE.py: command not found"

NCEichner:
Hi @mblue9!
Thanks for adopting my suggestions and adding the changes from @yhoogstrate. I admit that in this special case I'm still not feeling very comfortable with this 'from_work_dir' construct because one could possibly see '.csv-zombiefiles' when clicking on the yellow history-entries just before the sed-commands are executed. But as this maybe a minor problem and the 'from_work_dir' tag seems to be the preferred method I will step back in this...
Humble Question: Are the two lines with redirection to 'std.out' really needed as you use the 'from_work_dir'-tags at the same time? Aren't the variable names (e.g.: $legend) already linked to the actual filenames when using these tags in the output-section?

About the 'additional column'-problem: On our system I see this additional column in the 'gene counts' file only when I'm setting the clustering option to 'No'. Did you also test this option?

As I still find no time to set up a planemo testsystem I actually cannot test your changes as you wish. Just copying the .xml file on our productive system and reset the metadata will not work... ;-)
I'm sorry but for now I have to hand this tests over to @yhoogstrate or others. Maybe @jennaj as the original thread starter (see #1322) could have a look on this? 
But 'in silico' your changes look quite interesting especially the strandness-parameters.

mvdbeek:
> Am wondering if a chmod of the prepDE.py needs to be added somewhere to get past the error?: "/tool_script.sh: line 25: prepDE.py: command not found"

it seems there's a problem with the conda package, prepDE.py is missing for me as well (even though it should be there according to the recipe at https://github.com/bioconda/bioconda-recipes/blob/master/recipes/stringtie/build.sh)

mvdbeek:
https://github.com/bioconda/bioconda-recipes/pull/6618/files should fix prepDE.py

bgruening:
@mblue9 in theory it work now. Can you update the test data please.

@NCEichner 
> I'm still not feeling very comfortable with this 'from_work_dir' construct because one could possibly see '.csv-zombiefiles' when clicking on the yellow history-entries just before the sed-commands are executed. 

`from_work_dir` is only executed after a job has finished. It will not be tricked during a job run by user-interaction. So this should be save.

mblue9:
Hi guys, @mvdbeek thanks for fixing the prepDE.py issue! I've updated the test data for that failed test 6 so fingers crossed it passes here now. 
@NCEichner, thanks for taking a look at the changes! The only reason I added those 2 std.out lines was to trap the stdout message from appearing in the history, otherwise you get something like "[] [].." appearing, which our group thinks looks a bit too ugly and error-like. If there's a better way to do that please let me know. I did try to see if I could see that 3rd column appearing with and without clustering and I didn't see it, but I only used the small stringtie deseq test data files from here: 
https://github.com/galaxyproject/tools-iuc/blob/master/tools/stringtie/test-data/stringtie_in.gtf
https://github.com/galaxyproject/tools-iuc/blob/master/tools/stringtie/test-data/stringtie_in1.bam
so I'll check with some real size files when I get a chance.

mblue9:
Yay! finally green here!

tools/stringtie/stringtie.xml
```diff
@@ -7,133 +7,185 @@
...
+#if str($guide.use_guide) == 'yes':
+    #if $guide.guide_source.guide_gff_select == "history":
+        ln -s '$guide.guide_source.ref_hist' guide.gff &&
+    #elif $guide.guide_source.guide_gff_select == "cached":
+        ln -s '$guide.guide_source.ref_builtin.fields.path' guide.gff &&
+    #end if
+#end if
+
+#if $input_bam.metadata.ftype == 'sam':
+    samtools sort -@ \${GALAXY_SLOTS:-1} '$input_bam' | stringtie
```
yhoogstrate: samtools is not defined in the dependencies and is not a dependency of stringtie. This means it will take the samtools provided by Galaxy which in theory could change version in a future release. Would it be handier to include samtools as dependency? I don't know if adding samtools as dependecy may cause problems?

tools/stringtie/stringtie.xml
```diff
@@ -7,133 +7,185 @@
...
+                    -k '$guide.special_outputs.key'
+                #end if
+                --legend '$legend'
+                > /dev/null
                 &&
-                sed -i.bak 's/,/\t/g' transcripts_count_matrix.tsv
+                sed -i.bak 's/,/\t/g' '$legend'
                 &&
-                sed -i.bak 's/,/\t/g' gene_cout_matrix.tsv
+                sed -i.bak 's/\r//g' '$legend'
```
yhoogstrate: perfect addition (https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/datatypes/tabular.py#L72)

tools/stringtie/stringtie.xml
```diff
@@ -7,133 +7,185 @@
...
             #end if
-        #end if
+
+        > /dev/null
```
yhoogstrate: does StringTie put logging information or data to stdout? If it is logging it might be good to not erase it.

tools/stringtie/stringtie.xml
```diff
@@ -7,133 +7,185 @@
...
+        &&
+        sed -i.bak 's/,/\t/g' '$gene_counts'
+        &&
+        sed -i.bak 's/\r//g' '$gene_counts'
+    #end if
+#end if
     ]]></command>
     <inputs>
-        <param name="input_bam" type="data" format="sam,bam" label="Mapped reads to assemble transcripts from" />
+        <param name="input_bam" type="data" format="sam,bam" label="Input mapped reads" help="Input one or more BAM files containing reads you want to assemble into transcripts"/>
```
nsoranzo: "one or more BAM files": there doesn't seem to be support for multiple inputs in the wrapper. Also you can mention SAM too.

tools/stringtie/stringtie.xml
```diff
@@ -7,133 +7,185 @@
...
+#else
+    stringtie '$input_bam'
+#end if
+
+-o '$output_gtf'
+-p "\${GALAXY_SLOTS:-1}"
+
+#if $rna_strandness:
+    $rna_strandness
+#end if
```
nsoranzo: You don't need the `#if`/`#end if`, just `$rna_strandness` should be enough.

tools/stringtie/stringtie.xml
```diff
@@ -7,133 +7,185 @@
...
+                        <option value="cached" selected="true">Use a built-in file</option>
+                        <option value="history">Use a file from history</option>
+                    </param>
+                    <when value="cached">
+                    <param name="ref_builtin" type="select" label="Use a built-in GTF" help="If the GTF file for your transcriptome of interest is not listed, contact your Galaxy administrator">
+                        <options from_data_table="gene_sets">
+                            <filter type="sort_by" column="2" />
+                            <validator type="no_options" message="No GTF file is available." />
+                        </options>
+                    </param>
```
nsoranzo: The above param is not indented enough.

tools/stringtie/stringtie.xml
```diff
@@ -7,133 +7,185 @@
...
+                    </when>
+                    <when value="history">
+                        <param name="ref_hist" type="data" format="gtf,gff3" label="Use a history file" help="Select a GTF/GFF3 file from the history" />
```
nsoranzo: I'd use "GTF/GFF3 dataset to guide assembly" as label and remove  `help`.

tools/stringtie/stringtie.xml
```diff
@@ -7,133 +7,185 @@
...
-                        <option selected="True" value="deseq2">DESeq2/EdgeR</option>
-                        <option value="no">No addional output</option>
+                        <option value="deseq2">DESeq2/edgeR</option>
+                        <option value="no" selected="True">No additional output</option>
                     </param>
                     <when value="ballgown" />
                     <when value="deseq2">
-                        <param name="read_length" type="integer" value="75" label="Average read length" />
-                        <param name="clustering" type="boolean" truevalue="--cluster" falsevalue="" label="Whether to cluster genes that overlap with different gene IDs" help="ignoring ones with geneID pattern" />
+                        <param name="read_length" type="integer" min="0" value="75" label="Specify the average read length" help="Default: 75" />
```
nsoranzo: I feel adding default value in help is redundant, but we don't have a policy about this.

tools/stringtie/stringtie.xml
```diff
@@ -7,133 +7,185 @@
...
+                        <param name="clustering" type="boolean" truevalue="--cluster" falsevalue="" checked="False" label="Cluster overlapping genes" help="Choose whether to cluster genes with different gene IDs that overlap. Transcripts containing the geneID prefix will be ignored. Default: No" />
```
nsoranzo: `argument="--cluster"` after `name`

tools/stringtie/stringtie.xml
```diff
@@ -7,133 +7,185 @@
...
+                        <param name="string" type="text" label="Prefix used for transcripts" help="If a different prefix was used for geneIDs assigned by StringTie than the default, specify it here. Only letters and numbers will be retained in this field. Default: MSTRG" >
```
nsoranzo: Replace `name` with `argument="--string"`

tools/stringtie/stringtie.xml
```diff
@@ -7,133 +7,185 @@
...
+                            <sanitizer>
+                                <valid initial="string.letters,string.digits"></valid>
+                            </sanitizer>
+                        </param>
+                        <param name="key" type="text" label="Prefix for clustering" help="If clustering, what prefix to use for geneIDs assigned by this script. Only letters and numbers will be retained in this field. Default: prepG">
```
nsoranzo: Replace `name` with `argument="--key"`

tools/stringtie/stringtie.xml
nsoranzo: Add `argument="--length"` after `name`

mblue9:
Thanks guys, I've made your suggested changes here now. @bgruening I left the sections in but if it's better if I remove them for the moment let me know.

tools/stringtie/stringtie.xml
mblue9: I added samtools 1.6 as a dependency and the tests pass, do you think that means it's ok to include?

mvdbeek: Yes, in fact this is needed. It just so happens that our travis setup includes a global samtools, but that requirement will go away in 18.01. If a tool uses samtools it must be declared in the requirements.

tools/stringtie/stringtie.xml
mblue9: Sorry @yhoogstrate I'm not quite sure what you mean with this, do you mean I should add in an if statement after the sed here like in that link?

tools/stringtie/stringtie.xml
mblue9: This is only to catch the output from the helper prepDE.py script, from what I've seen Stringtie doesn't put logging info to stdout (but I haven't tested it extensively with all possible options!)

tools/stringtie/stringtie.xml
mblue9: Personally I like having the default value in the help for situations where I change a value in the form from the default but then want to see what the default value was. If the value is not in the help then I either have to refresh the form and lose whatever other values I've entered, or open a new tab to see (or is there a better way to do that?). I'm happy to go with whatever the consensus here is.

tools/stringtie/stringtie.xml
yhoogstrate: Nope, it's great that carriage returns are removed because it is in concordance with the Galaxy datatype of a Tabular file :+1: I wouldn't have thought about something like this myself, so I'm giving a compliment

tools/stringtie/stringtie.xml
yhoogstrate: Thanks for clearing up @mvdbeek :+1:

bgruening:
Thanks all for the reviews and @mblue9 for the great enhancements!

tools/stringtie/stringtie.xml
nsoranzo: > Personally I like having the default value in the help for situations where I change a value in the form from the default but then want to see what the default value was.

@mblue9 Does this happen for you when you re-run the tool? Or when you fill the form the first time?

I was thinking if there may be a way to automatically show the default value in the tool form, like we do with the `argument`, which is added to the help. Any idea @guerler @jmchilton ?

tools/stringtie/stringtie.xml
mblue9: Yes thanks @mvdbeek and thanks for catching that samtools requirement @yhoogstrate!

tools/stringtie/stringtie.xml
mblue9: oh ok, well that compliment goes to @NCEichner then as he added that, I wouldn't have thought of that myself either! 😄

mblue9:
By the way it was @NCEichner who came up the bug fix in this PR (originally here https://github.com/galaxyproject/tools-iuc/pull/1517) that should help fix the long-standing bug from May here https://github.com/galaxyproject/tools-iuc/issues/1322, the rest of this PR is cosmetic changes, so big credit should go to him! and hooray if that bug is fixed at last!!

tools/stringtie/stringtie.xml
mblue9: @nsoranzo it could happen with a rerun, or happen the first time, or it could be if I'm looking at a user's history as, take a look at the screenshot I've attached here of the new FLASH tool, if that was a user's history how would I tell if those are default values for the tool or if the user changed them? 
so if it was possible to show the default option automatically I think that would be great! It would save me from adding it to every option (I'm doing it for all the booleans aswell!)
screen shot 2017-11-10 at 11 08 18 am

NCEichner:
Thanks @mblue9 for your appreciation! 
When I see all the commits and comments in #1322, #1517 and above I see an active community in which everyone spends time and genius for identifying problems (@jennaj), for programmming, reviewing and improving the code (@mblue9, @mvdbeek , @nsoranzo and @yhoogstrate) and for supervising (@bgruening). So I wanted to use this oportunity and send thanks back to all of you (and to the ones I may have forgotten to mention) because you keep this galaxy project alive, constantly improve it and make the users happy in working with the tools!

NCEichner:
Hi @mblue9!
(1) I tried to find out a bit more about the mysterious third column and 
(2) can now give a some more info about the '0 [ ] [ ]' output that you trapped by redirecting to /dev/null:

to (1):
I ran my older 'intermediate' version of stringtie with the test files you mentioned above and could not reproduce the 3rd column in the 'Gene counts' output. So it seems to be a matter of prepDE and how it processes our 'Assembled_transcripts.gtf' files.

to (2):
Summing up the following explanations I think it's OK to hide this because it is either redundant or unnecessary info. 
This strange output is generated by prepDE together with the test files and it's only shown if you use the clustering option. Actually these brackets represent the content of the empty 'legend' file.
If you have clustering transcripts 't1', 't2', etc. then these will be listed in the legend file (before the 'sed' commands are executed) like this:
`
NoOfClustersInFile
t1,t2
t3,t4,t5,t6
(...)

`

Each line represents one cluster of transcripts. After some reformatting this will also be sent to stdout. Being in there it's shown in all history entries of this particular stringtie job like this:
`
NoOfClustersInFile
[['t1', 't2'], ['t3', 't4', 't5', 't6'], (...)]

`

As the test files do not have any clustering transcripts the value for 'NoOfClustersInFile' is zero. This ends up in the stdout:`

0
[]
[]

`
 
The following output line '0 sample1' is just telling that there is only one 'sample1' subfolder containing one '.gtf' file recognized in the special_output folder. But it cannot be more than one as you know from setting up the input in the GUI.
PrepDE.py was originally designed to process more than one sample and create a list with all the samples (shown in columns) and all the genes/transcripts (representing the rows). This tabular file can directly be used e.g. as 'countdata' input in DESeq2. 
But as far as prepDE.py is not getting an own .xml-wrapper and then can process more than one sample this 'sample1' info is rather useless.

mblue9:
@NCEichner sorry for the delayed reply here I've been a bit tied up. Thanks for the the nice comments and sharing this info, that is good to know :smile:. I'd be happy to take a look at seeing if I can make the prepDE bit more efficient when I get some time (might be a few weeks). I'm not currently using this tool so would you have a small workflow with a couple of samples that you could put in usegalaxy.org or somewhere so I can see what you're doing. I could also use that for testing. Are you doing the 5 step recommended workflow for DE from here: https://ccb.jhu.edu/software/stringtie/index.shtml?t=manual

PR #1548, opened 2017-10-30
Bug fixes and improvements for the upgrade to featureCounts 1.5.3

AsierGonzalez:
This PR includes bug fixes and improvements for [#1515](https://github.com/galaxyproject/tools-iuc/pull/1515), namely:

- Corrected issue with filtering junction counts output. Now it creates this output file only if `-J` is selected.

- Logic has been changed so that using `-G` with a genome in fasta format to aid exon-spanning read counting is an option (as it should be) rather than compulsory when `-J` is selected.

- The first test has been expanded so that it also checks that an additional output is created when using the `-J` option **(*)**. The corresponding `output_1_jcounts.tab` output file has been added to the `test-data` directory. However, note that this has only the header line because the input file does not have any reads that span exon-exon junctions.

- The three tests that created `output_2_*` files have been removed because they were exact repeats of the first three tests (instead of using `featureCounts_input2.bam` as an input). All the unused files (featureCounts_input2.bam, output_2_short.tab, output_2_medium.tab and output_2_full.tab) have been removed from `test-data`.

- `expect_num_outputs` have been added to the tests to check that the correct number of output files are being created **(*)**.

**(*)** Note that neither the first test nor the `expect_num_outputs`  have been tested because our planemo is causing problems


tools/featurecounts/featurecounts.xml
```diff
@@ -274,9 +276,17 @@
                        label="Exon-exon junctions"
                        help="If specified, reads supporting each exon-exon junction will be counted" />
                 <when value="-J">
-                    <param name="genome" argument="-G" type="data" format="fasta"
```
nsoranzo: You can obtain the same result without breaking backward-compatibility by just adding `optional="true"` here instead of using the `conditional`.

AsierGonzalez: That sounds neat! Thank you, I didn't know that option existed.

AsierGonzalez: And how can I check within `<command>` whether it has a value? Does something like `#if argument="-G" not empty` exist?

nsoranzo: It's just:
```
#if $extended_parameters.exon_exon_junction_read_counting_enabled.genome
```

AsierGonzalez: Beautiful! Thanks for the suggestions. Everything is working now.

tools/featurecounts/featurecounts.xml
```diff
@@ -274,9 +276,9 @@
...
-                           label="Reference sequence (genome) file"
-                           help="The FASTA-format file that contains the reference sequences used in read mapping can be used to improve read counting for junctions" />
+                      <param name="genome" argument="-G" type="data" format="fasta" optional="true"
```
nsoranzo: Can you restore the indentation to be 4 spaces compared to the parent element (i.e. `<when>`)?

nsoranzo:
Thanks a lot @AsierGonzalez!

AsierGonzalez:
Thanks to you @nsoranzo for all your support and suggestions. I'm happy to contribute to such a nice project. More to come in the furture! :)

PR #1550, opened 2017-11-02
Add DISCO

bebatut:
This PR add a wrapper for DISCO (a metagenomics assembler)


tools/disco/disco.xml
```diff
@@ -0,0 +1,412 @@
+<tool id="disco" name="DISCO" version="@WRAPPER_VERSION@.0">
+    <description>to assemble metagenomics data using an overlap-layout-consensus (OLC) approach</description>
+    <macros>
+        <token name="@WRAPPER_VERSION@">1.2</token>
+        <xml name="assembly_params">
+            <param name="minSequenceLengthTobePrinted" argument="minSequenceLengthTobePrinted" type="integer" value="1000" label="Minimum length of contigs or scaffolds to be printed"/>
```
bgruening: most of them can use a min=0, isn't it?

tools/disco/disco.xml
```diff
@@ -0,0 +1,412 @@
...
+            <param name="minOverlapDifference4ClipBranches" argument="minOverlapDifference4ClipBranches" type="integer" value="25" label="Minimum overlap length difference (bp) to clip branches" help="If a read has multiple edges, Disco clips the branches with overlap lengths less than the largest overlap of this read by this difference or more. Increase this to reduce N50 and mis-assemblies"/>
+            <param name="minUniquePEsupport" argument="minUniquePEsupport" type="integer" value="3" label="Minumum number of paired-end reads that provide unique support to merge two edges" help="Increase this to reduce N50 and mis-assemblies"/>
+            <param name="minNonUniquePEsupport" argument="minNonUniquePEsupport" type="integer" value="0" label="Minumum number of paired-end reads that provide non-unique support to merge two edgess"/>
```
bgruening: the name is redundant here if you use argument

tools/disco/disco.xml
```diff
@@ -0,0 +1,412 @@
...
+        ]]></configfile>
+    </configfiles>
+    <inputs>
+        <conditional name="input">
+            <param name="library_type" type="select" label="Type of library?">
+                <option value="single">Single-end reads</option>
+                <option value="paired">Paired-end reads</option>
+            </param>
+            <when value="single">
+                <param name="inS" type="data" format="fastq,fasta" multiple="true" label="Single-read files" argument="-inS"/>
```
bgruening: the name is redundant here if you use argument

bgruening: and below

tools/disco/disco.xml
```diff
@@ -0,0 +1,412 @@
...
+                        <param name="inP" type="data" format="fastq,fasta" multiple="true" label="Interleaved paired-read files" argument="-inP"/>
+                    </when>
+                </conditional>
+            </when>
+        </conditional>
+        <param name="obg" argument="-obg" type="boolean" truevalue="-obg" falsevalue="" checked="false" label="Only build overlap graph?"/>
+        <param name="osg" argument="-osg" type="boolean" truevalue="-osg" falsevalue="" checked="false" label="Only simplify existing overlap graph?"/>
+        <param name="MinOverlap4BuildGraph" argument="MinOverlap4BuildGraph" type="integer" value="20" label="Minimum overlap length (bp) required to insert an edge between two reads during graph construction" help="Increase this to reduce N50 and mis-assemblies"/>
+        <param name="PrintContigs" argument="PrintContigs" type="boolean" truevalue="true" falsevalue="false" label="Print contigs?" help="Printing takes a non-trivial amount of wall-clock time"/>
+        <param name="PrintScaffolds" argument="PrintScaffolds" type="boolean" truevalue="true" falsevalue="false" label="Print scaffolds?"/>
```
bgruening: should we make a few of these outputs disabled by default

tools/disco/disco.xml
```diff
@@ -0,0 +1,412 @@
...
+        </section>
+        <section name="second_assembly_iter_param" title="Parameters for 2nd assembly iteration" expanded="False">
+            <expand macro="assembly_params"/>
+        </section>
+        <section name="third_assembly_iter_param" title="Parameters for 3rd assembly iteration" expanded="False">
+            <expand macro="assembly_params"/>
+        </section>
+    </inputs>
+    <outputs>
+        <data format="txt" name="log" from_work_dir="disco.log" label="${tool.name} on ${on_string}: Log file"/>
```
bgruening: Can we print this to stdout or make it non-default?

tools/disco/disco.xml
bebatut: I was not sure if it is working with only one `-` and not two... Is it?

tools/disco/disco.xml
bebatut: true

tools/disco/disco.xml
bebatut: Same as below. I thought `argument` needs two `-` to be used as both argument and name. I will remove the redundancy

tools/disco/disco.xml
bebatut: I will do that

tools/disco/disco.xml
bebatut: I will make it as non-default

bgruening:
Thanks B!

PR #1560, opened 2017-11-07
save stdout to separate dataset

Takadonet:
Reason is that parse out "Read combination statistics" in downstream analyses.

### DIFF
--- a/tools/flash/flash.xml
+++ b/tools/flash/flash.xml
@@ -27,7 +27,7 @@
                 #else:
                     -p 64
                 #end if
-                > $log
+                > '$log'
         ]]>
     </command>
     <inputs>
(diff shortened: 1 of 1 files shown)


tools/flash/flash.xml
```diff
@@ -69,6 +70,7 @@
             <filter>layout['select_layout'] == 'individual'</filter>
         </data>
         <data format="tabular" name="hist" from_work_dir="out.hist" label="${tool.name} on ${on_string}: Unmerged reads tabular histogram" />
+        <data format="txt" name="log" label="${tool.name} on ${on_string}: Raw Log" />
```
bgruening: Can we please make this optional, by default off? We should not clutter the history for users that don't need it.

mvdbeek: Hmm, it would be nice if we could control the `hidden` attribute here, so that you could go back and unhide it if necessary.

Takadonet: I made log file optional by default and adjusted test accordingly.

bgruening:
Thanks @Takadonet!

PR #1562, opened 2017-11-08
Use no_ref option for CRAM output in minimap2

mvdbeek:
This avoids reference-based compression (otherwise alignments against custom
references cannot be uncompressed).


bgruening:
Thanks!

PR #1567, opened 2017-11-13
change defaults for spliced alignments

bgruening:
Is this ok?
It was pretty confusing for a few of our users to have spliced alignments deactivated by default.


bgruening:
I don't think so. Rerun should not be affected as you said.

bgruening:
@mblue9 not your fault. This was included also in old version, just wanted to have your feedback here as you touched the code last time.

bgruening:
@yhoogstrate please merge :)

mblue9:
Ah I see and phew!... all good so :smile:

PR #1574, opened 2017-11-15
correct bz2 file ending condition

ieguinoa:


yhoogstrate:
@bgruening Apart from test files, can files of datatype `.fastq.gz` have `.gz`/`.gzip` suffixes or can they also be `.dat` files? I think only the integral datatype of the history item (`opts.informat`) needs to be used ?

bgruening:
I agree with you @yhoogstrate! This looks wired.
I think we need to add additional checks to it, like ` or ftype[-1] == "gzip" or informat.endswith('.gz'):` in the lines above.

@ieguinoa thanks for the fixes!

yhoogstrate:
I found another few small issues that I would like to see fixed:

 - [ ] Since BAM is an available datatype for input, it needs a test
 - [ ] Test-data on the .txt files has `lines_diff="100"`, which needs to be much less and requires updating test data to Fastqc v11.5.
 - [ ] The .zip file format is mentioned but not further used. Is this really functional or can this be removed?
 - [ ] All datatyping should ideally be done using Galaxy's datatyping system and not based on filenames
   * [ ] If possible and desired move pieces of the python file that generate the shell command it into the `<command />`-section.

tools/fastqc/rgFastQC.xml
```diff
@@ -37,25 +37,30 @@
...
-            <param name="input_file" value="1000gsample.fastq.gz" ftype="fastq.gz" />
+            <param name="input_file" value="1000gsample.fastq.gz.dat" ftype="fastq.gz" />
             <param name="contaminants" value="fastqc_contaminants.txt" ftype="tabular" />
             <output name="html_file" file="fastqc_report.html" ftype="html" lines_diff="100"/>
-            <output name="text_file" file="fastqc_data.txt" ftype="txt" lines_diff="100"/>
+            <output name="text_file" file="fastqc_data.txt" ftype="txt" lines_diff="2"/>
         </test>
         <test>
-            <param name="input_file" value="1000gsample.fastq.bz2" ftype="fastq.bz2" />
+            <param name="input_file" value="1000gsample.fastq.bz2.dat" ftype="fastq.bz2" />
```
bgruening: I'm not sure this is needed. planemo is uplaoading data using the normal route a normal web-user would also use, isn't it? So the file will end up in /tmp as a copy ending with dat, isn't it?

yhoogstrate: If you are sure we can absolutely revert them.

mvdbeek: It's nice in the sense that we can see the proper datatype extension being added in the raw text output, let's leave it.

mvdbeek:
Very nice @ieguinoa and @yhoogstrate, this a good example of when to get rid of an external argument parser!

mvdbeek:
I did test this with multiqc as well, all is good. I'm just going to bump the version, since these are some substantial changes (even though the output should remain the same).

PR #1590, opened 2017-11-19
Adding --reorder option

joachimwolff:


bgruening:
@joachimwolff can you add a test case.
I'm still not sure this will work.

PR #1591, opened 2017-11-19
Adding --reorder option

joachimwolff:


mvdbeek:
I'd love to see this option, but keep in mind that many tools in galaxy assume coordinate-sorted input, and that this only works on galaxy instances that use samtools >= 1.x due to a bug in the BAM datatype, not an actual design decision. For galaxy instances that still use 0.1x this will automatically be converted back to a coordinate-sorted BAM file.

That said you'll also need to change the samtools call on line 295, since that does coordinate sorting.

bgruening:
This could now be implemented using the `bam_native` datatype and the `profile="18.01"` tag.
https://docs.galaxyproject.org/en/latest/dev/schema.html#attributes

bgruening:
@mvdbeek could you please have a look.

mvdbeek:
it would be unfortunate to output `bam_native` in this case since we know that most tools that need adjacent mates would work with this. to make robust tool that consume bam_native you'd still need to sort. We don't have that datatype in galaxy yet, so I also don't feel like it's reasonable to block this PR.

mvdbeek:
If you can wait until tomorrow I can make some additional datatypes for this since this has (not sufficiently up to now) been bugging me too.

bgruening:
>  in this case 
In which case?

We need this tool for Friday if possible.
Imho tools that consume bam_native needs to know that they can not rely on anything. Its not sorted, does not have an index - is not manipulated by Galaxy.

bgruening:
@joachimwolff we also need to update this one for the new datatypes.

PR #1592, opened 2017-11-20
added exit code error detection for quast

bernt-matthias:
I had the problem that the quast package was not installed
correctly in galaxy. despite a !=0 exit code the result files
were shown as correct


bgruening:
We could also use `<command detect_errors="aggressive">`

yhoogstrate:
Good catch @bernt-matthias. The suggestion of @bgruening is indeed the standard I believe. Would it be okay to add this change?

bernt-matthias:
Any objections against removing the txt and tex output? Seems redundant (contains the same info as the tsv). 

Also: in the html report there is a dysfunctional link "View in Icarus contig browser". It links to `https://galaxy.intranet.ufz.de/datasets/a27a287a56220d09/display/icarus_viewers/alignment_viewer.html` and results in an error: `Could not find 'icarus_viewers/alignment_viewer.html' on the extra files path /gpfs1/data/galaxy_server/galaxy/database/files/008/dataset_8612_files/icarus_viewers/alignment_viewer.html`. 

Wondering if it is possible to bundle the dataset_8612_files with the html report. Then there would be one output data set less.

tools/quast/quast.xml
```diff
@@ -76,14 +76,28 @@
         <param name="min_contig" type="integer" value="500" label="Lower Threshold"  argument="--min-contig"
             help="Set the lower threshold for a contig length. Shorter contigs won't be taken into account (default is 500)"/>
         <param name="threshold_contig" type="text" value="0,1000" label="Thresholds" argument="--contig-thresholds"
-            help="Set the thresholds for contig length. Comma-separated list of contig length thresholds.(default is 0,1000)"/>
+		help="Set the thresholds for contig length. Comma-separated list of contig length thresholds.(default is 0,1000)"/>
+	<conditional name="add_output">
```
bgruening: This could be a simple optional, multiple select box instead of a conditional.

bgruening:
@bernt-matthias please also increase the version number of the tool. Thanks!

bernt-matthias:
Currently the tool version is the same as the program (which is the latest in bioconda). What would be the suggestion? Something like 4.5_1 / 4.5.1? Does the tool version need to increase if its only a bug fix / minor UI improvement? 

See also a recent discussion in https://github.com/galaxyproject/galaxy/issues/5069#issuecomment-347221110

bernt-matthias:
no idea why all these commits ended up in this PR. any ideas how to proceed?

bgruening:
Does `git rebase upstream/master` helps to get rid of them?

rchikhi:
note that, in the recent versions of Quast (>=4.6), the `--meta` flag does not exist anymore, therefore this wrapper is unable to work in metagenomics mode.

bernt-matthias:
@rchikhi thanks for the hint. Would it be an option to simply remove `--meta`

It also seems that there are quite some options that are not used by the wrapper:
```

-f 
--glimmer
--gene-thresholds <int,int,...>
--gage
--use-all-alignments
--min-alignment <int>
--min-identity <float>
--ambiguity-usage <none|one|all>
--ambiguity-score <float>
--strict-NA
-x  --extensive-mis-size  <int>
--scaffold-gap-max-size  <int>
--unaligned-part-size  <int>
--fragmented
--fragmented-max-indent  <int>
--plots-format  <str>  [Supported formats: emf, eps, pdf, png, ps, raw, rgba, svg, svgz]
--memory-efficient
-1  --reads1  <filename>
-2  --reads2  <filename>
--sam  <filename>
--bam  <filename>
--sv-bedpe  <filename>
```

Since I'm actually no experienced quast user I don't know if I'm able to extend the tool wrt these flags.

rchikhi:
From my experience, I'd say that two potentially demanded flags would be `--fragmented` and `--use-all-alignments`.  I'd guess that the others are for more or less niche uses-cases. (arguably these other two are too)

bgruening:
@bernt-matthias anything we can help here?

bernt-matthias:
Since I don't know the tool well I'm not confident to make the changes (adding the missing options, and removing --meta). Despite the superficial --meta option we could also leave the tool as it is see if the options are demanded... ? Or just open an issue and see if someone jumps in.

bgruening:
@bernt-matthias fair enough. But this PR seems to contain more then needed.

bgruening:
@rchikhi could you fill an issue with your recommendations so we do not forget about it?

nekrut:
Referencing #1714

rchikhi:
Hi @bgruening, I'll tone down what I said earlier, regarding "potentially demanded flags".  I actually believe that all these flags fill niche uses-cases, so it would make little sense to implement an arbitrary subset.

PR #1596, opened 2017-11-20
Add Genehunter-Modscore 3.0

mtekman:
A long-standing linkage and haplotype analysis program

https://www.helmholtz-muenchen.de/en/ige/service/software-download/genehunter-modscore/index.html


tools/ghm/ghm.xml
```diff
@@ -0,0 +1,467 @@
+<tool id="ghm" name="Genehunter-Modscore" version="3.0" >
```
bebatut: Can you use version 3.0.0 here? The idea is to have an extra num for the wrapper version compared to the tool version, such that if we need to change the wrapper we can increase the version number of the wrapper but stay close to the tool version. Is it clear?
You can use like `@VERSION@.0` here with `<token name="@VERSION@">3.0</token>` define in your `macros` section and in `requirement` using `version="@VERSION@"`. It helps when changing the version

tools/ghm/ghm.xml
```diff
@@ -0,0 +1,467 @@
...
+                   label="Type of NPL scoring">
+                <option value="all" selected="true" />
+                <option value="pairs" />
+                <option value="hom" />
+            </param>
+        </xml>
+    </macros>
+    <requirements>
+        <requirement type="package" version="3.0" >ghm</requirement>
+    </requirements>
```
bebatut: Can you add the `version_command` section please?

tools/ghm/ghm.xml
```diff
@@ -0,0 +1,467 @@
...
+#end if
+
+&& mv linkage.allegro_lod '${fparam}'
+#if $section_haplo.analysis_haplo.extra_haplotype
+&& mv linkage.allegro_haplo '${ihaplo}'
+#end if
+    ]]>
+    </command>
+    <configfiles>
+        <configfile name="setup_file" >
```
bebatut: I think the `CDATA` should be used here also

tools/ghm/ghm.xml
```diff
@@ -0,0 +1,467 @@
...
+    </configfiles>
+    <inputs>
+        <param name="inp_ped" type="data" format="linkage_pedin" label="Pedigree" />
+        <param name="inp_dat" type="data" format="linkage_datain" label="Recombination Freqs" />
+        <param name="inp_map" type="data" format="linkage_map" label="Marker Positions" />
+
+        <section name="section_haplo" title="Haplotypes" expanded="true" >
+            <conditional name="analysis_haplo" >
+                <param name="extra_haplotype" type="boolean" truevalue="on" falsevalue="off" value="false"
+		       label="Haplotype Analysis" />
```
bebatut: Can you put the param only one line? thanks :smile: (it is helping to review)

bebatut: Can you change `value` by `checked`?

tools/ghm/ghm.xml
```diff
@@ -0,0 +1,467 @@
...
+                <when value="on">
+                    <param name="extra_haplotype_method" type="select" label="Haplotype Reconstruction Algorithm. MaxProb is fastest, and has more global solution." >
+                        <option value="MaxProb" selected="true" />
```
bebatut: An option should be `<option value="value">Text</option>`

tools/ghm/ghm.xml
```diff
@@ -0,0 +1,467 @@
...
+                               label="Calculate P-values only at the best LOD positions" />
+                    </when>
+                    <when value="global" />
+                    <when value="off" />
+                </conditional>
+
+                <param name="extra_seq" type="integer" value="0" min="0"
+                       label="Number of sequential simulations in P-value calculations" />
+                <param name="extra_storereplicates" type="select" label="Store replicates during P-value calculations" >
+                    <option value="pre" >pre</option>
```
bebatut: Have all the options the same text?

tools/ghm/ghm.xml
mtekman: Hey thanks, I am a little confused by this tag -- does it override the version parameter set in `<tool>`? Is it supposed to match the version given by `<requirement>`?

tools/ghm/.shed.yml
```diff
@@ -0,0 +1,7 @@
+name: ghm
```
bgruening: maybe genehunter_modscore?

tools/ghm/ghm.xml
```diff
@@ -1,22 +1,51 @@
-<tool id="ghm" name="Genehunter-Modscore" version="3.0" >
+<tool id="ghm" name="Genehunter-Modscore" version="3.0.0" >
```
bgruening: the id could be as well `genehunter_modscore`

bgruening: as version you could take @VERSION@.0

tools/ghm/ghm.xml
```diff
@@ -1,22 +1,51 @@
...
+                <assert_contents>
+                    <has_text_matching expression="1\s+206006\s+206001\s+206002\s+2\s+2(\s+[0-2])+" />
+                </assert_contents>
+            </output>
+        </xml>
+        <!-- End Output test file(s) -->
     </macros>
     <requirements>
-        <requirement type="package" version="3.0" >ghm</requirement>
+        <requirement type="package" version="@VERSION@.0" >ghm</requirement>
```
bgruening: isn't the name `linkage2allegro` and the version 2017.3?

tools/ghm/ghm.xml
mtekman: Macros work on their parent?

bgruening:
Cool thanks a lot @mtekman!

PR #1598, opened 2017-11-21
Add Swiftlink 1.0

mtekman:
Swiftlink is the most up-to-date linkage analysis tool (2013). It currently lacks haplotype analysis, but more than makes up for it with fast parallel linkage analysis.

Experimental support for X-linked analyses.


tools/swift/swift.xml
```diff
@@ -0,0 +1,115 @@
...
+            <param argument="-b" type="integer" min="1" value="50000" label="Burn-ins" />
+            <param argument="-s" type="integer" min="1" value="1000" label="Sequential Imputation" />
+            <param argument="-x" type="integer" min="1" value="10" label="Scoring Period" />
+            <param argument="-n" type="integer" min="1" value="5" label="Lod Scores" />
+            <param argument="-R" type="integer" min="1" value="1" label="Runs" />
+            <param argument="-l" type="float" min="0" max="1" value="0.5" label="L-Sampler Probability" />
+        </section>
+
+        <section name="elod" title="Expected LOD (ELOD) Options" expanded="true" >
+            <param argument="-e" type="boolean" truevalue="-e" falsevalue="" value="false"
```
bebatut: Can you use a conditional and a selector here? Because the possible parameters depends on the choice for `e` and it is also a good practice (http://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#dynamic-options). Sth like

```
<conditional name="sth">
    <param argument="-e" type="select" label="Calculate Expected LOD scores from pedigree only.">
        <option value="true">Yes</option>
        <option value="false">No</option>
    </param>
    <when value="true">
        <param argument="-f" type="float" min="0" max="1" value="1.0e-04" label="Frequency" />
        ...
    </when>
    <when value="false">
        <param argument="-X" type="boolean" truevalue="-X" falsevalue="" value="false" label="X-linked"/>
        ...
    </when>
</conditional>
```

tools/swift/swift.xml
```diff
@@ -0,0 +1,132 @@
...
+        <section name="misc" title="Miscellaneous" expanded="true" >
+            <param argument="-X" type="boolean" truevalue="-X" falsevalue="" value="false" label="X-linked"/>
+            <param argument="-a" type="boolean" truevalue="-a" falsevalue="" value="false" label="Affecteds only"/>
+            <param argument="-q" type="integer" min="1" value="1000000" label="Peel-sequence Iterator"/>
+            <param argument="-r" type="data" format="txt" label="Random seeds (single-column)" optional="true" />
+        </section>
+    </inputs>
+
+    <outputs>
+        <data name="out_fparam" format="allegro_fparam" label="${tool.name} on ${on_string}: MPT Linkage File" />
```
bgruening: Both outputs are now created everytime, but one of them is always empty right?
I think you need to use `<filter>` on both and filter for elod_opts.e

mtekman: Makes sense and very handy, thanks

tools/swift/swift.xml
```diff
@@ -2,6 +2,11 @@
     <description>Parallel MCMC Linkage Analysis</description>
     <macros>
         <token name="@VERSION@">1.0</token>
+        <xml name="macro_inputs" >
```
bgruening: test_inputs? it should be clear its a macro, right?

mtekman: hah true, renaming

bgruening:
@mtekman awesome! Thanks a bunch!

mtekman:
thanks again @bgruening @bebatut

PR #1599, opened 2017-11-21
Migrate BWA to tools-iuc

mvdbeek:
bwa (and especially bwa-mem) is still very widely used and actively developed.
We're already maintaining tools like bwameth in here, so I think is makes sense
to migrate it here.


bgruening:
@martenson can you please grant IUC permissions? :)

martenson:
@bgruening done 👍

bgruening:
Danke @mvdbeek!

tools/bwa/bwa-mem.xml
```diff
@@ -0,0 +1,401 @@
...
+    -v 1                                                                                      ## Verbosity is set to 1 (errors only)
+
+    #if str( $fastq_input.fastq_input_selector ) == "paired_iv":                              ## For interleaved fastq files set -p option
+      -p
+      #if str( $fastq_input.iset_stats ):                                                     ## check that insert statistics is used
+        -I "${fastq_input.iset_stats}"
+      #end if
+    #end if
+
+    #if str( $analysis_type.analysis_type_selector ) == "pacbio":
```
yhoogstrate:
```
#if str( $analysis_type.analysis_type_selector ) == "full":
  ...
#else
    -x '${analysis_type.analysis_type_selector}'
#end if
```

nsoranzo: @yhoogstrate This PR has been merged already, do you want to open a new PR?

mvdbeek: I thought that's a good suggestion at first, but there's no `-x illumina`. I'll open a PR with some other improvements (we use single-threaded samtools sorting ... Can easily shave down 25 % of the time here).

nsoranzo: Maybe you can also clean up indentation while you're at it.

PR #1600, opened 2017-11-23
Update Discosnp to v2.3.0

cmonjeau:


tools/colibread/discosnp_pp.xml
```diff
@@ -1,11 +1,11 @@
 <?xml version='1.0' encoding='utf-8'?>
-<tool profile="16.04" id="discosnp_pp" name="DiscoSnp++" version="2.2.10">
+<tool profile="16.04" id="discosnp_pp" name="DiscoSnp++" version="2.3.0">
     <description>is an efficient tool for detecting SNPs without a reference genome.</description>
     <macros>
         <import>macros.xml</import>
     </macros>
     <requirements>
-        <requirement type="package" version="2.2.10">discosnp</requirement>
+        <requirement type="package" version="2.3.0">discosnp</requirement>
```
bgruening: Do you want to put the version number also in a macro? You can use it here and in the tool version.
Also the entire requirement could move into the macro as well as citation and parts of the help?

tools/colibread/discosnp_RAD.xml
```diff
@@ -0,0 +1,112 @@
...
+            @discosnp_paired_reads@
+        #end if
+
+        ## get the absolute path for short_reads_connector
+        CONNECTOR_PATH=`which short_read_connector.sh` &&
+        CONNECTOR_DIR=`dirname \$CONNECTOR_PATH` &&
+
+        run_discoSnpRad.sh
+        @discosnp_inputs@
+        -S \$CONNECTOR_DIR
```
bgruening: Is this folder just used for reading or also for writing?

bgruening: I guess it needs to be double quoted in cases where the path contains white spaces.

cmonjeau: The folder is just for reading. I added the double quoted.

tools/colibread/discosnp_pp.xml
cmonjeau: Done

bgruening:
Nice thanks a bunch @cmonjeau!

bgruening:
@cmonjeau for some reason the tests failed on master. Can you look at it? https://travis-ci.org/galaxyproject/tools-iuc/jobs/306399323

nsoranzo:
@cmonjeau To clarify, that means that the `colibread` repositories were not updated on the ToolShed.

PR #1603, opened 2017-11-24
Bowtie2 upgrade, new options and samtools multicore

mvdbeek:


tools/bowtie2/bowtie2_wrapper.xml
```diff
@@ -379,7 +373,6 @@
             <param name="own_file" type="data" format="fasta" label="Select reference genome" />
```
nsoranzo: According to the Version 2.3.3 release notes at http://bowtie-bio.sourceforge.net/bowtie2/index.shtml : "bowtie2-build now supports gzip-compressed FASTA inputs", so we could add `fasta.gz` here.

tools/bowtie2/bowtie2_wrapper.xml
```diff
@@ -4,321 +4,324 @@
...
+        ${analysis_type.input_options.solexa_quals}
+        ${analysis_type.input_options.int_quals}
+    #end if
+
+    #if str( $analysis_type.alignment_options.alignment_options_selector ) == "yes":
+        -N ${analysis_type.alignment_options.N}
+        -L ${analysis_type.alignment_options.L}
+        -i '${analysis_type.alignment_options.i}'
+        --n-ceil '${analysis_type.alignment_options.n_ceil}'
+        --dpad '${analysis_type.alignment_options.dpad}'
```
nsoranzo: You can remove the quotes, it's an integer.

tools/bowtie2/bowtie2_wrapper.xml
```diff
@@ -4,321 +4,324 @@
...
+        --gbar '${analysis_type.alignment_options.gbar}'
```
nsoranzo: You can remove the quotes, it's an integer.

tools/bowtie2/bowtie2_wrapper.xml
mvdbeek: but we don't have a fasta.gz datatype for now, right ?

tools/bowtie2/bowtie2_wrapper.xml
```diff
@@ -4,321 +4,324 @@
...
-            #set _aligned_root, _aligned_ext = splitext( str( $output_aligned_reads_l ) )
-            && mv "${ _aligned_root }.1${_aligned_ext}" '$output_aligned_reads_l'
-            && mv "${ _aligned_root }.2${_aligned_ext}" '$output_aligned_reads_r'
+    #if str( $analysis_type.scoring_options.scoring_options_selector ) == "yes":
+        #if ( str( $analysis_type.alignment_options.alignment_options_selector ) == "yes" and str( $analysis_type.alignment_options.align_mode.align_mode_selector ) == "local" ):
+            --ma ${analysis_type.scoring_options.ma}
         #end if
+        --mp '${analysis_type.scoring_options.mp}'
+        --np ${analysis_type.scoring_options.np}
+        --rdg ${analysis_type.scoring_options.rdg_read_open},${analysis_type.scoring_options.rdg_read_extend}
```
nsoranzo: Here you need to re-add the single quotes, there are 2 options separated by a comma.

tools/bowtie2/bowtie2_wrapper.xml
```diff
@@ -4,321 +4,324 @@
...
+        --rfg ${analysis_type.scoring_options.rfg_ref_open},${analysis_type.scoring_options.rfg_ref_extend}
```
nsoranzo: Here you need to re-add the single quotes, there are 2 options separated by a comma.

tools/bowtie2/bowtie2_wrapper.xml
mvdbeek: Hmm, are you sure? What does the quote change ?

tools/bowtie2/bowtie2_wrapper.xml
nsoranzo: Uh, I assumed we have. Probably I was confused by the fact that there is a (partial) support for compressed FASTA files in galaxy_sequence_utils.

tools/bowtie2/bowtie2_wrapper.xml
nsoranzo: You're right, the quotes shouldn't be needed anyway.

nsoranzo:
Last thing would be to increase the tool version.

mvdbeek:
Alright, as always, thanks for the thorough review @nsoranzo !

tools/bowtie2/bowtie2_wrapper.xml
mvdbeek: It would make sense to support fasta.gz, but I'm a bit hesitant to do this just yet in galaxy, I think the fastq.gz transition went not as smooth as I had hoped.

nsoranzo: The framework is there now, should be easier/smoother this time.

nsoranzo:
Thanks @mvdbeek!

PR #1604, opened 2017-11-24
Colibread: fix symlink creation outside working dir

abretaud:
While looking at the failing test in #1600, I found out that the colibread tools are creating symlinks outside the working dir, so here's a fix for this problem
I still don't understand why the tests in #1600 failed, maybe it's related? Let's see if it goes green

ping @cmonjeau


nsoranzo:
For this kind of fixes you don't need to increase the version number, if you prefer.

abretaud:
Ah yes, thanks @nsoranzo !

cmonjeau:
Thanks @abretaud and @nsoranzo !!

PR #1610, opened 2017-11-28
only run the deploy step on master

bgruening:


nsoranzo:
Nice, I didn't know that was possible!

Issue #1581, opened 2017-11-16
MultiQC for featurecounts fails when using default dataset names


lparsons:
It seems that when metadata column names for featureCounts summary output has commas (','), MultiQC will fail. This line (https://github.com/galaxyproject/tools-iuc/blob/master/tools/multiqc/multiqc.xml#L111) attempts to do some cleanup on the metadata column names and reinsert them back into the output (they seem to have been scrubbed here (https://github.com/galaxyproject/tools-iuc/blob/master/tools/featurecounts/featurecounts.xml#L100). MultiQC requires that the file be intact (as output from featureCounts) and so the MultiQC wrapper must "put it back together". However, the substitute of tabs for comma's causes problems.

As an aside, I've noticed that this same error is made elsewhere in Galaxy (e.g. the display of tabular data with commas in the metadata column names is messed up as the commas have been changed to tabs). See https://github.com/galaxyproject/galaxy/issues/5026.

lparsons:
@bebatut Do you have any thoughts on how to do this substitution properly? I've not dealt with column metadata before and I'm not really sure what is actually looks like (nor how to easily debug this issue).

dpryan79:
Honestly, I think [this whole if statement](https://github.com/galaxyproject/tools-iuc/blob/master/tools/multiqc/multiqc.xml#L110-L113) can be removed. That would make sense if featureCounts in Galaxy were written to output count files containing multiple samples, but currently those are all batch jobs, so such a file can't/won't be created. My suggestion would be to use the following code:

    #elif str($repeat.software_cond.software) == "featureCounts"
        #for $file in $repeat.software_cond.input
            #set file_prefix = $software_dir + '/' + str($file.element_identifier)
            ln -s '$file' '${file_prefix}.summary' &&
        #end for

Presumably one can get rid of the `file_prefix` all together.

PR #1611, opened 2017-12-01
Fix a typo in the multiqc --version command

dpryan79:
Currently, it runs `mulitqc --version` rather than `multiqc --version`.


dpryan79:
I'm running into #1581 now, I'll try to fix that real quick too (though it might make sense to just move to version 1.3 and see if the issue is gone there).

dpryan79:
Meh, for now I'll just relabel the items in the workflows rather than play around with the sanitizer settings.

mvdbeek:
great, thanks @dpryan79 !

PR #1612, opened 2017-12-01
Update orthofinder

Mataivic:
I found that in some cases under Galaxy, the input files for OrthoFinder miss a .fasta/.faa/.fa extension, causing a crash of OrthoFinder.

For example, I discovered this problem when I tried to send an output collection from TransDecoder to OrthoFinder (the collection is know to contain fasta files, but the filenames in the history were missing an explicit '.fasta' extension)

 Here is an updated wrapper which automatically add a ".fasta" extension to each input file.


tools/orthofinder/orthofinder_only_groups.xml
```diff
@@ -9,7 +9,7 @@
         #if $init.start=="fasta":
             #set $infiles = ""
             #for $input in $init.input_fasta
-                ln -s '$input' '$input.element_identifier' &&
+                ln -s '$input' '$input.element_identifier'.fasta &&
```
nsoranzo:
```
ln -s '$input' '${input.element_identifier}.fasta' &&
```

bgruening:
Thanks @Mataivic!

PR #1614, opened 2017-12-01
Change 'genome' to 'transcriptome'

davebx:
Per suggestion by @jxtx, it would be more accurate to refer to the reference input as a transcriptome.

### DIFF
--- a/tools/kallisto/kallisto_pseudo.xml
+++ b/tools/kallisto/kallisto_pseudo.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<tool id="kallisto_pseudo" name="Kallisto pseudo" version="@VERSION@.0">
+<tool id="kallisto_pseudo" name="Kallisto pseudo" version="@VERSION@.1">
     <description>- run pseudoalignment on RNA-Seq transcripts</description>
     <macros>
         <import>macros.xml</import>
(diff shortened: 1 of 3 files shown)


mblue9:
This looks good to me. Wonder if people think it would be good to add to the help or somewhere some tips on how to make the transcriptome file? As if you use pizzly after kallisto the transcriptome fasta and gtf need to match, and I don't know what anyone else did, but after a little trial and error we ended up generating a transcriptome with this 2 step procedure e.g. for hg38:

1) gffread -W -G -w Homo_sapiens.GRCh38.90.transcripts.fa -g Homo_sapiens.GRCh38.dna.toplevel.fa Homo_sapiens.GRCh38.90.gtf

2) a sed command to add the transcript version to the ENST id in the fasta headers which pizzly needs 

I tried a couple of other ways to get the transcriptome but then went with gffread because:
1) Homo_sapiens.GRCh38.cdna.all.fa (the Ensembl provided cDNA fasta) doesn't match the Ensembl gtf e.g. it doesn't include ncRNA and includes haplotypes like CHR_HSCHR8_4_CTG7, see here: http://seqanswers.com/forums/showthread.php?t=72828

2) gtf_to_fasta (a Tophat helper script) didn't reverse-complement the sequence for transcripts on the negative strand which affects stranded data, see here:
https://github.com/chapmanb/bcbio-nextgen/issues/2025

bgruening:
Thanks @davebx!
@mblue9 I think we have gffread in Galaxy. Would be nice to add a help to all these tools kallisto/salmon and so on describing how to use the Galaxy tool.

PR #1615, opened 2017-12-01
Fix #1581, bump multiqc wrapper version

### LINKED ISSUE #1581: MultiQC for featurecounts fails when using default dataset names

dpryan79:
This primarily fixes #1581. The initial implementation of this wrapper had issues in part due to the `Status` line being stripped in the featureCounts wrapper (which is itself due to `featureCounts` not using the `element_identifier`). This PR does two things:

1. It uses the `element_identifier` if the `Status` line is empty. This respects whatever people have labeled things and plays well with dataset collections (at least in my test, since I'm using this now in our local installation).
2. If a user uploads a featureCounts summary file, then it leaves that untouched and all file names that the user originally used (outside of Galaxy) will still be displayed as normal.

Thankfully, the `featureCounts` wrapper in Galaxy doesn't run featureCounts on multiple files at once (hopefully it will never support this since it would also break the DESeq2 wrapper).

### DIFF
--- a/tools/multiqc/multiqc.xml
+++ b/tools/multiqc/multiqc.xml
@@ -1,7 +1,7 @@
 <tool id="multiqc" name="MultiQC" version="@WRAPPER_VERSION@.0">
     <description>aggregate results from bioinformatics analyses into a single report</description>
     <macros>
-        <token name="@WRAPPER_VERSION@">1.2</token>
+        <token name="@WRAPPER_VERSION@">1.2.1</token>
         <token name="@LN_FILES@">
 <![CDATA[
 #for $file in $repeat.software_cond.input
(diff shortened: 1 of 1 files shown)


dpryan79:
@lparsons Can you confirm that this solves the issues that you've observed with the featureCounts module and default history item names?

mvdbeek:
Thanks a lot @dpryan79 & @lparsons!

Issue #1616, opened 2017-12-04
Bedtools coverage cannot iterate over multiple bam files... it always reduces


bwlang:

bwlang:
I'm working on a patch for this now...

bwlang:
I'm working on a patch for this now...

bwlang:
fixed in #1617

PR #1617, opened 2017-12-04
add reduce/iterate conditional to bedtools

bwlang:
This is intended to allow bedtools to create multiple output files (one for each input B)

### DIFF
--- a/tools/bedtools/coverageBed.xml
+++ b/tools/bedtools/coverageBed.xml
@@ -1,4 +1,4 @@
-<tool id="bedtools_coveragebed" name="Compute both the depth and breadth of coverage" version="@WRAPPER_VERSION@.2">
+<tool id="bedtools_coveragebed" name="Compute both the depth and breadth of coverage" version="@WRAPPER_VERSION@.1">
     <description>of features in file B on the features in file A (bedtools coverage)</description>
     <macros>
         <import>macros.xml</import>
--- a/tools/bedtools/coverageBed.xml
+++ b/tools/bedtools/coverageBed.xml
@@ -23,7 +23,7 @@
         $reciprocal_overlap
         $a_or_b
         -a '$inputA'
-        -b 'inputBs'
+        -b '$inputBs'
         | sort -k1,1 -k2,2n
         > '$output'
 ]]>
--- a/tools/bedtools/coverageBed.xml
+++ b/tools/bedtools/coverageBed.xml
@@ -7,7 +7,7 @@
     <expand macro="stdio" />
     <command>
 <![CDATA[
-        #set inputBs = "' '".join([str($file) for $file in $inputB])
+        #set inputBs = "' '".join([str($file) for $file in $reduce_or_iterate.inputB])
 
         bedtools coverage
         $d
@@ -23,7 +23,7 @@
         $reciprocal_overlap
         $a_or_b
         -a '$inputA'
-        -b '$reduce_or_iterate.inputBs'
+        -b 'inputBs'
         | sort -k1,1 -k2,2n
         > '$output'
 ]]>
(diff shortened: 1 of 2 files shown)


bwlang:
This does not currently work correctly in galaxy (only a single output file is generated even though my dataset collection contains two items.  Also the workflow output noodle is rendered as a single item... I'm not sure what I'm missing here.  
Any ideas?

tools/bedtools/coverageBed.xml
```diff
@@ -7,7 +7,7 @@
     <expand macro="stdio" />
     <command>
 <![CDATA[
-        #set inputBs = "' '".join([str($file) for $file in $inputB])
+        #set inputBs = "' '".join([str($file) for $file in $reduce_or_iterate.inputB])
```
mvdbeek: when `reduce_or_iterate` is `iterate`  `$reduce_or_iterate.inputB` is the path to the file, not an iterable of datasets, so `inputBs` should just be '$reduce_or_iterate.inputB'.

Not sure that fixes all of your issues though.

bwlang:
ahhh -thanks

tools/bedtools/coverageBed.xml
bwlang: I thought that would be it, but I'm still seeing only a single job running. I guess now that I think about it the inputBs should be ok... it will just be a list of 1 item for the iterate case.  Right?
If that's right, i can avoid adding a reduce/iterate conditional to the command line builder.

bwlang:
image
Should the  noodle coming out of the tool should be thick if i got the syntax for iterate/reduce right?

tools/bedtools/coverageBed.xml
jmchilton: @mvdbeek Is correct about this being an issue as it stands. I'm not sure if it is the only issue. Hmm...

jmchilton:
The output noodle should be think if the input noodle is thick and if iterate is chosen. If you changed the tool without changing the workflow though you may need to disconnect everything and reconnect?

bwlang:
@jmchilton I tried the disconnect/reconnect trick.... no joy. 
I did also tried changing the order of inputA and inputB in the parameter list, that flipped them in the tool but provided the same thin noodle and single output when I reconnected everything.
I'm pretty stumped... next step - try to find the place in the galaxy workflow code that decides thick vs thin noodles...

bwlang:
I had the sense wrong... this version is working fine in my system now.
Ready to merge?

bwlang:
fixes #1616

nsoranzo:
@bwlang So WIP can be removed from the PR title, right?

bwlang:
@nsoranzo  yep - thought i already did, but must have forotten to save.

tools/bedtools/coverageBed.xml
```diff
@@ -23,14 +21,32 @@
...
+            -b '$inputBs'
+        #end if
         | sort -k1,1 -k2,2n
         > '$output'
 ]]>
     </command>
     <inputs>
         <param format="bam,bed,gff,gff3,vcf" name="inputA" type="data" label="File A (on which coverage is calculated)" help="BAM/BED/GFF/VCF format" />
-        <param format="bam,bed,gff,gff3,vcf" name="inputB" type="data" multiple="true" label="File(s) B (for which coverage is calculated)" help="BAM/BED/GFF/VCF format" />
+        <conditional name="reduce_or_iterate">
```
bgruening: This could go into the macro isn't it? Also the cheetah part.

bwlang: I could not find generic enough help language that made sense  for both coverage and intersect (and any future case)... so I did not put it in the macro.  I think users need the extra help to understand which is A and which is B so I thought it's worth the duplication. Thoughts?

bwlang:
@bgruening  just rebased... hope that makes this easier to merge.

bwlang:
@bgruening  Let me know if you'd like other changes to this PR

bgruening:
Thanks @bwlang!

bgruening:
Sorry, that it took that long :(

bwlang:
no problem... the world is a busy place :)

PR #1620, opened 2017-12-06
Incorporate enhancements from devteam's wrapper.

davebx:
Implements the `$GALAXY_MEMORY_MB` variable added in galaxyproject/galaxy#4958


tools/trinity/trinity.xml
```diff
@@ -67,9 +77,18 @@
         #end if
 
         ## CPU and butterfly options.
-        --CPU \${GALAXY_SLOTS:-4} --max_memory \${TRINITY_MAX_MEMORY:-1G} --bflyHeapSpaceMax \${TRINITY_MAX_MEMORY:-1G} --bfly_opts '-V 10 --stderr'
+        --CPU \${GALAXY_SLOTS:-4} --max_memory \${GALAXY_MEMORY_GB:-1}G --bflyHeapSpaceMax \${GALAXY_MEMORY_GB:-1}G --bfly_opts '-V 10 --stderr'
```
mvdbeek: This doesn't support M instead of G ?

davebx: It does not. I was quite irritated when I saw this in the test log:

`Error, cannot parse jellyfish_ram value of 1024M.  Set it to 'xG' where x is a numerical value`

mvdbeek:
You'll have to increase the profile to 18.01

tools/trinity/trinity.xml
```diff
@@ -1,11 +1,21 @@
...
+<tool id="trinity" name="Trinity" version="@WRAPPER_VERSION@.0" profile="18.01">
     <description>de novo assembly of RNA-Seq data</description>
     <macros>
         <import>macros.xml</import>
     </macros>
     <expand macro="requirements" />
     <command detect_errors="aggressive"><![CDATA[
-       Trinity --no_version_check
+        if [ ! -z \$GALAXY_MEMORY_MB ] ; then
+            GALAXY_MEMORY_GB = 1 ;
```
nsoranzo: Remove spaces around `=`

tools/trinity/trinity.xml
```diff
@@ -1,11 +1,21 @@
-<tool id="trinity" name="Trinity" version="@WRAPPER_VERSION@.0">
+<tool id="trinity" name="Trinity" version="@WRAPPER_VERSION@.0" profile="18.01">
     <description>de novo assembly of RNA-Seq data</description>
     <macros>
         <import>macros.xml</import>
     </macros>
     <expand macro="requirements" />
     <command detect_errors="aggressive"><![CDATA[
-       Trinity --no_version_check
+        if [ ! -z \$GALAXY_MEMORY_MB ] ; then
```
nsoranzo: I think that the `!` shouldn't be there. Also, it's better to quote the env variable:
```bash
        if [ -z "\$GALAXY_MEMORY_MB" ] ; then
```

tools/trinity/trinity.xml
```diff
@@ -1,11 +1,21 @@
...
+            GALAXY_MEMORY_GB = 1 ;
+        else
+            GALAXY_MEMORY_GB `expr \$GALAXY_MEMORY_MB / 1024` ;
```
nsoranzo: Missing `=`. Also `$((..))` is recommended over `expr`:
```bash
            GALAXY_MEMORY_GB=\$((GALAXY_MEMORY_MB / 1024)) ;
```

tools/trinity/trinity.xml
```diff
@@ -1,11 +1,21 @@
...
+        fi ;
+
+        if [ ! -z \$TRINITY_SCRATCH_DIR ] ; then
+            cd \$TRINITY_SCRATCH_DIR ;
```
nsoranzo: Double-quotes around `\$TRINITY_SCRATCH_DIR` in the previous 2 lines.

tools/trinity/trinity.xml
```diff
@@ -67,9 +77,18 @@
...
 
         ## CPU and butterfly options.
-        --CPU \${GALAXY_SLOTS:-4} --max_memory \${TRINITY_MAX_MEMORY:-1G} --bflyHeapSpaceMax \${TRINITY_MAX_MEMORY:-1G} --bfly_opts '-V 10 --stderr'
+        --CPU \${GALAXY_SLOTS:-4} --max_memory \${GALAXY_MEMORY_GB:-1}G --bflyHeapSpaceMax \${GALAXY_MEMORY_GB:-1}G --bfly_opts '-V 10 --stderr'
 
         ## > $trinity_log 2>&1
+
+        &&
+
+        if [ ! -z \$TRINITY_SCRATCH_DIR ] ; then
```
nsoranzo: Double-quotes around `\$TRINITY_SCRATCH_DIR`

tools/trinity/trinity.xml
```diff
@@ -67,9 +77,18 @@
...
+            mkdir -p "\$workdir/trinity_out_dir";
```
nsoranzo: Is `$workdir` an env variable defined by Galaxy? It's not documented on https://docs.galaxyproject.org/en/latest/dev/schema.html#reserved-variables

nsoranzo:
> You'll have to increase the profile to 18.01

@mvdbeek Is that really necessary? The tool still works if the env variables are not defined, and admins of older releases can set `GALAXY_MEMORY_GB` in `job_conf.xml` as they were doing before with `TRINITY_MAX_MEMORY`.

mvdbeek:
You're right, that isn't necessary.

davebx:
I'll remove the profile, then. Which will also rerun the test that timed out on conda env creation.

tools/trinity/trinity.xml
```diff
@@ -1,11 +1,22 @@
...
     <macros>
         <import>macros.xml</import>
     </macros>
     <expand macro="requirements" />
     <command detect_errors="aggressive"><![CDATA[
-       Trinity --no_version_check
+        if [ -z "\$GALAXY_MEMORY_MB" ] ; then
+            GALAXY_MEMORY_GB=1 ;
+        else
+            GALAXY_MEMORY_GB=\$((expr \$GALAXY_MEMORY_MB / 1024)) ;
```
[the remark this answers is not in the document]
davebx: Oh, my mistake.

nsoranzo:
Some tests are failing due to https://github.com/conda/conda/issues/5536#issuecomment-350067404 .

tools/trinity/trinity.xml
```diff
@@ -1,11 +1,22 @@
...
     <macros>
         <import>macros.xml</import>
     </macros>
     <expand macro="requirements" />
     <command detect_errors="aggressive"><![CDATA[
-       Trinity --no_version_check
+        if [ -z "\$GALAXY_MEMORY_MB" ] ; then
+            GALAXY_MEMORY_GB=1 ;
+        else
+            GALAXY_MEMORY_GB=\$((GALAXY_MEMORY_MB / 1024)) ;
```
natefoo: The amount used needs to have a buffer under `$GALAXY_MEMORY_MB` or else it'll still hit the limit. Maybe we should change `$GALAXY_MEMORY_MB` to include a buffer so it can be used by tools directly rather than doing it on a per-tool basis. And `$GALAXY_MEMORY_HARD_MB` could be the hard limit in case it's relevant to a tool author.

tools/trinity/trinity.xml
natefoo: On Bridges, we set:

```
                --bflyHeapSpaceMax "32G"
                --bflyHeapSpaceInit "2G"
                --bflyGCThreads "6"
```

Not sure about the second two but the first is because I believe there was no benefit (and perhaps even a penalty?) for allowing more than 32GB for that option. If so, we could set it to `$GALAXY_MEMORY_GB` if < 32 and then 32 if `$GALAXY_MEMORY_GB` is greater than 32.

Maybe @pdblood can comment on whether these are still recommendations for the latest versions of Trinity.

davebx: @natefoo a problem with this is that for some reason, as I recall, one of the submodules sets the *maximum* memory value to the amount specified with the `--bflyHeapSpaceInit` parameter

nsoranzo:
@davebx I've rebased this to make it cleaner. Can you update `tools/trinity/README.rst` for these changes?

tools/trinity/trinity.xml
davebx: I imagine it would be sufficient to set `$GALAXY_MEMORY_GB` to one GB less than `$GALAXY_MEMORY_MB` / 1024, with a minimum of 1GB.

bgruening:
@nsoranzo is this ready to go?

bgruening:
Thanks a bunch @davebx!

PR #1622, opened 2017-12-07
Upgrade bedtools to 2.27.0

mvdbeek:


jmchilton:
It shouldn't be this easy - people will get lazy. Please make it look more difficult!

mvdbeek:
I had to dig through a bit of C++ to fix wrong tabs in 2.26.0, does that count ?
There are some new options, I'm checking if they make sense in galaxy.

bgruening:
Thanks @mvdbeek!

PR #1626, opened 2017-12-09
remove unused data table from the BWA wrappers

bgruening:


tools/bwa/bwa-mem.xml
```diff
@@ -1,13 +1,12 @@
...
+<tool id="bwa_mem" name="Map with BWA-MEM" version="@VERSION@.2">
     <description>- map medium and long reads (&gt; 100 bp) against reference genome</description>
     <macros>
         <import>read_group_macros.xml</import>
         <import>bwa_macros.xml</import>
     </macros>
     <expand macro="requirements"/>
-    <expand macro="stdio"/>
-    <command>
+     <command detect_errors="aggressive">
```
nsoranzo: Indentation

tools/bwa/bwa.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<tool id="bwa" name="Map with BWA" version="@VERSION@.3">
+<tool id="bwa" name="Map with BWA" version="@VERSION@.4">
```
nsoranzo: This tool doesn't require a version bump.

nsoranzo: Or probably it's the other one that doesn't need it. It may be less confusing if the remaining macro would keep the `dbKeyActionsBwaMem` name.

bgruening: Both tools are in one repo. If we don't increase the version number, Galaxy will still have a new version of this tool right? But this time two version with the same version number?

PR #1629, opened 2017-12-11
Fix macros.xml when input was *_bgzip files

Takadonet:
Forgot two lines where  '&&' were missing.

### DIFF
--- a/tools/bcftools/macros.xml
+++ b/tools/bcftools/macros.xml
@@ -49,7 +49,7 @@ export BCFTOOLS_PLUGINS=`which bcftools | sed 's,bin/bcftools,libexec/bcftools,'
   bgzip -c '$input_file' > $input_vcf &&
   bcftools index $input_vcf &&
 #elif $input_file.is_of_type('vcf_bgzip')
-  ln -s '$input_file' $input_vcf
+  ln -s '$input_file' $input_vcf &&
 #elif $input_file.is_of_type('bcf')
   #set $input_vcf = 'input.bcf'
   ln -s '$input_file' $input_vcf &&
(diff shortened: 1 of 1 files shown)


tools/bcftools/macros.xml
```diff
@@ -59,7 +59,7 @@ export BCFTOOLS_PLUGINS=`which bcftools | sed 's,bin/bcftools,libexec/bcftools,'
     bcftools index $input_vcf &&
   #end if
 #elif $input_file.is_of_type('bcf_bgzip')
-  ln -s '$input_file' $input_vcf
+  ln -s '$input_file' $input_vcf &&
```
bgruening: This looks wrong. The command would end with `&&` which will yield to an exception.

Takadonet: So running without this change. I get the following error when converting from a bcf_bgzip to vcf

```
Fatal error: Exit code 1 ()
ln: unrecognized option '--output-type'
Try 'ln --help' for more information.
```
Below is the command that is being build. The '&&' is missing between the ln  command and bcftools view
```
export BCFTOOLS_PLUGINS=`which bcftools | sed 's,bin/bcftools,libexec/bcftools,'`;     ln -s '/galaxy/data/000/361/dataset_361296.dat' input.vcf.gz            bcftools view                          --output-type 'v'   --threads ${GALAXY_SLOTS:-4}    input.vcf.gz  > '/galaxy/data/000/361/dataset_361732.dat'
```

nsoranzo: @bgruening This a `token`, not the whole `command`, so this is correct.

PR #1631, opened 2017-12-12
Exclude R from the channels used for testing.

davebx:
Workaround for conda/conda#5536 to properly test galaxyproject/tools-iuc#1620


nsoranzo:
Can you add the commit directly to #1620? So we can see if it fixes it.

bgruening:
Thanks!

PR #1635, opened 2017-12-15
Fix output type for compressed BCF

Takadonet:
Change format for compressed BCF to its corresponding extension of 'bcf_bgzip' Otherwise, you cannot output a compressed BCF at all.

### DIFF
--- a/tools/bcftools/macros.xml
+++ b/tools/bcftools/macros.xml
@@ -254,7 +254,7 @@ $vcfs_list_file
   <xml name="macro_vcf_output">
       <data name="output_file" format="vcf">
         <change_format>
-          <when input="output_type" value="b" format="bcf" />
+          <when input="output_type" value="b" format="bcf_bgzip" />
           <when input="output_type" value="u" format="bcf" />
           <when input="output_type" value="z" format="vcf_bgzip" />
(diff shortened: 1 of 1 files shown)


tools/bcftools/macros.xml
```diff
@@ -254,7 +254,7 @@ $vcfs_list_file
   <xml name="macro_vcf_output">
       <data name="output_file" format="vcf">
         <change_format>
-          <when input="output_type" value="b" format="bcf" />
+          <when input="output_type" value="b" format="bcf_bgzip" />
```
nsoranzo: In Galaxy 17.09 `bcf_bgzip` is just an alias for `bcf`, for details see https://github.com/galaxyproject/galaxy/pull/4477

Takadonet: and we are currently stuck on 17.05

bgruening: @Takadonet is this needed or can you use this with a local patch or update to 17.09?

Takadonet: I will update to 17.09

bgruening: Ok, thanks. Closing this PR now.

PR #1637, opened 2017-12-18
bwa: added a regular expression to stdio

bernt-matthias:
This happens when unexpected input is provided. For instance a fa.tar file is used (I also read that spaces in the sequence may cause this message).

### DIFF
--- a/tools/bwa/bwa_macros.xml
+++ b/tools/bwa/bwa_macros.xml
@@ -45,6 +45,7 @@
             <exit_code range=":-1" />
             <regex match="Error:" />
             <regex match="Exception:" />
+            <regex match="[bns_restore_core] Parse error reading" />
         </stdio>
     </xml>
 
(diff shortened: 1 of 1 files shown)


tools/bwa/bwa_macros.xml
```diff
@@ -45,6 +45,7 @@
             <exit_code range=":-1" />
             <regex match="Error:" />
             <regex match="Exception:" />
+            <regex match="[bns_restore_core] Parse error reading" />
```
mvdbeek: I think you'll have to escape the braces here, i.e `\[bns_restore_core\] Parse error reading`, otherwise the pattern doesn't match the text `[bns_restore_core] Parse error reading`

bernt-matthias: Sure (while it still should have matched .. but also on scrambled text).

bgruening:
Thanks!

PR #1638, opened 2017-12-18
initial commit of mykrobe tools

Takadonet:
- [x] Add help sections


tools/mykrobe/mykrobe_genotype.xml
```diff
@@ -110,6 +111,8 @@
 
 **MyKrobe predict - Antibiotic resistance predictions**
 
+Rapid antibiotic-resistance predictions from genome sequence data for Staphylococcus aureus and Mycobacterium tuberculosis
```
bgruening: Maybe a macro file would pay of there as well? requirements and citations could go in there as well.

tools/mykrobe/.shed.yml
```diff
@@ -0,0 +1,8 @@
+name: mykrobe
+owner: iuc
+description: Antibiotic resistance predictions
+long_description: Rapid antibiotic-resistance predictions from genome sequence data for Staphylococcus aureus and Mycobacterium tuberculosis
+categories:
```
bgruening: This would be better a suite definition, so that every tool gets its own repo, see here for an example: https://github.com/galaxyproject/tools-iuc/blob/643293a896a2ccfac10fe995f48c7f01c1a89a7f/tools/biom_format/.shed.yml

tools/mykrobe/mykrobe_genotype.xml
```diff
@@ -0,0 +1,122 @@
...
+  <command detect_errors="exit_code">
+<![CDATA[
+
+         #set $name=str($data_type.fastq_input1.name)
+         #set $type='fastq'
+         ## Adding sample name, indicate specie and inputs fastq(s) files
+          #if $data_type.type == "paired":
+
+            ln -s '$data_type.fastq_input1' '$data_type.fastq_input1.name'_1.fastq &&
+            ln -s '$data_type.fastq_input2' '$data_type.fastq_input2.name'_2.fastq &&
```
bgruening: Is it important to carry around the name for the tool? Using the name can have unwanted implications and is not encouraged :(
 
You could just name it first.fastq if this is possible?

tools/mykrobe/mykrobe_genotype.xml
```diff
@@ -0,0 +1,122 @@
...
+          *.bam
+        #else
+          *.fasta  
+        #end if
+         > $json
+]]>
+  </command>
+  <inputs>
+    <conditional name="data_type">
+      <param name="type" type="select" label="Specify the read type.">
```
bgruening: Perfect for a macro.xml file.

tools/mykrobe/mykrobe_genotype.xml
Takadonet: moved help, requirements and citations to macro.xml file

tools/mykrobe/mykrobe_genotype.xml
Takadonet: moved to macro.xml

tools/mykrobe/.shed.yml
Takadonet: changed to suite

tools/mykrobe/mykrobe_genotype.xml
Takadonet: It would be best to keep the name around since it gets directly embeeded into the json output file. Name is only really important when the user gives a collection.  

Do you have a better solution to this? I can change and to only keep name data_type.type is a collection

Takadonet: So changed it so that it will only use the name when user passes a paired collection and all other cases will use 'sample'.

Takadonet: Will this work for you  @bgruening ?

bgruening:
@Takadonet thanks! This looks like a very interesting tool!

PR #1642, opened 2017-12-20
add wrapper for umi_tools deduplication & update umi_tools version

dmaticzka:
this adds a wrapper for the command "umitools dedup" (in addition to the existing "umitools extract" and "umitools group".

the dedup command is very similar to the group command, however, group only marks groups and dedup actually extracts a representative alignment.

also updates umi_tools version from 0.5.0 to 0.5.1


tools/umi_tools/umi-tools_dedup.xml
```diff
@@ -0,0 +1,292 @@
+<tool id="umi_tools_group" name="UMI-tools group" version="@VERSION@.0">
```
bgruening: ID and name is wrong, right?

tools/umi_tools/umi-tools_dedup.xml
```diff
@@ -0,0 +1,292 @@
...
+    <description>Extract UMI from fastq files</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements">
+        <requirement type="package" version="1.5">samtools</requirement>
```
bgruening: 1.6 should be available already.

tools/umi_tools/umi-tools_dedup.xml
```diff
@@ -0,0 +1,292 @@
...
+            ln -sf '${input}' 'input.bam' &&
+            ln -sf '$input.metadata.bam_index' 'input.bam.bai' &&
+            #set $input_file = 'input.bam'
+        #end if
+
+        umi_tools dedup
+            --random-seed 0
+            --extract-umi-method $extract_umi_method
+            #if str($extract_umi_method) != 'read_id':
+                --umi-separator '$umi_separator' --umi-tag $umi_tag
```
bgruening: all text parameters and data parameters should be single-quoted

tools/umi_tools/umi-tools_dedup.xml
```diff
@@ -0,0 +1,292 @@
...
+        <param name="soft_clip_threshold" argument="--soft-clip-threshold" type="integer" value="4" label="Soft clip threshold" help="Mappers that soft clip, will sometimes do so rather than mapping a spliced read if there is only a small overhang over the exon junction. By setting this option, you can treat reads with at least this many bases soft-clipped at the 3' end as spliced." />
+        <param name="read_length" argument="--read-length" type="boolean" truevalue="--read-length" falsevalue="" label="Use the read length as as a criterion when deduping" />
+        <param name="whole_contig" argument="--whole-contig" type="boolean" truevalue="--whole-contig" falsevalue="" label="Consider all alignments to a single contig together" help="This is useful if you have aligned to a transcriptome multi-fasta" />
+        <param argument="--subset" type="float" min="0" max="1" value="1" label="Only consider a random selection of the reads" />
```
bgruening: value should be 1.0?

bgruening:
@dmaticzka 0.5.2 was also released a few minutes ago :)

tools/umi_tools/umi-tools_dedup.xml
dmaticzka: 4 is the official default

bgruening: 4? You have a min/max here from 0-1 and a float value :)

dmaticzka: ah, you were talking about that one :) didn't write this part, it's copied from umi-tools_group.xml. will change it in both files

dmaticzka:
okay, this should handle everything. further increased umi_tools to 0.5.3

tools/umi_tools/umi-tools_group.xml
```diff
@@ -72,12 +72,6 @@
         </data>
     </outputs>
     <tests>
-        <test>
-            <param name="input" value="group_in1.sam" ftype="sam" />
```
bgruening: @dmaticzka do you know what is going on here? Why is the BAM file different?

dmaticzka: yes, I spent some time investigating... the bam file is different, but the output stays the same.

reason is that entries are ordered differently on different machines, so multiple entries at the same coordinate will also have a different order.

bgruening: Can we compare the size (https://github.com/galaxyproject/tools-iuc/blob/88c494b035a34b097247a29bfd3515875d94ce2b/tools/qiime/qiime_core/summarize_taxa_through_plots.xml#L89) of the BAM instead and leave the test included? If not I suspect a few test files can be removed.

dmaticzka: I'll check, but I doubt it since there's compression involved. otherwise I'll delete the files related to this test...

dmaticzka:
@bgruening You were right, it uncompresses bam before comparison (but only if the target filetype is set).

Should be all done now!

bgruening:
Thanks @dmaticzka!

PR #1645, opened 2017-12-21
Featurecount: Add output option to keep header

scholtalbers:
For e.g. multiqc we need the column header in the output. However, in the current state this is always stripped. This adds another option to the output to keep it in.

### DIFF
--- a/tools/featurecounts/featurecounts.xml
+++ b/tools/featurecounts/featurecounts.xml
@@ -24,13 +24,13 @@
             -s  $extended_parameters.strand_specificity
                 $extended_parameters.multimapping_enabled.multimapping_counts
 
-                #if str($extended_parameters.multimapping_enabled.multimapping_counts) == " -M"
+                #if str($extended_parameters.multimapping_enabled.multimapping_counts) == " -M":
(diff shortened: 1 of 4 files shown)


bgruening:
@scholtalbers is the Galaxy MultiQC not working properly? If so can we fix the multiQC integration instead?

scholtalbers:
@bgruening generally the samples are spread across histories, making the use of multiqc through galaxy difficult. See also: https://gitter.im/galaxyproject/Lobby?at=5a3ba4a403838b2f2a1fa7ec

yhoogstrate:
>  could we please have a keep header option for featurecounts?

This would have my preference too. Any opinion maybe also from @mvdbeek @lparsons ?

lparsons:
If I understand correctly, an option to export the file with the header is really what is needed. While that would be ideal, I think it's probably fairly challenging to ensure that is made available with all the various ways people export data (via tools, history downloads, etc.), so an option to simply put the header into the output file would be a good way to handle things.

bgruening:
@scholtalbers this can be closed now isn't it?

scholtalbers:
yes! 👍

PR #1647, opened 2017-12-21
Update to Version 1.3, adding HiCExplorer support

joachimwolff:


joachimwolff:
To be honest I do not know what is going on with the failing tools. I added the HiCExplorer support and checked it locally with planemo and it was working.

bgruening:
@joachimwolff one error is that `tools/multiqc/test-data/hicexplorer1.log` is missing.

joachimwolff:
For which reasons the test-data is in .gitignore?

mvdbeek:
> For which reasons the test-data is in .gitignore?

https://github.com/galaxyproject/tools-iuc/blob/master/.gitignore#L48
it's the .log file that is in gitignore, not test-data.

joachimwolff:
@mvdbeek Thanks, this makes sense. Maybe I rename my files even if they are log files.

PR #1627, opened 2017-12-11
[WIP] MACS2 Callpeak should also work for ATAC-Seq data

zfrenchee:
This PR should address #1501 

@mvdbeek, cc. @bgruening 

@mvdbeek helped me figure out this issue locally, but for reproducibility, we now need to merge this. I'm not sure how to go about doing this exactly, but I'm happy to work through it with you. 

How do I add a conditional at each of these places?

Issue #1501, opened 2017-10-02
macs2 should be able to take many treatment files and a single control file. 


zfrenchee:
Hello, 
I've been working on this issue with @mvdbeek|:

The macs2 callpeak tool in this repo works in two scenarios: when you have 1 treatment file and 1 control file, or when you have a collection of treatment files and control files of the same length. Both these use cases make sense for ChIPseq data (which is what macs was originally developed for) but for ATACseq, which many use macs for as well, it is common to use a single control file for many treatment files. 

When I put this in my workflow editor: 

<img width="539" alt="screen shot 2017-10-02 at 1 14 32 pm" src="https://user-images.githubusercontent.com/2761597/31089887-b4efa7d2-a773-11e7-9d7a-55d5aaed175b.png">

Macs outputs a single file: peaks using the first treatment file in the collection and the control file -- none of the other treatment files is used, they're all ignored. 

Visually, if I have dataset collection [a, b, c, d, e, f, g] and control file [H], I get the results of callpeak(a, H) and nothing else. I'd like to be instead getting [callpeak(a, H), callpeak(b, H), callpeak(c, H), callpeak(d, H), ...] in a collection. 

Can we change the macs tool to operate this way?

jmchilton:
Is this a modified version of [this tool](https://github.com/galaxyproject/tools-iuc/blob/master/tools/macs2/macs2_callpeak.xml) - like did you make "treatment" non-multiple?

zfrenchee:
@jmchilton I'm using exactly that tool you linked. 

@mvdbeek says that by changing both the inputs to be non-multiple (and then changing the command slightly to account for that) he gets the right results. I'm trying that right now to see if I get the same. 

Assuming that fixes the issue, I'd be interested in seeing if we can add in all the possible behaviors for macs into one tool: 
1. single treatment/control files
2. equal length treatment/control collections
3. treatment collection/single control

zfrenchee:
**Update**: that did indeed solve my problem. I'd love to see this fixed in this repo as well.

mvdbeek:
> Visually, if I have dataset collection [a, b, c, d, e, f, g] and control file [H], I get the results of callpeak(a, H) and nothing else.

[...]

PR #1648, opened 2017-12-23
macs2: modify inputs, outputs, help and add r-base 3.4 as requirement

mblue9:
This PR has suggested modifications for MACS2, mainly because I ran into the same issue as @zfrenchee here: https://github.com/galaxyproject/tools-iuc/issues/1501

The changes here are to:

- add conditionals for more flexible handling of treatment and control inputs
- add r-base 3.4 as requirement to macros.xml (to remove of some errors due to defaulting to use R 3.2)
- change to only produce narrowPeaks file as default output
- add more info to Help section
- increase wrapper version for callpeak

@zfrenchee what do you think of the changes here for the conditionals you wanted? (here: https://github.com/galaxyproject/tools-iuc/pull/1627)
I think this should allow MACS2 to accept as input:

- single treatment/control files
- equal length treatment/control collections
- treatment collection/single control

but would be great if you could check

### DIFF
--- a/tools/macs2/macs2_callpeak.xml
+++ b/tools/macs2/macs2_callpeak.xml
@@ -1,4 +1,4 @@
-<tool id="macs2_callpeak" name="MACS2 callpeak" version="@VERSION_STRING@.0">
+<tool id="macs2_callpeak" name="MACS2 callpeak" version="@VERSION_STRING@.1">
     <description>Call peaks from alignment results</description>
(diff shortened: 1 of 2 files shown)


zfrenchee:
@mblue9 This looks terrific!

bgruening:
Thanks a lot @mblue9 and merry Christmas!

PR #1649, opened 2017-12-23
bump fastqc version number

bgruening:
https://github.com/bioconda/bioconda-recipes/pull/7151


tools/fastqc/rgFastQC.xml
```diff
@@ -63,30 +62,30 @@
...
             <output name="html_file" file="fastqc_report.html" ftype="html" lines_diff="100"/>
-            <output name="text_file" file="fastqc_data.txt" ftype="txt" lines_diff="2"/>
+            <output name="text_file" file="fastqc_data.txt" ftype="txt" lines_diff="4"/>
         </test>
         <test>
             <param name="input_file" value="1000gsample.fastq.bz2" ftype="fastq.bz2" />
             <param name="contaminants" value="fastqc_contaminants.txt" ftype="tabular" />
             <output name="html_file" file="fastqc_report.html" ftype="html" lines_diff="100"/>
-            <output name="text_file" file="fastqc_data.txt" ftype="txt" lines_diff="2"/>
+            <output name="text_file" file="fastqc_data.txt" ftype="txt" lines_diff="4"/>
```
yhoogstrate: Apart from the Filename entries, I don't see anything changed at all. Is lines_diff=4 really needed if the appropriate test data is updated?

bgruening: Imho we should design our test in a way that we do not need to update test-data anytime. Ideally we bump just the version number and test should pass if there are not any big changes to the outputs. This would enable automatic updates in a long run - the tests still needs to be strict enough to catch errors, but not to strict so that we need to update those every time.

yhoogstrate: Maybe I'm a bit too conservative or paranoid. Please merge :+1:

bgruening: Thanks @yhoogstrate!

PR #1659, opened 2018-01-07
Delete Star readme

bgruening:
Can anyone confirm that this Readme is outdated and can be removed?


yhoogstrate:
It can be removed. It's annotated in comment sections here: https://github.com/galaxyproject/tools-iuc/blob/master/data_managers/data_manager_star_index_builder/data_manager/rna_star_index_builder.xml#L103

Thanks !

PR #1660, opened 2018-01-08
fix for 'Read distribution from RSeQC' in MultiQC 

ashvark:
Hi Team, 
I am trying to use mulitiqc to summarise the outputs of 'Read Distribution from RSeQC'. I realised that this module is not working properly. The condition for 'read_distribution' is missing in the 'command' tag of multiqc.xml. I believe this pull request will fix this issue.


bgruening:
@ashvark thanks a lot. This looks good to me!

PR #1661, opened 2018-01-08
added missing echo to version command

bernt-matthias:
the missing echo gives a `command not found` error on the info page of data sets created with enasearch.


bernt-matthias:
Maybe one could add the links to the documentation of the command line utility. I started for enasearch_retrieve_taxons.xml

bgruening:
Thanks!

PR #1646, opened 2017-12-21
Featurecount: add header and remove hidden attribute

scholtalbers:
Links to #1645 but this also removes the hidden attribute on the summary output data.

PR #1664, opened 2018-01-10
Add options to featureCounts for built-in annotation and header output

mblue9:
This PR contains the changes from @scholtalbers PR https://github.com/galaxyproject/tools-iuc/pull/1646 to have an option to output the header (can we please have this option??? I really need this!!) and also makes featureCounts built-in annotation available as an option.

 - add @scholtalbers changes from PR #1646 to have option to output header
 - add option to use featureCounts built-in RefSeq annotation for hg38, hg19, mm10 and mm9
 - add test for built-in annotation


tools/featurecounts/featurecounts.xml
```diff
@@ -1,17 +1,27 @@
...
     <description>Measure gene expression in RNA-Seq experiments from SAM or BAM files.</description>
     <requirements>
         <requirement type="package" version="1.6.0">subread</requirement>
     </requirements>
 
     <version_command>featureCounts -v 2&gt;&amp;1 | grep .</version_command>
-    <command><![CDATA[
+    <command detect_errors="exit_code"><![CDATA[
+        ## Export fc path for its built-in annotation
+        export FC_PATH=\$(dirname \$(which featureCounts | sed 's/bin//')) &&
```
yhoogstrate: if featureCounts is at: `/home/youri/src/binaries/subread-1.6.0-source/bin/featureCounts`

You will get: `/home/youri/src/aries/subread-1.6.0-source/bin`


You need to make sure bin/featureCounts is at the end using the dollar sign:

`command -v featureCounts  | sed 's@bin/featureCounts$@@'`

mblue9: Thanks for reviewing this @yhoogstrate!

But I think this is exactly what I want `/home/youri/src/aries/subread-1.6.0-source/bin` as I then strip off the bin directory and add on the annotation directory later here:
https://github.com/mblue9/tools-iuc/blob/6a24a4abca251a86761c2281d89c6c2f264c9730/tools/featurecounts/featurecounts.xml#L23

Should I not do it like this?

yhoogstrate: It does not strip off the bin directory, it replaces the first occurance of bin, regardless of whether this is at the end or not. If the first occurance of bin gets substituted instead, then if the path is:

/home/youri/src/**binaries**/subread-1.6.0-source/bin/featureCounts

it becomes after regex:

/home/youri/src/**aries**/subread-1.6.0-source/bin

instead of:

/home/youri/src/**binaries**/subread-1.6.0-source/ `[+annotation/]`

mblue9:
Sorry @yhoogstrate! I had completely missed your point about bin _within_ the path! I'll blame those dodgy eyes again (it's great to have backups here 😄 ) Thanks!! I've made that change now.

PR #1665, opened 2018-01-11
HiCExplorer 2.0: Setting system environment variables for mkl, numexpr and open mp to …

joachimwolff:
…control multicore usage

@bgruening I think this fixing the multicore numpy issue as discussed in https://github.com/deeptools/HiCExplorer/issues/167


bgruening:
@natefoo @jmchilton I'm fine with merging this for the time being, but should we also set/control this on the framework level?

joachimwolff:
Obviously no interest. Closing this issue.

mvdbeek:
Sorry @joachimwolff, I guess this flew under the radar. If you export an environment variable it is scoped to the current process and its subprocess, so no need to unset them (try exporting a var in one shell, and then echoing in another shell). The rest seems fine to me, I don't think setting those variables on a framework level is necessary for merging this change.

tools/hicexplorer/hicBuildMatrix.xml
```diff
@@ -61,6 +62,7 @@
         && mv $qc.files_path/*.log raw_qc
         && mv matrix.$outputFormat matrix
         && samtools sort ./unsorted.bam -o sorted.bam
+        && @unset_multi_thread_variables@
```
bgruening: these once can be removed, we do not need to unset them

tools/hicexplorer/macros.xml
```diff
@@ -192,7 +192,22 @@
...
 
+    <token name="@set_multi_thread_variables@">
+    <![CDATA[    
+        export MKL_NUM_THREADS=@THREADS@ &&
+        export NUMEXPR_NUM_THREADS=@THREADS@ &&
+        export OMP_NUM_THREADS=@THREADS@
+    ]]>
+    </token>
 
+    <token name="@unset_multi_thread_variables@">
```
bgruening: can be completely removed

tools/multiqc/multiqc.xml
```diff
@@ -136,7 +136,10 @@ mkdir multiqc_WDir &&
         @LN_FILES@
     #elif str($repeat.software_cond.software) == "hicexplorer"
         #set $pattern = "Min rest. site distance"
-        @LN_FILES@
```
bgruening: this does not belong into this PR

PR #1668, opened 2018-01-16
MACS2 - fix for BAMPE and some changes to defaults

mblue9:
- --format BAMPE is not getting output when selected since my PR of 3 weeks ago https://github.com/galaxyproject/tools-iuc/pull/1648, fixed here
- added a BAMPE test
- made some parameters optional as there were lots of parameters being output and some combinations don't make sense afaik (--nomodel with --mfold and --bw), only the treatment parameter is required
- changed the effective genome sizes (for hs, mm, dm, ce) to the numbers given on the MACS website here: https://github.com/taoliu/MACS. Or should I keep the other numbers? Where do they come from?


mblue9:
Hi, if anyone has time to take a look at these would love to know what people think? are they ok?

tools/macs2/macs2_callpeak.xml
```diff
@@ -205,17 +220,17 @@
...
             <when value="on">
-                <param name="to_large" type="boolean" truevalue="--to-large" falsevalue="" checked="False"
+                <param name="to_large" type="boolean" truevalue="--to-large" falsevalue="" checked="False" optional="True"
                     label="When set, scale the small sample up to the bigger sample"
                     help="By default, the bigger dataset will be scaled down towards the smaller dataset, which will lead to smaller p/qvalues and more specific results. Keep in mind that scaling down will bring down background noise more. (--to-large)"/>
-                <param name="nolambda" type="boolean" truevalue="--nolambda" falsevalue="" checked="False"
+                <param name="nolambda" type="boolean" truevalue="--nolambda" falsevalue="" checked="False" optional="True"
                     label="Use fixed background lambda as local lambda for every peak region" help="up to 9X more time consuming (--nolambda)"/>
-                <param name="ratio" type="float" value="1.0"
+                <param name="ratio" type="float" value="1.0" optional="True"
```
bgruening: if we set it to 1.0 by default, it is set and the `$advanced_options.ratio` will always be true (the only situation where this false is when the value is removed to empty, isn't it). Is this intended ...

tools/macs2/macs2_macros.xml
```diff
@@ -13,19 +13,17 @@
...
             <param name="effective_genome_size_options_selector" type="select" label="Effective genome size"
                 help="The effective genome size is the portion of the genome that is mappable. Large fractions of the genome are stretches of Ns that should be discarded.
-                Also, if repetitive regions were not included in the mapping of reads, the effective genome size needs to be adjusted accordingly. (--gsize)">
-                <option value="2451960000">H. sapiens (2,451,960,000)</option>
-                <option value="2150570000">M. musculus (2,150,570,000)</option>
-                <option value="121400000">D. melanogaster (121,400,000)</option>
-                <option value="93260000">C. elegans (93,260,000)</option>
-                <option value="12400000">S. cerevisiae (12,400,000)</option>
+                Also, if repetitive regions were not included in the mapping of reads, the effective genome size needs to be adjusted accordingly. Sizes are from https://github.com/taoliu/MACS(--gsize)">
+                <option value="2700000000">H. sapiens (2.7e9)</option>
```
bgruening: From where do you get these numbers, can we specify this in a the help section so people can reference this or update there numbers.

tools/macs2/macs2_callpeak.xml
mblue9: It sure isn't intended, I'm trying to remove params! Thanks! I've changed that.

tools/macs2/macs2_macros.xml
mblue9: Numbers are from the MACS2 website and I've added that to the help now. Although the numbers are old and refer to hg18(!) but at least there's a ref for them I guess.

zfrenchee:
@mblue9 @bgruening what are the chances it would be possible to remove the dependency on $HOME?

tools/macs2/macs2_macros.xml
```diff
@@ -13,19 +13,17 @@
         <conditional name="effective_genome_size_options">
             <param name="effective_genome_size_options_selector" type="select" label="Effective genome size"
                 help="The effective genome size is the portion of the genome that is mappable. Large fractions of the genome are stretches of Ns that should be discarded.
-                Also, if repetitive regions were not included in the mapping of reads, the effective genome size needs to be adjusted accordingly. (--gsize)">
-                <option value="2451960000">H. sapiens (2,451,960,000)</option>
-                <option value="2150570000">M. musculus (2,150,570,000)</option>
-                <option value="121400000">D. melanogaster (121,400,000)</option>
-                <option value="93260000">C. elegans (93,260,000)</option>
-                <option value="12400000">S. cerevisiae (12,400,000)</option>
+                Also, if repetitive regions were not included in the mapping of reads, the effective genome size needs to be adjusted accordingly. Sizes are from the MACS2 website (--gsize)">
```
bgruening: We are using in deeptools the old numbers as well. Are you sure the old ones are not appropriate? 
https://github.com/deeptools/deepTools/blob/eca29f60663d169a80494358732ea9ce8929e4cd/galaxy/wrapper/deepTools_macros.xml

There is also a link to a paper included.

bgruening:
Overall this looks good to me. Anyone else as these are quite substantial changes? @nekrut?

tools/macs2/macs2_macros.xml
mblue9: I am not sure the old numbers are not appropriate but I think they are not accurate. As they were calculated using old assemblies e.g. hg18/hg19 and effective genome size also varies with read length which those numbers don't take into account. 

I have not had time yet to look at how effective genome size may vary results but I did come across one user reporting a difference (albeit with old macs1.4) here https://groups.google.com/forum/#!topic/macs-announcement/-iIDkVwenn8

I've had a look at those 2 papers referenced in the deeptools wrapper and they're old info too. But the deeptools website does have some info on how to calculate effective genome size here https://github.com/deeptools/deepTools/wiki/General-deepTools-FAQs#effGenomeSize

Maybe we could add in one of those methods instead? Or do you know about these deeptools plans from that webpage

>How do I calculate the effective genome size for an organism that's not in your list?
We plan on including a module that will calculate the effective genome size for you, but it's not ready yet, so you will have to find a solution outside of deepTools for the time being.

mblue9:
@zfrenchee sorry I don't get your question what do you mean by the dependency on $HOME?

I have made some big changes here and I know macs is widely used so if @nekrut or others could take a look that would be great!

zfrenchee:
@mblue9 I believe this MACS2 tool uses $HOME as a working directory, which isn't good form (see [number 3 in these guidelines](https://github.com/bgruening/coding_for_big_data)). This is what caused the error I messaged you about, but @bgruening and I were able to resolve it. Do you have anything to add, @bgruening ?

bgruening:
```
export PYTHON_EGG_CACHE=`pwd`
```
or `HOME` should help here. We can add this to the wrapper.

Just as a note 18.01 will have a setting for this and will set HOME appropriate. `profile=18.01` tools will get this also automatically.

mblue9:
Ah I see, @zfrenchee thanks for the info and the link to those guidelines!

@bgruening thanks for the command! I would not have known that! Just realised I only added that export to callpeak, should I instead add the export to the macros.xml and then call it using a token in the 9 other xmls?

bgruening:
> should I instead add the export to the macros.xml 

:+1:

mblue9:
I could really use this updated macs asap. If I changed the effective genome size numbers back to the way they were (if that's the main holdup) would that help get this PR merged?

bgruening:
Thanks @mblue9!

PR #1670, opened 2018-01-17
add jaspscan tool for EMBOSS suite.

FredericBGA:
I've add one xml for jaspscan and one related test file.

Somebody will have to create a dedicated page here: [https://galaxy-iuc.github.io/emboss-5.0-docs/jaspscan.html](url)

I can provide also a perl script that uses the jaspar API: [http://jaspar.genereg.net/api/v1/live-api/](url) in order to create the files needed by EMBOSS (in INSTALL_DIR/share/EMBOSS/data/JASPAR_*)

### DIFF
--- a/tools/emboss_5/macros.xml
+++ b/tools/emboss_5/macros.xml
@@ -4,6 +4,11 @@
             <requirement type="package" version="5.0.0">emboss</requirement>
         </requirements>
     </xml>
+    <xml name="requirements_emboss6">
+        <requirements>
+            <requirement type="package" version="6.0.0">emboss</requirement>
+        </requirements>
+    </xml>
     <xml name="citations">
         <citations>
             <citation type="doi">10.1016/S0168-9525(00)02024-2</citation>
--- a/tools/emboss_5/macros.xml
+++ b/tools/emboss_5/macros.xml
@@ -6,7 +6,7 @@
     </xml>
     <xml name="requirements_emboss6">
         <requirements>
-            <requirement type="package" version="6.0.0">emboss</requirement>
+            <requirement type="package" version="6.5.7">emboss</requirement>
         </requirements>
     </xml>
     <xml name="citations">
--- a/tools/emboss_5/emboss_jaspscan.xml
+++ b/tools/emboss_5/emboss_jaspscan.xml
@@ -1,0 +1,62 @@
+<tool id="EMBOSS: jaspscan" name="jaspscan" version="5.0.0.1">
+  <description>Scans DNA sequences for transcription factors</description>
+  <macros>
+    <import>macros.xml</import>
+  </macros>
+  <expand macro="requirements" />
+  <code file="emboss_format_corrector.py" />
+  <command>jaspscan -sequence '$input1' -menu '$menu' -threshold '$threshold' $both -outfile '$out_file1' -rformat2 '$out_format1' -auto</command>
+  <inputs>
+    <param name="input1" type="data" format="data" label="Sequence" /> 
+    
+    <param name="threshold" type="float" value="80" label="If the matrix score is greater than or equal to this percentage then a hit will be reported (Any numeric value)" />
+    <param name="both" type="boolean" truevalue="-both" falsevalue="" checked="false" label="If set then both the forward and reverse strands are searched" />
+    <param name="menu" type="select" label="Jaspar matrix set">     
+      <option value="C">Core</option>
+      <option value="F">Fam</option>
+      <option value="P">Phylofacts</option>
+      <option value="N">CNE</option>
+      <option value="O">POLII</option>
+      <option value="S">SPLICE</option>
+    </param>   
+    <param name="out_format1" type="select" label="Output sequence file format">
+      <option value="seqtable">seqtable</option>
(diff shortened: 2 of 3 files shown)


bgruening:
@FredericBGA I think the jaspscan tool is not part of EMBOSS 5 isn't it? 
Can we do the following: Lets bump the conda version of emboss to be the lastest version 6.6 and use this requirement for this single tool.

FredericBGA:
@bgruening I missed that point indeed! Sorry. EMBOSS 5 is now ten years old. I'm working with 6.5.7 on the server, it works well with these EMBOSS 5 XML files (I have not tested all tools of course).
Do I need to add then a new requirements in the macro.xml?
I have no idea how to deal with conda.
It can wait, the Japscan tool works for me, I wanted only to share for others users.

bgruening:
If you are updating the requirements in the macro.xml file, than we would need to update all other 100 tools as well I guess ... this will be a lot of work - not sure if you have the time - it would be highly appreciated :)

We had several attempts to generate these wrappers automatically out of the ACD files - but never finished it.

As a quick solution I would add the new requirement only to this tool. We could also put this into its on repository so outside of this directory. The situation is not ideal for EMBOSS - sorry.

FredericBGA:
I was meaning: add a new requirement for jaspscan, I'm sure that I won't have time for all others tools.

```
<xml name="jaspscan_requirements">
        <requirements>
            <requirement type="package" version="6.0.0">emboss</requirement>
        </requirements>
</xml>
```
I have never heared about jaspscan until yesterday when a coleague needs it in Galaxy. All the jaspscan data have to be integrated using jaspextract. But the procedure is deprecated (see: [http://emboss.sourceforge.net/apps/release/6.6/emboss/apps/jaspextract.html](url). It's why I have also a perl script in order to grab all the data needed by japscan. It's not perfect, it difficult to get all the previous fields. But it works at last.

It's fine for me. I will add the new requirement for jaspscan.

EMBOSS under Galaxy is great, even if the state and version of tools are not perfect.

FredericBGA:
Here is the perl script.
[getJASPAR.pl.txt](https://github.com/galaxyproject/tools-iuc/files/1642184/getJASPAR.pl.txt)

tools/emboss_5/macros.xml
```diff
@@ -4,6 +4,11 @@
             <requirement type="package" version="5.0.0">emboss</requirement>
         </requirements>
     </xml>
+    <xml name="requirements_emboss6">
+        <requirements>
+            <requirement type="package" version="6.0.0">emboss</requirement>
```
bgruening: I think we need to use this version: 6.5.7 as this is the only one available one bioconda: https://anaconda.org/bioconda/emboss/files

bgruening:
@FredericBGA do you know how big the downloaded data is?

FredericBGA:
It's not big at all.

The script takes less than 10 minutes.
```
du -kh
416K    ./JASPAR_CORE
0       ./JASPAR_POLII
64K     ./JASPAR_PHYLOFACTS
0       ./JASPAR_FAM
96K     ./JASPAR_PBM
96K     ./JASPAR_PBM_HOMEO
0       ./JASPAR_SPLICE
0       ./JASPAR_PBM_HLH
96K     ./JASPAR_CNE
768K    .
```
I've choosen to retrieve JASPAR_PBM* as well, even if the jaspscan tool seems to be not ready to deal with theses files (, at least in 6.5.7 version of EMBOSS, see the comments in the perl script)

bgruening:
@FredericBGA can you point me to the deprecation of the jaspextract tool? I only could find this one here: http://emboss.sourceforge.net/apps/release/6.6/emboss/apps/jaspextract.html

Do you think you have time to update the bioconda package to the latest EMBOSS version and if needed integrate your perl script into the build process, so that we ship the small amount of files with the package?

Thanks!

FredericBGA:
@bgruening  The tool is not really deprecated, this is the Jaspar site which does not provided the file(s?) that jaspextract needs. The tool works which the data I've been able to built using the API of Jaspar and following the example here [EMBOSS GIT test data](https://github.com/kimrutherford/EMBOSS/tree/master/test/data/jaspar)

I have understood that the files (Archive.zip or JASPAR_CORE.tgz,
JASPAR_FAM.tgz and JASPAR_PHYLOFACTS.tgz) were no longer available on Jaspar site when visiting theses URLs:
[old mail](http://mailing.unix.emboss-users.narkive.com/tCd5rOUI/emboss-data-for-jasextract)
[old book](https://books.google.fr/books?id=9bXC7ueiGVQC&pg=PA23&lpg=PA23&dq=jaspextract&source=bl&ots=dqeLppezkB&sig=P7iUwXLVc5ZlQBE-Z1EL3f9i3q0&hl=fr&sa=X&ved=0ahUKEwjH9qS-6P_YAhVIhqQKHTHjBnw4ChDoAQgrMAE#v=onepage&q=jaspextract&f=false)
[comment in jaspextract source code](https://github.com/pjotrp/EMBOSS/blob/master/doc/programs/master/emboss/apps/jaspextract.html)

With help, I will try to update the bioconda package to the latest EMBOSS version and try to embed the perl script for Jaspscan data.

PR #1672, opened 2018-01-19
adds scale_metagenome wrapper for picrust

skanwal:


tools/picrust/scale_metagenome.xml
```diff
@@ -0,0 +1,42 @@
+<tool id="scale_metagenome" name="Scale_Metagenome" version="1.1.3">
```
bgruening: Can we remove the `_` from the name?

tools/picrust/scale_metagenome.xml
```diff
@@ -0,0 +1,42 @@
...
+    </macros>
+    <requirements>
+        <requirement type="package" version="1.1.3">picrust</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        scale_metagenome.py -i '$input1' -s '$input2' -o tempbiom
+        @OUTPUT_CONVERSION_COMMANDS@
+    ]]></command>
+    <inputs>
+        <param type="data" name="input1" format="biom1" />
```
bgruening: indentation is wrong here

tools/picrust/scale_metagenome.xml
```diff
@@ -0,0 +1,42 @@
...
+	<param type="data" name="input2" format="tabular" />
+	<expand macro="biom_format_select"/>
+    </inputs>
+    <outputs>
+        <expand macro="biom_output"/>
+    </outputs>
+    <tests>
+        <test>
+            <param name="input1" value="otus.biom"/>
```
bgruening: indentation

bgruening:
Thanks a lot @skanwal! This looks wonderful. Just a few tiny comments.

tools/picrust/scale_metagenome.xml
bgruening: Should the version be `version="@WRAPPER_VERSION@.0"`, reusing the macro? @Slugger70?

skanwal: Sure, we can change the name to "Scale Metagenome"

tools/picrust/scale_metagenome.xml
skanwal: Tried fixing it in the new commit

tools/picrust/scale_metagenome.xml
skanwal: Sorry, couldn't identify where it was incorrect?

tools/picrust/scale_metagenome.xml
```diff
@@ -0,0 +1,46 @@
+<tool id="scale_metagenome" name="Scale Metagenome" version="1.1.3">
+    <description>Converts metagenomic relative abundance back to sequence counts</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <requirements>
+        <requirement type="package" version="1.1.3">picrust</requirement>
```
bgruening: @skanwal I see that this one tool depends on 1.1.3, all other tools in the suite belongs to 1.1.1 https://github.com/skanwal/tools-iuc/blob/65f6bb5e5df0bdc322436ce3714d635ff5397198/tools/picrust/macros.xml#L5 is there a special reason?

What do you think of upgrading all other tools as well, I can help with this of course. 
Its more or less increasing the version in the macro and using the macro also here in your tool.

tools/picrust/scale_metagenome.xml
bgruening: I fixed this for you. You can do now `git pull origin scale_metagenome` to get my changes locally.
Thanks.

tools/picrust/scale_metagenome.xml
skanwal: I just rechecked the link we used for picrust documentation in the workshop today. 
http://picrust.github.io/picrust/scripts/

This is using 1.1.3 - for all tools infact. Not sure why is macro using 1.1.1, probably it was written a while ago?

tools/picrust/scale_metagenome.xml
skanwal: Thank you :)

tools/picrust/scale_metagenome.xml
bgruening: Yes most likely. Galaxy supports different versions of the same tools - for reproducibilty and the older one used an older version.

skanwal:
Following-up, is this pull request waiting for any further changes?

bgruening:
I think we should update the other wrappers to the newer version as well. Do you want to do this?

skanwal:
Definitely - when I find sometime, hopefully soon.
Can you please point out other wrappers that need to be updated?

PR #1674, opened 2018-01-19
Markduplicate regex fix

bwlang:
the regex is no longer needed by modern versions of picard...


bgruening:
@bwlang this one might be interesting for you as well. https://github.com/galaxyproject/tools-iuc/pull/1425
I think only the tests needs to be updated or relaxed.

bgruening:
Thanks Brad!

Issue #508, opened 2015-12-31
Roadmap: JBrowse


erasche:
# Changelog / Upcoming Changes

## v0.5
- [x] update existing jbrowse instances on disk (adding new tracks, also support pulling info from apollo) f4206bc58448bb39e5fae31942477c45642a9fca
- [x] generate-names.pl http://gmod.org/wiki/JBrowse_Configuration_Guide#Name_Searching_and_Autocompletion
- [x] Allow forcing HTML rather than canvas ef1646fc375513143fd743841ac618720267dbc9
- [x] defaultTracks e6ef382e885381ab1606e44f397477014434aa82
- [x] http://gmod.org/wiki/JBrowse_Configuration_Guide#General_configuration_options acb5b6696880d517bf76b790e70496b714dd46c3, f93e44ee9cc01d2f83ccbfe7151b408c12d4b333
- [x] Bugfix: >12 sequences would exceed brewer colour limit and crash. Now cycles back to start. https://github.com/galaxyproject/tools-iuc/commit/3dadcef329a8881837293e354b21880a6ecd3ca9
- [x] Bugfix: Incorrect XML could be generated, causing a crash for automatically coloured, automatically scored bigwig tracks 2fd44398452799138fa5c95fa78996814c72d386
- [x] Bugfix: BlastXML tracks without a gff3 feature parent would crash. 4be3385

## v0.6

- [x] log support for bigwig tracks
- [x] REST endpoints
- [x] Accept data from chrom/len files, and from data tables
- [x] templated values in name / text field. E.g. ability to use "#date#" to get a dated category. Useful when updating pre-existing JBrowses with re-run analyses.
- [x] track metadata
- [x] Conda packages

## v0.7

- [x] Editing menus
- [x] Plugins
  - [x] https://github.com/cmdcolin/gccontent
  - [x] https://github.com/Arabidopsis-Information-Portal/ComboTrackSelector
- [x] Themes

## v0.8+

- [x] SPARQL endpoints
- [ ] more JBrowse plugins! (Specifically the one that links db_xrefs to upstream databases, such as IPR:000000 -> `http://....`)
  - [x] https://github.com/cmdcolin/multibigwig
  - [ ] https://github.com/cmdcolin/sashimiplot
- [ ] disable left-click action
- [ ] auto create blat?
- [ ] ... ? Ideas?
- [ ] enhance compat with blast XML data
- [ ] full support for protein files
  - [ ] fix chrom selector upstream
[...]

PR #1683, opened 2018-01-23
jbrowse-next (0.8.0)

erasche:
- [x] remove old emails that bounce
- [x] Fix metadata issues
    yeah... this is how it was supposed to work from the beginning. Oops. 
    auswahl_251
- [x] SPARQL
- [ ] support blast tabular data in 12 and 24 column variants
- [x] https://github.com/elsiklab/multibigwig
- [x] configurable chunk size

xref https://github.com/galaxyproject/tools-iuc/issues/508


erasche:
Never mind, can't include multibigwig due to lack of license. (@cmdcolin no chance this could be resolved? I'd love to ship this plugin for use in the galaxy tool)

cmdcolin:
It is mit license https://github.com/elsiklab/multibigwig/blob/master/bower.json :)

erasche:
Oh, I missed that. Awesome, thanks so much!

erasche:
TODO: "if you can see this text you need to whitelist this tool"

abretaud:
Just pushed some changes:
- updated to latest 1.15.4 version (not yet 1.16 as it's not yet supported by apollo)
- reworked/fixed the sparql and rest track, it should work now

I worked with @xgaia to test sparql, it almost works, but the last error is probably not in this jbrowse galaxy tool (maybe jbrowse, maybe the sparql query, this kind of stuff is... complicated).

I'll need to work on including the jbrowse plugins inside the conda package since with newest jbrowse releases they need to be built with webpack...
I'd like to test indexing gff with tabix too...

erasche:
> updated to latest 1.15.4 version (not yet 1.16 as it's not yet supported by apollo)

Would that be an issue for us? I'm assuming the data directory format doesn't change between 1.15 and 1.16, since only the data dir is sent we could probably upgrade 

> I'll need to work on including the jbrowse plugins inside the conda package since with newest jbrowse releases they need to be built with webpack...

urgh, that's... gross, ok. I really loved that we could load plugins from github/rawgit, that was super convenient.

abretaud:
> > updated to latest 1.15.4 version (not yet 1.16 as it's not yet supported by apollo)
> 
> Would that be an issue for us? I'm assuming the data directory format doesn't change between 1.15 and 1.16, since only the data dir is sent we could probably upgrade
Yep, maybe it would work great, we can try. I don't know why needs to be done to integrate 1.16 in apollo.

> 
> > I'll need to work on including the jbrowse plugins inside the conda package since with newest jbrowse releases they need to be built with webpack...
> 
> urgh, that's... gross, ok. I really loved that we could load plugins from github/rawgit, that was super convenient.

I agree! I haven't found an issue for this on https://github.com/GMOD/jbrowse, we could open one maybe

abretaud:
Ok, some plugins seem to work with the plugins built in the conda package
Looks like ComboTrackSelector is broken, I don't know why yet

abretaud:
@erasche when you have time, I think it's ready for a review
I added the multibigwig plugin, it's a bit hacky, but it seems to work
Apart from the themes, all plugins should work

tools/jbrowse/readme.rst
```diff
@@ -25,18 +25,31 @@ likely to change at any time! Beware. ;)
...
--  0.5.1 Support for contextual menus. Conda tests.
--  0.5 Update existing instances on disk. Index names. Support HTML tracks
-   instead of Canvas. Support default tracks. General JBrowse optinos
--  0.4 Support for dataset collections and customisation of tracks including
-   labelling, colours, styling. Added support for genetic code selection.
-   Fixed package installation recipe issues.
--  0.3 Added support for BigWig, etc.
--  0.2 Added support for BAM, Blast, VCF.
--  0.1 Initial public release.
+- 0.8 aka 1.16.1+galaxy0
```
erasche:
```suggestion
- 1.16.1+galaxy0
```

firefox made crazy version number changes, we can too :)

bgruening:
Merci @erasche and @abretaud!

erasche:
@abretaud could we maybe use `exit_code`? I worry that someday we have some python warning that causes the tool to fail unexpectedly.

bgruening:
Super cool. Thanks all!

PR #1693, opened 2018-01-30
Bump edgeR to v3.20.7

mblue9:
As I was updating the edgeR requirement to v3.20.7 here https://github.com/galaxyproject/tools-iuc/pull/1692 
I thought I should update it in the edgeR tool too, so this is a PR to do that.


bgruening:
Thanks @mblue9!

PR #1695, opened 2018-01-30
Add R-based charts tool

guerler:
+ [x] condafication needed
+ [x] fixing the last failing test


tools/charts/tool_dependencies.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0"?>
```
nsoranzo: We don't accept new `tool_dependencies.xml` files in this repo, you need to specify the corresponding conda packages in the tool `<requirements>` section.

bgruening:
@nsoranzo Aysam pinged me and ask for help to migrate this important tool to use conda and I told him to push what he has here as a PR. I will work on this and get it updated.

nsoranzo:
Cool, thanks @bgruening !

bgruening:
The linting will be fixed with https://github.com/galaxyproject/galaxy/pull/5421
But the last test is failing for me because R wants to allocate 6**TB** of memory ;)

guerler:
@bgruening Thanks for working on this. I suggest to remove the test case which requests too much memory. Are you sure its 6TB, I wonder how this test worked for me locally.

bgruening:
This is what R told me, I would rather try to fix it. I'm not sure what happens here at the moment, but it can also be that the newer R version 3.4 is causing this - unlikely but possible.

guerler:
I agree that it has to be fixed. I assume its the matrix clustering which causes this problem.

mblue9:
Hi, I had a look at this and I think that failing test could be due to an old R or r-matrix version issue. As if I run the test script that's failing e.g.

`Rscript '/Users/maria/galaxy/tools-iuc/tools/charts/charts.r' -w '/Users/maria/galaxy/tools-iuc/tools/charts/' -m heatmap -i '/private/var/folders/zn/m_qvr9zd7tq0wdtsbq255f8xypj_zg/T/tmp4621gs/files/000/dataset_1.dat' -c 'key1: 1, key2: 2, key3: 3' -s '' -o '/private/var/folders/zn/m_qvr9zd7tq0wdtsbq255f8xypj_zg/T/tmp4621gs/files/000/dataset_2.dat'`

It fails with a "Killed: 9" error and I can see it's using these versions:
R version 3.3.2
Matrix_1.2-7.1
getopt_1.20.0

But it passed for me when I ran the above command on the command line where I had these versions installed:
R version 3.4.1
Matrix_1.2-10
getopt_1.20.0

bgruening:
@mblue9 new version, new error :(

mblue9:
Looks like that error can be fixed by creating a Bioconda R 3.4 version of r-getopt 
(Bioconda currently only has r-getopt available for R 3.3 but conda-forge has an r-getopt for R 3.4 and if I use that one the test passes)

bgruening:
@mblue9 thanks a lot. I was able to fix it now.
@guerler one Q, the old tools were pushed to your personal TS account, correct? Now it will be pushed to an `iuc` account? Does this break anything in Galaxy?

guerler:
Thanks a lot. Pushing it to an IUC account is fine.

bgruening:
Ok, great. Thanks @guerler!

martenson:
Is the old charts repository installation in Galaxy newer than 18.01 useless now? Can it be uninstalled?

The fact that admins have to install a new repository for charts to continue working also need to be part of the release notes. xref https://github.com/galaxyproject/galaxy/pull/6006

bgruening:
@guerler ping.
We discussed this with Aysam and he said that people should install it manually or update it.
I guess this should be on the general Admin checklist.

guerler:
Are there other alternatives?

bgruening:
We could push into your TS account, but even then the Admin need to update the tool manually. Update instead of Install is the alternative.

martenson:
Let's use this new repo then, I installed it on https://test.galaxyproject.org/ so feel free to give it a shot.

PR #1697, opened 2018-01-31
adds -sorted support for bedtool coverage when using bams

bwlang:
also fixes gff and gff3 output sorting

### DIFF
--- a/tools/bedtools/macros.xml
+++ b/tools/bedtools/macros.xml
@@ -5,7 +5,7 @@
             <yield/>
         </requirements>
     </xml>
-    <token name="@WRAPPER_VERSION@">2.27.1</token>
+    <token name="@WRAPPER_VERSION@">2.27.0</token>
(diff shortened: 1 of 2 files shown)


tools/bedtools/macros.xml
```diff
@@ -5,7 +5,7 @@
             <yield/>
         </requirements>
     </xml>
-    <token name="@WRAPPER_VERSION@">2.27.0</token>
+    <token name="@WRAPPER_VERSION@">2.27.1</token>
```
bgruening: This should be untouched I think, instead please bump the version of coverageBed
Thanks! :)

bwlang: one of these days i'll figure out the versioning ;)

bgruening:
@bwlang do you think we can add a test for both changes? I would feel more comfortable with it, but I'm also ok with how it is now.

PR #1699, opened 2018-01-31
Add antismash 4.0.2

cmonjeau:
Based on the antismash tool made by @bgruening -> https://github.com/bgruening/galaxytools/tree/master/tools/antismash

Based on @abretaud 's advice.


bgruening:
@cmonjeau its working for me locally. I think the travis image is just too small for this installation :(
Maybe we leave it in galaxytools for now if tests are failing and I make sure tests passing locally.

abretaud:
Ok, I'm now sure it's a problem of disk space: I've made some ugly hacks to .travis.yml to get [some logs](https://travis-ci.org/abretaud/tools-iuc/jobs/338001604#L2401)  => "No space left on device"
I guess the antismash data requires just a little more than the available space, but even [removing unneeded files](https://github.com/abretaud/tools-iuc/pull/3/files#diff-354f30a63fb0907d4ad57269548329e3R77) is not enough
So +1 to leave antismash in https://github.com/bgruening/galaxytools/tree/master/tools/antismash

(I also had this idea that _perhaps_ we could create a loopback compressed zfs or btrfs partition to store conda environments. But well. I'm not brave enough to try it :D)

bgruening:
> (I also had this idea that perhaps we could create a loopback compressed zfs or btrfs partition to store conda environments. But well. I'm not brave enough to try it :D)

Before we do this, lets simply switch to Circle-CI ... they seems to have way more disc-space available :)

abretaud:
> Before we do this, lets simply switch to Circle-CI ... they seems to have way more disc-space available :)

Yes, it sounds much more reasonable!

PR #1701, opened 2018-02-01
Update to version 0.4.4

nekrut:
This release of unicycler introduces new option `--existing_long_read_assembly`. Unfortunately it cannot yet be used in our case because Racon bioconda package is in conflict with unicycler due to zlib version incompatibility.


bgruening:
Let's see it works for me locally.

nekrut:
@bgruening `<requirement type="package" version="3.5">python</requirement>` is needed because it installs with py36 while spades are only available for py35, so if you install unicycler0.4.4 from conda the way it is right now it will not work because spades will die. So the only way to run unicycler right now if you are installing it from conda is by something like this `conda create -n unicycler unicycler=0.4.4 python=3.5`

bgruening:
I fixed this upstream instead of adding this hack. It should be all good.

PR #1705, opened 2018-02-05
detection of insufficient memory for seqret (EMBOSS)

bernt-matthias:
I found that emboss' seqret gives a specific message on insufficient memory: `Uncaught exception:  Allocation failed, insufficient memory available, raised at ajstr.c:2089`

So I added a stdio tag that detects this. Could be added to all tools. Even if the emboss documentation suggests that the error code is always 0 this was not the case here. So I left the exit code 1: detection. I you think this is useful I could add it to all other tools of the suite.

Question is `<code file="emboss_format_corrector.py" />` used at all? When is this called? I could not find a corresponding call in the job directory.

### DIFF
--- a/tools/emboss_5/emboss_seqret.xml
+++ b/tools/emboss_5/emboss_seqret.xml
@@ -4,6 +4,7 @@
     <import>macros.xml</import>
   </macros>
   <expand macro="requirements" />
+  <expand macro="stdio" />
   <code file="emboss_format_corrector.py" />
   <command>seqret -sequence '$input1' -outseq '$out_file1' -feature $feature -firstonly $firstonly -osformat2 $out_format1 -auto</command>
   <inputs>
--- a/tools/emboss_5/macros.xml
+++ b/tools/emboss_5/macros.xml
@@ -5,6 +5,12 @@
             <requirement type="package" version="@VERSION@">emboss</requirement>
         </requirements>
     </xml>
+    <xml name="stdio">
+        <stdio>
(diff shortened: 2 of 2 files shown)


bernt-matthias:
For curiosity: why is the ID seqret**84**?

mvdbeek:
> Question is <code file="emboss_format_corrector.py" /> used at all? When is this called.

This file contains a `exec_before_job` job hook. I think those are deprecated, if you look at the content it does some fairly "bad things" like dynamically changing output datatypes to be galaxy compatible (e.g ncbi -> fasta).

I don't know this tool well enough, but I think those things may not be required (and be silently ignored) in a modern galaxy.

mvdbeek:
Actually this may still work and be required, since you can choose output datatypes that galaxy doesn't know. I think that may have been an artefact of wanting to support the tools with minimal manual modification of the wrappers.

bernt-matthias:
At least I do dot see a trace in the job files. Lets ask @nsoranzo  and @martenson who seem to have contributed these tools.

nsoranzo:
We just imported these old tools from https://github.com/galaxyproject/tools-devteam

bernt-matthias:
OK. Should we use the macro in all emboss tools?

nsoranzo:
@bernt-matthias Thanks and sorry for the late merge! If you want to take a look at https://github.com/galaxyproject/tools-iuc/pull/2261 in exchange :wink:

bernt-matthias:
Thanks, seems that I was to late for https://github.com/galaxyproject/tools-iuc/pull/2261

PR #1709, opened 2018-02-07
Add genomicfeatures to requirements

mvdbeek:
this is required by out wrapper in https://github.com/galaxyproject/tools-iuc/blob/76065f751b678abf87e6c723be680fecaa8cb589/tools/deseq2/deseq2.R#L240, which is not always used, but it will be used for running deseq2 on salmon data.


bgruening:
Thanks for the catch? Any idea why the salmon test does not catch this, should be enhance the test?

mvdbeek:
I guess this is only used if you have GTF input, and we test with tabular files, so it should be sufficient to get the GTF variant of the tabular file.

nekrut:
This is badly needed!

PR #1715, opened 2018-02-11
Some fixes for EGSEA

mblue9:
- add statmod as requirement (to fix error obtained if fry is selected in gene set methods), and add test to verify that all gene set methods work
- add option to retrieve and use updated KEGG pathways and add test
- fix RData file output (was not outputting when selected)
- increase wrapper version


tools/egsea/egsea.xml
```diff
@@ -186,6 +189,7 @@ Rscript '$__tool_directory__/egsea.R'
                 <option value="keggsig">Signalling pathways</option>
                 <option value="keggdis">Disease pathways</option>
             </param>
+            <param name="kegg_updated" type="boolean" truevalue="True" falsevalue="False" checked="False" label="Download KEGG pathways?" help="Select Yes if you want to download the most recent KEGG pathways. Default: No"/>
```
bgruening: This will break potentially break reproducibility, maybe we should mention this.

mblue9:
Thanks for the review @bgruening, I've added a sentence on reproducibility for the KEGG download option.

mblue9:
Except now it's failing here because looks like the pathview server is down (needs to access e.g. https://pathview.uncc.edu/data/khier.tsv)

bgruening:
As a reference: http://genome-source.cse.ucsc.edu/gitweb/?p=kent.git;a=blob_plain;f=src/hg/lib/encode/narrowPeak.as;hb=HEAD

mblue9:
@bgruening did you mean to put that narrowPeak reference in the bigBed issue https://github.com/galaxyproject/galaxy/issues/5419

bgruening:
Blubb, yes wrong issue :(

bgruening:
Thanks @mblue9!

PR #1716, opened 2018-02-11
update conda version

bgruening:
another bugfix release


mvdbeek:
I'd prefer to start with a PR to galaxy, otherwise we end up in the situation where tests pass here, but new, clean instances can't run the tools that we tested. Apart from that 👍

nsoranzo:
@mvdbeek We've done that before, basically to beta test a conda version before updating Galaxy.

PR #1718, opened 2018-02-12
rolling over lastz from devteam. major upgrade is coming

nekrut:
Major upgrade to lastz

 - switch to conda
 - incorporate multiple additional parameters


mvdbeek:
@nekrut can we just port the unmodified version in a first PR, and then modify things ? This would be a lot easier to review.

nekrut:
so it is not modified at this point. What do you want me to do?

mvdbeek:
I'd favor merging as is, if that's OK with the rest of the IUC

nekrut:
ok. then I can start modifying

bgruening:
👍

nsoranzo:
As mentioned on Gitter, if we merge this as it is, all future PRs will fail at the linting stage.

mvdbeek:
At that point we could blacklist lastz, but it's probably not worth the trouble. Sorry for the noise, it's probably better to get it in on one PR.

nekrut:
This tool will *completely* gutted, so I think future builds will be just fine

bgruening:
@nekrut you should be all set. Here is the kill-lastz PR for devteam: https://github.com/galaxyproject/tools-devteam/pull/502

nsoranzo:
@nekrut @bgruening I've pushed a commit to master to fix the flake8 issues in `tools/lastz/lastz_wrapper.py`, because only tools in `deprecated/` are ignored by flake8. `.tt_blacklist` is used only for planemo shed_lint/test/shed_update.

nekrut:
`lastz_wrapper.py` will no longer be there after this update

nsoranzo:
@nekrut I understand, the problem was for all the unrelated PRs between the merge of this PR and the PR that will remove `lastz_wrapper.py` . For simplicity we always lint all Python files, not just the ones touched in a PR.

PR #1720, opened 2018-02-12
Add Msaboot tool

matthew-fogel:
New tool for bootstrapping multiple sequence alignment files.


tools/msaboot/.shed.yml
```diff
@@ -0,0 +1,9 @@
+categories:
+- Fasta Manipulation
+description: A multiple sequences alignment bootstrapping tool.
+homepage_url: https://github.com/phac-nml/msaboot
+long_description: This tool can be used to generate bootstrapped replicates of multiple sequence alignments. The bootstrapping is done on the columns of the alignment.
+name: msaboot
+owner: phac-nml
```
bgruening: This tool is not yet in the toolshed, so we could make the owner IUC if you agree.

tools/msaboot/.shed.yml
```diff
@@ -0,0 +1,9 @@
...
+remote_repository_url: https://github.com/phac-nml/msaboot
```
bgruening: https://github.com/galaxyproject/tools-iuc/tree/master/tools/msaboot

tools/msaboot/msaboot.xml
```diff
@@ -0,0 +1,75 @@
+<?xml version="1.0"?>
+<tool id="msaboot" name="MSABOOT" version="@VERSION@">
+    <description>Output PHYLIP file with bootstrapped multiple sequence alignment data.</description>
```
bgruening: the dot can be removed.

tools/msaboot/msaboot.xml
```diff
@@ -0,0 +1,75 @@
...
+        <token name="@VERSION@">0.1.0</token>
+    </macros>
+    <requirements>
+        <requirement type="package" version="@VERSION@">msaboot</requirement>
+    </requirements>
+    <version_command>msaboot</version_command>
+    <command detect_errors="aggressive">
+        <![CDATA[
+	    msaboot
+      -i $input -n $number -o $output > $log 2>&1
```
bgruening: * input and output needs to be single quoted
* the indentation could be adjusted

tools/msaboot/msaboot.xml
```diff
@@ -0,0 +1,75 @@
...
+        ]]>
+    </command>
+    <inputs>
+        <param name="input" format="fasta" type="data" label="FASTA input file" />
+        <param name="number" type="integer" value="1" label="Specify the number of bootstrapping replicates to generate." />
```
bgruening: do you think a min="1" would make sense?

tools/msaboot/msaboot.xml
```diff
@@ -0,0 +1,75 @@
...
+    </inputs>
+    <outputs>
+        <data name="output" format="phylip" label="${tool.name} on ${on_string}:out.phylip" />
+        <data name="log" format="txt" hidden="true" label="${tool.name} on ${on_string}:log_msabootpl_run.txt"/>
```
bgruening: should the log file be an optional output? In this case you could add a boolean parameter asking for this log file.

tools/msaboot/msaboot.xml
```diff
@@ -0,0 +1,75 @@
...
+
+You can unhide the hidden output by clicking on the cog wheel next to the History panel and select "Unhide Hidden Datasets".
+
+(A) The bootstrapped replicates in Phylip format.
+
+(B) The hidden log file, containing information about the msaboot run to bootstrap the multiple sequence alignment data.
+
+]]>
+    </help>
+    <citations>
```
bgruening: You could add here some bibtex to reference your github repo, like here: https://github.com/galaxyproject/tools-iuc/blob/eea727c3bdfe36d9d16036d5ab79fb8b27c4e82e/tools/umi_tools/macros.xml#L6

tools/msaboot/msaboot.xml
matthew-fogel: I will make this log file an optional output, thank you for your advice!

tools/msaboot/.shed.yml
matthew-fogel: Yes, I will change it accordingly.

tools/msaboot/msaboot.xml
matthew-fogel: Thank you, I have removed it.

tools/msaboot/msaboot.xml
matthew-fogel: Yes.

tools/msaboot/.shed.yml
matthew-fogel: Thank you again!

matthew-fogel:
Thank you for your review(s)! I added a sample log file for the test case and other than that implemented the changes that you suggested.

bgruening:
@matthew-fogel thanks a lot!

Issue #1383, opened 2017-06-23
Mothur_aligndb loc file should be divided


arbernard:
After testing the latest mothur wrappers, I feel like the mothur_aligndb loc file should be divided.

At the moment this file can both reference files that are alignments (used as reference for the align.seqs tool for example) and simple fasta files (used as reference for the classify.seqs tool for example).

IMHO, this situation is an issue for the end users since it might lead them in selecting improper/incompatible references in their analysis.

(ping @shiltemann)

shiltemann:
thanks @arbernard, I agree completely, will work on this after GCC (or perhaps during hackathon)

bgruening:
Two loc files are probably a good idea, but in theory you could also add an additional column and filter by this ... e.g. a fasta column.

shiltemann:
good idea @bgruening, some tools can take both aligned or unaligned files as input, this may very well be true for these reference data as well for some of the tools, this approach would keep that flexibility at least

bernt-matthias:
I guess the easiest way would be to add a column to the end (then we have backward compatibility).

PR #1721, opened 2018-02-14
mothur data_manager add UNITE and update SILVA reference DB

bernt-matthias:
- adds SILVA132 and UNITE
- data manager tests 
- adds column to aligndb differentiating aligned/unaligned data

TODOs

- [x] check latest data here: https://unite.ut.ee/repository.php with tests
- [x] add dm tests
- [x] subdivide loc file
  - [x] for downloads
  - [x] for files from the local file system
-  [x] fix git history. no idea what went wrong again.
- [ ] update tools using aligndb
  - [ ] chimera.uchime, chimera.ccode, chimera.vsearch, chimera.pintail, chimera.slayer, chimera.check
  - [ ] classify.seqs 
  - [x] seq.error: accepts both verified with new test
  - [x] align.seqs: now only accepts aligned data
  - [ ] make.lookup


bgruening:
bump version number?

bernt-matthias:
Did not know about the versioning scheme used here. So I increased by the smallest possible.

shiltemann:
@bernt-matthias yes, the aligndb loc file currently contains both aligned and unaligned references, so you can add UNITE there as well. 

Some tools can take both formats, some only one or the other. We should probably split the loc file at some point (as also suggested by @arbernard in #1383) and also update the datamanager accordingly, however, I need to work out which of the tools that use it can take which formats and the documentation isn't always super clear about this.

P.S. tools that use the locfile: `chimera.uchime.xml chimera.ccode.xml chimera.vsearch.xml classify.seqs.xml chimera.pintail.xml chimera.slayer.xml chimera.check.xml seq.error.xml align.seqs.xml make.lookup.xml`

bernt-matthias:
Got it. I also asked our mothur users at which steps and how they use UNITE.

bernt-matthias:
First, I would like to ask if there is the possibility to test a data_manager with planemo?

I started to work on including UNITE. Since the file naming scheme is a bit awkward I was wondering if the data_manager should _apply some search and replace actions for the file names_: 

- The file names follow the scheme `UNITEv6_sh_THRESHOLD[_s]` (Threshold is 97, 99, or dynamic) and the _s indicates if the so called singleton RepS are included. I guess it is hard for users to make sense of this. 
- The worst thing is the `v6` even if its a newer version (I already contacted the UNITE guys/girls). 
- For the UNITE + INSD data sets are named `UNITE_public_mothur[_full]_01.12.2017` which does not describe the data sufficiently (INSD is missing .. and what is the meaning of full).

bgruening:
I don't think planemo supports data_managers yet.

bernt-matthias:
@bernt-matthias finish this :)

bernt-matthias:
Started a bit research: 

- chimera.uchime, chimera.ccode, chimera.vsearch, chimera.pintail, chimera.slayer, chimera.check seem to accept both
- classify.seqs according to the wiki "Does the reference need to be aligned? No", So I assume that both is OK
- seq.error has an aligned parameter which should specify if the references are aligned. implementing this seems tricky. lets see
- align.seqs this one needs aligned reference
- make.lookup no idea, but `help(make.lookup)` led me to the wild guess that sequences are required

So I would suggest to adapt only seq.error, and align.seqs accordingly and leave the others until someone complains :)

bernt-matthias:
Two important design decisions: 

1. How to handle the upgrade of the data table format (new column)? I discussed a bit with @nsoranzo on gitter and it seems we either replace the data tables by new ones, or we upgrade and require the admins to update existing data tables. In both ways admins will probably update manually, because I guess that they won't download the reference data again. So I would prefer the first option.

I also think that we should update the data table format of all tables to the new style format, since this is can be used in tests. And if we require admins to update the tables, then we should do it right.

2. Should be replace the upload from the Filesystem in the data manager by upload from history. Via import from a library I guess we can still have the same functionality. The big plus is that we can write tests for this (which seems impossible for integration from the filesystem).

shiltemann:
@bernt-matthias I agree, let's update the data tables properly if we can't avoid requiring admins to do some manual curation anyway.

Thanks for all your work on this!

data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
```diff
@@ -65,15 +70,94 @@
         <data name="out_file" format="data_manager_json" />
     </outputs>
     <tests>
+<!-- TODO       <test>-->
```
bgruening: is that still a todo?

tools/mothur/align.seqs.xml
```diff
@@ -50,7 +50,10 @@ echo 'align.seqs(
             </param>
```
bgruening: bump tool version?

tools/mothur/seq.error.xml
```diff
@@ -78,6 +78,19 @@ echo 'seq.error(
         <param name="count" type="data" format="mothur.count_table" optional="true" label="count table"/>
```
bgruening: bump tool version?

data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
bernt-matthias: done

tools/mothur/align.seqs.xml
bernt-matthias: done

data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
bernt-matthias: Nope. Not done. 

I would like to have a test, but I don't know how to test with a local file...? Was wondering if this is a good idea anyway. As far as I remember there is a generic data manager to import files from the local filesystem to arbitrary data tables ...? So maybe we can remove this here or adapt it to data sets from the history.

tools/mothur/align.seqs.xml
bernt-matthias: done

tools/mothur/seq.error.xml
bernt-matthias: done

pmpcosta:
This is long overdue, it is already stale (SILVA is at version 138). Can I do something to help? Why has the last build failed?

bernt-matthias:
Could add 138. Do you have the download link(s) at hand?

bernt-matthias:
> This is long overdue, it is already stale (SILVA is at version 138). Can I do something to help? Why has the last build failed?

I guess the reason is because the parts of the PR  might be controversial: i.e. the introduction of the new column in `mothur_aligndb`.

pmpcosta:
Here:
https://mothur.s3.us-east-2.amazonaws.com/wiki/silva.nr_v138.tgz
https://mothur.s3.us-east-2.amazonaws.com/wiki/silva.seed_v138.tgz

bernt-matthias:
Thanks @pmpcosta 

I just noticed on the mothur wiki that it also lists `Silva-based alignment of template file for chimera.slayer (5,181 sequences)` .. seems to be the same for all versions, or. 

Problem here is also that this file is written to the `aligndb` data table .. As mentioned in https://github.com/galaxyproject/tools-iuc/issues/1383 aligndb contains two different types of data .. this seems to be a third. Do you know if this is correct? If so we could use a third value in the new column.

Of course we should also update the tools that use this data table. The comments above might contain sufficient information to do so. If you have further insights they would be very welcome.

pmpcosta:
It is complicated. The original "gold" alignment (the greengenes one) was specially crafted for the original implementation of ChimeraSlayer at the Broad Institute. The SILVA gold alignment was a "new" gold alignment specifically made for mothur reimplementation of the algorithm, chimera.slayer, which is now considered legacy. Both these specially crafted alignments are meant to be used with either chimera.slayer or chimera.uchime (a newer related algo), and nothing else. No new versions are likely. I would suggest that these silva "gold" alignment be separated from the other silva files, as the greengenes already is. I don't believe they should be considered a third type of alignment.

Going through this, I noticed some inaccuracies in the mothur READ.ME file, I think I will do a PR for it.

bernt-matthias:
> I don't believe they should be considered a third type of alignment.

Thanks for the explanations. Fine with me.

> Going through this, I noticed some inaccuracies in the mothur READ.ME file, I think I will do a PR for it.

Great.

bernt-matthias:
So little update from me. Hope that all tools work now. Still a bit puzzled how to manage the data table update (a column was added ). 

The new tool versions will ignore old entries ... which might be acceptable. Not sure how old versions will react on the new data table column. Also untested how the merging of the data tables on startup will react. 

Alternative would be to introduce a new data table. But I'm also unsure how the update of `tool_data_table_conf.xml` will work out at the end. 

Anyway, we need a decision here. This PR is open for way to long .. I'm happy to finish this .. the sooner the better.

Also a bit unsure about the decisions on which of the changed tools take aligned data, which take sequence data, and which take both for the two affected inputs (`name="fasta"` and `name="template/reference"`). Maybe @pmpcosta can shed some light on this.

bgruening:
Shipping?

bernt-matthias:
Now in https://github.com/galaxyproject/tools-iuc/pull/3765

Issue #1644, opened 2017-12-21
Mothur Make.biom does not recognize tabular datatype dataset as "metadata" input


jennaj:
Make.biom (version 1.36.1.0)
https://toolshed.g2.bx.psu.edu/view/iuc/mothur_make_biom/5dc4d49a238a

Dataset 10 in this test history is an example of proper tabular input for the "metadata" tool form options, but it cannot be selected: https://galaxy.uni-freiburg.de/u/jenj/h/test-history-mothur-subsample

True also at https://usegalaxy.org, so either a problem with the tool wrapper or Galaxy itself (v17.09). If not the tool, we can move this over to the /galaxy repo. The other two primary inputs capture appropriate input dataset fine.

Used in the [16S metagenomics GTN tutorial](http://galaxyproject.github.io/training-material/topics/metagenomics/tutorials/mothur-miseq-sop/tutorial.html), step 'Hands-on: Phinch"

from `make.biom.xml`
```
 <inputs>
        <param name="otu" type="data" format="mothur.shared" label="shared - OTU Shared file"/>
        <param name="constaxonomy" type="data" format="mothur.cons.taxonomy" optional="true" label="contaxonomy - consensus taxonomy" help="The contaxonomy file is the taxonomy file outputted by classify.otu"/>
        <param name="metadata" type="data" format="tabular" label="metadata" optional="true" help="You can add sample data support here"/>
```

vparreir:
same problem. any suggestions?

shiltemann:
as a workaround you can change the format of your metadata file to `tabular`. 

I have been working on updating the wrappers to the latest version of mothur and have incorporated a fix for this issue there (hope to submit the PR by the end of the day)

vparreir:
Hi Saskia,


Yes, the format of the metadata file is tabular, I also have another copy as txt, but still doesn't work.

Thank you for your quick response.

Valeria






________________________________

shiltemann:
@vparreir oh hmm, thanks for letting me know, that is strange, I thought the problem was that the metadata file from the tutorial gets sniffed wrongly, ok, I will look into it in more detail

shiltemann:
hmm, I don't have this issue locally or on our server. @jennaj I dont have permission to view the files in the history you shared, could you share it with me (saskia, zazkia@gmail.com on that server)

shiltemann:
[...]

Issue #1643, opened 2017-12-20
Mothur Venn does not produce output with format/datatype set to "svg"


jennaj:
Venn Generate Venn diagrams for groups (Galaxy Version 1.36.1.0)
Galaxy Tool ID: | toolshed.g2.bx.psu.edu/repos/iuc/mothur_venn/mothur_venn/1.36.1.0
Galaxy Tool Version: | 1.36.1.0
Tool Version: | v.1.36.1
https://toolshed.g2.bx.psu.edu/view/iuc/mothur_venn/cc4a296dd5d0

Problem: The .svg output has no "format" assigned (the tabular and logfile output are fine). When downloaded, a file name like `Galaxy223-[Venn_on_data_221__svg__0.03.sharedsobs.F3D0-F3D1-F3D2-F3D3] (1)` is given. The default extension is `.null` - but this can be changed by the user in the download dialog to be `.svg` (at least on MAC OSX). It can be difficult to modify this after download (user must know how to access file extensions....). Other Mothur tools that produce .svg output are labeled and download with the .svg extension (so users can click on the output locally and view easily)

Example of a tool with .svg properly assigned is:

Galaxy Tool ID: | toolshed.g2.bx.psu.edu/repos/iuc/mothur_heatmap_sim/mothur_heatmap_sim/1.36.1.0
Galaxy Tool Version: | 1.36.1.0

Example history contains a run-through of the [GTN 16S metagenomics tutorial](http://galaxyproject.github.io/training-material/topics/metagenomics/tutorials/mothur-miseq-sop/tutorial.html#otu-based-analysis), section "Hands-on: Venn diagram"). See dataset 223 for the venn output, dataset 220 for the heatmap output: https://usegalaxy.org/u/jen/h/16s-microbial-analysis-with-mothur---proccess

If I am misunderstanding what this tool should be outputting, please correct me. Thanks!

bgruening:
Hopefully fixed with: https://github.com/galaxyproject/tools-iuc/pull/1722

jennaj:
Thanks - version 1.39.5.0 resolved both tool problems!

Issue #1625, opened 2017-12-08
Mothur - error on libboost_iostreams.so.1.57.0


vebaev:
Hi,
I have installed some particular tools from the Mothur tutorial (installed ok, green), however, the first tool Merge.files produce an error (Galaxy v17.09):

Fatal error: Exit code 127 ()
mothur: error while loading shared libraries: libboost_iostreams.so.1.57.0: cannot open shared object file: No such file or directory

bgruening:
Cab you please test the latest version and let us know if it still is not working?

Issue #1621, opened 2017-12-07
mothur view log files starts download of binary file


bernt-matthias:
Instead of displaying the log viewing the log files of some tools triggers a download (of a binary file). 
Examples are: 
-  Align.seqs
- Screen.seqs
- Make.contigs

Some work as expected: Unique.Seqs, Summary.Seqs

bgruening:
Hopefully fixed with: https://github.com/galaxyproject/tools-iuc/pull/1722

PR #1722, opened 2018-02-14
Mothur version update

### LINKED ISSUE #1688: Mothur screen.seqs with a contigs report
Hello everyone,

One of my colleague that uses Mothur for her analysis recently show me that there are two issues with the current Screen.seqs wrapper when you execute it with an input contigs report file (generated by Make.contigs).

The first one is that the option displayed in the help of the contigsreport parameter (Ie. minoverlap, ostart, oend and mismatches) can't be set with the current Screen.seqs wrapper.

The second one is that, in these conditions, Mothur generates a modified contigsreport (as displayed in its log) that is not recovered by the wrapper.

It would be nice if these parameters could be made available (and the output file recovered) in the next iteration of the screen.seqs wrapper.

This is not a major issue so it can probably wait for the major update of the wrapper (migration to Mothur 1.39.5).

Thanks in advance for the enhancements :)

ping @shiltemann

shiltemann:
started at GCC with @Takadonet and @ValentinChCloud, this updates mothur to the latest version, 1.39.5, and does some general cleanup etc of wrappers

should also fix #1688, #1644, #1643, #1472 and incorporates changes from #1569 (thanks @arbernard!) 
I have not encountered #1625, #1572 locally,  let's see if travis agrees
I could not reproduce #1621, let me know if you're still running into this with new wrappers @bernt-matthias 

ping @arbernard @Takadonet @bernt-matthias @jennaj thanks for all the help and bugreports, maybe you would like to test/review this? :)


Takadonet:
Will for sure take a look! Awesome work.

shiltemann:
@bernt-matthias am pushing to the testtoolshed now, thanks!

bernt-matthias:
@shiltemann I guess you have an answer to my latest question in https://github.com/galaxyproject/tools-iuc/pull/1721 on the integration of the UNITE DB in the mothur data manager...?

bernt-matthias:
@shiltemann will you also update suite_mothur in the testtoolshed?

shiltemann:
@bernt-matthias ah yeah, something went wrong the first time, but should be good to go now :)

shiltemann:
@nsoranzo the last tool I added seems to have just pushed this over the max time limit for travis again, shall I make a 4th chunk?

nsoranzo:
I guess there's no other choice!

bgruening:
@shiltemann @bernt-matthias @abretaud @Takadonet ready to go?

bernt-matthias:
I forgot about this one. Sorry. I will test this afternoon.

Some quick comments (see https://github.com/galaxyproject/tools-iuc/issues/1472): 
- I still think that the 3x log output (as data set, in the stdio/stderr, and in the data set preview) is very redundant. I my opinion the stdio/stderr in the data set information page plus an **optional** log data set  would be great.
- have you tried to add the -q switch to mothur (to remove unnecessary log output)
- are citations now automatically added (command(help))?

shiltemann:
@bernt-matthias oh yeah, I have made the logfile an optional output now, which is off by default, remembering your comment (but maybe only after I pushed to the test toolshed), is that ok for you?

I will look into the citations thing, but it may be nicer to just add that the "Galaxy way" ..I will have a look at which tools are missing citations and add those

bernt-matthias:
Currently redoing http://galaxyproject.github.io/training-material/topics/metagenomics/tutorials/general-tutorial/tutorial.html with the new tools.

One first comment: 

Unique.seqs set "output format" needs to be changed to "Name file" (expected as input for count.seqs)
- maybe the default changed / the tutorial was buggy
- alternatively: can the tool output both: Name file and count table at the same time

jennaj:
@bernt-matthias This is a new issue for Unique.seqs afaik -- I haven't rerun the tutorial again at https://usegalaxy.org since late Dec. Open issues I am tracking to run the tutorial on that server are here and includes some issues that probably impact usage anywhere: https://github.com/galaxyproject/usegalaxy-playbook/issues/74

@shiltemann Thanks the updates!

shiltemann:
@bernt-matthias ah yes, they changed the default here. Thanks a lot for testing the training materials, I will update the tutorial accordingly :)

bernt-matthias:
Screen.seqs: The minlength default is 10. For all other parameters its -1. Does -1 (probably ignore) also work here?

Cluster.split  seems to have more options. is Opticlust the Clustering Method of choice

shiltemann:
@bernt-matthias mothur uses a default minlength of 10, so I did same in wrapper. Same for Cluster.split, opti is indeed the new default in this version. 

We will need to update both the mothur tutorials to explicitly set any parameters that are no longer the default. I started this here: https://github.com/galaxyproject/training-material/pull/715

bernt-matthias:
Otherwise everything seems fine.

shiltemann:
@bernt-matthias thanks a lot! I suspect that is due to the new clustering method (if different clustering levels give same results, mothur only reports biggest) I changed the tutorial to set the method back to the previous default of `average neighbour`

bernt-matthias:
@shiltemann you are right average neighbour gives the usual results

bgruening:
Thanks all this is an amazing team work!

PR #1723, opened 2018-02-15
Lastz rewrite

nekrut:
complete rewrite of lastz greatly expanding options and output formats


bgruening:
@nekrut can you rebase your PR against master?

tools/lastz/lastz.xml
```diff
@@ -0,0 +1,508 @@
+<tool id="lastz" name="align long sequences" version="@VERSION@.1">
```
bgruening: this can start with 0

tools/lastz/lastz.xml
```diff
@@ -0,0 +1,508 @@
...
+  <macros>
+    <token name="@VERSION@">1.0.4</token>
+  </macros>
+  <requirements>
+    <requirement type="package" version="@VERSION@">lastz</requirement>
+  </requirements>
```
bgruening: indentation should be 4 here as well

tools/lastz/lastz.xml
```diff
@@ -0,0 +1,508 @@
...
+    <command detect_errors="exit_code"><![CDATA[
+      lastz
+
+      #if $source.ref_source=="history":
+        "${source.target}"
+      #else:
+        "${source.target_2bit.fields.path}"
+      #end if
+      #if $query:  
+        "${query}"
```
bgruening: single quotes for text, data and output parameters

tools/lastz/lastz.xml
```diff
@@ -0,0 +1,508 @@
...
+      <param name="filter_ngap" type="integer" min="0" optional="true" argument="--filter=ngap:0.." label="Filter alignments by the number of gaps" help="Allow no more than this number of gaps. Gap count, or ngap, is the number of runs of gapped columns in the alignment (each run is counted as one gap)."/>
+      <param name="filter_cgap" type="integer" min="0" optional="true" argument="--filter=cgap:0.." label="Filter alignments by the number of gap columns" help="Allow no more than this number of gaps. Gap column count, or cgap, is the number of gapped columns in the alignment (each column is counted as one gap)."/>
+      <param name="notrivial" type="boolean" truevalue="--notrivial" argument="--notrivial" label="Do not output a trivial self-alignment block if the target and query sequences are identical" help="Note that using ‑‑self automatically enables this option."/>
+    </section>
+
```
bgruening: these empty lines can be removed

nekrut: I need them to see where blocks begin and end

tools/lastz/lastz.xml
nekrut: will fix

tools/lastz/lastz.xml
nekrut: this is the way my editor works. I'll fix it later

tools/lastz/lastz.xml
```diff
@@ -0,0 +1,640 @@
+<tool id="lastz" name="LASTZ" version="@VERSION@.1">
```
bgruening: @nekrut if we want to keep the tool linage we should keep the old ID `lastz_wrapper_2`.

bgruening:
can you rebase it against master or should I try this during merge?

tools/lastz/lastz.xml
nekrut: I don't see the point in keeping the lineage

nekrut:
If I only knew what rebase it against master means ;)

nekrut:
It's not yet ready anyway ... still needs tests

tools/lastz/lastz.xml
nsoranzo: Two points for keeping it:
- the new and lastz will appear as a single entry in the tool panel instead of 2
- it will be possible to update a workflow to use the new version without removing and readding the tool.

nekrut:
This is done. @bgruening can you help me get the tests going? Tx.

bgruening:
@nekrut I fixed the tests and rebased the branch. I also changed the ID back and needed to set the version number to a higher value as the previous one was bigger. Hope that is ok.

nekrut:
Let's not merge yet -> I want to add substitution score estimator

nekrut:
@bgruening why tests do not run?

bgruening:
@nekrut now it would be good to use a shared marco file and to adopt the shed.yml file to be a suite.

nekrut:
Hmmm ... not sure where tabs came from ... Thanks! Yes, I'll get macros and shed.

nekrut:
@bgruening @nsoranzo @bebatut @shiltemann can someone merge (if you think it's ok)

nekrut:
@bgruening @nsoranzo @bebatut @shiltemann = PLEASE!

PR #1724, opened 2018-02-15
Krona: make wording for text input more generic

shiltemann:
I made a tool to convert mothur taxonomy files to krona text input format, so it is no longer restricted to MetaPhlAn output, changed wording to reflect this.

ping @bebatut


bebatut:
Nice! Thanks @shiltemann !

nsoranzo:
Maybe use `tabular` instead of `text`?

bgruening:
@nsoranzo  http://www.crs4.it can not be accessed.

nsoranzo:
@bgruening It's back.

PR #1725, opened 2018-02-16
fix get_ancestor_terms input type

bernt-matthias:
was text -> now obo (was probably forgotten)


bgruening:
Thanks @bernt-matthias!

PR #1726, opened 2018-02-16
STAR: Add two pass mode

TorHou:
This feature is now available in STAR 2.5 and I would really like to use it.


bgruening:
@TorHou version needs to be bumped and a test would be nice.
@yhoogstrate can you look at this?

TorHou:
I just realized that all tests that use the star-fusion setting will probably fail now, because I have changed those settings. Twopass mode for star-fusion makes perfect sense however and is the default setting, so I think the test data output should be updated.

TorHou:
ok. I checked the ouput of the tests, they are all as expected (it's a simple case, as should be). So I think this change is all good.

yhoogstrate:
@TorHou Indeed twopass makes also sense for fusion analysis. I don't see any reason why this implementation would fail and given that the test passes it's an OK from my side.

Thanks everyone :+1:

bgruening:
Thanks @TorHou and @yhoogstrate!

PR #1727, opened 2018-02-17
Fixes for MAGeCK

mblue9:
Found some things that needed fixing and made some small improvements.

I assume I need to update the wrapper version for this but do I update it in all the .xmls or just somewhere in the macros.xml? (I only modified some xmls and to varying degrees, ranging from cosmetic changes to changing defaults, so am not sure what version changes to make here.)


bgruening:
You need to change this in the XML that will potentially change there results due to your modifications. So not the macro but the single XML and by changing/adding a last digit `.1`

bgruening:
@mblue9 thanks.

PR #1729, opened 2018-02-20
Add Cutadapt

mblue9:
This PR is to add @lparsons's Cutadapt wrapper (from here https://bitbucket.org/lance_parsons/cutadapt_galaxy_wrapper) to IUC and to update if from v1.6 to the current Cutadapt version v1.16.

So far I've just:
 - changed the existing wrapper to use Conda cutadapt
 - updated to cutadapt v.1.16
 - got the existing tests work
 - removed some (old?) parameter attributes like length= and size=

I'll have a look at the changes since the existing wrapper (here: http://cutadapt.readthedocs.io/en/stable/changes.html) and see what options to add or modify.


mblue9:
Hmm I am not sure how I ended up with my diffbind commits in this PR, the 1st 4 - f7f5f7c, 0438838, b09ed51, ccea9bc. I was planning to add those in a separate PR (and really thought I had them in another branch).

What is the best way to remove those 4 commits from here and put in another PR?

bgruening:
ping @lparsons can you have a look at this?

bgruening:
@mblue9 you could do an interactive rebase against master and skip the commits. `git rebase master -i`

mblue9:
Ok thanks for the info @bgruening I'll attempt the rebase.

That would be great @lparsons if you could take a look. You might want to wait until after I've added in paired end support, the additional outputs etc. Might take me a couple of days, I'll say when I'm done.

Also there's just been a new release of Cutadapt- v1.16 yesterday and there's already a Bioconda package available (how cool!) so I'm also updating to that version now.

mblue9:
Oh hooray! I think that rebase worked.

lparsons:
Will do, let me know when you're ready @mblue9

mblue9:
@lparsons I'm done for now. Thanks in advance for the review :)

tools/cutadapt/README
```diff
@@ -0,0 +1,55 @@
+Galaxy tool definition for cutadapt (http://code.google.com/p/cutadapt/)
```
lparsons: Update URL to http://cutadapt.readthedocs.io/en/stable/index.html

tools/cutadapt/README
```diff
@@ -0,0 +1,55 @@
...
+------------------------
+
+The recommended way to install cutadapt as a tool in Galaxy is to the use the
+Galaxy Tool Shed (http://wiki.galaxyproject.org/Tool%20Shed).
+
+This will allow cutadapt to be installed automatically and keep track of older
+versions of cutadapt and the tool wrapper.
+
+
+Installation - Manual
```
lparsons: Consider removing the "manual" installation instructions. They are probably out of date anyway.

bgruening:
@lparsons one thing you need to do is to grant IUC permissions to push to your repo. Can you please do this on the TTS and TS?

tools/cutadapt/README
```diff
@@ -0,0 +1,55 @@
...
+
+Reading adapters from a fasta file is not supported
+Colorspace data support is not implemented
+Only one "Strip suffix" is suppored
+
+
+Galaxy Wrapper Development
+--------------------------
+
+Author: Lance Parsons <lparsons@princeton.edu>
```
lparsons: @mblue9 Probably add your name/email here

tools/cutadapt/.shed.yml
```diff
@@ -0,0 +1,14 @@
+name: cutadapt
+owner: lparsons
+description: Flexible tool to remove adapter sequences (and quality trim) high throughput sequencing reads (fasta/fastq).
+homepage_url: https://cutadapt.readthedocs.org/en/stable/
+long_description: |
+    Cutadapt finds and removes adapter sequences, primers, poly-A tails and other types of unwanted sequence from your high-throughput sequencing reads.
+
+    http://dx.doi.org/10.14806/ej.17.1.200
+remote_repository_url: https://bitbucket.org/lance_parsons/cutadapt_galaxy_wrapper
```
lparsons: Update this to the tools-iuc location.

tools/cutadapt/README
```diff
@@ -0,0 +1,55 @@
...
+
+Repository: [https://bitbucket.org/lance_parsons/cutadapt\_galaxy\_wrapper](https://bitbucket.org/lance_parsons/cutadapt_galaxy_wrapper)
```
lparsons: Update this to tools-iuc.

tools/cutadapt/cutadapt.xml
```diff
@@ -0,0 +1,595 @@
...
+            #end if
+            ln -f -s '${library.input_1}' ${read1} &&
+        #end if
+
+
+        ## Run Cutadapt
+
+        cutadapt
+
+        -j \${GALAXY_SLOTS:-4}
```
lparsons: The backslash looks odd here...

tools/cutadapt/cutadapt.xml
```diff
@@ -0,0 +1,595 @@
...
+
+        #if $library.input_1.extension.startswith( "fastq"):
```
lparsons: Is this perhaps redundant now with the added code to create properly named links?

bgruening:
I think you need to go to "manage repositories" -> "manage repository admins" and add IUC to this.
> Also, review is in progress... :-)

Yeah!

lparsons:
> I think you need to go to "manage repositories" -> "manage repository admins" and add IUC to this.

Done. I thought there was one more thing...

tools/cutadapt/cutadapt.xml
nsoranzo: @lparsons The backslash is needed to prevent Cheetah from trying to map the variable, which instead is going to be done by bash when the job is run.

mblue9:
Thanks for the review @lparsons ! The cutadapt wrapper was already pretty good I thought 😄 

I've made those edits now. But I decided to remove the README and just keep the important info, let me know what you think.

tools/cutadapt/cutadapt.xml
lparsons: Thanks @nsoranzo, that makes sense. Somehow I didn't realize that it's actually a BASH variable.

lparsons:
@mblue9 Looks great. I agree about removing redundant info. One the the README is useful for though, is communicating additional steps an administrator might want to take after tool installation. Two that I can think of are:

1. Updating the list of adapters. It would be nice to mention that the tools reads from a data table file and also point out the [info presented in the cutadapt docs about Illumina adapters](http://cutadapt.readthedocs.io/en/stable/guide.html#illumina-truseq), specifically
    > The adapter sequences can be found in the document [Illumina TruSeq Adapters De-Mystified](http://tucf-genomics.tufts.edu/documents/protocols/TUCF_Understanding_Illumina_TruSeq_Adapters.pdf).
2. The fact that the tool uses `GALAXY_SLOTS` and thus the admin might want to add the tool to the `job_conf.xml`

mblue9:
Thanks for the feedback @lparsons! I would be happy to add that info to the README but just to check first - is this the recommended way to provide info on data tables and GALAXY_SLOTS for IUC tools? 

As I did not know that (and have not been doing that 😟) and can see a bunch of IUC tools that use GALAXY_SLOTS but have no READMEs. 

If this is the best way to provide this info to admins then I assume it should be added to all the IUC tools?  

If all those tools need READMEs I could add it to my to dos if noone else gets to it first but please let me know what the thoughts on this are.

bgruening:
@mblue9 we do not ship those infos via Readme. If this is really needed we should to this via the TS interface and link to admin docs etc ... I don't think a tool dev should have the need to care about this, we should do better in Galaxy to make admins aware. An other point is that more and more Admins installing tool via scripts. In this case you also don't see the readme.

bgruening:
@lparsons please feel free to merge this one if you think its ready.

lparsons:
@bgruening I agree that the README is not ideal, but at the moment, it's the best (only) mechanism we have of getting that type of information out and I, for one, find it useful. IMHO, until there is a better way to disseminate that information, we should use what we have.

@mblue9 I certainly don't think it's worth going back through all of the old tools and adding README info, just thought since this was going through an update it was a good time to catch it.

Looks great though! Thanks so much @mblue9 ! Great work!

PR #1743, opened 2018-02-26
msaboot - increment version (bug fix)

matthew-fogel:
Incremented version number for msaboot wrapper due to small bug fix in the source code.


bgruening:
@matthew-fogel thanks a lot!

PR #1746, opened 2018-02-27
OrthoFinder_onlygroups : add -S option (choice between blast, blast_g…

Mataivic:
- Updated the version of OrthoFinder (2.1.2)
- Add the -S option, which allows to choose between blast/blast_gz/diamond
- Add a functional test for the -S option

### DIFF
--- a/tools/orthofinder/orthofinder_only_groups.xml
+++ b/tools/orthofinder/orthofinder_only_groups.xml
@@ -1,4 +1,4 @@
-<tool name="OrthoFinder OnlyGroups" id="orthofinder_onlygroups" version="1.1.4">
+<tool name="OrthoFinder OnlyGroups" id="orthofinder_onlygroups" version="2.1.2">
     <description>finds orthogroups in a set of proteomes</description>
     <requirements>
         <requirement type="package" version="2.1.2">orthofinder</requirement>
@@ -35,7 +35,8 @@
         ## start Orthofinder
         orthofinder
         #if $init.start=="fasta":
-            -f . -S $init.search_program
+            -f . 
+            -S $init.search.search_program
         #elif $init.start=="blast":
             -b .
         #end if
@@ -44,10 +45,12 @@
 
         #if $init.start=="fasta":
             mv Results_* results
-            #if $init.keepblastout=="yes":
-                && mkdir -p results/WorkingDirectory/blast results/WorkingDirectory/fa &&
-                mv results/WorkingDirectory/Blast* results/WorkingDirectory/blast/ &&
-                mv results/WorkingDirectory/*.fa results/WorkingDirectory/fa/
+            #if $init.search.search_program=="blast":
+                #if $init.search.keepblastout=="yes":
+                    && mkdir -p results/WorkingDirectory/blast results/WorkingDirectory/fa &&
+                    mv results/WorkingDirectory/Blast* results/WorkingDirectory/blast/ &&
+                    mv results/WorkingDirectory/*.fa results/WorkingDirectory/fa/
+                #end if
             #end if
         #elif $init.start=="blast":
             mkdir results  &&
@@ -67,12 +70,18 @@
 
             <when value="fasta">
                 <param name="input_fasta" type="data" format="fasta" multiple="true" label="Select input fasta proteomes" help="One fasta file per species; species and sequences names in the results will remain the same than in the input files."/>
-                <param name="search_program" type="select" label="Sequence search program" help="">
-                    <option value="blast" selected="true">blast</option>
-                    <option value="blast">blast_gz</option>
-                    <option value="blast">diamond</option>
-                </param>
-                <param name="keepblastout" type="boolean" checked="true" truevalue="yes" falsevalue="no" label="Do you want to get the blast results ?" help="Used to re-run OrthoFinder_OnlyGroups from pre-computed blast results"/>
+                <conditional name="search">
+                    <param name="search_program" type="select" label="Sequence search program" help="">
+                        <option value="blast" selected="true">blast</option>
+                        <option value="blast_gz">blast_gz</option>
+                        <option value="diamond">diamond</option>
+                    </param>
+                    <when value="blast">
+                        <param name="keepblastout" type="boolean" checked="true" truevalue="yes" falsevalue="no" label="Do you want to get the blast results ?" help="Used to re-run OrthoFinder_OnlyGroups from pre-computed blast results"/>
+                    </when>
+                    <when value="diamond"></when>
+                    <when value="blast_gz"></when>
+                </conditional>
             </when>
 
             <when value="blast">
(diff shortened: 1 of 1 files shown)


tools/orthofinder/orthofinder_only_groups.xml
```diff
@@ -76,7 +81,7 @@
                 <param name="specIDs" type="data" format="txt" label="Select the SpeciesIds file" help="SpeciesIDs.txt file from the blast output files of a previous OrthoFinder_OnlyGroups run."/>
                 <param name="seqIDs" type="data" format="txt" label="Select the SequencesIds file" help="SequencesIDs.txt file from the blast output files of a previous OrthoFinder_OnlyGroups run."/>
             </when>
-        </conditional>
+        </conditional>
```
abretaud: Can you remove the trailing spaces?

tools/orthofinder/orthofinder_only_groups.xml
```diff
@@ -67,7 +67,12 @@
 
             <when value="fasta">
                 <param name="input_fasta" type="data" format="fasta" multiple="true" label="Select input fasta proteomes" help="One fasta file per species; species and sequences names in the results will remain the same than in the input files."/>
-                <param name="keepblastout" type="boolean" checked="true" truevalue="yes" falsevalue="no" label="Do you want to get the blast results ?" help="Used to re-run OrthoFinder_OnlyGroups from pre-computed blast results"/>   
+                <param name="search_program" type="select" label="Sequence search program" help="">
+                    <option value="blast" selected="true">blast</option>
+                    <option value="blast">blast_gz</option>
+                    <option value="blast">diamond</option>
+                </param>
+                <param name="keepblastout" type="boolean" checked="true" truevalue="yes" falsevalue="no" label="Do you want to get the blast results ?" help="Used to re-run OrthoFinder_OnlyGroups from pre-computed blast results"/>
```
abretaud: Does it make sense when diamond is selected? If yes, maybe rename the option, if not a conditional would be good

tools/orthofinder/orthofinder_only_groups.xml
```diff
@@ -67,7 +67,12 @@
 
             <when value="fasta">
                 <param name="input_fasta" type="data" format="fasta" multiple="true" label="Select input fasta proteomes" help="One fasta file per species; species and sequences names in the results will remain the same than in the input files."/>
-                <param name="keepblastout" type="boolean" checked="true" truevalue="yes" falsevalue="no" label="Do you want to get the blast results ?" help="Used to re-run OrthoFinder_OnlyGroups from pre-computed blast results"/>   
+                <param name="search_program" type="select" label="Sequence search program" help="">
+                    <option value="blast" selected="true">blast</option>
+                    <option value="blast">blast_gz</option>
```
abretaud: If blast_gz is selected with keepblastout==yes, will the corresponding output will be a compressed blast file? I'm not sure there is a datatype for this in galaxy?

tools/orthofinder/orthofinder_only_groups.xml
```diff
@@ -35,7 +35,7 @@
         ## start Orthofinder
         orthofinder
         #if $init.start=="fasta":
-            -f .
+            -f . -S $init.search_program
```
abretaud: Could you move the -S option to a new line to make it more readable?

tools/orthofinder/orthofinder_only_groups.xml
```diff
@@ -1,7 +1,7 @@
 <tool name="OrthoFinder OnlyGroups" id="orthofinder_onlygroups" version="1.1.4">
```
abretaud: Update the version here too

tools/orthofinder/orthofinder_only_groups.xml
Mataivic: I realized I might be too tired today, very silly mistakes here ...

Mataivic:
I fixed the mistakes and I thinks it's good now. Concerning the -M, -A, -T, -R options, this wrapper only deals with the first part of OrthoFinder (the -og option, which stops the program when orthogroups are computed), so these options are not involved yet.

I wrapped only the first part because the orthofinder user of my team only needs the orthogroups for now ; and I did not take time to extend the wrapper to the whole program :/. If you wish to have the whole program under Galaxy, just tell me and I'll try to find some time in the upcoming weeks :).

tools/orthofinder/orthofinder_only_groups.xml
```diff
@@ -339,6 +418,13 @@ Input files
         - The SpeciesIDs.txt file
         - The SequencesIDs.txt file
 
+----------
+Parameters
+----------
+    - Sequence search program : You can choose either blast, blast_gx, or diamond (diamond is faster)
```
abretaud: a little typo on blast_gz

abretaud:
Thanks @Mataivic!
I think there's a problem with travis so I can't see if tests are passing
I'll need the other options, but if you don't have the time don't worry, I'll add them later

tools/orthofinder/orthofinder_only_groups.xml
```diff
@@ -92,17 +106,17 @@
         <!-- working directory : blast outputs-->
         <collection name="wdblast" type="list" label="Blast_outputs">
             <discover_datasets pattern="__name_and_ext__" directory="results/WorkingDirectory/blast/" />
-            <filter>init['start']=="fasta" and init['keepblastout']</filter>
+            <filter>init['start']=="fasta" and init['search']['search_program']=="blast" and init['keepblastout']</filter>
```
abretaud: Green!
A last thing: I think here (and below) it should be `init['search']['keepblastout']` instead of `init['search']['keepblastout']`
Can you check this and modify the tests to check this?

Mataivic: @abretaud I fixed the mistake but I can't figure out why the associated test passed (the 3rd, the 3 last outputs tested are linked to the keepblastout param). How can I modify it to be sure it works fine ?

abretaud: I don't understand either why it passed... Anyway you'd better write `init['search']['keepblastout'] == "yes"` in the filter (and maybe this is why the test passed previously = init['keepblastout'] got evaluated to True even though it was not "yes"?)

Mataivic: So that's a bit weird ; I checked with planemo test and planemo serve, and with ``init['search']['keepblastout'] == "yes"``, it does not work, and without the ``=="yes"`` it seems to work perfectly. Perhaps the tags ``truevalue`` and ``falsevalue`` are set only for the command an not for the tests ?

abretaud:
I think [in the tests](https://github.com/abims-sbr/tools-iuc/blob/f6b507bc986a400a29968153ca50991b98adb8eb/tools/orthofinder/orthofinder_only_groups.xml#L242), you need to set the value to `true` or `false` instead of `yes` or `no`

And then, you should be able to use `== "yes"` in the filter tag

Mataivic:
@abretaud When I do this, the test also fails (all the others are green), I don't understand why yet. The log indicates the correct commands, all the files and directories are present (I checked with the ``--no_cleanup`` option), but the output collection are not found (for example I got a log saying ``failed to find output 'wdblast' : []``. Do I keep searching or is it okay without the ``=="yes"`` ?

abretaud:
Ok, I took the liberty to make a little commit on your pr, now it passes the test on my laptop, let's see if it gets green

Mataivic:
Green !

PR #1747, opened 2018-02-28
fixing macs for multiple samples

bgruening:


mblue9:
This fix looks good to me. But a separate issue I just discovered is, if I use collections with macs2 the Control element.identifier gets used for the output identifier instead of the Treatment. And if you input multiple Treatments with the same control they all get the same identifier (from the control). I don't know where/how to change that 😕

mvdbeek:
@mblue9 @bgruening you can control which input is supposed to be the source of the identifier with `default_identifier_source `, see https://docs.galaxyproject.org/en/latest/dev/schema.html#id63

mblue9:
Thanks @mvdbeek ! How do you use `default_identifier_source`? as I just tried adding that into the <data../> params like so:

`<data name="output_tabular" format="tabular" default_identifier_source="input_treatment_file" label="${tool.name} on ${on_string} (Peaks in tabular format)">
            <filter> outputs and 'peaks_tabular' in outputs</filter>
        </data>`

but it fails linting with Planemo, I'm getting: 

`Schemas validity error : Element 'data', attribute 'default_identifier_source': The attribute 'default_identifier_source' is not allowed.`

mvdbeek:
That's the correct usage, but it's a new feature -- maybe you need a more recent planemo ?
0.46.1 fails with 
```
.. ERROR: Invalid tmpFP4UnE found. Errors [/var/folders/df/6xqpqpcd7h73b6jpx9t6cwhw0000gn/T/tmpFP4UnE:14:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'data', attribute 'default_identifier_source': The attribute 'default_identifier_source' is not allowed.
/var/folders/df/6xqpqpcd7h73b6jpx9t6cwhw0000gn/T/tmpFP4UnE:15:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'data', attribute 'default_identifier_source': The attribute 'default_identifier_source' is not allowed.]
```
0.48.0 passes.

mblue9:
I get that planemo version 0.46.1 when I install through conda (as described here: https://github.com/galaxyproject/planemo/issues/752#issuecomment-370204966) 
I've got conda v4.3.24. Should I be using a newer conda to get that planemo 0.48.0 version? or can I update planemo to 0.48.0 through conda?

nsoranzo:
@mblue9 Planemo recipe in bioconda needs to be update to 0.48.0

mvdbeek:
In the spirit of keeping things simple and atomic I'm going to merge this. We can add the `default_identifier_source` in another PR!

Issue #1733, opened 2018-02-21
BWA does not properly report errors


blankenberg:
if e.g. ```bwa sampe``` command fails with error ```[fread] Unexpected end of file``` the error code is not able to be used, because its output is being piped into ```samtools sort```. Since a header is being written by ```bwa sampe```, the  ```samtools sort``` command completes successfully and the output dataset is an OK green header-only bam file. 

This output should be in the error state.

This is a generalizable problem with using pipes (```|```) within tools, as return codes are not propagated down the line, but can be found by examining e.g. ```$PIPESTATUS```.

An easy workaround is to add a special catch for ```[fread] Unexpected end of file``` to the growing lists of error ```<regex match>``` strings in the tool, but this is advisable nor sustainable, as it becomes a 'game' of trying to find all of the errors, instead of properly relying on the command's exit codes. 

example command:
```
ln -s '/export/galaxy-central/database/files/024/dataset_24239.dat' 'localref.fa' && bwa index 'localref.fa' &&                            bwa aln -t "${GALAXY_SLOTS:-1}"   'localref.fa' '/export/galaxy-central/database/files/024/dataset_24063.dat' > first.sai &&  bwa aln -t "${GALAXY_SLOTS:-1}"   'localref.fa' '/export/galaxy-central/database/files/024/dataset_24064.dat' > second.sai &&  bwa sampe    -r '@RG\tID:str8'  'localref.fa' first.sai second.sai '/export/galaxy-central/database/files/024/dataset_24063.dat' '/export/galaxy-central/database/files/024/dataset_24064.dat'    | samtools sort -@${GALAXY_SLOTS:-2} -O bam -o '/export/galaxy-central/database/files/024/dataset_24253.dat'
```

example tool stderr:
```
[bwa_index] Pack FASTA... 0.06 sec
[bwa_index] Construct BWT for the packed sequence...
[bwa_index] 4.03 seconds elapse.
[bwa_index] Update BWT... 0.07 sec
[bwa_index] Pack forward-only FASTA... 0.05 sec
[bwa_index] Construct SA from BWT and Occ... 1.77 sec
[main] Version: 0.7.17-r1188
[main] CMD: bwa index localref.fa
[main] Real time: 9.673 sec; CPU: 5.970 sec
[bwa_aln] 17bp reads: max_diff = 2
[bwa_aln] 38bp reads: max_diff = 3
[bwa_aln] 64bp reads: max_diff = 4
[...]

PR #1748, opened 2018-02-28
Give samtools sort information on available memory

bernt-matthias:
Problem: for huge alignments creates a large number of temporary files and is slow because of disk IO. 

The -m parameter gives `samtools sort` the information on how many memory it can use (default is 768M in all samtools versions) . 

I also changed the number of cores to 1 if this was different. 

I would also be willing to implement an alternative for the use of the pipe in these tools (https://github.com/galaxyproject/tools-iuc/issues/1733) .. but a consensus on the "best" solution would be needed.

### DIFF
--- a/tool_collections/samtools/sam_to_bam/sam_to_bam.xml
+++ b/tool_collections/samtools/sam_to_bam/sam_to_bam.xml
@@ -25,7 +25,7 @@
         samtools sort
             -O bam
             -@ \${GALAXY_SLOTS:-1}
-            -m ${GALAXY_MEMORY_MB:-768}M
+            -m \${GALAXY_MEMORY_MB:-768}M
             -o '$output1'
             -T temp
     ]]></command>
--- a/tool_collections/samtools/sam_to_bam/sam_to_bam.xml
+++ b/tool_collections/samtools/sam_to_bam/sam_to_bam.xml
@@ -25,6 +25,7 @@
         samtools sort
             -O bam
             -@ \${GALAXY_SLOTS:-1}
+            -m ${GALAXY_MEMORY_MB:-768}M
             -o '$output1'
             -T temp
     ]]></command>
--- a/tools/htseq_count/htseq-count.xml
+++ b/tools/htseq_count/htseq-count.xml
@@ -32,9 +32,9 @@
     #end if
 
     #if $samfile.extension == 'bam':
-        samtools sort -n --output-fmt=SAM -o '$name_sorted_alignment_filename' '$samfile' &&
+        samtools sort -@ \${GALAXY_SLOTS:-1} -m ${GALAXY_MEMORY_MB:-768}M -n --output-fmt=SAM -o '$name_sorted_alignment_filename' '$samfile' &&
     #else
-        samtools view -Su -t '${reference_fasta_filename}.fai' '$samfile' | samtools sort -n --output-fmt=SAM -o '$name_sorted_alignment_filename' - &&
+        samtools view -Su -t '${reference_fasta_filename}.fai' '$samfile' | samtools sort -@ \${GALAXY_SLOTS:-1} -m ${GALAXY_MEMORY_MB:-768}M -n --output-fmt=SAM -o '$name_sorted_alignment_filename' - &&
     #end if
 
     htseq-count
@@ -66,7 +66,7 @@
             && samtools view -Su
                 -t '${reference_fasta_filename}.fai'
                 '$__new_file_path__/${samoutfile.id}_tmp.sam'
-            | samtools sort -o '$samoutfile' -
+            | samtools sort -@ \${GALAXY_SLOTS:-1} -m ${GALAXY_MEMORY_MB:-768}M -o '$samoutfile' -
         #end if
     #end if
     ]]>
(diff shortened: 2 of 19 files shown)


mvdbeek:
That's a good change in principle, but above a certain threshold adding more core to samtools sort likely reduces the sorting speed and increases the memory usage (since the value for -t is per thread).
Could we limit this to maximum 4 cores or something like that ?

bernt-matthias:
Does somebody know if GALAXY_MEMORY_MB is meant to give the per core memory or overall?

mvdbeek:
It's overall.

bernt-matthias:
Ah OK. Good to know for my recent changes in https://github.com/galaxyproject/galaxy/pull/4857.

mvdbeek:
The failure is because of ongoing work on planemo/galaxy-lib.

bernt-matthias:
Limiting the number of cores does not make sense in my opinion because the job has been given a certain number of cores. If we just don't use them all then the unused ones are wasted completely, but if samtools can not use them efficiently then they are just not used optimally. 

I guess it is the task of the admin to start the jobs in an appropriate queue.

mvdbeek:
I don't agree, the aligner is going to spawn 32 threads if galaxy slots is 32, and samtools is going to spawn another 32 threads at the same time. In strict environments your job is going to be killed. Ideally the slots would be divided in a fixed ratio between aligner and samtools.

mvdbeek:
Maybe the long term fix for this problem is in tighter integration with datatype converters, so that we could pipe in converters, which would allow specifying exactly how many cores to use for the sorting step (if sorting is even required).

bernt-matthias:
> Does somebody know if GALAXY_MEMORY_MB is meant to give the per core memory or overall?

Then the question is if samtools expects a per core / overall value. 

> Maybe the long term fix for this problem is in tighter integration with datatype converters, so that we could pipe in converters, which would allow specifying exactly how many cores to use for the sorting step (if sorting is even required).

You are right. Hmm. 

- Then we should limit the number of threads such that the producer and the consumer of the pipe are "balanced". I guess this will not be possible. 
- Or use a temporary file and call the two tasks successively (this might be better in terms of not using to many threads).

Would also be a clean solution for https://github.com/galaxyproject/galaxy/pull/4857.

mvdbeek:
> Then the question is if samtools expects a per core / overall value.

it's per thread, from `samtool sort`:
```
-m INT     Set maximum memory per thread; suffix K/M/G recognized [768M]
```

> Or use a temporary file and call the two tasks successively (this might be better in terms of not using to many threads).

I was thinking named pipes could work, but anyway this is probably something that is far down the road.

tool_collections/samtools/sam_to_bam/sam_to_bam.xml
```diff
@@ -25,6 +25,7 @@
         samtools sort
             -O bam
             -@ \${GALAXY_SLOTS:-1}
+            -m ${GALAXY_MEMORY_MB:-768}M
```
nsoranzo: You need to add a `\` before `${GALAXY_MEMORY_MB:-768}` (like for GALAXY_SLOTS). Similarly for all other occurrences below.

bernt-matthias:
I decided for using a temporary file between hisat2 and samtools. Seemed to be the cleanest solution (no use of pipes and both tools can use all cores & memory).  

Furthermore I used GALAXY_MEMORY_MB_PERSLOT which might be introduced via https://github.com/galaxyproject/galaxy/pull/5625 .. ?

tools/htseq_count/htseq-count.xml
```diff
@@ -66,7 +66,7 @@
             && samtools view -Su
                 -t '${reference_fasta_filename}.fai'
                 '$__new_file_path__/${samoutfile.id}_tmp.sam'
-            | samtools sort -o '$samoutfile' -
+            | samtools sort -@ \${GALAXY_SLOTS:-1} -m ${GALAXY_MEMORY_MB:-768}M -o '$samoutfile' -
```
mvdbeek: s/${GALAXY_MEMORY_MB/\${GALAXY_MEMORY_MB/

bernt-matthias: fixed

bernt-matthias:
Don't merge as it is at the moment: https://github.com/samtools/samtools/issues/831 The tools will fail in cluster environments with strict memory checking. 

Realized this while working on https://github.com/galaxyproject/tools-iuc/pull/1801.

bernt-matthias:
In the light of https://github.com/samtools/samtools/issues/831 the question now is if we should try to give samtools sort as much memory as possible (e.g. a fraction of GALAXY_MEMORY_MB) in order to minimize the number of temporary files. Or if we should use a reasonable fixed number (or the default) which results is a reasonable number of temp files. 

In the latter case the PR would be obsolete (I guess this would be the best). 

But due to my recent work on https://github.com/galaxyproject/tools-iuc/pull/1801 I would volunteer to 
- update samtools in all tools to the latest version 
- fix a bug in the number of threads (the parameter gives actually the number of additional threads, i.e. it uses one thread more than it should).

PR #1750, opened 2018-03-01
renamed extract genomic dna tool

bernt-matthias:
The name was identical to the legacy tool. This has the side effect that the galaxy environment is preferred over the conda environment (because of preserve_python_environment = legacy_only in galaxy.in). 

(You won't believe how long I needed to find this out ...)

### DIFF
--- a/tools/extract_genomic_dna/extract_genomic_dna.xml
+++ b/tools/extract_genomic_dna/extract_genomic_dna.xml
@@ -1,4 +1,4 @@
-<tool id="Extract genomic DNA 1" name="Extract Genomic DNA" version="3.0.3">
+<tool id="Extract genomic DNA" name="Extract Genomic DNA" version="3.0.3">
     <description>using coordinates from assembled/unassembled genomes</description>
     <requirements>
         <requirement type="package" version="0.7.1">bx-python</requirement>


bernt-matthias:
Alternativ PR: https://github.com/galaxyproject/galaxy/pull/5618 .. but I also don't care if both are accepted :)

mvdbeek:
If the version has changed it would be better to limit the upper version in Galaxy when selecting tools that need Galaxy's environment, so that we keep the tool lineage intact.

bernt-matthias:
So, this is covered in your PR? Or is this something I need to configure?

bernt-matthias:
Little note that might become relevant later. The problem was: `class GFFReaderWrapper(Iterator, NiceReaderWrapper)` and one of the subclasses of `NiceReaderWrapper` was also derived from `Iterator` in bx-python==0.8.1 which caused an Exception. In the version in the conda package this is not the case and works fine. 

This will become relevant when the requirement will be updated.

mvdbeek:
Yeah, I think this PR shouldn't be needed once https://github.com/galaxyproject/galaxy/pull/5626 is in and merged forward. The other problem you mentioned should be gone in galaxy's dev branch, I think. In any case there's a code comment in there that mentions this.

nsoranzo:
https://github.com/galaxyproject/galaxy/pull/5626 is in and merged forward, closing.

Issue #1754, opened 2018-03-03
FeatureCounts 1.6.0.2 has the first "built-in" Gene annotation file auto-selected. Would avoid usage errors to require active selection. 


jennaj:
**Tool**: featureCounts Measure gene expression in RNA-Seq experiments from SAM or BAM files. (Galaxy Version 1.6.0.2)

Users do not always notice when an option is pre-filled on a tool form.

I can select an input with the database metadata set as "mm10" but still have a form filled out with "hg38", the first index available, selected.

Enhancement ideas to avoid usage problems:

 - **Force users to actively choose a built-in annotation file**. Meaning, the default on the tool form would be blank or "unspecified (?)". 
 - **If nothing is chosen, do NOT submit the job**, but give a warning on the tool form that a selection must be made (built-in or from the history).
 - We could **use the input database metadata to auto-select annotation, if the input database matches one of the built-in indexes**. Nice to have, but is a niche use-case if there are only 4 indexed annotations available. 
 - AND, **IF there is a mismatch between the input datatype assignment and the built-in annotation selected (not for history annotation), also do not submit the job**, instead, warn the user at submission that they need to correct the mismatch. Mapping tools always tag the BAM/SAM output with the genome/database used for mapping -- unless a custom genome is used, which would not use one of the built-in annotations anyway.

Many existing tools do one of the above. It avoids very common usage errors -often users do not even notice the problem until some downstream tool errors or produces odd results. This is frustrating after doing the heavy lifting of getting through many mapping jobs + intermediate tools that might not fail but produce odd but unnoticed results. All jobs have to be rerun from the start when the wrong built-in index is used.

We are considering doing the same for more tools (mapping tools as a priority https://github.com/galaxyproject/galaxy/issues/4499) that have still built-in indexes auto-selected. **These are passively and randomly "selected"** - what is used is just the first in the index list on that server. Instead, tools should require actively-selected built-in indexes/genomes at runtime or data mismatch problems result.

Discussion welcome - just trying to make things easier for our end-users. Alternative ideas/solutions are great.

bgruening:
On the tool side we could make use of more validators and check for the dbkey more often. But this requires that the user actively annotates all datasets with the dbkey. 
A fix on the Galaxy side could be to not select anything if the select box is filled from a location file.

ping @blankenberg @guerler

mblue9:
[...]

PR #1755, opened 2018-03-04
featureCounts fix - require active selection of built-in annotation

mblue9:
This PR is a start towards a fix for https://github.com/galaxyproject/tools-iuc/issues/1754 as reported by @jennaj

### DIFF
--- a/tools/featurecounts/featurecounts.xml
+++ b/tools/featurecounts/featurecounts.xml
@@ -6,21 +6,22 @@
 
     <version_command>featureCounts -v 2&gt;&amp;1 | grep .</version_command>
     <command detect_errors="exit_code"><![CDATA[
+
         ## Export fc path for its built-in annotation
+
         export FC_PATH=\$(command -v featureCounts | sed 's@/bin/featureCounts$@@') &&
 
         ## Check whether all alignments are from the same type (bam || sam)
         featureCounts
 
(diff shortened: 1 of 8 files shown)


mvdbeek:
I like this. The other option could have an additional validator (https://docs.galaxyproject.org/en/latest/dev/schema.html#id39), but those don't work very well in workflows.

tools/featurecounts/featurecounts.xml
```diff
@@ -134,8 +134,8 @@
                help="The input alignment file(s) where the gene expression has to be counted. The file can have a SAM or BAM format; but ALL files must be in the same format" />
         <conditional name="anno">
             <param name="anno_select" type="select" label="Gene annotation file">
+                <option value="gtf" selected="True">GTF file</option>
```
mvdbeek: We need a bump here, I think, since the defaults have changed.

mblue9:
Thanks @mvdbeek I've bumped the wrapper version and also made some more changes to try to make it more robust, and eliminate some of those opportunities for errors reported by @jennaj. I've:
- made the built-in annotation selection a data table
- added some validators
- added tests for the built-in annotation and gtfs

mvdbeek:
Very nice, thanks a lot @mblue9 !

Issue #1704, opened 2018-02-03
single end dataset collection as input of Trinity


drosofff:
Dear trinity maintainers,

I have doubts that https://github.com/galaxyproject/tools-iuc/blob/master/tools/trinity/trinity.xml#L53 is sufficient to handle a collection of single end read datasets as input of the trinity tool.

Can you help on this ?

abretaud:
I've just tested, it seems to work as expected. What exactly is your problem?

drosofff:
Hi Antony,
Sorry I did not see you reply.

The tool does not have a test for a collection of *single end* reads.
When I test this setting, it fails. I though that the  https://github.com/galaxyproject/tools-iuc/blob/master/tools/trinity/trinity.xml#L53 may explain this bug.

I can give you access to a fresh galaxy instance, but this is a GCE VM so I need to start it up before. Ask me if you want to have a look (but testing a single end read collection locally should in principle show you the same problem)

abretaud:
I don't see the problem, at https://github.com/galaxyproject/tools-iuc/blob/master/tools/trinity/trinity.xml#L54 we loop on all the single end datasets provided. As the param is `multiple="true"`, the user can select multiple datasets or collections and run trinity on it.

mvdbeek:
> multiple="true", the user can select multiple datasets or collections and run trinity on it.

`multiple="true"` is a reduction operation, so this only makes sense if the samples you select there are supposed to be merged and treated as one. It probably makes more sense to either remove `multiple="true"` or wrap the merging in a conditional.

abretaud:
> merged and treated as one

That's the intended beaviour, it is an assembler which takes all available fastq to produce a single result

mvdbeek:
I wouldn't call that expected, maybe for metagenomics it makes sense, but for a multi-sample analysis where you have one fastq per sample this is not what you'd expect. We've had that discussion a few times before and I think until the galaxy ui lets you specify the behaviour we should wrap this in aconditional, as we've done for macs2 recently (https://github.com/galaxyproject/tools-iuc/pull/1648)

abretaud:
You mean https://github.com/galaxyproject/tools-iuc/pull/1648/files#diff-77d32f534e7d91585bd3f97991dbd623R127 ?

For trinity I don't think it makes sense to add a conditional like "Are you pooling Treatment Files?", you always want to pool the input files

mvdbeek:
Yep, that's the one

[...]

PR #1756, opened 2018-03-04
Adapt trinity to process separately elements of single or paired-end collection

drosofff:
@abretaud @mvdbeek 
In response to the recent issue #1704, I modified the trinity.xml wrapper.

- 2 cases of collection inputs (single end and paired-end) are added, in which datasets are treated separately. This allow, for instance, to use the tool in situation of patient diagnostics (obviously merging patient sequence datasets is non sense).

- I added the corresponding tests. They could be more stringent but I am not super confortable with testing collection inputs/ outputs.

- I bumped the version token to 2.4.1

I hope you'll be ok to review this PR: I really need this functionality in trinity !


tools/trinity/macros.xml
```diff
@@ -7,7 +7,7 @@
         </requirements>
     </xml>
 
-    <token name="@WRAPPER_VERSION@">2.4.0</token>
+    <token name="@WRAPPER_VERSION@">2.4.1</token>
```
abretaud: This is supposed to be the trinity version, so you'd better leave it as 2.4.0 here, and bump the suffix in https://github.com/ARTbio/tools-iuc/blob/1bf6f1415275996ba281fc988d4a4434080761a3/tools/trinity/trinity.xml#L1

tools/trinity/trinity.xml
```diff
@@ -112,8 +138,10 @@
         <conditional name="inputs">
             <param name="paired_or_single" type="select" label="Paired or Single-end data?">
                 <option value="single">Single-end</option>
+                <option value="unmerged_single_collection">Unmerged Single-end collection</option>
```
abretaud: I don't know if there are some convention for this yet, but I'd write "Pooled single-end" on previous line and "Single-end" on this line

tools/trinity/trinity.xml
```diff
@@ -112,8 +138,10 @@
...
                 <option value="paired" selected="true">Paired-end</option>
                 <option value="paired_collection">Paired-end collection</option>
+                <option value="unmerged_paired_collection">Unmerged Paired-end collection</option>
```
abretaud: Here too, "Pooled paired-end collection" on previous line and "Paired-end collection" here

tools/trinity/trinity.xml
```diff
@@ -138,6 +180,10 @@
                 <param name="pair_input" type="data_collection" collection_type="list:paired" format="fasta,fastqsanger" label="FASTA/FASTQ dataset collection with R1/R2 pair" help="Can be lists of pair dataset collection"/>
                 <expand macro="input_paired_strand_jaccard" />
             </when>
+            <when value="unmerged_paired_collection">
+                <param name="pair_input" type="data_collection" collection_type="paired" format="fasta,fastqsanger" label="FASTA/FASTQ dataset collection with R1/R2 pair" help="Pair dataset collection. The paired datasets will NOT be merged"/>
```
abretaud: Is it intentional to use paired collection_type instead of list:paired as above?

mvdbeek: This is correct, I think, otherwise you're consuming the entire list, instead of running one job per list item.

tools/trinity/trinity.xml
mvdbeek: Hmm, this is actually mixing 3 concepts here
 - Single end vs Paired end
 - Collection vs Single dataset
 - Pooled or not

Could we implement this change as in
https://github.com/galaxyproject/tools-iuc/blob/master/tools/macs2/macs2_callpeak.xml#L130-L141 ?

This would result in a nested and repeated conditional, but I think this would make it easier to understand for users.

drosofff:
@abretaud @mvdbeek @jmchilton I hope these commits address the concern above. I am not sure that the tests are testing as accurately as we could do but I am not comfortable with simulating dataset collections in the xml wrapper

abretaud:
Sorry for the delay, LGTM, are you ok with it too @mvdbeek?

Issue #1751, opened 2018-03-01
Broken dependency links for package_python_2_7


luke-c-sargent:
The link in [this line](https://github.com/galaxyproject/tools-iuc/blob/427913231c5894bde2b2ed428da2b8aee30bc311/packages/package_python_2_7/tool_dependencies.xml#L56) of the `package_python_2_7` package is broken;
```https://bitbucket.org/pypa/setuptools/get/18.2.tar.bz2```
is no longer accessible, impacting the installation of tools that use this dependency.

**Comment by luke-c-sargent** [comment]
the github mirror:
```
https://github.com/pypa/setuptools/archive/18.2.tar.gz
```
seems workable; will put this in a PR shortly.

nsoranzo:
You'll need to update also the md5sum to 3a13d4dab5b5120c8e75866371745c82, the tar options (`s/j/z/`) on the next line, and the directory from `pypa-setuptools-1a981f2e5031` to `setuptools-18.2` .

luke-c-sargent:
do you have suggestions on how to test this? i tried running tests on the only tool that uses `package_python_2_7` as a direct dependency in the IUC suite, `progressivemauve`, but it installs setuptools before running tests.

i am using: `planemo test --install_galaxy progressivemauve.xml`

the tests fail, but it does not appear to be because of setuptools.

my fork's changes [can be seen here](https://github.com/luke-c-sargent/tools-iuc/blob/aab1f378a948b476964098f5cf7a061721c026a7/packages/package_python_2_7/tool_dependencies.xml#L56).

I used the sha256 as that was being used for the previous link, but am happy to switch it to md5 as you mentioned earlier (is it just `md5sum="..."` instead of `sha256sum="..."`?)

also, the tar options i went with are: `tar -xsf`

nsoranzo:
Unfortunately the only way to test a package is to upload it to a local ToolShed, configure a local Galaxy to use this ToolShed, and install it from this ToolShed.

You're right about keep using SHA256, which is a bit better than MD5, but the value still need to be updated (to `132d9b7a69facc1a7d78d2dfa2fb5a6e93f1250c5a320634d6061854d8729aaf` in this case).

Finally, the tar command should be `tar -xzf`, I used a cryptic regular expression above (`s/j/z/`) which means substitute the `j` with a `z`, sorry for the jargon.

luke-c-sargent:
whoops, i even computed the sha and didn't change the value :(

[...]

PR #1762, opened 2018-03-07
Setuptools url fix

luke-c-sargent:
fix per issue #1751


nsoranzo:
Thanks @luke-c-sargent!

PR #1765, opened 2018-03-10
Fastp - minor modifications

mblue9:
Now that I've tried fastp a bit I have some suggestions for modifications to the wrapper here:

- add sample id (element identifier) to report title
- group and reorder options similar to Cutadapt
- bump wrapper version (because of changing output - adding sample id above)

### DIFF
--- a/tools/fastp/fastp.xml
+++ b/tools/fastp/fastp.xml
@@ -15,13 +15,18 @@
 #else
     #set ext = 'fastq'
 #end if
-#set $filename = str($in1.element_identifier)
 
 ln -s '$in1' input1.$ext &&
 
 #if str($single_paired.single_paired_selector) == 'paired':
     ln -s '$in2' input2.$ext &&
 #end if
+
+
+## Set filename for output report
+
+#import re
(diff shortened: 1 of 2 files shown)


tools/fastp/fastp.xml
```diff
@@ -8,14 +8,14 @@
...
     <command detect_errors="exit_code"><![CDATA[
-
 ## Link input files
 
 #if $in1.is_of_type('fastq.gz')
     #set ext = 'fastq.gz'
 #else
     #set ext = 'fastq'
 #end if
+#set $filename = str($in1.element_identifier)
```
mvdbeek: Since this appears on the command line I think we'll have to escape this as in https://github.com/galaxyproject/tools-iuc/blob/master/tools/fastqc/rgFastQC.xml#L8

mblue9:
Thanks @mvdbeek and @bgruening! I've fixed the identifier and added the new options for v0.12.5

bgruening:
Cool, thanks @mblue9!

PR #1766, opened 2018-03-11
AnnotateMyIDs - add filter for duplicate gene ids in output

mblue9:
When our users try to e.g. add gene symbols to ids they can end up doing an ugly Excel dance to remove the duplicated gene ids from the output (ids with more than one symbol). This adds an option to annotateMyIDs to remove duplicate gene ids from the output (just keeps the first).


tools/annotatemyids/annotateMyIDs.xml
```diff
@@ -118,7 +125,16 @@ write.table(result, file='$out_tab', sep="\t", row.names=FALSE, quote=FALSE)
             <param name="id_type" value="ENSEMBL"/>
             <param name="organism" value="Hs"/>
             <param name="output_cols" value="ENSEMBL,GO,ONTOLOGY,EVIDENCE" />
-            <output name="out_tab" file="out_gokegg.tab" compare="contains" />
```
bgruening: Why have you changed this? It seems to cause one of the failing tests.

bgruening:
@mblue9 why not using the `unique` tool or the datamash tool to group on the gene-id column?

tools/annotatemyids/annotateMyIDs.xml
mblue9: I think the failing tests were due to me somehow accidentally using old versions of the org.db package's (3.4.1 instead of 3.5.0) and I downgraded the test data which then failed here. Think I've fixed it now.

mblue9:
@bgruening yes I could use another tool but I'd like to have an all-in-one option. The users are using this tool to get the annotation anyway so I'd like to make it easier for them and have an option within this tool. Good to know about the other ways though I haven't used those yet!

PR #1767, opened 2018-03-12
Wrap samtools fasta and fastq operations

davebx:
This also adds filtering options to macros.xml


tool_collections/samtools/samtools_fastx/samtools_fastx.xml
```diff
@@ -0,0 +1,102 @@
...
+    <command detect_errors="exit_code">
+    <![CDATA[
+        samtools $output.output_format
+            #if $output.output_format == 'fastq':
+                -v $output.default_quality $output.output_quality
+            #end if
+            #if $reference:
+                --reference $reference
+            #end if
+            $copy_flags
```
nsoranzo: `$omit_read_number` is missing here

tool_collections/samtools/samtools_fastx/samtools_fastx.xml
```diff
@@ -0,0 +1,102 @@
...
+        <test>
+            <param name="input" value="samtools_fastx-in3.sam" ftype="sam" />
+            <param name="output_format" value="fasta" />
+            <output name="nonspecific" file="samtools_fastx-out3-1.fasta" />
+            <output name="forward" file="samtools_fastx-out3-2.fasta" />
+            <output name="reverse" file="samtools_fastx-out3-3.fasta" />
+        </test>
+    </tests>
+    <help>
+        This tool uses SamTools to extract sequences from a SAM or BAM file in FASTA or FASTQ format.
```
nsoranzo: Maybe `Samtools` can be a link to http://www.htslib.org/ (also it seems the preferred spelling is with a lowercase `t` now).

tool_collections/samtools/samtools_fastx/samtools_fastx.xml
```diff
@@ -0,0 +1,102 @@
...
+            <param name="output_format" type="select" label="Output format">
+                <option value="fasta">FASTA</option>
+                <option value="fastq">FASTQ</option>
+            </param>
+            <when value="fastq">
+                <param name="default_quality" argument="-v" type="integer" value="1" label="Default quality if none is given" />
+                <param name="output_quality" argument="-O" type="boolean" truevalue="-O" falsevalue="" label="Output quality in the OQ tag if available" />
+            </when>
+            <when value="fasta" />
+        </conditional>
```
nsoranzo: Should the output options be at the end of the `<inputs>` section, as is our usual pattern?

bernt-matthias:
Just a note: 

There are a few bugs in the in the implementation of the flags (see https://github.com/samtools/hts-specs/blob/master/SAMv1.pdf) 
- 768 (The read is a PCR or optical duplicate) should be 1024
- 1024 (Supplementary alignment) should be 2048

I will fix this in https://github.com/galaxyproject/tools-iuc/pull/1801 ..

Issue #1770, opened 2018-03-15
Trinotate: Pre-generated Resource SQLite database not found


loraine-gueguen:
The download of the Trinotate Pre-generated Resource SQLite database fails because it has moved (now here: https://data.broadinstitute.org/Trinity/__deprecated_trinotate_resources/Trinotate_v3_RESOURCES/Trinotate_v3.sqlite.gz)

A quick fix is to update the URL in trinotate.xml.

But it seems that the sqlite database can now be build using the trinotate tool Build_Trinotate_Boilerplate_SQLite_db.pl (see https://trinotate.github.io/#SequencesRequired). I haven't tried, I am not an expert of Trinotate, but if this tool is ok, should be changed when the trinotate package version will be updated in the requirements.

PR #1771, opened 2018-03-15
Trinotate: fix sqlite database url

loraine-gueguen:
The quick fix for https://github.com/galaxyproject/tools-iuc/issues/1770

### DIFF
--- a/tools/trinotate/trinotate.xml
+++ b/tools/trinotate/trinotate.xml
@@ -1,4 +1,4 @@
-<tool id="trinotate" name="Trinotate" version="3.0.1.0">
+<tool id="trinotate" name="Trinotate" version="3.0.1.1">
     <requirements>
         <requirement type="package" version="3.0.1">trinotate</requirement>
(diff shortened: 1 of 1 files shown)


tools/trinotate/trinotate.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="trinotate" name="Trinotate" version="3.0.1.0">
+<tool id="trinotate" name="Trinotate" version="3.0.1.1">
```
nsoranzo: No need to update the tool id for a bug fix that doesn't change the tool behaviour.

loraine-gueguen: ok. Done

bgruening:
@loraine-gueguen maybe we should cache this file ourself in https://github.com/galaxyproject/cargo-port ?

bgruening:
Thanks for the bugfix!

loraine-gueguen:
@bgruening I would make sense. I will have a look. Thanks

PR #1772, opened 2018-03-15
Add data_manager_meme_motif_databases_downloader

gregvonkuster:


gregvonkuster:
ping @bgruening

data_managers/data_manager_meme_motif_database_downloader/data_manager/data_manager_meme_motif_databases_downloader.xml
```diff
@@ -0,0 +1,37 @@
+<tool id="data_manager_meme_motif_databases_downloader" name="MEME Motif Databases Downloader" version="1.1.0" tool_type="manage_data">
+    <description></description>
+    <command detect_errors="exit_code"><![CDATA[
+python $__tool_directory__/data_manager_meme_motif_databases_downloader.py
```
bgruening: can you please also quote the file+tool-dir?

data_managers/data_manager_meme_motif_database_downloader/data_manager/data_manager_meme_motif_databases_downloader.xml
```diff
@@ -0,0 +1,37 @@
...
+--name '$name.replace(' ', '_')'
```
bgruening: If you sanitize the text input and don't allow spaces, you don't need to replace it here and I guess it's more save.

data_managers/data_manager_meme_motif_database_downloader/data_manager/data_manager_meme_motif_databases_downloader.py
```diff
@@ -0,0 +1,124 @@
...
+        src = urllib2.urlopen(req)
+        dst = open(file_path, 'wb')
+        while True:
+            chunk = src.read(2**10)
+            if chunk:
+                dst.write(chunk)
+            else:
+                break
+    except Exception, e:
+        print >>sys.stderr, str(e)
```
bgruening: Can you convert this to py3k as well? I guess we should not use py2k specific code anymore.

gregvonkuster:
This data table is needed only if the Centrimo feature is needed for the meme-chip tool - see https://github.com/galaxyproject/tools-iuc/pull/1773/files#diff-288fc08f13ea171da15363728a91a6f6R19.  

To see how this works with mem-ChIP - try the tool here http://alternate.meme-suite.org/tools/meme-chip, selecting an option from the first select list under the label ```Input the motifs``` and watch the options change in the 2nd select list.

But using Centrimo requires a fix to the Galaxy dynamic options support for rendering multiple dynamic select lists where selecting an option in the first list re-renders the [changed] options in the second list.  
This used to work since I used this feature years ago for this tool https://github.com/gregvonkuster/galaxy-csg/blob/master/tools/icqsol_color_surface_field/icqsol_color_surface_field.xml#L32, and I contributed the section titled "Examples -> Dynamic Options" here https://docs.galaxyproject.org/en/latest/dev/schema.html which describes the feature.  

But I have no idea when it was broken in the Galaxy framework.  Obviously not many tools use this feature, and since meme-chip is useful even without the Centrimo features, I'm wondering how to proceed here.  Should I open an issue to get the dynamic options fixed for handling multiple associated dynamic select lists like this, or should I just close this PR?  

Ideally, I would like this feature supported for both the meme-chip tool and the CSG tool linked above, but I don't believe the CSG tool is being used at the current time, so I'm not pushing for the framework fix at the current time.  Also, if I should open an issue to get the framework fixed, I won't have the time to work on a fix myself, at least for a while.

PR #1744, opened 2018-02-26
Move cdhit to IUC

jj-umn:
Defined as tool suite.
Using conda cd-hit version 4.6.8

PR #1778, opened 2018-03-19
joint cd-hit[-est][-2d] wrapper

bernt-matthias:
Follow up to: https://github.com/galaxyproject/tools-iuc/pull/1744

This wrapper unites the cd-hit, cd-hit-est wrappers of @jj-umn  and @bebatut  and adds cd-hit-2d and cd-hit-est-2d. Since program calls and options largely overlap I suggest to solve it this way instead of macros and 4 xml files .. but I could change this easily.

For comparison I left the two old xml files. These need to be deleted before merging.

Some TODOs are in the xml where I could need your input.

- [x] delete old wrappers


tools/cdhit/cd-hit.xml
```diff
@@ -0,0 +1,349 @@
+<!-- TODO are special characters in the name a problem? -->
+<tool id="cd_hit" name="cd-hit[-est][-2d]" version="4.6.8.1">
```
bgruening: Most likely not a problem, but I think it is confusing. I guess cd-hit is just good enough.

tools/cdhit/cd-hit.xml
```diff
@@ -0,0 +1,349 @@
...
+    #end if
+    #if $print_alnovl.print_alnovl_select == "yes":
+      -p 1
+      #if str($print_alnovl.desclen) != '':
+        -d $print_alnovl.desclen
+      #end if
+    #end if
+
+    ## TODO instead of 800 (default) we could also use 0:unlimited?
+    -M \${GALAXY_MEMORY_MB:-800}
```
bgruening: I'm more a friend of unlimited by default.

tools/cdhit/cd-hit.xml
```diff
@@ -0,0 +1,349 @@
...
+        <when value="-2d">
+          <param name="fasta_in2" type="data" format="fasta" label="Other sequences to cluster/compare" help="(-i2)"/>
+          <conditional name="advancedtwod">
+            <param name="advancedtwod_select" type="select" label="Use advanced settings for est?">
+              <option value="no" selected="true">No</option>
+              <option value="yes">Yes</option>
+            </param>
+            <when value="no"/>
+            <when value="yes">
+             <!-- TODO these parameters work only if local alignment is chosen in the advanced options. We could also move them there. Or is there a way to combine the results of multiple selects -->
```
bgruening: Moving to the advanced sections seems to be the only way for the moment.

tools/cdhit/cd-hit.xml
```diff
@@ -0,0 +1,349 @@
...
+              <param name="match" type="integer" value="2" optional="true" label="match score" help="match score, default 2 (1 for T-U and N-N) (-match)"/>
+              <param name="mismatch" type="integer" value="-2" optional="true" label="mismatch score" help="mismatch score, default -2 (-mismatch)"/>
+              <param name="gap" type="integer" value="-6" optional="true" label="gap opening score" help="gap opening score, default -6 (-gap)"/>
+              <param name="gapext" type="integer" value="-1" optional="true" label="gap extension score" help="gap extension score, default -1 (-gap-ext)"/>
+            </when>
+          </conditional>
+        </when>
+      </conditional>
+      <!-- TODO org tools had 0.4-1.0 (protein) and 0.75-1.0 (est) can/should we implement a validator that distinguishes the choice, or put this element in the two when branches above with different limits -->
+      <param name="similarity" type="float" value="0.9"  label="Sequence identity threshold" min="0.4" max="1.0" help="global sequence identity: number of identical alignment positions divided by the full length of the shorter sequence (-c)"/>
```
bgruening: Moving this parameter seems good to me, or we offer a select with multiple defaults (default for EST, default for Protein) and  a `set manual option`.

bgruening:
Thanks @bernt-matthias!

tools/cdhit/cd-hit.xml
bernt-matthias: OK

tools/cdhit/cd-hit.xml
bernt-matthias: Fine

tools/cdhit/cd-hit.xml
bernt-matthias: OK

bernt-matthias:
OK did the fixes. 

Any comments to the last comments. 

Furthermore it seems that local mode implies that two parameters need to be larger that 0 (https://github.com/weizhongli/cdhit/issues/57) .. but 0 is the default :) . Is there a way to check (validator) if a value is larger than 0. With min=0 one can only check for larger or equal...

tools/cdhit/cd-hit.xml
```diff
@@ -0,0 +1,344 @@
...
+      <param name="similarity" type="float" value="0.9"  label="Sequence identity threshold" min="0.4" max="1.0" help="global sequence identity: number of identical alignment positions divided by the full length of the shorter sequence. For nucleotide sequences [0.75,1.0] and for proteins [0.4:1.0] (-c)"/>
+
+      <conditional name="advanced">
+        <param name="advanced_select" type="select" label="Use advanced settings?">
+          <option value="no" selected="true">No</option>
+          <option value="yes">Yes</option>
+        </param>
+        <when value="no"/>
+        <when value="yes">
+          <!-- TODO most parameters in the original version had a validator which is replaced here by min/max. OK? -->
```
bgruening: yes, perfect!

tools/cdhit/cd-hit.xml
```diff
@@ -0,0 +1,344 @@
...
+              <param name="align_coverage_short" type="float" value="0.0" optional="true" label="Alignment coverage for the shorter sequence" min="0.0" max="1.0" help="Fraction of the shorter sequence that must be covered (-aS)"/>
+              <param name="align_coverage_short_control" type="integer" value="99999999" optional="true" label="Alignment coverage control for the shorter sequence" min="0" help="Maximum number of residues uncovered residues of the shorter sequence (-AS)"/>
+              <param name="align_coverage_min" type="integer" value="0" optional="true" label="Minimal alignment coverage control for the both sequences" min="0" help="Minimum number of residues of both sequences that must be covered (-A)"/>
+              <!-- TODO many parameters had no default value set, but had value="" and optional="true". I have now chosen defaults as given on the web site. Does this make sense? The downside is that now all options in a conditional are set once it is chosen. Then I would remove the optional="" and the if statements in the command call. Otherwise I would remove the value="..." and add the default value to the help.-->
```
bgruening: Do you know what happens when you don't set the param on the commandline, will then the default value be taken?
If so I would leave if optional with no default value set.

bgruening:
> Furthermore it seems that local mode implies that two parameters need to be larger that 0 (weizhongli/cdhit#57) .. but 0 is the default :) . Is there a way to check (validator) if a value is larger than 0? With min=0 one can only check for larger or equal...

I do not think this is possible, but not sure. (@blankenberg )
I would make the optional=false and no default for the moment

tools/cdhit/cd-hit.xml
bernt-matthias: Just tried, in local mode the program fails with default options, i.e. cd-hit uses 0 which fails.

bernt-matthias: So I just added a little note to the select where the local mode is chosen.

bernt-matthias:
I decided to just remove the default for these three options, since a value of 0 for one or two of them (i.e. the default) is OK. 

Just 1 TODO left .. somehow connected.

bernt-matthias:
I will now install the tool in our local instance and let the user who requested the tool test it.

bernt-matthias:
One question. The conda package seems to compile with OMP if possible. If OMP is not available then the -T switch will not work. Should we deal with this case or just assume that OMP is available?

bgruening:
Ideally we should consider all different options, but I don't think this is feasible. Its a sane choice to assume that it is compiled with OMP .

bernt-matthias:
OK

tools/cdhit/cd-hit.xml
```diff
@@ -0,0 +1,344 @@
...
+          <option value="" selected="true">Cluster sequences</option>
+          <option value="-2d">Compare with 2nd sequence data set</option>
+        </param>
+        <when value=""/>
+        <when value="-2d">
+          <param name="fasta_in2" type="data" format="fasta" label="Other sequences to cluster/compare" help="(-i2)"/>
+        </when>
+      </conditional>
+      <conditional name="est">
+        <param name="est_select" type="select" label="Sequence type? For nucleotid the -est variant of cd-hit is called">
```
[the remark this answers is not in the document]
bernt-matthias: done

tools/cdhit/cd-hit.xml
```diff
@@ -0,0 +1,344 @@
...
+          <when value="">
+          <param name="wordsize" type="integer" value="5" label="Word size" min="2" max="5" help="">
+            <help>Suggested word size:
+            5 for thresholds 0.7 ~ 1.0
+            4 for thresholds 0.6 ~ 0.7
+            3 for thresholds 0.5 ~ 0.6
+            2 for thresholds 0.4 ~ 0.5 (-n)
+            </help>
+          </param>
+          <param name="redtol" type="integer" value="2" optional="true" label="tolerance for redundance" help="(-t)"/>
```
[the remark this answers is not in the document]
bernt-matthias: done .. all over the xml :)

tools/cdhit/cd-hit.xml
```diff
@@ -0,0 +1,344 @@
...
+        </when>
+      </conditional>
+    </inputs>
+
+    <outputs>
+      <data format="txt" name="clusters_out" label="${tool.name} on ${on_string}: Clusters" from_work_dir="rep_seq.clstr"/>
+      <data format="fasta" name="fasta_out" label="${tool.name} on ${on_string}: Representatives sequences" from_work_dir="rep_seq"/>
+    </outputs>
+
+    <tests>
```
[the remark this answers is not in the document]
bernt-matthias: OK. You add the tests (seems that you have some ideas) and I fix problems that are uncovered. Deal? :)

mvdbeek:
This is ready, right ?

mvdbeek:
Agreed, about the additional tests. The other wrappers seem to never have been upload to the tool shed (oops), so that's OK.

nsoranzo:
> The other wrappers seem to never have been upload to the tool shed (oops), so that's OK.

@mvdbeek I think the reason is that this PR was open against the `cdhit` branch, i.e. this has now been merged in PR https://github.com/galaxyproject/tools-iuc/pull/1744 .

PR #1781, opened 2018-03-20
mothur: fix filter

shiltemann:
@nsoranzo this should fix the failing test you mentioned on gitter


nsoranzo:
Thanks @shiltemann!

PR #1477, opened 2017-09-14
fix typo

ieguinoa:
Just noticed that the output files number was wrong.

PR #1782, opened 2018-03-20
MACS2 -fix output naming for collections when using Control

mblue9:
Currently if you use MACS2 with collections and have a control sample, the control sample name gets used as the output identifier (as mentioned here: https://github.com/galaxyproject/tools-iuc/pull/1747#issuecomment-369484476)

@mvdbeek I've tried using default_identifier_source with planemo 0.48.0 and it passes linting now, but I'm still getting the control file name used 😕. What I tried is in this PR, am I doing something wrong?


mvdbeek:
I'll have a look. Did you test this with 18.01 ?

mblue9:
In case it helps to know, I'm testing it using a couple of the macs2 test-data files like so (where "treat" is a collection containing the ChIP bed file and "input" is a collection containing the Control bed file):

screen shot 2018-03-21 at 8 01 50 am

mblue9:
@mvdbeek that was fast I didn't see your reply! Yes 18.01

mvdbeek:
Hmm, it looks correct to me:
screen shot 2018-03-20 at 22 28 01

mvdbeek:
If it wasn't working we'd get 'Control 200K.bed' as the element identifier, right ?

mblue9:
Yep, that's what I'm getting

mblue9:
oops also missed your screenshot above! that is weird. Are you also using planemo 0.48.0? I'm on 18.01 `a73a1d6`

mvdbeek:
Yes, `planemo, version 0.48.0`. The `default_identifier_source` was introduced in 17.09, so that should be fine. How are you testing this, just a `planemo serve` ? Any chance that would pick up an old release ?

mvdbeek:
You could hit <galaxy_url>/api/version, just to be sure nothing weird is happening there.

mvdbeek:
I'll do the reverse and revert your changes, then I guess I should be seeing the wrong result.

mvdbeek:
Yep, without your changes in this PR:
screen shot 2018-03-20 at 22 45 51

mvdbeek:
Humm, switching back I now see the wrong behavior as well. Uff, that's probably a galaxy bug then :(.

mblue9:
Ok good to know it's not just me :smile: 

Fwiw I was testing it with `planemo s . --galaxy_root /path/to/local/galaxy`

and if I do `<galaxy_url>/api/version` I get:

// 20180321085037
// http://127.0.0.1:9090/api/version

{
  "extra": {
    
  },
  "version_major": "18.01"
}

mblue9:
Seeing as I was in the macs2 xmls I thought I may aswell:
 - update the Effective Genome Size info 
 - add @ieguinoa's  typo fix for macs2_bdgdiff.xml from here https://github.com/galaxyproject/tools-iuc/pull/1477 (and fix the resulting failing test)

bgruening:
@mvdbeek is this one ready to go? Should it be pinned to >17.09?

mvdbeek:
Pinning to 17.09 should be enough.

mvdbeek:
Thanks @mblue9 and @bgruening, I've assigned myself https://github.com/galaxyproject/galaxy/issues/5756. This really was my motivation to implement `default_idenitifier_source`, so I'll dig into what is going on there.

PR #1784, opened 2018-03-21
Use version compatible with packaging.version

davebx:
Unfortunately the tools appending .n to the version string, in this case, causes the tool panel code to parse the version as `packaging.version.LegacyVersion`, which always sorts as older than a PEP-440 compliant `packaging.version.Version`

This change will make the correct version appear in the tool panel when a 0.0.x version has also been installed. I have confirmed that `packaging.version.parse('1.0.0_rc1') < packaging.version.parse('1.0.0.1')` evaluates as `True`


martenson:
The question is "How many non pep-440 versions do we have in TS?".

tool_collections/vcflib/macros.xml
```diff
@@ -10,7 +10,7 @@
             <exit_code range="1:" level="fatal" />
         </stdio>
     </xml>
-   <token name="@WRAPPER_VERSION@">1.0.0_rc1</token>
+   <token name="@WRAPPER_VERSION@">1.0.0.1</token>
```
nsoranzo: I think @mvdbeek suggested to just use `1.0.0`

martenson:
From discussion with @bgruening an alternative is to use `<tool id="vcfbedintersect" name="VCF-BEDintersect:" version="@WRAPPER_VERSION@-0">` (dash instead of dot) which makes it PEP440 I think and preserves the original version of the tool's author (vcflib in this case).
```
>>> packaging.version.parse('1.0.0_rc1-0')
<Version('1.0.0rc1.post0')>
```

peterjc:
Suggest holding off merging this until https://github.com/galaxy-iuc/standards/issues/51 is agreed,  e.g.

``<tool id="vcfbedintersect" name="VCF-BEDintersect:" version="@TOOL_VERSION+galaxy@WRAPPER_VERSION@">``

tool_collections/vcflib/vcf2tsv/vcf2tsv.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="vcf2tsv" name="VCFtoTab-delimited:" version="@WRAPPER_VERSION@-0">
+<tool id="vcf2tsv" name="VCFtoTab-delimited:" version="@WRAPPER_VERSION+galaxy0">
```
martenson: I think this is invalid macro token. Also you might want to rebase since vcffilter has been updated: https://github.com/galaxyproject/tools-iuc/pull/1791

@mvdbeek chose `+galaxy1` there, so this should follow imo.

mvdbeek:
@davebx you'll have to rebase against master because of #1791

PR #1786, opened 2018-03-22
MAGeCK - improvements

mblue9:
After being notified that I should escape the element_identifier I added to fastp (here: https://github.com/galaxyproject/tools-iuc/pull/1765#discussion_r175140809, thanks @mvdbeek!), I realised I did not escape the element_identifier I added to mageck count. So have fixed that in this PR.


mvdbeek:
Thanks @mblue9! I guess we need a test data update (https://travis-ci.org/galaxyproject/tools-iuc/jobs/356738090#L1679)

mblue9:
@mvdbeek this passes locally but here it's failing with a pdflatex error: https://travis-ci.org/galaxyproject/tools-iuc/jobs/356738090#L1584. So not sure what to do?

Here with Travis it gives:

```
INFO  @ Thu, 22 Mar 2018 07:03:11:    1 : keep.source term verbatim (label = funcdef, output_countsummary.Rnw:28) 
INFO  @ Thu, 22 Mar 2018 07:03:11:    2 : keep.source term tex (label = tab1, output_countsummary.Rnw:156) 
INFO  @ Thu, 22 Mar 2018 07:03:11:    3 : keep.source term tex (label = tab2, output_countsummary.Rnw:174) 
INFO  @ Thu, 22 Mar 2018 07:03:11:    4 : keep.source term tex (label = tab3, output_countsummary.Rnw:188) 
INFO  @ Thu, 22 Mar 2018 07:03:11:    5 : keep.source term verbatim pdf  (output_countsummary.Rnw:221) 
INFO  @ Thu, 22 Mar 2018 07:03:11:    6 : keep.source term verbatim pdf  (output_countsummary.Rnw:228) 
INFO  @ Thu, 22 Mar 2018 07:03:11:    
INFO  @ Thu, 22 Mar 2018 07:03:11:   You can now run (pdf)latex on ‘output_countsummary.tex’ 
INFO  @ Thu, 22 Mar 2018 07:03:11:   Error in texi2dvi("output_countsummary.tex", pdf = TRUE) :  
INFO  @ Thu, 22 Mar 2018 07:03:11:     pdflatex is not available 
INFO  @ Thu, 22 Mar 2018 07:03:11:   Execution halted 
INFO  @ Thu, 22 Mar 2018 07:03:11:    
INFO  @ Thu, 22 Mar 2018 07:03:11: End command message. 
```

Locally I get:

```
INFO  @ Thu, 22 Mar 2018 16:49:45:    1 : keep.source term verbatim (label = funcdef, output_countsummary.Rnw:28) 
INFO  @ Thu, 22 Mar 2018 16:49:45:    2 : keep.source term tex (label = tab1, output_countsummary.Rnw:156) 
INFO  @ Thu, 22 Mar 2018 16:49:45:    3 : keep.source term tex (label = tab2, output_countsummary.Rnw:174) 
INFO  @ Thu, 22 Mar 2018 16:49:45:    4 : keep.source term tex (label = tab3, output_countsummary.Rnw:188) 
INFO  @ Thu, 22 Mar 2018 16:49:45:    5 : keep.source term verbatim pdf  (output_countsummary.Rnw:221) 
INFO  @ Thu, 22 Mar 2018 16:49:45:    6 : keep.source term verbatim pdf  (output_countsummary.Rnw:228) 
INFO  @ Thu, 22 Mar 2018 16:49:45:    
INFO  @ Thu, 22 Mar 2018 16:49:45:   You can now run (pdf)latex on ‘output_countsummary.tex’ 
INFO  @ Thu, 22 Mar 2018 16:49:45:    
INFO  @ Thu, 22 Mar 2018 16:49:45: End command message. 
```

tools/mageck/mageck_count.xml
```diff
@@ -99,18 +111,19 @@ $adv.test_run
...
             <param name="gmt_file" argument="--gmt-file" type="data" format="tabular" optional="true" value="" label="Pathway file for QC" help="TThe pathway file used for QC, in GMT format. By default it will use the GMT file provided by MAGeCK" />
             <param name="trim5" argument="--trim-5" type="integer" min="0" optional="true" label="5' Trim length" help="Length of trimming the 5' of the reads. Default: 0" />
-            <param name="norm_method" argument="--norm-method" type="select" label="Method for normalization" help="Methods include: None (no normalization), Median (median normalization), Total (normalization by total read counts), Control (normalization by control sgRNAs specified by the --control-sgrna option). Default: Median" >
+            <param name="norm_method" argument="--norm-method" type="select" optional="true" label="Method for normalization" help="Methods include: None (no normalization), Median (median normalization), Total (normalization by total read counts), Control (normalization by control sgRNAs specified by the --control-sgrna option). Default: Median" >
+                <option value="median">Median</option>
```
bgruening: you can leave selected=True

mblue9: done

mblue9:
I'm not sure how to get around this `pdflatex is not available` issue that's causing Travis to fail. I think this may have passed Travis previously because I didn't have pdflatex in the VM I was using. However, now that I'm testing this on my Mac where I do have pdflatex I get the full report with no pdflatex error.

So plan B - I could revert the PDF back to just being the plots (no full report) and then add to the Help that the report can be generated locally with R, if you have pdflatex installed, using the .Rnw and the normalized counts file.

Should I do that?

bgruening:
@mblue9 maybe I'm blind but you haven't defined a dependency on pdflatex isn't it?
pdflatex can be a big package, so not sure if its worth it to include. Your call.

mblue9:
Thanks @bgruening and nope you're not blind but could you maybe be suffering from a small touch of amnesia 😄 as I told you it looks like I need the texlive-core conda package to use pdflatex and it has this issue https://github.com/conda-forge/texlive-core-feedstock/issues/19
(also should texlive-core be added to the conda mageck package instead of this wrapper? as the conda package also can't generate the PDF report unless pdflatex is already installed on the system)

Have now confirmed that if I add texlive-core as a requirement to mageck I get the same error that @nsoranzo reported here https://github.com/conda-forge/texlive-core-feedstock/issues/19#issuecomment-323746351 `can't find the format file 'pdflatex.fmt'!`

So ok, looks like pdflatex is not an option atm (correct me if I'm wrong) so I'll go with plan B ^^.

nsoranzo:
I can confirm that `pdflatex` from the `texlive-core` conda-forge package is broken, an alternative seems to be [tectonic](https://tectonic-typesetting.github.io), for which a working conda-forge package exists, but:
- you need to change the calls to `pdflatex` to calls to `tectonic` (or create a `pdflatex` bash script that map them)
- `tectonic` requires an Internet connection to download LaTeX packages on the fly, which means that the tool may not work e.g. on cluster nodes with no Internet access.

mblue9:
Thanks for that info @nsoranzo! I like the sound of `tectonic` (hadn't known about that). 

For this point:

>tectonic requires an Internet connection to download LaTeX packages on the fly, which means that the tool may not work e.g. on cluster nodes with no Internet access.

Some questions I have are:

1.  Is needing an Internet connection acceptable or does that rule out using tectonic here? 
2. Is there any way to provide the files tectonic needs to create this report (as should be same every time?) without downloading them on the fly?

and for your other point:

>you need to change the calls to pdflatex to calls to tectonic (or create a pdflatex bash script that map them)

3.  If tectonic was to be used instead of pdflatex should that be changed in:

- a) this wrapper
-  b) the mageck conda package
-  c) mageck itself

bgruening:
Imho we should go to fix the texlive-core package. Internet connection is to be avoided if we can.

mblue9:
Ok, I don't know how to fix the texlive-core package but happy to add it to mageck when it's ready. 
If fixing texlive could take a while, what about merging this PR in the meantime?

bgruening:
Fine with me :)
I hope I will find time to work on this conda package soon.

mblue9:
Thanks @bgruening ! 

I think I may have messed up the versioning here, as when I went to try to update mageck in our instance from the Admin panel I tried updating `suite_mageck` first which didn't pull in the changes :( Looks like I have to go through all the mageck modules individually to update them (mageck count, mageck test etc)? Is there a better way I should have updated the wrapper so I could just have updated mageck through updating `suite_mageck` in the Admin panel? 

(My strong preference would be to only have one wrapper version for a suite, rather than one for each xml, as that's unnecessarily confusing and cumbersome imho.)

bgruening:
I'm all in for one wrapper-version per suite.
The other problem sounds more like a general suite problem. I don't think suites do work like we all expect in the moment.

PR #1793, opened 2018-03-26
Fix advanced scoring options in Hisat2

### LINKED ISSUE #1792: HISAT2 2.1.0 fails when options under "Scoring options" used
Tool version 2.1.0 (most current): https://toolshed.g2.bx.psu.edu/view/iuc/hisat2/40e1083a4430

Test history: https://usegalaxy.org/u/jen/h/test-history-hisat2-soft-clip

 - Dataset 6: uses all defaults and is successful
 - Dataset 7: has the "scoring options" section opened at job submission, but no changes made
 - Dataset 8: has "scoring options" section opened and soft-clip set to Yes (default is No)
 - Datasets 9-18: Test jobs using the two prior versions of the tool (2.0.5.2, 2.0.5.1) are running try to find out when this was introduced. Same test pattern. 
 - Dataset 19: Rerun of dataset 7, but with reference annotation also added under "spliced alignment options" -- expecting it to fail, is just a test to see if a reference annotation (GTF) is for some reason required. If it is, then that input dependency will need to be clarified on the tool form.

I'll also test at https://usegalaxy.eu to make sure this isn't server-specific. The error does not look like a dependency problem, but a tool wrapper problem:

screen shot 2018-03-26 at 11 32 21 am

mvdbeek:
Should fix #1792

### DIFF
--- a/tools/hisat2/hisat2.xml
+++ b/tools/hisat2/hisat2.xml
@@ -233,8 +233,8 @@ hisat2
     --np ${adv.scoring_options.ambiguous_penalty}
     --rdg ${adv.scoring_options.read_open_penalty},${adv.scoring_options.read_extend_penalty}
     --rfg ${adv.scoring_options.ref_open_penalty},${adv.scoring_options.ref_extend_penalty}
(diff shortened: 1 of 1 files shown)


bgruening:
Urgs .... thanks!

Issue #1454, opened 2017-08-30
Incoming tool wrappers for pRESTO


tfenne:
Hi iuc team!  I'm creating this issue pursuant to the IUC Best Practices for Tool Creation to inform you that I'm working on a set of `tool.xml`s for the tools contained within the [pRESTO](https://presto.readthedocs.io/) toolkit for analyzing immune repertoire sequencing experiments.

I'm working in collaboration with @bwlang, and have an in progress branch - diff vs. tools-iuc master [here](https://github.com/galaxyproject/tools-iuc/compare/master...tfenne:tf_presto_tools).

I still have a little more work to do, mostly wrapping one or two more tools and then manually inspecting the tool interfaces to make sure the params and help look good.  Once that's done I'll open a PR here and request a review.  If there's anything else I can do to help make this process easier/smoother please just let me know.  Cheers!

shiltemann:
Sounds great @tfenne! looking forward to the PR :) ..if you have any questions feel free to ask, and you are also welcome to open a WIP PR here if you would like some input from us as you go, great work!

tfenne:
Thanks @shiltemann!  I do have on question.  I read that, after merge, tools are automatically published to the test and main toolsheds.  Will it screw anything up if I independently use planemo to publish to the test toolshed?  I'd like to make the tools easily available to a couple of collaborators to help test them before generating a PR.

bgruening:
I just looked briefly over the diff and this already looks great. A few tips which might already help.

* try to use `argument="--maxerror"` https://docs.galaxyproject.org/en/master/dev/schema.html#common-attributes
* you can use `from_work_dir` instead of your `mv` commands: https://docs.galaxyproject.org/en/master/dev/schema.html#id53
* better help sections can be easily written with http://rst.ninjs.org/
* you can macrofy the version for example
* all the `Copy/pasted values from align` could be macrofied I think

bgruening:
@tfenne you can use the TTS as you like you will not screw something up :)

tfenne:
Thanks @bgruening!  I'm new to a lot of this (galaxy, tool.xml wrappers, cheetah, rst).  I appreciate the feedback and should be able to fix most of those things pretty quickly.  One question I can't find an answer to online, does `from_work_dir` support any kind of dynamic eval/globbing?  In most cases I know the output filename ahead of time but in at least one case the name includes a dynamic component - it actually embeds an input parameter.

bgruening:
[...]

PR #1797, opened 2018-03-27
Wrappers for pRESTO suite of immune-sequencing analysis tools

tfenne:
I opened issue #1454 months ago when this work was 80-90% done and then promptly got pulled off onto other projects.  I think this is finally ready for review and merging!


tools/presto/presto_buildconsensus.xml
```diff
@@ -0,0 +1,118 @@
...
+    <expand macro="requirements"/>
+    
+    <version_command>BuildConsensus.py --version</version_command>
+    <command detect_errors="exit_code"><![CDATA[
+        ln -s $fastq_in in.fastq &&
+        BuildConsensus.py
+          --nproc "\${GALAXY_SLOTS:-1}"
+          -s in.fastq
+          -n $n
+          --bf $bf
```
bgruening: single-quote as well as the other text parameters
bf also seems to be used a lot, maybe put them into the macro?

tools/presto/presto_assemblepairs.xml
```diff
@@ -0,0 +1,186 @@
...
+                <option value="usearch">USearch</option>
+            </param>
+        </xml>
+    </macros>
+    
+    <expand macro="requirements"/>
+    
+    <version_command>AssemblePairs.py --version</version_command>
+    <command detect_errors="exit_code"><![CDATA[
+        ln -s $r1_in r1.fastq &&
```
bgruening: single-quotes and the other tools as well ...

tools/presto/presto_alignsets.xml
```diff
@@ -0,0 +1,55 @@
...
+            --div
+          #end if
+          --outname=tmp
+          #if $capture_log
+            --log $log_out
+          #end if
+    ]]></command>
+
+    <inputs>
+        <param argument="-s"    name="fastq_in"  type="data" format="fastq" label="Input FASTQ file" help="FASTQ file of sequences with barcodes/UMIs in an annotation."/>
```
bgruening: can you try to use simple single spaces

tools/presto/presto_alignsets.xml
```diff
@@ -0,0 +1,55 @@
...
+    
+    <expand macro="requirements"/>
+    
+    <version_command>AlignSets.py --version</version_command>
+    <command detect_errors="exit_code"><![CDATA[
+        ln -s $fastq_in in.fastq &&
+        AlignSets.py muscle
+          --nproc "\${GALAXY_SLOTS:-1}"
+          -s in.fastq
+          --bf $bf
```
bgruening: needs also to be single-quoted

tools/presto/presto_parseheaders.xml
```diff
@@ -0,0 +1,150 @@
...
+    
+    <expand macro="requirements"/>
+    
+    <version_command>ParseHeaders.py --version</version_command>
+    <command detect_errors="exit_code"><![CDATA[
+        ln -s $fastq_in in.fastq &&
+        ParseHeaders.py $cc.command
+          -s in.fastq
+          
+          #if $varExists('cc.f')
```
bgruening: @galaxyproject/iuc how to you feel about $varExists?

tools/presto/presto_parseheaders.xml
```diff
@@ -0,0 +1,150 @@
...
+          #if $varExists('cc.act')
+            --act $cc.act
+          #end if
+          #if $varExists('cc.k')
+            -k $cc.k
+          #end if
+
+          --outdir=.
+          --outname=tmp &&
+        mv ./tmp_*header* $output
```
bgruening: output also needs to be quoted

tools/presto/presto_parselog.xml
```diff
@@ -0,0 +1,44 @@
...
+        <import>presto_macros.xml</import>
+    </macros>
+    
+    <expand macro="requirements"/>
+    
+    <version_command>ParseLog.py --version</version_command>
+    <command detect_errors="exit_code"><![CDATA[
+        ParseLog.py 
+          -l $log_file
+          -f $fields
```
bgruening: need to be quoted as well

tools/presto/presto_parselog.xml
```diff
@@ -0,0 +1,44 @@
...
+          --outdir=.
+          --outname=tmp
+    ]]></command>
+
+    <inputs>
+        <param argument="-l" name="log_file" type="data" format="txt" label="pRESTO log file" help="Detailed log file generated from a pRESTO tool."/>
+        <param argument="-f" name="fields"   type="text" label="Field(s)" help="Space-separated list of fields to extract."/>
```
bgruening: this field should probably be sanitized and only letters/numbers allowed?

tfenne: It's hard to say.  There's nothing stopping users from outputting field names with non-alphanumeric characters in them, so I'm reluctant to restrict the input.

tools/presto/presto_alignsets.xml
```diff
@@ -0,0 +1,55 @@
...
+    
+    <macros>
+        <import>presto_macros.xml</import>
+    </macros>
+    
+    <expand macro="requirements"/>
+    
+    <version_command>AlignSets.py --version</version_command>
+    <command detect_errors="exit_code"><![CDATA[
+        ln -s $fastq_in in.fastq &&
```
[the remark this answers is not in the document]
tfenne: I've added single quotes to _most_ text and data parameters throughout the PR.  There are some text parameters that take in multiple values that are user entered, e.g. `fields` in `presto_parselog.xml`.  Those I've left unquoted because otherwise values like `PRIMER CONSCOUNT` are incorrectly parsed by presto as a single field name instead of 2.  Admittedly I could use `multiple="true"` on all these parameters, but I think that's actually more cumbersome when the values are short text strings typed in by the user.

tools/presto/presto_alignsets.xml
tfenne: I've gone through and everywhere I could spot remove multiple spaces within xml elements.

bgruening:
Thanks @tfenne! Can you have a look at the tests? They are still failing.

tfenne:
Will do, sorry @bgruening .

tfenne:
@bgruening  Fingers crossed this commit does it.  It appears the failures were caused by differences in the ordering of the records output to the fastq files.  The tests were passing locally, but apparently the travis tests run in a different enough environment that the ordering changes.  

Since the order in the fastqs isn't important, I've just added `sort="true"` to a bunch of test outputs.  That feels a bit hacky for a structured file like fastq where ideally you'd like to sort records and not lines, but I think it's probably ok and certainly a lot less of a time-sink that trying to track down why the ordering is different.

tfenne:
@bgruening I'm a at a bit of a loss on how to proceed.  The tests are passing for me locally.  With that last commit I went from 5 failures to 4 in travis, but the change is actually two of the previous failures passed, and one new failure popped in.

Is there a way to get more detailed information on the test failures than what's available in the travis console log?  Or alternatively is there a better way to run my tests than just `planemo test` in order to better simulate the environment in travis?

bgruening:
It seems to be still the odering. Have you updated your test data? You can use `--update_test_data` for this. Or try to use `contains` as test method or a regular expression. Both just check a part of the file that is created and can be therefor independent of the ordering.

bgruening:
ping @tfenne

tfenne:
@bgruening Sorry for taking so long on this.  I've been trying to get to the bottom of why the tests are non-deterministic on linux but not on mac.  I've given up though and have essentially reduced the tests to the point where most of them are running on fastqs with a single read in them.  I think this is probably fine as it demonstrates that the tools run successfully under galaxy, but it doesn't feel awesome.  

I just pushed an update to the branch.  Fingers crossed the build passes this time.

tools/presto/presto_alignsets.xml
```diff
@@ -0,0 +1,55 @@
...
+            --div
+          #end if
+          --outname=tmp
+          #if $capture_log
+            --log '$log_out'
+          #end if
+    ]]></command>
+
+    <inputs>
+        <param argument="-s" name="fastq_in" type="data" format="fastq" label="Input FASTQ file" help="FASTQ file of sequences with barcodes/UMIs in an annotation."/>
```
bgruening: Do these tools also work on fastqz files?

tools/presto/presto_assemblepairs.xml
```diff
@@ -0,0 +1,186 @@
...
+          -2 r2.fastq
+          --coord $coord
+          #if $rc
+            --rc '$rc'
+          #end if
+          #if $r1_annotations
+            --1f $r1_annotations
+          #end if
+          #if $r2_annotations
+            --2f $r2_annotations
```
bgruening: single quotes around all text parameters

tools/presto/presto_assemblepairs.xml
```diff
@@ -0,0 +1,186 @@
...
+    <inputs>
+        <param argument="-1" name="r1_in" type="data" format="fastq" label="Read 1 FASTQ file"/>
+        <param argument="-2" name="r2_in" type="data" format="fastq" label="Read 2 FASTQ file"/>
+        <param argument="--rc" type="select" label="Reverse Complement" value="" help="Reverse complement neither, R1, R2 or both reads before assembly.">
+            <option value="">Neither</option>
+            <option value="head">Read 1 Only</option>
+            <option value="tail">Read 2 Only</option>
+            <option value="both">R1 and R2</option>
+        </param>
+        <param argument="--1f" name="r1_annotations" type="text" optional="true" label="R1 Fields To Copy" help="Annotation fields to copy from R1 into the output."/>
```
bgruening: What are fields here?

tools/presto/presto_assemblepairs.xml
```diff
@@ -0,0 +1,186 @@
...
+    </inputs>
+    
+    <outputs>
+        <data name="fastq_out" format="fastq" from_work_dir="tmp_assemble-pass.fastq"/>
+        <expand macro="presto-log-output"/>
+    </outputs>
+    
+    <tests>
+        <test>
+            <param  name="command" value="join"/>
```
bgruening: We usually don't do these alignments with spaces ...

tools/presto/presto_assemblepairs.xml
```diff
@@ -0,0 +1,186 @@
...
+            <param  name="r" value="reference_repertoire.fasta"/>
+            <param  name="aligner" value="blastn"/>
+            <output name="fastq_out" file="presto_assemblepairs_test_sequential_out.fastq" sort="true"/>
+        </test>
+    </tests>
+    
+    <help><![CDATA[
+Assembles paired-end reads into a single sequence.
+
+See the See the `pRESTO online help <@PRESTO_URL_BASE@/AssemblePairs.html>`_ for more information.
```
bgruening: See the See the ...

bgruening:
Upps, completely dropped the ball here, sorry.

tools/presto/presto_assemblepairs.xml
bgruening: The problem with the unqoted once is that a user could include something like `; rm -rf /; echo` in your text box and you would execute it.
Can we make this parameter a select box with multiple options or a repeat with single text params, so a user needs to specify one string after the other?

tools/presto/presto_assemblepairs.xml
bgruening: Gotcha. Would be a repeat with a text field on option when you don't know the strings a priori?

tools/presto/.shed.yml
```diff
@@ -0,0 +1,18 @@
...
+  paired-end reads. It includes features for quality control, primer masking, annotation of reads with sequence
+  embedded barcodes, generation of unique molecular identifier (UMI) consensus sequences, assembly of
+  paired-end reads and identification of duplicate sequences. Numerous options for sequence sorting, sampling
+  and conversion operations are also included.
+name: presto
+owner: iuc
+remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/master/tools/presto
+type: unrestricted
+auto_tool_repositories:
+  name_template: "presto_{{ tool_id }}"
```
bgruening: the tool_id is already `presto_alignsets` so the repo-name would become `presto_presto_alignsets`.
`name_template: {{ tool_id }}` would be better I think. But please make sure all tool.ids are starting with presto_

tools/presto/presto_partition.xml
```diff
@@ -0,0 +1,53 @@
+<tool id="presto_partition" name="pRESTO Partition" version="@PRESTO_VERSION@">
+    <description>Partition a file in two.</description>
```
bgruening: the dot can be removed

tools/presto/prestor_abseq3.xml
```diff
@@ -0,0 +1,71 @@
...
+        ln -s '$qual1'    ./quality-1_table.tab &&
+        ln -s '$qual2'    ./quality-2_table.tab &&
+        ln -s '$primer1'  ./primers-1_table.tab &&
+        ln -s '$primer2'  ./primers-2_table.tab &&
+        ln -s '$cons1'    ./consensus-1_table.tab &&
+        ln -s '$cons2'    ./consensus-2_table.tab &&
+        ln -s '$assemble' ./assemble_table.tab &&
+        ln -s '$maskqual' ./maskqual_table.tab &&
+        ln -s '$cregion'  ./cregion_table.tab &&
+        echo "library(prestor); report_abseq3('.', sample='$sample', output_dir='.', output_file='report.html', format='html')" > ./tmp.R &&
```
bgruening: You could use a configfile here as shown here: https://github.com/galaxyproject/tools-iuc/blob/master/tools/vegan/vegan_rarefaction/vegan_rarefaction.xml#L19

tools/presto/prestor_abseq3.xml
```diff
@@ -0,0 +1,77 @@
...
+    <configfiles>
+        <configfile name="prestor_script"><![CDATA[
+library(prestor)
+report_abseq3('.', sample='${sample}', output_dir='.', output_file='report.html', format='html')
+        ]]>
+        </configfile>
+    </configfiles>
+
+    <inputs>
+        <param name="sample" type="text" label="Sample Name" help="Sample Name. Must not contain spaces."/>
```
bgruening: You could add a sanitizer here as well?

bgruening:
Thanks a lot @tfenne and sorry that it took so long!

tfenne:
Woohoo!  Thanks @bgruening!

PR #1798, opened 2018-03-27
Fix versioning for snpEff tools.

davebx:
Also accidentally fixed trailing whitespace.

Old: `LegacyVersion('4.3.1t.1')`
New: `Version('4.3.1.1')`

### DIFF
--- a/tool_collections/snpeff/snpEff_macros.xml
+++ b/tool_collections/snpeff/snpEff_macros.xml
@@ -13,7 +13,7 @@
 snpEff -version
     ]]></version_command>
   </xml>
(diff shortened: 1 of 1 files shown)


nekrut:
@davebx let's use something like `<token name="@wrapper_version@“>4.3+T.galaxy1</token>` per our conversatin with @mvdbeek

martenson:
this change to the macro will result in `4.3+T.galaxy1.1` because tools specify it as  `<tool id="snpEff" name="SnpEff eff:" version="@wrapper_version@.1">`

PR #1800, opened 2018-03-28
Prokka version bump to 1.13 and fixed locus tag parameter

Slugger70:
to make it compatible with Roary tool


tools/prokka/prokka.xml
```diff
@@ -1,14 +1,11 @@
-<tool id="prokka" name="Prokka" version="1.12.0">
-    <description>prokaryotic genome annotation</description>
+<tool id="prokka" name="Prokka" version="1.13.galaxy0">
```
bgruening: I'm just not sure about the version here. Did we agreed on something on `galaxy` already? If not leave the number schema fro before until we have settled on the new schema.

Slugger70: It's already been used in a couple of other tools. Dave B did it for the new snpeff

Slugger70: But I'm happy to change it back too.

erasche: If you're doing this it should probably be `+galaxy` instead.

davebx:
Thanks @Slugger70

PR #1801, opened 2018-03-28
[WIP] upgrade samtools and add merge and view

bernt-matthias:
This is at the moment only for note keeping (to check all the PRs after they are merged): 

- [ ] check all tool versions 
- [ ] re test everything
- [ ] introduce regions macros/tokens (see stats) in the other lools (view is buggy at the moment)

- [ ] Do we want/need suite_samtools as in the devteam repo + take care of https://github.com/galaxyproject/tools-iuc/issues/1814

- [ ] http://www.htslib.org/doc/samtools.html#GLOBAL_OPTIONS
- [ ] samtools cat, tview, and some others


mvdbeek:
Yes please, if we can update everything that'd be great, so many fixes and improvements in the newer htslib releases!

tool_collections/samtools/samtools_merge/samtools_merge.xml
```diff
@@ -0,0 +1,82 @@
+<!-- TODO can we restrict the input to sorted alignments -->
+<!-- TODO can we set the output sort order based on the input sort order -->
```
bgruening: Yes, @mvdbeek can added in 18.01 different types of BAM datatypes. Sorted in different ways.

tool_collections/samtools/samtools_merge/samtools_merge.xml
```diff
@@ -0,0 +1,82 @@
...
+<!-- TODO 2.0.1. is the somtools sort's version .. shouls I change this? -->
```
bgruening: We can unfortunately not decrease it, we can only increase it ... :(

bernt-matthias: Then we increase it in wee steps .. maybe samtools version numbers catch up eventually.

bernt-matthias:
Question CRAM needs a reference sequence for reading and writing. Is this stored along the data set? Then I could pass this implicitly for tools that read CRAM.

mvdbeek:
That isn't possible atm, you can disable reference-based compression with `--output-fmt=cram,no_ref`

bernt-matthias:
Work is going on. I have added 'some' TODOs to the xml files where I could need input from you. 

Some general questions have been added to the are now in the 1st message of this PR. 

Can you point me to some documentation where I can read about how to determine the type of input (sam,bam,...) in the tool code and how to access properties of the data sets. I have seen some examples in the xml files already (e.g. accessing the bam index) - and I would like to read more about it.

mvdbeek:
Determining the input type is documented [here]( https://planemo.readthedocs.io/en/latest/writing_how_do_i.html#check-input-type-in-command-blocks). The closest thing to a documentation about metadata elements is [here](https://docs.galaxyproject.org/en/latest/dev/schema.html#id9).
The actual attributes that you can use are not really documented systematically, but they are registered  in the datatype class in question, e.g in the [CRAM class](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/datatypes/binary.py#L549)

tool_collections/samtools/samtools_merge/samtools_merge.xml
```diff
@@ -0,0 +1,82 @@
...
+    <description>merge multiple sorted alignment files</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements"/>
+    <expand macro="stdio"/>
+    <expand macro="version_command"/>
+    <command><![CDATA[
+        samtools merge
+        #if $comression == 'levelone'
```
mvdbeek: s/comression/compression/

tool_collections/samtools/samtools_merge/samtools_merge.xml
```diff
@@ -0,0 +1,82 @@
...
+        #else if $compression == 'uncompressed'
+            -u
+        #end if
+        ## 
+        #if str($headerbam) != 'None'
+            -h '$headerbam'
+        #end if
+        ##The input alignments are sorted by read names rather than by chromosomal coordinates 
+        ##-n
+        ##TODO for -n: can we find out the sorting type state of the input?
```
mvdbeek: #if $input.is_of_type('qname_sorted.bam', 'qname_input_sorted.bam')
  some_command
#end if

tool_collections/samtools/samtools_fastx/samtools_fastx.xml
```diff
@@ -1,3 +1,5 @@
+<!--TODO Samtools fastq can now create compressed fastq files, by giving the
```
mvdbeek: This comment breaks the linting

tool_collections/samtools/samtools_merge/samtools_merge.xml
bernt-matthias: Corrected. Question is if such options should be visible to the user or if we just force max/default compression in Galaxy...?

tool_collections/samtools/samtools_merge/samtools_merge.xml
bernt-matthias: Thanks. Am I right, that bam can be assumed to be coordinate sorted?

tool_collections/samtools/samtools_fastx/samtools_fastx.xml
bernt-matthias: odd. the command runs locally for me. But I moved the comment one line down. Maybe this helps.

tool_collections/samtools/samtools_merge/samtools_merge.xml
bernt-matthias: btw. why are there special data types if the metadata contains the sorting information: 
https://github.com/galaxyproject/galaxy/blob/4fc6380708454c15360f0940ee56998c08fc6c8a/lib/galaxy/datatypes/binary.py#L200 ? Or is it because `visible=False` is set for this metadata element?

Are the data types already used somewhere?

tool_collections/samtools/samtools_merge/samtools_merge.xml
mvdbeek: I'd be in favor of default, which is a decent middle ground IMO.

tool_collections/samtools/samtools_merge/samtools_merge.xml
mvdbeek: Yes, bam should be coordinate sorted, but you can double check using the existence of the index file, which can only be produced for coordinate sorted files.
Having the sort_order in the datatype doesn't really help when galaxy should be doing implicit conversions, it's just a metadata field at the moment (though we do want to change this eventually)

tool_collections/samtools/samtools_fastx/samtools_fastx.xml
```diff
@@ -10,15 +10,24 @@
     <![CDATA[
         samtools $output.output_format
             #if $output.output_format == 'fastq':
-                -v $output.default_quality $output.output_quality
+               -v $output.default_quality
+               $output.output_quality
+               $ilumina_casava
```
mvdbeek: This needs to be prefixed with the conditional name, e.g ($output.illumina_casave), but I'd also change the name of the conditional to something more descriptive (`output_format` ?), since $output is often used as the actual output file parameter.

tool_collections/samtools/samtools_fastx/samtools_fastx.xml
```diff
@@ -29,18 +38,50 @@
...
-            '$input'
+            #if str($exclusive_filter) != 'None':
+                #set $filter = $exclusive_filter_all
+                @FLAGS@
+                -G $flags
+            #end if
+            #if str($write_singletons) == 'yes':
+                -s $singletons
+            #end if
+            #if str($index_format) != '':
```
mvdbeek: same as above, needs to be prefixed with the conditional

bernt-matthias:
Just comitted a good 1st candidate for samtools_view

- fixed index data
- added possibility (now default) to keep the format
- added sorting for SAM to BAM/CRAM conversions
- region filter bed/text
- added possibility to use a random seed for subsampling
- added tests (I recycled the tests from the samtools converters in the tool suite)

Would be happy about comments (in particular on my TODOs). 

Currently some tests fail: 
- some due to size differences (maybe the new version generates larger files?)
- I struggle to get the examples with cached reference data working

From my side I hope that I can give the help section a bit more love and add tests for the main filters.

bernt-matthias:
Made quite a bit of progress. Would be happy for comments, in particular to my questions which are hidden as `TODO` in the xml files.

bernt-matthias:
I need to stop working on this now. I would suggest to fix the remaining problems with some tests and all other things that you find. For this I need your input @bgruening and @mvdbeek.

bernt-matthias:
Hi @peterjc. Seems to be the case, since the planemo tests do not complain. I assume that for planemo tests GALAXY_SLOTS=1, or am I wrong?

mvdbeek:
These are a whole lot of questions and I guess for most of them the answer would be "probably, if you have time ?". Why not split the PR in smaller tasks, that would make reviewing a lot easier, and travis tests should be quicker as well.

bernt-matthias:
OK. Will do :)

bernt-matthias:
First 6 PRs are out :) These are the tools where I actually need the update. More PRs will follow soon.

PR #1802, opened 2018-03-28
Updated umi_tools whitelist, paired input

mtekman:
Adds paired collection input.

Works better in conjunction with other demultiplexing tools such as Je, when feeding inputs inside a workflow


bgruening:
You need to bump the tool version number.

Issue #1694, opened 2018-01-30
diffbind should be able to take collections for case/control


zfrenchee:
DiffBind is often used with multiple samples / replicates. In Galaxy, you can input these manually, one by one, but if you have 10 or 100, this isn't particularly feasible.

Migrating from https://github.com/bgruening/galaxytools/issues/670 because of https://github.com/galaxyproject/tools-iuc/pull/1678

(cc @bgruening , @mblue9 )

zfrenchee:
I believe this was solved by @mblue9

PR #1808, opened 2018-04-03
DiffBind - rework to make collection-friendly and other improvements

mblue9:
Inspired by the Great Work of @jmchilton on the new Rule Builder for collections, I found some fresh enthusiasm to go back and retackle diffbind - the most collection-unfriendly of all the Galaxy tools I know. As diffbind's current wrapper only allows collections to have one dataset, which is awkward and very limiting. And there's also an excessive amount of manual typing/copy-paste required for each sample in the tool form.  See issue https://github.com/galaxyproject/tools-iuc/issues/1694

ping @zfrenchee


mblue9:
Thanks for the review @bgruening! I found some other things that look like they could be changed to improve the wrapper, but at the risk of turning this into another bomb of unrelated commits and delaying things, I've put them into an issue here: https://github.com/galaxyproject/tools-iuc/issues/1809

bgruening:
No hurry from my side :)

mblue9:
Ok cool, if noone objects I might try to add some of the https://github.com/galaxyproject/tools-iuc/issues/1809 changes here then, which I could group loosely into an atomic PR called "diffbind - rework to make collection-friendly and other improvements"  :). As I need to use this tool and would prefer if it had some of those things incorporated. Will aim to have the extra changes in by sometime next week.

zfrenchee:
One small request on my side: If the FDR cutoff could be a text field rather than a slider that could be nice.

mblue9:
@zfrenchee the FDR cutoff is currently both a slider and a text box and you can use either, see here:

screen shot 2018-04-05 at 1 57 17 pm

Is that ok?

tools/diffbind/diffbind.xml
```diff
@@ -143,10 +147,13 @@ Rscript '$__tool_directory__/diffbind.R'
         <data name="rdata" format="rdata" from_work_dir="DiffBind_analysis.RData" label="${tool.name} on ${on_string}: RData file">
             <filter>out['rdata']</filter>
         </data>
+        <data name="rscript" format="txt" from_work_dir="*.txt" label="${tool.name} on ${on_string}: Rscript">
```
bgruening: I guess the `from_work_dir` is not needed here?

tools/diffbind/diffbind.R
```diff
@@ -6,6 +6,7 @@ Sys.setlocale("LC_MESSAGES", "en_US.UTF-8")
 suppressPackageStartupMessages({
 	library('getopt')
 	library('DiffBind')
+    library('rjson')
```
bgruening: indentation

tools/diffbind/diffbind.xml
```diff
@@ -223,7 +223,11 @@ Rscript '$__tool_directory__/diffbind.R'
             <output name="binding_matrix" value="out_binding.matrix" />
             <output name="rdata" value="DiffBind_analysis.RData" compare="sim_size"/>
             <output name="rscript" value="out_rscript.txt"/>
-            <output name="analysis_info" value="out_analysis_info.txt"/>
+            <output name="analysis_info" value="out_analysis_info.txt" compare="sim_size" >
```
bgruening: I think if you have assert_content you don't need to have sime_size? Or is both, the sim_size and the assert, checked in this case?

tools/diffbind/diffbind.xml
```diff
@@ -19,69 +20,143 @@
...
+            <option value="narrow" selected="true">narrowPeaks</option>
+            <option value="macs">MACS .xls</option>
+            <option value="swembl">SWEMBL</option>
+            <option value="bayes">bayesPeak</option>
+            <option value="fp4">FindPeaks v4</option>
+         </param>
+         <param name="peakformat" type="select" optional="True" label="Peak File Format" help="Specify format for peak files. Default: narrowPeaks">
+            <option value="raw">raw (scoreCol=4)</option>
+            <option value="bed">bed (scoreCol=5)</option>
+            <option value="narrow" selected="true">narrowPeaks (scoreCol=8)</option>
```
bgruening: Narrowpeaks is also from MACS, right?

tools/diffbind/diffbind.xml
```diff
@@ -19,69 +20,143 @@
...
+                <param name="bamreads" type="data" format="bam" multiple="true" label="Read BAM file" help="Specify the Read BAM file used for Peak calling."/>
+                <param name="bamcontrol" type="data" format="bam" multiple="true" optional="True" label="Control BAM file" help="If specifying a control BAM file, all samples are required to specify one."/>
+            </repeat>
         </repeat>
-        <param name="th" type="float" value="1" min="0" max="1"
-                label="FDR Threshold"
-                help="Significance threshold; all sites with FDR less than or equal to this value will be included in the report. A value of 1 will include all binding sites in the report. Default: 1"/>
-        
+
+        <param name="peakcaller" type="select" optional="True" label="Peak Caller" help="Specify peak caller. If Peaks is not a bed file, this will determine how the Peaks file is parsed. Default: narrowPeaks">
```
bgruening: What is the difference between the format and the caller? Currently I can mix a caller with a different format, right? Let's say bayesPeak with SCIER.

bgruening:
This gets me super excited :)

tools/diffbind/diffbind.xml
mblue9: I've removed this peakFormat parameter and the peakCaller one below also as you just confirmed what I was wondering, adding these params is probably only going to create more confusion than they're worth. These options are described in the diffbind vignette but I think just the scorecol is all that's really important. Thanks for confirming :smile:

tools/diffbind/diffbind.xml
mblue9: removed, see comment above

mblue9:
Yes I'm also excited about this one, I'd never used diffbind til I started working on this wrapper and now I can see it looks very useful, well worth the wrapping effort :smile:

tools/diffbind/diffbind.xml
bgruening: I was just wondering if Galaxy is such a case really checks to tests at once the test for similar size and the assert. If so I was not aware of this :)

bgruening:
Yeah, sorry for the first version :)
Let get this deployed.

tools/diffbind/diffbind.xml
mblue9: well I just added it in and received no complaints from planemo, got the green light so went with it..but does it actually work..hmm

PR #1810, opened 2018-04-04
mothur: chimera.vsearch fixes

shiltemann:
add count table output when a count table is given as input, and fix the output labels


bgruening:
Thanks @shiltemann!

PR #1812, opened 2018-04-04
Collection column join: add new 'Keep original column header' option

abretaud:
Hi,
A new little option for the collection_column_join tool, we'll need it in a workflow.
The option makes more sense when joining multiple files with 2 column (ie id -> value). It allows to keep only the dataset name as column header instead of <dataset_name>_<col_name>.


mblue9:
Brilliant! I also wanted this change but hadn't got around to doing anything about it. So thanks @abretaud 👍

abretaud:
Thanks @mblue9! Can I go ahead and merge it?

PR #1819, opened 2018-04-08
DiffBind: tweaks

mblue9:
Now that I can see the new diffbind wrapper live in usegalaxy.eu (that was quick!), I have some minor suggested edits in this PR e.g. removal of the unnecessary factorName parameter and some edits to text to clarify things.


mblue9:
From spending more time on this tool today, I discovered the headers in the main file output (the differential binding sites) looked odd, see screenshot below, as the output is not bed, it's a table. So I've now changed the format to tabular.

screen shot 2018-04-11 at 3 10 04 pm


I also removed the options to output gff or wig instead of the "bed" as they don't work, they also give the same "bed" output, see other screenshot below, and I couldn't see mention of bed/gff/wig format for this output in the diffbind Bioconductor vignette.

screen shot 2018-04-11 at 3 11 35 pm

bgruening:
@jennaj is this still an interval format?

mblue9:
This diffbind output comes from GRanges which is 1-based afaik and not 0-based like bed and the Galaxy interval format I see described here: https://galaxyproject.org/learn/datatypes/#interval.
So should we just stick with tabular?

bgruening:
Can we use something like this: https://www.biostars.org/p/89341/
And use 0-based. I have the feeling that the 1 based stuff yields to a lot of errors down the road.

If not I'm fine with tabular.

mblue9:
What about leaving this output as tabular format and instead including in training material how to make a bed file out of this (if so desired)? As the R diffbind tool doesn't output bed format (and it's not mentioned in diffbind training material that I'm aware of e.g. https://bioinformatics-core-shared-training.github.io/cruk-autumn-school-2017/ChIP/Materials/Practicals/Practical9_diffbind_DB.pdf so imho it would be better to not add potential for confusion by outputting bed format in this wrapper and instead to stick with tabular.

bgruening:
Mh, I think it better to provide standard formats to user and prevent errors. This tabular is not a standard and its not that easy to convert it to a BED afaik. But that just my opinion ... ready to merge.

mblue9:
Thanks for merging this @bgruening! I'm planning to come back to diffbind again when I can get a chance (e.g. to add the edgeR method option) and I was thinking I could add an option then to output bed or interval. 

One slight complication is this diffbind table is not standard bed format afaik (e.g. see
https://genome.ucsc.edu/FAQ/FAQformat.html) because the columns are:

`chr, start, end, width, strand, Conc, Conc_Group1, Conc_Group2, Fold, p.value, FDR`

but I guess I could chop out columns to make it bed-like? or output interval format instead if that's better in this case?

jennaj:
For next revision, I vote for `Interval` and correcting the `[start]` coordinate value to be 0-based for end users. This meets specification, retains all information, and avoids users hunting around for help/FAQs and/or reporting wierd usage errors, or getting technically invalid results in a presumably "successful" green result dataset (from this tool, or others used downstream).

Even if the tool form help itself stated the data is 1-based with in-line tool help/instructions about how to manipulate the result to be an actual `Interval` (or `BED`), is a multi-step hassle: involves "add column" >> "cut" >> metadata "datatype" change. Sounds like we all agree would be a really nice-to-have next time.

----

~Glorious~ details

If we go with `Interval`, the descions are what columns/data to assign to `[score]`, `[name]`, and `[comment`]. The order doesn't really matter, although if the first three are `[chrom]`, `[start]`,` [end]`, that makes it slightly more friendly as an input (worth it imho and the data is already ordered that way).

 - where `[start]` is the original `start` with "1" already subtracted (to make it 0-based)
 - where `[name`] can be any single alphanumeric "word", or a placeholder value like "diffbind"
 - where `[score]` can be any integer in 0-1000 range, or a placeholder value like "0"
 - where `[strand]` can be +, -, or "." ("." means no strand assigment, and most tools will interpret that as "+" or ignore it)
 - where `[comment]` is a like a blob, eg: alphanumeric plus ascii chars like `[:|;|-|_|.]`. Technically could contain spaces, but I tend to dislike internal spaces in tabular datasets, can cause a raft of problems, Galaxy or line-command, and requires more data manipulations to get rid of them or parse out content. But is a corner-case issue, meaning spaces may or may not matter big picture - you two decide.

Would suggest not using/creating `BED` as the primary output -- the data does not fit that specification and if made to, there is information loss. But it might be helpful to include it as an optional output on the tool form (and still always produce the full interval output). 

Just include this to make a proper `BED` (bed6 variant):

```[chr] [start] [end] [name] [score] [strand]```

Same format/content requirements as `Interval` for the common data columns.

Could make a `BED3` instead, but if can keep strand, that is better, and it requires at least a `BED6` format. A `BED3` is very simple for a user to create from a `BED6` (single "Cut").

`BED` spec: https://galaxyproject.org/learn/datatypes/#bed && http://genome.ucsc.edu/FAQ/FAQformat.html#format1

`Interval` spec: https://galaxyproject.org/learn/datatypes/#interval

**Just to complicate it: Can the tool produce negative `start` coordinates and/or overhanging `end` coordinates (eg. extend past the end of the `chr`)?** Some tools do and that data formatting can produce errors with many tools. If this tool does either (I don't know), a tool form option to "Trim overhanging coordinates" could be included and set to "Yes" at a default. And then the tool trims based on the database assignement (start is easy, if negative make it "0", end more work and would pull actual chrom end from the database "dbkey.len" file/data table). Should include trimming even when a custom genome from the history is used (best) or optionally the custom build "database" metdata is assigned (not best, requires extra steps where the user needs to first create then assign a custom build to inputs). Would apply to both the `interval` and the optional `bed` output (if we include that). Would be preferable if the "trim or not to trim" setting is somewhere in Advanced options nested away from primary form options, so is not as easy to find/uncheck unless the end user makes an active decision and has a specific usage goal in mind.

mblue9:
Thanks @jennaj, I'll work on this shortly as I need diffbind bed format now for this ChIPseeker tool https://github.com/galaxyproject/tools-iuc/pull/1898

mblue9:
@jennaj I've changed to outputing interval and optional bed in the PR here: https://github.com/galaxyproject/tools-iuc/pull/1899

See what you think.

For this question:
>Just to complicate it: Can the tool produce negative start coordinates and/or overhanging end coordinates (eg. extend past the end of the chr)?

I don't know that it can so what about waiting to see if users report errors and dealing with it then.

PR #1805, opened 2018-03-30
fix for `-cvsStats` option of snpEff

nekrut:
Additional fixes for snpEff:

 - Added `-cvsStats` option
- Modified multiQC to render SnpEff reports

Issue #1658, opened 2018-01-04
MultiQC for list of pairs (of FastQC output)


bernt-matthias:
If I run FastQC on a List of Paired data I get also the FastQC results as list of pairs. These are not selectable as input in a MultiQC analysis. 

I'm unsure if this should be fixed in FastQC (maybe it should just output a list of (unpaired) reports) or MultiQC which maybe should also support list of paired data as input...?

mblue9:
Yes that would be great if MultiQC could work with paired collections and FastQC! I ran into this issue aswell.

bernt-matthias:
Do you know why it does not accept lists of pairs as input? Is there some thing wrong in the tool xml form description?

mvdbeek:
fastqc (the software, not the wrapper) is not aware of paired end datasets.

bernt-matthias:
Thanks for the information. 

What I do not understand is why the tool form does not list the available list of pairs as input. Would this need an additional parameter in the `<param>` tag? It might be reasonable because I guess that also the `<command>` part would need changes for iterating over the list of pairs. 

For the application I think it does not matter that FastQC does not know about paired data sets. One could treat them separately with FastQC. This is also what actually happens: a list of pairs of fastq data is transformed by FastQC into a list of pairs of reports. So I would suggest to modify the MultiQC tool such that also list of pairs are accepted and the result should be the same as if a joint list of single fastq files is used as input. The toolbox function of MultiQC allows to color by sample name or forward/reverse information (or any other info contained in the filename). If users prefer to analyze interleaved data they can still do so, but I just learned that forward and reverse read may differ in their qualities, i.e. it makes sense to treat them separately.

mvdbeek:
I think http://multiqc.info/docs/#afterqc may be a good option to replace fastqc for this purpose.

mblue9:
Yes I'd love to see a Galaxy tool for AfterQC anyway and if helps get around this issue with FastQC then that would be great!

bernt-matthias:
Just a note: according to the readme of https://github.com/OpenGene/AfterQC afterqc has been reimplemented: https://github.com/OpenGene/fastp

bebatut:
Do you think we should handle the paired collection for FastQC module of MultiQC? And then only for FastQC?

bernt-matthias:
Thanks for reviving this issue @bebatut . 

[...]

Issue #1608, opened 2017-11-28
MultiQC - samtools flagstat weird output


mblue9:
Just tried MultiQC (v1.2.0) on samtools flagstat (v2.0.1) and while the "General Statistics" part looks ok, the "Samtools Flagstat" part does not, it has blue dots randomnly plotted (see screenshot below), wondering if anyones's got that to work or if it's a bug?
![screen shot 2017-11-28 at 4 23 01 pm](https://user-images.githubusercontent.com/5591821/33303652-7b9d4c34-d458-11e7-9099-6a4fdaa232cf.png)

mblue9:
This weird plot issue also happens in Galaxy Freiburg with MultiQC 1.2.0 (it is pretty cool to have multiple Galaxies to use for testing!)

bebatut:
I have the same type of output

![screen shot 2018-04-09 at 20 55 27](https://user-images.githubusercontent.com/1842467/38516808-5edfafc2-3c38-11e8-9eb3-3b4dad8a0f07.png)

But it is not specific to MultiQC in Galaxy. I reproduce it by running MultiQC locally, so maybe a MultiQC issue

mblue9:
Thanks for checking this @bebatut! I'll close this issue so as it's not a wrapper issue.

Issue #1595, opened 2017-11-20
MultiQC - Use "element_identifier" as "sample name" for all tools


lparsons:
In order to take full advantage of MultiQC's aggregation report, it needs to find consistent sample names across various datasets. Thus, whenever possible, MultiQC should be using `element_identifier` as the "sample name". The way MultiQC parses the sample name varies for each tool. In some cases, the underlying tool must be updated to use the `element_identifier` (e.g. featureCounts https://github.com/galaxyproject/tools-iuc/pull/1583). This issue is intended to track the tools that do not properly parse the `element_identifier` when used with MultiQC. If someone can verify the status of a tool, please update this issue accordingly.

Verified tools (checked tools properly use `element_identifier`):
[x] fastqc
[ ] featureCounts - (tool is [fixed](https://github.com/galaxyproject/tools-iuc/pull/1583), pending Galaxy bug #???)
[x] rseqc
[ ] samtools
[x] star

Unknown status (the status of `element_identifier` parsing for these tools is unverified):
* bamtools
* bcftools
* bismark
* bowtie2
* busco
* cutadapt
* flexbar
* gatk
* hicup
* hisat2
* htseq
* kallisto
* picard
* prokka
* quast
* rsem
* salmon
* samblaster
* snpeff
* sortmerna
* tophat
* trimmomatic
* vcftools

mvdbeek:
Awesome, thanks for compiling the list!

mvdbeek:
@bebatut the element identifier is the only dataset attribute that should describe a sample identifier. It falls back to the `.name` attribute outside of collections. Tags are really problematic to track datasets, that is waht the element identifier is for.

bebatut:
Tools properly using `element_identifier` as sample name
- Bamtools
- Bismark
- Bowtie 2
- BUSCO
- GATK (BaseRecalibrator or VariantEval output)
- HISAT2
- HTSeq
- Samtools
- SnpEff
- STAR
- TopHat2: Sample name = `element_identifier` + `align`

Tools using some file content as sample name
- Bcftools: Sample name = 3rd column of the line starting by`ID` (https://github.com/ewels/MultiQC/blob/v1.5/multiqc/modules/bcftools/stats.py#L41)
- Cutadapt/Trim Galore!: Sample name = last name of line starting with `Command line` (https://github.com/ewels/MultiQC/blob/v1.5/multiqc/modules/cutadapt/cutadapt.py#L104)
[...]

PR #1821, opened 2018-04-08
MultiQC: Update + fixing issues

### LINKED ISSUE #1538: Update MultiQC to reject or report in error/help that HTML input is not valid
Users are reporting usage problems because they are inputting the HTML report from FastQC, not the Raw Data report, and sometimes both (!!).

Tool version 1.2.0: https://toolshed.g2.bx.psu.edu/view/iuc/multiqc/3bad335ccea9

Tutorials cover appropriate inputs here (and other places) but it doesn't seem to be enough: https://galaxyproject.org/tutorials/ngs/#assessing-data-quality

Could we consider these updates, with item 1 being the best (avoid this error case completely, making items 2 & 3 unnecessary, plus avoid wasting resources on a bad job that will eventually fail):

1. Trap when HTML input is given to the tool at runtime and aborting the submission with a red box warning near the input select stating "Inputs included HTML formatted reports. Input raw (plain text) reports instead." - OR - by filtering HTML datatypes out in the input select list of available datasets?

2. If item 1 above is not possible, improve the error message to state "Inputs included HTML formatted reports. Input raw (plain text) reports instead." ("Raw text" is the actual name given to the FastQC text version of the report - so just stating "plain text" is language that is not always interpreted by sci users well - and doesn't map to commonly input dataset names).

3. If item 1 above is not possible, state similar text as item 2 above _in the help section_ of the tool form, example "Input raw (plain text) reports only. HTML reports will trigger a tool failure".

### LINKED ISSUE #1542: MultiQC: Add module to error reports
It can be tricky to debug a large MultiQC run. Some error messages are hard to identify (e.g. `'Collect' not found in the file`). It would be very helpful if the error could contain the module that had the error (e.g. FastQC, Tophat2, etc.)

### LINKED ISSUE #1589: MultiQC - escape element_identifiers
`$element_identifier` should be escaped when used as filenames (see https://github.com/galaxyproject/tools-iuc/commit/b47bba4052185e45c39c5de35d0c42f7b3ea4a49). One potentital issue is that we'd like the "Sample Name" for each file to match up as much as possible so MultQC can connect the data (it produces some nice summaries).

### LINKED ISSUE #1624: MultiQC - error 'pytz' distribution was not found
After successful installation on MultiQC on Galaxy v17.09 it outputs error:

```
Fatal error: Exit code 1 ()
  File "/export/tool_deps/_conda/envs/__multiqc@1.2/bin/multiqc", line 4, in 
    __import__('pkg_resources').run_script('multiqc==1.2', 'multiqc')
  File "/export/tool_deps/_conda/envs/__multiqc@1.2/lib/python2.7/site-packages/pkg_resources/__init__.py", line 3144, in 
    @_call_aside
  File "/export/tool_deps/_conda/envs/__multiqc@1.2/lib/python2.7/site-packages/pkg_resources/__init__.py", line 3128, in _call_aside
    f(*args, **kwargs)
  File "/export/tool_deps/_conda/envs/__multiqc@1.2/lib/python2.7/site-packages/pkg_resources/__init__.py", line 3157, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/export/tool_deps/_conda/envs/__multiqc@1.2/lib/python2.7/site-packages/pkg_resources/__init__.py", line 666, in _build_master
    ws.require(__requires__)
  File "/export/tool_deps/_conda/envs/__multiqc@1.2/lib/python2.7/site-packages/pkg_resources/__init__.py", line 984, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/export/tool_deps/_conda/envs/__multiqc@1.2/lib/python2.7/site-packages/pkg_resources/__init__.py", line 870, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'pytz' distribution was not found and is required by matplotlib
```

### LINKED ISSUE #1820: MultiQC: does it work with HISAT2/STAR/any-RNA-seq-mapper and collections?
Wondering if anyone here is using MultiQC with HISAT2? As I can't get it to work e.g. if I pass in a collection of HISAT2 (v2.1.0) summary files to MultiQC (v1.3) I just get an "empty" green output :( Would be great to be able to summarise the HISAT2 stats for collections of samples if possible. Or if not HISAT2, then STAR or another mapper, I'd take anything that works at this stage. If anyone's successfully summarising RNA-seq QC stats (mapped reads etc) for collections with MultiQC with any mapper please let me know!

bebatut:
Solved with this PR:
- [X] Update to version 1.5
- [X] Add new supported tools
- [X] Add snpeff support as started in #1805
- [X] Add error if FastQC HTML report is submitted, to fix #1538 
- [X] Escape element_identifiers, to fix #1589
- [x] Support collection, to fix #1820 
- [X] Add module to error reports, to fix #1542 
- [X] Solve the error 'pytz' distribution was not found, to fix #1624 (Thanks @lparsons for the check)

Not solved with this PR:
- Support list of pairs for FastQC output (#1658): technical limitation
- Fix samtools flagstat weird output (#1608): it seems to be a MultiQC error
- Use "element_identifier" as "sample name" for all tools (#1595)
   - Tools properly using `element_identifier` as sample name
       - Bamtools
       - Bismark
       - Bowtie 2
       - BUSCO
       - GATK (BaseRecalibrator or VariantEval output)
       - HISAT2
       - HTSeq
       - Samtools
       - SnpEff
       - STAR
       - TopHat2: Sample name = `element_identifier` + `align`
    - Tools using some file content as sample name
       - Bcftools: Sample name = 3rd column of the line starting by`ID` (https://github.com/ewels/MultiQC/blob/v1.5/multiqc/modules/bcftools/stats.py#L41)
       - Cutadapt/Trim Galore!: Sample name = last name of line starting with `Command line` (https://github.com/ewels/MultiQC/blob/v1.5/multiqc/modules/cutadapt/cutadapt.py#L104)
       - deepTools: Sample names = `element_identifier`(s) on which deepTools has been run
       - FastQC: Sample name = 2nd column of line starting with `Filename` 
Unknown status (the status of `element_identifier` parsing for these tools is unverified)
       - featureCounts: Sample names = column names with the `element_identifier`(s) on which featureCounts has been run
       - HiCExplorer: Sample name = concatenation of the path + 2nd column of the line starting with `File`
       - Flexbar: Sample name = 2nd column of line starting with `Filename`
       - Kallisto: Sample name = 2nd column of line starting with `Run File` (looks like a log output needs to be added to the kallisto wrapper to work with multiqc, see here https://github.com/ewels/MultiQC/issues/440)
       - MACS2: multiqc can work with macs2 .xls, could try to add to multiqc wrapper, see here: https://github.com/ewels/MultiQC/issues/632
       - Picard: Sample name = value after `INPUT` on the line starting with `# picard.analysis`
       - Prokka: Sample name = what is after the first 2 words on the line starting with `organism`
       - QUAST: Sample names = column names with the `element_identifier`(s) on which QUAST has been run 
       - RSeQC: Gene Body Coverage uses element_identifier the other modules (xmls) don't
       - Salmon: multiqc can work with salmon, user request here: https://biostar.usegalaxy.org/p/27537/#27547
       - Samblaster
       - SortMeRNA: Sample name = extracted from the command line in the input file
       - Trimmomatic: Sample name = extracted from the command line in the input file
       - VCFTools


lparsons:
@bebatut I can't replicate #1624 anymore, so I guess I'd say it's "fixed". Conda dependency resolution is black magic to me, and I suspect that something changed with an underlying package (and not with MultiQC packages) that both broke and then subsequently fixed things...  :man_shrugging: 

At any rate, this is some excellent work! Thanks.

nekrut:
See also https://github.com/galaxyproject/tools-iuc/pull/1805

bebatut:
@nekrut I added the SnpEff module too and some tests 😄

bebatut:
Thanks @lparsons for the check!

bebatut:
I think this PR is ready for review 😄

lparsons:
Fantastic work @bebatut! One thought I had was that for tools that put the input filename into the output (e.g. BCFTools, cutadapt, etc.) perhaps it would be a good "best practice" to first link the input files using the `element_identifier`. That would be helpful not only when MultiQC parses the files, but when the user is looking at them too. @bgruening any thoughts?

bgruening:
@lparsons you mean linking inside the cutadapt wrapper?

lparsons:
@bgruening Yes, within the wrapper of tools that include the input filenames as part of the output, link the input files first to a name that includes the `element_identifier` so they output files look a bit nicer and are more interpretable. We already do that for some tools, but we could include it as "best practice".

bgruening:
I feel a little bit if this is really needed and a best practice it should be part of the job staging and Galaxy should provide a variable that we can simply use. So Galaxy does the linking for us and a tool can use it.

We should move this discussion to an other issue I guess. If you are fine with the PR please approve and/or merge it.

mblue9:
Thanks a lot for working on this @bebatut I'm excited to use it!
I tested a few of the modules with collections and they seem to work well e.g. fastqc, hisat2, featurecounts so looks good to me 👍 
@lparsons I'll try to get the cutadapt wrapper working properly with element_identifier, as you're right, would be good to have that, as otherwise cutadapt can't be easily used with multiqc (samples get called meaningless names e.g. "input_f")

tools/multiqc/multiqc.xml
```diff
@@ -411,15 +511,18 @@ sp:
                     <!--<option value="salmon">Salmon</option>-->
```
mblue9: Remove the module names from RSeQC here? "(bam_stat, gene_body_coverage, infer_experiment, ...)"
as why list them for rseqc and then not picard etc? I don't think they're needed here imho.

tools/multiqc/multiqc.xml
```diff
@@ -647,6 +769,9 @@ sp:
                 </when>
             </conditional>
         </repeat>
+        <param name="title" type="text" value="" optional="true" label="Report title" help="It is rinted as page header"/>
```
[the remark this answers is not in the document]
bebatut: Fixed

tools/multiqc/multiqc.xml
bebatut: Done... Thanks

bebatut:
@mblue9 Thanks for having tested the title and comment. I fixed the issue...

mblue9:
Thanks a lot @bebatut ! LGTM Will I merge it?

bebatut:
This isuse is probably not limited to HiCExplorer module. We probably need to fix that for all modules by adding a sub folder for each file (if we want to keep the file identifier intact)

mblue9:
>HiCExplorer does not work if you use files with identical name in Galaxy.

Just to check as am curious, are there use cases for analysing files with identical names? As that doesn't seem like a good idea to me, wouldn't it be better to recommend using unique file names? (saying that I do have the current situation where macs gives the same name to all files if using a control with collections which is not good imo)

joachimwolff:
I appreciate that this is a special case. It happens e.g. if you run a workflow and all QC reports are collected in the end, if they have an identical Galaxy name, MultiQC crashes. 

It is quite easy to fix it, just add an unique prefix/suffix to each file and we are more stable.

lparsons:
@joachimwolff My concern would be ensuring that the sample identifiers remain the unadulterated `element_identifier` across all tools. I'd be more in favor of some sort of check in the tool xml for uniqueness first. Not sure if that's doable with existing validators, but perhaps adding a new validator or falling back to coding it into the wrapper.

bgruening:
@joachimwolff that is a bigger issue as we want to preserver the `element_identifier`.
@lparsons yes this is doable, I discussed this with @bebatut already.

As this was "broken" before and currently is I don't consider this as blocking and will merge now as it contains so many other useful fixes.

Thanks @bebatut!

lparsons:
@bebatut Fantastic work on MultiQC, I find it a very useful tool. I was wondering if you had any insight as to why Salmon is commented out. I was looking into enabling it in the Galaxy wrapper but thought I'd see if there was already some work done... Thanks!

Issue #1634, opened 2017-12-14
Do not ignore the hisat2 exit code.


jmchilton:
I'm not sure why it is being ignored - seems odd. https://github.com/galaxyproject/tools-iuc/blob/master/tools/hisat2/hisat2.xml

If there is good reason to ignore it - we should then update the stdio regexes to fail a job is "out of memory" is detected. https://github.com/galaxyproject/galaxy/pull/5196#issuecomment-351003278

PR #1823, opened 2018-04-10
Use pipefail when piping into samtools

mvdbeek:
This should fix https://github.com/galaxyproject/tools-iuc/issues/1733 and https://github.com/galaxyproject/tools-iuc/issues/1634.

### DIFF
--- a/tools/bowtie2/bowtie2_wrapper.xml
+++ b/tools/bowtie2/bowtie2_wrapper.xml
@@ -10,7 +10,7 @@
     <version_command>bowtie2 --version</version_command>
     <command detect_errors="exit_code"><![CDATA[
 ## Use pipefail if available to quit with first non-zero exit code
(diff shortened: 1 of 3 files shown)


tools/bowtie2/bowtie2_wrapper.xml
```diff
@@ -9,6 +9,8 @@
     </requirements>
     <version_command>bowtie2 --version</version_command>
     <command detect_errors="exit_code"><![CDATA[
+## Use pipefail if available to quit with first non-zero exit code
+set -o | grep pipefail && set -o pipefail;
```
nsoranzo: What about adding ` -q` after `grep`?

mvdbeek: good point, I just added that

blankenberg:
Thanks!

PR #1828, opened 2018-04-14
Cutadapt: Add element_identifier for MultiQC

mblue9:
- use element_identifer in input filename (@lparsons suggestion here: https://github.com/galaxyproject/tools-iuc/pull/1821#issuecomment-381175523)
- add mention of MultiQC to report output help

### DIFF
--- a/tools/cutadapt/cutadapt.xml
+++ b/tools/cutadapt/cutadapt.xml
@@ -142,51 +142,7 @@
         #if $output_options.report:
             > report.txt
         #end if
-        ; cp '$out_config' galaxy.json
     ]]></command>
-    <configfiles>
-      <configfile name="out_config">{
-#set out1_format = "fastq"
-#set out2_format = "fastq"
-#set library_type = str($library.type)
-#if $library_type == 'paired':
-#set out1_format = $library.input_1.ext
-#set out2_format = $library.input_2.ext
-#else if $library_type == 'paired_collection'
-#set out1_format = $library.input_1.forward.ext
-#set out2_format = $library.input_1.reverse.ext
-#else
-#set out1_format = $library.input_1.ext
-#end if
-    "out1": {
-        "ext": "$out1_format"
-    },
-    "out2": {
-        "ext": "$out2_format"
-    },
-    "rest_output": {
-        "ext": "$out1_format"
-    },
-    "untrimmed_output": {
-        "ext": "$out1_format"
-    },
-    "untrimmed_paired_output": {
-        "ext": "$out2_format"
-    },
-    "too_short_output": {
-        "ext": "$out1_format"
-    },
-    "too_short_paired_output": {
-        "ext": "$out2_format"
-    },
-    "too_long_output": {
-        "ext": "$out1_format"
-    },
-    "too_long_paired_output": {
-        "ext": "$out2_format"
-    }
-}</configfile>
-    </configfiles>
     <inputs>
 
         <!-- Reads -->
@@ -269,9 +225,12 @@
     </inputs>
 
     <outputs>
-        <data name="out1" format="auto" metadata_source="input_1" from_work_dir="out1*" label="${tool.name} on ${on_string}: Read 1 Output"/>
-        <data name="out2" format="auto" metadata_source="input_2" from_work_dir="out2*" label="${tool.name} on ${on_string}: Read 2 Output" >
+        <data name="out1" format="fasta" metadata_source="input_1" from_work_dir="out1*" label="${tool.name} on ${on_string}: Read 1 Output">
+            <expand macro="inherit_format_1" />
+        </data>
+        <data name="out2" format="fastqsanger" metadata_source="input_2" from_work_dir="out2*" label="${tool.name} on ${on_string}: Read 2 Output" >
             <filter>(library['type'] == 'paired' or library['type'] == 'paired_collection')</filter>
+            <expand macro="inherit_format_2" />
         </data>
 
         <data name="report" format="txt" from_work_dir="report.txt" label="${tool.name} on ${on_string}: Report">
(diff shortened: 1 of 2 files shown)


tools/cutadapt/cutadapt.xml
```diff
@@ -13,104 +13,98 @@
...
+                #set read2 = $read2 + ".fq"
                 #set out2 = "out2.fq"
             #end if
-            ln -f -s '${library.input_2}' ${read2} &&
-
+            ln -f -s '${library.input_2}' $read2 &&
 
         #else if str($library.type) == 'paired_collection':
 
+            #set read2 = re.sub('[^\w\-\s]', '_', str($library.input_2.element_identifier))
```
nsoranzo: #set read1 = re.sub('[^\w\-\s]', '_', str($library.input_1.forward.element_identifier))

tools/cutadapt/cutadapt.xml
```diff
@@ -13,104 +13,98 @@
...
-                #set read1 = "input_f.fasta"
+                #set read1 = $read1 + ".fa"
                 #set out1 = "out1.fa"
             #else:
-                #set read1 = "input_f.fastq"
+                #set read1 = $read1 + ".fq"
                 #set out1 = "out1.fq"
             #end if
-            ln -s '${library.input_1.forward}' ${read1} &&
+            ln -s '${library.input_1.forward}' $read1 &&
```
nsoranzo: Add:
```
#set read2 = re.sub('[^\w\-\s]', '_', str($library.input_1.reverse.element_identifier))
```

tools/cutadapt/cutadapt.xml
```diff
@@ -13,104 +13,98 @@
 
         ## Link in the input and output files, so Cutadapt can tell their type
 
-        #set compressed="False"
+        #import re
         #set format = "fastq"
+        #set read1 = re.sub('[^\w\-\s]', '_', str($library.input_1.element_identifier))
```
nsoranzo: Does this actually work when `$library.input_1` is a paired collection? We probably need a test for this.

tools/cutadapt/cutadapt.xml
```diff
@@ -13,104 +13,98 @@
...
-                #set read1 = "input_f.fasta"
+                #set read1 = $read1 + ".fa"
                 #set out1 = "out1.fa"
             #else:
-                #set read1 = "input_f.fastq"
+                #set read1 = $read1 + ".fq"
                 #set out1 = "out1.fq"
             #end if
-            ln -f -s '${library.input_1}' ${read1} &&
+            ln -f -s '${library.input_1}' $read1 &&
```
bgruening: Jupp that is how I imagined it :)

tools/cutadapt/cutadapt.xml
mblue9: Good point @nsoranzo! I don't know that it will work, this was just being hopeful and not based on any testing. I've removed the paired collections now as I couldn't see how to get the outputs working and don't have time to work on that atm. (I'm also not using paired collections atm as I don't think they'll work with multiqc will they.)

tools/cutadapt/cutadapt.xml
mblue9: removed, see comment below

tools/cutadapt/cutadapt.xml
mblue9: removed, see comment below

tools/cutadapt/cutadapt.xml
mblue9: cool 😄

tools/cutadapt/cutadapt.xml
nsoranzo: Wait, that seems a bit extreme, I think cutadapt was working on paired collections before this PR, it was just missing a test for it.
And cutadapt can be used indepently of MultiQC.

mblue9: Well I got as far as adding your changes and trying below, which didn't work and I really really don't have time to mess around with this atm. If you can see what I'm doing wrong though I'll add it. Do I need to output out1 and out2 in a collection?
```
     <!-- Ensure paired collection works -->
    <test>
        <param name="type" value="paired" />
        <param name="input_1">
            <collection type="paired">
                <element name="forward" ftype="fastq.gz" value="bwa-mem-fastq1.fq.gz" />
                <element name="reverse" ftype="fastq.gz" value="bwa-mem-fastq2.fq.gz" />
            </collection>
        </param>
        <param name="adapter_source_list" value="user"/>
        <param name="adapter" value="AGATCGGAAGAGC"/>
        <param name="adapter_source_list2" value="user"/>
        <param name="adapter2" value="AGATCGGAAGAGC"/>
        <output name="out1" decompress="True" file="cutadapt_out1.fq.gz"/>
        <output name="out2" decompress="True" file="cutadapt_out2.fq.gz"/>
    </test>
```

lparsons: From my experience, paired collections cause more problems that they are worth. I'd vote for pushing out an update that works, but is missing paired collection support. If someone wants to look into fixing paired collections, it would be a nice update.

mblue9: Totally agree with @lparsons, I've stopped using paired collections for the same reason. I'd also vote for pushing this through now and adding in paired collection support later.

jmchilton: People use them, I don't get a vote but I'm totally with @nsoranzo in spirit - we shouldn't break something that works to add an enhancement in. We point people at this repository as examples of best practices - I don't think just breaking this would be that. Is there something I can do in the framework to help here? I can try to take a look at the this PR tomorrow morning maybe...

mblue9: Thanks @jmchilton your opinion definitely counts!! To include paired collections here we need to get the test I pasted above (or similar) working, if you know how to do that that would be really helpful. 

From my perspective this PR is not an enhancement, it's a critical usability fix, as I can't use cutadapt with multiqc (or teach it) until this PR is merged, because of the meaningless "input_f" file naming. So if that test for paired collections can't be added by this weekend I still vote to push this through for the moment without paired collections and add in paired collections later.

jmchilton: @mblue9 Thanks for bringing this up and explaining the problem. I've opened a PR against your branch (https://github.com/mblue9/tools-iuc/pull/2) that has two commits. The first restores the existing functionality - allowing paired collections with no use of the element identifier in the file name. I think the only problem with your test was that ``<param name="type" value="paired" />`` needed to be ``<param name="type" value="paired_collection" />``.

> From my perspective this PR is not an enhancement, it's a critical usability fix, as I can't use cutadapt with multiqc (or teach it) until this PR is merged, because of the meaningless "input_f" file naming. 

I mean it has to be useful to someone right? It got this far? I get that MultiQC is a popular tool and I'm excited people are working toward getting it to work within Galaxy but I feel like this wrapper vastly predates it and that I've seen in important pipelines for a long time.

> And just to add, I don't know if @lparsons has other reasons for not using paired collections, but my reason for stopping using paired collections was due to multiqc, getting samples called "forward" and "reverse". If there's a way to use the element_identifier instead I'd be happy to try using paired collections again.

The second commit I think should fix this - if multiqc reads this information from some report that depends on the filename here then I think the code I've included should set the filename to the element identifier of the list element containing the pair. This should be more robust than feeding two separate lists since the researcher can be more confident the lists are of the same length and corresponding elements are paired properly - basically it is harder to imagine things getting out of order for instance or unmatched - from the collection builder all the way down. Hopefully less visual clutter in the history as well.

mblue9: Thanks for working on the paired collections @jmchilton !!! I merged what you did (hopefully that was the right thing to do) as I think we're almost there, just need the output datatype to not output as "data" (see screenshot here: https://github.com/mblue9/tools-iuc/pull/2#issuecomment-383231387)

jmchilton:
@mblue9 - responding here since that comment has been collapsed by Github but this is an interesting discussion and others may want to weigh in on the approach. There were some other problems with output types also - format_source="txt" and such. I've opened a PR with https://github.com/mblue9/tools-iuc/pull/3 that replaces most of the ``format_source=`` calls with ``format="auto"`` which will instruct Galaxy to check galaxy.json for the format after the job is complete. The galaxy.json is generated as a config file as part of the job creation. This uses the newer version of galaxy.json that is expected in version 17.09+ of Galaxy so it adds a profile version to the tool of 17.09. You can support older versions of Galaxy by dropping the profile version and using the older version of galaxy.json that requires the output dataset id - https://github.com/galaxyproject/galaxy/blob/dev/test/functional/tools/tool_provided_metadata_1.xml. The PR also add ftype checks to the outputs - so Galaxy will verify data types generated as part of Planemo testing.

You asked another question somewhere about whether this should output a collection. I have an unpopular opinion about this - I think of tools as functions and workflows as programs - functions with multiple return types are gross to work with in general - you need to dispatch on the output type and such - and there is just no way to do that in the Galaxy workflow editor. So for tools like this - I'd argue it should be split into three tools with no conditional - one tool that takes in one output and produces one output, one tool that takes in two outputs and produces two outputs, and one tool that takes in a paired collection and produces a paired collection. Alternatively the first and third version of this would be enough IMO - if three tools is too many - I think from a use perspective "cutadapt - unpaired" vs "cutadapt - paired" isn't a terrible amount of overload in the tool panel. The format source hackery would go away, the conditional outputs would be vastly simpler, and the tool would work more robustly in the workflow editor. It isn't the IUC way however - so I think for now outputting two individual datasets - one of which happens to be conditional on the input structure is ok - if you need a collection output in a workflow or such - you can use the zip tool to restructure things.

bgruening:
> I think from a use perspective "cutadapt - unpaired" vs "cutadapt - paired" isn't a terrible amount of overload in the tool panel. 

Not for cutadapt, but we have >50 tools that fall into this category. Shouldn't we fix the workflow editor instead or the tool form?

nsoranzo:
Yep, duplicate tools is a no-go, sorry!

jmchilton:
> Shouldn't we fix the workflow editor instead or the tool form?

I'm all for trying to improve conditionals in workflows and conditional output structures but no one is pushing this or working on it - I'm skeptical we will be very successful in such an endeavor. 

The way differing output types are represented in the tool form right now can't really be remedied with a better workflow editor. We'd need to start by capturing that information in the tool descriptions in a more systematic way, then we would need to update the workflow editor (and a million things below it), etc.. So we'd need to update 50 tools anyway I think. And after all that we won't capture enough flexibility for everyone, it will all fall apart on conditional parameters and dynamically discovered tool stuff, etc.. Complex tools will always suck in a workflow language, we should be pushing for simple tools and complex workflow structures - it just isn't the Galaxy culture currently though.

Hopefully I'm proven wrong.

mblue9:
I've merged your PR @jmchilton. Thanks a lot for working on that! Especially as I would not have figured that out myself. And thanks for the info on collections. I agree with @bgruening and @nsoranzo, having duplicates of tools for paired/unpaired inputs would not be my preference, as already said, many tools would need it, and my concern would be that it might add clutter and confusion for users and add confusion and increase the maintenance burden for admins and tool devs. 

I'm happy to try cutadapt this like it is now but I agree that it would be better to have simple tools and my preference would be for this collections vs paired collections stuff to not be embedded into the tool form like it is currently, and instead for it to be abstracted, or to have some easier way to implement. As it really adds to the burden of wrapping tools (for me anyway) and I've avoided adding it to tools like fastp as I just didn't have the energy to figure how to incorporate the paired collections (and I'm not using fastp atm anyway as it doesn't work with multiqc yet).

bgruening:
@galaxyproject/iuc can someone please look at it. I feel manipulating the galaxy.json directly decreases our standards once again and makes development again harder. I don't want to block it but don't feel like merging it :(

@jmchilton; the tool cutadapt is not complicated, what it makes complicated is our handling of inputs, the file-extensions and so on. Collections have increased the development burden of tools to a large degree and we failed in making it more convenient or align it with the overall vision of the workflow editor.

I'm happy to organize an IUC call once 18.05 is out, if people think its worthwhile to discuss this.

jmchilton:
> we failed in making it more convenient or align it with the overall vision of the workflow editor.

That seems pretty rude and I really regret spending my weekend trying to help here. I’ve put a lot work both into the tool development process and into the workflow editor - both had many problems predating collections.

jmchilton:
galaxy.json have been a mistake - it might be possible to fix the output types with change format actions https://github.com/galaxyproject/galaxy/blob/dev/test/functional/tools/output_action_change_format.xml - that would certainly be better if it was. I’m not sure.

bgruening:
This was by no means an attack against you John. By no means. If you say `workflows and conditional output structures ... no one is pushing this or working on it ` than we need to change this or at least get a vision how to proceed here. If we can avoid it, we should not put the burden on the tool-developer (as here) or the UI-user (multiple tools etc ...).

I apologize if I sounded rude to you.

jmchilton:
I've pushed a commit to use output actions hacks instead of beautiful galaxy.json, I'll admit that despite the XML being yukky it is nice from a usability perspective that the formats are set at job start instead of job end.

mblue9:
Thanks a lot @jmchilton! Much appreciated!!

bgruening:
Thanks John and Maria!

PR #1829, opened 2018-04-15
Fix to umi_tools

mtekman:
* Added paired inputs for `extract`
 * Macrofied common options for `extract` and `whitelist`
 * Only `extract` wrapper version has been bumped due to added functionality.

### DIFF
--- a/tools/umi_tools/macros.xml
+++ b/tools/umi_tools/macros.xml
@@ -15,7 +15,7 @@
             </when>
         </conditional>
     </macro>
-    <xml name="macro_input_types">
+    <xml name="input_types">
         <conditional name="input_type">
             <param name="type" type="select" label="Library type">
                 <option value="single">Single-end</option>
--- a/tools/umi_tools/macros.xml
+++ b/tools/umi_tools/macros.xml
@@ -1,5 +1,41 @@
 <?xml version="1.0"?>
 <macros>
+    <macro name="barcode2_conditional" >
+        <conditional name="barcode">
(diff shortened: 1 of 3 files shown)


bgruening:
@mtekman can you please bump the version number.

bgruening:
error seems to be: `|  U.error("No local minima was accepted. Recommend checking the plot "`

mtekman:
@bgruening  - the wrapper version for the extract tool was bumped because it adds extra functionality (accepts paired inputs), but the whitelist tool has not changed because it was a fix rather than a feature. (i.e. the last release of whitelist had an error that went unnoticed but is now fixed).

Should I still bump it? I am worried that users might accidentally use the previous version of the tool

bgruening:
I guess its on the edge, it changes reproducibility ....
Can you look at the failing test?

tools/umi_tools/macros.xml
```diff
@@ -1,5 +1,41 @@
...
+            <when value="first_read_only"/>
+            <when value="both_reads">
+                <param name="bc_pattern2" argument="--bc-pattern2" type="text" value="" label="Barcode pattern for second read"
+                       help="Use this option to specify the format of the UMI/barcode for
+                             the second read pair if required." >
+                </param>
+            </when>
+        </conditional>
+    </macro>
+    <xml name="macro_input_types">
```
bgruening: I would also not call it macro, as it is part of the macro file.

bgruening:
Thanks @mtekman

PR #1830, opened 2018-04-15
Limma-voom: Update limma version and add option to output Rscript

mblue9:


mblue9:
I have updated the test data but it's failing here (as are the edgeR tests) due to tiny differences in floating point numbers e.g. 0.457332061341022 vs 0.457332061341017. I could add compare="sim_size" to work around it? Or what would be a better way to deal with this?

nsoranzo:
I've had that problem too with rdock. You can use the test explained at https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-tests-test-output-assert-contents

mblue9:
Great, thanks @nsoranzo! I've changed to using assert_contents and the tests pass here now. Will go fix the edgeR ones next 😄

tools/limma_voom/limma_voom.xml
```diff
@@ -300,17 +308,25 @@ cp '$outReport.files_path'/*.tsv output_dir/
...
             <output name="outReport" >
                 <assert_contents>
                     <has_text text="Limma Analysis Output" />
                     <not_has_text text="RData" />
                 </assert_contents>
             </output>
-        </test>
-        <!-- Ensure annotation file input works -->
+       </test>
+   <!-- Ensure annotation file input works -->
```
nsoranzo: Align comment with `<test>`?

tools/limma_voom/limma_voom.xml
```diff
@@ -300,17 +308,25 @@ cp '$outReport.files_path'/*.tsv output_dir/
             </repeat>
             <param name="normalisationOption" value="TMM" />
             <output_collection name="outTables" count="2">
-                <element name="limma-voom_Mut-WT" ftype="tabular" file="limma-voom_Mut-WT.tsv" />
-                <element name="limma-voom_WT-Mut" ftype="tabular" file="limma-voom_WT-Mut.tsv" />
+                <element name="limma-voom_Mut-WT" ftype="tabular" >
+                    <assert_contents>
```
nsoranzo: As for the edgeR PR, what about adding a `has_text_matching` also for the header line?

mblue9: The annotated output also wasn't working for limma-trend but fixed now. Thanks @nsoranzo !

PR #1831, opened 2018-04-15
EdgeR: Minor fixes and add option to output Rscript

mblue9:


nsoranzo:
It seems the small differences in floats is a problem here too.

tools/edger/edger.xml
```diff
@@ -278,8 +287,16 @@ cp '$outReport.files_path'/*.tsv output_dir/
             </repeat>
             <param name="normalisationOption" value="TMM" />
             <output_collection name="outTables" count="2">
-                <element name="edgeR_Mut-WT" ftype="tabular" file="edgeR_Mut-WT.tsv" />
-                <element name="edgeR_WT-Mut" ftype="tabular" file="edgeR_WT-Mut.tsv" />
+                <element name="edgeR_Mut-WT" ftype="tabular" >
+                    <assert_contents>
```
nsoranzo: What about matching also the header line (for each tabular file)?

nsoranzo: Can you add `<has_text_matching expression="GeneID.*logFC.*logCPM.*F.*PValue.*FDR" />` here too?

mblue9: done

bgruening:
Thanks @mblue9!

PR #1835, opened 2018-04-19
MAGeCK: Add single quotes to filenames used for sample-label

mblue9:
The filenames used in the sample labels weren't single quoted so have fixed that here. It was also causing mageck count to fail outright if you input files that had spaces in the name e.g "Tool on data N".


nsoranzo:
Thanks @mblue9!

PR #1842, opened 2018-04-21
MAGeCK: replace spaces in element_identifier with _

mblue9:
I found that mageck count really doesn't like spaces in filenames (e.g. "Tool on data N") as, even after adding the single quotes to the filenames and labels passed in on the command line, it runs fine but blank pdfs are produced instead of visible plots. The filename with spaces is also output in the counts files column header and if you try to read that into R it doesn't like it. So I've replaced spaces in the element_identifier with _ to fix i.e. I changed
`#set name = re.sub('[^\w\-\s]', '_', str($sample.element_identifier))`
to
`#set name = re.sub('[^\w\-]', '_', str($sample.element_identifier))`

I'm not sure about this one - I've also removed numpy as a requirement as 1) it's in the bioconda package and 2) my attempt to output the version for the `version_command` was giving an ugly error in every output file in the history (`/tool_script.sh: line 89: [python: command not found`). Or I guess the error could be fixed if it's better to record the numpy version.


tools/mageck/mageck_count.xml
```diff
@@ -17,7 +17,7 @@
 
     #for $sample in $reads.sample:
 
-        #set name = re.sub('[^\w\-\s]', '_', str($sample.element_identifier))
+        #set name = re.sub('[^\w\-]', '_', str($sample.element_identifier))
```
nsoranzo: If you add single quotes around `$infile` at line 30, you don't need this.

bgruening:
@nsoranzo is this ok?

mblue9:
Thanks for the help with this! Afraid the proposed single quotes fix doesn't seem to work though, the spaces in the name still lead to a blank pdf and spaces in the normalised counts header that will cause issues with generating the report in R. If I'm not allowed to remove spaces to help users avoid these issues, as an alternative then, I've added this line to the Help section "If you wish to generate the PDF plots or report ensure your input files do not have spaces in the dataset names, for example, "Cutadapt on data N" won't work, change to e.g. Cutadapt_on_data_N." Hopefully they won't have too many files to rename one-by-one.

mblue9:
This has a weird (to me) new error here.  `mageck mle` is complaining about some parameters not having a value but they're not used in the test and this passed fine previously here and locally, but is also now failing for me locally with 18.05, error is: 
`RunToolException: Error creating a job for these tool inputs - Parameter sgrnaeff_name_col requires a value, but has no legal values defined., Parameter sgrnaeff_score_col requires a value, but has no legal values defined.` 

(also `mageck test`'s 2 tests are failing due to different outputs found but they pass fine locally. I guess I can change to using the `has_text_matching` to work around that)

mvdbeek:
That sounds suspiciously like a bug we fixed in https://github.com/galaxyproject/galaxy/pull/6222.
This is already in 18.05 and the master branch, so it really should have worked :/

mvdbeek:
This time bisect points at https://github.com/galaxyproject/galaxy/commit/9febc519944dcfb5c8907b8a302fd6aff6f0f79f

nsoranzo:
> In this instance that's a bit silly and quite obvious, but maybe the better option from a usability perspective is to have these 3 options (the file and the coumns to choose from) in a conditional ?

I had basically the same suggestion (along with removing `optional="true"` from the 3 params) in a comment that somehow was not submitted, that seems the best option to me.

mblue9:
Thanks @mvdbeek and @nsoranzo! I've moved those params into a conditional now.

mblue9:
>In this case then it's probably better to replace the spaces then ask the user to do it manually.

Ok great, I've reverted back to that now. Thanks for the review @nsoranzo!

PR #1847, opened 2018-04-24
MultiQC: Fix error for input files with same identifier

bebatut:
- Keep all input path, check if path exists, add index if path already used
- Add test with 2 inputs with same name


bebatut:
@joachimwolff can you check if this PR solves your issue? Thanks

joachimwolff:
Works for me, thanks a lot for fixing.

bgruening:
Thanks!

PR #1852, opened 2018-04-26
Define and use @STD_BEDTOOLS_INPUTS@ token / allow bedgraph in most tools

mvdbeek:
Most bedtools tools take a standard set of bed/bedgraph/gff/vcf interval files. Using this token we can control this set in a single place.

In addition this commit allows bedgraph input in most commands without additional conversion and adds non-bed inputs in groupbybed, multicov, multiintersectbed and sortbed which all support these
additional formats (they can also take gzipped input ... if we'd want to toggle this in Galaxy).

### DIFF
--- a/tools/bedtools/multiCov.xml
+++ b/tools/bedtools/multiCov.xml
@@ -30,7 +30,7 @@
 ]]>
     </command>
     <inputs>
-        <param name="input" format="bed" type="data" label="Sorted BED file" />
+        <param name="input" format="@STD_BEDTOOLS_INPUTS@" type="data" label="Sorted BED file" />
         <!-- Additional files, if the user needs more -->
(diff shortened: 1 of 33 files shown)


tools/bedtools/multiCov.xml
```diff
@@ -30,7 +30,7 @@
 ]]>
     </command>
     <inputs>
-        <param name="input" format="bed" type="data" label="Sorted BED file" />
+        <param name="input" format="@STD_BEDTOOLS_INPUTS@" type="data" label="Sorted BED file" />
```
bgruening: I guess we should also marcofy the label?

bgruening:
Thanks @mvdbeek!

PR #1854, opened 2018-04-27
HiCExplorer update

joachimwolff:


joachimwolff:
I hope your changes @bgruening are not overwritten?

PR #1857, opened 2018-04-29
fix htseq

bgruening:
The old filter did not work and Galaxy produced an empty dataset, which Galaxy was trying to index and crashed (probably also a Galaxy bug).

But this should solve the htseq tool at least.


mvdbeek:
Empty, i.e 0-sized BAM files should be tool errors IMO, so that sounds correct. BAM files without any sequence would be OK, but that's not what was happening here.

PR #1858, opened 2018-04-30
fix ggplot2 outputs

lecorguille:
A little confusion about outputs filename

heatmap2.xml
```xml
<command>
mv output_plot.pdf $output1
</command>
<configfiles>
    <configfile name="script">
ggsave(file='output_plot.pdf', width=gg_width, height=gg_height)
    </configfile>
</configfiles>
<outputs>
    <data name="output1" format="pdf" from_work_dir="Rplot.pdf"/>
</outputs>
```

### DIFF
--- a/tools/ggplot2/heatmap2.xml
+++ b/tools/ggplot2/heatmap2.xml
@@ -8,8 +8,7 @@
     </expand>
     <command detect_errors="exit_code"><![CDATA[
 cat '$script' &&
-Rscript '$script' &&
-mv output_plot.pdf $output1
+Rscript '$script'
     ]]></command>
     <configfiles>
         <configfile name="script"><![CDATA[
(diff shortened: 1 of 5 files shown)


nsoranzo:
This is going to conflict with https://github.com/galaxyproject/tools-iuc/pull/1426 , can you check if that would solve the issue?

bgruening:
I fixed in the meantime the conda issue.

tools/ggplot2/heatmap2.xml
```diff
@@ -64,8 +63,7 @@ hclustfun=hclust_fun
   clabs = FALSE
 #end if
 
-pdf(file='output_plot.pdf')
-colorscale
+pdf(file='$output1')
```
bgruening: I think we need to have here the pdf file ending.

bernt-matthias:
Apparently this has been solved elsewhere

https://github.com/galaxyproject/tools-iuc/blob/43f3e91f557114060e6efc69e1d4ea711bb1ea96/tools/ggplot2/macros.xml#L280
https://github.com/galaxyproject/tools-iuc/blob/43f3e91f557114060e6efc69e1d4ea711bb1ea96/tools/ggplot2/macros.xml#L79

PR #1865, opened 2018-05-03
Fix default_identifier_source in macs2 to include conditional

mvdbeek:
Sorry that I didn't notice this right away @mblue9 !
I tested this a couple of times and this works correctly for me now.
Without the conditional prefix we're not finding the source in https://github.com/galaxyproject/galaxy/blob/release_18.01/lib/galaxy/tools/execute.py#L228.

### DIFF
--- a/tools/macs2/macs2_callpeak.xml
+++ b/tools/macs2/macs2_callpeak.xml
@@ -243,34 +243,34 @@
     </inputs>
     <outputs>
         <!--callpeaks output-->
-        <data name="output_tabular" format="tabular" default_identifier_source="input_treatment_file" label="${tool.name} on ${on_string} (Peaks in tabular format)">
+        <data name="output_tabular" format="tabular" default_identifier_source="treatment|input_treatment_file" label="${tool.name} on ${on_string} (Peaks in tabular format)">
             <filter> outputs and 'peaks_tabular' in outputs</filter>
         </data>
-        <data name="output_broadpeaks" format="bed" from_work_dir="MACS2_peaks.broadPeak" default_identifier_source="input_treatment_file" label="${tool.name} on ${on_string} (broad Peaks)">
+        <data name="output_broadpeaks" format="bed" from_work_dir="MACS2_peaks.broadPeak" default_identifier_source="treatment|input_treatment_file" label="${tool.name} on ${on_string} (broad Peaks)">
             <filter>
             ((
               advanced_options['broad_options']['broad_options_selector'] == "broad"
             ))
             </filter>
         </data>
-        <data name="output_gappedpeaks" format="bed" from_work_dir="MACS2_peaks.gappedPeak" default_identifier_source="input_treatment_file" label="${tool.name} on ${on_string} (gapped Peaks)">
+        <data name="output_gappedpeaks" format="bed" from_work_dir="MACS2_peaks.gappedPeak" default_identifier_source="treatment|input_treatment_file" label="${tool.name} on ${on_string} (gapped Peaks)">
             <filter>
             ((
               advanced_options['broad_options']['broad_options_selector'] == "broad"
             ))
             </filter>
         </data>
-        <data name="output_narrowpeaks" format="bed" from_work_dir="MACS2_peaks.narrowPeak" default_identifier_source="input_treatment_file" label="${tool.name} on ${on_string} (narrow Peaks)">
+        <data name="output_narrowpeaks" format="bed" from_work_dir="MACS2_peaks.narrowPeak" default_identifier_source="treatment|input_treatment_file" label="${tool.name} on ${on_string} (narrow Peaks)">
             <filter>
             ((
               advanced_options['broad_options']['broad_options_selector'] == "nobroad"
             ))
             </filter>
         </data>
-        <data name="output_summits" format="bed" from_work_dir="MACS2_summits.bed" default_identifier_source="input_treatment_file" label="${tool.name} on ${on_string} (summits in BED)">
+        <data name="output_summits" format="bed" from_work_dir="MACS2_summits.bed" default_identifier_source="treatment|input_treatment_file" label="${tool.name} on ${on_string} (summits in BED)">
             <filter>outputs and 'summits' in outputs</filter>
         </data>
-        <data name="output_plot" format="pdf" from_work_dir="MACS2_model.pdf" default_identifier_source="input_treatment_file" label="${tool.name} on ${on_string} (plot)">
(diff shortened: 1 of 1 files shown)


mblue9:
Thanks a lot for working on this @mvdbeek !!

I just tested it and it works well with 1 treatment sample in a collection vs 1 control but if I try with 2 treatment files in a collection vs the 1 control I get a pop up with this error:

```
The server could not complete the request. Please contact the Galaxy Team if this error persists. Unknown error occurred while processing request.

{
    "tool_id": "macs2_callpeak",
    "tool_version": "2.1.1.20160309.3",
    "inputs": {
        "treatment|t_multi_select": "No",
        "treatment|input_treatment_file": {
            "values": [
                {
                    "hid": 4,
                    "id": "2891970512fa2d5a",
                    "keep": false,
                    "name": "treat",
                    "src": "hdca",
                    "tags": []
                }
            ],
            "batch": true
        },
        "control|c_select": "Yes",
        "control|c_multiple|c_multi_select": "No",
        "control|c_multiple|input_control_file": {
            "values": [
                {
                    "hid": 5,
                    "id": "5729865256bc2525",
                    "keep": false,
                    "name": "control",
                    "src": "hdca",
                    "tags": []
                }
            ],
            "batch": true
        },
        "format": "BED",
        "effective_genome_size_options|effective_genome_size_options_selector": "2700000000",
        "nomodel_type|nomodel_type_selector": "create_model",
        "nomodel_type|mfold_lower": 5,
        "nomodel_type|mfold_upper": 50,
        "nomodel_type|band_width": 300,
        "cutoff_options|cutoff_options_selector": "qvalue",
        "cutoff_options|qvalue": "0.05",
        "outputs": null,
        "advanced_options|to_large": "false",
        "advanced_options|nolambda": "false",
        "advanced_options|ratio": "",
        "advanced_options|slocal": "",
        "advanced_options|llocal": "",
        "advanced_options|broad_options|broad_options_selector": "nobroad",
        "advanced_options|broad_options|call_summits": "false",
        "advanced_options|keep_dup_options|keep_dup_options_selector": "1"
    }
}
```

mvdbeek:
That is not specific to the default_identifier_source, right ? This fails here: https://github.com/galaxyproject/galaxy/blob/14259b11d9a5f6238153556ed5a28bf89468e2b4/lib/galaxy/util/permutations.py#L34-L33 which is supposed to make sure that the input parameters line up, while in this case we'd want to multiply the input. I think the workflow form allows you to do this, but not the tool form.
The workaround for now would be to use `merge collection` to merge the control file as many times as you need it, but I'll see if we can improve this in galaxy.

jmchilton:
Why is the one control in a collection? That is making me nervous, shouldn't it just be a single dataset if there is only one in this case?

mvdbeek:
That'd be another option, sure, but then you'd loose all the identifier goodness of having everything in collections.

jmchilton:
> you'd lose all the identifier goodness of having everything in collections and have everything streamlined up until this point

😭 indeed. I understand the impulse for sure. We do now have name tags as a workaround for single dataset identifier tracking that doesn't bring the added semantics of things being in a list (https://github.com/galaxyproject/galaxy/pull/6059#issuecomment-386580794). Is that helpful in these workflows?

These solutions are all messy, I'm not sure what the right answer is.

I don't love sticking things in singleton lists just for the identifier tracking, but if it is a reality we need to deal with should we develop tools for that use case? A collection operation tool that does [1] -> 1 might be helpful in this case? One can imagine other workflows where you have nested lists and you want to reduce the inner lists by finding the "best" dataset in some way and then apply this tool to build a flat list from the result Nx1 result of that? We could also do a collection operation tool for the inverse 1 -> [1].

mvdbeek:
> Is that helpful in these workflows?

It'd be a stop gap I think and fall apart where the control is derived though processing with other nametagged datasets, though better than nothing. 

> A collection operation tool that does [1] -> 1 might be helpful in this case? One can imagine other workflows where you have nested lists and you want to reduce the inner lists by finding the "best" dataset in some way and then apply this tool to build a flat list from the result Nx1 result of that?

I think that may be a better option than nametags. In the simple [1] -> 1 case, could we do this automatically if the n == n check fails and the input is [1] ?

> One can imagine other workflows where you have nested lists and you want to reduce the inner lists by finding the "best" dataset in some way and then apply this tool to build a flat list from the result Nx1 result of that

I can see that being useful, yes.

mvdbeek:
The other thing that'd work today is what I mentioned in https://github.com/galaxyproject/tools-iuc/pull/1865#issuecomment-386537501, I don't find this all that terrible since the user explicitly builds the list of treatment/control, which could also accomodate cases of 3 treatments 1 control, then 2 treatments 1 control, 5 treatments 1 control and so forth. That's fairly common in chip-seq I think.

jmchilton:
@mvdbeek Totally agree on this last point https://github.com/galaxyproject/tools-iuc/pull/1865#issuecomment-386537501, seems like a good direction to move in.

I'm thinking a "1 or [1]" -> [N] explicit collection operation tool might be more intuitive than merge collection operation though, do you agree with that? If yes, I'll create a Galaxy issue for adding these three singleton operation tools to Galaxy.

This tool could also take another list instead of an integer and use its size and identifiers to guide the new list creation. That would be usable in workflows too then. Hmm....

mvdbeek:
> I'm thinking a "1 or [1]" -> [N] explicit collection operation tool might be more intuitive than merge collection operation though, do you agree with that?

Oh yeah, absolutely. Sort of like 
```R
> rep('bla', 3)
[1] "bla" "bla" "bla"
```
in R ? That'd be so cool!

> This tool could also take another list instead of an integer and use its size and identifiers to guide the new list creation. That would be usable in workflows too then. Hmm....

so instead of N we'd use the length of collection x ? yeah, that'd be so helpful in workflows.

jmchilton:
> could we do this automatically if the n == n check fails and the input is [1] ?

I don't love that... I'm worried it will give users the wrong impression about what is happening under the hood (and it is hard to implement - kind of looks more complicated than implicit conversions and those are a bit of a mess both in the backend and in the user experience).

What about just a hint in the GUI - the GUI should be doing that n == n check anyway and if that fails we could have a little popup with different ideas (including explicit advice for the [1] case if that applies)?

mvdbeek:
> What about just a hint in the GUI - the GUI should be doing that n == n check anyway and if that fails we could have a little popup with different ideas

Actually we do do this check when building the form -- we have a nice 400 error, but we just seem to be ignoring that!

PR #1866, opened 2018-05-03
Integrate Pureclip into Galaxy + updating MEME (DREME, FIMO, MEME, meme_psp_gen) to 4.12.0 with new test files

michauhl:
Hello,
I wrote a wrapper for PureCLIP and also updated the MEME tools in Galaxy to the current version (4.12.0). For meme.xml, dreme.xml, fimo.xml, meme_psp_gen.xml I generated new test files (deleted most of the old ones), and rewrote some of the content (especially fimo.xml). In the case of FIMO I also deleted the wrapper script (fimo_wrapper.py) since it caused problems with the changed output format of 4.12.0 and was not easily fixable (at least for me). All the functionality should still be there (apart from outputting separate files for each motif, but you can call FIMO several times with option single motif output to achieve the same result).
Best,
Michael


peterjc:
This would probably be easier to review if you split the Pureclip and MEME work into two pull requests.

I'm not sure about the change to FIMO to only allow one output - I can check with local users of this tool. Making each output optional would be nicer I think.

tools/meme/macros.xml
```diff
@@ -1,10 +1,11 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <macros>
-    <token name="@WRAPPER_VERSION@">4.11.2</token>
+    <token name="@WRAPPER_VERSION@">4.12.0</token>
     <xml name="requirements">
         <requirements>
             <requirement type="package" version="1.3.23">graphicsmagick</requirement>
-            <requirement type="package" version="4.11.2">meme</requirement>
+            <requirement type="package" version="4.12.0">meme</requirement>
+            <yield/>
```
peterjc: What is this ``<yield/>`` line doing?

tools/meme/meme_psp_gen.xml
```diff
@@ -5,9 +5,11 @@
     </macros>
     <expand macro="requirements" />
     <command detect_errors="exit_code"><![CDATA[
+ln -s '${primary_sequence}' meme_psp_input_pos.fa && 
+ln -s '${control_sequence}' meme_psp_input_neg.fa &&
 psp-gen
--pos '$primary_sequence'
--neg '$control_sequence'
+-pos meme_psp_input_pos.fa
+-neg meme_psp_input_neg.fa
```
peterjc: This looks wrong, replacing a variable with a fixed value.

tools/meme/macros.xml
michauhl: And sorry for not splitting Pureclip and MEME, it is only my second pull request (I will keep it separate in the future).

tools/meme/meme_psp_gen.xml
peterjc: Ah, I understand now - using a symlink with a nice human readable filename - good idea 👍

tools/meme/meme_psp_gen.xml
```diff
@@ -78,14 +80,14 @@ $adv.raw
     </outputs>
     <tests>
         <test>
-            <param name="primary_sequence" value="meme_psp_protein_input.fasta" ftype="fasta"/>
-            <param name="control_sequence" value="meme_psp_protein_input.fasta" ftype="fasta"/>
+            <param name="primary_sequence" value="meme_psp_input_pos.fa" ftype="fasta"/>
```
peterjc: See discussion above about using a symlink called ``meme_psp_input_neg.fa`` internally. It would seem better not to rename the test input file now?

tools/pureclip/pureclip.xml
```diff
@@ -0,0 +1,318 @@
...
+    -bai target.bam.bai
+    -g genome.fa
+    #if $learn_params_contigs
+        -iv '$learn_params_contigs'
+    #end if
+    #if $apply_hmm_contigs
+        -iv '$apply_hmm_contigs'
+    #end if
+    -dm $merge_dist
+    #if str( $control_data.control_data_selector ) == 'supply_control_data':
```
bgruening: I don't think you need the str() cast here and below.

tools/pureclip/pureclip.xml
```diff
@@ -0,0 +1,318 @@
...
+            -g2kmin $advanced_params.g2kmin_value
+        #end if
+        #if $advanced_params.g2kmax_value
+            -g2kmax $advanced_params.g2kmax_value
+        #end if
+        $advanced_params.fk_value
+        -mkn $advanced_params.mkn_value
+        -mtp $advanced_params.mtp_value
+        #if $advanced_params.mk_value
+        -mk $advanced_params.mk_value
```
bgruening: indentation

tools/pureclip/pureclip.xml
```diff
@@ -0,0 +1,318 @@
...
+        #end if
+        #if $advanced_params.pa_value
+        -pa $advanced_params.pa_value
```
bgruening: indenation

tools/pureclip/pureclip.xml
```diff
@@ -0,0 +1,318 @@
...
+        #end if
+        $advanced_params.ea1_value
+        $advanced_params.ea2_value
+        $advanced_params.et1_value
+        $advanced_params.et2_value
+        #if $advanced_params.mrtf_value
+        -mrtf $advanced_params.mrtf_value
```
bgruening: indentation

tools/pureclip/pureclip.xml
```diff
@@ -0,0 +1,318 @@
...
+            <sanitizer>
+                <valid initial="string.printable">
+                    <remove value="&apos;"/>
+                </valid>
+            </sanitizer>
+        </param>
+        <param name="merge_dist" type="integer" value="8" min="1"
+               label="Distance used to merge individual crosslink sites to binding regions" argument="-dm"/>
+
+        <conditional name="control_data">
```
bgruening: A conditional with only one param in it, can usually be written without a conditional. Please consider to make `control_bam_file` just `optional=True` and remove the entire conditional.

tools/pureclip/pureclip.xml
```diff
@@ -0,0 +1,318 @@
...
+                <param name="b2p_value" type="float" value="0.15"
+                       label="Initial value for binomial probability parameter of 'crosslink' state" argument="-b2p"/>
+                <conditional name="antp_option">
+                    <param name="antp_option_selector" type="select" label="Choose n threshold for estimating crosslink state parameters" help="Either automatically choose n threshold (-ntp, -ntp2) to estimate parameters linked to crosslink states based on expected read start count at crosslink sites, or manually set values">
+                        <option value="antp_select" selected="true">Automatically choose n threshold (-ntp, -ntp2)</option>
+                        <option value="manual_select">Manually set -ntp, -ntp2</option>
+                    </param>
+                    <when value="antp_select" />
+                    <when value="manual_select">
+                        <param name="ntp_value" type="integer" value="10" argument="-ntp"
```
bgruening: using argument is great! If you use `$ntp` in your command section, you can even remove the name attribute.

bgruening:
@gregvonkuster could you please review this and let us know if the fimo/meme update looks good to you?

tools/meme/fimo.xml
```diff
@@ -1,135 +1,123 @@
...
-        --bgfile '${options_type.bgfile_type.bgfile}'
-    #end if
-    ${options_type.max_strand}
-    --max_stored_scores '${options_type.max_stored_scores}'
-    #if str($options_type.motifs_cond.motifs_selector) == 'no':
-        #for $motif in $options_type.motifs:
-            --motif '${motif.motif}'
-        #end for
-    #end if
-    --output_separate_motifs ${options_type.output_separate_motifs}
```
gregvonkuster: It looks like you are eliminating this option - is this no longer available with the latest version?

tools/meme/fimo.xml
```diff
@@ -1,135 +1,123 @@
...
-    --motif_pseudo '${options_type.motif_pseudo}'
-    ${options_type.no_qvalue}
-    ${options_type.norc}
-    #if str($options_type.parse_genomic_coord_cond.parse_genomic_coord) == 'yes':
-        --parse_genomic_coord 'yes'
-        --remove_duplicate_coords ${options_type.parse_genomic_coord_cond.remove_duplicate_coords}
```
gregvonkuster: Is this option no longer available in the new version?

tools/meme/fimo.xml
```diff
@@ -1,135 +1,123 @@
...
---txt_output '${txt_outfile}'
---xml_output '${xml_outfile}'
---gff_output '${gff_outfile}'
+    '$input_motifs'
+    '$input_fasta'
     ]]></command>
     <inputs>
-        <param name="input_motifs" type="data" format="memexml" label="'MEME output' formatted file"/>
-        <conditional name="fasta_type">
-            <param name="fasta_type_selector" type="select" label="Source for sequence to search">
```
gregvonkuster: This is a question on my part as I may be out of sync with the IUC on this.  Are we no longer using the ```from_data_table``` feature to allow multiple options like this for associating genome indexes?

gregvonkuster:
Just a few questions on the fimo changes - meme looks fine!

tools/meme/fimo.xml
michauhl: This was one of the options provided by the fimo_wrapper.py script. I guess one needs an additional script to do that, as it's not part of the FIMO options. You can specify the motifs you want to search for (motif_selector) to emulate the behavior (run FIMO n times for each of n motifs). But I'm not sure how frequently this old feature was used? If popular I might have to add an additional script.

tools/meme/fimo.xml
michauhl: remove_duplicate_coords was another option provided by the fimo_wrapper.py script. It removed identical motif occurrences in the .gff file (if same chromosome, start end and motif), using two "sort" command line calls. Again I'm not sure how often this was used (or if it's still necessary in 4.12.0), but it's not part of the FIMO options.

tools/meme/fimo.xml
gregvonkuster: I may have added this option and the one below (remove_duplicate_coords) a few years ago for one of my customers for whom I built the ChIP-exo pipeline.  If no one else finds these 2 parameters useful, feel free to merge without them.  I'm no longer working with that particular customer, and I'm not sure they'll ever upgrade their current environment to use this new version any way.

tools/meme/fimo.xml
gregvonkuster: If others are ok with no longer supporting the above feature, I'm ok with it too.

peterjc:
Earlier I wrote:

> I'm not sure about the change to FIMO to only allow one output - I can check with local users of this tool. Making each output optional would be nicer I think.

Local consensus is this would likely be an improvement - we use the tool in an in-house training course, and currently have ask users to delete 4 of the 5 the output files leaving just the "(text)" tabular output.

bgruening:
Meme looks good now, but for pureclip you are missing a .shed.yml file, so no test was running until now.

bgruening:
@michauhl thanks a lot! Thanks @peterjc and @gregvonkuster for the review.

PR #1870, opened 2018-05-04
RepeatMasker: small bug fixes

abretaud:
Hey,
I found some issues with the new RepeatMasker (thanks for updating it @pvanheus!), so here's PR to fix them.
I also added a (hacky) test case for the RepBase, with a fake embl file (just a tiny extract from the dfam embl file bundled with repeatmasker).

### DIFF
--- a/tools/repeatmasker/repeatmasker.xml
+++ b/tools/repeatmasker/repeatmasker.xml
@@ -1,4 +1,4 @@
-<tool id="repeatmasker_wrapper" name="RepeatMasker" version="4.0.7" profile="17.01">
+<tool id="repeatmasker_wrapper" name="RepeatMasker" version="4.0.7+galaxy1" profile="17.01">
   <description>RepeatMasker</description>
 
   <requirements>
@@ -11,7 +11,7 @@
     export REPEATMASKER_LIB_DIR=\$(pwd)/lib &&
       for file in \$(ls \$RM_LIB_PATH) ; do  ln -s \$RM_LIB_PATH/\$file lib/\$file ; done &&
     #if $repeat_source.source_type == "repbase":
-      cp '${repeat_source.repbase_file}' lib/RMRBSeqs.embl &&
+      cp '${repeat_source.repbase_file}' 'lib/${repeat_source.repbase_file_name}' &&
     #end if
     ln -s '${input_fasta}' rm_input.fasta &&
     RepeatMasker -dir \$(pwd)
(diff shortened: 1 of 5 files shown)


tools/repeatmasker/repeatmasker.xml
```diff
@@ -126,7 +127,7 @@
       </when>
     </conditional>
     <param type="boolean" argument="-gff" truevalue="-gff" falsevalue="" label="Output annotation of repeats in GFF format" checked="false" />
-    <param name="ignore_n_stretches" type="boolean" argument="-excln" falsevalue="" label="Ignore stretches of Ns when computing statistics" checked="true" help="Scaffolds are sometimes joined with stretches of 25 or more Ns. This option ignores them when calculating repeat statistics" />
+    <param name="ignore_n_stretches" type="boolean" truevalue="-excln" falsevalue="" label="Ignore stretches of Ns when computing statistics" checked="true" help="Scaffolds are sometimes joined with stretches of 25 or more Ns. This option ignores them when calculating repeat statistics" />
```
nsoranzo: Can you also keep `argument`?

nsoranzo:
Looks good, but the new test fails.

tools/repeatmasker/repeatmasker.xml
```diff
@@ -79,6 +79,7 @@
       </param>
       <when value="repbase">
         <param name="repbase_file" type="data" format="embl" label="RepBase (RMRBSeqs.embl) file" />
+        <param name="repbase_file_name" type="hidden" value="RMRBSeqs.embl"/> <!-- This is an ugly hack to allow testing with a fake repbase -->
```
pvanheus: I thought RMRBSeqs.embl was a magic name for RepeatMasker, thus the static name in the code. Does your testing suggest this is not the case?

tools/repeatmasker/repeatmasker.xml
```diff
@@ -145,8 +146,9 @@
         <option value="-s">Slow (0-5% more sensitive, 2.5 times slowdown)</option>
       </param>
       <param type="integer" argument="-frag" value="40000" label="Maximum contiguous sequence searched" help="Maximum length of sequencing that is search without fragmenting" />
-      <param type="integer" argument="-maxsize" value="4000000" label="Maximum length for IS or repeat clipped sequences" />
-      <param type="integer" argument="-gc" optional="True" label="Select matrices for this GC%" help="Valid values are a percentage or -1 to choose the default" />
+      <!-- -maxsize option is in the help, but not in the code of repeatmasker-->
```
pvanheus: Perhaps remove this (and the part in the <command> above rather than commenting it out?

pvanheus:
I left a few comments on the commits - I hope you can see them @abretaud . Otherwise I'll comment here.

tools/repeatmasker/repeatmasker.xml
abretaud: It's kind of magic, because there is another file named share/RepeatMasker/Libraries/RMRBMeta.embl which is needed when using RMRBSeqs.embl
There are some things (like species) defined in Meta, and the rest in Seqs
I'm not sure the embl file I wrote makes much sense, but at least it allows to check that it's running

abretaud:
I see them :) except one comment about the gff param, but maybe you deleted it?
For -maxsize, yes I could have removed it, but well... it's merged now, so it guess it can wait the next release

PR #1873, opened 2018-05-07
Limma-voom: Improvements to plots and add TREAT function

mblue9:
- Output Volcano plot for each contrast with top genes highlighted
- Output MDS plots for additional factors
- Output density plots for before and after filtering low counts
- Output box plots for before and after normalising counts
- Add TREAT function (Testing significance relative to a fold-change threshold)
- Other minor changes


bgruening:
Thanks @mblue9!

PR #1874, opened 2018-05-07
Update STAR to 2.6.0b and add a few of the newer options. --outBAMsor…

dpryan79:
…tingBinsN in particular is the preferred way to handle sorting memory requirements.

### DIFF
--- a/tools/rgrnastar/macros.xml
+++ b/tools/rgrnastar/macros.xml
@@ -1,7 +1,7 @@
 <macros>
     <xml name="requirements">
         <requirements>
-            <requirement type="package" version="2.5.2b">star</requirement>
+            <requirement type="package" version="2.6.0b">star</requirement>
             <requirement type="package" version="0.1.19">samtools</requirement>
         </requirements>
     </xml>
--- a/tools/rgrnastar/rg_rnaStar.xml
+++ b/tools/rgrnastar/rg_rnaStar.xml
@@ -1,4 +1,4 @@
-<tool id="rna_star" name="RNA STAR" version="2.5.2b-2" profile="17.01">
+<tool id="rna_star" name="RNA STAR" version="2.6.0b-1" profile="17.01">
     <description>Gapped-read mapper for RNA-seq data</description>
     <macros>
         <import>macros.xml</import>
@@ -94,6 +94,7 @@
             --outSAMattributes $output_params.outSAMattributes
             --outSAMstrandField $output_params.outSAMstrandField
             --outFilterIntronMotifs $output_params.outFilterIntronMotifs
+            --outFilterIntronStrands $output_params.outFilterIntronStrands
             #if str( $output_params.output_params2.output_select2 ) == "yes":
                 --outSAMunmapped $output_params.output_params2.outSAMunmapped
                 --outSAMprimaryFlag $output_params.output_params2.outSAMprimaryFlag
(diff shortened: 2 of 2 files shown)


tools/rgrnastar/macros.xml
```diff
@@ -1,7 +1,7 @@
 <macros>
     <xml name="requirements">
         <requirements>
-            <requirement type="package" version="2.5.2b">star</requirement>
+            <requirement type="package" version="2.6.0b">star</requirement>
             <requirement type="package" version="0.1.19">samtools</requirement>
```
bgruening: Do we want to use a newer samtools here as well?

dpryan79: Probably. If the current changes pass CI testing then I can bump that to 1.8

tools/rgrnastar/rg_rnaStar.xml
```diff
@@ -269,6 +276,10 @@
                     <option value="RemoveNoncanonical">Remove alignments with non-canonical junctions</option>
                     <option value="RemoveNoncanonicalUnannotated">Remove alignments with unannotated non-canonical junctions</option>
                 </param>
+                <param argument="--outFilterIntronStrands" type="select" label="Filter alignments containing junctions with inconsistent strands">
```
bgruening: Could this be a boolean with truevalue="--outFilterIntronStrands RemoveInconsistentStrands" falsevalue="--outFilterIntronStrands None"

dpryan79: It could, though I wouldn't be surprised if additional options get added to this in the future, so I'm doubtful of the benefit.

dpryan79:
I'll have to do some testing of this locally, there seem to be a couple issues.

dpryan79:
This seems to be working now.

bgruening:
> Given that it may be a good idea to simply raise the default to be higher than what STAR presets. The only down side to this is slightly slower performance and a bit higher temporary disk usage.

This point boils down to how much do we trust the users to take a sane decision about memory usage here. What we usually do is, we define some default as tool author, set that as bash default `${FOO:-BAR}` and let the admin overwrite it if wished and do not offer this option to the user at all. Like with SLOTS.

dpryan79:
Even I don't know a perfect value, users have no chance. The default that's in there now works OK for most genomes, but will fail about 30% of the time on fruit fly datasets.

dpryan79:
The genome is tiny, therefore the memory allocation is less. However the sequencing amount is rarely changed in parallel, so the ratio of BAM file size to genome size ends up much higher than typical mammalian RNAseq experiments, leading to the memory exhaustion issue. That's why I suggested perhaps raising the default in the wrapper higher than it is in the tool (I assume others receive error messages sent to them from users related to this too).

bgruening:
Please feel free to increase the value and maybe use the bash-magic from above to make it overwritable by an admin.

tools/rgrnastar/rg_rnaStar.xml
```diff
@@ -113,9 +113,9 @@
                 --outFilterMatchNminOverLread "$output_params.output_params2.outFilterMatchNminOverLread"
                 --outSAMmultNmax "$output_params.output_params2.outSAMmultNmax"
                 --outSAMtlen "$output_params.output_params2.outSAMtlen"
-                --outBAMsortingBinsN "$output_params.output_params2.outBAMsortingBinsN"
             #end if
         #end if
+        --outBAMsortingBinsN "$output_params.output_params2.outBAMsortingBinsN"
```
bgruening: The parameter is still in the conditional so this will most likely not work.

dpryan79: Could be, I'm using travis to test that :)

dpryan79:
Changing the default results in larger files, which is probably not worth it for 90+% of experiments. I'll just suggest people change this in their workflows when needed then.

PR #1875, opened 2018-05-07
Lorikeet spoligotyping wrapper

pvanheus:
Wrapper for the [lorikeet](https://github.com/AbeelLab/lorikeet) tool for digital spoligotyping (i.e. DNA fingerprinting) of *M. tuberculosis* sequences.

Currently only the `spoligotype` sub-tool from lorikeet is wrapped and its output is presented as text, not a new text subclass (i.e. a `spoligotype` format type). These are left as TODO items for the future.


tools/lorikeet/lorikeet.xml
```diff
@@ -0,0 +1,65 @@
...
+    <param name="input" label="Input sequences" type="data" format="bam" help="Reads of sample(s) being spoligotyped" />
+    <conditional name="spacer_source">
+      <param type="select" name="source_type" label="Source for spacer sequences">
+        <option selected="true" value="builtin">Built-in</option>
+        <option value="file">File with custom spacers</option>
+      </param>
+      <when value="builtin">
+      </when>
+      <when value="file">
+        <param argument="--spacer" name="spacer_file" type="data" format="fasta" label="Spacer file" help="FASTA format file containing customer spacer library" />
```
bgruening: I would remove the `name` here. And call it `-s '${spacer_source.spacer}'` above.

tools/lorikeet/lorikeet_macros.xml
```diff
@@ -0,0 +1,3 @@
+<macros>
```
bgruening: You can also include these small macros directly into the other xml file. e.g. https://github.com/galaxyproject/tools-iuc/blob/fd5d2211338239af6c577d66aeee3bb275c9b15b/tools/hicexplorer/hicPlotDistVsCounts.xml#L4

PR #1876, opened 2018-05-07
SpoTyping - spoligotyping tool

pvanheus:
Another tool for spoligotyping (DNA fingerprinting) M. tuberculosis sequence data


tools/spotyping/spotyping.xml
```diff
@@ -0,0 +1,104 @@
...
+    #set $input_file='input.' + $input.extension
+    ln -s '${input}' $input_file &&
+    SpoTyping.py
+    --noQuery
+    $advanced.seq
+    $advanced.swift
+    $advanced.filter
+    $advanced.sorted
+    $input_file &&
+    cat SpoTyping.log SpoTyping > '${output_txt}'
```
bgruening: Should the log not go to stdout?

pvanheus: Perhaps. The log shows which spacers were matched and is it a useful analysis output. Other similar tools (`lorikeet`) incorporate this kind of information in their spoligotyping report, thus I chose to combine the two outputs into a single output rather than "leaving it behind" in the job standard output.

PR #1886, opened 2018-05-16
Enabling using compressed files as inputs to LASTZ

nekrut:
Files are uncompressed on the fly without producing uncompressed intermediaries


blankenberg:
You probably need to set `ftype` in the tests for the datasets that you want to test the gzip functionality, otherwise galaxy is probably decompressing the files on upload.

tools/lastz/lastz_d.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="lastz_d_wrapper" name="LASTZ_D" version="1.3">
+<tool id="lastz_d_wrapper" name="LASTZ_D" version="@LASTZ_CONDA_VERSION@.galaxy4">
```
bgruening: Shit, this will not work. LASTZ_CONDA_VERSION is 1.0.4 ... so it will be lower than the previous one. I would stick to 1.3.1 or something until LASTZ catches up :(

tools/lastz/lastz_macros.xml
```diff
@@ -2,11 +2,24 @@
     <token name="@LASTZ_CONDA_VERSION@">1.0.4</token>
     <token name="@TARGET_INPUT_COMMAND_LINE@">
         #if $source.ref_source=="history":
-            '${source.target}'
+            #if $source.target.is_of_type('fasta.gz'):
```
bgruening: You could use CDATA here to make it more obvious.

tools/lastz/lastz_macros.xml
```diff
@@ -19,7 +32,7 @@
                 </param>
             </when>
             <when value="history">
-                <param name="target" type="data" format="fasta" label="Select a reference dataset" />
+                <param name="target" type="data" format="fasta, fasta.gz" label="Select a reference dataset" />
```
bgruening: there is a space between the datatypes, I had bad experience with this maybe more save to remove it.

tools/lastz/lastz_macros.xml
jmchilton: @bgruening Any clue if the biocontainers have gunzip and bunzip2 available and if not do we have recommendations for requirements to add to this to enable that? I should probably write that down somewhere in the planemo docs...

bgruening:
Thanks all!

nekrut:
I should probably do the same thing for blast+ tools

PR #1887, opened 2018-05-17
Modification of LASTZ behavior

nekrut:
- RPlot is no longer built by default. The old behavior was annoying when running on large collections. Additionally, the current version of dot plot is only really useful if query and target only contain one seq each. This is something that needs to be modified in the future. I have a Python script that makes PIP (Percent identity plot) for multiple sequences, but it needs to be polished. There is a new option to trigger plot creation.
- `--ambiguous=iupac` is now set by default to prevent erroring out on sequences containing ambiguous nucleotides
- Added housekeeping options


tools/lastz/lastz.xml
```diff
@@ -476,6 +488,10 @@
         <section name="interpolation" expanded="false" title="Interpolation">
             <param name="inner" type="integer" optional="true" argument="--inner" label="Perform additional alignment between the gapped alignment blocks, using (presumably) more sensitive alignment parameters" help="This value is used as the threshold for both the gap-free and gapped extension sub-stages; see the discussion of interpolation for more details. This option is only valid if gapped extension is performed."/>
         </section>
+        <section name="housekeeping" expanded="false" title="Housekeeping">
+            <param name="traceback" type="integer" optional="True" label="Set the amount of memory to allocate (in RAM) for trace-back information during the gapped extension stage (bytes)" help="The default traceback space is 83886080 or 80M" argument="--traceback"/>
```
bgruening: I don't think its a good idea to expose this option to a user.

nekrut: it is no exposed: `expanded="false"`

nekrut: oh, you mean because they can wrangle with RAM amounts? Well ... setting values there does not actually allocate anything, it just causes lastz not to die. It will not exceed what is set by Galaxy

bgruening: That does not matter, if a user can influence the memory consumption @natefoo needs to adopt the scheduler config. People can shoot them self in the foot here, no?

nekrut: I don't think so, you want me to hide it?

nekrut: I think that this option is well buried to cause any issues, but it can be useful occasionally

jmchilton: So as a compromise, can you set this automatically to some percent of GALAXY_MEMORY_MB if GALAXY_MEMORY_MB is set? Using some shell magic?

nekrut: It really depends on query size though. I'll just set it to 100 Mb

nekrut:
ping @bgruening

PR #1890, opened 2018-05-17
featureCounts: Output header as default and remove option for the headerless deseq2-specific output 

mblue9:
This PR 

- changes featurecounts to output the header (single row) as default, as per the upstream featurecounts tool
- removes the option to output deseq2-specific format (no header) as, with the related PR here https://github.com/galaxyproject/tools-iuc/pull/1889, deseq2 can accept files with a header row.
- also replaces `sed --expression=` with `sed -e` as otherwise the tests wouldn't run on my Mac


bgruening:
Thanks!

PR #1893, opened 2018-05-22
Merge BLAST XML tool

pvanheus:
A simple tool to merge BLAST XML datasets into a single BLAST XML dataset (based on the code in Galaxy itself).

### DIFF
--- a/tools/merge_blastxml/merge_blastxml.py
+++ b/tools/merge_blastxml/merge_blastxml.py
@@ -29,8 +29,8 @@
 # DEALINGS IN THE SOFTWARE
 
 import re
-import sys
 import shutil
+import sys
 xmlfiles = sys.argv[2:]
 xmlout = sys.argv[1]
 
@@ -46,7 +46,7 @@ def merge(split_files, output_file):
     iter_num = 2  # we only start using this from the 2nd file onwards
     for f in split_files:
         h = open(f)
-        body = False
+        # body = False
         first_line = False
         header = h.readline()
         if not header:
--- a/tools/merge_blastxml/merge_blastxml.py
+++ b/tools/merge_blastxml/merge_blastxml.py
@@ -1,0 +1,133 @@
+#!/usr/bin/env python
+
+# This file is derived from code in the Galaxy project that, but due to historical
+# reasons reflecting time developed outside of the Galaxy Project, this file is under
+# the MIT license.
+#
+# The MIT License (MIT)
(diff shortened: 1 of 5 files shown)


peterjc:
I don't mind putting this on the IUC repository, but equally it might have a home at https://github.com/peterjc/galaxy_blast/

What's the use case for this tool - does Galaxy's collection support not suffice?

pvanheus:
@peterjc I have a little workflow where I split a FASTA query, run a bunch of BLAST jobs and then get BLAST XMLs. I know this could be done using the Galaxy tasks support - effectively I was reproducing that behaviour, but using collections. Then finally the collection needs a "reduce" step to get to a single XML. Is that possible with Galaxy collections right now? If so I've probably written a tool for nothing. 

As to its location - the galaxy_blast collection is probably a more natural home.

peterjc:
@jmchilton can collections do what @pvanheus wants to do?

e.g. Reduce a collection of files of type XXX to a single file of type XXX, via the existing merge methods defined on the XXX datatype's Python (base) class (here BLAST XML, but equally FASTA, FASTQ, etc)

jmchilton:
Not yet @peterjc - it is a good idea though - https://github.com/galaxyproject/galaxy/issues/5464.

pvanheus:
Given @jmchilton 's answer, can I propose that I re-PR this as a tool over at https://github.com/peterjc/galaxy_blast/ and we move the discussion there?

bgruening:
fyi: Galaxy has native support for this: https://github.com/galaxyproject/galaxy/blob/30e3658b8b0e2f6b975dc6ccccb0cc8cc040247c/lib/galaxy/datatypes/blast.py#L91

The bad news is just that this feature is not really well maintained and currently broken afaik.

peterjc:
@bgruening That's the code which has been turned into a tool here - as far as I know it is working fine, but yes the task splitting is not a top tier Galaxy feature (not used at http://usegalaxy.org and not enabled by default).

pvanheus:
@bgruening yes, it does have that support (that this code is based on, with a few minor tweaks) but it is not exposed as a tool and is, as @peterjc says, part of the "task splitting" parallelisation framework that has been languishing in the code for a while. This tool splits that functionality out into an independent tool and the map / reduce can be achieved using FASTA split -> BLAST -> merge BLAST XML.

peterjc:
I'm happy to close this in favour of https://github.com/peterjc/galaxy_blast/pull/105

PR #1895, opened 2018-05-23
Repeat masker: fix crash when catalog in gzipped

abretaud:
It seems like sometimes repeatmasker decides to gzip its catalog output file... I experienced it with bigger input files, so it's hard to add a test case for it here.
Here's a fix for the failing mv command


nsoranzo:
Ping @pvanheus

pvanheus:
Thanks @abretaud - the switch is in RepeatMasker (line 755):

```perl
  my $totseqlen = $db->getSeqLength();
  my $compressCatFile = 1 if ( $totseqlen > 10000000 );
```
where `$db` is derived from the FASTA files being masked. I don't think it is feasible counting the bases of the input FASTA datasets (unless there is some metadata I'm not aware of) so I think your fix seems appropriate.

Issue #1892, opened 2018-05-22
Wrap HOMER for IUC?


mblue9:
>[HOMER]( http://homer.ucsd.edu/homer/)(Hypergeometric Optimization of Motif EnRichment) is a suite of tools for Motif Discovery and next-gen sequencing analysis.  It is a collection of command line programs for unix-style operating systems written in Perl and C++. HOMER was primarily written as a de novo motif discovery algorithm and is well suited for finding 8-20 bp motifs in large scale genomics data.  HOMER contains many useful tools for analyzing ChIP-Seq, GRO-Seq, RNA-Seq, DNase-Seq, Hi-C and numerous other types of functional genomics sequencing data sets.

As per conversation on Gitter, there are a few old wrappers already, could/should they be updated? 

I think the annotatePeaks module especially could be useful, as afaik, there's no other tool for easily adding gene names to peaks e.g. from macs2 or diffbind (please correct me if I'm wrong as I need a tool for that atm)

Current HOMER wrappers:
https://github.com/bgruening/galaxytools/tree/master/tools/homer
https://toolshed.g2.bx.psu.edu/view/kevyin/homer/f0b5827b6051 (this one has the annotatePeaks module)

ping @jennaj, @bgruening, @davebx

mblue9:
Thanks for the info @mvdbeek! That's useful to know. I've ended up wrapping ChIPseeker (https://github.com/bgruening/galaxytools/tree/master/tools/chipseeker) and that seems to work well for annotating peaks. I thought about using bedtools intersect or closest, but then one issue for us is, what annotation file to provide for users to intersect with (and how to version/maintain), and I'd rather not have users faff around with extra manipulation steps/regexs for this if they don't have to. I was looking for a one-step annotation solution and with chipseeker I think I found it :)

bernt-matthias:
I guess https://github.com/bgruening/galaxytools/tree/master/tools/homer is sufficient for now.

bgruening:
@lldelisle has also now added new/shiny HOMER tools here: https://github.com/galaxyproject/tools-iuc/tree/master/tools/homer

PR #1898, opened 2018-05-23
Add ChIPseeker to IUC

mblue9:
This PR adds [ChIPseeker](http://bioconductor.org/packages/release/bioc/html/ChIPseeker.html) (the functions for annotation and some plots), as I don't have time to wait for HOMER (https://github.com/galaxyproject/tools-iuc/issues/1892), I need a solution that works now. HOMER does look useful but also looks to be a bit painful to install in it's current form, as it requires manually installing HOMER and its data files, in addition to the wrapper. ChIPseeker seems to be a cleaner and simpler solution if you just want to annotate peaks.
This wrapper currently supports hg38, hg19 and mm10 and I think a future version could add support for any species with a GTF file.


tools/chipseeker/chipseeker.xml
```diff
@@ -0,0 +1,142 @@
...
+    ]]></version_command>
+    <command detect_errors="exit_code"><![CDATA[
+        #if $rscript:
+            cp '${chipseeker_script}' '${out_rscript}' &&
```
bgruening: can we just echo this, so its available as stdout?

tools/chipseeker/chipseeker.xml
```diff
@@ -0,0 +1,142 @@
+<tool id="chipseeker" name="ChIPseeker" version="1.14.2">
+    <description>for ChIP peak annotation and visualization</description>
+    <requirements>
+        <requirement type="package" version="1.14.2">bioconductor-chipseeker</requirement>
+        <requirement type="package" version="3.4.0">bioconductor-txdb.hsapiens.ucsc.hg38.knowngene</requirement>
```
bgruening: As a small warning, this will create an ENV which is probably a few GB big.

tools/chipseeker/chipseeker.xml
```diff
@@ -0,0 +1,142 @@
...
+            <option value="hg19">hg19</option>
+            <option value="mm10">mm10</option>
+        </param>
+
+        <param name="pdf" type="boolean" truevalue="True" falsevalue="" checked="True" label="Output a PDF file of plots?" help="Default: Yes" />
+        <param name="rscript" type="boolean" truevalue="TRUE" falsevalue="FALSE" checked="False" label="Output Rscript?" help="If this option is set to Yes, the Rscript used to annotate the IDs will be provided as a text file in the output. Default: No" />
+    </inputs>
+
+        <outputs>
+            <data name="out_tab" format="tabular" from_work_dir="out.tab" label="${tool.name} on ${on_string}: Annotated Peaks" />
```
bgruening: want to add metadata for the header?

tools/chipseeker/chipseeker.xml
mblue9: I guess I could but I'd like to have this script available as a file for teaching. Can I leave as is?

tools/chipseeker/chipseeker.xml
mblue9: ok have moved to your galaxytools repo as suggested

tools/chipseeker/chipseeker.xml
mblue9: Not particularly, not for this first version anyway, I could add later.

mblue9:
Ok, I've moved this over to @bgruening's galaxytools repo as it's big, PR here: https://github.com/bgruening/galaxytools/pull/728

PR #1906, opened 2018-05-30
Hyperlink DOIs against preferred resolver

katrinleinweber:
Same as https://github.com/galaxyproject/training-material/pull/793 ;-)

PR #1908, opened 2018-05-31
Follow-up to "Hyperlink DOIs against preferred resolver #1906"

katrinleinweber:
I had deleted my fork, so my commit a047997 that now [addresses the review comment](https://github.com/galaxyproject/tools-iuc/pull/1906#discussion_r191839604) was not auto-associated with #1906 m-(


nsoranzo:
@bgruening The failed build is due to an unrelated khmer issue (probably due to the conda package), merge?

bgruening:
Sure, but are we stopping travis and prevent the push?

nsoranzo:
No need, the build will fail, so no push stage.

bgruening:
So let it fail. Thanks!

PR #1912, opened 2018-06-01
FeatureCounts: small modifications (move up stranded param and add info on built-in annotation to Help)

mblue9:
@jennaj I've added info on the built-in annotation now and moved up the stranded parameter out of Advanced options, as discussed in Gitter. What do you and/or anyone else think.


bgruening:
Thanks @mblue9 and @lparsons for the review.

PR #1916, opened 2018-06-04
Update umi tools extract

mtekman:
* Adds whitelist w. test cases
 * ~~Adds regex barcodes~~


tools/umi_tools/umi-tools_extract.xml
```diff
@@ -57,12 +74,31 @@
...
+            <option value="string" selected="true">String</option>
+        </param>
+            
         <param name="bc_pattern" argument="--bc-pattern" type="text" label="Barcode pattern for first read"
             help="Use this option to specify the format of the UMI/barcode. Use Ns to
                     represent the random positions and Xs to indicate the bc positions.
                     Bases with Ns will be extracted and added to the read name. Remaining
                     bases, marked with an X will be reattached to the read.">
         </param>
+
```
bgruening: Can you add something along these lines:

```
<sanitizer invalid_char="">
      <valid initial="string.letters,string.digits"><add value="&gt;"/><add value="&lt;"/><add value="!="/>
      <add value="-"/></valid>
</sanitizer>
```

bgruening: This will maybe solve your problem from above?

tools/umi_tools/umi-tools_extract.xml
```diff
@@ -126,18 +162,38 @@
...
+            <param name="input_read1" value="scrb_seq_fastq.1.gz" ftype="fastq.gz" />
+            <param name="input_read2" value="scrb_seq_fastq.2.gz" ftype="fastq.gz" />
+            <param name="extract_method" value="string" />
+            <param name="bc_pattern" value="CCCCCCNNNNNNNNNN" />
+            <param name="use_barcodes" value="yes" />
+            <param name="filter_barcode_file" value="scrb_seq_barcodes" />
+            <output name="out2" file="scrb_extract.fastq.gz" decompress="true" />
         </test>
+        <!-- <test><!-\- same as above but with regex barcode-\-> -->
+        <!--     <param name="type" value="paired" /> -->
```
bgruening: why is this commented out?

tools/umi_tools/umi-tools_extract.xml
mtekman: perfect, thank you! I saw this but thought it would only allow for those characters to be typed by the uses but that they would still be parsed badly when the commandline was generated.

tools/umi_tools/umi-tools_extract.xml
mtekman: now enabled with the above fix for regex characters

bgruening:
Thanks @mtekman!

PR #1920, opened 2018-06-05
Limma-voom: Add Glimma interactive MD plot and DE table to report

mblue9:
This PR adds a [Glimma](https://bioconductor.org/packages/release/bioc/html/Glimma.html) interactive MD plot and DE genes table to the output report.

>The interactive MD plot contains three key components which interact with each
other to show multiple aspects of the data in the one display.
The main component is a plot of gene-wise summarised statistics
which takes the top-left panel of the html page. Gene-wise logFCs are plotted against
gene-wise average logCPM values where each point on the plot represents a single gene.
Hovering your cursor over or clicking on a gene (or point) within the main plot brings
the expression of each sample for the selected gene in a plot in the top-right panel. At
the same time, associated gene information is displayed in the table below.
Users can simply scroll through the table looking for any gene that is of interest,
or hone into specific genes or groups of genes using the search function in the table.
Clicking on a gene (or row) in the table interacts with both of the plots simultaneously –
the selected gene is highlighted in the MD plot and next to it, the expression each sample
is displayed.
The order of genes displayed in the table can be re-ordered in an increasing or decreasing
fashion by clicking on the header of a column. This is useful to see which genes
have the smallest raw or adjusted p-value, or for sorting genes into those that are most
up- or down-regulated in terms of logFC. The ordering function in the table used in
conjunction with the search function can be especially powerful as an exploratory tool, for example, one can search by a keyword of interest, say “structural maintanence” and
then order the reduced table of genes by adjusted p-value.


bgruening:
Cool thanks!

PR #1923, opened 2018-06-06
DiffBind: replace spaces in element_identifier and fix control BAM input (if using same controls for both groups)

mblue9:
To fix the issue reported here: https://github.com/galaxyproject/usegalaxy-playbook/issues/127#issuecomment-394713249


mblue9:
Have added a fix for this control BAM issue reported by @natefoo (when same controls are used for both groups) here: https://github.com/galaxyproject/usegalaxy-playbook/issues/127#issuecomment-395073384

tools/diffbind/diffbind.R
```diff
@@ -69,6 +69,8 @@ if ( length(ctrls) != 0 ) {
                         stringsAsFactors=FALSE)
 }
 
+print(sampleTable)
```
bgruening: can this be removed?

mblue9: yep, done

PR #1925, opened 2018-06-06
Add wrapper for qfilt.

davebx:
This re-adds the qfilt wrapper from #1021


tools/qfilt/qfilt.xml
```diff
@@ -0,0 +1,167 @@
...
+                <param name="mismatch" argument="-t" type="integer" value="0" label="Mismatch" help="if PREFIX is supplied, prefix matching tolerates at most MISMATCH mismatches" />
+                <param name="format" argument="-f" type="select" label="Format" help="output in FASTA or FASTQ format (default=FASTA)" >
+                    <option value="FASTA">FASTA</option>
+                    <option value="FASTQ">FASTQ</option>
+                </param>
+            </when>
+        </conditional>
+    </inputs>
+    <outputs>
+        <data format="fasta" name="filtered_output"/>
```
bgruening: You should change the format depending on the format argument which you are asking from the user.

tools/qfilt/qfilt.xml
```diff
@@ -0,0 +1,172 @@
...
+    </tests>
+    <help><![CDATA[
+qfilt
+=====
+
+This simple program is meant to filter sequencing data,
+optionally removing or splitting reads with poor quality scores
+and to optionally _only_ retain fragments from reads that are tagged with a given 5' sequence.
+
+### OUTPUT: ####
```
bgruening: The output is missing here? Is the rest about stderr and stdout intended?

davebx: Yeah, I'll change that to `### OUTPUTS: ####`

tools/qfilt/qfilt.xml
```diff
@@ -0,0 +1,172 @@
...
+qfilt
+=====
+
+This simple program is meant to filter sequencing data,
+optionally removing or splitting reads with poor quality scores
+and to optionally _only_ retain fragments from reads that are tagged with a given 5' sequence.
+
+### OUTPUTS: ####
+
+#### stderr: ####
```
bgruening: @davebx I was more concerned about the stderr and stdout. Is this really just accessible on stdout? Most of the users will not now what stderr means. Is the stderr really needed?
Can we not simply write, example output ... fasta content ... done?

Also `#### foo ####` is not valid rst, isn't it?

PR #1926, opened 2018-06-06
Add wrapper for tn93.

davebx:
This re-adds the TN93 wrapper from #1021


tools/tn93/tn93.xml
```diff
@@ -0,0 +1,111 @@
...
+                    label="Only count pairs that overlap by at least N bases" />
+                <param name="counts_in_name" argument="-d" type="text"
+                    label="Sequence name component immediately preceding a copy number" />
+            </when>
+        </conditional>
+    </inputs>
+    <outputs>
+        <data format="csv" name="tn93_distances">
+            <change_format>
+                <when input="format" value="csvn" format="csvn" />
```
bgruening: @davebx csvn is not a valid Galaxy format isn't it?

davebx: You're right. Neither is hyphy.

bgruening: ping

davebx: I decided to just have `csv` as the format, and if/when we add the hyphy-specific datatype, we can create a new version of this wrapper.

tools/tn93/tn93.xml
```diff
@@ -0,0 +1,106 @@
...
+        #end if
+        '$input_fasta'
+        ]]></command>
+    <inputs>
+        <param name="input_fasta" type="data" format="fasta" label="Input in FASTA format" />
+        <param name="second_fasta" argument="-s" type="data" format="fasta" optional="true"
+            label="Additional input FASTA for pairwise comparisons" />
+        <param name="format" argument="-f" type="select" label="Output format">
+            <option value="csv">csv</option>
+            <option value="csvn">csvn</option>
```
bgruening: @davebx ??

tools/tn93/tn93.xml
```diff
@@ -0,0 +1,106 @@
...
+                    label="Sequence name component immediately preceding a copy number" />
+            </when>
+        </conditional>
+    </inputs>
+    <outputs>
+        <data format="csv" name="tn93_distances" />
+    </outputs>
+    <tests>
+        <test>
+            <!-- <param name="history_reference" value="tn93-in-ref-1.fa" /> -->
```
bgruening: Can this be removed?

bgruening:
ping @davebx

tools/tn93/.shed.yml
```diff
@@ -0,0 +1,11 @@
...
+owner: iuc
+description: Compute distances between sequences
+long_description: |
+  This is a simple program meant to compute pairwise distances between aligned
+  nucleotide sequences in sequential FASTA format using the Tamura Nei 93 distance
+remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/master/tools/tn93/
+homepage_url: https://github.com/veg/tn93/
+type: unrestricted
+categories:
+ - Next Gen Mappers
```
bgruening: Maybe sequence analysis?

tools/tn93/tn93.xml
```diff
@@ -0,0 +1,99 @@
...
+                    label="Maximum tolerated fraction of ambiguous characters" />
+                <param name="overlap" argument="-l" type="integer" value="100"
+                    label="Only count pairs that overlap by at least N bases" />
+                <param name="counts_in_name" argument="-d" type="text"
+                    label="Sequence name component immediately preceding a copy number" />
+            </when>
+        </conditional>
+    </inputs>
+    <outputs>
+        <data format="csv" name="tn93_distances" />
```
bgruening: Same here, I think we should convert the output to a tabular and output this instead.

tools/tn93/tn93.xml
```diff
@@ -0,0 +1,99 @@
...
+    <command detect_errors="exit_code"><![CDATA[
+    tn93
+        -o '$tn93_distances'
+        #if str($options.advanced) == 'advanced':
+            -t $options.threshold
+            -a $options.ambigs
+            -g $options.fraction
+            -f csv
+            -l $options.overlap
+            #if len(str($options.counts_in_name)) > 0:
```
bgruening: instead of doing this, you could add a validator that check if the text is non-empty.

tools/tn93/.shed.yml
davebx: That makes sense, I'm not sure why I had that category. Probably because I was going to ask someone, but got distracted by other obligations.

tools/tn93/tn93.xml
davebx: I'm not convinced that's a good idea in this case, since I suspect that some downstream tools designed with tn93 in mind expect csv rather than tsv. @stevenweaver or @rdvelazquez would be more qualified to say, though.

rdvelazquez: @davebx, correct. HIV-clustering (https://github.com/veg/tools-iuc/tree/master/tools/hivclustering) consumes the output of tn93 and expects a csv file.

davebx: Thanks, Ryan. I had a feeling that was the case, but I'd rather hear from someone who _knows_.

tools/tn93/tn93.xml
davebx: It should have been `optional`, and now is

tools/tn93/readreduce.xml
```diff
@@ -0,0 +1,75 @@
+<tool id="readreduce" name="Merge matching reads" version="1.0.4">
+    <description>into clusters with TN-93</description>
+    <requirements>
+        <requirement type="package" version="1.0.4">tn93</requirement>
+    </requirements>
+    <version_command><![CDATA[tn93 --version]]></version_command>
+    <command detect_errors="exit_code"><![CDATA[
+    readreduce
+    -q
+        -o '$output'
```
bgruening: indentation

tools/tn93/readreduce.xml
```diff
@@ -0,0 +1,75 @@
+<tool id="readreduce" name="Merge matching reads" version="1.0.4">
```
bgruening: maybe tn93_readreduce?

tools/tn93/.shed.yml
```diff
@@ -0,0 +1,11 @@
+name: tn93
+owner: iuc
+description: Compute distances between sequences
```
bgruening: should that be a suite file as it contains two tools?

rdvelazquez:
Any updates on this?

bgruening:
ping @davebx

tools/tn93/tn93.xml
```diff
@@ -0,0 +1,99 @@
+<tool id="tn93" name="TN93" version="1.0.4">
+    <description>compute distances between aligned sequences</description>
+    <requirements>
+        <requirement type="package" version="1.0.4">tn93</requirement>
```
bgruening: Should the requirement be macrofield and the version?

PR #1933, opened 2018-06-11
Pygenometracks

joachimwolff:
I took the wrapper from hicPlotTADs and added the `multiple` function. Any new feature that I missed? @fidelram @vivekbhr


fidelram:
What is multiple? That you can overlay tracks?

Here is a list of relevant changes in the latest version:

> - Allows a track be plotted over the previous. There is not limit on the tracks that can be plotted on top of each other
> - Added option to plot links as 'triangles'. This is useful to overlay links with Hi-C data to identify the contact pixel.
> - Added line width to bed track
> - Added 'bed graph' track
> - Added a track that can plot .narrowPeaks

joachimwolff:
Multiple is to plot n tracks with the same settings with just one command for one image.

bgruening:
.shed.yml file is missing to get the tests running.

vivekbhr:
@joachimwolff is this pulled from master branch of pyGenomeTracks? Some of the descriptions still mention hicPlotTADs.

joachimwolff:
@vivekbhr It is a renamed version of the hicPlotTADs wrapper from HiCExplorer and I removed all not used code from the macros.xml file. Moreover I extended the wrapper a bit by adding the `multiple` parameter to the tracks. With this it is possible to plot n tracks with the identical settings at once and users do not have to add n tracks manually. It is a Galaxy wrapper parameter, we just loop over these files and add n tracks in the background. I am using the latest conda version of pyGenomeTracks which is 2.0. 

@bgruening I always forget this file.. :disappointed:

tools/pygenometracks/pyGenomeTracks.xml
```diff
@@ -0,0 +1,742 @@
...
+            <param name="image_file_format" value="png" />
+            <output name="outFileName" file="master_TADs_plot.png" ftype="png" compare="sim_size" delta="35000" />
+        </test>
+    </tests>
+    <help><![CDATA[
+
+pyGenomeTracks
+==============
+
+**pyGenomeTracks** is a visualization tool to plot the topologically associating domains (TADs) in a given region.
```
bgruening: this needs also to be adjusted, this tool should plot everything, TADs is just one feature of many, and it should be possible to plot tracks without having a TAD track.

tools/pygenometracks/pyGenomeTracks.xml
```diff
@@ -0,0 +1,742 @@
...
+            <repeat name="tracks">
+                <conditional name="track_file_style_conditional">
+                    <param name="track_file_style_selector" value="vlines_track_option" />
+                    <param name="track_input_bed_single" value="domains.bed" ftype="bed" />
+                    <param name="type" value="vlines" />
+                </conditional>
+            </repeat>
+            <param name="image_file_format" value="png" />
+            <output name="outFileName" file="master_TADs_plot.png" ftype="png" compare="sim_size" delta="35000" />
+        </test>
```
bgruening: a second test, without HIC would be nice.

tools/pygenometracks/pyGenomeTracks.xml
joachimwolff: I changed the description a bit to cover this.

joachimwolff:
Any idea why travis quits after 20 mins because of no output from the test cases?

mvdbeek:
Could just take too much time to download and unpack the dependency. I've restarted it now, maybe it'll pass

bgruening:
Finally, the conda resolver had problems again. Fixed it upstream.

PR #1935, opened 2018-06-12
Fix cutadapt in multicore mode

### LINKED ISSUE #1934: Cutadapt 1.16.1 bug -q incompatible with --format
Hello,
I tried to run cutadapt with a fastq, a fastqsanger and a fastq.gz and I always have the message:
ERROR: Running in parallel is currently not supported for the given combination of command-line parameters.
These options are not supported: --info-file, --rest-file, --wildcard-file, --untrimmed-output, --untrimmed-paired-output, --too-short-output, --too-short-paired-output, --too-long-output, --too-long-paired-output, --format, --colorspace
Indeed, I found in the xml parser:
cutadapt

        -j \${GALAXY_SLOTS:-4}

        --format=$format

but in the documentation:
http://cutadapt.readthedocs.io/en/stable/guide.html
I found:
Additionally, the following command-line arguments are not compatible with multi-core:

        --format
        --colorspace

Thanks,

Lucille

mvdbeek:
We can't really test this since the testing framework sets GALAXY_SLOTS to 1, but I've tested this locally and all tests pass.

Fixes #1934.

### DIFF
--- a/tools/cutadapt/cutadapt.xml
+++ b/tools/cutadapt/cutadapt.xml
@@ -75,8 +75,6 @@
 
         -j \${GALAXY_SLOTS:-4}
 
-        --format=$format
-
        #if str( $library.type ) == "single":
             @read1_options@
             --output='$out1'
--- a/tools/cutadapt/cutadapt.xml
+++ b/tools/cutadapt/cutadapt.xml
@@ -10,139 +10,138 @@
     <version_command>cutadapt --version</version_command>
 
     <command detect_errors="exit_code"><![CDATA[
+## Link in the input and output files, so Cutadapt can tell their type
 
-        ## Link in the input and output files, so Cutadapt can tell their type
+#import re
+#set format = "fastq"
+#set read1 = "input_f"
+#set read2 = "input_r"
+#set paired = False
+#set library_type = str($library.type)
+#if $library_type == 'paired':
+#set paired = True
+#set read1 = re.sub('[^\w\-\s]', '_', str($library.input_1.element_identifier))
+#set read2 = re.sub('[^\w\-\s]', '_', str($library.input_2.element_identifier))
+#set input_1 = $library.input_1
+#set input_2 = $library.input_2
+#else if $library_type == 'paired_collection'
+#set paired = True
+#set input_1 = $library.input_1.forward
+#set input_2 = $library.input_1.reverse
+#set read1 = re.sub('[^\w\-\s]', '_', str($library.input_1.name))
+#else
+#set input_1 = $library.input_1
+#set read1 = re.sub('[^\w\-\s]', '_', str($library.input_1.element_identifier))
+#end if
 
-        #import re
-        #set format = "fastq"
-        #set read1 = "input_f"
-        #set read2 = "input_r"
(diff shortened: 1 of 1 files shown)


tools/cutadapt/cutadapt.xml
```diff
@@ -10,138 +10,138 @@
...
+    ln -f -s '${input_2}' '$read2' &&
+#end if
+
+## Run Cutadapt
+
+cutadapt
+
+## cutadapt (up to version 1.16) can't be run in multicore mode with these options
+#if not any(($output_options.info_file, $output_options.rest_file, $output_options.wildcard_file, $output_options.too_short_file, $output_options.too_long_file, $output_options.untrimmed_file))
+     -j \${GALAXY_SLOTS:-4}
```
nsoranzo: I'd also change the `GALAXY_SLOTS` default to 1 to be on the safe side.

tools/cutadapt/cutadapt.xml
```diff
@@ -10,138 +10,138 @@
...
-            '${read2}'
-        #end if
-
-        #if $output_options.report:
-            > report.txt
-        #end if
+## Link in the input and output files, so Cutadapt can tell their type
+
+#import re
+#set format = "fastq"
```
nsoranzo: You can now remove this line.

tools/cutadapt/cutadapt.xml
```diff
@@ -10,138 +10,138 @@
...
+#end if
+
+#if $input_1.is_of_type("fastq.gz", "fastqsanger.gz"):
+    #set read1 = $read1 + ".fq.gz"
+    #set out1 = "out1.gz"
+#else if $input_1.is_of_type("fastq.bz2", "fastqsanger.bz2"):
+    #set read1 = $read1 + ".fq.bz2"
+    #set out1 = "out1.bz2"
+#else if $input_1.is_of_type('fasta'):
+    #set format = "fasta"
```
nsoranzo: You can now remove this line.

tools/cutadapt/cutadapt.xml
```diff
@@ -10,138 +10,138 @@
...
+
+#if $paired:
+    #if $input_2.is_of_type("fastq.gz", "fastqsanger.gz"):
+        #set read2 = $read2 + ".fq.gz"
+        #set out2 = "out2.gz"
+    #else if $input_2.is_of_type("fastq.bz2", "fastqsanger.bz2"):
+        #set read2 = $read2 + ".fq.bz2"
+        #set out2 = "out2.bz2"
+    #else if $input_2.is_of_type('fasta'):
+        #set format = "fasta"
```
nsoranzo: You can now remove this line.

tools/cutadapt/cutadapt.xml
mvdbeek: To be on the safe side wrt available cores ? Not that I don't think 1 is the better default, but this should only be included if none of the incompatible options have been chosen.

nsoranzo: If Galaxy doesn't define `GALAXY_SLOTS` for a cluster scheduler.

PR #1937, opened 2018-06-12
Trinity 2.6.6

abretaud:
Here's an update from Trinity 2.4.0 to 2.6.6
Now there's a wrapper to generate [super transcripts](https://github.com/trinityrnaseq/trinityrnaseq/wiki/SuperTranscripts).
As things would have been too simple with just changing the `@WRAPPER_VERSION@`, I also add to update a few dependencies, countless test data, and update some wrappers a now trinity generates a gene map that needs to be reused by some of the scripts (replacing `--trinity_mode`)


tools/trinity/abundance_estimates_to_matrix.xml
```diff
@@ -19,13 +19,50 @@
...
     <inputs>
         <param name="samples" label="Abundance estimates" type="data" multiple="true" format="tabular" help="output(s) from 'Align reads and estimate abundance' tool" />
 
+        <conditional name="gene_map">
+            <param name="has_gene_map" type="select" label="Use gene level estimates?">
+                <option value="yes">Yes</option>
+                <option value="no">No</option>
+            </param>
+            <when value="yes">
+                <param format="tabular" name="gene_trans_map" argument="--gene_trans_map" type="data" label="Gene to transcript correspondence ('gene(tab)transcript' lines)" />
```
bgruening: could you not just use the param and make it optional and remove the conditional?

tools/trinity/super_transcripts.xml
```diff
@@ -0,0 +1,35 @@
+<tool id="trinity_super_transcripts" name="Generate SuperTranscripts" version="@WRAPPER_VERSION@">
+    <description>from a Trinity assembly</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements"/>
+    <command detect_errors="aggressive"><![CDATA[
+       Trinity_gene_splice_modeler.py --incl_malign --trinity_fasta "$assembly"
```
bgruening: single quotes

bgruening:
Thanks @abretaud!

lldelisle:
Hi,
In my private instance, the dependency is not solved. Is there a package trinity 2.6.6 to download?
Many thanks,

Lucille

abretaud:
@lldelisle there's a conda package for this version of trinity (https://anaconda.org/bioconda/trinity)
If your galaxy is configured to use conda for dependency resolution, it should install automatically
We no longer maintain old-style tool_dependency.xml for trinity

PR #1943, opened 2018-06-15
spades: added regex for detection of out of memory

bernt-matthias:


tools/spades/spades.xml
```diff
@@ -1,10 +1,14 @@
-<tool id="spades" name="SPAdes" version="3.11.1">
+<tool id="spades" name="SPAdes" version="3.11.1.1">
```
nsoranzo: Can you use `3.11.1+galaxy1` instead? See the new guideline at https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tool-versions

tools/spades/spades.xml
```diff
@@ -1,10 +1,14 @@
...
     <description>genome assembler for regular and single-cell projects</description>
     <requirements>
         <requirement type="package" version="3.11.1">spades</requirement>
     </requirements>
     <stdio>
         <exit_code range="1:" />
+        <regex match="The reads contain too many k-mers to fit into available memory"
+           source="stdio"
```
nsoranzo: I think `source` can be only `stderr`, `stdout` or `both`, see https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-stdio-regex

tools/spades/spades.xml
bernt-matthias: done. I like this guideline

tools/spades/spades.xml
bernt-matthias: my bad. should have done linting. sorry for that

tools/spades/spades.xml
```diff
@@ -1,10 +1,14 @@
...
+<tool id="spades" name="SPAdes" version="3.11.1+galaxy1">
     <description>genome assembler for regular and single-cell projects</description>
     <requirements>
         <requirement type="package" version="3.11.1">spades</requirement>
     </requirements>
     <stdio>
         <exit_code range="1:" />
+        <regex match="The reads contain too many k-mers to fit into available memory"
+           source="stdout"
+           level="fatal_oom"
```
nsoranzo: It looks like that `fatal_oom` level was not added to the `LevelType` simpleType in `lib/galaxy/tools/xsd/galaxy.xsd` for Galaxy 18.05, so this fails linting.

bernt-matthias: OK. Opened an issue https://github.com/galaxyproject/galaxy/issues/6340 - would have created a PR, but I guess this needs kind of back porting?

nsoranzo: You just need to open a PR against the `release_18.05` branch. From your local git fork:
```
git checkout --no-track -b release_18.05_fix_xsd upstream/release_18.05
```
and when you push:
```
git push --set-upstream origin release_18.05_fix_xsd
```

bernt-matthias: thanks https://github.com/galaxyproject/galaxy/pull/6341

bernt-matthias: could someone restart travis. linting should work now.

bernt-matthias: heureka. linting works finally :) :+1:

bgruening:
Thanks!

PR #1944, opened 2018-06-17
Add @TMPDIR_OPTION@ to set working directory

pvanheus:
I recently ran into a problem with `picard` running out of disk space on our cluster as it was writing to `/tmp` on worker nodes. This behaviour is controlled by the TMP_DIR command line parameter to the picard tools. This PR adds this parameter to each wrapper and propagates the setting from the `$TEMP` environment variable

The Galaxy admin can set `$TEMP` using a `<env>` clause in `job_conf.xml`. For jobs running with the local job runner `$TEMP` is [already set](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/runners/local.py#L50) by default.

An alternative way to set the temporary working directory for picard would be to set `-Djava.io.tmpdir=` to a sensible value as recommended on [Biostars](https://www.biostars.org/p/42613/). The effect seems to be the same as setting `TMP_DIR` thus setting `TMP_DIR` was chosen as the more straightforward of the two options.

### DIFF
--- a/tools/picard/picard_macros.xml
+++ b/tools/picard/picard_macros.xml
@@ -1,6 +1,6 @@
 <macros>
     <token name="@TMPDIR_OPTION@">
-     `if [ -n "\$TMPDIR" ] ; then echo "TMP_DIR=\$TMPDIR" ; else if [ -n "\$TEMP" ] ; then echo "TMP_DIR=\$TEMP" ; fi ; fi`</token>
+     `if [ -n "\$TMPDIR" ] ; then echo 'TMP_DIR=\$TMPDIR' ; else if [ -n "\$TEMP" ] ; then echo 'TMP_DIR=\$TEMP' ; fi ; fi`</token>
     <xml name="VS">
         <param name="validation_stringency" type="select" label="Select validation stringency" help=" Setting stringency to SILENT can improve performance when processing a BAM file in which variable-length data (read, qualities, tags) do not otherwise need to be decoded.">
           <option value="LENIENT" selected="True">Lenient</option>
--- a/tools/picard/picard_macros.xml
+++ b/tools/picard/picard_macros.xml
@@ -1,6 +1,6 @@
 <macros>
     <token name="@TMPDIR_OPTION@">
-     `if [ -n "\$TMPDIR" ] ; then TMP_DIR="\$TMPDIR" ; else if [ -n "\$TEMP" ] ; then TMP_DIR="\$TEMP" ; fi ; fi`</token>
+     `if [ -n "\$TMPDIR" ] ; then echo TMP_DIR="\$TMPDIR" ; else if [ -n "\$TEMP" ] ; then echo TMP_DIR="\$TEMP" ; fi ; fi`</token>
     <xml name="VS">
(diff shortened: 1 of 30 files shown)


pvanheus:
BTW I believe that using `$TMP_DIR` or `$TEMP` if they are set means that if the admin wants this functionality in recent (18.01 and later) Galaxy they should set `tmp_dir` in `job_conf.xml` as per https://github.com/galaxyproject/galaxy/issues/4606 and https://github.com/galaxyproject/galaxy/pull/5193. Alternatively they can use `<env>` tag to set one of these to environment variables.

tools/picard/picard_macros.xml
```diff
@@ -1,4 +1,6 @@
 <macros>
+    <token name="@TMPDIR_OPTION@">
+     `if [ -n "\$TMPDIR" ] ; then TMP_DIR="\$TMPDIR" ; else if [ -n "\$TEMP" ] ; then TMP_DIR="\$TEMP" ; fi ; fi`</token>
```
nsoranzo: I think you need to add `echo`s as below, otherwise nothing gets added to the command line:
```
`if [ -n "\$TMPDIR" ] ; then echo TMP_DIR="\$TMPDIR" ; else if [ -n "\$TEMP" ] ; then echo TMP_DIR="\$TEMP" ; fi ; fi`
```

tools/picard/picard_macros.xml
```diff
@@ -1,4 +1,6 @@
 <macros>
+    <token name="@TMPDIR_OPTION@">
+     `if [ -n "\$TMPDIR" ] ; then echo TMP_DIR="\$TMPDIR" ; else if [ -n "\$TEMP" ] ; then echo TMP_DIR="\$TEMP" ; fi ; fi`</token>
```
nsoranzo: In my testing the only quoting that works properly is:
```
echo 'TMP_DIR=$TEMP'
```
so I think this line should be:
```
`if [ -n "\$TMPDIR" ] ; then echo 'TMP_DIR=\$TMPDIR' ; else if [ -n "\$TEMP" ] ; then echo 'TMP_DIR=\$TEMP' ; fi ; fi`</token>
```

pvanheus: In my testing both `echo "TMP_DIR=\$TEMP"` and `echo "TMP_DIR=\$TEMP"` works.

nsoranzo: @pvanheus I guess you meant that both `echo 'TMP_DIR=\$TEMP'` and `echo "TMP_DIR=\$TEMP"` worked for you.

Unfortunately the double quoted version breaks in the silly case where the value of `TEMP` contains a space

tools/picard/picard_macros.xml
```diff
@@ -1,4 +1,6 @@
 <macros>
+    <token name="@TMPDIR_OPTION@">
+     `if [ -n "\$TMPDIR" ] ; then echo "TMP_DIR=\$TMPDIR" ; else if [ -n "\$TEMP" ] ; then echo "TMP_DIR=\$TEMP" ; fi ; fi`</token>
```
nsoranzo: Unfortunately the double-quoted echo breaks in the silly case where the value of TEMP contains a space, any reason not to use the single quotes?

PR #1947, opened 2018-06-19
add -R to featurecounts

bgruening:


tools/featurecounts/featurecounts.xml
```diff
@@ -108,6 +111,9 @@
             && sed -e 's|${alignment}|${alignment.element_identifier}|g' 'output.jcounts' > '${output_jcounts}'
         #end if
 
+        #if $extended_parameters.R:
+            && samtools sort *.featureCounts.bam -o '$output_bam'
```
yhoogstrate: The sam usage says `samtools sort [options...] [in.bam]`. If desired, this could be rephrased to `samtools sort -o '$output_bam' *.featureCounts.bam`

tools/featurecounts/featurecounts.xml
```diff
@@ -389,6 +395,13 @@
                    label="Ignore reads marked as duplicate"
                    help="If specified, reads that were marked as duplicates will be ignored. Bit Ox400 in the FLAG field of a SAM/BAM file is used for identifying duplicate reads. In paired end data, the entire read pair will be ignored if at least one end is found to be a duplicate read." />
 
+            <param type="boolean"
+                   truevalue="-R BAM"
+                   falsevalue=""
+                   argument="-R"
+                   label="Output detailed assignment results for each read or read-pair"
```
yhoogstrate: After reading this I wasn't sure what it precisely does. I think it would be handy to describe which tag is used and that this annotates an alignment file. Maybe we could proceed with:

`label="Annotates the alignment file with 'XS:Z:'-tags to described per read or read-pair the corresponding assigned feature(s)."`

tools/featurecounts/featurecounts.xml
```diff
@@ -500,14 +519,18 @@
...
             <param name="anno_select" value="builtin"/>
             <param name="format" value="tabdel_short" />
+            <section name="extended_parameters">
+                <param name="R" value="true" />
+            </section>
             <output name="output_short" file="output_builtin_hg19.tab">
                 <metadata name="column_names" value="Geneid,pairend_strandspecific_51mer_hg19_chr1_1-100000.bam"/>
             </output>
             <output name="output_summary" file="output_summary_builtin_hg19.tab"/>
+            <output name="output_bam" file="output.bam" ftype="bam"/>
```
yhoogstrate: This test makes more sense with a GTF included, since all features are set to `XS:Z:Unassigned_NoFeatures`

yhoogstrate:
sourceforge is under maintainance ...

tools/featurecounts/featurecounts.xml
bgruening: Agreed, but I searched for a low complex test to include it. We see that the tag assignment works, this is what counts, isn't it?

bgruening:
> sourceforge is under maintainance ...

Hopefully they fix it today. Could you approve when you are happy so I can merge when sourceforge is back?
Thanks @yhoogstrate for the review!

bgruening:
Yeah, thanks a lot!

PR #1948, opened 2018-06-20
Add Umi tools count

mtekman:
Adds the count method from umi tools. Useful in count matrix generation from multiplexed FASTQ files


tools/umi_tools/umi-tools_count.xml
```diff
@@ -0,0 +1,441 @@
...
+
+Finally, if you have used umis to extract the UMI +/- cell barcode,
+you can specify --extract-umi-method=umis
+
+The start position of a read is considered to be the start of its alignment
+minus any soft clipped bases. A read aligned at position 500 with
+cigar 2S98M will be assumed to start at position 498.
+
+
+Input/Output Options
```
bgruening: I guess this can be removed as it is described in the params above.

tools/umi_tools/umi-tools_count.xml
```diff
@@ -0,0 +1,441 @@
+<tool id="umi_tools_extract" name="UMI-tools extract" version="@VERSION@.1">
```
bgruening: its the count tool, you have not changed the ID, the name and the description.

tools/umi_tools/umi-tools_count.xml
```diff
@@ -0,0 +1,181 @@
...
+    <macros>
+        <import>macros.xml</import>
+        <xml name="sanitize_tag" >
+            <sanitizer invalid_char="">
+                <valid initial="string.letters,string.digits" />
+            </sanitizer>
+        </xml>
+    </macros>
+    <expand macro="requirements" >
+        <requirement type="package" version="1.7">samtools</requirement>
```
bgruening: Is samtools still needed?

mtekman: no, you're right -- I added it because I thought it was causing the indexing problems earlier, but everything runs fine without it.

bgruening:
@mtekman You have now removed a tool. Before you added a new test but forget to include the test file, thats why the test was failing.

mtekman:
@bgruening thank you

PR #1950, opened 2018-06-21
emboss sixpack fix empty parameter uppercase

bernt-matthias:
- sixpack does not work with empty parameter uppercase
- readded test which is now working
- added missing test files
- also added version macro


bgruening:
Thanks!

PR #1951, opened 2018-06-24
Small Change to intersectBed.xml

Florian-H-Lab:
Small change due to a fatal error when no overlap mode is selected. Leading to:

***** ERROR: Unrecognized parameter: None *****

However the intersect tool of bed allows for a default behavior which is not supported if the user is forced to pick an overlap mode.


nekrut:
You might want to up the wrapper version

PR #1956, opened 2018-06-30
Upgrade jbrowse to 1.12.5, bump tool version to 4

pvanheus:
Upstream (in bioconda) 1.12.5 build 2 has fixed an error in that Bio::FeatureIO was not added as a dependency when it was needed. This PR updates the jbrowse requirement from 1.12.3 to 1.12.5 to take advantage of that update.

The versioning of the tool is a little opaque. Wrapper version is 0.7.0, then an extra number is appended. I presume 0.7.0 refers to the versioning of the tool semantics and the version (3 before, 4 now) that is added to that refers to the version of the tool implementation (updated here because of dependency updates).


pvanheus:
The failing check is due to a too-specific test (the `test-data/gff3/test.xml`).

abretaud:
Thanks @pvanheus!
There's also #1683 still opened, but I see no objection to merge this sooner than #1683 if it fixes a bug
The version number looks good. Are you looking into the failing test?

ping @erasche

erasche:
Thanks for the ping @abretaud, I'm +1 on it.

pvanheus:
Thanks @abretaud and @erasche - I'll work on fixing the tests today and then it can be merged.

pvanheus:
The latest patches are from my work going back to the jbrowse tool that was in tools-iuc, fixing it up so that all tests pass and then merging those changes into the updated jbrowse tool repo.

erasche:
0.7.0 was an arbitrary number. I've been bumping them minor version as I add new features

pvanheus:
So @erasche there are 2 "versions" - 0.7.0 for the "wrapper version" and then a final ".3" (or in this PR's case ".4") that is added after the "wrapper version". So the final version that is currently in the IUC collection is 0.7.0.3. I could just bump it to 0.7.1.3 and ensure that all version number changes are in the macros.

erasche:
LGTM.

PR #1958, opened 2018-07-01
Add intermine galaxy interchange formatter tool.

blankenberg:


tools/intermine_galaxy_exchange/intermine_galaxy_exchange.xml
```diff
@@ -0,0 +1,93 @@
+<tool id="galaxy_intermine_exchange" name="Create InterMine Interchange" version="0.0.1">
+    <description>Dataset</description>
+    <requirements>
+        <requirement type="package" version="8.25">coreutils</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        awk -F $'\t' ' { 
+        #if str($feature_text):
+        f = "${feature_text}";
```
bgruening: single quotes? Otherwise, we need to sanitize the text field?

tools/intermine_galaxy_exchange/intermine_galaxy_exchange.xml
```diff
@@ -0,0 +1,93 @@
...
+            <param name="feature_text" value=""/>
+            <param name="forganism_column" value="2"/>
+            <param name="organism_text" value=""/>
+            <output name="tabular_output" file="intermine-format-example.tsv" ftype="intermine_tabular"/>
+        </test>
+    </tests>
+    <help>
+        <![CDATA[
+
+https://raw.githubusercontent.com/intermine/galaxy-integration/master/intermine-file-format/intermine-format-example.tsv
```
bgruening: mh, this should be removed?

tools/intermine_galaxy_exchange/intermine_galaxy_exchange.xml
```diff
@@ -0,0 +1,93 @@
...
+or perhaps a protein accession, or some other identifier.
+
+## Column 3 (optional)
+**Organism name** - e.g. `H. sapiens` or `Drosophila melanogaster`.
+This is optional and does not have to be precise, but is good to provide if
+it is known.
+
+        ]]>
+    </help>
+    <citations>
```
bgruening: Intermine citation?

tools/intermine_galaxy_exchange/intermine_galaxy_exchange.xml
```diff
@@ -0,0 +1,93 @@
...
+
+
+
+-----
+
+## InterMine-Galaxy exchange format
+
+This file format exists as an intermediate step to import from Galaxy to
+InterMines and is not in itself a data format intended to be used for data
+storage or any purpose apart from Galaxy->InterMine interoperability.
```
bgruening: Mh, if this is the case can we not do this in the data_source tool?

tools/intermine_galaxy_exchange/intermine_galaxy_exchange.xml
yochannah: I'm not familiar with the correct formatting, but you can point here: InterMine: extensive web services for modern biology. Kalderimis A et al. (2014). Nucleic Acids Res. https://www.ncbi.nlm.nih.gov/pubmed/24753429

bgruening: This could work :)
```
<citation type="doi">10.1093/nar/gku301</citation>
```

tools/intermine_galaxy_exchange/intermine_galaxy_exchange.xml
yochannah: I think that data_source is hardcoded to a single tool, similar to https://github.com/galaxyproject/galaxy/blob/71cea6604d43c5fe6215f5656462ba6c1af69bb6/tools/data_source/flymine.xml - but with over 30 different intermines we didn't want to have to hardcode each one. The way @blankenberg did it allows us to workaround and let the use choose _any_ InterMine to port the data to using the InterMine registry.

tools/intermine_galaxy_exchange/intermine_galaxy_exchange.xml
blankenberg: Standard built-in sanitizer should work well, or are there additional concerns?

bgruening: Are there any downsides to use single quotes?

bgruening:
Thanks @blankenberg!

PR #1960, opened 2018-07-03
Add table compute (for general-purpose) table operations

mtekman:
This tool subselects and duplicates rows/columns, performs general and custom math or string functions on rows/columns/elements, and produces histograms.

 * ~~A custom environment stripped of all file operations is implemented for safety, but no expressions are evaluated within it at the moment due to an odd bug (vector methods do not appear to work on vectors within the environment).~~
 * ~~Ideally all table operations should run within this custom environment.~~
 * ~~For now, all operations run in the global environment.~~
 * Safe environment now fully implemented. 
 * Security testing is required before merge.


tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,457 @@
...
+                    <add value="&#93;"/> <!-- right square bracket -->
+                    <add value="&#40;"/> <!-- left parenthesis -->
+                    <add value="&#41;"/> <!-- right parenthesis -->
+                </valid>
+            </sanitizer>
+        </macro>
+    </macros>
+
+    <requirements>
+        <requirement type="package" version="3.1.2" >r</requirement>
```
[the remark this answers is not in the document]
mblue9: Yes, I tested this as is and all the tests fail for me with that R 3.1.2, if you change to 3.4.1 all except 1 test passes (last test for .png needs a sim_size)

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,457 @@
...
+            <output name="table" value="small.element.custom.tsv" />
+        </test>
+        <test><!-- Histogram -->
+            <expand macro="test_inputs" >
+                <param name="mode" value="histogram" />
+                <param name="histo_mode" value="log10" />
+                <param name="histo_lim_rowx" value="1,10" />
+                <param name="histo_lim_coly" value="1,3" />
+            </expand>
+            <output name="histogram" value="small.rowx1-10.coly1-3.png" />
```
mblue9: This test with the .png output fails, can you add compare="sim_size"

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,457 @@
...
+                </valid>
+            </sanitizer>
+        </macro>
+    </macros>
+
+    <requirements>
+        <requirement type="package" version="3.1.2" >r</requirement>
+        <!-- 3.2.2, and 3.4.3 appear to be broken... -->
+    </requirements>
+
```
mblue9: Add version_command here? e.g.
<version_command><![CDATA[
echo $(R --version | grep version | grep -v GNU)
    ]]></version_command>

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,457 @@
...
+            <output name="histogram" value="small.rowx1-10.coly1-3.png" />
+        </test>
+    </tests>
+    <help><![CDATA[
+This tool computes table expressions on the element, row, and column basis. It can subselect, duplicate, sort, perform general and custom expressions on rows/columns/elements, and produces histograms based on row and column sums.
+
+Only a single operation can be performed upon the data. Multiple operations can be performed by chaining successive runs of this tool. This is to provide a more transparent workflow for complex operations.
+
+]]></help>
+    <citations></citations>
```
mblue9: Is there any citation that can be put here?

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,457 @@
...
+                                        # END USER CONFIG
+        </configfile>
+    </configfiles>
+    <inputs>
+        <param name="input" type="data" format="tsv" label="Input Table" />
+        <param name="reader_header" type="boolean" checked="true" truevalue="T" falsevalue="F" label="Has header for column names?" />
+        <param name="reader_row_col" type="integer" value="-1" label="Column number for row names" help="Leave as -1 to force automatic row numbering" />
+        <conditional name="user" >
+            <param name="mode" type="select" label="Type of table operation" >
+                <option value="sort" selected="true" >Sort Row or Column</option>
```
mblue9: s/Row or Column/Rows or Columns ?

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,457 @@
...
+                <option value="histogram">Generate Histograms for Row sums and Column sums</option>
+            </param>
+
+            <when value="sort">
+                <param name="sort_mode" type="select" label="By" >
+                    <option value="rows" selected="true" >Rows</option>
+                    <option value="cols">Columns</option>
+                </param>
+
+                <param name="sort_fields" type="text" label="Fields to sort by" help="This should be a list of row or column ids seperated by comma, with the first number being the primary sorting field, and the second number being the secondary sorting field and so on. Negative numbers sort in descending order. (e.g. '-5,1,2' will sort by the 5th column in descending order, then by the 1st column in ascending order, and then by the 2nd column in ascending order, where columns are set to be the mode of sorting.)" >
```
mblue9: s/seperated by comma/separated by commas

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,457 @@
...
+                    <expand macro="sanitizer_vector" />
+                </param>
+            </when>
+
+            <when value="select">
+                <param name="select_cols_wanted" type="text" label="List of columns to keep" help="comma seperated. (e.g1. 1,5:10,6,3,99  will select columns 1, 5, 6, ..., 9, 10, 6 again, 3 again, and 99) ">
```
mblue9: s/seperated/separated

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,457 @@
...
+                    <expand macro="sanitizer_vector" />
+                </param>
+            </when>
+        </conditional>
+    </inputs>
+    <outputs>
+        <data name="table" format="tsv" >
+            <filter>user_mode != "histogram"</filter>
+        </data>
+        <data name="log" format="txt" >
```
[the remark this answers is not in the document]
mblue9: s/user_mode/user['mode']
(the filter isn't working so currently the histogram png is always output instead of only when histogram is selected)

mblue9: Could you add labels to label the different outputs? e.g. Table/Log/Histogram as they currently all output with the same name in the history "Table compute on data N"

mtekman:
@bgruening @mblue9 thanks for the comments, I think I got everything (except citations, I could not find anything suitable).

I'm happy with this as it is, and plan to use it soon in a workflow if possible.  There were mentions of:
 * Removing the histograms
 * Rewriting this in pandas.

I can remove the histogram option easily if it's really not wanted, but I am not planning a re-write in pandas anytime soon!

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,496 @@
...
+        <test><!-- umi2transcript test2 -->
+            <expand macro="umi2trans" >
+                <param name="input" value="mat2.umi.tsv" />
+                <output name="table" value="mat2.trans.tsv" />
+            </expand>
+        </test>
+    </tests>
+    <help><![CDATA[
+This tool computes table expressions on the element, row, and column basis. It can subselect, duplicate, sort, perform general and custom expressions on rows/columns/elements, and produces histograms based on row and column sums.
+
```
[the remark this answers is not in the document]
mtekman: examples are mostly in the help text for each mode, but okay I will add a few

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,496 @@
...
+            <when value="select">
+                <param name="select_cols_wanted" type="text" label="List of columns to keep" help="Comma separated. (e.g1. 1,5:10,6,3,99  will select columns 1, 5, 6, ..., 9, 10, 6 again, 3 again, and 99) ">
+                    <expand macro="sanitizer_vectorpositive" />
+                </param>
+                <param name="select_rows_wanted" type="text" label="List of rows to keep" >
+                    <expand macro="sanitizer_vectorpositive" />
+                </param>
+
+                <param name="select_cols_unique" type="boolean" truevalue="T" falsevalue="F" checked="false" label="No duplicate columns" help="Remove duplicates when specifying ranges (e.g. 1:5,3:6  will yield 1,2,3,4,5,6 instead of 1,2,3,4,5,3,4,5,6). Columns are also chosen in ascending order with this option" />
+                <param name="select_rows_unique" type="boolean" truevalue="T" falsevalue="F" checked="false" label="No duplicate rows" />
```
bgruening: help for rows is missing.

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,496 @@
...
+
+                <param name="sort_fields" type="text" label="Fields to sort by" help="This should be a list of row or column ids separated by commas, with the first number being the primary sorting field, and the second number being the secondary sorting field and so on. Negative numbers sort in descending order. (e.g. '-5,1,2' will sort by the 5th column in descending order, then by the 1st column in ascending order, and then by the 2nd column in ascending order, where columns are set to be the mode of sorting.)" >
+                    <expand macro="sanitizer_vector" />
+                </param>
+            </when>
+            <when value="select">
+                <param name="select_cols_wanted" type="text" label="List of columns to keep" help="Comma separated. (e.g1. 1,5:10,6,3,99  will select columns 1, 5, 6, ..., 9, 10, 6 again, 3 again, and 99) ">
+                    <expand macro="sanitizer_vectorpositive" />
+                </param>
+                <param name="select_rows_wanted" type="text" label="List of rows to keep" >
```
bgruening: help and mini-example like in columns is missing

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,496 @@
...
+            <option value="exp" >Exponential</option>
+            <option value="floor" >Floor</option>
+            <option value="ceiling" >Ceiling</option>
+            <option value="round" >Round</option>
+            <option value="sqrt" >Square root</option>
+            <option value="as.hexmode" >Hex</option>
+            <option value="as.logical" >Boolean</option>
+            <option value="as.character" >Character</option>
+            <option value="as.octmode" >Oct</option>
+            <option value="toString" >String</option>
```
bgruening: `to string` also in the description?

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,496 @@
...
+
+        <!-- select macros -->
+        <macro name="select_onevalueoperator">
+            <option value="log" >Ln</option>
+            <option value="log10" >Log10</option>
+            <option value="abs" >Absolute</option>
+            <option value="exp" >Exponential</option>
+            <option value="floor" >Floor</option>
+            <option value="ceiling" >Ceiling</option>
+            <option value="round" >Round</option>
```
bgruening: Round does what in comparision to Ceil and Floor?

bebatut:
I am testing the tool currently: it is really useful! Thanks a lot for that!!!!!!!

I had a bug: The custom function of rows does not work with: `mean(vec)` or `sd(vec)`:

   ```
    Fatal error: Exit code 1 ()
    Error in UseMethod("mean") : 
      no applicable method for 'mean' applied to an object of class "c('double', 'numeric')"
    Calls: evaluate -> evaluate -> eval -> eval -> apply -> FUN -> mean
    Execution halted
    ```

Few "requests":
- Having as the column name the type of function used?
- Having several functions to create several columns/rows?
- Giving the possibility to have something like: `table - apply(table, 1, mean)/apply(table, 1, sd)`

mtekman:
Ah, okay I must have not added the right mean function to the safe evaluation environment.

The requests:
 * Sure, will do
 * Like an optional input in the "Compute Expression across Rows or Columns" section, with the ability to `cbind` or `rbind`? Would that work?
 * Ah, did not consider that use-case. Okay, I think I will add a Full Table Operations mode, and add a custom section which would allow for that.

Thanks for the review!

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,645 @@
...
+In Galaxy we would select the following:
+
+ * *Has header for column names?* → **Yes**
+ * *First column provides row names?* → **Yes**
+ * *Type of table operation* →  **Perform a Full Table Operation**
+   * *Custom Expression* →
+
+::
+
+         return(table - apply(table, 1, mean)/apply(table, 1, sd))
```
bebatut: The `return` is not needed and may disturb some users

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,645 @@
...
+                    <when value="modify" >
+                        <param name="modify_op" type="select" label="Operation to modify selected elements" help="e.g. to square root all values, select 'Square Root'" >
+                            <expand macro="select_onevalueoperator" />
+                        </param>
+                    </when>
+                    <when value="scale">
+                        <param name="scale_op" type="select" label="Operation to scale selected elements by" help="e.g. to raise all values to the power of 5, select 'Power' and and type in '5' in the box below" >
+                            <expand macro="select_twovaluenumericoperator" />
+                        </param>
+                        <param name="scale_value" type="float" label="Value to scale using selected operation" value="0" />
```
bebatut: Can you move the `value` before the `label`? Thanks

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,645 @@
...
+            <option value="neq" >Not Equal to</option>
+            <yield />
+        </macro>
+        <macro name="select_applyops">
+            <option value="mean" />
+            <option value="max" />
+            <option value="min" />
+            <option value="sum" />
+            <option value="median" />
+            <option value="summary" />
```
bebatut: Could you also add `sd` there? Thanks

tools/table_compute/table_compute.xml
mtekman: For single line evaluation, I agree -- but what about multi-line code with variables?  
e.g.  

    tmp1 <- apply(table, 1, mean)
    tmp2 <- apply(table, 2, sd)
    return(table - tmp1/tmp2)

One solution is that I could wrap the last line of a custom expression with a `return`, but I do not know how bulletproof this would be.

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,785 @@
...
+                            </param>
+                            <when value="T">
+                                <param name="custom_func" type="text" label="Custom function" help="The parameter name is 'vec', referring to a vector of a specific row or column. Most operators and if/else statements are supported. e.g. sum(vec) + median(vec)" >
+                                    <expand macro="sanitizer_functiondef" />
+                                </param>
+                            </when>
+                            <when value="F">
+                                <param name="inbuilt_op" type="select" label="In-built function">
+                                    <option value="mean" >Mean</option>
+                                    <option value="sd" >Standard Devation</option>
```
bebatut: s/Standard Devation/Standard Deviation

bebatut: It is failing when I tried to use it:

```
Error in getVectorNumericOperator(evaluate(se, matrixapply.op)) : 
  sd is not a vector operator
Calls: addToEnv -> addToEnv -> getVectorNumericOperator
```

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,785 @@
...
+    #end for
+    fulltable.customop = trimws('$singtabop.fulltable_custom_expr.value')
+#end if
+
+outtable = '$table'
+]]>
+        </configfile>
+    </configfiles>
+    <inputs>
+        <param name="reader_header" type="boolean" checked="true" label="Has header for column names?" />
```
bebatut: With the multiple table, you probably need the 3 next parameters for each table

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,785 @@
...
+        <param name="reader_row_col" type="boolean" checked="true" label="First column provides row names?" help="Rows will be automatically numbered if false." />
+        <param name="keep_nas" type="boolean" checked="true" label="Ignore NA values" />
+
+        <conditional name="singtabop" >
+            <param name="use_type" type="select" label="Table operation acting upon" help="Operations on single tables are better tailored towards more general use-cases. Multiple table operations must be specified manually by the user." >
+                <option value="single">Single Table</option>
+                <option value="multiple">Multiple Tables</option>
```
bebatut: I tried to add / substract two tables: one with 131 lines and 15 columns and the other with 131 lines but 1 column.
It failed with:
```
Error in Ops.data.frame(table1, table2) : 
  ‘+’ only defined for equally-sized data frames
```
which is expected...

But if the matrix with one column is managed as a vector, it should work.
But I am not sure if it is a real use case

tools/table_compute/table_compute.xml
mtekman: oh right, forgot to add it to that list

tools/table_compute/table_compute.xml
mtekman: Yeah I thought about this, and reasoned that the user wouldn't upload tables of different formats.... but I can implement this easily.

tools/table_compute/table_compute.xml
mtekman: Ah okay, I will add a check that converts a 1 column table to a vector

mtekman:
Huh, that's so strange -- the test with the custom function name passed. Okay I will dig into that.

Well, I mentioned that table operation on different dimensions will not always work in Example 6 in the Help text, and I reference example 6 in the actual parameter. Should I be more explicit then?

mtekman:
@bebatut  can you give me a test set for the failing function name? I cannot reproduce on my side

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,800 @@
+<tool id="table_compute" name="Table Compute" version="@VERSION@.1">
```
mblue9: Should this just be @VERSION@ without the .1 if it's the first version?

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,800 @@
...
+    <description>computes operations upon table data</description>
```
mblue9: s/upon/on

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,800 @@
...
+    fulltable.customop = trimws('$singtabop.fulltable_custom_expr.value')
+#end if
+
+outtable = '$table'
+]]>
+        </configfile>
+    </configfiles>
+    <inputs>
+        <conditional name="singtabop" >
+            <param name="use_type" type="select" label="Table operation acting upon" help="Operations on single tables are better tailored towards more general use-cases. For multiple tables usage, all tables should ideally be of the same dimensions for simple operations, although complex operations are also supported for unequally sized tables (see Help section)." >
```
mblue9: "Table operation acting upon" sounds a little strange to me, what about something like "Input Single or Multiple Tables" or "Operate on Single or Multiple Tables"

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,800 @@
...
+ === ===== ===== =====
+  .     c1    c2    c3
+ === ===== ===== =====
+ g1  FALSE FALSE FALSE
+ g2  TRUE  TRUE  TRUE
+ g3  TRUE  TRUE  FALSE
+ g4  FALSE TRUE  TRUE
+ === ===== ===== =====
+
+To get to our desired table, we would then run this table through galaxy once more:
```
mblue9: s/galaxy/Galaxy

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,800 @@
...
+                </repeat>
+                <param name="fulltable_custom_expr" value="(3 * table1) + log(table3 + 20) - table2" />
+            </conditional>
+            <output name="table" value="small.multiple.tsv" />
+        </test>
+    </tests>
+    <help><![CDATA[
+This tool computes table expressions on the element, row, and column basis. It can subselect, duplicate, as well as perform general and custom expressions on rows/columns/elements.
+
+Only a single operation can be performed upon the data. Multiple operations can be performed by chaining successive runs of this tool. This is to provide a more transparent workflow for complex operations.
```
mblue9: s/upon/on

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,800 @@
...
+ g2   3   3   9
+ g3   4   4  12
+ g2   3   3   9
+ === === === ===
+
+In Galaxy we would select the following:
+
+ * *Has header for column names?* → **Yes**
+ * *First column provides row names?* → **Yes**
+ * *Table operation acting upon* → **Single Table**
```
mblue9: If you end up changing that label "Table operation acting upon" then these entries in the help would need to be changed to match.

tools/table_compute/table_compute.xml
```diff
@@ -0,0 +1,800 @@
...
+ * *Type of table operation* →  **Perform a Full Table Operation**
+   * *Custom Expression* →
+
+::
+
+         table - apply(table, 1, mean)/apply(table, 1, sd)
+
+
+Example 6: Perform operations upon multiple tables
```
mblue9: s/upon/on

PR #1963, opened 2018-07-04
samtools view update to v1.9

bernt-matthias:
This updates samtools view to v1.9

- [x]  upgrade to 1.9

Discussion:

- [ ] Do we need samtools view based tools: slice_bam, bam_to_cram, bam_to_sam, cram_to_bam, and sam_to_bam. The logic with reference data and indices could potentially integrated into samtools_view.
  - keep converters for compatibily, remove/deprecate slice
    - [ ] mark slice deprecated / remove it? How? In any case move the tests here.
  - Note that I would have also the updated tools which I could push here, but lets decide first
    - [x] No need to push. I checked: no changes are necessary.
- [x] tool version is now 2.xx can we change it to 1.7+galaxy2.xx?
  - [x] if no problem for galaxy. is it?
    - not for this tool, because new
- [x] macro PREPIDX
  - [x] for cram there was a test for the presence of the index, do we need this also for bam
    - [x] implement this
  - [x] do we want indexing for SAM files? then tabix is needed which seems out-of-date in bioconda (0.2.5 available, latest 1.9 in htslib
    - [x] no indexing for sam files
- [ ] do we need to allow explicitely for the different BAM data types, I guess we need them all.
  - [ ] yep we need them: Q: do we need to allow them explicitely?
  - [ ] in PREPIDX: how can we check for the other types? is there a simple way. 
- [ ] check logic of setting outtype. also if statements for other types necessary? Can we simplify this?
- [x] readtags as repeat or text field
  - repeat is fine
- [x] input of reference fasta: allow fai? planemo complains about fasta.gz
  - [x] fix this!
- [ ] output sam: dbkey action needed? change_format needs to be replaced or improved
change_format deprecated?


tool_collections/samtools/samtools_view/samtools_view.xml
```diff
@@ -0,0 +1,438 @@
...
+
+        infile
+	
+	## region filter need to be at the end
+	#if $cond_filter.select_filter == 'yes' and $cond_filter.cond_region.select_region == 'text':
+		'$cond_filter.cond_region.regions'
+        #end if
+
+## if data is converted from an unsorted file (SAM, CRAM, or unsorted BAM) to BAM
+## then sort the output by coordinate,
```
bwlang: not so sure about the auto sorting... maybe make it more explicit by asking the user to choose from SAM, CRAM (maybe sort versions of cram - I don't use them?), unsorted.bam, qname_sorted.bam, indexed bam (coord sorted). With the default being indexed bam?

bwlang:
Commendable work on these important tools!

I'll respond to your discussion points in order:
1. I think we should keep the format conversion tools (sam <-> bam <-> cram) for usability/dicoverability reasons. Slice could be eliminated.
1. New version scheme seems good to me, assuming that we detect 1.7+g is > 2.x

1. I don't know about cram indexes, but I think we can assume in galaxy that all files of type "bam" are indexed. I say no indexing for sam files 

1. Would be nice to add the other galaxy bam input types(that are not indexed)  for tools like view that don't need an index:
     - qname_input_sorted.bam
     - qname_sorted.bam
     - unsorted.bam
1. I think the tool should ask about output type (.e.g. bam, sam, cram, indexed bam(implies coord sorting), indexed cram) and remove the "same" type - which, IMHO, adds complexity without much benefit. It would be tough to render a workflow from a tool using a same type. (left an inline comment to this effect)
1. Read tags support looks fine to me as a repeat
1. I think it's reasonable to require an fa even if the tool only *really* needs the index. fasta.gz should be supported
1. Sorry - I don't know about the change_format bits - maybe someone else will

bernt-matthias:
> I think we should keep the format conversion tools (sam <-> bam <-> cram) for usability/dicoverability reasons. Slice could be eliminated.

Agreed. What would be the IUC way to remove slice? Delete it? Or mark it deprecated..? How?

>     New version scheme seems good to me, assuming that we detect 1.7+g is > 2.x

Jep. Can someone answer if this works? Otherwise I would suggest to increase the version slowly, such that samtools version will catch up eventually. 

>    I don't know about cram indexes, but I think we can assume in galaxy that all files of type "bam" are indexed. I say no indexing for sam files

OK. 

>     Would be nice to add the other galaxy bam input types(that are not indexed) for tools like view that don't need an index:
>         qname_input_sorted.bam
>         qname_sorted.bam
>         unsorted.bam

OK. Do we need to add them explicitely? Or are they implicitely allowed anyway, because of the data types class structure?

>     I think the tool should ask about output type (.e.g. bam, sam, cram, indexed bam(implies coord sorting), indexed cram) and remove the "same" type - which, IMHO, adds complexity without much benefit. It would be tough to render a workflow from a tool using a same type. (left an inline comment to this effect)

OK

>     Read tags support looks fine to me as a repeat

OK

>     I think it's reasonable to require an fa even if the tool only really needs the index. fasta.gz should be supported

I will check

    Sorry - I don't know about the change_format bits - maybe someone else will

tool_collections/samtools/samtools_view/samtools_view.xml
```diff
@@ -0,0 +1,438 @@
+<tool id="samtools_view" name="Samtools view" version="2.0.1">
```
nsoranzo: `version="@TOOL_VERSION@"`

nsoranzo:
> >     New version scheme seems good to me, assuming that we detect 1.7+g is > 2.x

> Jep. Can someone answer if this works? Otherwise I would suggest to increase the version slowly, such that samtools version will catch up eventually. 

This doesn't apply here, since this is a new tool. Any way `2.0` > `1.7+g`, which can be verified with:
```python
>>> from packaging.version import parse
>>> parse('2.0') > parse('1.7+g')
True
```

bernt-matthias:
Somehow I have a problem with planemo on my local install (only this branch): 

All tests fail immediately, I just get:

```
  File "/tmp/tmpH13feS/galaxy-dev/test/functional/test_toolbox.py", line 88, in test_tool
    self.do_it(tool_version=tool_version, test_index=test_index)
  File "/tmp/tmpH13feS/galaxy-dev/test/functional/test_toolbox.py", line 41, in do_it
    verify_tool(tool_id, self.galaxy_interactor, resource_parameters=resource_parameters, test_index=test_index, tool_version=tool_version, register_job_data=register_job_data)
  File "/tmp/tmpH13feS/galaxy-dev/lib/galaxy/tools/verify/interactor.py", line 653, in verify_tool
    stage_data_in_history(galaxy_interactor, tool_id, testdef.test_data(), test_history)
  File "/tmp/tmpH13feS/galaxy-dev/lib/galaxy/tools/verify/interactor.py", line 59, in stage_data_in_history
    upload_wait = galaxy_interactor.stage_data_async(test_data, history, tool_id)
  File "/tmp/tmpH13feS/galaxy-dev/lib/galaxy/tools/verify/interactor.py", line 267, in stage_data_async
    file_name = self.test_data_path(tool_id, fname)
  File "/tmp/tmpH13feS/galaxy-dev/lib/galaxy/tools/verify/interactor.py", line 232, in test_data_path
    assert response.status_code == 200
AssertionError
```

I also tried to reset my venv and planemo  configuration directories. I'm completely stuck on this one. Any ideas?

bernt-matthias:
got planemo running again (problem was that I linked to the test data of the converters). my todos are done. the remaining points are open questions.

tool_collections/samtools/samtools_view/samtools_view.xml
bernt-matthias: So you mean that we need to remove: ` or $input.metadata.sort_order == 'queryname'` 

Otherwise I hope that its correct. The main point is that we need to coordinate sort if the output is bam (by assumptions of Galaxys bam data type). I just check if the input is sorted .. because then we can skip this. Just not sure if the metadata of the bam data type is set correctly.

tool_collections/samtools/samtools_view/samtools_view.xml
bernt-matthias: done

bgruening:
I think I would not include the different BAM datatypes here. My intuition is we should keep the tools simple in this regard and offer one sort tool that can create the other datatypes if needed.

Also please note, that Galaxy ships by default with a few converters, so BAM to CRAM and so on should be already in Galaxy.

tool_collections/samtools/samtools_view/samtools_view.xml
bwlang: I think user should specify what they want for an output format (or choose the default -which i think shoudl be bam). Since the output format implies the sort order, think that alone shoud determine whether or not to sort (e.g.  sort for at least bam, maybe cram too? )

bwlang:
@bgruening : I value simplicity too, but in this case I wonder whether the cost of duplicating the potentially very large data files for a format conversion is worth a bit more code complexity.

I don't think having a list of 4 output types (sam, bam, cram, count) is much more complex for users than "bam (coordinate sorted), cram (coordinate sorted), sam, unsorted.bam, qname_input_sorted.bam, qname_sorted.bam, unsorted.bam,count"

Or maybe you were referring to the input formats?  I think there is even less user complexity if that's your concern.

bernt-matthias:
@bwlang has a point with the data duplication, but users can decide to delete intermediate data (in workflows you can do this automatically).

bwlang:
agreed - also, these other bam formats are proably not typically used much...

tool_collections/samtools/samtools_view/samtools_view.xml
```diff
@@ -0,0 +1,416 @@
...
+        $adv_output.header
+        $adv_output.collapsecigar
+        #if $adv_output.outputpassing == 'yes'
+            -U inv_outfile
+        #end if
+        -o outfile
+
+        ## additional reference data
+        #if $reffa!=None:
+            -T '$reffa'
```
mvdbeek: This doesn't really work and is reproducible only on the galaxy server you are working on as it stores the path to the index. And without the index and not knowing what exact reference had been used the file is useless. For now if the output is CRAM it's always going to have to be `--output-fmt-option no_ref`.

mvdbeek:
@bernt-matthias do you mind if I rebase this against the new macros.xml file and address my last comment ?

bernt-matthias:
I merged the macros.xml file using the github frontend and commited an update using nicer names for the new macros.

mvdbeek:
thanks, looks good. what do you think about https://github.com/galaxyproject/tools-iuc/pull/1963#discussion_r207997309 ? I'm a little worried about users generating CRAM files and then being disappointed that they don't work when downloading them from Galaxy.

tool_collections/samtools/samtools_view/samtools_view.xml
bernt-matthias: Sorry for the late answer. Somehow github does not inform me on code comments, but only if there are comments to the PR.

First of all I'm open to any suggestion. Still I would like to dig into this -- I'm sure I can learn something :)

> This doesn't really work

The tests indicate that it works..?

> reproducible only on the galaxy server you are working on as it stores the path to the reference. 

Hmm. I have to admit that I do not know exactly what samtools/CRAM is doing with the reference (even after studying the manual). My guess is that the @SQ (sequence names, length, md5,...) headers are set/checked based on the data. Is the path to the sequence actually stored in the CRAM file? 

Furthermore, according to the samtools manual (but this is often not really helpful): 

```
The -t and -T options provide additional reference data. One **of these two options is required when SAM input does not contain @SQ headers**, and the **-T option is required whenever writing CRAM output**. 

noref:     CRAM output only; defaults to 0 (off). If 1, sequences will be stored verbatim with no reference encoding. This can be useful if no reference is available for the file. 
```

Is the CRAM format still useful if sequences are stored verbatim? 

If the path to the reference is stored in the CRAM file I agree, that there might be an issue with reproducibility. But I'm not sure about this. Can you give me an example.

bernt-matthias:
> I'm a little worried about users generating CRAM files and then being disappointed that they don't work when downloading them from Galaxy.

You have a point with this. But this might be treated by $reffa!=None which adds `--output-fmt-option no_ref` (note this is the default). One could state in the help (which needs some love anyway) that in case the user wants to use CRAM outside of Galaxy without the reference available then no refernce must be specified in the tool. But I think that programs using CRAM as input should allow to specify the path to the reference (and the md5 check verifies that its the correct reference).

bernt-matthias:
It seems that only the md5 is stored in CRAM. 

https://www.htslib.org/workflow/#mapping_to_cram : 

```
In CRAM format the reference sequence is linked to by the md5sum (M5 auxiliary tag) in the CRAM header (@SQ tags). This is mandatory and part of the CRAM specification.
```

bernt-matthias:
@mvdbeek took some time to provide a somewhat useful help.

tool_collections/samtools/samtools_view/samtools_view.xml
```diff
@@ -401,16 +429,18 @@ Examples of region specifications:
...
+This filters based on the MAPQ column of the SAM format. 
 
-The -t and -T options provide additional reference data. One of these two options is required when SAM input does not contain @SQ headers, and the -T option is required whenever writing CRAM output.
+.. class:: infomark 
 
-The -L, -r, -R, -q, -l, -m, -f, and -F options filter the alignments that will be included in the output to only those alignments that match certain criteria.
+Note that, 
 
-The -x, -B, and -s options modify the data which is contained in each alignment.
+- the SAM specification defines the MAPQ column insufficiently: as the PHRED scaled probability that a **position** is wrong -- which does not define at all how this value is computed for an alignment.
```
mvdbeek: I think we don't need to judge the SAM specification or aligners here. Users that filter on MAPQ will know this, and those that don't know will not understand if we don't explain what MAPQ is (which si not the scope of the tool help IMO)

tool_collections/samtools/samtools_view/samtools_view.xml
```diff
@@ -385,7 +384,36 @@
...
+
+**Alignment format conversion**
+
+Inputs of type SAM, BAM, and CRAM are accepted and can be converted to each of these formats (alternatively alignment counts can be computed) by selecting the appropriate "Output type". 
+
+.. class:: infomark 
+
+samtools view allows to specify a reference sequence. This is required for SAM input with missing @SQ headers (which include sequence names, length, md5, etc) and useful (and sometimes necessary) for CRAM input and output. In the following the use of reference sequence in the CRAM format is detailed. 
+CRAM is (primarily) a reference-based compressed format, i.e. only differences between the stored sequences and the reference are stored. As a consequence the reference that was used to generate the alignemnts is always needed in order to interpret the alignments (a checksum stored in the CRAM file is used to verify that the only the correct sequence can be used), i.e. the CRAM file on its own is not useful per default. This allows for a more space efficient storage compared to BAM. But it is also possible to use CRAM without a reference with the disadvantage that the reference is stored explicitely (as in SAM and BAM).
```
mvdbeek: The part about the checksum is right, but also the path or URL to the reference is stored in the CRAM file. So IMO we should just disable reference based compression, as there is no easy way to look at the alignment file for a user.

tool_collections/samtools/samtools_view/samtools_view.xml
mvdbeek: Yeah, just take the test file and try samtools view on another machine:
```samtools view test.cram                                                                                                                                                                                                         
[E::cram_get_ref] Failed to populate reference for id 0
[E::cram_decode_slice] Unable to fetch reference #0 2..100
[E::cram_next_slice] Failure to decode slice
[main_samview] truncated file.
```

tool_collections/samtools/samtools_view/samtools_view.xml
bwlang: I agree - no judgement needed. However many people wrongly think that MapQ is about how closely a sequence matches the reference.  Might be nice to mention that it's about correct placement instead and that aligners do not follow a consistent definition.

tool_collections/samtools/samtools_view/samtools_view.xml
bwlang: I agree - ref based compression is cool, but portability issues make it hard to justify

tool_collections/samtools/samtools_view/samtools_view.xml
bernt-matthias: LOL. Surprisingly it worked for me.. So I ran strace and found out that there is a cache ($HOME/.cache/hts-ref/73/2d/956b0927aa063007fb18c839890a) because the default is to get the sequences from EBI .. which is also described in the samtools manual. But, one can always overwrite this behaviour using -T. 

In the light of these side effects (which I have not thought of so far) I agree to disable these features for now.  Seems just to much work for the moment for me. 

But what to disable: remove all reference specification (also for BAM and SAM)? What are the reference sequences good for in SAM and BAM?

tool_collections/samtools/samtools_view/samtools_view.xml
bernt-matthias: OK. Good suggestion

tool_collections/samtools/samtools_view/samtools_view.xml
bernt-matthias: done. nice suggestion.

tool_collections/samtools/samtools_view/samtools_view.xml
mvdbeek: IIRC they will re-write the header, in case you got a broken BAM or SAM file. I'd just disable the reference-based compression

bernt-matthias: OK. I see two options to do this: 

1. have a conditional for the input type selector and show the refernce input only for BAM/SAM input.
2. just add the `--output-fmt-option no_ref` if CRAM is selected (mayme still allowing -t/-T?)

mvdbeek: I'd prefer option 2, this should be the simpler user experience

bernt-matthias: OK. Done. Also updated the help accordingly.

mvdbeek:
We still need a .shed.yml for samtools view.

mvdbeek:
@bernt-matthias could you add the .shed.yml file ?

bernt-matthias:
Done

tool_collections/samtools/samtools_view/samtools_view.xml
```diff
@@ -0,0 +1,439 @@
...
+                <param name="subsample" type="float" optional="False" value="0.5" min="0" max="1" label="Subsampling fraction" help="sets the fraction of templates/pairs to subsample." />
+                <param name="seed" type="integer" optional="True" label="Seed for random number generator" help="If empty a random seed is used." />
+            </when>
+            <when value="no"/>
+        </conditional>
+        <conditional name="addref_cond">
+            <param name="addref_select" type="select" label="Reference data">
+		    <help>Reference data as fasta(.gz). Required for SAM input without @SQ headers and useful/required for writing CRAM output (see help).</help>
+                <option value="no, see help (-output-fmt-option no_ref)" selected="True">No</option>
+		<option value="history (-t/-T)">Use a genome/index from the history</option>
```
mvdbeek: I think you want to make this:
```
		<option value="history">Use a genome/index from the history</option>
		<option value="cached">Use a built-in genome</option>
```
That's why the tests are failing

bernt-matthias: Indeed, stupid mistake.

PR #1964, opened 2018-07-04
samtools sort: upgrade to v1.9

bernt-matthias:
upgrade to version 1.9. new options + tests from samtools sources.
- [x]  upgrade to 1.9
- [x] fix samtools sort memory requirement https://github.com/samtools/samtools/issues/831 + also change to GALAXY_MEMORY_PER_SLOT
  - [x] should we? as suggested in the issue only 75% of the available mem should be used
- [x] check tool versions (currently 2.0.1. can it be changed to 1.7+galaxy3.0 ... or similar)
  - set to 2.0.2
- [x] should the input be bam / unsorted.bam? or do we need to allow all types explicitly anyway?
  - should be fine (see discussion below)
- [x] do we want sam/cram output
  - should be fine (see discussion below)
- [x] most likely unsorted bam can not be fed into any other samtool at the moment..?
  - prefer to adapt downstream tools (see discussion below)
- [x] for tag sorted bam format is currently unsorted.bam. or should a tag sorted BAM datatype be added OR dont allow for tag sorting in the Galaxy tool? 
  - leave it for the moment

I guess it would be worth to discuss the state of the art of Galaxy's sam/bam/cram data types. In my opinion a single sam/bam/cram data type with properly set metadata should be sufficient. Tools can access metadata, or? Otherwise for every combination of primary/secondary sort key a data type would be needed. .. But I guess this should go to a Galaxy issue :)


tool_collections/samtools/samtools_sort/samtools_sort.xml
```diff
@@ -7,39 +7,164 @@
...
+                    <option value="">coordinate</option>
+                    <option value="-n">name (-n)</option>
+                </param>
+           </when>
+        </conditional>
+        <!--<param name="compression" type="integer" argument="-l" optional="True" min="0" max="9" label="compression level" help="0 (uncompressed) to 9 (best)"/>-->
     </inputs>
     <outputs>
-        <data name="output1" format="bam" />
+       <!-- TODO do we want sam/cram output -->
```
mvdbeek: Ultimately I guess we'll want CRAM as well, but let's keep it simple for the start.

mvdbeek:
> should the input be bam / unsorted.bam ?

Unsorted bam should be fine, this will accept everything without conversions.

> do we want sam/cram output

Let's keep it simple and revisit that later on. Might need changes to the CRMA datatype anyway

> most likely unsorted bam can not be fed into any other samtool at the moment..?

That depends, without implicit conversion that is correct for tools that demand `bam` for instance. Is that a concern ? Those tools that can accept additional formats need to be adapted.


> for tag sorted bam format is currently unsorted.bam. or should a tag sorted BAM datatype be added OR dont allow for tag sorting in the Galaxy tool?

If we can come up with a useful datatype, sure, let's add that. Otherwise unsorted.bam is fine if the downstream tool does its own verification.

> In my opinion a single sam/bam/cram data type with properly set metadata should be sufficient

That'd be ideal, but probably be quite a big effort. I guess the momentum of the discussion is that we convert as part of the job if the metadata isn't what we'd like it to be. If that would be inefficient (i.e tag sorted output, but 10 jobs consuming coordinate-sorted that are going to consume the tag sorted file) the user can always explicitly convert. There's some discussion in https://github.com/galaxyproject/galaxy/issues/5790 (especially about eliminating implicit conversions) that centers on compressed or uncompressed, but at large this also applies to metadata differences IMO.

bernt-matthias:
ready from my side

mvdbeek:
The macros are conflicting now

bernt-matthias:
fixed

PR #1965, opened 2018-07-04
samtools merge: new tool in v1.9

bernt-matthias:
adds missing `samtools merge`
- [x]  upgrade to 1.9
- [ ] since I guess this will be the last merged file I will retest all tools since macro rebasing/merging might have gone wrong. Also make sure that now all tools are v 1.9
- [ ] can we restrict the input to sorted alignments (ie bam and maybe qname sorted bam)? Would simplify logic.
- [ ] it would be really nice to ignore all the sam/bam/cram data types that are currently implemented in galaxy and rather get the sort order from the metadata. is this possible?
- [ ] force overwrite -f seems necessary (but I do not understand why ...)
- [ ] sorting by TAG currently not supported since there is no data type to support this
- [ ] -r can set RG info based on file name. can we access tags/dataset names/etc to make this useful?
  - use maybe `re.sub('[^\s\w\-]', '_', str($file.element_identifier)`
- [ ] support sam, bam, and cram input? or even more?
- [ ] can we set the output sort order based on the input sort order
- [ ] while checking for inconsistently sorted input I output an error to stderr if it fails. should I just call false here?


tool_collections/samtools/samtools_merge/samtools_merge.xml
```diff
@@ -0,0 +1,130 @@
...
+        ## TODO -r makes no sense with the link names, is there some data set metadata (tags,...) that could be used?
+        $idrg
+        $idpg
+        $output
+        #for $i, $bam in enumerate( $bamfiles ):
+            ${i}.sam
+        #end for
+    ]]></command>
+    <inputs>
+        <!-- TODO I guess that sam and bam are valid, but documentation says only bam? -->
```
mvdbeek: Looks like that worked in the tests, so can you drop the comment or change it to
`documentation says only bam, but sam and bam are valid` ?

tool_collections/samtools/samtools_merge/samtools_merge.xml
```diff
@@ -0,0 +1,130 @@
...
+        ##         #end if
+        #if str($headerbam) != 'None'
+            -h '$headerbam'
+        #end if
+        #if $sortby=='name'
+            -n
+        #end if
+        ## TODO since galaxy can't represent this as data type at the moment this option is unsupported
+        ## -t TAG    The input alignments have been sorted by the value of TAG, then by either position or name (if -n is given). 
+	#if str($region) != ''
```
mvdbeek: The indentation is a little off here, and overall I think we prefer completely left-aligned command blocks

bernt-matthias: So no spaces on the left? But, `if` and `for` still indented?

mvdbeek: Right, so that we make best use of the screen estate. Yes, control statements still indented.

tool_collections/samtools/samtools_merge/samtools_merge.xml
bernt-matthias: done

tool_collections/samtools/samtools_merge/samtools_merge.xml
bernt-matthias: done. will follow this scheme in the future.

PR #1967, opened 2018-07-04
samtools fixmate and markdup in v1.9

bernt-matthias:
- [x]  upgrade to 1.9
- [ ] `samtools markdup`
  - [ ] implement also the negative tests
- `samtools fixmate`
  - [ ] output change_format help needed
- [ ] old rmdup tool needs to be removed or marked obsolete


tool_collections/samtools/samtools_fixmate/samtools_fixmate.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="samtools_fixmate" name="Samtools fixmate" version="2.0.1">
+<tool id="samtools_fixmate" name="Samtools fixmate" version="@TOOL_VERSION@">
```
bgruening: We can not do this here, as the previous version was higher than the TOOL_VERSION (1.7).

bernt-matthias: for this tool we could, since its a new one. 

or do we want the same tool version for all samtools?

bgruening: No this is not needed, but you did this for all other tools as well, isn't it?

bernt-matthias: I tried not to, but maybe by mistake. 

In https://github.com/galaxyproject/tools-iuc/pull/1971 I forgot to use the token, but at least set it to v1.8

tool_collections/samtools/samtools_fixmate/samtools_fixmate.xml
```diff
@@ -0,0 +1,99 @@
...
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements"/>
+    <expand macro="stdio"/>
+    <expand macro="version_command"/>
+    <command><![CDATA[
+        @ADDTHREADS@
+        ## name sort input 
+        #if not $bamfile.is_of_type('qname_sorted.bam'):
```
mvdbeek: `['qname_sorted.bam', 'qnamed_input_sorted.bam']`

tool_collections/samtools/samtools_fixmate/samtools_fixmate.xml
```diff
@@ -0,0 +1,99 @@
...
+        <param name="bamfile" type="data" format="sam,bam,cram" optional="false" label="Alignment in BAM format" help="Set of aligned reads." />
+        <param name="remsec" type="boolean" argument="-r" truevalue="-r" falsevalue="" checked="false" label="Remove secondary and unmapped reads" />
+        <param name="noprop" type="boolean" argument="-p" truevalue="-p" falsevalue="" checked="false" label="Disable FR proper pair check" />
+        <param name="tempcigar" type="boolean" argument="-c" truevalue="-c" falsevalue="" checked="false" label="Add template cigar ct tag" />
+        <param name="addms" type="boolean" argument="-m" truevalue="-m" falsevalue="" checked="false" label="Add ms (mate score) tags" help="These are used by markdup to select the best reads to keep." />
+    </inputs>
+    <outputs>
+        <data name="output" format="sam" />
+	<!--
+        TODO I would prefer this (in combination with -O bam, but I could not get the test running, apparently the binary files differ slightly in size but the SAM content is identical). Alternative: have a select for the out type
```
mvdbeek: Yes please, sam isn't practical anymore ... . If you restore the test I'll figure out what's different. Maybe the testing framework needs some work for this, or maybe a combonation of `sort`, `compare` and `lines_diff` may be needed (https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-tests-test-output).

tool_collections/samtools/samtools_fixmate/samtools_fixmate.xml
bernt-matthias: done

tool_collections/samtools/samtools_fixmate/samtools_fixmate.xml
bernt-matthias: ok I changed it to bam. test do fail now. file sizes are identical (maybe this would be sufficient?). I guess the bam format stores some information that changes - seemingly timestamps are incuded..?

mvdbeek:
can you add a `.shed.yml` for fixmate to get the tests started on travis ?

bernt-matthias:
> can you add a .shed.yml for fixmate to get the tests started on travis ?

done

tool_collections/samtools/samtools_fixmate/samtools_fixmate.xml
mvdbeek: No, the text representation is exactly the same. I think samtools from bioconda now links against a new compression library, so probably only the compression differs. I'll push a commit with updated test data.

tool_collections/samtools/samtools_fixmate/samtools_fixmate.xml
```diff
@@ -0,0 +1,97 @@
...
+            $remsec
+            $noprop
+            $tempcigar
+            $addms
+            -O BAM
+            namesorted.bam
+            '$output'
+    ]]></command>
+    <inputs>
+        <param name="bamfile" type="data" format="sam,bam,cram" optional="false" label="Alignment in BAM format" help="Set of aligned reads." />
```
mvdbeek: s/Alignment in BAM format/Select alignment ?

tool_collections/samtools/samtools_fixmate/samtools_fixmate.xml
```diff
@@ -0,0 +1,97 @@
...
+    <help>
+**What it does**
+
+Fill in mate coordinates, ISIZE and mate related flags from a name-sorted alignment.
+
+**Note**
+
+The Galaxy tool sorts SAM and coordinate sorted input automatically by query name. 
+Output is in query name sorted BAM format.  
+<!--Output is in BAM format. If the input is coordinate sorted BAM then the output is as well. -->
```
mvdbeek: That is how it is implemented now, right ?

tool_collections/samtools/samtools_fixmate/samtools_fixmate.xml
bernt-matthias: done

tool_collections/samtools/samtools_fixmate/samtools_fixmate.xml
bernt-matthias: Yes. In the beginning the tool checks if the input is query-name or input sorted. If so it takes the files as they are and if not it resorts them by name. 

In any case the output is treated as qname_sorted .. but maybe it should be qname_sorted OR input_sorted - depending on the input?

mvdbeek: Yeah, technically that would be more correct, but if fixmate worked than we know it is qname_sorted, so that's fine. We can drop the comment now because we don't sort back to coordinate sorted, right ?

bernt-matthias: exact

bernt-matthias:
But: what happens in the change_format. Is this than silently converted to bam and sorted? 

Maybe we shouldn't do this, but leave it qname_sorted?

mvdbeek:
Oh yeah, no, this is just an assignment, and it will be wrong. So for the code as it is right now the output should always be qname sorted. Otherwise we have to implement the correct sorting int he command section.

bernt-matthias:
fixed

tool_collections/samtools/samtools_markdup/samtools_markdup.xml
```diff
@@ -0,0 +1,83 @@
+<tool id="samtools_markdup" name="Samtools mark" version="@TOOL_VERSION@">
+    <description>mark duplicate alignments</description>
```
mvdbeek: The `Samtools mark mark duplicate alignments` will look a bit funny in the tool panel, how about `name="Samtools markdup"` and `marks duplicate alignments` as description ?

tool_collections/samtools/samtools_markdup/samtools_markdup.xml
```diff
@@ -0,0 +1,83 @@
...
+        <param name="bamfile" type="data" format="sam,bam,cram" optional="false" label="Alignment" />
+        <param name="remove" type="boolean" argument="-r" truevalue="-r" falsevalue="" label="Remove duplicate reads" />
+        <param name="maxlen" type="integer" optional="true" argument="-l" min="0" label="Expected maximum read length of INT bases. (default 300)"/>
+	<param name="stats" type="boolean" argument="-s" truevalue="-s" falsevalue="" label="Print basic statistics" />
+	<param name="supp" type="boolean" argument="-S" truevalue="-S" falsevalue="" label="Mark supplementary reads of duplicates as duplicates" />
+    </inputs>
+    <outputs>
+	<!-- output bam, if input was name sorted then restore this sorting order -->
+        <data name="output" format="bam">
+            <change_format>
```
mvdbeek: We should drop change_format here too, right ?

tool_collections/samtools/samtools_markdup/samtools_markdup.xml
bernt-matthias: yep

tool_collections/samtools/samtools_markdup/samtools_markdup.xml
bernt-matthias: yep.

bernt-matthias:
Hi @mvdbeek .. thanks for all you comments. 

I will go through my checklist on https://github.com/galaxyproject/tools-iuc/pull/1801 and make sure that I have forgotten nothing and everything works.

bernt-matthias:
How about the deprecated tools. Is there anything that needs to be done?

mvdbeek:
We need to move the old tools into the deprecated folder (https://github.com/galaxyproject/tools-iuc/tree/master/deprecated) -- I think that should be all that is needed.

PR #1968, opened 2018-07-04
samtools fastx upgrade to v1.9

bernt-matthias:
- [x] update tool version
- [x] option -0 seems not to work (https://github.com/samtools/samtools/issues/885) should we remove it until fixed?
  - [x] actually works. docs need update, see the above issue and below PR
- [x] implement compressed fast[aq] files, by giving the output filenames an extention of .gq, .bgz, or .bgzf. 
- [x] samtools fast[aq] needs name sorted / collated input. see updates docs https://github.com/samtools/samtools/pull/887
- [ ] someone please check new output behaviour
  - [x] output that goes to stdout needs to be gziped 'manually'
- [x] someone please check `idxout_cond`. I have no idea how the options `-i1` `-i2` are supposed to be used. apparently they need also the index format specified. There is one successful test, maybe this is sufficient for the moment.
  - marked beta


peterjc:
Supporting gzipped FASTA / FASTQ output would be welcome given other work to more widely support these compressed datatypes in Galaxy.

bernt-matthias:
> Supporting gzipped FASTA / FASTQ output would be welcome given other work to more widely support these compressed datatypes in Galaxy.

I will implement this. With the 4 potential output files (-1,-2,-0 and stdout) in 4 formats (fasta[a|q|a.gz|q.gz]) it seems that I need quite a bit outputs with `<fillter>` and `<change_format>`. Any suggestion on how to implement this most efficiently are welcome. Just tried the format auto detection.. but `planemo t` got stuck (galaxy just hanging with 100% CPU usage)

peterjc:
Maybe offer ``fasta`` vs ``sangerfastq`` as one parameter, and compression as another (either a boolean for gzipped, or if you want a drop-down with a range of compressions)?

bernt-matthias:
ok done

- does somebody know if its possible to to the label of an output per <action>? 
- somehow the optional="false" does not work for the index format parameter

tool_collections/samtools/samtools_fastx/samtools_fastx.xml
```diff
@@ -1,24 +1,43 @@
...
-                --reference '$reference'
-            #end if
-            $copy_flags $omit_read_number
-            -0 '$nonspecific'
-            -1 '$forward'
-            -2 '$reverse'
+    <command><![CDATA[
+	## name sort input file if neccessary
+        @ADDTHREADS@ 
+        #if not $input.is_of_type('qname_sorted.bam'):
```
mvdbeek: can you make this `[qname_sorted.bam', 'qname_input_sorted.bam']` ?

mvdbeek:
> somehow the optional="false" does not work for the index format parameter

There is no such thing as optional="false", parameters are required by default. Do you perhaps mean optional="true" that doesn't work ? This should be fixed by https://github.com/galaxyproject/galaxy/pull/6222

mvdbeek:
> someone please check idxout_cond. I have no idea how the options -i1 -i2 are supposed to be used. apparently they need also the index format specified. There is one successful test, maybe this is sufficient for the moment.

I am not familiar with this either. Maybe label it beta in the tool form and ask for feedback?

mvdbeek:
> does somebody know if it's possible to to the label of an output per <action>?

You mean to label using cheetah variables ? That should be possible in the top level `label`

tool_collections/samtools/samtools_fastx/samtools_fastx.xml
bernt-matthias: done

bernt-matthias:
> optional="falseMaybe label it beta in the tool form and ask for feedback?"

removed

> Maybe label it beta in the tool form and ask for feedback?

Done

tool_collections/samtools/samtools_fastx/samtools_fastx.xml
```diff
@@ -29,70 +48,459 @@
...
+                #if str($idxout_cond.quality_tag) != '':
+                    --quality-tag $idxout_cond.quality_tag
+                #end if
+            #end if
+            input 
+            #if str($outputs)=='None' or 'other' in str($outputs):
+                ## since currently stdout in not zipped (samtools doesn't know about the extension)
+                > output.$command
+                #if str($output_fmt_cond.output_fmt_select).endswith('.gz'):
+                    && gzip output.$command
```
bwlang: any way to do this on a pipe? maybe also consider pigz so we can use multiple threads?

bwlang:
This looks pretty good to me... 
I think the -i1 and -i2 bits are valuable.
I'm not 100% sure about how to get the BC tags in to a bam iun suable format, but I think bwa-mem's -C argument will do it.

bernt-matthias:
> I think the -i1 and -i2 bits are valuable.

good that you like it.

> I'm not 100% sure about how to get the BC tags in to a bam iun suable format, but I think bwa-mem's -C argument will do it.

Any suggestion is welcome. I think I won't have time for experiments.

tool_collections/samtools/samtools_fastx/samtools_fastx.xml
```diff
@@ -1,24 +1,43 @@
-<?xml version="1.0"?>
-<tool id="samtools_fastx" name="Samtools extract" version="@TOOL_VERSION@">
-    <description>FASTA or FASTQ from a SAM file</description>
+<tool id="samtools_fastx" name="Samtools fastx" version="@TOOL_VERSION@">
+    <description>extract FASTA or FASTQ from a SAM file</description>
```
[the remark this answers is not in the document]
bernt-matthias: done

tool_collections/samtools/samtools_fastx/samtools_fastx.xml
```diff
@@ -1,24 +1,43 @@
...
-            -0 '$nonspecific'
-            -1 '$forward'
-            -2 '$reverse'
+    <command><![CDATA[
+	## name sort input file if neccessary
+        @ADDTHREADS@ 
+        #if not $input.is_of_type('qname_sorted.bam', 'qname_input_sorted.bam'):
+            samtools sort -@ \$addthreads -n $input > input &&
+        #else:
+            ln -s $input input &&
```
mvdbeek: gotta quote all paths here

tool_collections/samtools/samtools_fastx/samtools_fastx.xml
```diff
@@ -29,70 +48,459 @@
...
-            <change_format>
-                <when input="output_format" value="fastq" format="fastq" />
-            </change_format>
+        <data name="nonspecific" format="fasta" from_work_dir="nonspecific" label="${on_string} converted to ${output_fmt_cond.output_fmt_select} (nonspecific)">
+            <filter>'r0' in str(outputs)</filter>
+    <actions>
+                <action type="format">
+                    <option type="from_param" name="output_fmt_cond" param_attribute="output_fmt_select" />
+                </action>
+</actions>
```
mvdbeek: The indentation is a bit off here

tool_collections/samtools/samtools_fastx/samtools_fastx.xml
bernt-matthias: done

tool_collections/samtools/samtools_fastx/samtools_fastx.xml
bernt-matthias: fixed

mvdbeek:
Any chance you forgot to add/push your last changes ? I don't see them

tool_collections/samtools/samtools_fastx/samtools_fastx.xml
bernt-matthias: done

tool_collections/samtools/samtools_fastx/samtools_fastx.xml
bernt-matthias: done

bernt-matthias:
guess the last open point is @bwlang s comment on using pipes for gzip. I see alternatives: 

- we could use samtool functionality for compression. According to the manual "The files will be automatically compressed if the file names have a .gz or .bgzf extension."
- Alternatively also pigz could be used with parallel processing turned on...
- or we leave it as it is :)

What do you think about this?

mvdbeek:
Oh, if the samtools builtin functionality works for us we should use it.

Issue #1952, opened 2018-06-28
bedtools intersect is missing the sorted option


bernt-matthias:
Would be important for large input. Because bedtools promises to be more memory efficient.

bernt-matthias:
Just one more question. Apparently Null Pointer Exceptions seem to indicate a out of memory condition. Would it be an idea to check for the fatal_oom level, I guess this would be to dirty.

bgruening:
Think so as well, Null Pointer can mean everything :(

I found this here:
```
If you are trying to intersect very large files and are having trouble with excessive memory usage, please presort your data by chromosome and then by start position (e.g., sort -k1,1 -k2,2n in.bed > in.sorted.bed for BED files) and then use the -sorted option. This invokes a memory-efficient algorithm designed for large files. This algorithm has been substantially improved in recent (>=2.18.0) releases.
```

Which indicated that a user should sort in advance, is that not true anymore?

bernt-matthias:
I think its still true. So we could either silently sort with gnu sort and activate `-sorted` or let the user decide if he wants this. We could also make a short run time comparison with `sort ... && bedtools intersect -sorted` vs. just `bedtools intersect`. I guess the former will be the same time or even faster.

PR #1969, opened 2018-07-04
bedtools intersect: add -sorted, -bed, -e options

bernt-matthias:
- `-sorted`: allow to use the faster algorithm for sorted input
- `-bed`: output bed for bam input
- `-e` A OR B for reciprocal overlap fractions

### DIFF
--- a/tools/bedtools/intersectBed.xml
+++ b/tools/bedtools/intersectBed.xml
@@ -33,16 +33,17 @@
                   -f '${fraction_cond.fractionA}'
                 #end if
                 #if str($fraction_cond.fractionB) != "None" and str($fraction_cond.fractionB):
-                  -f '${fraction_cond.fractionB}'
+                  -F '${fraction_cond.fractionB}'
                 #end if
             #else if $fraction_cond.fraction_select == "joint":
-                $fraction_cond.reciprocal fraction_cond.fraction
+                $fraction_cond.reciprocal -f $fraction_cond.fraction
             #end if
             $invert
             $once
             $header
             $modes
             $sorted
+            $bed
             $count
             > '${output}'
 ]]>
--- a/tools/bedtools/intersectBed.xml
+++ b/tools/bedtools/intersectBed.xml
@@ -1,4 +1,4 @@
-<tool id="bedtools_intersectbed" name="Intersect intervals" version="@WRAPPER_VERSION@.2">
+<tool id="bedtools_intersectbed" name="Intersect intervals" version="@WRAPPER_VERSION@+galaxy3">
     <description>find overlapping intervals in various ways</description>
     <macros>
         <import>macros.xml</import>
@@ -28,14 +28,21 @@
             #end if
             $split
             $strand
-            #if str($fraction) != "None" and str($fraction):
-              -f '${fraction}'
+            #if $fraction_cond.fraction_select == "separate":
+                #if str($fraction_cond.fractionA) != "None" and str($fraction_cond.fractionA):
+                  -f '${fraction_cond.fractionA}'
+                #end if
+                #if str($fraction_cond.fractionB) != "None" and str($fraction_cond.fractionB):
+                  -f '${fraction_cond.fractionB}'
+                #end if
+            #else if $fraction_cond.fraction_select == "joint":
+                $fraction_cond.reciprocal fraction_cond.fraction
             #end if
-            $reciprocal
             $invert
             $once
             $header
             $modes
+            $sorted
             $count
             > '${output}'
 ]]>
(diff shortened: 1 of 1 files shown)


tools/bedtools/intersectBed.xml
```diff
@@ -66,12 +74,29 @@
...
 
         <expand macro="split" />
-        <!-- -f -->
-        <param name="fraction" type="text"
-            label="Minimum overlap required as a fraction of the BAM alignment"
-            help="Alignments are only retained if the overlap with the an interval in the BED file comprises at least this fraction of the BAM alignment's length.  For example, to require that the overlap affects 50% of the BAM alignment, use 0.50. (-f)"/>
-        <!-- -r -->
-        <expand macro="reciprocal" />
+        <conditional name="fraction_cond">
+            <param name='fraction_select' type='select' label='Required overlap'>
```
mvdbeek: This breaks simple tool upgrades. The default parameters (overlap of 1e-9) are effectively the same as the more formal `none` option introduced here. If accuracy is a concern we could bump this to 1e-12 ? Is there anything else I missed that this would fix ?

bernt-matthias: What exactly is breaking the upgrades? I can revert to setting the value 1e-09 explicitly .. will use the macro again (and add one for `-F`). 

I'm currently working on tests. 

Maybe a question. There are the `-names` and -filenames parameter which add a column to the output which allows to see more easily from which file the overlapping entries come.  Is there a way to get some useful metadata info from the data sets that could be used here?

mvdbeek: > What exactly is breaking the upgrades?

The change of parameters names, from fraction to fraction_cond.fraction. Which is fine if its needed, but I feel this will still behave the same ultimately?



> There are the -names and -filenames parameter which add a column to the output which allows to see more easily from which file the overlapping entries come. Is there a way to get some useful metadata info from the data sets that could be used here

I guess the proper dataset element identifier might be helpful ?

mvdbeek: > might be helpful because it gives a bit more structure to the tool form

I don't agree, while the initial label (`Minimum overlap required as a fraction of the BAM alignment`) for instance doesn't make it easy for the user, setting `Required overlap` to `None` also makes no sense, since if we take this literally, we're not intersecting anymore. Setting `Required overlap` to `separate` and `joint` is also weird, since that just describes *how* you want to specify the required overlap.


> Could you point me to some docs. Could not find it.

I mean the regular `.element_identifier` attribute (https://planemo.readthedocs.io/en/latest/writing_advanced.html?highlight=element%20identifier#processing-identifiers)

bernt-matthias: yep -- OK. Meanwhile I also learned a bit on bedtools intersect. I already have switched back to the original implementation, but then I learned that -r and -F can not be used together. So my current plan is to have a default/advanced conditional for the overlaps. The advanced will show -f and -r (and if -r is unselected then -F and -e are shown). Unfortunately this would imply that the boolean reciprocal needs to be transformed to a select. 

Thanks for the link.

bernt-matthias:
@mvdbeek Just commited an improved version. If you prefer I can remove the outer conditional of the overlap specification. 

@bwlang just implemented the technique that you introduced in the coverageBed tool. But wouldn't it be better to use `-sorted -g ...` also for sorted bed input and not only for bam input? Also it seems that some of the macros could be also used in the coverage tool...

tools/bedtools/intersectBed.xml
```diff
@@ -54,24 +72,38 @@
...
+                <option value='default' selected='true'>Default: 1bp</option>
```
mvdbeek: Thanks, I like that option!

tools/bedtools/intersectBed.xml
```diff
@@ -3,49 +3,59 @@
...
-            #else:
-                #set inputBs = "' '".join([str($file) for $file in $reduce_or_iterate.inputB])
-                -b '$inputBs'
+        #if str($reduce_or_iterate.reduce_or_iterate_selector) == 'iterate':
+            -b '$reduce_or_iterate.inputB'
+        #else:
+            #set inputBs = "' '".join([str($file) for $file in $reduce_or_iterate.inputB])
+            -b '$inputBs'
+            #if $reduce_or_iterate.names:
+                #set namesB = "' '".join([str($file.element_identifier) for $file in $reduce_or_iterate.inputB])
```
mvdbeek: we have the element_identifier, we usually use ` re.sub('[^\s\w\-]', '_', str($file.element_identifier)` for this, which you can define as a token, similar to what we do in https://github.com/galaxyproject/tools-iuc/blob/master/tools/multiqc/multiqc.xml#L5

bwlang:
> But wouldn't it be better to use -sorted -g ... also for sorted bed input and not only for bam input? Also it seems that some of the macros could be also used in the coverage tool...
@bernt-matthias  I'm not sure I understand...
how can one tell if bed input is sorted? 
also - where to get the contig names/ordering? (cut -f 1 | uniq ?)

bernt-matthias:
@bwlang For this tool I currently let the user decide using a checkbox. In https://github.com/galaxyproject/tools-iuc/issues/1952 I tried to start a discussion .. an alternative would be to sort tabular input within the tools.

bwlang:
@bernt-matthias hmm - i'm of two minds.  unix sort + bedtools might be good, but maybe it's best to fix bedtools?
maybe this is less of an issue for bed files - i rarely see massive ones.  it's the bams that are huge.
could also consider a new data type 'sorted_bed'

mvdbeek:
> i rarely see massive ones. it's the bams that are huge.

bedtools also accepts VCFs and some other formats. VCFs can definitely be huge.

> could also consider a new data type 'sorted_bed'

or at least a metadata element, to avoid sorting if it's already sorted

bernt-matthias:
> or at least a metadata element, to avoid sorting if it's already sorted

this would be better in my opinion .. would also need to store the chromosome order .. the idea would be that the sorting tools set the metadata? But I guess I have not sufficient time for this at the moment. So we can decide for 

a) sort within the tool 
b) let the user decide

nsoranzo:
I'd keep asking the user as last resort, if the use case actually emerges.

bgruening:
There is a new version of bedtools (https://anaconda.org/bioconda/bedtools). Maybe we should this update as well.

bernt-matthias:
there you are, I guess adding "bedtools" to the tool names is also a good idea

peterjc:
I think we do want ``bedtools ....`` for the human readable naming, as without it many of these names are very generic, and likely to cause confusion due to the ambiguity.

bernt-matthias:
For instance "Convert from BAM to FastQ" could as well be samtools

bernt-matthias:
but of course also the Galaxy admin could do this by adding the bedtools to an extra category

tools/bedtools/intersectBed.xml
bernt-matthias: @mvdbeek do you think the current implementation is fine? We still have a conditional, but now for joint/separate specification of the overlap fraction -- which I think is necessary. 

But I might need to add a check for None in the `joint` branch / make the fraction parameters non-optional?

bernt-matthias:
Can this be merged?

PR #1970, opened 2018-07-04
Add featurecounts fix

mtekman:
Subread version bump to (hopefully) fix a segmentation fault, and added a test for -R flag.


mtekman:
@bgruening nice save, thanks

PR #1971, opened 2018-07-06
samtools depth new tool in v1.8

bernt-matthias:
- [x] `samtools_depth`
  - [x] I did not expect values > 4 in the output, but there are ... 
- [x] choose a version: 1.8 (using the samtools version) or 2.0.2 (just increasing the version of the other galaxy tools)


pvanheus:
Looks good, I'm +1 on this.

peterjc:
See also https://github.com/peterjc/pico_galaxy/blob/master/tools/samtools_depth/samtools_depth.xml which made it to the Test Tool Shed, but not the main Tool Shed (it got held up with samtools's max depth settings - long since fixed in the released samtools binaries).

Other than the help text, yours looks much more comprehensive.

tool_collections/samtools/samtools_depth/samtools_depth.xml
```diff
@@ -0,0 +1,111 @@
...
+
+            <param name="basequality" value="11"  />
+            <param name="mapquality" value="12" />
+            <output name="output" file="samtools_depth_out4.tab" />
+        </test>
+    </tests>
+    <help>
+**What it does**
+
+Calculates read depth for regions listed in a BED dataset using ``samtools bedcov`` command::
```
peterjc: Misleading help text - the default is all regions, can do BED regions, or text region.

Also would be helpful to describe the output format, e.g. copying my text from https://github.com/peterjc/pico_galaxy/blob/master/tools/samtools_depth/samtools_depth.xml

The output is a tabular file, with one line for each base of each reference
with any coverage (bases with no coverage may not appear).

The first column is the reference name, the second column is the reference
position, and then there is one column for each SAM/BAM file giving the
coverage depth at that position.

tool_collections/samtools/samtools_depth/samtools_depth.xml
```diff
@@ -0,0 +1,111 @@
...
+
+ samtools bedcov [INPUT BED] [INPUT BAM1] ... [INPUT BAMn] > [OUTPUT]
```
peterjc: This is not ``samtools bedcov`` but ``samtools depth`` (which may explain why the help text seemed unclear)

tool_collections/samtools/samtools_depth/samtools_depth.xml
```diff
@@ -0,0 +1,111 @@
...
+        #end for
+
+	> '${output}'
+    ]]></command>
+    <inputs>
+	<param name="input_bams" type="data" format="bam" multiple="true" label="BAM file(s)" />
+        <conditional name="cond_region">
+            <param name="select_region" type="select" label="Filter by regions">
+                <option value="no" selected="True">No</option>
+                <option value="text">Manualy specify a region</option>
```
peterjc: This need help text with an example of how to specify a region.

tool_collections/samtools/samtools_depth/samtools_depth.xml
bernt-matthias: Copy paste error (I started with the bedcov xml and forgot to update the help).

tool_collections/samtools/samtools_depth/samtools_depth.xml
bernt-matthias: fixed

tool_collections/samtools/samtools_depth/samtools_depth.xml
```diff
@@ -0,0 +1,116 @@
...
+
+            <param name="basequality" value="11"  />
+            <param name="mapquality" value="12" />
+            <output name="output" file="samtools_depth_out4.tab" />
+        </test>
+    </tests>
+    <help>
+**What it does**
+
+Computes the depth at each position or region using the ``samtools depth`` command::
```
peterjc: You have a trailing double colon which is RST syntax for the following being a literal code block - like a command line example?

bernt-matthias: fixed

peterjc:
I'm happy now (but have only given this a superficial check), and we have a +1 for @pvanheus - anyone ready to merge this?

bernt-matthias:
Apart from my two questions 
- can you explain the strange depth of >4 in the last test
- how about the version? I would prefer 1.8, but then we diverge from the other samtools tools which are stuck in the old version >2.x.x

peterjc:
Regarding the max-depth, what samtools produces is (or at least in samtools 1.4.1 was) a little fuzzy - see also my comments in https://github.com/peterjc/pico_galaxy/blob/master/tools/coverage_stats/coverage_stats.py - in particular variable ``depth_margin``. I concluded it could under report coverage if a region was rich in CIGAR deletions.

--

As this is effectively a new tool, we can follow https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html and use ``@TOOL_VERSION@+galaxyN`` here?

tool_collections/samtools/samtools_depth/samtools_depth.xml
```diff
@@ -0,0 +1,113 @@
+<tool id="samtools_depth" name="Samtools depth" version="1.8">
```
nsoranzo: s/1.8/@TOOL_VERSION@/

bernt-matthias:
I'm satisfied now with the answer regarding the depth in https://github.com/samtools/samtools/issues/889 and adapted the help accordingly. Ready to merge from my side.

bgruening:
Thanks @bernt-matthias and all other for the review and discussions!

peterjc:
+1 on the new max-depth help text, good job 👍

PR #1972, opened 2018-07-06
samtools flagstats, idxstats, and stats in v1.9

bernt-matthias:
- [x] `samtools stats`
  - [x] could generate stats per prefix or tag (-P -S) not implemented, yet. Should we?
   - keep it for later
  - [x] some of the dataset discovery does not work
  - [x] should the tool output empty tables (currently it does not for split outputs)
    -  jep: for consistence
  - [x] plot bamstats: remove `<a>` and `</a>` (`<a class="thumbnail" href="insert-size.png">`) and css from html output
   - seems to be no problem (tested with `planemo s`)


mvdbeek:
> could generate stats per prefix or tag (-P -S) not implemented, yet. Should we?

Sounds cool, but one step at a time ?

> should the tool output empty tables (currently it does not for split outputs)

When would that happen ?

tool_collections/samtools/samtools_idxstats/samtools_idxstats.xml
```diff
@@ -7,18 +7,23 @@
     <expand macro="stdio"/>
     <expand macro="version_command"/>
     <command><![CDATA[
-        ln -f -s '${input.metadata.bam_index}' '${input}.bai' &&
-
-        samtools idxstats '${input}' > '${output}'
+        @PREPIDX@
```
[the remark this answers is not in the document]
bernt-matthias: changed it

need to keep in mind that this affects also some other samtools tools which are in the other PRs.

bernt-matthias:
>>    should the tool output empty tables (currently it does not for split outputs)
> When would that happen ?

In the unlikely case that there are no values for some category, e.g. if there are no indels. If we decide to generate also empty tables then we can remove the `if grep...` statement.

mvdbeek:
Thanks, in that case I think I prefer empty tables, for the sake of being consistent, but that's up to you.

bernt-matthias:
OK. All done -- incl update to v1.9. 

One question out of interest: I thought that I need to replace the image paths in the html file from image.png to html_file.files_path/image.png, but it seems (tested with `planemo s`) that it works out of the box. Is there some magic happening in the background?

bernt-matthias:
should be fixed by the last commit

tool_collections/samtools/samtools_stats/samtools_stats.xml
```diff
@@ -1,184 +1,132 @@
...
             #end if
+            infile
 
-            #if 'ic' in $outputs_to_split:
-                && echo "# Indels per cycle. The columns are: cycle, number of insertions (fwd), .. (rev) , number of deletions (fwd), .. (rev)" > 'split/Indels per cycle.tab' &&
-                if grep -q ^IC '${output}'; then
-                    grep ^IC '${output}' | cut -f 2- >> 'split/Indels per cycle.tab';
-                fi
+            #if $cond_region.select_region == 'text':
+                $cond_region.regions
```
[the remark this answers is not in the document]
bernt-matthias: I added single quotes. 

If I interpret this correctly single quotes are not in the default set of allowed characters:  https://github.com/galaxyproject/galaxy/blob/ed6f5d45a9ecbf0cf8a85341e0213b82a9132a46/lib/galaxy/tools/parameters/sanitize.py#L47

bernt-matthias:
problem is the quoting that the last commit introduced. if quoted its interpreted as a single region which changes the results

according to the samtools online man page the specification is

```
[region...]
```

where each region in 

```
chr:from-to
```

Of course I could split the parameter and quote the elements, but this would still lead to wrong results if the user adds wrong spaces.

In my opinion the best solution is documentation (question is if its sufficient at the moment). I guess that (simple) validators do not help here, since filtering/checking for spaces does not help. Otherwise we would need to parse the content of the parameter....

For the moment I just revert the last commit.

mvdbeek:
I don't think this diagnosis is correct:
samtools view ./spen_dnaseq_analysis/spen.bam 2L:156969-176969|wc -l
    7780
samtools view ./spen_dnaseq_analysis/spen.bam '2L:156969-176969'|wc -l
    7780

mvdbeek:
And the diff for instance has this line:
```
+# This file was produced by samtools stats (1.9+htslib-1.9) and can be plotted using plot-bamstats
+# This file contains statistics for all reads.
+# The command line was:  stats -g 4 infile ref1:10-24 ref1:30-46 ref1:39-56
```
which is absent form the test data file.

mvdbeek:
Ah these are multiple regions, I see, sorry for the noise. So this should a repeat for each region ideally

bernt-matthias:
Repeat is perfect. Just changed it. I think the text param + regex validator is more powerful since it can check ecactly for `RNAME[:STARTPOS[-ENDPOS]]` as specified in the man page. We couldn't do this with three params for RNAME, STARTPOS and ENDPOS. 

There are some more tools in the suite that would profit form this, maybe we include this in this PR, ie add a macro and token and include it in the other xmls? 

- samtools_bedcov.xml
- samtools_depth.xml
- samtools_mpileup.xml
- samtools_view.xml

as well as the view based tools (which I would not like to touch): 

- samtools_cram_to_bam.xml
- samtools_slice_bam.xml

tool_collections/samtools/samtools_stats/samtools_stats.xml
```diff
@@ -189,113 +132,249 @@
...
             </when>
-            <when value="no" />
         </conditional>
-
     </inputs>
 
     <outputs>
         <data name="output" format="tabular" label="${tool.name} on ${on_string}">
-            <discover_datasets pattern="(?P&lt;designation&gt;.+)\.tab" ext="tabular" visible="true" directory="split" />
+            <discover_datasets directory="split" pattern="(?P&lt;designation&gt;.+)\.tab" format="tabular" visible="true" assign_primary_output="true"/>
```
mvdbeek: How do you feel about making this a collection ? I think this would be easier to deal with in downstream tools and should be helpful if you want to include this output in workflows.

mvdbeek:
To me this is fine, I'm happy to merge as is and we think about macros in the next round ?

tool_collections/samtools/samtools_stats/samtools_stats.xml
bernt-matthias: added a collection if the output is splitted.

bernt-matthias:
Can be merged when the test is successful.

tool_collections/samtools/samtools_stats/samtools_stats.xml
mvdbeek: Thanks, that's awesome!

mvdbeek:
Awesome, really appreciate the work!

Issue #1921, opened 2018-06-05
Samtools Mpileup 2.1.3 metadata output problem with BCF format + warnings about options


jennaj:
MTS: https://toolshed.g2.bx.psu.edu/view/devteam/samtools_mpileup/583abf29fc8e

Test history: https://usegalaxy.org/u/jen/h/test-history-mpileup-bcftools-call

Data 10 presents with the metadata problem if BCF is chosen as the output along with "uncompressed" set (uncompressed does not change to compressed when BCF is chosen instead of VCF). Is there a reason why this option is on the tool form? Couldn't it be hidden from the user? Choosing VCF implies uncompressed already, and BCF compressed. The user shouldn't need to specify this twice -- it leaves open an easy way to create a problematic output. If BCF/compressed are both set, the job is successful (Data 13).

<img width="426" alt="screen shot 2018-06-05 at 9 02 09 am" src="https://user-images.githubusercontent.com/8999735/40988484-0a9098f2-68a0-11e8-88e3-4d3a706f7ae5.png">
<img width="250" alt="screen shot 2018-06-05 at 9 02 26 am" src="https://user-images.githubusercontent.com/8999735/40988485-0aab6556-68a0-11e8-90b4-d535241ee8c6.png">
<img width="244" alt="screen shot 2018-06-05 at 9 02 45 am" src="https://user-images.githubusercontent.com/8999735/40988486-0ac76d8c-68a0-11e8-92b7-cf572f82167c.png">

Warnings in all outputs. Looks like the tool needs an update anyway. Maybe can add in the above at the next rev?

```
[warning] tag DPR functional, but deprecated. Please switch to `AD` in future.
[warning] tag DV functional, but deprecated. Please switch to `AD` in future.
[warning] tag DP4 functional, but deprecated. Please switch to `ADF` and `ADR` in future.
[warning] tag INFO/DPR functional, but deprecated. Please switch to `INFO/AD` in future.
[mpileup] 1 samples in 1 input files
<mpileup> Set max per-file depth to 8000
```

Found during testing for https://github.com/galaxyproject/usegalaxy-playbook/issues/123
Update for `.org` tracked here: https://github.com/galaxyproject/usegalaxy-playbook/issues/133

bernt-matthias:
[...]

PR #1973, opened 2018-07-06
samtools mpileup in v1.9

bernt-matthias:
- [x] mpileup: 
  - [x]  upgrade to 1.8
  - [x] https://github.com/galaxyproject/tools-iuc/issues/1921
  - [x] compressed bcf output does not work yet: 
    - the latter two are obsolete in 1.9


tool_collections/samtools/samtools_mpileup/samtools_mpileup.xml
```diff
@@ -262,38 +262,57 @@ ${pasted_data}
...
+                    <param name="coefficient_for_modeling_homopolymer_errors" value="100" />
+                    <param name="skip_indel_calling_above_sample_depth" value="250" />
+                    <param name="gap_open_sequencing_error_probability" value="40" />
+                    <param name="platform_list_repeat" value="0" />
+                </conditional>
+            </conditional>
+            <conditional name="advanced_options">
+                <param name="advanced_options_selector" value="basic" />
+            </conditional>
+            <output name="output_file" file="samtools_mpileup_out_2.vcf" ftype="vcf" lines_diff="8" />
```
bgruening: This should not be output_file, but rather `output_file_vcf`

bgruening:
I think your tests were just wrong, can this be?
Matthias, do you think we can put all the tools in one directory and use planemo demultiplexing to spit them put into different repos without the symlinking and keeping the currently repo working?

bgruening:
> the output name was definitely wrong. odd that the planemo tests still worked. is this an issue?

We could be a lot smarter in linting and so on, but I guess no one of us has time to work on it.

> I'm wondering if the tool accepts all bam data types as input? Explicitely allowed is bam, but I guess also qname_sorted bam etc is accepted by the tool

Not sure. I would leave BAM, Galaxy will convert if needed and I assume an idex is required.

> for bam the index should be always present .. ?

Yes.

nsoranzo:
> > the output name was definitely wrong. odd that the planemo tests still worked. is this an issue?

> We could be a lot smarter in linting and so on, but I guess no one of us has time to work on it.

Actually this is already checked by `planemo lint`, but it is only a warning, see the first Travis build for this PR: https://travis-ci.org/galaxyproject/tools-iuc/jobs/400921611

We should probably change it to an error.

bernt-matthias:
made it an issue https://github.com/galaxyproject/planemo/issues/858

mblue9:
@bernt-matthias did you see there's now a samtools v1.9 https://anaconda.org/bioconda/samtools

P.S Welcome to IUC!

bernt-matthias:
@mblue9 thanks for the info. seems that they have deprecated samtools' mpileup in favour of bcftools. will see if I find the time.

bernt-matthias:
Guess I'm done here: 
- updated to 1.9
- removed vcf and bcf output which is now deprected in `samtools mpileup` (they refer to bcftools)

bernt-matthias:
One comment/question: since one of my samtools PR is already merged: how is this handled with the shared macros? Is each samtools tool deposited separately to the TS? Then I guess its OK, because each gets its separate copy of the macros.xml, or?

mvdbeek:
Yeah, that should be ok, I don't think planemo detects this as a change, but we should not forget to adjust all to the latest macro version once we're done here.

mvdbeek:
Thanks for all the much needed work, this is much appreciated!

PR #1974, opened 2018-07-08
update fastp to version 0.18.0

bgruening:


mblue9:
Just to add here, looks like MultiQC should have a module for fastp in the next release: https://github.com/ewels/MultiQC/pull/791

(also this tool can't handle paired-end collections so maybe something similar to what was done for cutadapt could be added at some point)

Issue #1901, opened 2018-05-29
Error while running DEXseq


bellenger-l:
Hello,

I tried to run DEXseq count reads with a DEXseq reference and a bam but I have the following error : 
```
Fatal error: Exit code 1 ()
Traceback (most recent call last):
  File "/home/galaxy/tool_dependency/_conda/envs/mulled-v1-683dc60555985982bcd77839627baa1ef503a6a15fb8cf27f5b492bbedd69ac4/bin/dexseq_count.py", line 98, in 
    features[f.iv] += f
  File "python3/src/HTSeq/_HTSeq.pyx", line 451, in HTSeq._HTSeq.ChromVector.__iadd__
  File "python3/src/HTSeq/_HTSeq.pyx", line 466, in HTSeq._HTSeq.ChromVector.apply
  File "python3/src/HTSeq/_HTSeq.pyx", line 449, in HTSeq._HTSeq.ChromVector.__iadd__.addval
TypeError: unhashable type: 'GenomicFeature'
```

It seems that there is no test for DEXseq counts, (only for DEXseq prepare reference), so can you help me figure it out ? (After some research, a problem with the version of python is mentioned : https://support.bioconductor.org/p/106685/)

Best,
Lea

mblue9:
@bellenger-l and @michelraymond this should be fixed now (xref: https://github.com/galaxyproject/tools-iuc/pull/1975) but let us know if not.

PR #1975, opened 2018-07-10
DEXSeq: Add python 2.7 requirement to dexseq_count

mblue9:
This PR adds python 2.7 as a requirement to DEXSeq-Count as otherwise the Count reads mode can fail with an error like:
`
Fatal error: Exit code 1 ()
  File "/usr/local/tools/_conda/envs/mulled-v1-683dc60555985982bcd77839627baa1ef503a6a15fb8cf27f5b492bbedd69ac4/bin/dexseq_count.py", line 98, in <module>
    features[f.iv] += f
  File "python3/src/HTSeq/_HTSeq.pyx", line 451, in HTSeq._HTSeq.ChromVector.__iadd__
  File "python3/src/HTSeq/_HTSeq.pyx", line 466, in HTSeq._HTSeq.ChromVector.apply
  File "python3/src/HTSeq/_HTSeq.pyx", line 449, in HTSeq._HTSeq.ChromVector.__iadd__.addval
TypeError: unhashable type: 'GenomicFeature'
`

### DIFF
--- a/tools/dexseq/dexseq_count.xml
+++ b/tools/dexseq/dexseq_count.xml
@@ -1,10 +1,11 @@
-<tool id="dexseq_count" name="DEXSeq-Count" version="@VERSION@.0">
+<tool id="dexseq_count" name="DEXSeq-Count" version="@VERSION@.1">
     <description>Prepare and count exon abundancies from RNA-seq data</description>
     <macros>
         <import>macros.xml</import>
     </macros>
     <expand macro="requirements">
         <requirement type="package" version="0.9.1">htseq</requirement>
+        <requirement type="package" version="2.7">python</requirement>
     </expand>
     <stdio>
         <!-- Anything other than zero is an error -->


nsoranzo:
xref. https://support.bioconductor.org/p/106685/

I think this workaround should be added at the Bioconda level at https://github.com/bioconda/bioconda-recipes/blob/master/recipes/bioconductor-dexseq/meta.yaml .

@mblue9 Maybe you can also try to report this to htseq upstream at https://github.com/simon-anders/htseq ?

mblue9:
Thanks @nsoranzo, good suggestions! 
I see @bgruening has made a PR to make the change for Bioconda here: https://github.com/bioconda/bioconda-recipes/pull/9854 and I've reported the issue upstream to HTSeq here: https://github.com/simon-anders/htseq/issues/59.

mblue9:
@nsoranzo the better fixes are looking like they might take a while so wondering what so you think about letting this through as a short-term fix? To fix it for people currently trying to install dexseq from the toolshed eg there's a report of this error here: https://github.com/galaxyproject/tools-iuc/issues/1901

nsoranzo:
I'd try to wait a bit more. If someone really needs the fix, they can apply this fix post-installation and reinstall the conda dependencies.

bgruening:
I fixed it on the bioconda side. However, this is also not perfect, as we also need changes more downstream - but it should now work for Galaxy.

@mblue9 I changes this already last week on usegalaxy.eu and it should work there already.

nsoranzo:
@mblue9 Do you have a test case that could be added to the tool to verify we are not regressing?

mblue9:
I don't, but I'll see if I can make one when I get  chance.

mblue9:
So I'm still getting the Python3 error with dexseq on our instance. I tried deleting the env and reinstalling dexseq and still get:
```
Fatal error: Exit code 1 ()
  File "/galaxy/production/database/dependencies/_conda/envs/mulled-v1-683dc60555985982bcd77839627baa1ef503a6a15fb8cf27f5b492bbedd69ac4/bin/dexseq_count.py", line 98, in 
    features[f.iv] += f
  File "python3/src/HTSeq/_HTSeq.pyx", line 451, in HTSeq._HTSeq.ChromVector.__iadd__
  File "python3/src/HTSeq/_HTSeq.pyx", line 466, in HTSeq._HTSeq.ChromVector.apply
  File "python3/src/HTSeq/_HTSeq.pyx", line 449, in HTSeq._HTSeq.ChromVector.__iadd__.addval
TypeError: unhashable type: 'GenomicFeature'
```

If I try to run the dexseq_count test I added on our instance it fails with the same Python3 error and if I activate the dexseq env the default python is python 3, whereas locally on my Mac the dexseq count test passes fine and the dexseq env is using python 2.

Can anyone confirm if dexseq_count works for them in their instance? Or does anyone know how to fix this?

bgruening:
@mblue9 can you try it again on usegalaxy.eu? Should be fixed.

mblue9:
Just tried it and yep it's fixed at .eu now 👍 Now how can I fix it in our local instance? And how to fix it for people installing from the toolshed?

PR #1979, opened 2018-07-11
hisat2 serialize samtools

bernt-matthias:
Currently hisat2 and samtools sort run in parallel each with the full number of threads (ie in theory up to 2x GALAXY_SLOTS). 

The suggested solution tries to reduce the over-usage, ie. it would use at most one CPU more than GALAXY_SLOTS. 

see also https://github.com/samtools/samtools/issues/891

I you think this is not a good idea we could serialize hisat2 and samtools sort using a temporary file. 

(plus samtools is updated to 1.8)


bgruening:
Thanks @bernt-matthias!

PR #1980, opened 2018-07-12
Enable meme_chip to run Centrimo

gregvonkuster:
Since https://github.com/galaxyproject/galaxy/pull/6374 was merged, tools will once again render multiple dependent dynamic select lists, where selecting an option in the first will render the options in the second.

With this fix, we can now include the option to run Centrimo here in the meme_chip tool, which requires installation of the ```meme motif databases downloader``` data manager tool.  This https://github.com/galaxyproject/tools-iuc/pull/1772 will have to be merged and made available in the MTS to enable this.

### DIFF
--- a/tools/meme_chip/meme_chip.xml
+++ b/tools/meme_chip/meme_chip.xml
@@ -16,29 +16,27 @@ meme-chip '$input'
 $sequence_alphabet
 -o '$output.files_path'
 #if str($options_type_cond.options_type)=='advanced':
-    ## FIXME: CentriMo cannot be run,  See the comments in the input section.
-    ## #set run_centrimo = $options_type_cond.run_centrimo_cond.run_centrimo
-    ## #if str($run_centrimo) == "yes":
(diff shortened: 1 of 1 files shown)


bgruening:
@gregvonkuster can you increase the tool version number and maybe include a profile to indicate from which Galaxy version this tool is supposed to work?

gregvonkuster:
@bgruening I've increased the tool version string.  Can you let me know what profile string you'd like to use?  It's difficult to know when support for this tool feature was broken in the Galaxy framework since I haven't messed around with the only other tool that uses it for a couple of years, so past profiles are probably not useful, and it should probably be the next release version.  Will that be 18.09?

bgruening:
Good question @gregvonkuster. When was it fixed? I think 18.09 is a save bet. But maybe the fix is already in 18.05? Can you check? Thanks.

gregvonkuster:
Here is the tag, so I set it to 18.09.  Thanks @bgruening 

```Fix for rendering multiple dependent dynamic select lists  kind/bug status/review
#6374 by gregvonkuster was merged 28 days ago • Approved    18.09```

PR #1982, opened 2018-07-14
adds support and tests for interleaved input files

bwlang:


bwlang:
any concerns about the new dependency on seqtk or the somewhat hacky way i'm supporting bzip2?

bgruening:
@galaxyproject/iuc this is an invasive change to Hisat, can you please comment and review :)

Thanks!

tools/hisat2/hisat2.xml
```diff
@@ -155,10 +176,14 @@ hisat2
     #end if
 
 #else:
-
-    -1 '${read1}'
-    -2 '${read2}'
-
+    ##quotes are embedded in r1 and r2 variables, needed to allow use of <()
```
mblue9: I don't know enough to know if this is ok or if there's a better way to do this, otherwise this PR is fine by me!

bwlang:
looks good to me - thanks for the fixes.

abretaud:
Thanks @bwlang!

tools/hisat2/hisat2.xml
jmchilton: This isn't super clean but looks entirely reasonable to me and I can't imagine a trick to clean it up any further. In this case ``${read1}`` doesn't have any unwanted characters in it as far as I can tell based on the above logic - so this seems secure enough to me.

mblue9: Cool, good to know, thanks @jmchilton !

PR #1992, opened 2018-07-17
update kraken version

rhpvorderman:
Hi I got some error when running the kraken database builder, so I figured the best way forward was to update to the next stable release of kraken. Currently it depends on some weird in between versions version.


data_managers/data_manager_build_kraken_database/data_manager/kraken_database_builder.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0"?>
-<tool id="kraken_database_builder" name="Kraken" tool_type="manage_data" version="1.0.2">
+<tool id="kraken_database_builder" name="Kraken" tool_type="manage_data" version="1.1.0">
```
nsoranzo: I'd just use `1.1`, see https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tool-versions

nsoranzo:
Thanks @rhpvorderman!

PR #1996, opened 2018-07-18
add configfile to stdout of bamtools_filter

bgruening:


mvdbeek:
I really dislike this practice, can we stop doing that? If this file is interesting it should be an output, so that we can continue operating on it. This is currently stored in the database, and is being truncated if it becomes too big (max_metadata_value_size in galaxy.yml, default 5MB).

PR #1998, opened 2018-07-19
Update umi-tools counts

mtekman:
Option to prepend columns by a custom string or filename.


tools/umi_tools/umi-tools_counts.xml
```diff
@@ -111,6 +123,29 @@ prep generates PCR duplicates with non-identical alignment positions such as CEL
             <param argument="--per-cell" name="per_cell" type="boolean" truevalue="--per-cell" falsevalue="" checked="true" label="Group reads only if they have the same cell barcode." />
             <param argument="--random-seed" name="random_seed" type="integer" min="0" value="0" label="Random Seed" />
         </section>
+
+        
+        <conditional name="cond_extra" >
+            <param name="prepender" type="select" label="Prepend a label to all column headers" help="This preserves uniqueness when merging with other files with the same headers. Note: filename must not contain a '.' character" >
+                <option value="none" selected="true" >No modifications</option>
+                <option value="string">Custom Label</option>
+                <option value="filename">Filename</option>
```
bgruening: maybe name it `dataset name`?

tools/umi_tools/umi-tools_counts.xml
```diff
@@ -10,36 +10,48 @@
...
+    --skip-tags-regex='$advanced.skip_tags_regex.value'
+#end if
+#if '$advanced.random_seed' != 0:
+    --random-seed='$advanced.random_seed'
+#end if
+    -S '$out_counts'
+
+
+#if str($cond_extra.prepender) != "none":
+#set $replacer = $input_bam.name.rsplit('.', 1)[0]
```
bgruening: You probably want to use $input.element_identifier as used here for example: https://github.com/galaxyproject/tools-iuc/blob/5ebd6c8453b49dd6a36e372eb1eb6e323bb7ad8a/tools/picard/picard_NormalizeFasta.xml#L13 - replacing spaces and so on.

mtekman: thanks, I was looking for something like this

PR #2007, opened 2018-07-27
Update deseq2.R

ashvark:
Colnames from the "sampleTable" should be assigned to the "tbl".

FOR CONTRIBUTOR:
* [x ] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [x ] - License permits unrestricted use (educational + commercial)
* [x ] - This PR updates an existing tool or tool collection

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help

### DIFF
--- a/tools/deseq2/deseq2.R
+++ b/tools/deseq2/deseq2.R
@@ -203,7 +203,7 @@ dir <- ""
 if (!useTXI & hasHeader) {
     countfiles <- lapply(as.character(sampleTable$filename), function(x){read.delim(x, row.names=1)})
     tbl <- do.call("cbind", countfiles)
-    rownames(sampleTable) <- colnames(tbl) # take sample ids from header
+    colnames(tbl) <- rownames(sampleTable) # take sample ids from header
 
     # check for htseq report lines (from DESeqDataSetFromHTSeqCount function)
     oldSpecialNames <- c("no_feature", "ambiguous", "too_low_aQual",
--- a/tools/deseq2/deseq2.xml
+++ b/tools/deseq2/deseq2.xml
@@ -196,7 +196,7 @@ Rscript '${__tool_directory__}/deseq2.R'
             <param name="normCounts" value="True"/>
             <output name="counts_out">
                 <assert_contents>
-                    <has_text_matching expression="untreat1\tuntreat2\tuntreat3\tuntreat4\ttreat1\ttreat2\ttreat3" />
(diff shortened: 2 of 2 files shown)


mvdbeek:
Thanks @ashvark, I think this is correct, but is there a chance you could add a test that would fail without this change? I'm a little surprised we haven't noticed this earlier.

ashvark:
Thanks for the feedback @mvdbeek. I will try to write a test and update.

mblue9:
@mvdbeek it's maybe not too surprising that it wasn't noticed as this change is just to take the sample ids from the filenames rather than from the header of the counts files. It only affects visualising the sample ids in the PCA and correlation heatmap (and the sampletable in the stdout). I don't know which is more preferred by people for this tool. I'm not currently using this tool so don't have a preference here. 

If this change is accepted the test of the header in the normalised counts output needs to be changed, it's failing with `No text matching expression 'untreat1\tuntreat2\tuntreat3\tuntreat4\ttreat1\ttreat2\ttreat3' was found in output file.` as it should now be the filenames `GSM461176_untreat_single.counts\tGSM461177_untreat_paired.counts\tGSM461178_untreat_paired.counts\tGSM461182_untreat_single.counts\tGSM461179_treat_single.counts\tGSM461180_treat_paired.counts\tGSM461181_treat_paired.counts`, and not the original header content.

mvdbeek:
Ah cool, then we just have to update the test data instead of adding a new test.

mblue9:
Nope, no updating test data needed, just this line in the test https://github.com/galaxyproject/tools-iuc/blob/master/tools/deseq2/deseq2.xml#L199

needs to be changed to:

`<has_text_matching expression="GSM461176_untreat_single.counts\tGSM461177_untreat_paired.counts\tGSM461178_untreat_paired.counts\tGSM461182_untreat_single.counts\tGSM461179_treat_single.counts\tGSM461180_treat_paired.counts\tGSM461181_treat_paired.counts" />`

ashvark:
Thanks @mvdbeek  and @mblue9 for the suggestions. I have updated the file with the suggested change.

bgruening:
Thanks all!

PR #2018, opened 2018-08-03
Trinotate: update to 3.1.1 + fixes

### LINKED ISSUE #1770: Trinotate: Pre-generated Resource SQLite database not found

abretaud:
FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help

So here's an update of trinotate to version 3.1.1
It also uses the new trinotate command to create the initial sqlite db (instead of downloading) (fixes #1770)
And it fixes some conda installation problems with the older trinotate packages


abretaud:
Ok, it's failing probably due to network: https://travis-ci.org/galaxyproject/tools-iuc/jobs/411786059#L2430
I'll retry next week

tools/trinotate/trinotate.xml
```diff
@@ -1,23 +1,22 @@
-<tool id="trinotate" name="Trinotate" version="3.0.1.0">
+<tool id="trinotate" name="Trinotate" version="3.1.1.0">
```
nsoranzo: Version can be just `3.1.1`

nsoranzo:
I think the error is `PASV ... couldn't connect to 141.161.180.197 port 1034: No route to host`, which I don't think would go away retrying. Is it possible to download using http instead of ftp?

abretaud:
Argh... The ftp.ebi.ac.uk mirror is accessible by http, I'll have to patch the conda package to change the url in the code

abretaud:
Aah green!

PR #2028, opened 2018-08-07
Workaround for using paired fastqc out in multiqc

mvdbeek:
I've arbitrarily chosen the reverse read, which IIRC often has lower Q
scores. Will be less necessary in the future if we start using fastp,
which was added in verison 1.6.

FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help


mblue9:
@mvdbeek looks good, is it possible to get both reads in the paired collection working? 
I tried this but it didn't work:
```
               #for $k, $file in enumerate($repeat2.fastqc_collection_type.input)
                    #set file_dir = os.path.join($repeat_dir, 'file_' + str($k))
                    #set file_path = os.path.join($file_dir, 'fastqc_data.txt')
                    mkdir '$file_dir' &&
                    #if str($repeat2.fastqc_collection_type.is_paired) == 'yes':
                        #set $file = $file.forward
                        #set file_dir2 = os.path.join($repeat_dir, 'file_' + str($k) + '_2')
                        #set file_path2 = os.path.join($file_dir2, 'fastqc_data.txt')
                        mkdir '$file_dir2' &&
                        #set $file2 = $file.reverse
                        ln -s '$file2' '$file_path2' &&
                    #end if
                    ln -s '$file' '$file_path' &&
                #end for
```

tools/multiqc/multiqc.xml
```diff
@@ -590,9 +596,21 @@ sp:
                             <option value="data">Raw data</option>
                             <option value="theoretical_gc">Theorectical GC</option>
                         </param>
-                        <param name="input" type="data" format="txt" multiple="true" label="FastQC output">
-                            <validator type="expression" message="MultiQC does not accept the HTML report generated by FastQC, only the Raw Data">value is not None and value.extension != "html"</validator>
-                        </param>
+                        <conditional name="fastqc_collection_type">
+                            <param name="is_paired" type="select" label="Is the fastqc output paired and in a list:paired collection ?">
```
mblue9: > Is the fastqc output paired and in a list:paired collection ?"

Will users know what list:paired is? I think I've seen "Paired collection" used more commonly on the user-facing side e.g. https://galaxyproject.org/tutorials/collections/  What about just "Is the fastqc output in a paired collection?"

tools/multiqc/multiqc.xml
```diff
@@ -590,9 +596,21 @@ sp:
...
+                                <option value="no">No, my data is single end</option>
+                                <option value="yes">Yes, the fastqc output is in a list:paired collection</option>
```
mblue9: See comment above for list:paired.

tools/multiqc/multiqc.xml
mvdbeek: That's unfortunate, we should definitely update the tutorial then :(.

mvdbeek:
> @mvdbeek looks good, is it possible to get both reads in the paired collection working?

Sure, but multiqc can't really do anything with it, at least last time I checked.

mblue9:
>Sure, but multiqc can't really do anything with it, at least last time I checked. If you add _1 or _2 it'll just show them as separate samples, making it harder to interpret in the overview. But I admit I haven't tried this recently.

Well I recently discovered the nifty MultiQC toolbox where you can hide/show samples, so you could hide/show e.g. all R1 or R2 files and so just see the first or second read. See screenshot below.

Personally I don't feel that comfortable with only letting people see the quality info for the second read and not the first, as what if there are cycle-specific issues? e.g. see here: https://sequencing.qcfail.com/articles/position-specific-failures-of-flowcells/ Would be better I think to let users be able to see it all, with the option to filter within multiqc, if this is possible.

screen shot 2018-08-09 at 9 50 01 am

mvdbeek:
I don't know if we still want to pursue it, fastp is the much better option here IMO ?

mblue9:
Yeah fastp does seem good, although it seems a little bit of a shame to leave this when it looks so close to working. Or is it just not possible to show both reads here? If not then yeah maybe no point in pursuing if only 1 read can be seen.

nsoranzo:
This PR needs a rebase (not related to the force push).

mvdbeek:
So the fastp output is not as comprehensive as fastqc, so I guess there is some merit in continuing this path. I'll have a look

mblue9:
Hi @mvdbeek just wondering if you managed to do any more with this?

As had a user discover paired collections and send me the screenshot below today reporting that they couldn't use MultiQC with their FastQC output.

image001

mvdbeek:
I didn't have time to look into this and am happy with fastp, but if you figure something out that'd be great. The excerpt you posted looks OK on first glance, what error did you get ?

mvdbeek:
If you put a list:paired collection into fastp you get a list back with a single report that can be used in multiqc.

mblue9:
ok good to know, thanks!

lparsons:
I admit, I haven't tried fastp yet as a replacement for FastQC, but I do find merit in presenting the FastQC data for each end along with other data in a MultiQC report. It's not perfect, but for me, it seems quite valuable and workable if the sample ids are `<sample>_forward` and `<sample>_reverse`.
image
This can be accomplished by using a join of the sample identifier and paired end indicator for FastQC reports with paired end data. Right now, I can achieve this by "flattening" the fastq collection before running FastQC. That's a bit of an extra step though.

lparsons:
Hmmm... Looking at `fastp` I don't think it's a reasonable alternative. It's going to output a new Fastq file, not just a report. I don't have a desire to filter or change the fastq, just report some basic stats. Filtering, etc. could be done later if needed. Still, fastp looks quite useful for those type of operations.

bernt-matthias:
I'm stumbling over this over and over again :) 

@lparsons suggestion seems to be the best option at the moment. The actual problem seems to be FastQC which sets the sample name in the raw data - which is the `element_identifier`. 

The change implemented here can equivalently be reached by using `unzip collection` and just passing the forward (or reverse reads). 

So maybe its just an matter of documentation, here in the help and ideally in the GTN (@bebatut).

An alternative might be to implement pair input for FastQC which would then start two calls to FastQC setting the sample name to `samplename_forward` and `samplename_reverse`. Which would yield the same result as @lparsons suggestion, but would save the user from creating "superficial" collections. Or implement pair input here and search replace the sample same, e.g. with sed. I don't see an alternative that allows to access the name of the sample name and forward and reverse data.

lparsons:
I like the idea of implementing pair input for FastQC which sets the sample name to `[element_id]_forward` and `[element_id]_reverse`. Seems the most elegant to me...

hexylena:
I hit this again, for the Nth time today and it's still incredibly annoying.

When iterating over list:paired, the element identifier just being forward/reverse seems very restricted. I really feel like it should include the parent element identifiters (joined with - or something)

I can fix it for myself just by prefixing with the loop index but I don't like this, it definitely needs to be fixed on the galaxy side, right? That when we have multiple=True and a collection input, that the iteration comes with more information about the path to that element within the collection?

I've added this to the GTN priorities since it's been bothering us for a long time, it's so difficult to train people to MultiQC for fastqc reports when we can't just throw any list/list:paired/paired at it and have it working.

bernt-matthias:
@pimarin fastp can take paired data https://github.com/galaxyproject/tools-iuc/blob/c75b6fda89f1cb0dce226ec983adda7f051a42e8/tools/fastp/fastp.xml#L232

pimarin:
I, Thank's @bernt-matthias, fastp can take a paired collection but not a `list of pairs` in fact

bernt-matthias:
@pimarin that is true. But as far as I know also the underlying tool can only process a single fastq / fastq pair.

Note that you can still use a paired collection as input, but galaxy will create a separate job for each pair in the collection.

pimarin:
Hi, this is not a problem to create multiple job for each input in the collection. In fact I tried the collection option as input using a generated `list of dataset pairs` to fastp only, it work (1 collection with 5 paired data gave me 5 paired output), but when I used fastp in a workflow it failed. So the fastp code seem to be ok but I don't understand the workflow layer impact, only an error as :
`cannot find 'forward' while searching for 'single_paired.paired_input.forward'`

pimarin:
And when I use manually fastp on list of dataset pairs followed by fastqc also manually it's fine, but not in workflow.
And also thank you for your help and reactivity

bernt-matthias:
I can't access the link to the workflow can you export it and share the file?

pimarin:
image
and the workflow : [workflow](http://192.168.53.17:8080/u/piermari/w/abromicpipelinebeta)

bernt-matthias:
The link to the workflow goes to your local network which I can't access. 

When you click on the workflow in the workflow menu, the dropdown offers the possibility to download the workflow. Please share this file .. or import it to a public Galaxy (usegalaxy.eu) where I can investigate it.

pimarin:
my bad... [workflow](https://drive.uca.fr/f/77ec02bb4ea04f2bb238/?dl=1)

lldelisle:
I propose a fix here: #7839

Issue #1354, opened 2017-06-09
SnpEff Download issues, originally v4.1.0 and still in v4.3k.0


jennaj:
The tool only makes the first genome from a nested .zip available in UI 

**Tool:** SnpEff Download Download a new database (Galaxy Version 4.1.0)

**Repo:** https://toolshed.g2.bx.psu.edu/repository?repository_id=93a5efea7e957b53&changeset_revision=7adfd0589f49

**Repo tool version:** 4.3k.0

_Is an update at http://usegalaxy.org the solution or is the problem still in the newer version of the tool as well or is there some configuration problem?_ 

ping @bgruening @davebx @natefoo 

---
**Version 4.1.0 issue**

The requested genome is in the list of the available genomes and the stderr for the database download tool _indicates that all of the data is downloaded_. However, only the first genome in the .zip bundle is available through the UI. 

Zip archives are not parsed through upload in a default Galaxy (nor at Galaxy Main). Only the first file is loaded. Not sure if related but seemed worth mentioning.

https://usegalaxy.org/u/jen/h/snpeff-database-download-test-history

<img width="970" alt="screen shot 2017-06-08 at 4 40 58 pm" src="https://user-images.githubusercontent.com/8999735/26955562-5aca7866-4c6c-11e7-92e2-c7e7cf682bf1.png">

**Genome requested by the tool:**

https://www.ncbi.nlm.nih.gov/genome/?term=txid264462[Organism:noexp]
**Bdellovibrio_bacteriovorus_HD100_uid61595**
refseq: NC_005363.1
INSDC: BX842601.2

**Genome accessible by SnpEff tool:**

https://www.ncbi.nlm.nih.gov/genome/486?genome_assembly_id=167053
**Bacillus_thuringiensis_BMB171_uid49135**
chr refseq: NC_014171.1	
chr INSDC: CP001903.1
plsm refseq: NC_014172.1
plsm INSDC: CP001904.1

Examining the target .zip when obtained directly (using the same path that the tool uses) shows that the zip does have the target genome in the set. The genome available through the UI is the first in the archive.

> $ curl https://managedway.dl.sourceforge.net/project/snpeff/databases/v4_1/snpEff_v4_1_NCBI_bacterial_4.zip > snpEff_v4_1_NCBI_bacterial_4.zip
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
[...]

PR #2059, opened 2018-08-31
Fixed snpeff download to result in correct genome version assignment in the metadata

hepcat72:
FOR CONTRIBUTOR:

I **believe** that this PR is a possible resolution to issue #1354, but I'm not 100% sure that's precisely what I was working on.  The issue I was trying to resolve was simply the incorrect assignment of the metadata item "genome_version".  That assignment is made in galaxy's text.py code, which assumes a single directory named with the genome version and containing a snpEffectPredictor.bin file.  I just changed the download tool to make that assumption correct.  I don't know about other tools and their possible issues with genome_version assignment.

It took me some effort to arrive at this solution.  Refer to PR #6658 (which I haven't yet closed) for some discussion on the topic.

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help


tool_collections/snpeff/snpEff_download.xml
```diff
@@ -9,7 +9,10 @@
     <expand macro="stdio" />
     <expand macro="version_command" />
     <command><![CDATA[
-snpEff download -dataDir '$snpeff_db.files_path' -v '$genome_version'
+        mkdir temp &&
+        snpEff download -dataDir `pwd`/temp -v '$genome_version' &&
```
lparsons: Not sure, but would `$PWD` be preferred here? Also, does it need to be in quotes (*e.g.* `"$PWD/temp"`)?

hepcat72: Definitely quoted would be better. I'll try the change.

hepcat72: Tool me awhile to realize I had to escape the dollar sign so that (I assume?) cheetah wasn't trying to mess with it, but I have made this change.

nsoranzo: Yep, the `\` is needed to prevent Cheetah from trying to assign a value to it. That's kind-of mentioned in https://docs.galaxyproject.org/en/latest/dev/schema.html#reserved-variables , but could be highlighted better.

hepcat72: Yeah, I didn't think to look in the Reserved Variables section.  I had bee-lined to the environment_variable stuff. Maybe a mention there...

nsoranzo:
@lparsons If you want to take a final look before merging, I've added a few more changes while at it.

lparsons:
Looks good to me.

PR #2067, opened 2018-09-06
WIP: Add bandage (Conda package in progress)

willemdek11:
FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help


tools/bandage/bandage.xml
```diff
@@ -0,0 +1,30 @@
+<tool id="bandage" name="bandage" version="0.8.1.0">
+    <description>a Bioinformatics Application for Navigating De novo Assembly Graphs Easily </description>
+    <requirements>
+        <requirement type="package" version="0.8.1">bandage</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+Bandage image '$gfa' 'out.png'
```
bernt-matthias: you may directly use '$outfile' here

tools/bandage/bandage.xml
```diff
@@ -0,0 +1,30 @@
...
+    <inputs>
+        <param name="gfa" type="data" format="tabular" label="Graphical Fragment Assembly" />
+    </inputs>
+    <outputs>
+        <data name="outputfile" format="png" label="${tool.name} on ${on_string}: Assembly Graph Image" />
+    </outputs>
+    <tests>
+        <test>
+            <param name="gfa" value="gfa.tabular"/>
+            <output name="outputfile" file="out.png"/>
```
bernt-matthias: specify the type of the input and output

bernt-matthias:
Excellent start. Few comments: 

- rename to bandage_image.xml  and tool to "bandage image" in order to allow to add the other commands (later)
- will you add the other options of the tool? How about the different input types?

tools/bandage/bandage.xml
```diff
@@ -4,18 +4,21 @@
...
     </inputs>
     <outputs>
-        <data name="outputfile" format="png" label="${tool.name} on ${on_string}: Assembly Graph Image" />
+        <data name="outfile" format="png" label="${tool.name} on ${on_string}: Assembly Graph Image" />
     </outputs>
     <tests>
         <test>
-            <param name="gfa" value="gfa.tabular"/>
-            <output name="outputfile" file="out.png"/>
+            <param name="gfa" format="tabular" value="gfa.tabular"/>
```
bgruening: This should be ftype="tabular" not format. I know its confusing, we need to fix this.

willemdek11:
@bernt-matthias @bgruening Thanks for reviewing. Bandage checks the extension of the output. So I'll leave it .png for the moment. I'll add the other output and input options and Bandage info option later.

bernt-matthias:
In this case you need to use the `<data ... from_work_dir="out.png"/>`

For other output types one could add a <filter> to produce the output only for the selected format.

willemdek11:
It seems like everything is working now! :)

tools/bandage/bandage.xml
```diff
@@ -0,0 +1,83 @@
...
+        <requirement type="package" version="0.8.1">bandage</requirement>
+    </requirements>
+    <version_command>Bandage --version</version_command>
+    <command detect_errors="exit_code">
+    <![CDATA[
+Bandage
+#if $mode.select_mode == 'info':
+    info
+    '$input_file'
+    #if $mode.tsv:
```
shiltemann: you don't need the if statement here, just adding `$mode.tsv` is enough (will be replaced by either `--tsv` or empty string)

tools/bandage/bandage.xml
```diff
@@ -0,0 +1,83 @@
+<tool id="bandage" name="bandage" version="0.8.1.0">
+    <description>a Bioinformatics Application for Navigating De novo Assembly Graphs Easily </description>
```
shiltemann: I think we can get rid of "a Bioinormatics Application" here to make it a bit more concise, just `navigate de-novo assembly graphs` orso?

tools/bandage/bandage.xml
```diff
@@ -0,0 +1,83 @@
...
+            <param name="select_mode" type="select" label="Info or Image">
+                <option selected="True" value="info">Info</option>
+                <option value="image">Image</option>
+            </param>
+            <when value="info">
+                <param argument="--tsv" type="boolean" checked="false" truevalue="--tsv" falsevalue="" label="Output the information in a single tab-delimited line starting with the graph file"/>
+            </when>
+            <when value="image">
+                <param argument="--height" type="integer" min="1" value="1000" optional="True" label="Image height (default: 1000). If only height or width is set, the other will be determined automatically. If both are set, the image will be exactly that size."/>
+                <param argument="--width" type="integer" min="1" optional="True" label="Image width (default: not set). If only height or width is set, the other will be determined automatically. If both are set, the image will be exactly that size."/>
```
shiltemann: can you keep the label small, and put the rest in `help="If only height or .."`, and same for height param

tools/bandage/bandage.xml
```diff
@@ -0,0 +1,83 @@
...
+                <param name="output_format" type="select" label="Produce jpg, png or svg file?">
+                    <option selected="True" value="jpg">.jpg</option>
+                    <option value="png">.png</option>
+                  <option value="svg">.svg</option>
+                </param>
+            </when>
+        </conditional>
+    </inputs>
+    <outputs>
+        <data name="outfile" format="auto" from_work_dir="out.*" label="${tool.name} on ${on_string}: Assembly Graph Image"/>
```
shiltemann: looks like autodetection of format is not working after all, try changing format more explicitly based on value of `output_format` parameter, see [here](https://github.com/galaxyproject/tools-iuc/blob/57493fdc88e6128bed127993464f2a6480e2774d/tools/hicexplorer/hicTransform.xml#L39) for an example

tools/bandage/bandage.xml
```diff
@@ -0,0 +1,83 @@
...
+            <param name="height" value="100"/>
+            <param name="width" value="100"/>
+            <param name="output_format" value="png"/>
+            <output name="outfile" ftype="data" file="out.png" compare="sim_size" delta="120000"/>
+        </test>
+        <test>
+            <param name="input_file" ftype="tabular" value="gfa.tabular"/>
+            <param name="select_mode" value="info"/>
+            <param name="tsv" value="True"/>
+            <output name="outfile" ftype="data" file="out.txt" compare="sim_size" delta="100"/>
```
shiltemann: assert contents with a regex?

shiltemann:
@willemdek11  just another thought, do we want to make a wrapper per bandage command? so make one `bandage_image`, and one `bandage_info` wrapper, instead of having both commands together in this one? it will simplify the logic in the wrapper, and be more intuitive for people already familiar with bandage via the commandline, and indeed facilitate adding the other bandage commands at a later time as well

willemdek11:
@shiltemann I created them separately now and applied your suggestions. Only thing I couldn't make work is the regex in the test of Bandage_info.

bernt-matthias:
> Only thing I couldn't make work is the regex in the test of Bandage_info. 

Can you show us the test that did not work?

willemdek11:
@shiltemann Updated the macros, hope I did it the right way.

willemdek11:
@shiltemann Now the "regex" should work :)

PR #2072, opened 2018-09-11
Jbrowse: small change to match/match_part tracks

abretaud:
FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help

Working on https://github.com/galaxyproject/training-material/pull/950, I have some gff which contains match/match_part features, but with varying parent type name (match, expressed_sequence_match, protein_match) in the same file. Currently I can't display all the features with their subfeatures as I'm forced to specify a single parent type.
In fact CanvasFeatures tracks don't really need to specify it, by default it will auto detect everything. So here's a patch to allow leaving the option empty and let it be autodetected.
For HTMLFeautres track type, it will just display features but not subfeatures which is not dramatic I think (same as if I entered the wrong name in the field).

What do you think @erasche?

### DIFF
--- a/tools/jbrowse/jbrowse.py
+++ b/tools/jbrowse/jbrowse.py
@@ -572,7 +572,9 @@ class JbrowseConnector(object):
 
         if 'match' in gffOpts:
             config['glyph'] = 'JBrowse/View/FeatureGlyph/Segments'
-            cmd += ['--type', gffOpts['match']]
+            if bool(gffOpts['match']):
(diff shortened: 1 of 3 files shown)


erasche:
Oh, yeah, that's definitely sub-optimal. I guess I'd been playing around and following the docs and interpreted it as requiring it.

PR #2075, opened 2018-09-12
snpeff - fix macro token names

martenson:
they are case sensitive I think

screenshot 2018-09-12 12 35 00


martenson:
ping @hepcat72  @mvdbeek

PR #2077, opened 2018-09-13
Quote database name in kraken

peterjc:
FOR CONTRIBUTOR:

Ensures the database name is quoted (and puts a space in the test case), and also switches filename quoting to use single quotes rather than double quotes.

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help

### DIFF
--- a/tool_collections/kraken/macros.xml
+++ b/tool_collections/kraken/macros.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <macros>
-    <token name="@WRAPPER_VERSION@">1.2.3</token>
+    <token name="@WRAPPER_VERSION@">1.2.4</token>
     <xml name="requirements">
         <requirements>
             <requirement type="package" version="0.10.6_eaf8fb68">kraken</requirement>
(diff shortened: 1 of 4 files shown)


peterjc:
TravisCI fails in samtools, logged as #2077 

See https://travis-ci.org/peterjc/tools-iuc/builds/428093510 where I removed the samtools collection, and these kraken changes pass

peterjc:
Interesting - whatever the samtools problem was, it didn't happen here - just on the forked repo's test.

tool_collections/kraken/macros.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <macros>
-    <token name="@WRAPPER_VERSION@">1.2.3</token>
+    <token name="@WRAPPER_VERSION@">1.2.4</token>
```
bgruening: We also need to bump the tool version number.

peterjc:
If #2080 is merged, it would include this work.

tool_collections/kraken/macros.xml
peterjc: This is the Galaxy tool version - or do you mean convert ``@WRAPPER_VERSION@`` to follow the ``@TOOL_VERSION@+galaxyN`` style?

bgruening: Ah, you are right, sorry!

peterjc: OK. No problem, this was worth clarifying.

Sadly given the tool version is currently ``0.10.6_eaf8fb68`` (and soon after moved to ``1.0``), I don't think we can switch to ``@TOOL_VERSION@+galaxyN`` as the sort order would break given the wrapper is already at ``1.2.x``

peterjc:
Thank you - I'll rebase the follow up PR now.

PR #2110, opened 2018-09-28
samtools: final changes for 1.9 update

bernt-matthias:
Some changes in the new 1.9 samtools that were missed in the previous commits. I already did some changes. I will definitely fix the regions which are buggy in some tools at the moment.

For the other open points I need comments!

- depth: fixed macro name
- rmdup: deprecated
- bam_to_cram: created links in data directory, tried to make it identical with view call (because it uses the same test data)
- bedcov: new parameters

- [x] Do we want/need suite_samtools as in the devteam repo
  - I would definitely say yes!
  - Need input on how to realize this
- [ ] introduce regions macros/tokens (see stats) in the other lools (view is buggy at the moment)
  - this seems complicated due to different parameter names and different nesting levels in tools
- [ ] samtools_bedcov I ould expect test 3 to fail (since mapq 40 should filter some reads)
- [ ] bam_to_cram don't understand why tests fails (should be the same as in samtools view which works - they use the same test data)
- [ ] samtools_split test fails?
- [ ] samtools_phase stumbles during metadata setting because of empty bam file which galaxy cant index
```
  File "/tmp/tmpH3VbtC/galaxy-dev/lib/galaxy/jobs/runners/local.py", line 152, in queue_job
    self._finish_or_resubmit_job(job_state, stdout, stderr, exit_code)
  File "/tmp/tmpH3VbtC/galaxy-dev/lib/galaxy/jobs/runners/__init__.py", line 433, in _finish_or_resubmit_job
    job_state.job_wrapper.finish(stdout, stderr, exit_code, check_output_detected_state=check_output_detected_state)
  File "/tmp/tmpH3VbtC/galaxy-dev/lib/galaxy/jobs/__init__.py", line 1301, in finish
    dataset.datatype.set_meta(dataset, overwrite=False)
  File "/tmp/tmpH3VbtC/galaxy-dev/lib/galaxy/datatypes/binary.py", line 429, in set_meta
    pysam.index(dataset.file_name, index_file.file_name)
  File "/home/berntm/.planemo/gx_venv_None/local/lib/python2.7/site-packages/pysam/utils.py", line 75, in __call__
    stderr))
SamtoolsError: 'samtools returned with error 1: stdout=, stderr=samtools index: "/tmp/tmpH3VbtC/files/000/dataset_5.dat" is in a format that cannot be usefully indexed\n'
```

FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help


mvdbeek:
I think you have to update the symlink for the samtools_rmdup repo (https://travis-ci.org/galaxyproject/tools-iuc/jobs/435254193#L659)

bernt-matthias:
I added the xml of the old tool. Seems more reasonable to me, instead of linking.

nsoranzo:
This PR needs a rebase or a merge.

bernt-matthias:
Would like to finish this. Tests are working locally. Here something fails.

bernt-matthias:
There seems to be a problem with `owner` in `.shed.yml`

tool_collections/samtools/.shed.yml
```diff
@@ -0,0 +1,12 @@
+auto_tool_repositories:
```
wm75:
```suggestion
name: samtools
owner: iuc
description: Galaxy wrappers for the SAMtools package
homepage_url: http://www.htslib.org/
long_description: |
  SAM (Sequence Alignment/Map) format is a generic format for storing large
  nucleotide sequence alignments. The Samtools package of command line
  utilities lets you read, write, edit, index and view data in SAM format,
  and in the related binary formats BAM and CRAM.
remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/master/tool_collections/samtools
type: unrestricted
categories:
- SAM
auto_tool_repositories:
```

You need to add top-level information, including owner!

tool_collections/samtools/.shed.yml
```diff
@@ -0,0 +1,12 @@
...
+  name_template: "{{ tool_id }}"
+  description_template: "Wrapper for samtools application {{ tool_name }}."
+suite:
+  name: "suite_samtools"
+  description: "A suite of Galaxy tools designed to work with version 1.2 of the SAMtools package."
```
wm75: I would just copy the top-level description and long_description from above here, so it will be shown for the suite.

tool_collections/samtools/.shed.yml
```diff
@@ -0,0 +1,12 @@
...
+  long_description: |
+    SAM (Sequence Alignment/Map) format is a generic format for storing large nucleotide sequence
+    alignments.  These associated Galaxy utilities consist of a Galaxy Data
+    Manager contained in the repository named data_manager_sam_fasta_index_builder and Galaxy tools
+    contained in several separate repositories.
+  owner: iuc
```
wm75:
```suggestion

```

tool_collections/samtools/.shed.yml
bernt-matthias: just adapted the version number. I think that duplicating the long_description would not look nice in the toolshed.

bernt-matthias:
@wm75 thanks for the comments. The shed.yml seems to work now and multithreading has been added where it was missing and is a documented command line parameter (axcluding the view based tools).

bgruening:
> RunToolException: Error creating a job for these tool inputs - Parameter input_reference requires a value, but has no legal values defined.

Rerunning does not help and it passes for me locally. Mh, any ideas? The tests that are failing are all with loc files as far as I can see.

wm75:
@bgruening can you trigger the CI build once more so that the errors get logged again (currently the tests are logged as canceled without any useful information).
The tests pass locally for me, too.

bgruening:
Is there any reason we keep the separate folders? Can we move everything into a suite-like structure without breaking the tool-linage?

Maybe something is not following the symlinks, which is wired as it also works for me locally with 19.01

bernt-matthias:
> Is there any reason we keep the separate folders?

If we move everything, then we should structure the test data in sub folders. This could be a bit of work. 

> Can we move everything into a suite-like structure without breaking the tool-linage?

Can't answer this. 

If we have a positive answer to this question I would be wiling to do it. 

I also do not understand the failing tests. They are all about locally cached data. Symlinks should not be a problem, because this is only the case in the `samtools view` case. My best guess is that I use the parameter `ref`twice (in two different paths of the conditional), could this be confusing. I like to do this since it simplifies the cheetah code a bit.

mvdbeek:
There are also 2 tools that fail to load altogther
```
2019-02-21 09:35:00,055 INFO  [galaxy.tools.toolbox.base] Parsing the tool configuration /tmp/tmpVCK_Qn/tool_conf.xml
2019-02-21 09:35:00,117 ERROR [galaxy.tools.toolbox.base] Failed to load potential tool /home/travis/build/galaxyproject/tools-iuc/tool_collections/samtools/samtools_mpileup/samtools_mpileup.xml.
  File "/tmp/tmpVCK_Qn/galaxy-dev/lib/galaxy/tools/toolbox/base.py", line 719, in quick_load
    tool = self.load_tool(tool_file)
  File "/tmp/tmpVCK_Qn/galaxy-dev/lib/galaxy/tools/toolbox/base.py", line 755, in load_tool
    tool = self.create_tool(config_file=config_file, repository_id=repository_id, guid=guid, **kwds)
  File "/tmp/tmpVCK_Qn/galaxy-dev/lib/galaxy/tools/__init__.py", line 272, in create_tool
    return self._create_tool_from_source(tool_source, config_file=config_file, **kwds)
  File "/tmp/tmpVCK_Qn/galaxy-dev/lib/galaxy/tools/__init__.py", line 275, in _create_tool_from_source
    return create_tool_from_source(self.app, tool_source, **kwds)
  File "/tmp/tmpVCK_Qn/galaxy-dev/lib/galaxy/tools/__init__.py", line 220, in create_tool_from_source
    tool = ToolClass(config_file, tool_source, app, **kwds)
  File "/tmp/tmpVCK_Qn/galaxy-dev/lib/galaxy/tools/__init__.py", line 448, in __init__
    raise e
KeyError: 'fasta_indexes'
2019-02-21 09:35:00,134 ERROR [galaxy.tools.toolbox.base] Failed to load potential tool /home/travis/build/galaxyproject/tools-iuc/tool_collections/samtools/sam_to_bam/sam_to_bam.xml.
  File "/tmp/tmpVCK_Qn/galaxy-dev/lib/galaxy/tools/toolbox/base.py", line 719, in quick_load
    tool = self.load_tool(tool_file)
  File "/tmp/tmpVCK_Qn/galaxy-dev/lib/galaxy/tools/toolbox/base.py", line 755, in load_tool
    tool = self.create_tool(config_file=config_file, repository_id=repository_id, guid=guid, **kwds)
  File "/tmp/tmpVCK_Qn/galaxy-dev/lib/galaxy/tools/__init__.py", line 272, in create_tool
    return self._create_tool_from_source(tool_source, config_file=config_file, **kwds)
  File "/tmp/tmpVCK_Qn/galaxy-dev/lib/galaxy/tools/__init__.py", line 275, in _create_tool_from_source
    return create_tool_from_source(self.app, tool_source, **kwds)
  File "/tmp/tmpVCK_Qn/galaxy-dev/lib/galaxy/tools/__init__.py", line 220, in create_tool_from_source
    tool = ToolClass(config_file, tool_source, app, **kwds)
  File "/tmp/tmpVCK_Qn/galaxy-dev/lib/galaxy/tools/__init__.py", line 448, in __init__
    raise e
KeyError: 'fasta_indexes'
```
so that may or may not be related, I'll check this out.

mvdbeek:
So I can reproduce the error I posted aboce locally if I test the tool_collections/samtools folder, while it works fine if I test tool_collections/samtools/samtools_mpileup ... which is using a different mechanism for loading tools.

bernt-matthias:
Which also brings us back to @bgruening's question if can flatten the collection to a single dir. But determining the source or the error is also a good idea.

wm75:
Thanks @mvdbeek, looking at kraken, which is organized similarly, it appears that test_database.loc, tool_data_table_conf.xml.test are supposed to live in the top-level directory. Inside the subfolders there are symlinks to these files. Maybe that structure is there to satisfy planemo no matter where the tests are run from?

mvdbeek:
That sounds reasonable ... worth a try ?

mvdbeek:
On the other hand I guess planemo could be smarter in detecting which directories have changed:
```
0.01s$ cat changed_repositories.list
deprecated/tool_collections/samtools/samtools_rmdup
tool_collections/samtools
tool_collections/samtools/bam_to_cram
tool_collections/samtools/samtools_bedcov
tool_collections/samtools/samtools_calmd
tool_collections/samtools/samtools_flagstat
tool_collections/samtools/samtools_idxstats
tool_collections/samtools/samtools_phase
tool_collections/samtools/samtools_split
tool_collections/samtools/samtools_stats
install.14
0.01s$ cat changed_repositories_chunk.list
deprecated/tool_collections/samtools/samtools_rmdup
tool_collections/samtools
tool_collections/samtools/bam_to_cram
tool_collections/samtools/bam_to_sam
tool_collections/samtools/cram_to_bam
tool_collections/samtools/sam_to_bam
```

wm75:
forget about kraken - turns out every single test in it fails with something like
`RunToolException: Error creating a job for these tool inputs - Parameter kraken_database requires a value, but has no legal values defined.` when run from the top-level directory.
This is really weird.

wm75:
ok, I think I've found what needs to be done:
1) There needs to be a `test-data` folder in the top-level `samtools` directory. This folder must contain the `cached_locally` folder copied over from any tool that currently contains one.
2) There also needs to be a top-level `tool_data_table_conf.xml.test` file (again you can copy it from one of the tools that brings one).
3) Now you can replace all actual `cached_locally` subfolders and the `tool_data_table_conf.xml.test` files in all tools with symlinks to the corresponding top-level items.
This makes all errors disappear when running planemo test directly from the samtools folder, and because of the symlinks it doesn't break things when you step inside any tool folder for testing it specifically.

wm75:
Note: The `kraken` tools have got it almost right, but don't put the top-level `test_database.loc` file in a top-level `test-data` folder. So the `${__HERE__}/test-data/test_database.loc` destination defined in their top-level `tool_data_table_conf.xml.test` file is not found.

bernt-matthias:
Thanks for the clarifications. 

To me it seems that removing the hierarchy and remove the separate folders is really an option that would simplify development. I don't see an advantage of keeping the separate folders. 

Opinions? Reasons? :)

bgruening:
The only reason I can imagine is that we want to keep the current toolshed repositories and don't break tool-linage. That means we need to make sure that the tools-ids and the suite-shed.yml file will create the same structure as the current one.

For example, someone needs to check if the tool_id is always the repository name atm. If this is the case its easy I assume.

bernt-matthias:
@bgruening tool-ids seems easy. The suite-shed.yml did not exist so far .. so this seems also no problem. But I guess all the separate .shed.yml files would go. Would this be a problem.

bgruening:
>  But I guess all the separate .shed.yml files would go. Would this be a problem?

No.

Imho, we just need to make sure the tool-id is the same as the repo-name it should end up.

wm75:
FWIW, I like PRs that are limited to one topic. That's just my opinion, and it's heavily biased by the fact that I'd like to see the multithreading support for calmd in the MTS rather sooner than later :wink:, but I'd only fix the tests here, then get it merged.
After that, one could still look at flattening the collection and moving everything into tools. If that turns out to be unproblematic for samtools, one could actually consider the same for all packages in tool_collections and merge them into tools in one big effort?

tool_collections/samtools/.shed.yml
```diff
@@ -0,0 +1,24 @@
...
+remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/master/tool_collections/samtools
+type: unrestricted
+categories:
+- SAM
+auto_tool_repositories:
+  name_template: "{{ tool_id }}"
+  description_template: "Wrapper for samtools application {{ tool_name }}."
+suite:
+  name: "suite_samtools"
+  description: "A suite of Galaxy tools designed to work with version 1.2 of the SAMtools package."
```
wm75: Mentioning a specific version of samtools in the suite description is unnecessary and, IMHO, is prone to getting outdated sooner or later, like in the current case ;)

tool_collections/samtools/.shed.yml
```diff
@@ -0,0 +1,24 @@
...
+  long_description: |
+    SAM (Sequence Alignment/Map) format is a generic format for storing large nucleotide sequence
+    alignments.  These associated Galaxy utilities consist of a Galaxy Data
+    Manager contained in the repository named data_manager_sam_fasta_index_builder and Galaxy tools
+    contained in several separate repositories.
```
wm75: Talking about the data manager in this way doesn't make sense when it's not included in the suite. Either make it a requirement for the suite, drop the mention, or state that it is desirable to install the DM alongside the suite.

tool_collections/samtools/.shed.yml
bernt-matthias: ok.

tool_collections/samtools/.shed.yml
bernt-matthias: requirement sounds good (but we might want a version bump there for consistency). Can you tell me where I find documentation on specifying requirements in shed.yml. I never could find recent docs for shed.yml anyway.

wm75: not sure whether that's current best practice, but the planemo docs mention

https://github.com/galaxyproject/planemo/blob/b5da69cb22028505df0e84a275f5078c1885902e/tests/data/repos/multi_repos_flat_flag_suite/.shed.yml#L13

to illustrate how you can add an additional repo to a suite.

wm75: oh, and just found this right next to it: https://github.com/galaxyproject/planemo/blob/master/tests/data/repos/suite_auto/.shed.yml

It uses the samtools tool collection as an example. Interstingly, it mentions all tools directly instead of relying on autodetection. I was actually wondering whether autodetection would find tools in subfolders.

bernt-matthias:
@wm75 How can we check what will be inluded in the suite in these two cases. I somehow would like to exclude the the converters that are based on samtools view.

bgruening:
@bernt-matthias you could push to the TTS.

bernt-matthias:
Test 1: no `include_repositories` all tools in the subdirs are included.
Test 2: addiing include for the data manager just adds the data manager to the tools in the subdirs

I guess that explicitly adding a subset of the tools in the subdirs will change nothing (will test tomorrow).

I would like to exclude some tools from the suite, but this seems not possible at the moment, or?

tool_collections/samtools/.shed.yml
```diff
@@ -0,0 +1,27 @@
...
+  name: "suite_samtools"
+  description: "A suite of Galaxy tools designed to work with the SAMtools package."
+  long_description: |
+    SAM (Sequence Alignment/Map) format is a generic format for storing large
+    nucleotide sequence alignments. The Samtools package of command line
+    utilities lets you read, write, edit, index and view data in SAM format,
+    and in the related binary formats BAM and CRAM.
+  include_repositories:
+    - owner: "devteam"
+      name: "data_manager_sam_fasta_index_builder"
```
bgruening: I would not put this into the suite.
We never did this before afaik and many people do not need any data_managers if they use CVMFS.

bernt-matthias: ok. good point.

bernt-matthias:
OK. I guess that's it from my side. There are a few points that could be addressed but these should go to a separate PR: 

- multiple tools use regions filters, these should be unified with macros.
- in my opinion the samtools view based tools should be deprecated.

bgruening:
@bernt-matthias you have tested that the tool-linage does not break with the TTS, right?

@nsoranzo @mvdbeek @wm75 is this also good to go for you?

bernt-matthias:
@bgruening no. I still do not have a clear understanding of this concept. But they are in the TTS and you may have a look.

wm75:
@bgruening in its current form, the PR does not change the folder structure of the tools so tool-lineage should not be affected because of this. However, I'm not sure how the global `owner: iuc` in the new .shed.yml would affect the ownership associated with individual tools (most are currently owned by devteam). I guess there is potential here for affecting the tool lineage? This can only be checked through a TTS upload by an iuc member, I think, because the current test upload is simply owned by @bernt-matthias altogether.

@bernt-matthias You should be able to exclude individual tools from the suite by using `ignore:` in the .shed.yml. It's used just like `include_repositories:`, *i.e.*, followed by a list of tools to exclude.

bernt-matthias:
will try `ignore` right away

bernt-matthias:
`ignore` seem not to work. I tested: 

```
  ignore:
    - name: samtools_bam_to_cram
      owner: matthias
    - name: bam_to_sam
      owner: matthias
...
```
and 
```
  ignore:
    - samtools_bam_to_cram
    - bam_to_sam
...
```

or do I need to specify a file that needs to be ignored?

nsoranzo:
It's `exclude:`, with filenames. See tools/bcftools/.shed.yml

nsoranzo:
Ah, and it seems to be broken! See https://travis-ci.org/galaxyproject/tools-iuc/jobs/496627461 , where last Thursday 2019-02-21 the 2 mentioned tools were pushed to the TSs nonetheless.

wm75:
So I've played around with the different options a bit locally and here's a summary:

1) `exclude:` doesn't seem to work (as @nsoranzo observed already)
2) `ignore:` works (though AFAIU it's considered legacy and one should use exclude instead), but it isn't doing exactly what I thought it would do:
it only works at the top level, not inside suite, so only influences which repositories get created in the toolshed, but not which ones get included in the suite. In fact, when a repo is ignored that already exists in the toolshed its latest revision seems to get changed to not installable.  So this is really meant for a different usecase.  Sorry for mixing things up.
3) Your current top-level .shed.yml is conflicting with the .shed.yml files in the subfolders (I didn't really think of those before): The subfolder .shed.yml files already take care of generating all the individual repos and they are owned by different users (some  iuc, some devteam), but your `auto_tool_repositories:` in the new top-level .shed.yml tries to specify all those repos again declaring `owner` to be iuc. This is a *different* task from declaring the suite further down in the file.
4) I think the cleanest approach would be something like:
```
owner: iuc
categories:
  - SAM
suite:
  name: suite_samtools
  description: "A suite of Galaxy tools designed to work with version 1.2 of the SAMtools package."
  long_description: |
    SAM (Sequence Alignment/Map) format is a generic format for storing large nucleotide sequence
    alignments.  These associated Galaxy utilities consist of a Galaxy Data
    Manager contained in the repository named data_manager_sam_fasta_index_builder and Galaxy tools
    contained in several separate repositories.
  include_repositories:
  - name: samtools_bam_to_cram
    owner: iuc
  - name: samtools_cram_to_bam
    owner: iuc
  - name: samtools_bedcov
    owner: devteam
  - name: samtools_calmd
    owner: devteam
  - name: samtools_flagstat
    owner: devteam
  - name: samtools_idxstats`
    owner: devteam
  - name: samtools_mpileup
    owner: devteam
  - name: samtools_phase
    owner: devteam
  - name: samtools_reheader
    owner: devteam
  - name: samtools_slice_bam
    owner: devteam
  - name: samtools_sort
    owner: devteam
  - name: samtools_split
    owner: devteam
  - name: samtools_stats
    owner: devteam
```
for the new .shed.yml, which does not try to generate any new repository, but just the suite, and explicitly states which repos identified by name and owner are to be part of that suite.
I checked that this works as expected locally. Feel free to modify the example to whatever set of repos looks reasonable to you.
Since you are then not messing with any existing repos, there is no concern over breaking any tool lineage.

bernt-matthias:
@wm75 thanks. I implemented exactly as suggested .. but removed the tools that derive from `samtools view`.

bernt-matthias:
Ready from my side .. if the tests still pass.

bernt-matthias:
They passed. Could be merged if no one has any open wishes.

mvdbeek:
Did you try pushing to the TTS to see if all works as expected ?

bernt-matthias:
No. Will do immediately.

mvdbeek:
planemo shed_update --shed_target testtoolshed --shed_email "$SHED_EMAIL" --shed_password "$SHED_PASSWORD" --force_repository_creation "$DIR" is what we let travis do, can you try this ?

bernt-matthias:
Unfortunately

`planemo shed_update --shed_target testtoolshed --force_repository_creation  --shed_email "m.bernt@ufz.de" --shed_password  *******`

also gives 

```Unexpected HTTP status code: 400: {"err_msg": "You already own a repository named <b>suite_samtools<\/b>, please choose a different name.", "err_code": 400008}
Repository [suite_samtools] does not exist in the targeted Tool Shed.
Failed to update repository 'suite_samtools' as it does not exist on the test Tool Shed.```

The error message is really strange: reads like "suite_samtools present.... suite_samtools not present"

bernt-matthias:
Unfortunately

`planemo shed_update --shed_target testtoolshed --force_repository_creation  --shed_email "m.bernt@ufz.de" --shed_password  *******`

also gives 

```Unexpected HTTP status code: 400: {"err_msg": "You already own a repository named <b>suite_samtools<\/b>, please choose a different name.", "err_code": 400008}
Repository [suite_samtools] does not exist in the targeted Tool Shed.
Failed to update repository 'suite_samtools' as it does not exist on the test Tool Shed.```

The error message is really strange: reads like "suite_samtools present.... suite_samtools not present"

bernt-matthias:
Unfortunately

`planemo shed_update --shed_target testtoolshed --force_repository_creation  --shed_email "m.bernt@ufz.de" --shed_password  *******`

also gives 

```Unexpected HTTP status code: 400: {"err_msg": "You already own a repository named <b>suite_samtools<\/b>, please choose a different name.", "err_code": 400008}
Repository [suite_samtools] does not exist in the targeted Tool Shed.
Failed to update repository 'suite_samtools' as it does not exist on the test Tool Shed.```

The error message is really strange: reads like "suite_samtools present.... suite_samtools not present"

bernt-matthias:
Unfortunately

`planemo shed_update --shed_target testtoolshed --force_repository_creation  --shed_email "m.bernt@ufz.de" --shed_password  *******`

also gives 

```Unexpected HTTP status code: 400: {"err_msg": "You already own a repository named <b>suite_samtools<\/b>, please choose a different name.", "err_code": 400008}
Repository [suite_samtools] does not exist in the targeted Tool Shed.
Failed to update repository 'suite_samtools' as it does not exist on the test Tool Shed.```

The error message is really strange: reads like "suite_samtools present.... suite_samtools not present"

bgruening:
With the latest commit I was able to push it to the TTS with:

planemo shed_upload -t testtoolshed --force_repository_creation tool_collections/samtools/


Can someone doublecheck? I would like to merge this finally ... thanks everyone!

bernt-matthias:
Wonderful. Linter errors seem not to related to samtools. I guess we need to rebase (also because of the recent force push)?

mvdbeek:
The linter error is not related, but we need to fix it (IMO just drop it, linking the reader to a cpp file is a bit of a joke).

mvdbeek:
and a rebase isn't required

bernt-matthias:
https://github.com/galaxyproject/tools-iuc/pull/2507 .. lets merge that .. then rebase this .. and hope for success :)

bernt-matthias:
hooray.. all green :)

PR #2115, opened 2018-10-01
fastqc: add some missing parameters

bernt-matthias:
I added --nogroup and --minlength which are really important if one wants to compare samples of different length (eg. before and after trimming) since otherwise FastQC might bin differently which makes the results difficult to compare (ie incomparable). 

Also added --kmer... 

Few other parameters are still missing.

FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help


lparsons:
@bernt-matthias I think you need something like:
```
    <assert_command>
        <has_text text="--kmers">
    </assert_command>
```

bernt-matthias:
Thanks for the tip. Should be running now.

Unfortunately FastQC still bins the Sequence Length Data .. just contacted the autor. But I guess we can still progress with the PR. 

Any tip for an alternative program?

mblue9:
@bernt-matthias maybe you could use [fastp](https://github.com/OpenGene/fastp) as it has a minimum `--length_required` and I don't think it groups bases into bins

bernt-matthias:
https://github.com/s-andrews/FastQC/issues/17

bernt-matthias:
Had some mail conversation with @s-andrews: new release will come soon. Lets wait with the PR until then.

bernt-matthias:
Should be it. At least the tool is finished. 

By separating the test data (1 per parameter) I have realized that many parameters seem to have no effect. 

- --kmer 3 definitely should show some over represented sequences
- -adapters file should add columns to the adapter section, but only fastqc default adapaters show up
- same for contamination (but this could be a problem of the test set)

bernt-matthias:
https://github.com/s-andrews/FastQC/issues/18

abretaud:
It looks good, do you prefer to wait for some fixes for s-andrews/FastQC#18 or can we merge now?

bernt-matthias:
Lets just merge for now since so far there was no response for https://github.com/s-andrews/FastQC/issues/18

abretaud:
Ok, thanks @bernt-matthias!

abretaud:
Upload failed, see #2141

abretaud:
Hum, it was [changed in this PR](https://github.com/galaxyproject/tools-iuc/pull/2115/files#diff-1d63b12823d13bf103a8af0c91fcbcc7L4) though it wasn't uploaded to TS, and then it should have been uploaded in #2141, isn't it right?

nsoranzo:
I mean that the tool requirement was updated, but not the wrapper version, so when an admin updates the tool, reproducibility is broken.

mvdbeek:
Yeah, that wasn't a good move, sorry :(

abretaud:
Ah ok, hadn't seen it... There's no way to remove a "broken" version from the TS?

mvdbeek:
Can only mark as malicious. That would be bad for reproducibility as well.

nsoranzo:
Should we have the TS block submission of a new repo revision changing requirements without modifying the tool version?

mvdbeek:
We wanted to do that for a long while already, right ?

nsoranzo:
@mvdbeek I think so, but I haven't found an issue at the galaxy repo with a cursory search.

erasche:
@mvdbeek @nsoranzo I proposed that way back years ago when I pushed for an alternative toolshed. That issue I got *SO* much pushback on, and was the sticking point that really made me give up on pushing for a new TS. People believed that it's too much of a burden on developers. But then things like this happen because of exactly that and it's much worse. And it's so messy from the galaxy update side. The example given is usually help text updates shouldn't be so cumbersome, but it ignores that help text updates can affect reproducibilty, if they forgot a 'no' in "do use this parameter for X". I strongly believe multiple tools with identical versions should be prohibited.

mvdbeek:
I think present-day reviewers/comitters think about this a little differently, I don't think that'd be a huge issue. Bumping deps without version bumps break reproducibility in a pretty major way after all.

erasche:
If we can get that changed, I'll be overjoyed.

PR #2124, opened 2018-10-08
Avoid blacklist/whitelist in user-facing stacks genotypes text

peterjc:
The terms 'whitelist' and 'blacklist' for inclusion or exclusion have undesirable associations, making it better to use more neutral language.

See also "Be careful in the words that we choose" in the main Galaxy Project Code of Conduct,
https://github.com/galaxyproject/galaxy/blob/dev/CODE_OF_CONDUCT.md and #2123 

Here we can't change the parameter names without breaking backward compatibility, so this is a cosmetic change only. The command line arguments -B and -W remain, but this reflects the underlying tool itself.

FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help

### DIFF
--- a/tools/stacks/stacks_genotypes.xml
+++ b/tools/stacks/stacks_genotypes.xml
@@ -1,4 +1,4 @@
-<tool id="stacks_genotypes" name="Stacks: genotypes" version="@WRAPPER_VERSION@.0">
+<tool id="stacks_genotypes" name="Stacks: genotypes" version="@WRAPPER_VERSION@.1">
     <description>analyse haplotypes or genotypes in a genetic cross ('genotypes' program)</description>
     <macros>
         <import>macros.xml</import>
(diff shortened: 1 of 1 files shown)


tools/stacks/stacks_genotypes.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="stacks_genotypes" name="Stacks: genotypes" version="@WRAPPER_VERSION@.0">
+<tool id="stacks_genotypes" name="Stacks: genotypes" version="@WRAPPER_VERSION@.1">
```
bgruening: You don't need to change the version if you just touch labels.

peterjc: OK, I'll revert that then...

bernt-matthias:
True, but it seems that `genotypes` is deprecated anyway, at least its not included in the recent stacks releases... 

Maybe we can join forces here: https://github.com/galaxyproject/tools-iuc/pull/2085 .. I hope that I can continue on this next week.

mvdbeek:
Yeah, this doesn't create a new version, the smaller the changes the better.

peterjc:
Thanks all.

PR #2127, opened 2018-10-10
gffcompare: fix for creating extra files in `database/files/`

bernt-matthias:
gffcompare creates extra files in the basedir of the input files. This leads to an error when jobs run as real user and is unwanted anyway.

For instance, for the input `database/files/001/295/dataset_1295486.dat` the files 

- `database/files/001/295/gffcmp.dataset_1295486.dat.tmap`
- `database/files/001/295/gffcmp.dataset_1295486.dat.tloc`

would be created. 

My solution is to create links, since this offers the possibility to add these files to an output collection. Alternatively with `-T` the creation could be forbidden. 

General question: would it be a good idea to link all input data to the working directory by default? 

FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help


tools/gffcompare/gffcompare.xml
```diff
@@ -10,33 +10,38 @@
...
-                    ln -s '${seq_data.seq_source.index.fields.path}' ref_seq.fa &&
-                #end if
-            #end if
-            gffcompare
-            ## Use annotation reference?
-            #if $annotation.use_ref_annotation == "Yes":
-                -r '$annotation.reference_annotation' $annotation.ignore_nonoverlapping_reference $annotation.ignore_nonoverlapping_transfrags
-            #end if
+    <command><![CDATA[
+#for $i, $gtf in enumerate( $inputs ):
```
nsoranzo: `#for $gtf in $inputs:`

tools/gffcompare/gffcompare.xml
```diff
@@ -10,33 +10,38 @@
...
-            $discard_single_exon -e $max_dist_exon -d $max_dist_group $discard_intron_redundant_transfrags
+$discard_single_exon
+-e $max_dist_exon
+-d $max_dist_group
+$discard_intron_redundant_transfrags
 
-            '$input_gtf'
-        ]]>
-    </command>
+#for $i, $gtf in enumerate( $inputs ):
```
nsoranzo: `#for $gtf in $inputs:`

bernt-matthias: good idea. 

any idea how we could name the files more usefully. maybe use the data set identifier. .. then we could create an output collection nicely...

nsoranzo: Which output collection?

bernt-matthias: There is none yet. The tmap and refmap files that are created by gffcompare are currently goint to nirvana. I was thinking about adding additional optional output collections for them .. But with the current loop the identifiers would be just 1,2,...

nsoranzo: Something like:
```
#set escaped_element_identifiers = [re.sub('[^\w\-]', '_', str(_.element_identifier)) for _ in $inputs]
#for $input, $escaped_element_identifier in zip($inputs, $escaped_element_identifiers):
    ln -s '$input' '$escaped_element_identifier' &&
#end for
...
#for $escaped_element_identifier in $escaped_element_identifiers:
    '$escaped_element_identifier'
#end for
```

How are the tmap and refmap files named?

bernt-matthias: Thanks for the suggestion I will give it a try tomorrow. 

The naming scheme is ´gffcmp.INFILENAME.[tmap|tloc]´. So this would become then ´gffcmp.$escaped_element_identifier.[tmap|tloc]´

tools/gffcompare/gffcompare.xml
```diff
@@ -10,33 +10,46 @@
...
-                    ln -s '${seq_data.seq_source.index.fields.path}' ref_seq.fa &&
-                #end if
-            #end if
-            gffcompare
-            ## Use annotation reference?
-            #if $annotation.use_ref_annotation == "Yes":
-                -r '$annotation.reference_annotation' $annotation.ignore_nonoverlapping_reference $annotation.ignore_nonoverlapping_transfrags
-            #end if
+    <command><![CDATA[
+echo $inputs &&
```
nsoranzo: Debugging statement?

bernt-matthias:
This should be nearly it: 
- also fixed wrong filter statements for the annotated and combined outputs

Tests are not working for me locally: 
```
Exception: Incorrect number of outputs - expected 6, found 4.
```
Exception: Incorrect number of outputs - expected 4, found 5.
```

When I try manually the number or outputs is correct. Also the tests without `expect_num_outputs` work.

tools/gffcompare/gffcompare.xml
bernt-matthias: Yep. Removed.

bernt-matthias:
Anyone has an idea why the number of outputs seems to be wrong. For me the planemo tests seem to be correct. Do I count wrong or is planemo?

mblue9:
@bernt-matthias the collection outputs don't get counted by planemo (don't know why) so the first test should be just 4 `expect_num_outputs="4"`.

Don't know why the second test is counting 5 as looks like should be 4 to me.

bernt-matthias:
Tests work now against this: https://github.com/galaxyproject/galaxy/pull/6931

Indeed galaxy counted wrong. 

Additionally the number of `gffinputs` is a list of HDAs in case of multiple inputs and a single HDA otherwise. Which is why the filters previously did not work.

bernt-matthias:
Something is odd with the testing. The errors are related to different tools. Any ideas. 

The tests of the tools should work now, since https://github.com/galaxyproject/galaxy/pull/6931 was just merged into dev.

nsoranzo:
> Something is odd with the testing. The errors are related to different tools. Any ideas.

@bernt-matthias See https://github.com/galaxyproject/tools-iuc/pull/2160

> The tests of the tools should work now, since galaxyproject/galaxy#6931 was just merged into dev.

We use `release_18.09` at the moment, not `dev`, see `.travis.yml`

nsoranzo:
@bernt-matthias The linting issue has been fixed. Now the first test fails with "Exception: Incorrect number of outputs - expected 6, found 4." because your fix is only in Galaxy `dev`, do you want to backport it to 18.09?

bernt-matthias:
@nsoranzo https://github.com/galaxyproject/galaxy/pull/6951

bernt-matthias:
hooray. all tests passed after  https://github.com/galaxyproject/galaxy/pull/6951 was merged.

PR #2133, opened 2018-10-12
Make certain tools conform to IUC best practices

wm75:
FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help

### DIFF
--- a/tools/sra-tools/fastq_dump.xml
+++ b/tools/sra-tools/fastq_dump.xml
@@ -12,7 +12,6 @@
     ## Need to set the home directory to the current working directory,
     ## else the tool tries to write to home/.ncbi and fails when used
     ## with a cluster manager.
-    (
     export HOME=\$PWD &&
     vdb-config --restore-defaults &&
     #if $input.input_select == "file":
--- a/tools/sra-tools/fastq_dump.xml
+++ b/tools/sra-tools/fastq_dump.xml
@@ -1,4 +1,4 @@
-<tool id="fastq_dump" name="Download and Extract Reads in FASTA/Q" version="@VERSION@.3">
+<tool id="fastq_dump" name="Download and Extract Reads in FASTA/Q" version="@VERSION@.4">
     <description>format from NCBI SRA</description>
     <macros>
         <import>sra_macros.xml</import>
--- a/tools/sra-tools/fastq_dump.xml
+++ b/tools/sra-tools/fastq_dump.xml
@@ -5,30 +5,14 @@
     </macros>
     <expand macro="requirements"/>
     <version_command>fastq-dump --version</version_command>
-    <command detect_errors="exit_code">
-        <![CDATA[
+    <command detect_errors="exit_code"><![CDATA[
 
-    #if $input.input_select=="file_list":
-    
-        for acc in `cat $input.file_list` ;
-        do
-    
-    #elif $input.input_select=="accession_number":
-
-        ## Stripping leading and trailing spaces in case user typed them in 
-        acc="${input.accession}" &&
-    
-    #end if
-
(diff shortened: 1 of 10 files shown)


martenson:
@wm75 please increase versions of all tools touched

bgruening:
@martenson I'm fine with these changes. But I think these are security enhancements and the version number bump can be avoided - maybe.

martenson:
@bgruening the version bump is mostly to prevent deployers from disregarding the update as minor. These updates are important.

bgruening:
But it leaves an old broken version. If we merge without a bump the current version is fixed, we can bump later as well and you get two fixed versions.

martenson:
@bgruening I am less concerned by that, since e.g. gemini has 4 installable revisions. However it is a fair approach (to update twice). I leave that to you.

mvdbeek:
I think it's important to bump, that way we can in the future block the execution of tools with known vulnerabilities.

mvdbeek:
Actually both is fine with me. Maybe your suggestion is better, causes less disruption.

wm75:
The failed check for the latest commit seems to be a timeout problem. The tests defined for the IQ-Tree tool involve computationally quite demanding construction of a phylogenetic tree, which takes quite some time on my local test machine (currently running for 15 min), but CI aborts after 20 min without response.
Seems worthwhile to simplify the test input file to speed things up in the future.

tools/sra-tools/fastq_dump.xml
```diff
@@ -1,34 +1,18 @@
...
-    
-        [ ""\$acc" =~ ^[E|S|D]RR[0-9]{1,}$" ] && (
-    
-    #end if
+    @SET_ACCESSIONS@
 
     ## Need to set the home directory to the current working directory,
     ## else the tool tries to write to home/.ncbi and fails when used
     ## with a cluster manager.
+    (
```
nsoranzo: I think having `(` here is going to break when `$input.input_select` is `'file'`, since this is not closed with `)` below in that case. We need a test case for this.

Similarly for `sam_dump.xml` .

nsoranzo:
@wm75 If it's not too complex, fixing it here would be nice. And more tests are also welcome :wink:

bgruening:
Can we fix SRA in a later PR @nsoranzo. We would like to get the patch in.

martenson:
If the PR does not break things that were working fine I urge a merge.

nsoranzo:
@bgruening You were arguing for not updating the version numbers, should these be reverted before merge?

nsoranzo:
It seems http://www.clustal.org is down (linked in https://github.com/galaxyproject/tools-iuc/blob/master/tools/iqtree/iqtree.xml#L695 ) which prevents the linting CI step to pass.

wm75:
clustal.org site seems to be up again

martenson:
all green!

nsoranzo:
I just noticed that gemini is not being tested (some planemo bug), I'll test locally and merge.

bgruening:
@nsoranzo gemini is on the black/skip list afaik. @wm75 is working on tests in a following PR.

nsoranzo:
Yeah, I also re-discovered it is in `.tt_skip`, not a Planemo bug at all (but should probably print the list of skipped repos).

martenson:
Given the version increases were removed, we need a subsequent PR that bumps the versions.

wm75:
I can do that. Nothing else required, right?

martenson:
@wm75 thank you! We should also update the suite_gemini repository.

wm75:
see #2146 for the version bumps

PR #2137, opened 2018-10-15
MACS2: Added option "--SPMR" to save signal per million reads

sveinugu:
--SPMR is needed to generate ChIP-seq signal tracks according to the MACS2 documentation:

https://github.com/taoliu/MACS/wiki/Build-Signal-Track

FOR CONTRIBUTOR:
* [x] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [x] - License permits unrestricted use (educational + commercial)
* [ ] - This PR adds a new tool or tool collection
* [x] - This PR updates an existing tool or tool collection
* [ ] - This PR does something else (explain below)

FOR REVIEWER:
* [ ] .shed.yml file ok
    - [ ] Toolshed user `iuc` has access to associated toolshed repo(s)
* [ ] Indentation is correct (4 spaces)
* [ ] Tool version/build ok
* [ ] `<command/>`
  - [ ] Text parameters, input and output files `'single quoted'`
  - [ ] Use of `<![CDATA[ ... ]]>` tags
  - [ ] Parameters of type `text` or having `optional="true"` attribute are checked with `if str($param)` before being used
* [ ] Data parameters have a `format` attribute containing datatypes recognised by Galaxy
* [ ] Tests
  - [ ] Parameters are reasonably covered
  - [ ] Test files are appropriate
* [ ] Help
  - [ ] Valid restructuredText and uses `<![CDATA[ ... ]]>` tags
* [ ] Complies with other best practice in [Best Practices Doc](http://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html)

PR #2139, opened 2018-10-15
JBrowse: Set any track config option + fix for math error

sveinugu:
Contains two changes (unsure whether it should have been two pull requests, but settled for one for convenience):

a) The code for calculating feature opacity logarithmically from score seemed wrong:

```javascript
var opacity = (score - ({min})) / (({max}) - ({min}));
opacity = Math.log10(opacity) + Math.log10({max});
```

Opacity should be a value from 0 to 1. Adding the log of max makes no sense, as this number will often be larger than 10, thus alwyas bringing the opacity to >1 indepentent of the actual score. Also if the score == max, the first term will be log10(1) == 0. Suggesting the following:

```javascript
var opacity = Math.log10(score - ({min})) / Math.log10(({max}) - ({min}));
```

b) The number of options for tailoring the tracks appearance and behaviour in JBrowse is too large to be conveniently wrapped in a single tool. I wanted to set two different settings that were not supported yet, 'displayMode' and 'style.strandArrow', and I am sure other users have other needs. Thus, I implemented a section for setting _any_ JBrowse track config option. Might be controversial, but very useful.

Do not know if I should add tests, as it is in any case very advanced usage, but can do if you think I should. Should I then add some settings to some existing tests or add completely new tests?

FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help


sveinugu:
Failing tests seems not to be relevant. Seems like there are several failing tests in he current build, experienced the same in https://github.com/galaxyproject/tools-iuc/pull/2137.

nsoranzo:
> Failing tests seems not to be relevant. Seems like there are several failing tests in the current build, experienced the same in #2137.

The test failures in the 2 PRs are unrelated. Unfortunately there are many moving parts (conda packages, tool versions, Galaxy releases, OS/library changes, ...) which may cause a previously working tool to fail a test, so there is a constant need to fix these.

Since we cannot merge a PR with failing tests (the updated repository will not be updated automatically), you may want to fix them yourself to speed up the process.

sveinugu:
@erasche @abretaud I have rebased the branch on top of the latest master and added tests for the custom track config. Also, I cleaned up the rest of the tests. From my point of view, everything is ready for review and merge.

abretaud:
@sveinugu sorry, I had missed your PR, I will have a look shortly!

tools/jbrowse/test-data/custom_config/test.xml
```diff
@@ -0,0 +1,127 @@
...
+                        <method>ignore</method>
+                        <scheme>
+                            <color>__auto__</color>
+                        </scheme>
+                </scaling>
+                <menus>
+                </menus>
+                <custom_config>
+                  <displayMode>
+                    <text>
```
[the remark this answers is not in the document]
sveinugu: @erasche Sorry, did not understand which whitespace you meant.. There was already a single blank line after the menu tag, which I kept, but perhaps I should go through and remove all blank lines throughout?

tools/jbrowse/jbrowse.xml
```diff
@@ -210,6 +210,17 @@ $trackxml &&
                   #end for
                 </menus>
             #end if
+            #if str($track.data_format.data_format_select) in ["gene_calls", "pileup", "blast", "wiggle", "vcf"]:
+                <custom_config>
+                #for $custom_option in $track.data_format.jb_custom_config.option:
+                  <${custom_option.opt_key}>
+                    <${custom_option.opt_value.val_type}>
```
erasche: Specifically whitespace within this tag, I think should be removed.

tools/jbrowse/test-data/custom_config/test.xml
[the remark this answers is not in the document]
erasche: @sveinugu sorry should have linked directly to source https://github.com/galaxyproject/tools-iuc/pull/2139/files#r273446021

tools/jbrowse/jbrowse.xml
sveinugu: Do you mean e.g.:
```
<style.strandArrow><boolean>false</boolean></style.strandArrow>
```

erasche: Doesn't need to be for the opt_key, but for opt_value.val_type, in case the type is a string, and the contents have the newlines and whitespace unexpectedly for the user

abretaud: Actually, the generated xml is invalid when you put a whitespace in opt_key, which makes jbrowse.py crash. I see no good reason to write a space there, but maybe adding a regex validator would be better than a big red error in the history?

Dangerous characters are sanitized in the generated xml (`'`=>`__sq__`, `"`=>`__dq__`, `<`=>`__lt__` etc), so I think this should be safe. opt_value [is unsanitized](https://github.com/galaxyproject/tools-iuc/pull/2139/files#diff-1aa143c7e46d808f79ed389dd7a6726cR697) before being written to trackList.json, but quotes are escaped, so should be safe too

sveinugu: I have now added a regex validator, as well as removed the whitespace within opt_value.val_type tag.

PR #2141, opened 2018-10-16
Fastqc test fix

abretaud:
FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help

For some reason (recent change in one of the conda package?) the tests failed after merging #2115, which means the tool was not pushed t the toolshed.
This PR should fix it


tools/fastqc/rgFastQC.xml
```diff
@@ -103,18 +103,18 @@
             <param name="kmers" value="3" />
             <output name="html_file" file="fastqc_report_kmer.html" ftype="html"/>
             <output name="text_file" file="fastqc_data_kmer.txt" ftype="txt"/>
-	    <assert_command>
-		    <has_text text="--kmers 3"/>
-	    </assert_command>
+    	    <assert_command>
+    		    <has_text text="--kmers 3"/>
+    	    </assert_command>
```
nsoranzo: Indentation is off here.

tools/fastqc/rgFastQC.xml
```diff
@@ -103,18 +103,18 @@
...
             <param name="input_file" value="1000trimmed.fastq" ftype="fastq" />
             <param name="nogroup" value="--nogroup" />
             <output name="html_file" file="fastqc_report_nogroup.html" ftype="html"/>
             <output name="text_file" file="fastqc_data_nogroup.txt" ftype="txt"/>
-	    <assert_command>
-		    <has_text text="--nogroup"/>
-	    </assert_command>
+    	    <assert_command>
+    		    <has_text text="--nogroup"/>
+    	    </assert_command>
```
nsoranzo: Indentation is off here.

bernt-matthias:
odd. did remember that the tests were successful.

abretaud:
> odd. did remember that the tests were successful.

Yes, the PR was green when I merged it, but maybe there was a change in a conda package: when I compared the html result, there were some (very minor) changes in the police used for the images.

bernt-matthias:
Looks like the html/image data differs. Would suggest to test only for the presence of the html files / allow a large diff.

abretaud:
Yeah, comparing the size should be enough I guess, there's already a check on the txt output

bernt-matthias:
Problem is: the html is on a single line which contains the date (fun fact: upload on the same date would have worked). 

So it should be sufficient to check for sim_size. I could do the changes (I have one more edit). Do you prefer that I open a PR against your fork or would you allow be to push to your branch?

abretaud:
Can you try to push to my branch? If you have write access to the iuc repo, you should have the right to push to any PR

bernt-matthias:
worked. 

adapter and kmer test now have an effect (ie produce a different result than the parameter less test)

still not completely satisfied that the contaminants parameter has no effect in the tests... see also 
https://github.com/s-andrews/FastQC/issues/18

bernt-matthias:
I'm happy now (all tests make a difference in the output).

tools/fastqc/rgFastQC.xml
```diff
@@ -3,7 +3,15 @@
     <requirements>
         <requirement type="package" version="0.11.8">fastqc</requirement>
     </requirements>
-    <command detect_errors="exit_code"><![CDATA[
+    <stdio>
+        <exit_code range="1:" level="fatal" description="FastQC returned non zero exit code" />
+        <exit_code range=":-1" level="fatal" description="FastQC returned non zero exit code" />
+    <regex match="There is insufficient memory for the Java Runtime Environment"
```
mvdbeek: This doesn't look great to me, a OOM kill should not quit with 0, and should also not write to stdout. Any chance the job runnner is doing something weird here ?

bernt-matthias: Problem is that `java` (as most programs) does not (sufficiently) differentiate errors by error code. 

Just checked a call to `java` with limited memory which gives 

```
Error occurred during initialization of VM
Could not allocate metaspace: 1073741824 bytes
```

on **stdout** and `echo $?` gives 1. 

(I really start to dislike JAVA for more than one reason.. to put it mildly :) )

Wrt. this problem I started with @jmchilton to change the behavior of Galaxy to check for oom errors before exit codes: https://github.com/galaxyproject/galaxy/pull/6685 (follow up to https://github.com/galaxyproject/galaxy/pull/6338).

This should make this stdio tag useful (you are right it isn't at the moment).

However, the zoo of possible oom error messages that java spits that would need to be treated by oom checks for each java based tool bugs me quite a bit. Maybe it would be an idea to add     `java_errors` to the `detect_errors` attribute of the `<command>` tag...? Btw. do you know if regexes defined in the `<stdio>` tag are considered if `detect_errors` is set to `exit_code` or `aggressive`?

bernt-matthias: SIGKILL, yes (but this could then be also a violation of the maximum run time). 

Some schedulers might set the state properly, but eg. UNIVA (and probably SGE as well) don't do a great job in helping to detect OOM errors, i.e. they don't help at all :( thus also the runners don't set an oom state...

mvdbeek:
Thanks @abretaud and @bernt-matthias !

PR #2145, opened 2018-10-17
Trinity: fix option name

### LINKED ISSUE #2143: trinity bug
hello
Trinity galaxy tool:
This affects all the version of Trinity 2.4.0 including the latest one 21 / ca0588b09438

Previously Trinity only did read normalisation if the --normalize_reads option was provided. This is no longer the case.

Trinity 2.4.0 does read normalisation by default. 

It uses the optional flag --no_normalize_reads to NOT normalise the reads. The galaxy tools use an older option that expected --normalize_reads to normalise the reads. 

This means that all galaxy runs with users that set normalisation to FALSE, always got their data normalised anyway ...

So the XML needs to be changed with  --no_normalize_reads and the true/false option switched around...

[sorry i can't do a diff patch right now...]
a

abretaud:
FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help

A little patch to fix #2143 (thanks for reporting @alpapan!)

### DIFF
--- a/tools/trinity/macros.xml
+++ b/tools/trinity/macros.xml
@@ -7,7 +7,7 @@
         </requirements>
     </xml>
 
-    <token name="@WRAPPER_VERSION@">2.8.3</token>
+    <token name="@WRAPPER_VERSION@">2.8.4</token>
 
     <token name="@COMMAND_PAIRED_STRAND_JACCARD@">
         #if $pool.inputs.strand.is_strand_specific:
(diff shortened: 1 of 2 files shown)


nsoranzo:
@abretaud I think that we need to update the `trinity` requirement to 2.8.4, whose conda package fix the missing `numpy` dependency https://github.com/bioconda/bioconda-recipes/pull/10909

PR #2149, opened 2018-10-18
Maker: update 2.31.10 + add support for mpi

abretaud:
FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help

A minor update for maker, and also added support for MPI (finally found a way to avoid systematic segfault).
I'm a little nervous with the MPI stuff, but tests are ok (= in a non-cluster environment) and it seems to work fine on our cluster with a properly configure job destination.

### DIFF
--- a/tools/maker/macros.xml
+++ b/tools/maker/macros.xml
@@ -6,7 +6,7 @@
             <yield />
         </requirements>
     </xml>
-    <token name="@VERSION@">2.31.9</token>
+    <token name="@VERSION@">2.31.10</token>
 
     <xml name="citations">
         <citations>
--- a/tools/maker/maker.xml
+++ b/tools/maker/maker.xml
@@ -35,7 +35,7 @@
             export AUGUSTUS_CONFIG_PATH=`pwd`/augustus_dir/ &&
         #end if
 
-        mpiexec -n \${GALAXY_SLOTS:-4} maker maker_opts.ctl maker_bopts.ctl maker_exe.ctl
+        mpiexec -n \${GALAXY_SLOTS:-4} maker maker_opts.ctl maker_bopts.ctl maker_exe.ctl < /dev/null
 
         &&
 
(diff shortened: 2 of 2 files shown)


abretaud:
(There's also this PR which should improve support for mpi on SGE cluster: https://github.com/conda-forge/openmpi-feedstock/pull/30)

abretaud:
humpf. maybe mpi is not so simple. we should wait before merging

abretaud:
ping @bgruening as we discussed: using maker in mpi mode would probably look like this
(but this pr is not yet ready I think)

bgruening:
Thanks for sharing @abretaud! Have a save trip home!

abretaud:
Hum, the error looks like https://github.com/bioconda/bioconda-recipes/issues/13562, I'll have a look next week

abretaud:
Green!
I think it's ready to merge now. To sum up, it uses mpich now to run maker in mpi mode:
- if it runs on a destination which supports mpi, fantastic, it can use thousands of cpus on hundreds of compute nodes. e.g. running this on a slurm queue with option `-n`
- if the destination doesn't support mpi, it works too, it will be able to use GALAXY_SLOTS number of cpus on the compute node (or even a single one if there is only one available)

ping @erasche @shiltemann with this PR the maker tutorial should run much faster, so it could be useful for the Utrecht session. Do you think it would be possible to try it, or do you prefer to keep the current non-mpi version?

I can open a Galaxy issue to start a discussion on a better way to standardize mpi execution

erasche:
@abretaud of course we want a faster version! :)

abretaud:
@erasche cool!
I opened the galaxy issue: https://github.com/galaxyproject/galaxy/issues/7884

bgruening:
Thanks for creating the issue and the update.

PR #2157, opened 2018-10-22
New option to annotate outputs of deseq2 and dexseq tools and new test cases for dexseq.

pavanvidem:
FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help


mvdbeek:
Would you be willing to implement this as a standalone tool ?
Both tools are already very complex, and I think it might be easier for users to have a separate tool.

pavanvidem:
I discussed it with @bgruening and agreed on integrating into the tools as an option. Implementing 
 a standalone tool is also fairly simple. In the end, it is all about the usability. I don't know what others will say about these two options. Maybe others want to comment about it?

mvdbeek:
It isn't just about usability, it's also about maintainability, additional dependencies and "doing one thing, and doing it well".

The fact that you're going to modify 2 tools instead of having a separate tool to do the job is a pretty good sign that this should be a separate step. Why stop here and not do this for limma, edgeR and co ? What if you've done the analysis already, or a colleague send you a result table or you downloaded them from the supplemental materials of an article, then you need to run deseq/dexseq again ?

There's also the issue that you can do this with a regex filter and bedtools today, but I agree that you'll have to be an advanced user, I can see this being useful as a separate tool.

This sounds a bit negative, but I care about this a lot and I totally appreciate your work.

pavanvidem:
I agree with the point that we cannot annotate already existing result tables. I am totally aware that it is a 1 liner in bash, and that is how I normally do. But I am not sure without writing a proper parser we can extract information from any GTF/GFF files. If you already have such a generic script, please let me know. We will build a tool/workflow out of that. Otherwise, I am glad to write an independent tool out of my work.

mblue9:
>I can see this being useful as a separate tool

I agree! I use the annotatemyids tool at the moment to annotate e.g. limma output but that only has info for 4 species (human, mouse, fly, zebrafish). You can also only get the info provided by org.db packages and in cases where you're working with different GTFs you may instead want to get the info from those GTFs rather than org.db. I had also been thinking of writing a tool to extract info from a GTF (just no time for it at the moment) as I'm working with some splicing tools where you provide a GTF and you don't always get the gene name output. So I think that a generic tool where you can extract e.g. a gene name from a GTF for a table of gene/transcript ids would be generally useful.

pavanvidem:
as per @mvdbeek  and @mblue9  suggestions, created a standalone version here:
https://github.com/galaxyproject/tools-iuc/pull/2186

PR #2121, opened 2018-10-05
Scanpy

bebatut:
FOR CONTRIBUTOR:
* [X] I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [X] License permits unrestricted use (educational + commercial)
* [X] This PR adds a new tool or tool collection
* [ ] This PR updates an existing tool or tool collection
* [ ] This PR does something else (explain below)

FOR REVIEWER:
* [ ] .shed.yml file ok
    - [ ] Toolshed user `iuc` has access to associated toolshed repo(s)
* [ ] Indentation is correct (4 spaces)
* [ ] Tool version/build ok
* [ ] `<command/>`
  - [ ] Text parameters, input and output files `'single quoted'`
  - [ ] Use of `<![CDATA[ ... ]]>` tags
  - [ ] Parameters of type `text` or having `optional="true"` attribute are checked with `if str($param)` before being used
* [ ] Data parameters have a `format` attribute containing datatypes recognised by Galaxy
* [ ] Tests
  - [ ] Parameters are reasonably covered
  - [ ] Test files are appropriate
* [ ] Help
  - [ ] Valid restructuredText and uses `<![CDATA[ ... ]]>` tags
* [ ] Complies with other best practice in [Best Practices Doc](http://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html)

---
The skeletons of the wrappers have been automatically generated using the script in `src`.

We then group them by themes. They now need to be cleaned, tested with addition of tests:
1. Filter
   - [x] `pp.filter_cells`
   - [x] `pp.filter_genes`
   - [x] `pp.filter_genes_dispersion`
   - [ ] `pp.highly_variable_genes`
   - [x] `pp.subsample`
   - [ ] `queries.gene_coordinates`
   - [ ] `queries.mitochondrial_genes`
3. Normalize
   - [x] `pp.normalize_per_cell`
   - [x] `pp.recipe_zheng17`
   - [x] `pp.recipe_weinreb17`
   - [x] `pp.recipe_seurat`
   - [x] `pp.log1p`
   - [x] `pp.scale`
   - [x] `pp.sqrt`
   - [x] `pp.downsample_counts`
4. Remove confounder
   - [x] `pp.regress_out`
   - [ ] `pp.mnn_correct`
   - [ ] `pp.dca`
   - [ ] `pp.magic`
[...]

PR #2159, opened 2018-10-29
First implementation of galaxy wrappers for scanpy-scripts.

nh3:
FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help


nh3:
@pcm32 Strangely, linter is failing me for wrappers of other tools that has nothing to do with this PR>

nsoranzo:
@nh3 Yep, that's not your fault. flake8 (the tool we use to lint our Python code) has a new release, which now has some extra checks. I'll have a look in a bit.

shiltemann:
Hi @nh3, thanks for your contribution! I see that @bebatut is also working on scanpy (see https://github.com/galaxyproject/tools-iuc/pull/2121), perhaps you could merge your work together?

pcm32:
Hi @shiltemann we have discussed this with @bgruening and @bebatut but we are starting from different assumptions. We want to be able to easily reuse cli calls from other workflow environments and for that we needed an intermediate bioconda package. We are also interested in certain level of granularity that is different to the PR started by @bebatut. I hope that having both sets of tools is not a problem. I presume that with good documentation it will be clear what are the differences and complementarity of them.

nsoranzo:
@nh3 Can you add the `.shed.yml` file for your suite of tools? See https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/shed_yml.html

nh3:
Thanks @nsoranzo . I've added `.shed.yml`. Tests failed because of missing `test-data/`. My test data are pretty large (~180MB). Should I add and push them here?

shiltemann:
Hi @nh3 :) test data needs to be included if you want to contribute it here, and it should really be as small as possible so please see if you could reduce it (a lot) first. 

One solution could be to combine with @bebatut's effort, but have different subfolders (like we have for [qiime](https://github.com/galaxyproject/tools-iuc/tree/master/tools/qiime)); for example a `scanpy_core` and `scanpy_scripts` folder? That way you can still be a bit separate, but easily re-use each other's test-data and collaborate on a shared macros file because I imagine there would still be a lot of overlap in these things even if you are going for different levels of granularity? What do you think? 

There appears to be a large collection of small test datasets in the other PR, which could be useful to you possibly?

nh3:
Hi @shiltemann , thanks for the advice. I'll definitely look into the options suggested.

nh3:
Hi @shiltemann , thank you very much for the detailed code review! Is there a way of controlling the order in which individuals tools are displayed on the side bar? Thanks.

nh3:
> Hi @nh3, thanks again for contributing, yes, if you can make the test-data folder approx 10 times smaller that would be great (not a strict cutoff or anything, just ballpark, but the smaller you can get it the better)

I've managed to reduce the size of test-data to under 8M, so should be fine for now I guess. I push a commit once the code style/quality-related edits are done. Thanks.

nsoranzo:
> Hi @shiltemann , thank you very much for the detailed code review! Is there a way of controlling the order in which individuals tools are displayed on the side bar? Thanks.

@nh3 The Galaxy admin can change that by editing `integrated_tool_panel.xml` after the tool installation, but nothing we can control in this repository.

tools/scanpy_scripts/.shed.yml
```diff
@@ -0,0 +1,15 @@
...
+    It includes preprocessing, visualization, clustering, pseudotime and
+    trajectory inference and differential expression testing. Scanpy-scripts
+    provides a set of command-line wrapper scripts that are composed from Scanpy
+    API for easy execution of common analysis.
+homepage_url: https://scanpy.readthedocs.io
+remote_repository_url: https://github.com/ebi-gene-expression-group/scanpy-scripts
+type: unrestricted
+categories:
+- Transcriptomics
+- Sequence Analysis
```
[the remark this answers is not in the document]
nh3: Hi @shiltemann, thank you for the review and suggestion. When making the CLI scripts for scanpy, I tried to align the CLI options with library function options where possible, but they are not always the same, as we also want scanpy CLI scripts to have somewhat similar interface as other scRNAseq CLI scripts we developing. Therefore, the sharing of code snippets is not that straightforward. With that said, I am open to the idea, as long as development/maintenance can be separated.

nh3:
Hi @shiltemann, I've made modifications and uploaded test-data. Hopefully, it has improved over the previous version.

shiltemann:
@nh3 thanks! I will have a closer look this week, but looking good :+1:

tools/scanpy_scripts/.shed.yml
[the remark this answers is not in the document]
shiltemann: hi @nh3 oh sure, any code or test file sharing would still be optional, just a bit easier to do when you do find overlaps at any point :)

tools/scanpy_scripts/scanpy-filter-cells.xml
```diff
@@ -0,0 +1,69 @@
...
+        -s '${subset}'
+    #end if
+    ]]></command>
+
+    <inputs>
+        <expand macro="input_object_params"/>
+        <expand macro="output_object_params"/>
+        <repeat name="parameters" title="Parameters used to filter cells" min="1">
+            <param name="name" type="text" value="n_genes" label="Name of parameter to filter on">
+                <option value="n_genes">n_genes</option>
```
shiltemann: did you mean for this to be a `select` parameter?

nh3:
Thank you so much @shiltemann for such detailed review and correction!

tools/scanpy_scripts/scanpy-filter-cells.xml
nh3: This is meant to be a text parameter where arbitrary field name can be entered, where n_genes and n_counts are commonly used options.

shiltemann: ok, these options are not visible anywhere in the interface if it is a text parameter, so might be better to add it as help text in that case, something like `help="for example n_genes or n_counts"` as a hint to users

shiltemann: thanks for the update @nh3, please also remove the `options` here since they do not do anything if the param type isn't a `select` :)

nh3: Sure. I was just following [this](https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-inputs-param) which seems to fit the intended usage scenario:

> As of 17.01, text parameters can also supply a static list of preset defaults options. The user may be presented with the option to select one of these but will be allowed to supply an arbitrary text value.
> ```
> <param name="foo" type="text" value="foo 1">
>     <option value="foo 1">Foo 1 Display</option>
>     <option value="foo 2">Foo 2 Display</option>
> </param>
> 
> ```

Users who are not that familiar with `scanpy` may want to filter on number of genes/cells and read counts but don't know the exact name of the property, e.g. `seurat` uses nGenes, so those options were meant to provide hints and/or convenience.

nh3: Yes, it works locally for me. The options popped up once I started typing in the input.

shiltemann: ok great :+1:

tools/scanpy/scanpy_scripts/scanpy-run-tsne.xml
```diff
@@ -0,0 +1,111 @@
...
+                    <option value="X">X, use normalised expression values</option>
+                    <option value="auto" selected="true">Automatically chosen based on problem size</option>
+                </param>
+                <param name="perplexity" argument="--perplexity" type="float" value="30" label="The perplexity is related to the number of nearest neighbours, select a value between 5 and 50"/>
+                <param name="early_exag" argument="--early-exaggeration" type="float" value="12" label="Controls the tightness within and between clusters"/>
+                <param name="learning_rate" argument="--learning-rate" type="float" value="1000" label="Learning rate, should be between 100 and 1000"/>
+                <param name="fast_tsne" type="boolean" checked="true" label="Use multicoreTSNE"/>
+                <param name="n_job" argument="--n-jobs" type="integer" optional="true" label="The number of jobs"/>
+                <param name="n_pc" argument="--n-pcs" type="integer" optional="true" label="The number of PCs to use"/>
+                <param name="seed" argument="--random-seed" type="integer" value="0" label="Seed for random number generator"/>
```
shiltemann: you do a check of `not None` on this parameter in the command section, did you want to make it optional="true"? (I assume a random seed is used if nothing provided, and that might be the best default behaviour?) and same Q for the other tools that use the seed parameter

tools/scanpy/scanpy_scripts/scanpy-run-umap.xml
```diff
@@ -0,0 +1,115 @@
...
+                <param name="neg_sample_rate" argument="--negative-sample-rate" type="integer" value="5" label="The ratio of negative to positive edge in optimisation"/>
+                <param name="init_pos" argument="--init-pos" type="text" label="Method to initialise embedding, any key for adata.obsm or choose from the preset methods">
+                    <option value="spectral" selected="true">spectral</option>
+                    <option value="paga">paga</option>
+                    <option value="random">random</option>
+                </param>
+                <param name="maxiter" argument="--maxiter" type="integer" optional="true" label="Number of iterations of optimisation"/>
+                <param name="A" argument="-a" type="float" optional="true" label="More specific parameter controlling embedding, automatically determined from --min-dist and --spread if unset"/>
+                <param name="B" argument="-b" type="float" optional="true" label="More specific parameter controlling embedding, automatically determined from --min-dist and --spread if unset"/>
```
shiltemann: can we get a bit more info on these parameters, perhaps in a help attribute? the labels for A and B are identical for both now and a tad vague.

tools/scanpy/scanpy_scripts/scanpy-find-cluster.xml
```diff
@@ -0,0 +1,92 @@
...
+
+    <help><![CDATA[
+**What it does**
+Cluster cells into subgroups.
+
+Cluster cells using the Louvain algorithm of Blondel et al. in the
+implementation of Traag et al. The Louvain algorithm has been proposed
+for single-cell analysis by Levine et al.
+
+@HELP@
```
shiltemann: I added a bit more per-tool info to the help sections of some of the wrappers (and expanded the suite-level help in the macro) perhaps you could do it for the remaining wrappers? just a few sentences on what the tool does (on a non-technical level), with the extra citations where needed. (and please check that what I added is correct, I just added what I could find in the docs, but you may have more/better descriptions)

tools/scanpy/scanpy_scripts/scanpy-run-tsne.xml
```diff
@@ -0,0 +1,111 @@
...
+                <param name="use_rep" argument="--use-rep" type="select" label="Use the indicated representation">
+                    <option value="X_pca">X_pca, use PCs</option>
+                    <option value="X">X, use normalised expression values</option>
+                    <option value="auto" selected="true">Automatically chosen based on problem size</option>
+                </param>
+                <param name="perplexity" argument="--perplexity" type="float" value="30" label="The perplexity is related to the number of nearest neighbours, select a value between 5 and 50"/>
+                <param name="early_exag" argument="--early-exaggeration" type="float" value="12" label="Controls the tightness within and between clusters"/>
+                <param name="learning_rate" argument="--learning-rate" type="float" value="1000" label="Learning rate, should be between 100 and 1000"/>
+                <param name="fast_tsne" type="boolean" checked="true" label="Use multicoreTSNE"/>
+                <param name="n_job" argument="--n-jobs" type="integer" optional="true" label="The number of jobs"/>
```
shiltemann: we can probably set a `min="0"` on some of these numerical parameters yeah?

tools/scanpy/scanpy_scripts/scanpy-run-tsne.xml
```diff
@@ -0,0 +1,111 @@
...
+            <when value="false">
+                <param name="use_rep" argument="--use-rep" type="select" label="Use the indicated representation">
+                    <option value="X_pca">X_pca, use PCs</option>
+                    <option value="X">X, use normalised expression values</option>
+                    <option value="auto" selected="true">Automatically chosen based on problem size</option>
+                </param>
+                <param name="perplexity" argument="--perplexity" type="float" value="30" label="The perplexity is related to the number of nearest neighbours, select a value between 5 and 50"/>
+                <param name="early_exag" argument="--early-exaggeration" type="float" value="12" label="Controls the tightness within and between clusters"/>
+                <param name="learning_rate" argument="--learning-rate" type="float" value="1000" label="Learning rate, should be between 100 and 1000"/>
+                <param name="fast_tsne" type="boolean" checked="true" label="Use multicoreTSNE"/>
```
shiltemann: here you can use `truevalue="" falsevalue="--no-fast-tsne"` and replace the `if` in the command section with a simple `$settings.fast_tsne`

tools/scanpy/scanpy_scripts/scanpy-run-tsne.xml
nh3: Thank you for the suggestion. The intention here is to have a default seed value (0), so that runs are reproducible by default. If one really wants to achieve something "more random", then  one needs to explicitly remove the default value. So perhaps having `value="0"` and `optional="true"` at the same time?

tools/scanpy/scanpy_scripts/scanpy-run-umap.xml
nh3: Yep, I agree it sounds vague. Unfortunately, that's how these options are explained in `scanpy`'s [API documentation](https://scanpy.readthedocs.io/en/latest/api/scanpy.api.tl.umap.html#scanpy.api.tl.umap)

tools/scanpy/scanpy_scripts/scanpy-find-cluster.xml
nh3: Yep, makes a lot of sense. I'll add them.

tools/scanpy/scanpy_scripts/scanpy-run-tsne.xml
nh3: That's actually something I wanted to ask. The galaxy best practice recommends making use of `truevalue=` and `falsevalue=` to have simpler command section. However, it also recommends putting single quotes around variables in command for security reasons. If both are followed, then when `fast_tsne` is set to true, the command line will have a `''` in it, which argparse will treat as a positional argument being an empty string which usually leads to error in command line parsing. What would be the right way to handle boolean parameters?

tools/scanpy/scanpy_scripts/scanpy-run-tsne.xml
shiltemann: sure, you can have both :)

tools/scanpy/scanpy_scripts/scanpy-run-tsne.xml
shiltemann: you would not need to put this in quotes in the command section since it will only be one of two known values, not arbitrary user input (we only require single quotes for file paths and text parameters) So I think you can use truevalue and falsevalue here :)

tools/scanpy/scanpy_scripts/scanpy-run-umap.xml
shiltemann: ok, thats unfortunate but not much you can do, if you ever do find out, just add it :)

bernt-matthias:
How does this relate to https://github.com/galaxyproject/tools-iuc/blob/master/tools/scanpy? Is this PR still needed?

pcm32:
You can close this. We ended up managing our scanpy tool set independently on our repos. It is independent of the iuc scanpy tools.

bernt-matthias:
Thanks @pcm32

PR #2162, opened 2018-10-30
Cutadapt : Addition of a multiple output option to split file depending on barcode

Delphine-L:
FOR CONTRIBUTOR:

Addition of a multiple output option in cutadapt. Instead of creating one file with all trimmed reads in one file, it created a collection of file divided by adapter.

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help


tools/cutadapt/cutadapt.xml
```diff
@@ -218,6 +226,7 @@ $read_mod_options.trim_n
             <param name="too_short_file" argument="--too-short-output" type="boolean" value="False" label="Too Short Reads" help="Write reads that are too short (according to minimum length specified) to a file. (default: discard reads)"/>
             <param name="too_long_file" argument="--too-long-output" type="boolean" value="False" label="Too Long Reads" help="Write reads that are too long (according to maximum length specified) to a file. (default: discard reads)"/>
             <param name="untrimmed_file" argument="--untrimmed-output" type="boolean" value="False" label="Untrimmed Reads" help="Write reads that do not contain the adapter to a separate file, instead of writing them to the regular output file.  (default: output to same file as trimmed)"/>
+            <param name="multiple_output" argument="" type="boolean" value="False" label="Multiple output" help="Create a separate file for each adapter trimmed  (default: all trimmed read are in a single file)"/>
```
bgruening: `all trimmed read__s__` ?

bernt-matthias:
Thanks for the contribution to this important tool (IMHO the only tool that is (for multiple reasons) worth considering for adapter trimming and quality filtering .. and many other things). I have a few comments (please tell me if you need help with this).

1. It seems that the adapter sequences need to be specified with `-a`/`-g`/'-b' (and `-A`/`-G`/`-B` for paired end data). Since these options seem to be missing from the tool I'm wondering what happens when you enable splitting. Furthermore it seems that you need named adapters to use `{name}` in the output option? 

2. I would really like to see a test for splitting.

3. `--suffix` can also make use of `{name}` 

4. I would suggest to move  `--no-trim` and `--mask-adapter` to the adapter section (makes more sense and would be consistent to the cutadapt help output)

5. would be great if you could add `--no-match-adapter-wildcards` then cutadapts adapter options would be (I hope) complete. 

I think 3 to 5 are optional. 

Here some links to the docs

https://cutadapt.readthedocs.io/en/stable/guide.html#multiple-adapters
https://cutadapt.readthedocs.io/en/stable/guide.html#demultiplexing
https://cutadapt.readthedocs.io/en/stable/guide.html#linked-adapters

It seems that there are some possibilities for the adapter parameters

- just adapter sequence `-a sequence` 
- multiple adapters `-a sequence1 -a sequence2....`
- named adapters `-a name1=sequence1 ... `
- adapters from file `-a file:filename`

Sequences might have anchors `^`/`$`. 

Something along the lines: 

```
<conditional>
   <param name="file_or_separate">
   <when file>
       <param type="data">
   </when>
   <when separate>
     <repeat>
         <param type="text" name="adapter_name" optional="true"/>
         <param type="text" name="adapter">
             ideally here would be a validator
         </param>
     </repeat>
   </when>
</conditional>
```

using a macro for the form and a token for the command should allow to treat all three options.

davebx:
@bernt-matthias 

1. You'll be happy to know that the -aAbBgG flags are covered in the `macros.xml` and implemented with the `@read1_options@` and `@read2_options@` tokens
2. We are in fact working on a test for that at the moment.

bgruening:
@bernt-matthias ready if you are ready :)

@Delphine-L thanks a bunch!

Delphine-L:
Thank you !

PR #2167, opened 2018-11-06
Use new group_tag parameter to allow selecting collection subsets

mvdbeek:
The idea here is that the user tags elements of a collection using tags that
start with `group:`. All of these tags can then be selected in the tool
form when choosing the factor levels.
This makes it very easy to select dataset combinations of interest
without having to re-shape collection contents for every contrast.

Technically the input does not need to be a collection, a
multiple="true" data input also works, but is currently not testable.
We might change this in the future.

The previous way of doing the analysis is still the default way of
using deseq2, as we currently don't auto-propagate group tags.
So in order to use this new feature users either have to manually tag collection
elements or use the `Tag elements from file from contents of a file`
to set the group tags. Hopefully we will enable auto-propagation
in 19.01 and enhance the apply rules tool to manipulate group tags.

FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help

### DIFF
--- a/tools/deseq2/deseq2.xml
+++ b/tools/deseq2/deseq2.xml
@@ -1,5 +1,8 @@
-<tool id="deseq2" name="DESeq2" version="2.11.40.5">
+<tool id="deseq2" name="DESeq2" version="2.11.40.6">
     <description>Determines differentially expressed features from count tables</description>
+    <macros>
+        <import>deseq2_macros.xml</import>
+    </macros>
     <requirements>
         <requirement type="package" version="1.20.0">bioconductor-deseq2</requirement>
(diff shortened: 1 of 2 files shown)


mvdbeek:
I guess we need to update the xsd schema for the tags attribute ...

mvdbeek:
Yay, all green with the new planemo release!

davebx:
I agree with Björn.

mvdbeek:
Corresponding mini-tuotiral here: https://github.com/galaxyproject/training-material/pull/1244

bgruening:
Merci @mvdbeek!

PR #2192, opened 2018-11-29
kallisto quant: add strand specificity options and test

pavanvidem:
this is an important missing parameter


bgruening:
Thanks Pavan!

PR #2205, opened 2018-12-07
Add mash_sketch tool

dfornika:
FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help


dfornika:
Currently not passing `planemo test`. When galaxy picks up the `RefSeqSketches.msh` file it renames it to have a `.dat` filetype. `mash screen` doesn't like that (it checks for the `.msh` extension`).

I'm not sure if galaxy needs a new datatype for mash sketches in order to handle that correctly(?)

I'm not sure if it's ok to include the ~100MB `RefSeqSketches.msh` file as test data.

tools/mash/mash_screen.xml
```diff
@@ -0,0 +1,82 @@
+<tool id="mash_screen" name="mash screen" version="2.1-0">
+    <description>Determine how well query sequences are contained within a pool of sequences.</description>
+    <requirements>
+        <requirement type="package" version="2.1">mash</requirement>
+    </requirements>
+    <version_command>mash --version</version_command>
+    <command detect_errors="exit_code"><![CDATA[
+        mash screen
```
bgruening: You can add a command like `        ln -s '$queries' input.msh &&` to link your mesh file to a file with proper file ending.

tools/mash/mash_screen.xml
```diff
@@ -0,0 +1,82 @@
...
+	     $queries
+	     $pool
+	     > $output
+    ]]></command>
+    <inputs>
+        <param type="data" name="queries" format="binary" />
+        <param type="data_collection" name="pool" format="fastq" />
```
bgruening: you probably want to have type="data" multiple="true"

tools/mash/mash_screen.xml
```diff
@@ -0,0 +1,82 @@
+<tool id="mash_screen" name="mash screen" version="2.1-0">
+    <description>Determine how well query sequences are contained within a pool of sequences.</description>
+    <requirements>
+        <requirement type="package" version="2.1">mash</requirement>
+    </requirements>
+    <version_command>mash --version</version_command>
+    <command detect_errors="exit_code"><![CDATA[
+        mash screen
+	     $queries
+	     $pool
```
bgruening: What is expected there as `$pool` is that a list of fastq files? Space separated?

bgruening:
>I'm not sure if galaxy needs a new datatype for mash sketches in order to handle that correctly(?)

If you assume that users upload those files and there will be a few other tools that can consume and manipulate them, I think it makes sense to include this datatype into Galaxy.

> I'm not sure if it's ok to include the ~100MB RefSeqSketches.msh file as test data.

This is unfortunately too big, is it possible to stay below 1MB?

Thanks @dfornika!

tools/mash/mash_screen.xml
dfornika: Yes. I think `mash screen` accepts `.fastq` or `.fa`, but it would be a space-separated list of sequence files.

dfornika:
I'll generate a smaller `.msh` file to use for testing.

tools/mash/mash_screen.xml
dfornika: That seems to work. Passing `planemo test` now with that fix.

tools/mash/mash_screen.xml
```diff
@@ -0,0 +1,91 @@
...
+	<param type="float" name="minimum_identity_to_report" value="0." min="-1." max="1." />
+	<param type="float" name="maximum_p_value_to_report" value="1." min="0." max="1."/>
+    </inputs>
+    <outputs>
+        <data name="output" format="tabular" />
+    </outputs>
+    <tests>
+        <test>
+            <param name="queries" value="NZ_MYON01000010.1.msh"/>
+            <param name="pool" value="ERR024951_seqtk_sample_1000.fastq"/>
```
bgruening: Can you include here at least 2 files or add a new test with more then one file?

tools/mash/mash_screen.xml
```diff
@@ -0,0 +1,91 @@
...
+             -i $minimum_identity_to_report
+             -v $maximum_p_value_to_report
+	     '$queries'.msh
+	     $pool
+	     > $output
+    ]]></command>
+    <inputs>
+        <param type="data" name="queries" format="binary" />
+        <param type="data" multiple="true" name="pool" format="fasta,fastq,fasta.gz,fastq.gz" />
+	<param type="boolean" name="winner_takes_all" />
```
bgruening: please adjust the indentation ... below as well :)

tools/mash/mash_screen.xml
```diff
@@ -0,0 +1,91 @@
...
+        ln -s '$queries' '$queries'.msh &&
+        mash screen
+	     #if $winner_takes_all
+                 -w
+             #end if
+             -i $minimum_identity_to_report
+             -v $maximum_p_value_to_report
+	     '$queries'.msh
+	     $pool
+	     > $output
```
bgruening: All data-parameters and text-parameters needs to be quoted.

tools/mash/mash_screen.xml
bgruening: Ok. If you add the mentioned test you will see that this will likely not work here. $pool will be a comma-separated list. You can do something like this `#str($pool).replace(',', ' ')#`. Do you know if the files need to have a proper file-ending as well?

nsoranzo:
> I'll generate a smaller .msh file to use for testing.

Thanks for replacing the test file! Can you please rebase the pull request squashing the commits, so that the big file will not be part of the repo history when this is merged?

dfornika:
@nsoranzo I'd be happy to do the rebase you suggested but could you suggest the exact sequence of git commands that I should execute to do that? I'm not sure how to do it and I don't want to make a mess of the commit history.

tools/mash/mash_screen.xml
dfornika: I've confirmed that `mash` doesn't seem to care about the file extension on the 'pool' file. It seems to detect valid fasta/fastq based on file contents.

dfornika:
I found this guide so I think I've done the rebase & squash:

https://github.com/servo/servo/wiki/Beginner's-guide-to-rebasing-and-squashing

tools/mash/mash_screen.xml
dfornika: sorry I blame emacs. fixed.

dfornika:
Thanks for the suggestion. I've got two out of three tests passing (at least when I run `planemo test` locally). I'm not sure at this point why the third test (with multiple input files) is failing.

dfornika:
@bgruening @nsoranzo I think this may be ready to merge (?)

nsoranzo:
@bgruening I have not done a full review, so feel free to merge!

tools/mash/mash_screen.xml
```diff
@@ -0,0 +1,102 @@
+<tool id="mash_screen" name="mash screen" version="@TOOL_VERSION@+galaxy0">
```
bgruening: @dfornika you probably want to add a profile tag here: https://docs.galaxyproject.org/en/latest/dev/schema.html#attributes 

Assuming your datatype gets into 19.01 this profile should reflect that.

dfornika: I don't actually refer to the `msh` datatype in this tool. For selfish reasons I'd like this tool to work on older galaxy versions if possible. If I were to add a `profile="19.01"` attribute, would that prevent the tool from installing on older galaxy versions?

Would it be ok to leave it out?

bgruening: Yes, that's what I mean. We need to give it a proper filetype for the input before we can merge this. Sorry.

> If I were to add a profile="19.01" attribute, would that prevent the tool from installing on older galaxy versions?

I think you can overrule this. Either way, you can always deploy this tool locally if you like but from the TS we should make sure people get the of-the-shelf-solution of your great tool.

dfornika: Ok, thanks for clarifying. I've added the `profile` attribute.

tools/mash/mash_screen.xml
```diff
@@ -0,0 +1,102 @@
...
+        mash screen
+             $winner_takes_all
+             -i $minimum_identity_to_report
+             -v $maximum_p_value_to_report
+             queries.msh
+             '${ str($pool).replace(',', '\' \'') }'
+             > '$output'
+    ]]></command>
+    <inputs>
+        <param type="data" name="queries" format="binary" />
```
nsoranzo:
```suggestion
        <param name="queries" type="data" format="msh" />
```

tools/mash/mash_screen.xml
```diff
@@ -0,0 +1,102 @@
...
+            <output name="output" file="mash_screen_NZ_MYON01000010.1_ERR024951_seqtk_sample_1000_2.tsv"/>
+        </test>
+        <test>
+            <param name="queries" value="NZ_MYON01000010.1.msh"/>
+            <param name="pool" value="ERR024951_seqtk_sample_1000_1.fastq,ERR024951_seqtk_sample_1000_2.fastq"/>
+            <output name="output" file="mash_screen_NZ_MYON01000010.1_ERR024951_seqtk_sample_1000_1and2.tsv"/>
+        </test>
+    </tests>
+    <help><![CDATA[
+Usage:
```
bgruening: @dfornika can we remove the usage and the options from the help, they are less meaningful in Galaxy.
Since the datatype was merged and is part of 19.01 I think we can merge this one :)

Thanks a lot!

dfornika: Done!

bgruening:
Merci @dfornika!

PR #2204, opened 2018-12-07
Fix the gemini data manager, data tables and wrapper tests

wm75:
This changeset should make the gemini data manager fully functional
and eliminate the need of any manual patching of the installed
gemini data files.
It also introduces a new gemini_versioned_databases data table that
allows gemini tool wrappers to filter the installed gemini
annotation files for those they can actually work with.

FOR CONTRIBUTOR:
* [x] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [ ] - License permits unrestricted use (educational + commercial)
* [ ] - This PR adds a new tool or tool collection
* [x] - This PR updates an existing tool or tool collection
* [ ] - This PR does something else (explain below)

FOR REVIEWER:
* [ ] .shed.yml file ok
    - [ ] Toolshed user `iuc` has access to associated toolshed repo(s)
* [ ] Indentation is correct (4 spaces)
* [ ] Tool version/build ok
* [ ] `<command/>`
  - [ ] Text parameters, input and output files `'single quoted'`
  - [ ] Use of `<![CDATA[ ... ]]>` tags
  - [ ] Parameters of type `text` or having `optional="true"` attribute are checked with `if str($param)` before being used
* [ ] Data parameters have a `format` attribute containing datatypes recognised by Galaxy
* [ ] Tests
  - [ ] Parameters are reasonably covered
  - [ ] Test files are appropriate
* [ ] Help
  - [ ] Valid restructuredText and uses `<![CDATA[ ... ]]>` tags
* [ ] Complies with other best practice in [Best Practices Doc](http://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html)

PR #1203, opened 2017-03-13
Condafy gemini tools and update to version 0.20.0

gregvonkuster:
Second part of handling this issue: https://github.com/galaxyproject/tools-iuc/issues/1130

PR #2208, opened 2018-12-14
Gemini version upgrade

wm75:
Now that #2204 is merged we can proceed with some confidence with upgrading the gemini tools to GEMINI version 0.20.1.

Fix https://github.com/galaxyproject/tools-iuc/issues/1130.

FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help


wm75:
This PR is intended to superseed #1203.

nsoranzo:
@wm75 What about integrating the requirement changes from https://github.com/galaxyproject/tools-iuc/commit/a0a575806c18c160014694f358168306390674f7#diff-92440ed36ad524a20f71d520097dbfef and https://github.com/galaxyproject/tools-iuc/pull/1203/commits/dae7a5ae41238d7e48daca30a9aa6366f44f4673 ?

wm75:
@nsoranzo ah right, tabix comes with the gemini conda package, of course. Why did you add htslib though?

nsoranzo:
Because tabix actually comes from htslib ;)
See http://www.htslib.org/download/

wm75:
right, but then we do not need to require either one?

nsoranzo:
Ah, I see now that it gets installed through an indirect dependency: gemini -> pysam -> htslib 1.9 .
Fine by me to drop both tabix and htslib then.

wm75:
Thanks, @nsoranzo, for working out the chain of dependencies and for thinking of the requirement change in the first place! Always nice to be able to simplify things :smile:

wm75:
Things look promising so far, but I'd like to combine this upgrade with some tool wrapper enhancements for an all shiny, new gemini experience :)

bgruening:
Thanks @wm75!

bgruening:
To get rid of the text families parameter, we could also include a repeat and do not allow whitespace. The current text box is also not sanitized I think.

wm75:
yeah, I played with repeat elements briefly, but when all you have to repeat is a single text field, they look & feel rather clunky.
The text box is not sanitized, and doesn't have to be. It's only when you start altering the set of valid chars that you have to think about these things. Of course, one current limitation here is that sample names can not contain shell tokens, quotes, etc. because these get sanitized by default. OTOH, there is really no need to use such characters in family or sample names, and most users would not even think about it.

wm75:
:relieved: This is it - from my side at least.

All tools and parameters that have been added to gemini since 0.18.1 are available through wrappers now.
There is one single exception I know of and just to have it documented:
it's that, according to the 0.20.0 changelog, gemini can now also parse annotations generated with bcftools bcsq (in addition to snpEff and VEP). If there's a need for it, I guess support for that in the gemini load wrapper can be added fairly easily, but it doesn't have to be in the very first release of the wrapper.

bgruening:
Awesome! Thanks a bunch, especially the query tool looks really nice.

PR #2209, opened 2018-12-14
Updates to VCFFilter documentation and filter field

VJalili:
This PR makes the following updates to the VCFFilter:
1. Update the documentation of VCFFilter for more intuitive description of how to specify filters;
2. Ignore `"` (double-quotation) and `'` (single-quotation) for the filter field (if given, the wrapper removes the quotations);
3. Require at least one filter.

### DIFF
--- a/tool_collections/vcflib/vcffilter/vcffilter.xml
+++ b/tool_collections/vcflib/vcffilter/vcffilter.xml
@@ -1,4 +1,4 @@
-<tool id="vcffilter2" name="VCFfilter:" version="@WRAPPER_VERSION@+galaxy1">
+<tool id="vcffilter2" name="VCFfilter:" version="@WRAPPER_VERSION@+galaxy2">
     <description>filter VCF data in a variety of attributes</description>
     <macros>
         <import>macros.xml</import>
--- a/tool_collections/vcflib/vcffilter/vcffilter.xml
+++ b/tool_collections/vcflib/vcffilter/vcffilter.xml
@@ -80,11 +80,18 @@ You can specify the following options within the **Specify filtering expression*
     -o, --or              use logical OR instead of AND to combine filters
     -r, --region          specify a region on which to target the filtering (must be used in conjunction with -f or -g)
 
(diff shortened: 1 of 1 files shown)


bgruening:
Thanks @VJalili!

tool_collections/vcflib/vcffilter/vcffilter.xml
```diff
@@ -30,7 +31,27 @@ input1.vcf.gz
...
+                </valid>
+                <mapping initial="none">
+                    <add source="&apos;" target=""/>
+                    <add source="&quot;" target=""/>
+                </mapping>
+            </sanitizer>
+        </param>
+
+        <!-- Additional filters are allowed. -->
+        <repeat name="filter_repeat" title="more filters">
```
bgruening: @VJalili you can add `min="1"` to a repeat, this means a user is forced to put in at least one iteration.

VJalili: Yes, that sounds much better. Thanks!

jennaj:
This should cover all the basic things we wanted to address. 

 - Help that matches updated input form !! (much clearer now)
 - Stripping quotes (users enter these by mistake due to older tutorial/form structure)
 - Requiring at least one filter term before job submission (users sometimes miss entering anything then not knowing what went wrong)

Thank you!

For reference, this is where we are tracking the fix all the way to usegalaxy.org: galaxyproject/usegalaxy-playbook#182

bgruening:
@VJalili please also increase the tool version.

VJalili:
I guess you're referring to the [wrapper version](https://github.com/galaxyproject/tools-iuc/blob/91a4a5aacf1e8bac5ddccb8504e46dae13b594aa/tool_collections/vcflib/macros.xml#L13). It says `rc`, not sure how we version these, but does `1.0.1` sound reasonable?

bgruening:
Thanks @VJalili and sorry for being so slow these days.

VJalili:
@bgruening Thank you for checking this.

PR #2223, opened 2019-01-04
gitignore: removed *log

bernt-matthias:
since I often missed to commit *log test data and was wondering why tests failed... 

I guess there is no django involved here, but the original entry was copied from somewhere 

FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,9 +44,6 @@ coverage.xml
 *.mo
 *.pot
 
-# Django stuff:
-*.log
-
 # Sphinx documentation
 docs/_build/
 


.gitignore
```diff
@@ -44,9 +44,6 @@ coverage.xml
 *.mo
 *.pot
 
-# Django stuff:
-*.log
```
bgruening: that can be potentially dangerous if people name test-data `*.log, isn't it?

bernt-matthias: But if there is a .log file in test-data then it should be in the git repo .. or?

Currently .log files (in test-data) are ignored by git (i.e. it is not shown in `git status` and can be added only with `git add -f`) which is confusing IMHO. In my case this led twice to the situation that  planemo tests were running on my local computer but they failed here (with the rather cryptic JobOutputsError). 

The only difference should be that log files show up in `git status` .. and can still be ignored if necessary.

bgruening: Oh, sorry, this was not entirely clear. it was a +1!

PR #2243, opened 2019-01-16
Fix the Only annotate ID field option of SnpSift Annotate

wm75:
From snpSift v4.1.h onwards, the -noInfo option is used instead
of the -id option of the annotate command is used to indicate that
the INFO field of the input VCF should be left untouched.

Unfortunately, the snpSift documentation was never updated to
reflect that change and the Galaxy wrapper was never changed.

FOR CONTRIBUTOR:

FOR REVIEWER:
* [ ] .shed.yml file ok
* [ ] Tool version/build ok
* [ ] `<command/>`
* [ ] Tests
* [ ] Help


tool_collections/snpsift/snpsift/snpSift_annotate.xml
```diff
@@ -10,27 +10,28 @@
...
-#elif str($annotate.info_ids).strip() != '':
-    -info "$annotate.info_ids"
-#end if
--q '$dbSnp' '$input' > '$output'
+        SnpSift -Xmx8G annotate
+        ${annotate.no_info}
+        #if str($annotate.info_ids).strip():
+            -info '$annotate.info_ids'
+        #end if
+        -q '$dbSnp' '$input' > '$output'
```
nsoranzo: I can't find a reference now, but I think it was agreed to not start the `<command />` with any indentation, like we do for `<help />`.

wm75: it makes for a slightly clearer diff, so why not

PR #2244, opened 2019-01-18
Allow multiple annotations to be extracted with gemini annotate.

wm75:
This functionality was always present in the gemini `annotate` command line tool, but was not previously exposed through the tool wrapper interface. This changeset improves also tool parameter help and removes major parts of the tool help that were describing command line issues not relevant in Galaxy.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/gemini/gemini_annotate.xml
+++ b/tools/gemini/gemini_annotate.xml
@@ -1,4 +1,4 @@
-<tool id="gemini_@BINARY@" name="GEMINI @BINARY@" version="@VERSION@.1">
+<tool id="gemini_@BINARY@" name="GEMINI @BINARY@" version="@VERSION@+galaxy1">
     <description>the variants in an existing GEMINI database with additional information</description>
     <macros>
         <import>gemini_macros.xml</import>
@@ -7,12 +7,8 @@
             <param argument="-c" name="column_name" type="text" value=""
             label="Database column name to use for recording annotations"
             help="A column with the name provided here will be added to the variants table of the GEMINI database to store the annotations">
-                <sanitizer invalid_char=" ">
-                    <valid initial="string.letters,string.digits">
-                        <add value="_" />
-                    </valid>
-                </sanitizer>
-                <validator type="expression" message="This field cannot be empty">value.strip()</validator>
+                <validator type="empty_field" />
+                <validator type="expression" message="Only alphanumeric characters and the underscore can be used in column names">value.replace('_', '').isalnum()</validator>
             </param>
         </xml>
     </macros>
--- a/tools/gemini/gemini_annotate.xml
+++ b/tools/gemini/gemini_annotate.xml
@@ -1,8 +1,20 @@
-<tool id="gemini_@BINARY@" name="GEMINI @BINARY@" version="@VERSION@">
-    <description>adding your own custom annotations</description>
+<tool id="gemini_@BINARY@" name="GEMINI @BINARY@" version="@VERSION@.1">
+    <description>the variants in an existing GEMINI database with additional information</description>
     <macros>
         <import>gemini_macros.xml</import>
         <token name="@BINARY@">annotate</token>
+        <xml name="add_as">
+            <param argument="-c" name="column_name" type="text" value=""
+            label="Database column name to use for recording annotations"
+            help="A column with the name provided here will be added to the variants table of the GEMINI database to store the annotations">
+                <sanitizer invalid_char=" ">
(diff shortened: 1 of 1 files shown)


tools/gemini/gemini_annotate.xml
```diff
@@ -1,8 +1,20 @@
-<tool id="gemini_@BINARY@" name="GEMINI @BINARY@" version="@VERSION@">
-    <description>adding your own custom annotations</description>
+<tool id="gemini_@BINARY@" name="GEMINI @BINARY@" version="@VERSION@.1">
```
nsoranzo:
```suggestion
<tool id="gemini_@BINARY@" name="GEMINI @BINARY@" version="@VERSION@+galaxy1">
```

tools/gemini/gemini_annotate.xml
```diff
@@ -18,73 +30,86 @@
...
+        label="Strict variant-identity matching of database and annotation records (VCF format only)"
+        help="The default is to consider VCF-formatted annotations only if a variant in the GEMINI database and a record in the annotation source describe the exact same nucleotide change at the same position in the genome. You can disable this option to make use of any annotation that overlaps with the position of a database variant. This setting is ignored for annotation sources in BED format, for which matching is always based on overlapping positions only." />
         <conditional name="a">
-            <param name="a_selector" type="select" label="How should the annotation file be used?" help="(-a)">
-                <option value="boolean">Did a variant overlap a region or not? (boolean)</option>
-                <option value="count">How many regions did a variant overlap? (count)</option>
-                <option value="extract" selected="True">Extract specific values from a BED/VCF file. (extract)</option>
+            <param argument="-a boolean|count|extract" name="a_selector" type="select"
```
nsoranzo:
```suggestion
            <param argument="-a" name="a_selector" type="select"
```

tools/gemini/gemini_annotate.xml
```diff
@@ -18,73 +30,86 @@
...
+            label="Type of information to add to the database variants"
+            help="">
+                <option value="boolean">Binary indicator (1=found, 0=not found) of whether the variant had any match in the annotation source</option>
+                <option value="count">Count of the number of matches found in the annotation source for the database variant</option>
+                <option value="extract" selected="True">Specific values extracted from matching records in the annotation source</option>
```
nsoranzo: I'd keep the option value in the description, e.g.:
```suggestion
                <option value="extract" selected="True">Specific values extracted from matching records in the annotation source (extract)</option>
```

tools/gemini/gemini_annotate.xml
```diff
@@ -1,8 +1,20 @@
...
+        <xml name="add_as">
+            <param argument="-c" name="column_name" type="text" value=""
+            label="Database column name to use for recording annotations"
+            help="A column with the name provided here will be added to the variants table of the GEMINI database to store the annotations">
+                <sanitizer invalid_char=" ">
+                    <valid initial="string.letters,string.digits">
+                        <add value="_" />
+                    </valid>
+                </sanitizer>
+                <validator type="expression" message="This field cannot be empty">value.strip()</validator>
```
nsoranzo:
```suggestion
                <validator type="empty_field" />
```

tools/gemini/gemini_annotate.xml
wm75: I can do this, of course, if it's the prefered way. @bgruening what do you think?

tools/gemini/gemini_annotate.xml
wm75: I changed this because I thought the option text was already quite long, but I don't have a strong opinion. Again, @bgruening, what do you prefer?

tools/gemini/gemini_annotate.xml
wm75: There's a subtle difference between these validators: my custom one will complain when the user enters only whitespace, while the predefined one will accept that, but whitespace-only is not a useful SQL column name.

**However**: your comment here made me realize that the sanitizer before isn't doing what it is supposed to do either, and needs to be fixed still, since
1) whitespace won't work anywhere in a column name
2) a sanitizer is a poor choice here because it is rather difficult for a user to find out what the sanitized name was, which means they cannot query that column later on.

=> I think I'll use another validator instead.

tools/gemini/gemini_annotate.xml
bgruening: It's the current IUC standard - I'm not a particular fan of it, but this is what the IUC prefers.

PR #2248, opened 2019-01-21
Fix genotype fields standardization of varscan_somatic

wm75:
The genotype fields modifications introduced in #2246 of the tool turned out to result in core dumps with real-world data with high coverage.
The issue is that htslib will allocate a fixed amount of space for each value of a given genotype field, when that field is first modified for any sample in a variant record. The amount of space depends on the
size of the largest value stored in the field for the sample and later attempts to add a value requiring more space to the same field (even for a different sample) will trigger the assert:
https://github.com/pysam-developers/pysam/blob/fadbafbe03c4340a6b817a07c408e7aa293bd5ef/htslib/vcf.c#L3496
resulting in a core dump.
This is fixed in this PR by modifying each genotype field starting with the sample that has the largest value for the field.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/varscan/varscan.py
+++ b/tools/varscan/varscan.py
@@ -465,7 +465,7 @@ class VarScanCaller (object):
         # genotype field starting with the sample with the largest value.
         new_gt_fields = {'AD': [], 'ADF': [], 'ADR': []}
         # store the current genotype field contents for each sample
-        per_sample_gts = list(record.samples.values())
+        per_sample_gts = record.samples.values()
         # calculate and store the new genotype field values
         for gt_field in per_sample_gts:
             # generate a standard AD field by combining VarScan's
--- a/tools/varscan/varscan.py
+++ b/tools/varscan/varscan.py
@@ -443,36 +443,60 @@ class VarScanCaller (object):
         # VarScan
         with pysam.VariantFile(varcall_template, 'r') as original_data:
             varscan_header = original_data.header
+        # don't propagate non-standard and redundant FORMAT keys written
+        # by VarScan
+        varscan_header.formats.remove_header('AD')
+        varscan_header.formats.remove_header('FREQ')
+        varscan_header.formats.remove_header('RD')
+        varscan_header.formats.remove_header('DP4')
         for sample in varscan_header.samples:
             common_header.samples.add(sample)
         common_header.merge(varscan_header)
-        # don't propagate non-standard and redundant FORMAT keys written
-        # by VarScan
-        common_header.formats.remove_header('FREQ')
-        common_header.formats.remove_header('RD')
-        common_header.formats.remove_header('DP4')
         return common_header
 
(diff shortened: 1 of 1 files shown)


tools/varscan/varscan.py
```diff
@@ -443,36 +443,60 @@ def _compile_common_header(self, varcall_template, no_filters=False):
...
-        for gt_field in read.samples.values():
+        # The key issue to consider here is that we need to modify
+        # genotype field values on a per-sample basis, but htslib
+        # reserves memory for the values of all samples upon the first
+        # modification of the field in the variant record.
+        # => We need to calculate all values first, then fill each
+        # genotype field starting with the sample with the largest value.
+        new_gt_fields = {'AD': [], 'ADF': [], 'ADR': []}
+        # store the current genotype field contents for each sample
+        per_sample_gts = list(record.samples.values())
```
nsoranzo: I don't think `list()` is needed.

wm75: Are you sure? I wasn't in the mood for experimenting and relying on the order of the values here and further down being the same. So I thought I'd play it safe. Also, because these values are needed twice, the cost of constructing the list is probably not that high?

wm75: Ah yes, you should be right. Since it's the values view that gets referenced, it should be fine without the list.

tools/varscan/varscan.py
```diff
@@ -443,36 +443,60 @@ def _compile_common_header(self, varcall_template, no_filters=False):
...
             )
+        # Modify the record's genotype fields.
+        # For each field, modify the sample containing the largest
+        # value for the field first.
+        # Without this precaution we could trigger:
+        # "bcf_update_format: Assertion `!fmt->p_free' failed."
+        # in vcf.c of htslib resulting in a core dump.
+        for field in sorted(new_gt_fields):
+            for new_field, sample_fields in sorted(
+                zip(new_gt_fields[field], per_sample_gts),
```
nsoranzo: I don't fully understand what you are doing here, but I'll raise a doubt any way.
Before Python 3.7, `per_sample_gts` entries may have a random order, so zipping with an array with fixed order seems suspicious.

wm75: That was exactly (my probably too paranoid) thinking that made me construct the list. The important thing here is just that the order of `per_sample_gts` has to be the same in this `for` loop as in the `for gt_field in per_sample_gts:` above. If that's guaranteed, then we know that the elements in the `new_gt_fields` lists are corresponding to their initial values in `per_sample_gts`.

Does that explain it?

nsoranzo: > The important thing here is just that the order of `per_sample_gts` has to be the same in this for loop as in the `for gt_field in per_sample_gts:` above.

Yes, that's guaranteed until you modify `record.samples`.

wm75:
@nsoranzo Thanks a lot!

PR #2257, opened 2019-01-24
Fix Snippy Paired Collection Handling

### LINKED ISSUE #2161: Snippy wrapper fails with Collection of Paired dataset
#set $dir_name = re.sub('[^\w_]', '_', $fastq_input.fastq_input1.element_identifier) this line of code will fail when the input is  collection of paired datasets

dfornika:
FOR CONTRIBUTOR:

Fixes #2161

### DIFF
--- a/tools/snippy/snippy.xml
+++ b/tools/snippy/snippy.xml
@@ -1,4 +1,4 @@
-<tool id="snippy" name="snippy" version="@VERSION@">
+<tool id="snippy" name="snippy" version="@VERSION@+galaxy1">
   <description>
       Snippy finds SNPs between a haploid reference genome and your NGS sequence reads.
   </description>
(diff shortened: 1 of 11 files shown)


dfornika:
There seems to be a bit of non-determinism in the test output, maybe introduced by some of the dependencies(?). I'll try to generate new test outputs.

dfornika:
I guess I should add a `+galaxy#` string onto the version:

```xml
<tool id="snippy" name="snippy" version="@VERSION@+galaxy0">
```
...but should it be `+galaxy0`, `+galaxy1` or some other value?

nsoranzo:
I'd use `+galaxy1` .

dfornika:
@bgruening I'm not sure why the tests are producing different output now, but I did purposely re-generate the test outputs to get the tests to pass.

I haven't changed any parameters or input files for the tests, so my best guess is that some of the dependencies aren't tightly pinned, resulting slightly different output compared to last time the test data was generated. Maybe a default parameter or something was changed in one of the dependencies? 

But it shouldn't be necessary to get to the root of that problem here, should it? Those new output files are consistent with what the tests produce.

bgruening:
Ok! Thanks @dfornika!

PR #2291, opened 2019-02-10
Collection column join: small changes to parameter text

mblue9:
I've had a few users a little confused by this tool so have made some small edits to try to clarify
Is this ok with you @abretaud?

FOR CONTRIBUTOR:

### DIFF
--- a/tools/collection_column_join/collection_column_join.xml
+++ b/tools/collection_column_join/collection_column_join.xml
@@ -56,8 +56,8 @@ cat header${ ( $i + 1 ) % 2 }.tmp output${ ( $i + 1 ) % 2 }.tmp > "${tabular_out
     <inputs>
         <param name="input_tabular" type="data" format="tabular" multiple="True" optional="False" label="Tabular files"/>
         <!-- <param name="identifier_column" type="data_column" data_ref="input_tabular" value="0" min="0" optional="False" label="Identifier column"/> -->
-        <param name="identifier_column" type="integer" value="1" min="0" optional="False" label="Identifier column" help="The column that will be used to join the input files"/>
-        <param name="has_header" type="integer" value="0" min="0" optional="False" label="Number of header lines in each input file" help="If this is set to 0, a header line will be added containing the names of the input files/columns and the identifier column will be named #KEY. If you have a header row in your input files set this to 1."/>
+        <param name="identifier_column" type="integer" value="1" min="0" optional="False" label="Identifier column" help="The column that will be used to join the input datasets"/>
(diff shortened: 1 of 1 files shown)


tools/collection_column_join/collection_column_join.xml
```diff
@@ -56,9 +56,9 @@ cat header${ ( $i + 1 ) % 2 }.tmp output${ ( $i + 1 ) % 2 }.tmp > "${tabular_out
     <inputs>
         <param name="input_tabular" type="data" format="tabular" multiple="True" optional="False" label="Tabular files"/>
         <!-- <param name="identifier_column" type="data_column" data_ref="input_tabular" value="0" min="0" optional="False" label="Identifier column"/> -->
-        <param name="identifier_column" type="integer" value="1" min="0" optional="False" label="Identifier column"/>
-        <param name="has_header" type="integer" value="0" min="0" optional="False" label="Number of Header lines in each item"/>
-        <param name="old_col_in_header" type="boolean" checked="true" label="Keep original column header" help="Disable if you want columns headers to be only composed of the input dataset names"/>
+        <param name="identifier_column" type="integer" value="1" min="0" optional="False" label="Identifier column" help="The column that will be used to join the input files"/>
```
bernt-matthias: maybe we should use "data set" instead of "file"?

nsoranzo: Or "dataset", which is the one most commonly used.

tools/collection_column_join/collection_column_join.xml
```diff
@@ -56,9 +56,9 @@ cat header${ ( $i + 1 ) % 2 }.tmp output${ ( $i + 1 ) % 2 }.tmp > "${tabular_out
...
+        <param name="has_header" type="integer" value="0" min="0" optional="False" label="Number of header lines in each input file" help="If this is set to 0, a header line will be added containing the names of the input files/columns and the identifier column will be named #KEY. If you have a header row in your input files set this to 1."/>
```
bernt-matthias: Still a bit unclear: 

- what happens for more than one header line (I suggest: "If you have one or more header lines in your input set this to the number of header lines".

- Suggestion for the 0 case: If this is set to 0, a header line will be added containing column names as follows: the identifier column will be named #KEY and the other columns are named by the input data set names /columns.

bernt-matthias:
In the example I would remove the ".tabular" from the data set names in order to make it more easy to comprehend the main point.

mblue9:
@bernt-matthias and @nsoranzo thanks for the review! I've made those changes.

PR #2299, opened 2019-02-14
get summary.pdf file from working directory

FredericBGA:
the pdf file was not imported in history.

### DIFF
--- a/tools/bcftools/bcftools_stats.xml
+++ b/tools/bcftools/bcftools_stats.xml
@@ -128,7 +128,7 @@ ${section.verbose}
     </inputs>
     <outputs>
         <data name="output_file" format="txt"/>
-        <data name="output_pdf" format="pdf">
+        <data format="pdf" label="${plot_title}.pdf" name="summary.pdf" from_work_dir="plot_tmp/summary.pdf">
             <filter>plot_title</filter>
         </data>
     </outputs>
--- a/tools/bcftools/bcftools_stats.xml
+++ b/tools/bcftools/bcftools_stats.xml
@@ -128,7 +128,7 @@ ${section.verbose}
     </inputs>
     <outputs>
         <data name="output_file" format="txt"/>
-        <data format="pdf" label="${plot_title}.pdf" name="summary.pdf" from_work_dir="plot_tmp/summary.pdf">
+        <data format="pdf" label="${plot_title}.pdf" from_work_dir="plot_tmp/summary.pdf">
             <filter>plot_title</filter>
         </data>
     </outputs>


nsoranzo:
@FredericBGA Thanks for the patch! I have also written a very similar fix as part of https://github.com/galaxyproject/tools-iuc/pull/1416 . Unfortunately I don't think that's sufficient to have working plots, I'll try again to iron the last issues there, if that's OK.

FredericBGA:
@nsoranzo Thanks! It works here, summary.pdf holds stat, Ts/Tv by sample [...] and Substitution types plots.

nsoranzo:
> Thanks! It works here, summary.pdf holds stat, Ts/Tv by sample [...] and Substitution types plots.

@FredericBGA I suppose that's because you have a LaTeX distribution installed on the Galaxy server, outside of Conda.

tools/bcftools/bcftools_stats.xml
```diff
@@ -128,7 +128,7 @@ ${section.verbose}
     </inputs>
     <outputs>
         <data name="output_file" format="txt"/>
-        <data name="output_pdf" format="pdf">
+        <data format="pdf" label="${plot_title}.pdf" name="summary.pdf" from_work_dir="plot_tmp/summary.pdf">
```
blankenberg: We should probably not change the `name` attribute here.

FredericBGA:
> 
> 
> > Thanks! It works here, summary.pdf holds stat, Ts/Tv by sample [...] and Substitution types plots.
> 
> @FredericBGA I suppose that's because you have a LaTeX distribution installed on the Galaxy server, outside of Conda.
LaTeX is then a requirements for plot-vcfstats, not Galaxy, right ?

tools/bcftools/bcftools_stats.xml
FredericBGA: ok, updated. thank you @blankenberg

nsoranzo:
> > > Thanks! It works here, summary.pdf holds stat, Ts/Tv by sample [...] and Substitution types plots.
> > 
> > 
> > @FredericBGA I suppose that's because you have a LaTeX distribution installed on the Galaxy server, outside of Conda.
> 
> LaTeX is then a requirements for plot-vcfstats, not Galaxy, right ?

I mean that I think it is working for you only because you have LaTeX installed where the tool is run, since `plot-vcfstats` requires `pdflatex`, which is not installed in the conda environment. That means that normally this is broken, and I am trying to fix it in #1416.

FredericBGA:
> 
> 
> > > > Thanks! It works here, summary.pdf holds stat, Ts/Tv by sample [...] and Substitution types plots.
> > > 
> > > 
> > > @FredericBGA I suppose that's because you have a LaTeX distribution installed on the Galaxy server, outside of Conda.
> > 
> > 
> > LaTeX is then a requirements for plot-vcfstats, not Galaxy, right ?
> 
> I mean that I think it is working for you only because you have LaTeX installed where the tool is run, since `plot-vcfstats` requires `pdflatex`, which is not installed in the conda environment. That means that normally this is broken, and I am trying to fix it in #1416.

ok, I understand.

nsoranzo:
Closing in favour of https://github.com/galaxyproject/tools-iuc/pull/1416 . Thanks for the contribution!

Issue #2301, opened 2019-02-15
XMFA to GFF3 converter may be outputing incomplete data


innovate-invent:
https://github.com/galaxyproject/tools-iuc/blob/16205b1cd782c5a4f7c08af7d4ee50dc5903418b/tools/progressivemauve/xmfa2gff3.py#L115

Perhaps I don't understand how this tool is representing alignments in GFF3 but shouldn't there be coordinates in the target field?

bgruening:
@innovate-invent this looks alright isn't it?

https://github.com/galaxyproject/tools-iuc/blob/16205b1cd782c5a4f7c08af7d4ee50dc5903418b/tools/progressivemauve/test-data/1.gff3

bgruening:
ping @abretaud

innovate-invent:
If you have a look at the gff3 standards description of the Target attribute you will see that the linked test data is incomplete.

PR #2306, opened 2019-02-19
Fix Target attribute output

innovate-invent:
The current output does not contain enough information to describe an alignment. The current output is also not GFF3 compliant. This commit is in reference to #2301

FOR CONTRIBUTOR:

### DIFF
--- a/tools/progressivemauve/xmfa2gff3.py
+++ b/tools/progressivemauve/xmfa2gff3.py
@@ -109,7 +109,7 @@ def convert_xmfa_to_gff3(xmfa_file, relative_to='1', sequences=None, window_size
         for other in others:
             targetlist = [label_convert.get(other['id'], other['id']), str(other['start']), str(other['end'])]
             if other['strand'] != '.':
-                targetlist.append(other['strand'])
+                targetlist.append(str(other['strand']))
             other['feature'] = SeqFeature(
                 FeatureLocation(parent['start'], parent['end'] + 1),
                 type="match", strand=parent['strand'],
--- a/tools/progressivemauve/xmfa2gff3.py
+++ b/tools/progressivemauve/xmfa2gff3.py
@@ -107,7 +107,7 @@ def convert_xmfa_to_gff3(xmfa_file, relative_to='1', sequences=None, window_size
         others = [seq for seq in lcb if seq['id'] != relative_to]
 
         for other in others:
-            targetlist = [label_convert.get(other['id'], other['id']), other['start'], other['end']]
+            targetlist = [label_convert.get(other['id'], other['id']), str(other['start']), str(other['end'])]
             if other['strand'] != '.': targetlist.append(other['strand'])
             other['feature'] = SeqFeature(
                 FeatureLocation(parent['start'], parent['end'] + 1),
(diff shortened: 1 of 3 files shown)


innovate-invent:
@abretaud Thanks for the help. Sorry I forgot to set this to WIP. I have some pending commits that remove the code you are editing.

innovate-invent:
Testing the latest change exposes a bug in BioPy URI encoding '+' in the attributes when it shouldn't be.
I am going to abandon this tool for a simpler awk script that converts the backbone to GFF as I don't need scoring.

abretaud:
Is there an issue for this encoding problem in https://github.com/biopython/biopython?

innovate-invent:
Looks like the BCBio library has been depreciated.
https://github.com/chapmanb/bcbb/blob/master/gff/README.rst

erasche:
@innovate-invent yeah it's deprecated the replacement for it has completely different semantics, and the current version of bcbio works sufficiently for most purposes. I tried to swap to gffutils at one point but had no success.

tools/progressivemauve/xmfa2gff3.py
```diff
@@ -112,7 +112,7 @@ def convert_xmfa_to_gff3(xmfa_file, relative_to='1', sequences=None, window_size
                 type="match", strand=parent['strand'],
                 qualifiers={
                     "source": "progressiveMauve",
-                    "target": label_convert.get(other['id'], other['id']),
+                    "Target": " ".join(map(str, [label_convert.get(other['id'], other['id']), other['start'], other['end'], '+' if other['strand'] > 0 else '-'])),
```
erasche: Would it make sense to add this to the `_id_tn_convert` function?

innovate-invent: Do you mean _id_tn_dict?
I don't think that would be right.

erasche: Yep that one, sorry, typo.

innovate-invent:
That is a standing bug in the aforementioned BCBio gff library.

hexylena:
Huh, that's unfortunate. Would ±1 work instead? Or we could consider to
fork and maintain it, it will be a dependency for a while due to sheer
amount of legacy code I have running on that lib. Do you have a link to the
issue? (On mobile or I'd look myself)

tir. 23. jun. 2020, 18:25 skrev Nolan Woods <notifications@github.com>:

innovate-invent:
I believe this issue relates to the problem we are seeing: https://github.com/chapmanb/bcbb/issues/86

hexylena:
@innovate-invent ok, I've added a small monkey patch for bcbio with the following logic:

- scope:
  - very limited scope of what we're patching, already abstracted into a function for us.
  - tiny function to override
  - most importantly: we have significant control over what data is in the key values, we know it will be safe overall, so the more advanced things the original function does, does not matter for us.
- bcbio is deprecated
  - I'm not interested in patching, other patches are upstreamed there but new releases are not forthcoming. the author has deprecated it, we should respect their wish to not be responsible for it, and forking is a significant undertaking.
  - I've had another look at the recommended replacement and it looks a lot more usable than it used to, so, switching to it is now a definite possibility
  - but dozens of tools depend on it, so, I'll do the migration to the new library in one go.

This patch has languished (my fault, I'm sorry) and it should definitely get in :) With this patch, I believe the tests will pass again.

PR #2311, opened 2019-02-22
Barcode splitter

hepcat72:
Note, there's a bioconda recipe pr for this tool waiting.

FOR CONTRIBUTOR:


tools/barcode_splitter/barcode_splitter.xml
```diff
@@ -0,0 +1,248 @@
...
+                #end for
+            #end if
+
+            ##Add the relative file argument positions of the index files in the list included above
+            --idxread
+            #for $n in range( 1, $num_index_files )
+                ${n}
+            #end for
+            
+            > $summary
```
bgruening: single quotes

tools/barcode_splitter/barcode_splitter.xml
```diff
@@ -0,0 +1,248 @@
...
+                    ${indf.input2}
+                    #set $num_index_files += 1
+                    #set $num_files += 1
+                #end for
+            #end if
+
+            ##Add the sequence files without indexes
+            #if self.varExists('seqfiles'):
+                #for $sf in $seqfiles
+                    ${sf.input}
```
bgruening: single quotes

tools/barcode_splitter/barcode_splitter.xml
```diff
@@ -0,0 +1,248 @@
...
+            mkdir split &&
+            barcode_splitter
+            
+            ##There must be an index file, so supply the extension of the first one
+            #for $indf in $indexfiles
+                --format "${indf.input2.extension}"
+                #break
+            #end for
+            
+            --bcfile $bcfile --mismatches "$mismatches" --galaxy $barcodes_at_end $split_all --prefix "split/" --suffix ".${indf.input2.extension}"
```
bgruening: please use for all paths and for all text-params always single-quotes

tools/barcode_splitter/barcode_splitter.xml
```diff
@@ -0,0 +1,248 @@
...
+    <version_command>barcode_splitter --version</version_command>
+
+    <command detect_errors="exit_code">
+        <![CDATA[
+            mkdir split &&
+            barcode_splitter
+            
+            ##There must be an index file, so supply the extension of the first one
+            #for $indf in $indexfiles
+                --format "${indf.input2.extension}"
```
bgruening: is fastqsanger really recognised?

tools/barcode_splitter/barcode_splitter.xml
```diff
@@ -0,0 +1,248 @@
...
+            #end for
+            
+            --bcfile $bcfile --mismatches "$mismatches" --galaxy $barcodes_at_end $split_all --prefix "split/" --suffix ".${indf.input2.extension}"
+
+            ##Add the sequence files with indexes (which must come first)
+            #set $num_files = 0
+            #set $num_index_files = 1
+            #if self.varExists('indexfiles'):
+                #for $indf in $indexfiles
+                    ${indf.input2}
```
bgruening: quotes

tools/barcode_splitter/barcode_splitter.xml
```diff
@@ -0,0 +1,248 @@
...
+                --format "${indf.input2.extension}"
+                #break
+            #end for
+            
+            --bcfile $bcfile --mismatches "$mismatches" --galaxy $barcodes_at_end $split_all --prefix "split/" --suffix ".${indf.input2.extension}"
+
+            ##Add the sequence files with indexes (which must come first)
+            #set $num_files = 0
+            #set $num_index_files = 1
+            #if self.varExists('indexfiles'):
```
bgruening: `#if $indexfiles` should work

tools/barcode_splitter/barcode_splitter.xml
```diff
@@ -0,0 +1,248 @@
...
+                    #else:
+                        && perl -e 'foreach(@ARGV){\$nn=\$_;\$nn=~s/-(read-[^\.]+)/-XxX-\$1/;rename(\$_,\$nn)}' split/*-read-${n}.${indf.input2.extension}
+                    #end if
+                #end for
+            #end if
+            #if self.varExists('seqfiles'):
+                #for $repeat in $seqfiles
+                    #set $n += 1
+                    #if str($repeat.readname) != "":
+                        && perl -e '\$r=shift(@ARGV);foreach(@ARGV){\$nn=\$_;\$nn=~s/-read-[^\.]+/-XxX-\$r/;rename(\$_,\$nn)}' "$repeat.readname" split/*-read-${n}.${indf.input2.extension}
```
bgruening: if you are using perl, please add perl to your requirements...

tools/barcode_splitter/barcode_splitter.xml
```diff
@@ -0,0 +1,248 @@
...
+            <param format="fasta,fastq,fastqsanger,fastqsolexa,fastqillumina" name="input2" type="data" label="Index Read File, typically 'Read 2 (Index Read)'" help="If there are multiple barcode columns in the barcode file, index sequence files must be supplied in the same number and order as the barcode columns.  If your reads have barcodes embedded in them, that file must be submitted here as an index file  and 'Split all files' below, must be true." />
+            <param name="indexname" type="text" label="Optional Name" help="E.g. index1, index2, etc.  Letters and numbers only.">
+                <sanitizer>
+                    <valid initial="string.ascii_letters,string.digits" />
+                </sanitizer>
+            </param>
+        </repeat>
+
+        <param name="mismatches" type="integer" size="3" value="0" max="2" min="0" label="Number of allowed mismatches" help="An integer between 0 and 2 (inclusive).  Warning: Make sure your barcodes all differ from one another by at least this many nucleotides plus 1, otherwise sequences that match both barcodes equally well will be thrown out as 'multimatched' reads." />
```
bgruening: size can be removed, it is not used anymore

tools/barcode_splitter/barcode_splitter.xml
```diff
@@ -0,0 +1,248 @@
...
+
+        <param name="barcodes_at_end" type="boolean" truevalue="--barcodes_at_end" falsevalue="" checked="false" label="Barcodes are at the end of all sequences" help="Default is the beginning of all sequences" />
```
bgruening: you could use argument="--barcodes_at_end" instead of name

tools/barcode_splitter/barcode_splitter.xml
```diff
@@ -0,0 +1,248 @@
...
+
+        <param name="split_all" type="boolean" truevalue="--split_all" falsevalue="" checked="false" label="Split all files" help="Use this if your desired output sequences have barcode sequences in them." />
```
bgruening: same here, argument=

nsoranzo:
Looks quite similar to https://github.com/galaxyproject/tools-iuc/tree/master/tool_collections/fastx_toolkit/fastx_barcode_splitter ?

tools/barcode_splitter/barcode_splitter.xml
hepcat72: That's a good question.  I didn't write the file reading code or add the related options.  I just dug through it and it looks like fastq is the only format the script reads (despite the help text saying the format can be fasta).  The value supplied to --format appears to only be involved in construction the output file's extension(s) (in certain cases?), e.g. .fastq.gz. I'll create an issue on the tool page to (at the least) remove the 'fasta' format mention, or at the most, extricate the format option.

hepcat72:
@nsoranzo - Yeah, it's similar.  I just asked Lance about it.  He said the original fastx one didn't have the ability to process index reads in a separate file, so he (forked? and) modified it.  Then he had created his own version in python (fastx's version is in perl).  He asked me to add the ability to handle multiple columns of barcodes and submit it to the IUC.  That's the main difference now.

bernt-matthias:
>  He said the original fastx one didn't have the ability to process index reads in a separate file, so he forked? and) modified it.

Seems to be not the case. There are separate parameters as far as I can see. 

Anyway I would go for `cutadapt`.

Don't like the additional perl dependency. I would have hoped that the program names them in a useful way in the first place.

hepcat72:
Don’t get me wrong. The fastx barcode splitter does a decent job, but it only handles one column of barcodes to split on sequences in a single index file. This tool in this PR handles multiple columns of barcodes to split on sequences residing in multiple index files.

I don’t understand how cutadapt can be applied here. Cutadapt could separate an index occurring at the beginning of a sequence after splitting, but how can it address situations involving multiple index files? Until we made this tool, splitting with 2 index files required 2 barcode splitting runs with intermediate files. This tool results in only needing to run once with no internediate files. Much simpler.

The Perl dependency is not necessary. I could take it out. I just wanted consistent/meaningful outfile names. The way Lance wrote the original, it names the files as read-1, read-2, etc, but those numbers simply refer to the order in which the read files were supplied, which can be particularly confusing if the input files were read1, read2, etc, but in a different order - which in this wrapper, they usually are, because of the way I had to implement the —index option, which uses a number indicating which files (in the order supplied) are index files. I used Perl only because that’s what I’m most familiar with. Do you have another file renaming method you can suggest?

bernt-matthias:
> The Perl dependency is not necessary. I could take it out. 

You are absolutely right, the data sets should be named reasonably. Maybe it can be done with cheetah or bash? Or the splitter can be convinced to do it?

> I don’t understand how cutadapt can be applied here.

https://cutadapt.readthedocs.io/en/stable/guide.html#demultiplexing

One can specify an arbitrary number of `-a` and `-A` options and also give names for adapters (https://cutadapt.readthedocs.io/en/stable/guide.html#named-adapters) for useful named outputs.

bernt-matthias:
Please don't get me wrong. The tool certainly is a valuable contribution that is probably useful (for some). And its always good to have several tools for one task. So, if you want to maintain the tool I'm happy. 

I just want to check if there are better alternatives for the task. If so, it might be (on the long run) better if the IUC would then combine forces toward these tools.

My biggest problem with the program (not the wrapper) is that it is essentially undocumented, i.e. nobody, who did not dig in the sources, knows what the tool is doing in detail (in particular how is adapter alignment done) -- a big plus of cutadapt where also the algorithmic details are documented.

hepcat72:
> One can specify an arbitrary number of -a and -A options and also give names for adapters

Ah, cool. I never thought to use cutadapt for barcode splitting. However, if I understand the documentation correctly, with cutadapt, you cannot handle multiple index files. Let me explain with a real world example, because I have a feeling perhaps there’s a miscommunication. We frequently have 4 files coming off the sequencer: 250bp forward, 250bd reverse, 8bp index 1, and 8bp index 2. We want to split forward and reverse. Say there are 40 samples. There are usually 20 unique barcodes in index 1 and 2 unique barcodes in index 2. When splitting forward and reverse, we only know which of the 40 samples the pair belongs to by knowing the value of its 2 barcodes in index files 1 & 2 together. It’s the combination that uniquely identifies the sample that the sequence belongs to. Neither cutadapt nor fastx barcode splitter can do that in 1 run.

Regarding renaming, I *might* be able to use cheetah and bash, but I would have to rely on (more) assumptions about the composition of the names and presence of the output files. I’ll think about it.

We use the tool internally on a daily basis. It has been integrated into out automated system, so both Lance and I actively support it.

As far as algorithmic documentation, I will add to the wrapper for now, and create an issue to add that to the tool itself. Briefly, the tool simply precomputes a dictionary of all possible mismatches and substrings the sequence to do a lookup in the dictionary. It catches barcode collisions during the dictionary construction, which results in warnings and/or errors and reads ending up in a “multimatched” file. The dictionary is multi-dimensional. The number of dimensions is determined by the number of barcode columns in the barcode file.

hepcat72:
OK.  I have re-implemented the file renaming to use bash & cheetah and took out the perl dependency.  I'll now write up something about the algorithm in the description.

hepcat72:
I added a description of the barcode matching mechanism to the help text.

All concerns have now been addressed.  Let me know if I missed anything or if there is anything further.

hepcat72:
Ah wait. I was just playing around with the tool in planemo and I realized I can improve the output collection construction.  I'm going to change this to WIP and see if I can optimize the collection output.

hepcat72:
OK.  I finished a revamp of the tool to create more appropriate output collections.  I also created 2 accompanying simplified interfaces for the most common use-cases of single-end & paired-end and a 3rd "flexible" interface for any scenario those don't cover (which I also stream-lined a bit, based on the first version I submitted).

The only thing I don't like about the result here is the hard-to-read filter code, but there doesn't appear to be a straight-forward way to count elements in a repeat that have specific values - and the main intent was to support list:paired, and avoid overlap among members in the different collections. I also wanted to separate pure index files from reads with the target data in them.  And I wanted the flexible interface to still have the ability to generate list:paired. So it is what it is, I guess.

@bernt-matthias @bgruening @nsoranzo @lparsons

hepcat72:
Cool.  All checks have passed.

tools/barcode_splitter/barcode_splitter.xml
```diff
@@ -0,0 +1,819 @@
...
+The output of this tool is a summary table displaying the split counts for each barcode identifier and the percentage of the total reads those represent.
+In addition, each fastq file produced will be loaded into the galaxy history as part of a collection list.
+        ]]>
+    </help>
+
+    <citations>
+        <citation type="bibtex">
+          @misc{paired_sequence_utils,
+              title = {{Barcode}-{Splitter}},
+              url = {https://bitbucket.org/hepcat72/paired_sequence_utils},
```
lparsons: That URL seems dead, I believe it should be: https://bitbucket.org/princeton_genomics/barcode_splitter/

tools/barcode_splitter/barcode_splitter.xml
```diff
@@ -0,0 +1,819 @@
...
+                                    <valid initial="string.ascii_letters,string.digits" />
+                                </sanitizer>
+                            </param>
+                        </when>
+                    </conditional>
+                </repeat>
+            </when>
+        </conditional>
+
+        <param name="mismatches" type="integer" value="0" max="2" min="0" label="Number of allowed mismatches" help="An integer between 0 and 2 (inclusive).  Warning: Make sure all your barcodes differ by at least double this value plus 1, otherwise sequences that match both barcodes equally well will be discarded as 'multimatched' reads." />
```
lparsons: Default of 1 seems most appropriate.

tools/barcode_splitter/barcode_splitter.xml
```diff
@@ -0,0 +1,819 @@
...
+
+        <param argument="--barcodes_at_end" type="boolean" truevalue="--barcodes_at_end" falsevalue="" checked="false" label="Barcodes are at the end of all index sequences" help="Default is the beginning of all sequences" />
+
+        <param argument="--split_all" type="boolean" truevalue="--split_all" falsevalue="" checked="false" label="Split index files too" help="This creates an output collection for split files whose selected 'Read Type' is 'Index'." />
+    </inputs>
+
+    <outputs>
+        <data format="tabular" name="summary" label="${tool.name} on ${on_string}: Summary" />
```
lparsons: Consider adding `column_names` action: https://docs.galaxyproject.org/en/latest/dev/schema.html#unconditional-actions-and-column-names

hepcat72:
> I like how the single end and paired end splitting work with single or dual indexes. However, when the index reads are split, they are all mashed into a single collection (instead of the expected one collection per index that mirrors the sample names of the other collection).
> 
> I think the "flexible" interface should create one collection per input read file and dispense with the "read type" and the ability to create paired collections. It seems simpler and the user could merge or filter the collections as needed after that.

You've mentioned this before, but I would have to set a max number of input files and create a static collection for each of that max number, and I'm not sure how I would label the collections. The names and types are discovered via discover datasets.  It would be easier to create a list:list.

tools/barcode_splitter/barcode_splitter.xml
hepcat72: Good catch.

- [ ] Fix bitbucket URL

tools/barcode_splitter/barcode_splitter.xml
hepcat72: Seems reasonable.

- [ ] Set mismatch default to 1

tools/barcode_splitter/barcode_splitter.xml
hepcat72: I would have to do something like this for the variable number of barcode columns, but I would have to use code similar to the current filter code to get the number of index files.

`"Geneid,${','.join([a.name for a in $input_files])}"`

Is that what you were thinking?

lparsons: Hmm... that could be tricky the way it's implemented. Might not be worth the effort/complexity. But yeah, that's what I was thinking.

lparsons:
> You've mentioned this before, but I would have to set a max number of input files and create a static collection for each of that max number, and I'm not sure how I would label the collections. The names and types are discovered via discover datasets. It would be easier to create a list:list.

The collections could be labeled by the "name" given to each input read. But it would require considerable changes. I'm ok with going the way it is and then making changes if we run into use cases that don't fit well.

tools/barcode_splitter/barcode_splitter.xml
hepcat72: Yeah.  While I already worked out how to essentially "grep" the files array, if they didn't enter a name, I would have to have logic to use a generic "-read-#" name, so a map with a conditional would be necessary too.  I think it's doable.  It would just be as difficult to read as the filter code currently is, so I'll forgo...

hepcat72:
> > You've mentioned this before, but I would have to set a max number of input files and create a static collection for each of that max number, and I'm not sure how I would label the collections. The names and types are discovered via discover datasets. It would be easier to create a list:list.
> 
> The collections could be labeled by the "name" given to each input read. But it would require considerable changes. I'm ok with going the way it is and then making changes if we run into use cases that don't fit well.

I'd either have to require the name be entered (and unique) or have code to determine what to name the collection if they don't provide a name... and I'm not sure you can enter code in the label attribute's value.  Also, I would still need some filter code that checks the size of the repeat and the read type (i.e. to determine whether it's going to be output) to hide collections that aren't populated.

hepcat72:
> > > You've mentioned this before, but I would have to set a max number of input files and create a static collection for each of that max number, and I'm not sure how I would label the collections. The names and types are discovered via discover datasets. It would be easier to create a list:list.
> > 
> > 
> > The collections could be labeled by the "name" given to each input read. But it would require considerable changes. I'm ok with going the way it is and then making changes if we run into use cases that don't fit well.
> 
> I'd either have to require the name be entered (and unique) or have code to determine what to name the collection if they don't provide a name... and I'm not sure you can enter code in the label attribute's value. Also, I would still need some filter code that checks the size of the repeat and the read type (i.e. to determine whether it's going to be output) to hide collections that aren't populated.

You know, I was thinking about this. If I wasn't supporting output of paired-end collections in flexible mode, I'd be able to implement your suggestion (with the caveat that I'd have to set the repeat max and require a read name be entered).  However, if anyone has paired-end or single-end data with embedded barcodes, they wouldn't be able to include it in a workflow that needed to connect it with a tool that takes list:paired as input.  I don't think(?) there's a way to (reliably?) create a list:paired from 2 lists in a workflow. And that was my primary motivation for the current output collections.

Anyway, that's just a point of discussion.  Presuming the CI succeeds, this PR should be ready.

hepcat72:
Any other suggestions/changes?

hepcat72:
Is there anything else anyone would like done to this PR?

bgruening:
@hepcat72 please add a .shed.yml.

@nsoranzo @bernt-matthias any concerns from your side?

hepcat72:
Let me know if the .shed.yml needs any edits.

tools/barcode_splitter/.shed.yml
```diff
@@ -0,0 +1,15 @@
+categories:
+- Fastq Manipulation
+description: A utility to split sequence files using multiple sets of barcodes
+long_description: |
+    Split multiple fastq files by matching barcodes in one or more of the
+    sequence files. Barcodes in the tab-delimited barcodes.txt file are matched
+    against the beginning (or end) of the specified index read(s). By default,
+    barcodes must match exactly, but --mistmatches can be set higher if desired.
+    Compressed input is read (from all files) if the first input file name ends
+    in .gz. Reading of compressed input can be forced with the gzipin option.
```
nsoranzo: The last 2 lines don't seem relevant for the tool description (and gzipin is not in the wrapper anyway).

hepcat72: Removed.  Thanks.

PR #2335, opened 2019-03-21
Fixes 3 bugs

cgirardot:
The tool was : 
- not passing the `min_mean_count` option from web form 
- not filtering on any `min_mean_count` (as it was always NULL) in the `create_seq_expression_set()` function
- not using the filtered data to create the `newSeqExpressionSet` in the `create_seq_expression_set()` function
- I also add trouble to actually deploy the tool at first and had to manually deploy a conda env. I reflected it by bumping up the packages versions and adding the needed getup for cmd line option parsing

I doubt the tool actually ever worked :-/

FOR CONTRIBUTOR:


abretaud:
Hi @cgirardot and thanks for the PR!
Looks ilke the [tests are failing](https://travis-ci.org/galaxyproject/tools-iuc/jobs/509520367#L3292), do you know where this could come from? (the new version or the changes you've made?)

cgirardot:
we are looking into this. First 2 issues are most likely because the test expects some precise value that should now be different. And the third one, we are not sure yet. Otherwise the code works on our installation.

cgirardot:
The `get_deseq_dataset()` routine was : 
- 1. Using a command line variable instead of the one passed to the method
- 2. Blindly setting header=FALSE instead of using the passed argument

The sailfish test needs `min_mean_count=0` to run

cgirardot:
@abretaud the last test still fails despite the additional fixes. The plugin executes correctly (did it interactively in my planemo). The problem is that we do not understand the error. Could you have a look please ?

cgirardot:
@abretaud ok finally found "by trying" what the problem was. The tests' tests were also incorrect. How this could previously pass the tests is a mystery to me !

abretaud:
Cool, thanks @cgirardot!

PR #2342, opened 2019-03-28
anndata: Add tools to manage AnnData datasets

bebatut:
FOR CONTRIBUTOR:

It used the h5ad format which has been added in https://github.com/galaxyproject/galaxy/pull/7268 (so it will fail with the current used version of Galaxy).


tools/anndata/import.xml
```diff
@@ -0,0 +1,234 @@
...
+&&
+cp '$in.matrix' mtx/matrix.mtx.gz
+&&
+cp '$in.tenx.features' mtx/features.tsv.gz
+&&
+cp '$in.tenx.barcodes' mtx/barcodes.tsv.gz
+    #end if
+#else if $in.adata_format == 'umi_tools'
+&&
+gunzip '$in.input'
```
bgruening: in.input is just a tabular isn't it? Do you want to zip it instead?

tools/anndata/import.xml
```diff
@@ -0,0 +1,234 @@
+<tool id="anndata_import" name="Import AnnData" version="@VERSION@+@GALAXY_VERSION@">
+    <description>from different format</description>
+    <macros>
+        <import>macros.xml</import>
+        <xml name="params_10x">
+            <param name="barcodes" type="data" format="tabular" label="Barcodes"/>
```
bgruening: How does this file looks like? Two columns, or one? If one maybe tabular and txt can work both?

tools/anndata/import.xml
```diff
@@ -0,0 +1,234 @@
...
+                <param name="first_column_names" value="true"/>
+            </conditional>
+            <assert_stdout>
+                <has_text_matching expression="ad.read_csv"/>
+                <has_text_matching expression="delimiter='\\t'"/>
+                <has_text_matching expression="first_column_names=True"/>
+            </assert_stdout>
+            <output name="anndata" value="import.tsv.h5ad" ftype="h5ad" compare="sim_size"/>
+        </test>
+        <!--<test>
```
bgruening: Should we include these tests before merging?

tools/anndata/manipulate.xml
```diff
@@ -0,0 +1,175 @@
+<tool id="anndata_manipulate" name="Manipulate AnnData" version="@VERSION@+@GALAXY_VERSION@">
+    <description>object</description>
+    <macros>
+        <import>macros.xml</import>
+        <xml name="param_join">
+            <param name="join" type="text" value="-" label="The connecting string between name and integer"/>
```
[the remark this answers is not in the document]
bebatut: I replaced it by a select with some possible characters

tools/anndata/import.xml
bebatut: I added txt also here

tools/anndata/import.xml
bebatut: It would be nice. But we need https://github.com/galaxyproject/galaxy/pull/7569 then

bebatut:
I expanded the help section also
Thanks @bgruening for the comments

bgruening:
Awesome!

PR #2352, opened 2019-04-02
Flash p gz bug

hepcat72:
FOR CONTRIBUTOR:


tools/flash/flash.xml
```diff
@@ -22,7 +22,7 @@
                     '$layout.reads.forward' '$layout.reads.reverse'
                     #set $input = $layout.reads.forward
                 #end if
-                #if $input.is_of_type('fastqsanger'):
+                #if $input.is_of_type('fastqsanger', 'fastqsanger.gz'):
```
mvdbeek: This translates to an `isinstance` check, so just `fastqsanger` should be enough here.

hepcat72: I did have just fastqsanger in the version Lance merged yesterday, but it was still supplying "-p 64”. After I added fastqsanger.gz, it supplies “-p 33”.

hepcat72: OK, yeah.  So I just tried changing it back to `#if $input.is_of_type('fastqsanger'):`, and it supplied `-p 64`:

```
flash --threads=${GALAXY_SLOTS:-1} -m 10 -M 65 -x 0.25  '/private/var/folders/8d/z84mms353qb2n0r3n2jhww0m0000gq/T/tmpKO5mvt/files/000/dataset_27.dat' '/private/var/folders/8d/z84mms353qb2n0r3n2jhww0m0000gq/T/tmpKO5mvt/files/000/dataset_28.dat' -p 64  > '/private/var/folders/8d/z84mms353qb2n0r3n2jhww0m0000gq/T/tmpKO5mvt/files/000/dataset_33.dat'
```

And the test fails with:

```
Fatal error: Exit code 1 ()
[FLASH] ERROR: Qual string contains character under phred_offset = 64 (file "/private/var/folders/8d/z84mms353qb2n0r3n2jhww0m0000gq/T/tmpKO5mvt/files/000/dataset_27.dat", near line 5)
[FLASH] ERROR: Qual string contains character under phred_offset = 64 (file "/private/var/folders/8d/z84mms353qb2n0r3n2jhww0m0000gq/T/tmpKO5mvt/files/000/dataset_28.dat", near line 5)
```

So I assume I have something wrong elsewhere?

lparsons:
I believe you need to supply the `-z` flag if you want compressed output (which I personally think it should always output).

hepcat72:
> I believe you need to supply the `-z` flag if you want compressed output (which I personally think it should always output).

I'd be happy if it just supplied the correct value for -p.  Do you have any idea why Marius's suggestion ends up resulting in `-p 64` being supplied?

lparsons:
No, that would require some troubleshooting. However, if you're going through the trouble to update the tool, it would be very helpful to output compressed files.

hepcat72:
Outputting gz would be nice, but it's moot if I can't get -p to set correctly.

hepcat72:
OK.  I suspected this, but I didn't say because I wasn't sure.  It apparently outputs in gz if the input is gz.  I just changed the tests to expect fastqsanger (not fastqsanger.gz) and re-added the 'fastqsanger.gz' in the is_of_type call.  It fails because it's outputting gz:

```
JobOutputsError: Dataset metadata verification for [file_ext] failed, expected [fastqsanger] but found [fastqsanger.gz].
```

So it does output gz.  I just can't get -p 33 when I only have `#if $input.is_of_type('fastqsanger'):` which @mvdbeek says should work...

mvdbeek:
There's no harm in specifying both `fastqsanger` and `fastqsanger.gz`, I'll add that just using fastqsanger doesn't work to the list of things to look into.

lparsons:
@hepcat72 This needs a version bump and then it looks good to me.

hepcat72:
I don't understand why the output test files (e.g. tools/flash/test-data/flash_unmerged_f_out4.fastqsanger) pass the tests, given they're not zipped. Is that a problem or does the test structure unzip the output files in order to compare them?

hepcat72:
Oh, never mind.  My files changed tab was stale.

hepcat72:
I think we're good to go now.

tools/flash/flash.xml
```diff
@@ -146,6 +149,35 @@
             <output name="unmerged_reads_r" file="flash_unmerged_r_out3.fastqillumina" ftype="fastqillumina" />
             <output name="hist" file="flash_hist_out3.tabular" />
         </test>
+        <test>
+            <param name="select_layout" value="individual" />
+            <param name="forward" value="flash_forward_in4.fastqsanger.gz" ftype="fastqsanger.gz" />
+            <param name="reverse" value="flash_reverse_in4.fastqsanger.gz" ftype="fastqsanger.gz" />
+            <param name="generate_histogram" value="false" />
+            <param name="save_log" value="true" />
+            <output name="merged_paired_reads" file="flash_merged_out4.fastqsanger.gz" ftype="fastqsanger.gz" compare="sim_size" />
```
lparsons: Does anyone know what the default `delta` is for `sim_size`? I didn't see it in the docs.

lparsons:
Looks good to me.

tools/flash/flash.xml
hepcat72: Oh right.  I forgot you said a delta should be set...

PR #2355, opened 2019-04-03
Samtools_sort: support for new BAM datatype variants

sveinugu:
I have added "unsorted.bam", "qname_sorted.bam" and "qnamed_input_sorted.bam" as input types to samtools_sort. This is needed to support the following workflow:

1. samtools_fixmate, which outputs "qname_sorted.bam"
2. samtools_sort, to coordinate sort, outputs "bam"
3. samtools_markdup

Currently, samtools_sort requires "bam" as input. so the workflow above is not possible to make.

(Automatically changing the output of samtools_fixmate to "bam" in the workflow triggers indexing, which fails (with a very strange error message) due to the file not being coordinate sorted, so that is not a solution)

FOR CONTRIBUTOR:

### DIFF
--- a/tool_collections/samtools/samtools_sort/samtools_sort.xml
+++ b/tool_collections/samtools/samtools_sort/samtools_sort.xml
@@ -1,4 +1,4 @@
-<tool id="samtools_sort" name="Samtools sort" version="2.0.2">
+<tool id="samtools_sort" name="Samtools sort" version="2.0.3">
     <description>order of storing aligned sequences</description>
     <macros>
         <import>macros.xml</import>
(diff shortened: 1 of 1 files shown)


bernt-matthias:
Interesting. I would have guessed that these data types are automatically accepted because they derive from bam. Would just BamNative work?

sveinugu:
@bernt-matthias Yes you are right. Changing "bam" to "unsorted.bam" (BamNative) also adds support for the other variants. Force pushed commits with this simpler change now.

tool_collections/samtools/samtools_sort/samtools_sort.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="samtools_sort" name="Samtools sort" version="2.0.2">
+<tool profile="18.01" id="samtools_sort" name="Samtools sort" version="2.0.3">
```
mvdbeek:
```suggestion
<tool id="samtools_sort" name="Samtools sort" version="2.0.3" profile="18.01">
```
just a person preference of having the id being the first element.

sveinugu: Yes, changed it myself after 15 secs, but you were quicker! :)

PR #2359, opened 2019-04-05
STAR wrapper update

wm75:
FOR CONTRIBUTOR:


wm75:
Just a first check to see how the wrapper tests behave.

wm75:
With the above changeset STAR builds tremendously smaller indexes for the tiny test fasta files.

data_managers/data_manager_star_index_builder/data_manager_conf.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <data_managers>
-    <data_manager tool_file="data_manager/rna_star_index_builder.xml" id="rna_star_index_builder" version="0.0.3">
-        <data_table name="rnastar_index2">
+    <data_manager tool_file="data_manager/rna_star_index_builder.xml" id="rna_star_index_builder" version="0.0.6">
+        <data_table name="rnastar_index2_versioned">
```
bgruening: Could we use the old table and add a new column, if this column does not exists or is empty it is implicitly <2.7?

bgruening: But I guess its not worth the effort ...

tools/rgrnastar/macros.xml
```diff
@@ -1,10 +1,39 @@
 <macros>
+    <!-- REMEMBER to bump the version of rna_star_index_builder_data_manager
```
bgruening: Should we maybe use the version of the data manager also from this macro?

tools/rgrnastar/rg_rnaStar.xml
```diff
@@ -89,110 +92,155 @@
...
+        #if 'exclude_unmapped' in $filter_options:
+            $filter_options.remove('exclude_unmapped')
+            --outSAMunmapped None
+        #else:
+            --outSAMunmapped Within
+        #end if
+        #echo ' '.join($filter_options)
+
+        ## Other Filters
+        #if str( $filter.output_params2.output_select2 ) == "yes":
```
bgruening: If you are on it I would transform all these to single quotes.

tools/rgrnastar/rg_rnaStar.xml
```diff
@@ -89,110 +92,155 @@
...
+                --chimScoreJunctionNonGTAG "$algo.params.align.chim_settings.chimScoreJunctionNonGTAG"
+                --chimJunctionOverhangMin "$algo.params.align.chim_settings.chimJunctionOverhangMin"
+                --chimMainSegmentMultNmax "$algo.params.align.chim_settings.chimMainSegmentMultNmax"
+                #if str($chimeric.chimOutType) == "Junctions":
+                    --chimMultimapNmax "$algo.params.align.chim_settings.chimMultimapNmax"
+                #else:
+                    --chimMultimapNmax 0
                 #end if
+                --chimMultimapScoreRange "$algo.params.align.chim_settings.chimMultimapScoreRange"
+                #set $chim_out_junctions_only = $algo.params.align.chim_settings.chimMultimapNmax >= 1
```
bgruening: this is not used, isn't it?

tools/rgrnastar/rg_rnaStar.xml
```diff
@@ -228,175 +276,229 @@
...
-                    <param argument="--limitSjdbInsertNsj" type="integer" min="0" value="1000000" label="Maximum number of inserts to be inserted into the genome on the fly." />
-                </section>
-
-                <conditional name="chim">
-                    <param name="chim_select" type="select" label="Would you like to set chimeric alignment parameters?">
-                        <option value="no" selected="true">No</option>
-                        <option value="yes">Yes</option>
+            <param argument="--outSAMattrIHstart" name="HI_offset" type="select" display="radio"
+            label="HI tag values should be">
+                <option value="1" selected="true">one-based</option>
```
bgruening: two times selected true

tools/rgrnastar/rg_rnaStar.xml
```diff
@@ -228,175 +276,229 @@
...
+                <option value="0" selected="true">zero-based</option>
+            </param>
+            <!-- Using - -outSAMprimaryFlag AllBestScore would cause a
```
bgruening: Should this be reported upstream?

tools/rgrnastar/rg_rnaStar.xml
```diff
@@ -228,175 +276,229 @@
...
+            <param argument="--outSAMmapqUnique" type="integer" value="60" min="30" max="255"
+            label="MAPQ value for unique mappers"
+            help="STAR bases the mapping quality scores of alignment records in its BAM output on the number of alternative mappings for the read. If a read maps to multiple locations on the reference genome, the following MAPQ scoring scheme is
+used: >=5 mappings => MAPQ=0; 3-4 mappings => MAPQ=1; 2 mappings => MAPQ=3. This setting lets you control the MAPQ used for reads mapped to a single location. Set to 255 for compatibility with Cufflinks." />
+        </section>
+        <section name="filter" title="Output filter criteria" expanded="True">
+            <param name="basic_filters" type="select" display="checkboxes" multiple="true" optional="true"
+            label="Exclude the following records from the BAM output">
+                <option value="exclude_unmapped" selected="true">Unmapped reads</option>
+                <!-- <param argument="- -outSAMunmapped" type="boolean" truevalue="Within" falsevalue="None" checked="false" label="Would you like unmapped reads included in the SAM?"/> -->
```
bgruening: remove?

data_managers/data_manager_star_index_builder/data_manager_conf.xml
wm75: wasn't sure whether that would work (filtering for a column that doesn't exist in a record). If it does, I'd prefer that, too.

tools/rgrnastar/rg_rnaStar.xml
wm75: yes. Left over from a failed attempt, thanks!

tools/rgrnastar/rg_rnaStar.xml
wm75: oops, nice catch.

tools/rgrnastar/rg_rnaStar.xml
wm75: sure. forgot to clean up.

tools/rgrnastar/macros.xml
wm75: @bgruening Not sure this would work well. Since `macros.xml` in the data manager is just a symlink to the file here, I think, DM version bumps would then not be automatically recognized as changes to the DM, would they?

wm75:
One thing I noticed during testing: BAM datasets created by STAR seem to use a low compression level (and STAR's --outBAMcompression option does not seem to help). A simple samtools view-based recompression can decrease file sizes by ~ a factor 3.
I guess this isn't crucial, but if somebody wants to investigate this further before the PR gets merged, it would be a nice extra :)

wm75:
@bgruening and others: Here's a simple workaround illustrating the effect of recompression. Look at the failing tests to see how much it shrinks the BAM output sizes.
I still haven't understood why STAR doesn't compress its output properly in the first place, but if you are fine with the workaround I could update the test BAMs to make the failing tests pass. Otherwise I'll just revert the last commit and we let STAR generate the unnecessarily large files as before.

bgruening:
@wm75 I also feel more confident to just use samtools to create the BAM file. So I prefer actually the workaround.

wm75:
https://github.com/galaxyproject/tools-iuc/issues/2428

bgruening:
Damn, this is amazing work! Thanks @wm75!

tools/rgrnastar/rg_rnaStar.xml
wm75: Probably yes. I'll do so when I have the time.

tools/rgrnastar/rg_rnaStar.xml
wm75: I have standardized the use of quotes throughout the rapper now. The ones you commented on were actually all integers and floats so they didn't require any quoting.

data_managers/data_manager_star_index_builder/data_manager_conf.xml
wm75: So, we decided to go for the new table because that hides newly generated indexes from old tool versions that don't know about versioning.

wm75:
I think this is it regarding the regular STAR functionality. The new STARsolo options are still not part of this PR, but since we are planning to implement these in the form of a separate tool wrapper anyway, this should be ok.

bgruening:
I'm happy to merge this one if no one objects. StarSolo would then be next?

bgruening:
ping @Slugger70

PR #2360, opened 2019-04-05
Requested rename made. Fixed typo causing linter to fail

KyleL1998:
FOR CONTRIBUTOR:

Shortened the output names as It was causing problems in some pipelines


bgruening:
@KyleL1998 can you explain what the issue with the output names is? The names should not influence anything and I would favor more descriptive ones.

KyleL1998:
Keeping the name as ‘Simulated of Illumina sequencing of ‘. It makes it so that read lengths are so long that they no longer work with downstream pipelines.

bgruening:
Which tools does not work @KyleL1998 downstream? Maybe this tool should be changed instead?

Takadonet:
@bgruening a good portion of  tools we use cannot handle having '(' ')' or white-spaces in their file name. We are in the process of converting them to use collection identifier instead but this fix allows my users to move forward with their work.

nsoranzo:
I agree with @bgruening, the changes to `tools/art/art_illumina.xml` are a workaround that is not in line with IUC goals of providing "best practice" tools.
If you really needs them, I'd suggest to patch the labels in your installed copy of the tool.

The fix to `tools/art/art_solid.xml` should be merged though.

Sorry for the hindrance.

KyleL1998:
One of those git disasters...

PR #2363, opened 2019-04-08
data_managers: create dirs with 755 instead of 777

bernt-matthias:
FOR CONTRIBUTOR:

### DIFF
--- a/data_managers/data_manager_bowtie2_index_builder/data_manager/bowtie2_index_builder.py
+++ b/data_managers/data_manager_bowtie2_index_builder/data_manager/bowtie2_index_builder.py
@@ -60,7 +60,7 @@ def main():
 
     params = loads( open( filename ).read() )
     target_directory = params[ 'output_data' ][0]['extra_files_path']
-    os.mkdir( target_directory )
+    os.mkdir( target_directory, mode=0o755 )
     data_manager_dict = {}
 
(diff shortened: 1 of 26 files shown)


bernt-matthias:
Maybe one should also post on the admin mailing list that dir permissions of existing data sets should be changed.

bernt-matthias:
Didn't understand the first one. So I fixed the first .. hoping that the first is related .. lets see

nsoranzo:
> Maybe one should also post on the admin mailing list that dir permissions of existing data sets should be changed.

Luckily this shouldn't be an issue if you have a standard umask (`022`).

For [Python >=3.7](https://docs.python.org/3.7/library/os.html#os.makedirs), there is an additional issue: "The mode parameter is passed to `mkdir()` for creating the leaf directory. To set the file permission bits of any newly-created parent directories you can set the umask before invoking `makedirs()`."

bernt-matthias:
> Luckily this shouldn't be an issue if you have a standard umask (022).

@nsoranzo Extremely helpful point. Thanks a lot. Turns out that there are some strange things going on with the ACLs in der directory that contains Galaxy's job directory. Just contacted our admins to understand the problem. 

In this light I guess the PR is obsolete.

bernt-matthias:
will be fixed by https://github.com/galaxyproject/galaxy/pull/8581

Issue #2314, opened 2019-02-26
Bedtools coverage 2.27.1 fails with scientific notation entered on tool form


jennaj:
Tool: bedtools Compute both the depth and breadth of coverage of features in file B on the features in file A (bedtools coverage) (Galaxy Version 2.27.1)

MTS: https://toolshed.g2.bx.psu.edu/repository?repository_id=8d84903cc667dbe7

Coverage values: 
* Default (blank) == pass
* Decimal (0.000000001) == pass
* Scientific notation (1E-9) == fail
* Scientific notation (1E-09) == fail

"Default" for this wrapper translates to "left blank on the tool form". This is not constent with other Bedtools wrappers like MultiCovBed (1E-09 is specifically pre-entered). The form also has the notation in the help written as 1E-9, not 1E-09 -- be good to standardize that across all tools in the repo when this is fixed. 

Error: `***** ERROR: -f must be in the range (0.0, 1.0]. *****`

Test history: https://usegalaxy.org:/u/jen/h/bedtools-coverage 

<img width="286" alt="screen shot 2019-02-25 at 6 31 02 pm" src="https://user-images.githubusercontent.com/8999735/53383195-8fc5e980-392b-11e9-821e-45e3920087e9.png">

jennaj:
Cross-link to update org once fixed/in MTS https://github.com/galaxyproject/usegalaxy-playbook/issues/193

bernt-matthias:
Is the tool actually started? The validator should actually not allow this: 

https://github.com/galaxyproject/tools-iuc/blob/73c6b2baf9dda26c6809a4f36582f7cbdb161ea1/tools/bedtools/coverageBed.xml#L79

Would be also worth to link this with the overlap parameters in the macros that are used in the intersect tool:

https://github.com/galaxyproject/tools-iuc/blob/fc3616bd1463afef9681cd7c431ac29f79e37e74/tools/bedtools/macros.xml#L29

jennaj:
hi @bernt-matthias -- I don't think anyone has done anything yet. The tool version in the MTS, usegalaxy.org, and usegalaxy.eu are still the same as before. Plus, I reran the two failed jobs in the test history, and still don't parse correctly/job fails.

Or, are you stating that Sci notation should not be allowed? Other bedtools functions accept that type of input, so I wasn't sure about your thoughts. Remove the validator (?) Change it (?). You can decide with @bgruening / IUC how to resolve this the best way :)

[...]

PR #2372, opened 2019-04-24
bedtools: code cleanup

bernt-matthias:
fixes https://github.com/galaxyproject/tools-iuc/issues/2314

FOR CONTRIBUTOR:


tools/bedtools/bamToFastq.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="bedtools_bamtofastq" name="bedtools Convert from BAM to FastQ" version="@WRAPPER_VERSION@">
+<tool id="bedtools_bamtofastq" name="Convert from BAM to FastQ" version="@WRAPPER_VERSION@.0">
```
nsoranzo: I think all others have `bedtools` in the name, all or none?

tools/bedtools/bamToFastq.xml
```diff
@@ -32,7 +32,7 @@
             <output name="output" file="bamToFastq_result1.fastq" ftype="fastq" />
         </test>
     </tests>
-<help>
+    <help>
```
nsoranzo: Add `CDATA` while at it?

jennaj:
Hi -- Could we also address this part of the change request? Cosmetic, but will help users.

> "Default" for this wrapper translates to "left blank on the tool form". This is not consistent with other Bedtools wrappers like MultiCovBed (1E-09 is specifically pre-entered). The form also has the notation in the help written as 1E-9, not 1E-09 -- be good to standardize that across all tools in the repo when this is fixed.

tools/bedtools/bamToFastq.xml
jennaj: All do, so this one should also retain it imo

bernt-matthias: will do. also need to remove .0 from the version...

bernt-matthias:
@jennaj in all tools the default for overlaps should now be an empty input field.  Is this fine with you?

tools/bedtools/macros.xml
```diff
@@ -1,11 +1,12 @@
...
     <xml name="requirements">
         <requirements>
-            <requirement type="package" version="@WRAPPER_VERSION@">bedtools</requirement>
+            <requirement type="package" version="@BEDTOOLS_VERSION@">bedtools</requirement>
             <yield/>
         </requirements>
     </xml>
-    <token name="@WRAPPER_VERSION@">2.27.1</token>
+    <token name="@BEDTOOLS_VERSION@">2.27.1</token>
+    <token name="@WRAPPER_VERSION@">@BEDTOOLS_VERSION@.3</token>
```
nsoranzo: Per https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tool-versions , we should use `+galaxyN` for the wrapper-specific versions.

Also I think it's useful to keep the wrapper versioning per-wrapper instead of as a macro, but that would mean change each wrapper version separately. If you like the idea, I can push a commit to your branch to do so.

bernt-matthias: True, but unfortunately one can't switch from `.x` to `+galaxyx+1` since `'+'>'.'` equals `false`.

If you like you may commit a PR - otherwise I would change to a common version. I don't care that much. I just see the the work if we bump the program version (eg to 2.27.2) then we need to touch every wrapper.

tools/bedtools/bamToFastq.xml
bernt-matthias: done

jennaj:
@bernt-matthias Yes, that would be fine. 

Question: Is the help text consistent now? One tool shouldn't state "1E-09" and another "1E-9" in the help/example sections. Couldn't tell from PRs.

tools/bedtools/macros.xml
nsoranzo: > I just see the the work if we bump the program version (eg to 2.27.2) then we need to touch every wrapper.

Yep, true. The advantage though is that Galaxy needs to load fewer versions of the same tool.

tools/bedtools/bamToFastq.xml
```diff
@@ -1,28 +1,28 @@
-<tool id="bedtools_bamtofastq" name="bedtools Convert from BAM to FastQ" version="@WRAPPER_VERSION@.0">
+<tool id="bedtools_bamtofastq" name="bedtools Convert from BAM to FastQ" version="@TOOL_VERSION@">
```
bernt-matthias: I guess we need to leave `.0` here

nsoranzo: @bernt-matthias I don't think so, the `.0` was introduced in one of the previous commits of this same PR, all the changes are cosmetic.

tools/bedtools/coverageBed.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="bedtools_coveragebed" name="bedtools Compute both the depth and breadth of coverage" version="@WRAPPER_VERSION@">
+<tool id="bedtools_coveragebed" name="bedtools Compute both the depth and breadth of coverage" version="@TOOL_VERSION@">
```
bernt-matthias: need to increase tool version

tools/bedtools/intersectBed.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="bedtools_intersectbed" name="bedtools Intersect intervals" version="@WRAPPER_VERSION@">
+<tool id="bedtools_intersectbed" name="bedtools Intersect intervals" version="@TOOL_VERSION@">
```
bernt-matthias: need to increase tool version

tools/bedtools/jaccardBed.xml
```diff
@@ -1,35 +1,33 @@
-<tool id="bedtools_jaccard" name="bedtools JaccardBed" version="@WRAPPER_VERSION@">
+<tool id="bedtools_jaccard" name="bedtools JaccardBed" version="@TOOL_VERSION@">
```
bernt-matthias: increase tool version

tools/bedtools/tagBed.xml
```diff
@@ -1,38 +1,38 @@
-<tool id="bedtools_tagbed" name="bedtools TagBed" version="@WRAPPER_VERSION@">
+<tool id="bedtools_tagbed" name="bedtools TagBed" version="@TOOL_VERSION@">
```
bernt-matthias: increase tool version

tools/bedtools/fisherBed.xml
```diff
@@ -1,37 +1,35 @@
-<tool id="bedtools_fisher" name="bedtools FisherBed" version="@WRAPPER_VERSION@">
+<tool id="bedtools_fisher" name="bedtools FisherBed" version="@TOOL_VERSION@">
```
bernt-matthias: increase tool version

tools/bedtools/macros.xml
```diff
@@ -22,44 +22,55 @@
...
+            <validator type="expression" message="Only values between 0 and 1 allowed">0&lt;=float(value)&lt;=1</validator>
+        </param>
+    </xml>
     <xml name="overlap">
-        <param name="overlap" argument="-f" type="float" value="0.000000001" label="Minimum overlap required as a fraction of A" help="Default is 1E-9, i.e. 1bp."/>
+        <expand macro="fraction" name="overlap" argument="-f" label="Minimum overlap required as a fraction of A" help="Default is 1E-9, i.e. 1bp."/>
     </xml>
     <xml name="overlapB">
-        <param name="overlapB" argument="-F" type="float" value="0.000000001" label="Minimum overlap required as a fraction of A" help="Default is 1E-9, i.e. 1bp."/>
+        <expand macro="fraction" name="overlapB" argument="-F" label="Minimum overlap required as a fraction of A" help="Default is 1E-9, i.e. 1bp."/>
```
bernt-matthias: unify overlap and overlapB macros and check help

tools/bedtools/bamToFastq.xml
bernt-matthias: OK. You are right

nsoranzo:
@bernt-matthias I've finished the cleanup from my part. Also, I've changed the `type` of the `fraction` param to `float`, I think scientific notation is supposed to be allowed in the tool form and we just need to fix https://github.com/galaxyproject/galaxy/issues/7836 .
Let me know if that works for you, otherwise I can revert the changes.

bernt-matthias:
Looks good to me - at least one tool has a test with scientific notation .. so fine from my side. Thanks for the contributions.

bgruening:
@bernt-matthias you are a IUC member, please feel free to merge :)

PR #2385, opened 2019-05-02
VCFFilter: Add an assertion for common incorrect operators.

VJalili:
Operators such as "<=" or ">=" are not accepted by this tool, tool help form clearly says what are the valid operators, but it seems still users provide some unsupported operators.

ping @jennaj

### DIFF
--- a/tool_collections/vcflib/vcffilter/vcffilter.xml
+++ b/tool_collections/vcflib/vcffilter/vcffilter.xml
@@ -35,7 +35,7 @@ input1.vcf.gz
                 <option value="-f">Info filter (-f)</option>
                 <option value="-g">Genotype filter (-g)</option>
             </param>
-            <param name="filter_value" type="text" value="DP &gt; 10" label="Specify filterting value" help="See explanation of filtering options below">
+            <param name="filter_value" type="text" value="DP &gt; 10" label="Specify filtering value" help="See explanation of filtering options below">
                 <sanitizer>
                     <valid initial="string.printable">
                         <remove value="&apos;"/>
@@ -46,7 +46,7 @@ input1.vcf.gz
                         <add source="&quot;" target=""/>
                     </mapping>
                 </sanitizer>
-                <validator type="expression" message="Invalid operator has provided, check the help section for a list of valid operators.">value is not None and not any(x in value for x in ["&lt;=", "=&lt;", "&gt;=", "=&gt;"])</validator>
+                <validator type="expression" message="Invalid operator provided, valid operators are =, !, &lt;, &gt;, |, &amp;.">value is not None and not any(x in value for x in ["&lt;=", "=&lt;", "&gt;=", "=&gt;"])</validator>
             </param>
         </repeat>
         <param name="filter_sites" argument="--filter-sites" type="boolean" truevalue="--filter-sites" falsevalue="" label="Filter entire records, not just alleles"/>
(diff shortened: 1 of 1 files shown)


jennaj:
cc @jmchilton

jennaj:
Hi - Could we update the message a bit?

**Currently is:**

              <validator type="expression" message="Invalid operator has provided, check the help section for a list of valid operators.">value is not None and not any(x in value for x in ["&lt;=", "=&lt;", "&gt;=", "=&gt;"])</validator>

**Suggested change:** Fixes syntax plus tell the user exactly where the problematic term was entered (directs users to the specific entry area with the problematic operator/expression, "named" with the same label user's see on the form).

              <validator type="expression" message="Invalid operator present in at least one filterting value entered. Please check the help section below for a list of valid operators and example filter values.">value is not None and not any(x in value for x in ["&lt;=", "=&lt;", "&gt;=", "=&gt;"])</validator>

jennaj:
Sent this privately, but also need a test for the invert yes/no toggle on tool form. If had to pick one thing, that would be it. We need the tool to fail to get some eyes on how to fix the public servers that have trouble with this functionality. cc @jmchilton 

Scope creep:

And while we are in this tool improving tests, two more:

**Capturing expressions that do not have proper spacing and tossing a meaningful error or just fixing the query would help in a big way**. We get quite a few reports of "errors" when the user enters this:

`DP=10` (no spaces)

Instead of:

`DP = 10` (proper spacing)

Could the tool possibly accept the unspaced input and add in spaces to avoid the whole "toss an error, tell the user how to fix it" route? These expressions all have the same format. If an operator char is somewhere in that line, but not surrounded by a space on each side, how to make it a correctly formatted term is probably something we can do programmatically. 

The worst case is when people enter two terms in the same filter entry box. Not sure how to capture that. But maybe an error. Also, toss an error if `-f` or `-g` is in there. Both are old usage and predictably come up in reported problems.

Something like this is entered:

`-f “QUAL>20” -f “DP>100”`

or even just

` -f “QUAL>20”`

I know we have help for it, but I often end up sending help like this back that fixes the problem. How to do this without the direct Q&A loop?

Screen Shot 2019-05-02 at 4 47 59 PM

Know I am asking for a lot. But this tool is used a lot and these same usage issues keep coming up over and over. Seems worth it.

jennaj:
Broke everything out into distinct tickets (4!!). This tool will be SO perfect soon :)

tool_collections/vcflib/vcffilter/vcffilter.xml
```diff
@@ -46,6 +46,7 @@ input1.vcf.gz
                         <add source="&quot;" target=""/>
                     </mapping>
                 </sanitizer>
+                <validator type="expression" message="Invalid operator has provided, check the help section for a list of valid operators.">value is not None and not any(x in value for x in ["&lt;=", "=&lt;", "&gt;=", "=&gt;"])</validator>
```
mvdbeek:
```suggestion
                <validator type="expression" message="Invalid operator provided, valid operators are =, !, &lt;, &gt;, |, &amp;.">value is not None and not any(x in value for x in ["&lt;=", "=&lt;", "&gt;=", "=&gt;"])</validator>
```

mvdbeek: You could also list the valid operators in the help attribute of the parameter, and could you fix the `filterting` typo in the label ?

jmchilton: Is that right?

mvdbeek: I just copied the operators form the main help text, should be fine I guess ?

VJalili:
Thanks @bgruening

PR #2402, opened 2019-05-07
Loompy

astrovsky01:
FOR CONTRIBUTOR:

Previous PRs separated into multiple. This just adds a converter from tsv to a single-layered loompy file.


nsoranzo:
@astrovsky01 Just an heads-up: we are not using pylint, but flake8 with the flake8-import-order plugin.

astrovsky01:
@nsoranozo that would explain why the errors I'm getting on these commits weren't showing up on my side...

tools/loompy_convert/tsv_to_loompy.py
```diff
@@ -0,0 +1,62 @@
+"""This module converts a tsv file into a binary loom file"""
+
+import argparse
+import os
```
nsoranzo: You need a blank line to separate standard library imports from other libs. We use the `smarkets` style explained in https://github.com/PyCQA/flake8-import-order#styles

astrovsky01:
Thanks @nsoranzo and sorry for the multiple pr attempts

tools/loompy_convert/tsv_to_loompy.py
```diff
@@ -0,0 +1,63 @@
...
+else:
+    ROW_INFO = "Gene"
+if ARGS.COL_ATTR:
+    COL_INFO = ARGS.COL_ATTR
+else:
+    COL_INFO = "Cell"
+DATA = ARGS.FILES
+OUTFILE = ARGS.OUTPUT
+if os.path.isfile(OUTFILE):
+    os.remove(OUTFILE)
```
mtekman: this should have no effect in galaxy, but perhaps it's good that it's there.
I would otherwise throw an error.

tools/loompy_convert/tsv_to_loompy.xml
```diff
@@ -0,0 +1,35 @@
+<tool id="tsv_to_loompy" name="tsv to loompy converter" version="0.1.0">
+    <requirements>
+      <requirement type="package" version="1.9.3">numpy</requirement>
+      <requirement type="package" version="2.0.16">loompy</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+python /Users/alexanderostrovsky/Desktop/galaxy/tools/myTools/tsv_to_loompy.py -f '${data}'
```
mtekman: Woops, this needs to be `$__tool_directory__/` as given [here](https://docs.galaxyproject.org/en/master/dev/schema.html#tool-version-command)

tools/loompy_convert/tsv_to_loompy.xml
```diff
@@ -0,0 +1,35 @@
...
+    #if $rownames == "true":
```
mtekman: When comparing strings, please change these from `$blah` to `'$blah'` or even `str($blah)` in order to shield against code injection.

Alttough because here you are testing for boolean expression, `#if $rownames:` should be enough.

tools/loompy_convert/tsv_to_loompy.xml
```diff
@@ -0,0 +1,35 @@
...
+    #if $row_attr:
+        --row_attr '${row_attr}'
+    #end if
+    #if $col_attr:
+        --col_attr '${col_attr}'
+    #end if
+    -o '${loompyloom}'
+    ]]></command>
+    <inputs>
+      <param name="data" type="data" format="tsv" label="Input tsv file" help="Tsv file with or without row/column names"/>
```
mtekman: Please change this to 4 spaces indent as given by the [Best Practices](https://planemo.readthedocs.io/en/latest/standards/docs/best_practices/tool_xml.html#coding-style)

tools/loompy_convert/tsv_to_loompy.xml
```diff
@@ -0,0 +1,35 @@
+<tool id="tsv_to_loompy" name="tsv to loompy converter" version="0.1.0">
+    <requirements>
+      <requirement type="package" version="1.9.3">numpy</requirement>
+      <requirement type="package" version="2.0.16">loompy</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
```
mtekman: Please include a `<version_command>`  before the command tag, as given [here](https://docs.galaxyproject.org/en/master/dev/schema.html#tool-version-command) that outputs the version of your script.

Then set the version of your tool (not your script) to `<tool id="blah" name="blah" version="script.version.1" >` where you replace script.version with the version of your script, and then add an extra `.1` at the end to refer the tool wrapper version.

tools/loompy_convert/tsv_to_loompy.xml
```diff
@@ -0,0 +1,35 @@
+<tool id="tsv_to_loompy" name="tsv to loompy converter" version="0.1.0">
+    <requirements>
```
mtekman: Please include a brief `<description>` before the requirements

tools/loompy_convert/tsv_to_loompy.xml
```diff
@@ -0,0 +1,35 @@
...
+      <requirement type="package" version="1.9.3">numpy</requirement>
+      <requirement type="package" version="2.0.16">loompy</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+python /Users/alexanderostrovsky/Desktop/galaxy/tools/myTools/tsv_to_loompy.py -f '${data}'
+    #if $rownames == "true":
+        -r
+    #end if
+    #if $colnames == "true":
```
mtekman: same as above

tools/loompy_convert/tsv_to_loompy.xml
```diff
@@ -0,0 +1,35 @@
...
+        --row_attr '${row_attr}'
+    #end if
+    #if $col_attr:
+        --col_attr '${col_attr}'
+    #end if
+    -o '${loompyloom}'
+    ]]></command>
+    <inputs>
+      <param name="data" type="data" format="tsv" label="Input tsv file" help="Tsv file with or without row/column names"/>
+      <param name="rownames" type="boolean" label="Row names present" truevalue="true" falsevalue="false" selected="true"/>
```
mtekman: The `label` tag should go after the `value|truevalue|falsevalue` 

Also, since the `truevalue` and `falsevalue` are the defaults, you can just remove them.

tools/loompy_convert/tsv_to_loompy.xml
```diff
@@ -0,0 +1,35 @@
...
+      <param name="colnames" type="boolean" label="Column names present" truevalue="true" falsevalue="false" selected="true"/>
```
mtekman: same as above

tools/loompy_convert/tsv_to_loompy.xml
```diff
@@ -0,0 +1,35 @@
...
+      <param name="row_attr" type="text" label="Rows represent: (If unspecified, default is 'Genes')" optional="true"/>
+      <param name="col_attr" type="text" label="Columns represent: (If unspecifed, default is 'Cells')" optional="true"/>
```
mtekman: typo 'unspecified'

tools/loompy_convert/tsv_to_loompy.xml
```diff
@@ -0,0 +1,35 @@
...
+    </inputs>
+    <outputs>
+      <data format="loom" name="loompyloom" label="Loompy converted file"/>
+    </outputs>
+    <help><![CDATA[
```
mtekman: We are going to need some tests to prove the tool does what it says it does

I recommend having 3 tests, two of which the use the same file input but with different tool parameters, and another test which uses a different file as input.

Test files should be no bigger than 5MB (ideally < 1MB).

Some good examples can be found [here](https://docs.galaxyproject.org/en/master/dev/schema.html#tool-tests-test)

tools/loompy_convert/tsv_to_loompy.xml
```diff
@@ -0,0 +1,35 @@
...
+Converts tsv file to a loompy file
+    ]]></help>
```
mtekman: The help text might need fleshing out too.... maybe talk about motivation for the tool, what use cases it serves, maybe a bit about the file formats?

tools/loompy_convert/tsv_to_loompy.py
```diff
@@ -0,0 +1,63 @@
+"""This module converts a tsv file into a binary loom file"""
+
```
mtekman: Please include a VERSION string somewhere

tools/loompy_convert/tsv_to_loompy.py
```diff
@@ -0,0 +1,63 @@
...
+import loompy
+import numpy as np
+
+PARSER = argparse.ArgumentParser(description="Import tsv files")
+PARSER.add_argument('-f', '--FILES', help="Tsv file to import")
+PARSER.add_argument('-r', '--ROWNAMES', action='store_true', help="tsv file contains row names")
+PARSER.add_argument('-c', '--COLNAMES', action='store_true', help="tsv file contains column names")
+PARSER.add_argument('-o', '--OUTPUT', help="Output file path")
+PARSER.add_argument('--ROW_ATTR', help="Specifies the title of the row attributes for loompy")
+PARSER.add_argument('--COL_ATTR', help="Specifies the title of the column attributes for loompy")
```
mtekman: Please add a VERSION argument

tools/loompy_convert/colsrows.tsv
```diff
@@ -0,0 +1,10 @@
+"PC2"	"PC3"	"PC4"	"PC5"	"PC6"	"PC7"	"PC8"	"PC9"	"PC10"	"PC11"	"PC12"	"PC13"	"PC14"	"PC15"	"PC16"	"PC17"	"PC18"	"PC19"	"PC20"	"PC21"	"PC22"	"PC23"	"PC24"	"PC25"	"PC26"	"PC27"	"PC28"	"PC29"	"PC30"	"PC31"	"PC32"	"PC33"	"PC34"	"PC35"	"PC36"	"PC37"	"PC38"	"PC39"	"PC40"	"PC41"	"PC42"	"PC43"	"PC44"	"PC45"	"PC46"	"PC47"	"PC48"	"PC49"	"PC50"	"PC51"	"PC52"	"PC53"	"PC54"	"PC55"	"PC56"	"PC57"	"PC58"	"PC59"	"PC60"	"PC61"	"PC62"	"PC63"	"PC64"	"PC65"	"PC66"	"PC67"	"PC68"	"PC69"	"PC70"	"PC71"	"PC72"	"PC73"	"PC74"	"PC75"	"PC76"	"PC77"	"PC78"	"PC79"	"PC80"	"PC81"	"PC82"	"PC83"	"PC84"	"PC85"	"PC86"	"PC87"	"PC88"	"PC89"	"PC90"	"PC91"	"PC92"	"PC93"	"PC94"	"PC95"	"PC96"	"PC97"	"PC98"	"PC99"	"PC100"
```
mtekman: I just did a diff, is it normal that only the headers are different?

tools/loompy_convert/tsv_to_loompy.py
```diff
@@ -0,0 +1,64 @@
...
+PARSER.add_argument('-v', '--VERSION', action='version', version='%(prog)s 0.1.0', help="Displays tool version")
```
mtekman: If the script version is 0.1.0, then the first tool wrapper version must be 0.1.0.1

tools/loompy_convert/tsv_to_loompy.xml
```diff
@@ -0,0 +1,54 @@
+<tool id="tsv_to_loompy" name="tsv to loompy converter" version="0.1.1">
+    <requirements>
+        <requirement type="package" version="1.9.3">numpy</requirement>
+        <requirement type="package" version="2.0.16">loompy</requirement>
+    </requirements>
+    <version_command>python '$__tool_directory__/tsv_to_loompy.py'</version_command>
```
mtekman: Does the version command need a `-v` flag?

Also, we just want the numbering - so can you cut the script name part out (either in the script, or by piping it into `cut -d' ' -f 2`

tools/loompy_convert/tsv_to_loompy.xml
```diff
@@ -0,0 +1,54 @@
...
+        <param name="col_attr" type="text" label="Columns represent: (If unspecified, default is 'Cells')" optional="true"/>
+    </inputs>
+    <outputs>
+        <data format="loom" name="loompyloom" label="Loompy converted file"/>
+    </outputs>
+    <tests>
+        <test>
+            <param name="data" value="cols.tsv"/>
+            <param name="colnames" value="true"/>
+            <param name="col_attr" value="Testing"/>
```
mtekman: Your tests all need an `<output name="loompyloom" value="some.output.file" />`

Also please put all your test data into a subdirectory called `test-data/`. You don't need to reference this subdirectory in your tests, this is just one of the default folders it looks in

tools/loompy_convert/tsv_to_loompy.xml
```diff
@@ -0,0 +1,54 @@
...
+        </test>
+        <test>
+            <param name="data" value="colsrows.tsv"/>
+            <param name="colnames" value="true"/>
+            <param nam="rownames" value="true"/>
+            <param name="row_attr" value="Testing"/>
```
mtekman: as above

tools/loompy_convert/tsv_to_loompy.xml
```diff
@@ -0,0 +1,54 @@
...
+        </test>
+        <test>
+            <param name="data" value="cols.tsv"/>
+            <param name="col_attr" value="Testing"/>
+            <param name="row_attr" value="Testing2"/>
```
mtekman: as above

tools/loompy_convert/colsrows.tsv
astrovsky01: The headers are different because of the structure of the file. When no row names exist but column names exist, you have a column name for every column. When there are row names, the first column is the row names, and therefore don't need a column name

tools/loompy_convert/tsv_to_loompy.xml
```diff
@@ -0,0 +1,57 @@
...
+    </inputs>
+    <outputs>
+        <data format="loom" name="loompyloom" label="Loompy converted file"/>
+    </outputs>
+    <tests>
+        <test>
+            <param name="data" value="cols.tsv"/>
+            <param name="colnames" value="true"/>
+            <param name="col_attr" value="Testing"/>
+            <output name="loompyloom" file="loompyloom"/>
```
mtekman: This needs to be `<output name="loompyloom" value="an.actual.output.file.in.your.test.data" />`

tools/loompy_convert/tsv_to_loompy.xml
```diff
@@ -0,0 +1,57 @@
...
+        </test>
+        <test>
+            <param name="data" value="colsrows.tsv"/>
+            <param name="colnames" value="true"/>
+            <param nam="rownames" value="true"/>
+            <param name="row_attr" value="Testing"/>
+            <output name="loompyloom" file="loompyloom"/>
```
mtekman: Would these two tests provide the exact same output loom file?

tools/loompy_convert/tsv_to_loompy.xml
```diff
@@ -0,0 +1,57 @@
...
+        </test>
+        <test>
+            <param name="data" value="cols.tsv"/>
+            <param name="col_attr" value="Testing"/>
+            <param name="row_attr" value="Testing2"/>
+            <output name="loompyloom" file="loompyloom"/>
```
mtekman: as above

tools/loompy_convert/tsv_to_loompy.xml
astrovsky01: No, one would have both column and row names built in, the other wouldn't. This would mean that the first column for the second test is 0,1,2,3..., while the first column for the first test are the "labels" which are just the numbers of the first column in the tsv file.

tools/loompy_convert/tsv_to_loompy.xml
astrovsky01: This labels the columns and rows, different from there being column and row names. This is a specific attribute for loom files. In this, all columns are "testing" while all rows are "testing 2". I do need to add in colnames based on the file, but they are separate pieces of data creating different outputs.

tools/loompy_convert/tsv_to_loompy.xml
```diff
@@ -0,0 +1,58 @@
+<tool id="tsv_to_loompy" name="tsv to loompy converter" version="0.1.0.1">
+    <description>Converts tsv file to a loom file. Loom files are a form of hd5 format</description>
+    <requirements>
+        <requirement type="package" version="1.9.3">numpy</requirement>
+        <requirement type="package" version="2.0.16">loompy</requirement>
+    </requirements>
+    <version_command>python '$__tool_directory__/tsv_to_loompy.py' --VERSION | cut -d' ' -f 2'</version_command>
```
mtekman: the last `'` at the end there gives an unexpected EOF in your outputs I think.

PR #2406, opened 2019-05-09
Alternate fastqc reversal

mvdbeek:
ping @wm75, this only reverts the merge commits. The update then only needs to revert the revert and bump the version

FOR CONTRIBUTOR:


wm75:
hmm, except for the wording I'm not seeing a difference between the two versions? Don't both ways make the exact same changes? Anyway, I'm fine with yours, too :)

mvdbeek:
It's the same changes, I just ran `git revert -m 1 <merge commit>` individually. I guess it doesn't mater too much.

nsoranzo:
One test is failing.

mvdbeek:
Yeah, but I don't think we should fix it. I'd suggest merging this PR, upload manually and open a PR that reverts this PR. Does that seem like a good idea ?

nsoranzo:
If we fix the test, use the normal TS upload and then revert this, won't we reach the same outcome?

mvdbeek:
Same outcome, but we'll have to resolve the conflict.

nsoranzo:
If we revert the commit merging this PR, there shouldn't be any conflict, I think.

wm75:
I think the real question is: are we sure that this is correctly reverting the wrapper.
I tested one thing which makes me worried: I ran the inputs of the failing test through fastqc on main and on EU. Result:
- on main, I'm getting a file of the exact same size as the html output expected by the test
- on EU, I'm getting a file of the exact same size as during the failing test

but main has the version we are trying to revert to!

nsoranzo:
You can run `planemo test --failed --update_test_data` and see if the changes to `test-data/fastqc_report2.html` make sense.

wm75:
well, I'm always getting the file size expected by the test and generated by main when I test locally. This doesn't seem to depend on the fastqc or wrapper version. I've inspected the html reports generated by various sources and they all *look* the same. @bgruening suggested that the size differences may reflect different image encoding by different library versions and that sounds like a good explanation to me.

If you want the failing test to pass, why not change the comparison to the same `lines_diff` as in the other tests. It's pointless because the html is mostly on a single line, but at least it's consistently pointless ;)

wm75:
LGTM. @bgruening once merged and in MTS we can reinstall the 0.72 version on eu.

nsoranzo:
Uploaded to TTS and MTS.

PR #2410, opened 2019-05-10
Revert fastqc reversal, bump fastqc version

mvdbeek:


nsoranzo:
Tests are failing unfortunately.

wm75:
This looks good, doesn't it?

PR #2418, opened 2019-05-15
RFC: remove TTS deploy

bgruening:
The ToolShed was again full today, this will help a little bit to not pollute the TTS unnecessarily. MTS and TTS are on the same hard drive afaik.

Is there any important use-case where we would need the TTS?

ping @galaxyproject/iuc


martenson:
I appreciate the intent, but given that IUC repos are well-designed and small I don't think this is worth doing.

nsoranzo:
I think deploying to the TTS is still useful to detect errors in upcoming ToolShed releases, since TTS is usually updated in advance w.r.t. the MTS.

PR #2421, opened 2019-05-17
Seqtk sample low ram

bwlang:
FOR CONTRIBUTOR:


bgruening:
I'm +0 on this. I don't think the average user can make this decision and it should be made by the admin.

bwlang:
@bgruening 
I don't know a way to toggle an option in a tool from an admin's perspective.  One could allocate more RAM to seqk sample jobs, but what about really big samplings e.g. you want 100M reads? most cluster nodes would not have enough ram to do that without 2-pass mode.

Any suggestions for better documentation of the option and when to turn it on?

bernt-matthias:
I would like to give the user the ability to change this parameter as long as it is documented properly. We might hide it in an "advanced options" section.

bgruening:
> but what about really big samplings e.g. you want 100M reads?

How likely is that?

We could also use the 2-pass by default and just accept slower runtime.

> I would like to give the user the ability to change this parameter as long as it is documented properly. We might hide it in an "advanced options" section.

Just saying this makes it for an admin way harder to predict the runtime resources of the tool and a potential dynamic-job-destination also more complicated.
I think it's better to have one way and then adopt the admin side.

bwlang:
@bgruening 
 I see your perspective here... but I'm of two minds about it. 

@natefoo 
I added this because some jobs were failing on usegalaxy.org .. maybe you have an opinion about this?

Would you rather have 2x IO + RAM ~= f(num_reads) or 1x IO + RAM ~= f(num_reads)*read_len*2 ?

natefoo:
Longer is generally better in our case if we can avoid allocating too much memory. Alternatively, can you make use of `$GALAXY_MEMORY_MB`? I guess you don't know how much memory you need until you have run through the reads once anyway?

jennaj:
Users are running into memory problems with seqtk tools at galaxy main (org), too. Send in reports via our bug clicks/internal mailing list. @natefoo already knows/is cc'd on one example. However addressed, good with me. 

People have somewhat small datasets that are still failing at the org server. The example I shared with @natefoo was already a subsample out of a larger original file. It is not clear what the max fastq dataset size/read count can be for seqtk tools (or is same for all, some handle more data, some don't, et cetera). They are trying to convert an F/R interlaced fastq to two fastqs (F + R). Kind of important, can't use many tools with interlaced fastq at all. Others fail with really large interlaced. Need to subsample & de-interlace, both are a bit easier to do with seqtk functions.

bwlang:
How about this plan:
We switch the default to use less RAM (2-pass mode)
and add an advanced section that allows users to switch to single pass mode if their datasets are small.
I'll do that unless I hear otherwise.

mvdbeek:
> that allows users to switch to single pass mode if their datasets are small.

but then there's no advantage since it'll be fast either way. I think we should just invariably use the 2 pass mode.

bwlang:
IO on my local cluster is at a premium, and we have lots of RAM on cluster nodes so I want 1-pass mode.  I accept that this may be niche - hopefully it's enough to default to 2-pass mode and hide the 1-pass mode in advanced.

bgruening:
@bwlang how about adding those parameters as hidden params and admins can overwrite those via dynamic destinations?

bernt-matthias:
@bgruening this sounds cool, didn't know that this is possible. Do you have an example of such a dynamic destination? Since I plan to switch to dynamic destinations in the near future I would be interested to learn about this.

bgruening:
@bernt-matthias have a look here: https://docs.galaxyproject.org/en/latest/admin/jobs.html#dynamic-destination-mapping

jennaj:
@bwlang Does this apply to all or more seqtk tools or just this one?

We saw memory issues some of the others. It is not necessarily a complete list, just what was reported, but I can share which those were. Maybe EU team/server also noticed seqtk tools failing for memory.... Lmk.

Also, will there be a tool version bump or how will we be able to know when the MTS is updated and we can upgrade Main?

Thanks!!

MTS repo for reference: https://toolshed.g2.bx.psu.edu/view/iuc/seqtk/a09586d5149a

Screen Shot 2019-06-18 at 12 37 57 PM
Screen Shot 2019-06-18 at 12 37 47 PM

bwlang:
I think just seqtk sample is affected since that's the only tool that tries to hold the entire set of reads in memory.

bgruening:
Oh we forgot to talk about this one during GCC. How do we all feel about it :)

mvdbeek:
I'm in favor of 2 pass mode only.

mvdbeek:
I guess ultimately having this in the advanced section is fine, nobody except for me was opposed to this, so I'll merge this.

bgruening:
> I'm in favor of 2 pass mode only.

Me too Marius, it's easier to estimate for an admin and less complicated for a user. Let see what the consequences are in real life :)

mvdbeek:
I guess it's simple, if you use it without the roper quote it'll fail, which I think is fine given the warning in the help text.

bwlang:
Sorry I never got to @bgruening 's hidden param suggestion. I think that might have been best. However, the suboptimality of having the advanced option is not among the most important issues facing galaxy right now, so not worth opening back up.

mvdbeek:
Anything specific ?

PR #2429, opened 2019-05-27
fix bug where 'Per-Base Alignment Quality' select param which fails cmd

Takadonet:
fix bug where 'Per-Base Alignment Quality' select param where put 'None' when no option was selected causing command to fail.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/bcftools/bcftools_mpileup.xml
+++ b/tools/bcftools/bcftools_mpileup.xml
@@ -251,7 +251,7 @@ ${pasted_data}
 --no-BAQ; BAQ is the Phred-scaled probability of a read base being misaligned. Applying this option greatly helps to reduce false SNPs caused by misalignments.
 --redo-BAQ; ignore existing BQ tags
                         </help>
-			<option value="">Default</option>
+                        <option value="">Default</option>
                         <option value="--no-BAQ">disable BAQ (per-Base Alignment Quality) (no-BAQ)</option>
                         <option value="--redo-BAQ">recalculate BAQ on the fly, ignore existing BQs (redo-BAQ)</option>
                     </param>
--- a/tools/bcftools/bcftools_mpileup.xml
+++ b/tools/bcftools/bcftools_mpileup.xml
@@ -1,5 +1,5 @@
 <?xml version='1.0' encoding='utf-8'?>
-<tool name="bcftools @EXECUTABLE@" id="bcftools_@EXECUTABLE@" version="@TOOL_VERSION@">
+<tool name="bcftools @EXECUTABLE@" id="bcftools_@EXECUTABLE@" version="@TOOL_VERSION@+galaxy1">
     <description>Generate VCF or BCF containing genotype likelihoods for one or multiple alignment (BAM or CRAM) files</description>
     <macros>
         <token name="@EXECUTABLE@">mpileup</token>
@@ -249,11 +249,12 @@ ${pasted_data}
                     <option value="adjust">Set base and mapping quality options</option>
                 </param>
                 <when value="adjust">
-                    <param name="baq" type="select" optional="true" label="per-Base Alignment Quality">
+                    <param name="baq" type="select" label="per-Base Alignment Quality">
                         <help>
 --no-BAQ; BAQ is the Phred-scaled probability of a read base being misaligned. Applying this option greatly helps to reduce false SNPs caused by misalignments.
 --redo-BAQ; ignore existing BQ tags
                         </help>
+			<option value="">Default</option>
                         <option value="--no-BAQ">disable BAQ (per-Base Alignment Quality) (no-BAQ)</option>
                         <option value="--redo-BAQ">recalculate BAQ on the fly, ignore existing BQs (redo-BAQ)</option>
                     </param>
--- a/tools/bcftools/bcftools_mpileup.xml
+++ b/tools/bcftools/bcftools_mpileup.xml
@@ -109,10 +109,7 @@ bcftools @EXECUTABLE@
 -d "${section.max_reads_per_bam}"
 ${section.skip_anomalous_read_pairs}
 #if str( $section.quality.quality_settings ) == "adjust":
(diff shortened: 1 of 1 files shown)


tools/bcftools/bcftools_mpileup.xml
```diff
@@ -109,7 +109,10 @@ bcftools @EXECUTABLE@
 -d "${section.max_reads_per_bam}"
 ${section.skip_anomalous_read_pairs}
 #if str( $section.quality.quality_settings ) == "adjust":
-    $section.quality.baq
+    #if str( $section.quality.baq ) != "None":
```
bgruening: I think `#if $section.quality.baq:` should also work. Do you have time to include a test case for this?

Takadonet: For sure!

Takadonet:
@bgruening Changes are done. No idea where to update the wrapper version number.

bgruening:
@Takadonet in this file here: https://github.com/phac-nml/tools-iuc/blob/bug_fix_bcftools_mpileup/tools/bcftools/macros.xml

@nsoranzo so you think a version bump is needed?

nsoranzo:
For the proposed changes, I wouldn't bump the tool version.

*But* I think this select param, while technically fine, is a bit unusual:
```xml
                    <param name="baq" type="select" optional="true" label="per-Base Alignment Quality">
                        <help>
--no-BAQ; BAQ is the Phred-scaled probability of a read base being misaligned. Applying this option greatly helps to reduce false SNPs caused by misalignments.
--redo-BAQ; ignore existing BQ tags
                        </help>
                        <option value="--no-BAQ">disable BAQ (per-Base Alignment Quality) (no-BAQ)</option>
                        <option value="--redo-BAQ">recalculate BAQ on the fly, ignore existing BQs (redo-BAQ)</option>
</param>
```

We don't have guidelines on this, but we normally implement this as a non optional select with a default "empty" option (explicit better than implicit):
```xml
                    <param name="baq" type="select" label="Per-Base Alignment Quality">
                        <help>
--no-BAQ; BAQ is the Phred-scaled probability of a read base being misaligned. Applying this option greatly helps to reduce false SNPs caused by misalignments.
--redo-BAQ; ignore existing BQ tags
                        </help>
                        <option value="">Default</option>
                        <option value="--no-BAQ">Disable BAQ (per-Base Alignment Quality) (--no-BAQ)</option>
                        <option value="--redo-BAQ">Recalculate BAQ on the fly, ignore existing BQs (--redo-BAQ)</option>
</param>
```

If you switch to this, you can remove the changes to the command but you would need to increase the tool version as follows:

```xml
<tool name="bcftools @EXECUTABLE@" id="bcftools_@EXECUTABLE@" version="@TOOL_VERSION@+galaxy1">
```

Takadonet:
> For the proposed changes, I wouldn't bump the tool version.
> 
> _But_ I think this select param, while technically fine, is a bit unusual:
> 
> ```
>                     <param name="baq" type="select" optional="true" label="per-Base Alignment Quality">
>                         <help>
> --no-BAQ; BAQ is the Phred-scaled probability of a read base being misaligned. Applying this option greatly helps to reduce false SNPs caused by misalignments.
> --redo-BAQ; ignore existing BQ tags
>                         </help>
>                         <option value="--no-BAQ">disable BAQ (per-Base Alignment Quality) (no-BAQ)</option>
>                         <option value="--redo-BAQ">recalculate BAQ on the fly, ignore existing BQs (redo-BAQ)</option>
> </param>
> ```
> 
> We don't have guidelines on this, but we normally implement this as a non optional select with a default "empty" option (explicit better than implicit):
> 
> ```
>                     <param name="baq" type="select" label="Per-Base Alignment Quality">
>                         <help>
> --no-BAQ; BAQ is the Phred-scaled probability of a read base being misaligned. Applying this option greatly helps to reduce false SNPs caused by misalignments.
> --redo-BAQ; ignore existing BQ tags
>                         </help>
>                         <option value="">Default</option>
>                         <option value="--no-BAQ">Disable BAQ (per-Base Alignment Quality) (--no-BAQ)</option>
>                         <option value="--redo-BAQ">Recalculate BAQ on the fly, ignore existing BQs (--redo-BAQ)</option>
> </param>
> ```
> 
> If you switch to this, you can remove the changes to the command but you would need to increase the tool version as follows:
> 
> ```
> <tool name="bcftools @EXECUTABLE@" id="bcftools_@EXECUTABLE@" version="@TOOL_VERSION@+galaxy1">
> ```

Did all changes describe above.

tools/bcftools/bcftools_mpileup.xml
```diff
@@ -246,11 +249,12 @@ ${pasted_data}
                     <option value="adjust">Set base and mapping quality options</option>
                 </param>
                 <when value="adjust">
-                    <param name="baq" type="select" optional="true" label="per-Base Alignment Quality">
+                    <param name="baq" type="select" label="per-Base Alignment Quality">
                         <help>
 --no-BAQ; BAQ is the Phred-scaled probability of a read base being misaligned. Applying this option greatly helps to reduce false SNPs caused by misalignments.
 --redo-BAQ; ignore existing BQ tags
                         </help>
+			<option value="">Default</option>
```
nsoranzo: Indentation

tools/bcftools/bcftools_mpileup.xml
```diff
@@ -109,7 +109,10 @@ bcftools @EXECUTABLE@
 -d "${section.max_reads_per_bam}"
 ${section.skip_anomalous_read_pairs}
 #if str( $section.quality.quality_settings ) == "adjust":
-    $section.quality.baq
+    #if $section.quality.baq:
+        $section.quality.baq
+    #end if
```
nsoranzo: You can restore the simpler previous version now.

Takadonet: Done

PR #2434, opened 2019-05-28
Featurecounts: update to v3.4.1 and change default MQ to same as upstream (0)

mblue9:
* update to v3.4.1 (and update coreutils dep to v8.31)
* change default mapping quality from 12 to 0 to be consistent with the upstream featurecounts tool. As had a user confused by this today when they were comparing results to others using the command-line tool and only then discovered Galaxy featurecounts was using a different default for mapping quality vs command-line featurecounts. Is it better to be consistent with the upstream tool?

FOR CONTRIBUTOR:

### DIFF
--- a/tools/featurecounts/featurecounts.xml
+++ b/tools/featurecounts/featurecounts.xml
@@ -297,10 +297,10 @@
 
             <param name="mapping_quality"
                    type="integer"
-                   value="12"
+                   value="0"
                    argument="-Q"
                    label="Minimum mapping quality per read"
-                   help="The minimum mapping quality score a read must satisfy in order to be counted. For paired-end reads, at least one end should satisfy this criteria. 12 by default." />
+                   help="The minimum mapping quality score a read must satisfy in order to be counted. For paired-end reads, at least one end should satisfy this criteria. 0 by default." />
 
             <conditional name="exon_exon_junction_read_counting_enabled">
                 <param name="count_exon_exon_junction_reads" argument="-J" type="boolean" truevalue="-J" falsevalue=""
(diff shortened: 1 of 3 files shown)


bernt-matthias:
Jep. In my opinion the program defaults should be chosen. 

Maybe consider to add the new options? 
- readShiftType and readShiftSize
- there also seem to be some options in earlier releases that are not wrapped: eg. nonOverlap and nonOverlapFeature

bgruening:
> Is it better to be consistent with the upstream tool?

I would say this depends on how good these defaults are and if this is a well-maintained tool. There are a few tools out there, where the default does make less sense imho. This is not the case here, so :+1: from me.

mblue9:
Ok great, thanks for the opinions @bernt-matthias and @bgruening. I can add those other options when I get a chance, might take me a few days to get to it.

mblue9:
Hi @bernt-matthias I was having a look and I could add these options you suggest but was wondering... I personally don't think we should add in every option just because a tool provides it as it increases the maintenance burden (esp. if we need to add tests aswell) but if a user (or voice or users like yourself) says it's useful then am open to adding them - do you think these are useful options and worth adding now?

bernt-matthias:
I understand the arguments: 

- Indeed I guess most users will go for the defaults for most parameters anyway. 
- The UI gets cluttered by many parameters used by nobody

My points in favor of adding them are as follows: 

- I'm just thinking of (advanced) users who would be disappointed if options are missing. 
- The burden to add the parameters gets larger and larger with every new version 

I'm fine with leaving them out.. completely up to you.

bgruening:
I will merge this for now. We can add the params in a new PR then but the version upgrade is good I think.

mblue9:
Thanks for the merge @bgruening and @bernt-matthias thanks for your good points! featurecounts is used a lot I think (it's one of our most used tools anyway) so maybe it would be good to add those parameters when there's time.

PR #2439, opened 2019-05-31
--window_size does not exists in stacks_population

bgruening:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/stacks/stacks_populations.xml
+++ b/tools/stacks/stacks_populations.xml
@@ -441,26 +441,26 @@ Output from denovo_map or ref_map
 
 **Output files**
 
-- XXX.tags.tsv file::
+- XXX.tags.tsv file:
 
 See `Stacks output description <http://catchenlab.life.illinois.edu/stacks/manual/#files>`_
 
 Notes: For the tags file, each stack will start in the file with a consensus sequence for the entire stack followed by the flags for that stack. Then, each individual read that was merged into that stack will follow. The next stack will start with another consensus sequence.
 
 
-- XXX.snps.tsv file::
+- XXX.snps.tsv file:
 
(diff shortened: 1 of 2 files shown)


bernt-matthias:
Good one. 

Tests are failing because Galaxy gets the data set names from the relative paths given in the tests .. which makes liking fail. I had suggested a fix https://github.com/galaxyproject/galaxy/pull/7372 here which was not merged for various (valid) reasons. 

Currently I don't see a simple solution for this problem... but I'm wondering when this broke. 

I'm quite sure that the tests build on each other and assume that the data set names are the sample names (ie the file's basenames). So if you replace the `/` in the file name its not the sample anymore. 

I guess on the long run it might be a good idea to have an `id="..."` attribute for all collections and data sets allowing to name everything explicitly...

nsoranzo:
> Currently I don't see a simple solution for this problem... but I'm wondering when this broke.

@bernt-matthias @mvdbeek I've bisected this issue and it was broken by https://github.com/galaxyproject/galaxy/commit/6791f3b99905ac0db0a8e2c23f38a4bce8b1d3c8

bernt-matthias:
@nsoranzo I'm not understanding what is done in the commit. my guess would have been that some `basename` calls need to be distributed in this this function

https://github.com/galaxyproject/galaxy/blob/6791f3b99905ac0db0a8e2c23f38a4bce8b1d3c8/lib/galaxy/tools/verify/interactor.py#L280

nsoranzo:
Rebased, it's green now.

bernt-matthias:
I integrated this here: https://github.com/galaxyproject/tools-iuc/pull/4194

bernt-matthias:
My bad... this is not stacks2...

bernt-matthias:
@bgruening good to merge?

bernt-matthias (approved):

PR #2474, opened 2019-06-26
Add STARSolo

mtekman:
FOR CONTRIBUTOR:

This is a wrapper of the STARSolo capabilities of the new version of RNA STAR which provides a single-click solution to 10X datasets following the CellRanger model. 

This PR is forked off the much-needed [facelift PR](https://github.com/galaxyproject/tools-iuc/pull/2359/) given by @wm75


bgruening:
> AssertionError: Test file (737K-august-2016.txt) is missing. If you use planemo try --update_test_data to generate one.

Maybe a rebase is needed as well?

tools/rgrnastar/rg_rnaStarSolo.xml
```diff
@@ -0,0 +1,283 @@
...
+                    </param>
+                    <when value="with-gtf">
+                        <expand macro="@SJDBOPTIONS@" optional="false"/>
+                    </when>
+                    <when value="without-gtf" />
+                </conditional>
+            </when>
+        </conditional>
+
+        <section name="solo" title="Star Solo Settings" expanded="true">
```
bgruening: its a little bit strange that the tool is called start solo but only here the "star solo settings" begin ... I would assume all settings are star solo settings :)

wm75: Yes, the idea was to get the tool working, then determine the extent of overlap with STAR and split all that out into macros.

mtekman:
I did not add an `auto_tool_repositories` or a `suite` to the shed because STARSolo is still very much a STAR tool, and derives all of its components from it, so I thought it would be odd to split these into separate toolshed repos.

bgruening:
During GCC we decided to split them always into separate repos, more for technical reasons. Is this ready for review?

mtekman:
I think so, I added the `auto_tool_repositories` based on the mothur model, but perhaps this is overkill, since there are only two wrappers need here?

tools/rgrnastar/.shed.yml
```diff
@@ -1,15 +1,31 @@
...
-  automated dependency installation by Ross Lazarus.
+  Most of this wrapper was originally written by Jeremy Goecks, with kibbitzing and automated dependency installation by Ross Lazarus.
+
+  The scRNA-seq mapping and quantification component is encapsulated within the STARSolo wrapper.
 homepage_url: https://github.com/alexdobin/STAR
 name: rgrnastar
 owner: iuc
 remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/master/tools/rgrnastar
 type: unrestricted
+auto_tools_repositories:
```
bgruening: The auto one, unfortunately, will not work. The tool id is `rna_star` but the repro was called `rgrnastar`.

I think you need to use something like this: https://github.com/galaxyproject/tools-iuc/pull/2517/files#diff-05cc38f519e7fe8f2effa3958a835c32R10

mtekman: ah good example, I will adopt this

mtekman: @bgruening I changed the shed.yml, would this work?

bgruening:
I uploaded the tools to the test-toolshed and it looks good to me.
@wm75 ready when you are.

bgruening:
I updated Star to the latest version which was released yesterday.

PR #2455, opened 2019-06-18
Update freebayes to 1.3.1

davebx:
As far as I have been able to determine, the freebayes and leftalign wrappers are completely compatible with version 1.3.1 as-is. I have confirmed that tests pass with updated test data.

FOR CONTRIBUTOR:
* [x] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [x] - License permits unrestricted use (educational + commercial)
* [x] - This PR updates an existing tool or tool collection

PR #2487, opened 2019-07-09
Update FreeBayes to version 1.3.1

davebx:
FOR CONTRIBUTOR:

This replaces #2455


tools/freebayes/macros.xml
```diff
@@ -1,5 +1,5 @@
 <macros>
-    <token name="@DEPENDENCY_VERSION@">1.1.0.46</token>
+    <token name="@DEPENDENCY_VERSION@">1.3.1</token>
     <xml name="requirements">
         <requirements>
             <requirement type="package" version="@DEPENDENCY_VERSION@">freebayes</requirement>
```
wm75: would it be possible to bump also the samtools requirement to 1.9? Not sure, but samtools seems to be used only for ref indexing and getting the input bam header so should be unproblematic

wm75: also, remove the `-0` from the freebayes and leftalign tool wrapper versions. Just use `@DEPENDENCY_VERSION@` there.

tools/freebayes/macros.xml
```diff
@@ -44,6 +44,8 @@
         </conditional>
     </xml>
     <xml name="par_min_cov">
-        <param name="min_coverage" argument="--coverage" type="integer" value="0" label="Require at least this coverage to process a site" />
+        <param name="min_coverage" argument="--min-coverage" type="integer" value="0" label="Require at least this coverage to process a site" />
+        <param name="limit_coverage" argument="--limit-coverage" type="integer" value="0" label="Downsample per-sample coverage to this level if greater than this coverage" />
+        <param name="skip_coverage" argument="--skip-coverage" type="integer" value="0" label="Skip processing of alignments overlapping positions with coverage greater than this" />
```
wm75: I don't think this is the best place to add in the new options. The way the wrapper structures things, this ties `--limit-coverage` and `--skip-coverage` to using standard filters unless you are willing to configure all options.
If you do this, then I would also add a boolean for turning on/off standard filters.

davebx:
Thanks for the suggestions, @wm75, I'll be making those changes soon.

tools/freebayes/macros.xml
davebx: I like this suggestion, but I want to implement it in a way that doesn't hinder rerunning jobs, so I'll have to give it a little more thought when I have time (which will be this week, of course)

tools/freebayes/macros.xml
davebx: Done and done, thanks.

tools/freebayes/macros.xml
wm75: Good point! Take your time.

bernt-matthias:
Did you check that all changes mentioned in https://github.com/ekg/freebayes/releases are reflected? In particular the changed defaults in version 1.2.

nsoranzo:
Can you please rebase the pull request? Unfortunately we had to do a force-push on the master branch.

Please let us know if you have issues and we can do it for you.

bgruening:
@bernt-matthias do you think we can merge this?
@davebx do you plan to work on this and adapt the PR?

bernt-matthias:
The defaults that changed in 1.2.0 should be changed: https://github.com/ekg/freebayes/releases/tag/v1.2.0

davebx:
Replaced with #2525

PR #2489, opened 2019-07-15
[Experimental] Refactor data manager logic

wm75:
FOR CONTRIBUTOR:

I'm creating this PR **to get feedback** on an experiment I've been doing after reference data discussions during and after GCC. It's an illustration (based on the bowtie2 data manager) of an idea to separate concerns between data_managers' .py and .xml files.

**What it does and why**:

- After refactoring, the python code no longer contains any Galaxy-specific logic, but concerns itself exclusively with building the requested index.
- The wrapper xml calls the python code through a very much simplified cli - basically only passing a metadata.yaml file. With that the python code could be called with [idc](https://github.com/galaxyproject/idc)-style metadata (thanks @Slugger70!) to generate the same ref data with and independent of Galaxy.
- The DM python code can update the passed in metadata with information only known to itself (like the command line instruction(s) it ran to obtain/generate the ref data.
@natefoo This first version just turns the updated metadata into a separate dataset, but, alternatively, it could also become one of the extra files produced by the DM, which could be used and stored by Galaxy for things like building vignettes for the ref data, _etc_.
- Hopefully, this also makes the architecture of data managers more similar to regular tools in that it would now be possible to conda-package the data manager .py (complete with unit tests) like a regular tool dependency and have the iuc handle only the wrapper part of it.
- As a next step, I'm envisioning a general DM package (available through conda) that would define a DM base class providing logic for handling the metadata.yaml parsing, _etc._, simplifying and standardizing the task of writing new DMs.

I've created https://testtoolshed.g2.bx.psu.edu/view/wolma/ng_data_manager_bowtie2_index_builder/ if anyone wants to see the refactored DM in action though really all it does right now is replicate the existing DM's action and generate the new metadata dataset :smile:

### DIFF
--- a/data_managers/data_manager_bowtie2_index_builder/data_manager/bowtie2_index_builder.py
+++ b/data_managers/data_manager_bowtie2_index_builder/data_manager/bowtie2_index_builder.py
@@ -2,83 +2,55 @@
 # Dan Blankenberg
 from __future__ import print_function
 
-import optparse
+import argparse
 import os
 import subprocess
 import sys
-from json import dumps, loads
 
-DEFAULT_DATA_TABLE_NAMES = ["bowtie2_indexes"]
+import yaml
 
-
-def get_id_name( params, dbkey, fasta_description=None):
-    # TODO: ensure sequence_id is unique and does not already appear in location file
-    sequence_id = params['param_dict']['sequence_id']
-    if not sequence_id:
-        sequence_id = dbkey
-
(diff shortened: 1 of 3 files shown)


nsoranzo:
Can you please rebase the pull request? Unfortunately we had to do a force-push on the master branch.

Please let us know if you have issues and we can do it for you.

bgruening:
Ping @galaxyproject/iuc and @blankenberg can you please also comment?

PR #2496, opened 2019-07-23
quast: default single core

bernt-matthias:
FOR CONTRIBUTOR:


bgruening:
@bernt-matthias what is the reason here?

bernt-matthias:
On our cluster single core jobs still use 4 cores. But when thinking more about it, this might also be a bug of the job runner... Have to check.

Anyway I consider one cpu always the better default. But this is only my opinion.

PR #2503, opened 2019-07-26
Update pygenometracks to v3.1

lldelisle:
Hi,
I am currently updating the wrapper for pygenometracks to use pygenometracksv3.0.

To make things work, I:
* forced python3.6 (as python3.7 and h5py are incompatible)
* do symlink for hic files to guess the format
* change track=line to track=fill for bedgraph track
* updated the test output
* solved an issue when the boundaries_file where not as much as HiC matrices (but I do not see where the boundary files are used so maybe we should remove them).

To improve the wrapper, I:
* allowed no transformation for hic
* allowed no height for hic track

I would like to:
* allow use reverse orientation
* allow use overlay

If my [PR for gtf](https://github.com/deeptools/pyGenomeTracks/pull/83) is accepted:
* allow input gtf

If my [PR for bed file bed4 bed5](https://github.com/deeptools/pyGenomeTracks/pull/84) is not accepted:
* make a pre-step so it will not fail.


bgruening:
@lldelisle this is super! Thanks a lot for working on this!

It would be wonderful if you could add a few more tests, so we make sure this also keeps working on future versions.

Thanks!!

lldelisle:
Hi,
I think the wrapper is ready.
I tested as much as I could and everything seems to work.
Here are the only options which are possible in the command-line tools which are not possible with galaxy:
1. bedgraph_matrix
  a. pos score in bin
  b. plot horizontal lines
  c. overlay
2. bedgraph
  a. extra 4C
3. bed
  a. interval height (interval_height) but I do not think it is used anywhere
  b. line width
  c. min_value and max_value when score is used.
It is not possible to plot epilogos, links, narrow_peak, domain.

For the moment, the biggest issue for me is that it will fail when bed4, bed5 are used. If my PR is accepted, there will be another version and I would update the wrapper...

Thanks,

gtrichard:
Great job! Thanks for working on this Lucile.

I have one question, why wouldn't you allow `transformation for hic`, I guess you mean log1p ?

Actually it is important to be able to log1p for regular matrices, while it is mandatory to not log1p in the case of log2ratio (hicCompareMatrices) or Pearson correlation matrices plotting.

lldelisle:
No I mean I allow both. And the default is still log1p...
image

gtrichard:
"allowed `no transformation`"

Got it !! Sorry for the misunderstanding

bgruening:
@lldelisle I updated the version in your PR and it passes. Anything you want to add before we release a first version?

lldelisle:
Yes I would like to add the alpha.

lldelisle:
and gtf... It should be quick.

lldelisle:
image
Do you think we could avoid this in the tool version?

lldelisle:
When travis validates, for me it is ready.
(The only thing would be the INFO in the tool version but this is more related to pygenometracks than to the wrapper).

bgruening:
@lldelisle I hope my last commit fixed the INFO issue.

lldelisle:
It should, I will tell you in a minute, thanks to planemo...

lldelisle:
It did not:
image
What galaxy launched:
`pyGenomeTracks --version 2> /dev/null > /tmp/tmpjx48e3ng/job_working_directory/000/3/COMMAND_VERSION 2>&1`
I think it is written in galaxy `2>&1`...

bgruening:
maybe `pyGenomeTracks --version | tail -n 1`

lldelisle:
I will try locally

lldelisle:
image
yes

lldelisle:
I just noticed that it was set python3.6 whereas it should work with the last version of python. I am doing test with planemo, now.

lldelisle:
It fails but it showed me that pygenometracks3.1 with python 3.7 in conda also fails...

bgruening:
Let's keep this for 3.6 only. I have opened https://github.com/deeptools/pyGenomeTracks/pull/94 to follow up on this.

lldelisle:
ok perfect.

lldelisle:
Super many thanks

bgruening:
Thanks to you!!!

lldelisle:
Hi @bgruening,
I just noticed there is a wrapper for hicplotTAD which is different from pygenometracks whereas it is the same program. Should it be possible to have like a symlink?

bgruening:
@lldelisle you mean on Europe Galaxy? hicplotTAD is using pygenometrack under the hood that is true. We could hide one of the tools to make it less confusing. But I need to double-check all our trainings before.

lldelisle:
I did not mean hide it as I think users do not know it is the same and indeed some training use it. I thought something like when you click on it, it brings you to the pygenometracks wrapper...

bgruening:
That is not easily possible at the moment.

lldelisle:
ok, then we should just copy paste I guess...

PR #2505, opened 2019-07-27
add phred-offset to rnaspades

bgruening:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/rnaspades/rnaspades.xml
+++ b/tools/rnaspades/rnaspades.xml
@@ -1,4 +1,4 @@
-<tool id="rnaspades" name="rnaSPAdes" version="3.9.0.1">
+<tool id="rnaspades" name="rnaSPAdes" version="3.9.0.2">
     <description>assembler for RNA-Seq data</description>
     <requirements>
         <requirement type="package" version="3.9.0">spades</requirement>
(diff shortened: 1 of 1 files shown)


tools/rnaspades/rnaspades.xml
```diff
@@ -93,6 +95,10 @@
         </repeat>
         <param optional="true" format="fasta,fastq" label="Trusted contigs" multiple="true" name="trusted_contigs" type="data" />
         <param optional="true" format="fasta,fastq" label="Untrusted contigs" multiple="true" name="untrusted_contigs" type="data" />
+         <param argument="--phred-offset" name="phred_offset" type="select" style="radio" label="Phred-Offset" help="Specify which type of FASTQ format you have. By default fastqsanger is assumed." >
```
mvdbeek: You don't want to just set this based on the data type ? That is the only reason we discriminate between fastqsanger/fastqillumina etc.

bgruening: The datatype is currently only "fastq". I was hopeing for some feedback from @davebx or any other expert with this tool. We would need to remove "fastq" from the datatypes in this case and include fastqsanger and fastqillumina explicitly ... which poeses all the other issues which we already know :)

bgruening: I'm happy to change it.

bernt-matthias:
What will be the value of the data set's `ext` attribute if a fastqsanger or Illumina file is given?

PR #2506, opened 2019-07-28
add moabs

lijinbio:
FOR CONTRIBUTOR:


tools/moabs/macros.xml
```diff
@@ -0,0 +1,18 @@
...
+            <requirement type="package" version="@VERSION@">moabs</requirement>
+            <yield />
+        </requirements>
+    </xml>
+
+    <token name="@VERSION@">1.3.4</token>
+
+    <xml name="citations">
+        <citations>
+						<citation type="doi">10.1186/gb-2014-15-2-r38</citation>
```
bgruening: Indentation should be 4 spaces instead of an TAB.

tools/moabs/moabs.xml
```diff
@@ -0,0 +1,133 @@
...
+			p=4
+
+			[MCOMP]
+			Path=mcomp
+			reference='$input3'
+		</configfile>
+	</configfiles>
+
+	<inputs>
+		<repeat name="g1_fastq" title="Group1: fastq files">
```
bgruening: The repeat if probably not needed you can use multiple="true" in a data parameter.

tools/moabs/moabs.xml
```diff
@@ -0,0 +1,133 @@
...
+				<when value="single">
+					<param name="input" type="data" format="fastq" label="Single reads" />
+				</when>
+				<when value="paired">
+					<param name="input1" type="data" format="fastq" label="Forward reads" />
+					<param name="input2" type="data" format="fastq" label="Reverse reads" />
+				</when>
+			</conditional>
+		</repeat>
+		<param type="data" name="input3" format="fasta" label="Reference FASTA file" help="Can be uploaded in the history" />
```
bgruening: You could be fancy here and add support for cached reference data, see here one example:

https://github.com/galaxyproject/tools-iuc/blob/master/tools/bwa/bwa_macros.xml#L54

With this you support globally installed reference data, from an admin, or locally provided FASTA files from a user.

bgruening: Here is the cheetah counterpart: https://github.com/galaxyproject/tools-iuc/blob/master/tools/bwa/bwa_macros.xml#L23

tools/moabs/moabs.xml
```diff
@@ -0,0 +1,133 @@
...
+	</inputs>
+	<outputs>
+		<data name="output1" format="txt" label="${tool.name} on ${on_string} (.txt)" />
```
bgruening: This looks like a BED file, isn't it? `format="bed"`

tools/moabs/moabs.xml
```diff
@@ -0,0 +1,133 @@
+<tool id="moabs" name="MOABS" profile="16.04" version="@VERSION@">
+	<description>MOdel based Analysis of Bisulfite Sequencing data</description>
+	<macros>
```
bgruening: please use 4 spaces instead of a TAB

tools/moabs/.shed.yml
```diff
@@ -0,0 +1,15 @@
+name: moabs
+owner: lijinbio
```
bgruening: this should be `iuc`

tools/moabs/moabs.xml
```diff
@@ -0,0 +1,133 @@
...
+
+			[MCOMP]
+			Path=mcomp
+			reference='$input3'
+		</configfile>
+	</configfiles>
+
+	<inputs>
+		<repeat name="g1_fastq" title="Group1: fastq files">
+			<conditional name="fastq_input">
```
bgruening: You might want to add support for paired collections, like here:

https://github.com/galaxyproject/tools-iuc/blob/master/tools/bwa/bwa-mem.xml#L114

(interleaved is probably not needed....)

tools/moabs/moabs.xml
```diff
@@ -0,0 +1,133 @@
...
+			s2_r${i+1}_2='$s.fastq_input.input2'
+			#else:
+			s2_r${i+1}='$s.fastq_input.input'
+			#end if
+			#end for
+
+			[TASK]
+			Program=MMAP
+			Label=g1,g2
+			Parallel=NONE
```
bgruening: You can use this, but it should be something like `"\${GALAXY_SLOTS:-1}"` (https://docs.galaxyproject.org/en/latest/dev/schema.html#reserved-variables)

bgruening:
@lijinbio the error what you get is here: https://travis-ci.org/galaxyproject/tools-iuc/jobs/564531462#L2147

> mv: cannot stat 'dmr_M3_g1.G.bed_vs_g2.G.bed.txt.dmr': No such file or directory

It seems the file is not created or is it in a subdirectory?

martenson:
@bgruening when merging we have to make sure to squash the 2M lines test data commit.

bgruening:
I see this here:

> |  Something went wrong. No supported configuration file syntax found at /home/travis/conda/envs/__moabs@1.3.4/lib/site_perl/5.26.2/Config/Simple.pm line 184, <FH> line 6.

You could `echo $cfg_file` at the beginning to debug this locally.

lijinbio:
> I see this here:
> 
> > |  Something went wrong. No supported configuration file syntax found at /home/travis/conda/envs/__moabs@1.3.4/lib/site_perl/5.26.2/Config/Simple.pm line 184,  line 6.
> 
> You could `echo $cfg_file` at the beginning to debug this locally.

Hi Björn, yes. I saw this error. The `$cfg_file` did not correctly recognize input variables. I am debugging this locally using `planemo test`. Thank you.

lijinbio:
> Thanks for your contribution @lijinbio! I have added a few comments inline.
> 
> * the test data is a little bit too big, can you please try to reduce the size to <1MB

Hi Björn,

The genome FASTA file in `test-data` is too big. In fact, any genome file may be very big. May I ask do we have a convention to use a genome file cached somewhere in the test case? Instead of uploading in the `test-data` directory? Thank you.

bgruening:
Have a look at https://github.com/galaxyproject/tools-iuc/blob/master/tools/minimap2/test-data/all_fasta.loc

This is using just the MT genome part to be very small.

lijinbio:
@bgruening Hi Björn, may I ask a question about `\${GALAXY_SLOTS:-4}` within `<configfile>`? 
I want to set the multiple cores using `p=\${GALAXY_SLOTS:-4}`, but `<configfile>` can not recognize its default value as 4 (becomes empty in https://travis-ci.org/galaxyproject/tools-iuc/jobs/566211175#L1729). Without a number assigned to `p`, my PR will not pass the Travis CI build.

lijinbio:
@bgruening Hi Björn, may I ask another question about memory limit in the virtual host?

One component `mcomp` will read a large pre-computed lookup table file (need ~6.8GB memory). The component can pass my local test, but it failed with exit code 35072 (https://travis-ci.org/galaxyproject/tools-iuc/jobs/566218611#L1801). This code indicates `mcomp` exited by running out of the memory. May I ask how to configure the memory usage? So that it can pass the build, and it will correctly run after deployment? Thank you.

bgruening:
> @bgruening Hi Björn, may I ask a question about \${GALAXY_SLOTS:-4} within <configfile>?

urgs, yeah this might not be possible. `\${GALAXY_SLOTS:-4}` is a bash variable and not a cheetah one. You could put a placeholder in your configfile, e.g. `FANCY_PLACEHOLDER` and then `sed 's/FANCY_PLACEHOLDER/\${GALAXY_SLOTS:-4}/'` (syntax needs to be adopted) in your command line section.

> @bgruening Hi Björn, may I ask another question about memory limit in the virtual host?

arg, this is a bummer. I do not think we can get more memory from Travis.
But according to this table you should have 7.5GB https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system

Any change to run in low memory mode?

lijinbio:
> @bgruening Hi Björn, may I ask a question about `\${GALAXY_SLOTS:-4}` within `<configfile>`?
> 
> I want to use multiple cores by `p=\${GALAXY_SLOTS:-4}`, but `<configfile>` can not recognize its default value as 4 (becomes empty in https://travis-ci.org/galaxyproject/tools-iuc/jobs/566211175#L1729). Without a number assigned to `p`, the program will fail the build. Can you help suggest why `<configfile>` can not recognize this global variable? Thank you.

Hi Björn, I realized environmental variables like `$GALAXY_SLOTS` are accessible in the `<command>` tag but not the `<configfile>` tag. I can fix my question by assigning the number of threads in `<command>`.  Please ignore this question. Of cause, it will be very helpful if you can confirm that environmental variables can be used in the `<command>` tag only. Thank you.

lijinbio:
> > @bgruening Hi Björn, may I ask a question about ${GALAXY_SLOTS:-4} within ?
> 
> urgs, yeah this might not be possible. `\${GALAXY_SLOTS:-4}` is a bash variable and not a cheetah one. You could put a placeholder in your configfile, e.g. `FANCY_PLACEHOLDER` and then `sed 's/FANCY_PLACEHOLDER/\${GALAXY_SLOTS:-4}/'` (syntax needs to be adopted) in your command line section.

It is a great solution to use a place holder. Your explanation confirms that bash variables can be accessed in the `<command>` tag but not the `<configfile>` tag. This information solves many mysteries to me. Thank you.

> 
> > @bgruening Hi Björn, may I ask another question about memory limit in the virtual host?
> 
> arg, this is a bummer. I do not think we can get more memory from Travis.
> But according to this table you should have 7.5GB https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system
> 
> Any change to run in low memory mode?

I may think about a parameter to switch off the read of the big file. Yet, the 7.5GB memory may be assigned for the whole virtual host, and one testing program may not be able to have this much memory. Thank you for this information about the 7.5GB memory limitation.

tools/moabs/moabs.xml
```diff
@@ -0,0 +1,186 @@
...
+        <section name="mcomp_advanced" title="Advanced options for MCOMP" expanded="False">
+            <param name="doComp" type="integer" value="1" label="--doComp: Run the comparison or not" help="0: not compare, using comparison result by `-c`; 1: compare"/>
+            <param name="compFile" type="data" format="txt" label="-c: Input comparison results" help="Previously generated comparison file from history"/>
+        </section>
+    </inputs>
+    <outputs>
+        <data name="output1" format="txt" label="${tool.name} on ${on_string} (.txt)" />
+    </outputs>
+    <tests>
+        <test>
```
bgruening: Can you add a test for the loc file case, the cached genome?
You can find examples in the link I shared before.

tools/moabs/moabs.xml
```diff
@@ -0,0 +1,186 @@
...
+            <output name="output1" file="dmr_M3_g1.G.bed_vs_g2.G.bed.txt.dmr" ftype="txt" lines_diff="1"/>
+        </test>
+    </tests>
+    <help>
+        <![CDATA[
+        **MOABS: MOdel based Analysis of Bisulfite Sequencing data**
+
+        MOABS is a comprehensive, accurate and efficient solution for analysis of large scale base-resolution DNA methylation data, bisulfite sequencing or single molecule direct sequencing.
+
+        MOABS seamlessly integrates alignment, methylation calling, identification of hypomethylation for one sample and differential methylation for multiple samples, and other downstream analysis.
```
bgruening: I have the feeling you need to explain the output a little bit more, maybe a small example?

tools/moabs/moabs.xml
lijinbio: Sure. I will add the test case for the cached genome.

tools/moabs/moabs.xml
lijinbio: Yes. I will add an example to demonstrate the output format.

bgruening:
It is green @lijinbio! :1st_place_medal:

lijinbio:
> It is green @lijinbio! 🥇

Cheers. Many thanks for your great help. I am finalizing the test case. I will submit the commit with changes soon. Thank you.

lijinbio:
@bgruening Hi Björn, I think I have addressed all your comments. The latest commit has also passed the Travis CI build. Can you help review and merge my PR? Thank you.

Best regards,
Jin

tools/moabs/.shed.yml
```diff
@@ -0,0 +1,15 @@
...
+homepage_url: https://github.com/sunnyisgalaxy/moabs
+long_description: |
+  MOABS is a comprehensive, accurate and efficient solution for analysis of large scale base-resolution DNA methylation data, bisulfite sequencing or single molecule direct sequencing.
+
+  https://github.com/sunnyisgalaxy/moabs
+remote_repository_url: https://github.com/sunnyisgalaxy/moabs
+type: unrestricted
+categories:
+- Epigenetics
+auto_tool_repositories:
```
bgruening: Strictly, if you do not plan to add more moabs tools into this folder you do not need this auto_tool_repositories. Its only useful for tool-collections.

tools/moabs/moabs.xml
```diff
@@ -0,0 +1,305 @@
...
+        </repeat>
+        <repeat name="g2_fastq" title="Group2: fastq files">
+            <conditional name="fastq_input">
+                <param name="fastq_input_selector" type="select" label="Single or Paired-end reads" help="Select between paired and single end data">
+                    <option value="paired">Paired</option>
+                    <option value="single">Single</option>
+                    <option value="paired_collection">Paired Collection</option>
+                </param>
+                <when value="paired">
+                    <param name="fastq_input1" type="data" format="fastqsanger,fastqsanger.gz,fasta" label="Select first set of reads" help="Specify dataset with forward reads"/>
```
bgruening: will fasta.gz also work?

lijinbio: No. Input reads currently support fastq and fastq.gz only. I have removed the fasta format here.

tools/moabs/.shed.yml
lijinbio: Sure. I have deleted this `auto_tool_repositories` block.

lijinbio:
> The help is great now! Thanks for adding additional tests. Just one small comment inline and the last question. Is there any change to reduce the fastq file sizes? Can you only take half of the reads per fastq file?
> 
> This looks super good now! Thanks a lot!

@bgruening Hi Björn, I have reduced the test files under 1MB (~800KB). Thank you for your comments. Without your comments, I was not thinking about creating a small test case like this. The latest commit has finally passed the build. Can you again help review and merge my PR? Thank you.

Best regards,
Jin

tools/moabs/moabs.xml
```diff
@@ -0,0 +1,307 @@
...
+                <when value="single">
+                    <param name="fastq_input1" type="data" format="fastqsanger,fastqsanger.gz" label="Select fastq dataset" help="Specify dataset with single reads"/>
+                </when>
+                <when value="paired_collection">
+                    <param name="fastq_input1" format="fastqsanger,fastqsanger.gz" type="data_collection" collection_type="paired" label="Select a paired collection" help="See help section for an explanation of dataset collections"/>
+                </when>
+            </conditional>
+        </repeat>
+        <section name="bsmap_advanced" title="Advanced options for BSMAP" expanded="False">
+            <param argument="-v" type="text" value="0.08" label="Mismatch rate or number" help="[0,1]: mismatch rate w.r.t to the read length; otherwise: maximum number of mismatches allowed on a read"/>
```
bgruening: Should that be a float param? If you can accept float and int, depending on the meaning ... you could model it maybe with two params:

* (optional="true") mismatch rate in float and you can add min="0" and max="1"
* (optional="true") mismatch number in integer and you can add min="0"

tools/moabs/moabs.xml
```diff
@@ -0,0 +1,307 @@
...
+            <param argument="-n" type="integer" value="1" label="Set mapping strand information" help="0: only map to 2 forward strands, i.e. BSW(++) and BSC(-+); 1: map SE or PE reads to all 4 strands, i.e. ++, +-, -+, --"/>
```
bgruening: This could be a nice boolean parameter. `<param argument="-n" type="boolean" truevalue="-n 0" falsevalue="-n 1"` or just `0` and `1`. It makes in the UI more sense to have it as boolean.

tools/moabs/moabs.xml
```diff
@@ -0,0 +1,307 @@
...
+            <param argument="-r" type="integer" value="0" label="How to report repeat hits" help="0=none(unique hit/pair); 1=random one; 2=all(slow)"/>
```
bgruening: maybe a select parameter? This will prevent user entering a 4

tools/moabs/moabs.xml
```diff
@@ -0,0 +1,307 @@
...
+        </section>
+        <section name="mcomp_advanced" title="Advanced options for MCOMP" expanded="False">
```
bgruening: Question, you could to this as a conditional, remove the section and use `Run the comparison or not` as the conditional. If the user open the conditional he/she can add a new input file.

One other question, where is this comparison file generated and how can I get it?

tools/moabs/moabs.xml
```diff
@@ -0,0 +1,307 @@
...
+    </macros>
+    <expand macro="requirements"/>
+    <command detect_errors="exit_code">
+        <![CDATA[
+        #if $mcomp_advanced.doComp == 0:
+            cp -f '$mcomp_advanced.compFile' comp.g1.vs.g2.txt &&
+        #end if
+        moabs -v 1 --def MMAP.p="\${GALAXY_SLOTS:-4}" --def MCALL.p="\${GALAXY_SLOTS:-4}" --def MCOMP.p="\${GALAXY_SLOTS:-4}" --cf '$cfg_file' &&
+        head -n 3 comp.g1.vs.g2.txt &&
+        cat moabs.mcall.g1.log moabs.mcall.g2.log moabs.mcomp.comp.g1.vs.g2.txt.log &&
```
bgruening: Is this debug output still needed?

tools/moabs/moabs.xml
lijinbio: @bgruening Hi Björn, I appreciate your detailed inspection. Your comments are very helpful to make MOABS much easier to use. I will address them one by one. As for the second question here, the comparison file is one intermediate result by comparing methylation ratios between the two groups in `mcomp`. In fact, I may add interfaces to access additional intermediate results like below.

- To download BAM files with mapping results, these BAM files can be directly used for other downstream analysis.

- To display bigwig tracks of some region files, such as differential methylated CpG sites or regions. These visualization tracks are very helpful for users to understand methylation globally or locally for some specific regions. For now, I am not sure what is the best practice for tracks visualization in Galaxy.

- To report some statistics in a table (it can be a `txt` output, but not sure if Galaxy supports a better result presentation for a table, like a table in a webpage?). These basic statistics may help show some insight about the data themselves.

Therefore, besides the current DMR region result, these intermediate results are also important for users to understand methylation knowledge derived from their BS-Seq data. May I ask do we have a convention or some good practices to report these results? Thank you.

Best regards,
Jin

bgruening: > the comparison file is one intermediate result by comparing methylation ratios between the two groups in mcomp

But this file is never exposed, right?

> To display bigwig tracks of some region files, such as differential methylated CpG sites or regions. These visualization tracks are very helpful for users to understand methylation globally or locally for some specific regions.

That would be super cool, as those files can be visualised with a bunch of different tools (deeptools, pyGenometrack, trackster, JBrowse, IGV ....)

> To report some statistics in a table (it can be a txt output, but not sure if Galaxy supports a better result presentation for a table, like a table in a webpage?). These basic statistics may help show some insight about the data themselves.

Just use `tabular` and Galaxy as a slightly better view.

> besides the current DMR region result

This is a BED file right? You should also put it as `format="bed"`

> May I ask do we have a convention or some good practices to report these results? Thank you.

You can add a select box where the user can choose between different additional outputs. Use `type="select" multiple="true" optional="true"`.

Then in the output section use `<filter>` to only add outputs if they were requested by the user.

lijinbio: > > the comparison file is one intermediate result by comparing methylation ratios between the two groups in mcomp
> 
> But this file is never exposed, right?
Yes, it is not exposed for now. I will add this comparison file in the output.
> 
> > To display bigwig tracks of some region files, such as differential methylated CpG sites or regions. These visualization tracks are very helpful for users to understand methylation globally or locally for some specific regions.
> 
> That would be super cool, as those files can be visualised with a bunch of different tools (deeptools, pyGenometrack, trackster, JBrowse, IGV ....)

It is great to have so many tools here. I will examine them and embed one in the Galaxy interface of MOABS.

> 
> > To report some statistics in a table (it can be a txt output, but not sure if Galaxy supports a better result presentation for a table, like a table in a webpage?). These basic statistics may help show some insight about the data themselves.
> 
> Just use `tabular` and Galaxy as a slightly better view.

Great. I know this `tabular` tag now. Thank you for letting me know.

> 
> > besides the current DMR region result
> 
> This is a BED file right? You should also put it as `format="bed"`

It is not a BED file strictly. For a BED file, "The order of the optional fields is binding: lower-numbered fields must always be populated if higher-numbered fields are used."
https://genome.ucsc.edu/FAQ/FAQformat.html#format1

Although our current DMR region has three required BED fields, the rest of the five fields do not strictly match the additional fields defined in the above like. Such as the 4th field should be the `name`, but the 4th column in our result is the methylation state of the region. Therefore, I stick with the `txt` format but not `BED` in `moabs.xml`.

> 
> > May I ask do we have a convention or some good practices to report these results? Thank you.
> 
> You can add a select box where the user can choose between different additional outputs. Use `type="select" multiple="true" optional="true"`.
> 
> Then in the output section use `<filter>` to only add outputs if they were requested by the user.

It is great to know these features in Galaxy. I will learn these features and add them to MOABS soon. Many thanks for your comprehensive suggestion and information.

bgruening: > It is not a BED file strictly. 

But then mark it as `tabular` or better `interval`. :)

lijinbio: @bgruening Hi Björn, may I ask a question about a Galaxy API?

To implement a UCSC genome browser tracks using generated files in Galaxy, Galaxy provides a public URL to the dataset in a history. For example, we can access a bigwig file in Galaxy using the URL `https://usegalaxy.org/datasets/bbd44e69cb8906b5ef5ae64707e108c3/display`. Using this URL, I can construct a customized track to this bigwig file in the UCSC genome browser. However, in order to get the public URL, I need to know the "History Content ID"  for the dataset, such as `bbd44e69cb8906b5ef5ae64707e108c3` here for my bigwig file. Do you mind helping suggest how could I get these "History Content IDs" for output files in tool's `*.xml`? Do we have some special variables indicating history IDs for `<output>` data files? Thank you.

Best regards,
Jin

bgruening: I would not implement this into this tool, just output the bigwig file. Galaxy - the framwork - or a special other tool will deal with the display later one. Do not put too many functions in one tool. The Galaxy idea is to compose things. Your tool outputs standard files and other tools consume them and to other cool things with it.

lijinbio: I find a solution to get the history ID now. With bigwig files, I will simply form a URL directing the tracks in one output file. This function is not enough for a separate tool but will make a quick visualization much easier for users, and this kind of quick visualization can be a mandatory demand now. Of cause, I will also output bigwig files as one select option, so that other more sophisticated tools can further process them.

bgruening: I need to see this commit, but here in IUC we don't recommend to do this. These tools should not rely on Galaxy internals. And every bigwig file is already visualizable in IUC by default I think. The same for bigwig. You really don't need to do this :)

Lets leave this out and if you are not satisfied, let's add this functionality to a second version, ok?

lijinbio: Sure. I am finalizing the testing of the first version without the visualization of bigwig files. I will submit the commit soon. Thank you very much for your information and suggestion.

lijinbio: @bgruening Hi Björn, I think I have addressed all your comments, and I have submitted the commit. It has passed the Travis build. Can you help review this version? Thank you.

Best regards,
Jin

tools/moabs/moabs.xml
```diff
@@ -0,0 +1,447 @@
...
+                </when>
+            </conditional>
+        </section>
+        <param name="output_selector" type="select" multiple="true" optional="true" label="Select output files" help="">
+            <option value="1"> dmr_M3_g1.G.bed_vs_g2.G.bed.txt.dmr </option>
+            <option value="2"> comp.g1.vs.g2.txt </option>
+            <option value="3"> BAM files </option>
+            <option value="4"> Bigwig files </option>
+        </param>
+        <param name="genomebuild" type="select" optional="true" label="Genome build for generating Bigwig files" help="Genome build for chromosome sizes">
```
bgruening: I don't think this is needed. And its a little unfortunate that it would only work for those 4 genomes. In addition you already have the fasta file in your reference_source.

Keep it simple we do have a tool https://usegalaxy.eu/root?tool_id=wig_to_bigWig that converts bedgraph to bigwig so you don't need to add this complexity.

lijinbio: It is great to have a tool to convert bedgraph to bigwig in Galaxy. I am removing this part and output its raw bed files.

bgruening: You mean bedgraph?

lijinbio: Not strictly a bed or bedgraph. One intermediate result is methylation counts in CpG sites of two groups. The file format is an interval file with 15 columns of necessary raw counts. These raw counts are important for downstream DMR calling and visualization. What I did was to extract first 4 columns to form a bedgraph, then convert to bigwig, but it turns out not necessary. So I will simply output these raw ratios and counts so that users can perform downstream DMR calling or visualization on their own using other sophisticated tools.

bgruening: If you like you can output the bedgraph file, but this complicated dance with bigwig is not needed I think.

lijinbio: You are right. If we have a well-tested tool to do the bigwig conversion, it is far more efficient to use them directly instead of implementing a limited function. I agree to keep the current tool simple. As for the bedgraph file, I think the bedgraph format will lose some information compared to the raw interval file, such as the total number of Cs and the number of methylated Cs will be missed but they are also important. So I think as a tool itself, it is better to keep all the information. Downstream bedgraph conversion can be easily done by selecting interesting columns.

lijinbio:
@bgruening Hi Björn, bigwig function has been removed, and the commit has passed the Travis build. Can you help review my PR again? Thank you very much.

Best regards,
Jin

lijinbio:
> @bgruening Hi Björn, bigwig function has been removed, and the commit has passed the Travis build. Can you help review my PR again? Thank you very much.
> 
> Best regards,
> Jin

@bgruening Hi Björn, can you help review my last commit? Thank you.

bgruening:
Urgs, sorry, dropped the ball here :( 

Sorry! It's good to go I think. Thanks for your contribution @lijinbio!

lijinbio:
@bgruening Hi Björn, many thanks for your help with merging my PR for MOABS. I am glad to have MOABS deposited in this repository maintained by IUC.

May I ask how could we use MOABS in a UseGalaxy server? I tried to search MOABS in usegalaxy.eu and usegalaxy.org, yet it showed `Could not find tool with id 'moabs'`. Do you mind suggesting the process that I can follow to make MOABS available in a UseGalaxy server? so that MOABS can be used by biologists who are interested in DNA methylation? Thank you very much.

Best regards,
Jin

bgruening:
Do add it to usegalaxy.eu you need to add it to https://github.com/usegalaxy-eu/usegalaxy-eu-tools/blob/master/tools_iuc.yaml ... 

Hope that helps!
Bjoern

bgruening:
Ah, these tools will be installed weekly, so every Saturday morning.

lijinbio:
@bgruening Hi Bjoern, thank you very much for your direction. I have just submitted the PR to add MOABS to usegalaxy.eu (https://github.com/usegalaxy-eu/usegalaxy-eu-tools/pull/222). Can you help review my PR? Many thanks for your help.

Best regards,
Jin

> Do add it to usegalaxy.eu you need to add it to https://github.com/usegalaxy-eu/usegalaxy-eu-tools/blob/master/tools_iuc.yaml ...
> 
> Hope that helps!
> Bjoern

bgruening:
@lijinbio should be installed enjoy!

lijinbio:
> @lijinbio should be installed enjoy!

@bgruening Hi Bjoern, yes, I saw it in the server. I am advertising this Galaxy web service of MOABS in usegalaxy.eu. Many thanks for all your help from the very beginning. I also appreciate the computational resource provided by usegalaxy.eu. Thank you.

Issue #2508, opened 2019-07-30
trinotate: wget failed


loraine-gueguen:
wget failed when running trinotate because of not found library. 
Cross-references: https://github.com/bioconda/bioconda-recipes/issues/14288, https://github.com/bioconda/bioconda-recipes/issues/16128

```
Fatal error: Exit code 2 ()
* Running CMD: wget "http://ftp.ebi.ac.uk/pub/databases/uniprot/current_release/knowledgebase/complete/uniprot_sprot.dat.gz"
wget: error while loading shared libraries: libssl.so.1.0.0: cannot open shared object file: No such file or directory
Error, cmd: wget "http://ftp.ebi.ac.uk/pub/databases/uniprot/current_release/knowledgebase/complete/uniprot_sprot.dat.gz" died with ret 32512 No such file or directory at /usr/local/genome2/lib/perl5/Pipeliner.pm line 186.
	Pipeliner::run(Pipeliner=HASH(0x2b9c809208c0)) called at /tmp/mc/envs/mulled-v1-a8259c1b65c5d282750e47963d88f7d939acc8a28c8ab855de159198845685cc/bin/Build_Trinotate_Boilerplate_SQLite_db.pl line 119
```

bgruening:
I will try to fix the old bioconda package as well.

bgruening:
@loraine-gueguen can you please merge https://github.com/bioconda/bioconda-recipes/pull/16705

Than you do not need to disable your tool :)

bgruening:
This should be fixed now, hopefully. But please we need to migrate to the conda-forge version still. Thanks for the report @loraine-gueguen!

PR #2509, opened 2019-07-30
Trinotate: fix wget dependency 

loraine-gueguen:
FOR CONTRIBUTOR:

Fix https://github.com/galaxyproject/tools-iuc/issues/2508
Use conda-forge::wget instead of bioconda::gnu-wget. And update test data.


loraine-gueguen:
I don't know why the tests always stop before the end on Travis (`No output has been received`). They pass with planemo on my local machine.

bernt-matthias:
Ideally this should be added to the requirements of the conda recipe. 

I'm fine with the change, but please add to comment, such that we might remove the dependency in the future (if it found its way in the conda recipe).

loraine-gueguen:
@bernt-matthias I have added wget in the trinotate conda recipe: https://github.com/bioconda/bioconda-recipes/pull/16651. So, I removed the requirement in the xml file.

bernt-matthias:
Few points: 

- the test just takes "to long", on my system it was 45min. is there a way to possibly speed this up. It seems that the perl script called in the beginning needs most of the time.
- the tests failed because of differences with the created files (trinotate_annotation_report_minus_tmhmm_signalp.xls) .. maybe the reference data changed since yesterday?
- I'm wondering if `Build_Trinotate_Boilerplate_SQLite_db.pl Trinotate` (which seems to download a lot of data bases) is a case for a data manager (this also seems to need most of the time). Somehow it seems to make less sense if all these DBs are downloaded (etc) separately for each job compared to doing it once (or regularly) in a data manager. The actual tool could then work with a copy of the provided sqlite db.

loraine-gueguen:
Thank you for having a look on it.
The tests should probably be changed to be less strict and not fail each time the dataset reference has changed.
I don't know data manager (and the trinotate tool) very well, but maybe it could be the best long-term solution. What do you think @lecorguille @abretaud  ?
Unfortunately I don't have time now (holiday) and I guess I will simply deactivate this tool version in Galaxy at the moment...

PR #2515, opened 2019-07-31
hisat2: correctly assign datatype to output Fastq datasets when inputs are compressed

pjbriggs:
FOR CONTRIBUTOR:

PR which addresses an issue in the `hisat2` tool: when the input Fastqs are in a compressed format (e.g. `fastqsanger.gz` or `fastqsanger.bz`), the tool runs the `hisat2` program so that it produces output Fastqs with the same compression type as the input, but then incorrectly assigned the datatypes of those datasets to `fastqsanger`. This had implications for downstream tools, causing them to fail.

With this patch the correct datatype should be assigned to the output Fastq datasets based on that of the inputs (based on code which performs a similar job in the `bowtie2` tool).

### DIFF
--- a/tools/hisat2/hisat2.xml
+++ b/tools/hisat2/hisat2.xml
@@ -581,25 +581,121 @@ hisat2
         <!-- Unaligned fastq (L) -->
         <data name="output_unaligned_reads_l" format="fastqsanger" label="${tool.name} on ${on_string}: unaligned reads (L)">
             <filter>adv['output_options']['output_options_selector'] == "advanced" and adv['output_options']['unaligned_file'] is True</filter>
-            <expand macro="dbKeyActions" />
+            <actions>
+              <conditional name="library.type">
+                <when value="single">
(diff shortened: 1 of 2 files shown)


bgruening:
@pjbriggs I think the dbkeyactions are already actions: https://github.com/galaxyproject/tools-iuc/blob/b5944211a94ae2c83d4cc6d4f2d2d3a7649c7334/tools/hisat2/hisat2_macros.xml#L32

So you need to remove the action__s__ from the macro.

pjbriggs:
@bgruening thanks - I just figured it out now myself! (Should have linted before initial push...)

mvdbeek:
Thanks @pjbriggs! I changed the indentation to 4 spaces and bumped the tool version, thanks for fixing this!

bgruening:
Thanks!

pjbriggs:
Thanks @mvdbeek & @bgruening

PR #2516, opened 2019-08-01
Stacks2: keep logs even in case of error

abretaud:
FOR CONTRIBUTOR:

I had a problem with a denovo_map job that failed, but I had no useful log to understand why. So here's a patch to make sure the log is moved to the output_log dataset even in case of error, making it much easier to understand why it failed


abretaud:
Ah yes you're right, trap is called even if the command works
Making a symlink would feel simpler yes, but I think galaxy will not like to have a dataset_xxx.dat pointing to a file in a (deleted) working dir, unless there is some magic at the end of jobs to take care of that?

bernt-matthias:
If I'm not completely wrong then $output_log (i.e. dataset_xxx.dat) is a real file and `stacks_outputs/gstacks.log` is the symbolic link. So I think it should work.

abretaud:
Ah, yes of course, this way it should work, will do the change (I had not seen your edit)

bernt-matthias:
Hmm (un)fortunate coincidence. We just fixed the assert_command handling in tests https://github.com/galaxyproject/galaxy/pull/8395 .. before that this tag was just ignored. 

Just shows how important this tag is. So now we have the chance to fix some bugs..

bernt-matthias:
Tried to fix some of the bugs. So far only in the tests.

abretaud:
Ok, I thought I was getting insane with all these new errors yesterday :)
I'm fixing some of them right now, the procrad error is strange though, can't reproduce it locally

abretaud:
Ok, now there's only one procrad test failing, but it passes on my laptop so... would you have time to look at it?

bernt-matthias:
Just ran into another problem on my local planemo: https://github.com/galaxyproject/planemo/issues/947 ... I remembered that this was solved already.

bernt-matthias:
I think I got it. Some boolean parameters still had value true/false in tests instead of the corresponding true/falsevalues.

While fixing I noticed that the I often have long options with underscores, which have been changed in 2.3 to dashes. The change log for stacks2.3 says

```
Feature: Backwards-compatibly switched to hyphens in command line options (underscores remain accepted where they previously were).
```

Currently this still works, but might be confusing for the user since the documentation mentions parameters with dashes and the help of the parameters mentions them with underscore. 

Seems worth changing, but seems a bit of work (maybe reduced by https://github.com/galaxyproject/galaxy/pull/8409).

bernt-matthias:
Hmm. Still no success. The difference seems to be that the test here filters failed failed illumina reads: 

```
 7000 failed Illumina filtered reads (50.0%)
```

in my local test the number is 0

bernt-matthias:
Added the log file to the test output ... maybe this helps to understand.

One more thing: I force pushed to the branch .. please take care.

bernt-matthias:
Thanks.

Issue #2312, opened 2019-02-25
make sure most used tools support compressed input if possible


martenson:
The first column is a number of times the given tool triggered implicit conversion (decompression) of a compressed fastq dataset in last ~6 months on Main.

If we could ensure the tools can work with compressed fastq files we would save a lot of quota for the users of these tools.

```
 4056 | toolshed.g2.bx.psu.edu/repos/nate/trinity_psc/trinity_psc/0.0.1
 3850 | toolshed.g2.bx.psu.edu/repos/lparsons/fastq_join/fastq_join/1.1.2-806.1
 3600 | toolshed.g2.bx.psu.edu/repos/devteam/tophat2/tophat2/2.1.1
 3329 | toolshed.g2.bx.psu.edu/repos/iuc/pear/iuc_pear/0.9.6.1
 3079 | toolshed.g2.bx.psu.edu/repos/devteam/fastq_groomer/fastq_groomer/1.0.4
 3074 | toolshed.g2.bx.psu.edu/repos/devteam/bwa_wrappers/bwa_wrapper/1.2.3
 2827 | toolshed.g2.bx.psu.edu/repos/devteam/bowtie_wrappers/bowtie_wrapper/1.1.2
 1884 | toolshed.g2.bx.psu.edu/repos/devteam/fastx_barcode_splitter/cshl_fastx_barcode_splitter/1.0.1
 1536 | toolshed.g2.bx.psu.edu/repos/iuc/mothur_make_contigs/mothur_make_contigs/1.39.5.0
 1228 | toolshed.g2.bx.psu.edu/repos/devteam/tophat2/tophat2/0.6
 1226 | toolshed.g2.bx.psu.edu/repos/iuc/seqtk/seqtk_sample/1.2.0
  952 | toolshed.g2.bx.psu.edu/repos/devteam/fastq_trimmer_by_quality/fastq_quality_trimmer/1.0.0
  772 | toolshed.g2.bx.psu.edu/repos/iuc/flash/flash/1.2.11.3
  588 | toolshed.g2.bx.psu.edu/repos/devteam/tophat2/tophat2/2.1.0
  490 | toolshed.g2.bx.psu.edu/repos/devteam/kraken/kraken/1.2.3
  436 | toolshed.g2.bx.psu.edu/repos/devteam/fastq_trimmer/fastq_trimmer/1.0.0
[...]

PR #2520, opened 2019-08-02
Trinity: add support for compressed input (gz/bz2)

pjbriggs:
FOR CONTRIBUTOR:

PR to enable the `trinity` tool to handle compressed input Fastqs (gzipped or bz2).

This is implemented by making decompressed copies of the inputs in the working directory prior to running `trinity`. It's quite ugly and doesn't seem very efficient as the decompression operations are performed serially, but does seem to work.

Finally: IMHO although this tool is not installed on main, this change does seem related to issue #2312.


abretaud:
Thanks for taking the time to write it, but I think we don't really need (un)fortunately: Galaxy is already able to automatically extract gzipped fastq when you use one in a param that only accepts unzipped fastq

pjbriggs:
Hi @abretaud we are running an older version of Galaxy (18.09) and Trinity seemed to be having problems dealing with the output from HISAT2 when this was a list of `fastqsanger.gz` pairs - I could see Galaxy uncompressing the inputs, but then the compressed Fastqs were still being passed to Trinity (causing it to fail). Hence this update.

Do you know whether the handling of a list of pairs of compressed datasets works in a later Galaxy version?

pjbriggs:
After a quick check, it looks like the implicit decompression works on Galaxy EU (which is running 19.05 and has Trinity 2.8.3), so it looks like the failure to properly propagate the implicitly uncompressed Fastqs is a bug in Galaxy 18.09 (possibly fixed by this PR in 19.05: https://github.com/galaxyproject/galaxy/pull/7683?).

I still need the Trinity tool changes locally as a workaround until we update our Galaxy version, but I'm closing the PR as it's not needed generally.

Issue #2513, opened 2019-07-31
README.rst for Trinity gives incorrect instructions for setting memory limit


pjbriggs:
The `README.rst` file for the `trinity` tool (https://github.com/galaxyproject/tools-iuc/tree/master/tools/trinity) has incorrect instructions for setting the memory consumption limit.

The instructions indicate that values of the form `1G`, `50G` etc (i.e. a number followed by `G`) can be assigned to the `GALAXY_MEMORY_MB` environment limits, but doing this results in the following message appearing in the tool stdout:

    .../job_scripts/597189: line 53: 50G: value too great for base (error token is "50G")

and the default value of `1G` is supplied to `trinity` on the command line.

It appears that the memory limit actually needs to be assigned as a number of MB without the units, e.g. for 50G limit this would be (50*1024 =) 51200:

    <env id="GALAXY_MEMORY_MB">51200</env>

bernt-matthias:
Fixed https://github.com/galaxyproject/tools-iuc/pull/2521

PR #2521, opened 2019-08-02
trinity: update README to clarify specification of memory limit

pjbriggs:
FOR CONTRIBUTOR:

PR to address issue #2513 by updating the `README.rst` instructions for specifying the memory limits for the `trinity` tool (NB there are no functional changes to the tool in this PR).

### DIFF
--- a/tools/trinity/README.rst
+++ b/tools/trinity/README.rst
@@ -9,7 +9,14 @@ trinity requires a large amount of memory to perform the assembly: "roughly
 By default, this tool is configured to limit the memory consumption to 1G.
 You might need to lower this limit if the machine(s) executing the jobs have less memory available.
 If you have a lot of reads to assemble and a machine with enough memory, you can increase it.
-In both cases, you can set the GALAXY_MEMORY_MB environmental variable in the destination section of the job_conf.xml file to the amount of available memory (in MB)::
+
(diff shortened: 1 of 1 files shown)


abretaud:
Maybe I'm wrong, but I think the `GALAXY_MEMORY_MB` env var is set automatically by Galaxy now depending on the destination config (e.g. if I write `--mem=50000` in `nativeSpecification` for a slurm destination), just as SLOTS.
I think it's only needed to add this var in job_conf.xml for running new trinity versions on an old galaxy server.
If I'm not wrong (not guaranteed! :D) it would be good to write it in the README too

pjbriggs:
@abretaud if you can provide a link to relevant documentation or similar for the automatic setting of the `GALAXY_MEMORY_MB` that you describe then I'll look and update the README again.

abretaud:
I think you can refer to https://planemo.readthedocs.io/en/latest/writing_advanced.html#cluster-usage and/or https://docs.galaxyproject.org/en/latest/admin/cluster.html#special-environment-variables-for-job-resources

abretaud:
Getting some 500 errors from TTS during upload https://travis-ci.org/galaxyproject/tools-iuc/jobs/611261627
I restarted 2 or 3 times this morning, but still the same, any idea what's going on?

bgruening:
TS was/is down. But I'm actually happy that this is not uploaded, it does not touch the tool isn't it?

abretaud:
No change to the tools, just the readme, so no hurry I guess

PR #2523, opened 2019-08-03
trigger a test - is that failing?

bgruening:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/volcanoplot/volcanoplot.xml
+++ b/tools/volcanoplot/volcanoplot.xml
@@ -76,6 +76,7 @@ Rscript '${__tool_directory__}/volcanoplot.R'
             <when value="file">
                 <param name="label_file" type="data" format="tabular" label="File of labels"/>
             </when>
+            <!--when value="none" /-->
         </conditional>
         <section name="plot_options" expanded="false" title="Plot Options">
             <param name="boxes" type="boolean" truevalue="True" falsevalue="False" checked="True" label="Label Boxes" help="If this is set to Yes, the labels for the points will be in boxes. Default: Yes"/>
--- a/tools/volcanoplot/volcanoplot.xml
+++ b/tools/volcanoplot/volcanoplot.xml
@@ -76,7 +76,7 @@ Rscript '${__tool_directory__}/volcanoplot.R'
             <when value="file">
                 <param name="label_file" type="data" format="tabular" label="File of labels"/>
             </when>
-            <!--when value="none" /-->
+            <when value="none" />
         </conditional>
         <section name="plot_options" expanded="false" title="Plot Options">
             <param name="boxes" type="boolean" truevalue="True" falsevalue="False" checked="True" label="Label Boxes" help="If this is set to Yes, the labels for the points will be in boxes. Default: Yes"/>


bgruening:
We have many of those `a when tag has not been defined for 'labels ` in our logs and for a long time I thought these are only old tools before planemo linting. But it seems that new tools do also slip in with this. So maybe planemo lint is not catching it.

I'm on the fence if Galaxy should actually spit out a warning for those missing "when"s. If planemo should probably complain, but not Galaxy.

bgruening:
@mvdbeek what do you think?

mvdbeek:
Yes, we should add a linter for missing when tags.
Re: logging, I remember we discussed this on the Galaxy repo somewhere (maybe @erasche remembers ?) but I can't find the issue/PR. I think the conclusion was that this should be logged on the Galaxy side (meaning all currently installed tools will always continue logging this message). Maybe another log level ("tool_warning", "tool_debug") might be a good solution for this ?

mvdbeek:
Actually it looks like we have a linter that is supposed to do it: https://github.com/galaxyproject/galaxy/blob/2cd95c48f6dabfe79ec9a5c440bff8cfc29c68df/lib/galaxy/tools/linters/inputs.py#L83

nsoranzo:
The planemo linter works fine, but it's a warning, not an error.

mvdbeek:
Ha, that's the conclusion I reached after writing some unit tests for the linter.
So, we bump it up to error on the IUC repo ?

bgruening:
My preference would be to either consider it as an error in planemo and a warning/ignore it in Galaxy. Or leave it as a warning but then ignore it in Galaxy.

An Admin should not need to deal with such warnings imho. The admin usually does not know the tool in depth and can not react at this time. The warning will stay forever without any chance to change it.

I think it made a lot of sense to warn in Galaxy before we had planemo, as Galaxy was used for tool dev, but with planemo I'm not convinced anymore.

mvdbeek:
Many people still develop that way though, and admins may installed old/broken tools on their instance. I think it should remain as warning in Galaxy.

bgruening:
> installed old/broken tools on their instance

They are not old, as shown in this PR and they are also not broken. 

Imho this warning is for tool-devs - "look there is a `<when>` missing you might have missed describing an option".
 This `<when>` warning is really useless for an Admin imho, as there is no way to fix it. Happy to discuss this aspect in the Galaxy issue if Helena finds it.

If you really think this should be a warning in Galaxy, then to save Admins from any headache we should make it an error in planemo for sure.

mvdbeek:
Yes, it should be an error. No, admins may need to figure out why tools don't work the way they're intended to. One reason might be because of missing when tags. Developers should make sure they're fully defined, but as you know it's hard to write perfect bugfree tools, hence this should be a warning.

mvdbeek:
> This `<when>` warning is really useless for an Admin imho, as there is no way to fix it. Happy to discuss this aspect in the Galaxy issue if Helena finds it.

It isn't useless though. It is useless from the perspective of looking at the logs to see if there's something that should be fixed. It is very useful though if you get a bug report from a user that says "hey, this setting is supposed to have an effect, but it seems it doesn't".

bgruening:
> Plus I could swear we discussed this before on the Galaxy repo and reached that conclusion (with Helena's involvement).

That's why I said let's move this to this issue and discuss it there.
At least the Galaxy part. It's good that we at least reached the conclusion to bump it to an error on the planemo part. Created an issue here: https://github.com/galaxyproject/planemo/issues/949

Some numbers for people that do care:

> galaxy@sn04:~$ journalctl _UID=999 | grep 'when tag' | wc -l
4750

bgruening:
Feel free to merge this one it should fix at least one warning.

nsoranzo:
Maybe we can decrease the log level to info or debug at https://github.com/galaxyproject/galaxy/blob/2cd95c48f6dabfe79ec9a5c440bff8cfc29c68df/lib/galaxy/tools/linters/inputs.py#L83 for Galaxy? But I suspect this is the same code used by Planemo, so it can be tricky.

erasche:
Morning!

> logging, I remember we discussed this on the Galaxy repo somewhere (maybe @erasche remembers ?) but I can't find the issue/PR. 

https://github.com/galaxyproject/galaxy/pull/5920, took me a minute to find, too, until I removed `is:issue`. I thought it was an issue too, not an ancient open PR lol.

> I think the conclusion was that this should be logged on the Galaxy side (meaning all currently installed tools will always continue logging this message). Maybe another log level ("tool_warning", "tool_debug") might be a good solution for this ?

the conclusion for me was: this should be moved to the frontend, not shown in backend logging that few people look at.

> Many people still develop that way though, and admins may install old/broken tools on their instance. I think it should remain as warning in Galaxy.

Sure, I develop tools in Galaxy but I don't look at startup logs. Even on a minimal startup there's already 1k other messages, I'm not watching for my specific tool's or scrolling back. I do the same things I do as an admin debugging a broken tool (next paragraph)

> No, admins may need to figure out why tools don't work the way they're intended to. 
> ...
> It is very useful though if you get a bug report from a user that says "hey, this setting is supposed to have an effect, but it seems it doesn't".

Speaking as an admin: I have *never*, not once looked at startup logs (except if tool isn't loading at all). When I debug broken tools I do look at:

- the XML
- planemo lint's output
- the user's supplied parameters
- the generated command line.

Do you find yourself looking at startup logs often? I'd be interested to hear another opinion. I find that there is way too much information there and very little of it is actionable, for startup logs I find that I really want actionable information. EU doesn't, I don't have the impression the US side is looking at startup logs for missing when tags either. Admittedly I am biased towards large sites rather than small ones, but even when I maintained a smaller server, I remember using the same debugging steps above?

> That's why I suggested adding another log level for this so you can filter on that. Plus I could swear we discussed this before on the Galaxy repo and reached that conclusion (with Helena's involvement).

John proposed that but Dannon and I were focused on sharing with tool users + authors, rather than with server admins. If it gets reclassed as another log level and I can mute it, great, whatever, works for me. As long as I don't have to see it and can filter for the information that might indicate more important issues.

mvdbeek:
> the conclusion for me was: this should be moved to the frontend, not shown in backend logging that few people look at.

I don't know, this seems even more unactionable and confusing than admins seeing this in the logs, though there is some logic to it as well.

Maybe you're right and the warning isn't really that useful. I haven't actually debugged a tool that way either, mostly because I'd write a test first, and then it'd be very obvious that a when option is missing. I'm not opposed to dropping the level to info or debug or a separate leve if we get some consensus here.

PR #2525, opened 2019-08-05
Update FreeBayes wrapper to 1.3.1

davebx:
FOR CONTRIBUTOR:

This is a clean version of #2487, since my attempts at rebasing that one ended up with one merge conflict after another.


bgruening:
@davebx I'm not a fan of dumping the parameters into the help. The parameters should be explained as part of the UI. But if you like it keep it we do not have rules for it atm.


@bernt-matthias do you have time to look at it?

bernt-matthias:
Hi @bgruening. Very busy the next two weeks. Could have a look afterwards. Have my points already been considered. The comments seems still to be pending.

nekrut:
I disagree vigorously with the fact that we do not need to provide parameters in help. I think it is vital. The complex tools we hide many parameters. Having them all listed is a way to let users scan the parameters and decide what they may or may not change. It also provides continuity with the command line version of the tool. If this is the only issue, please accept this PR.

bgruening:
@nekrut I'm fine with merging. The help is not a stopper I think we don't have a rule for that afaik. 
But if your only argument is that "a list of all parameters is needed", then this list should be generated automatically and we should have a UI for that - to fulfill your use-case - a coherent one for all tools.
I don't see a point for duplicating information twice. And one of them in an unstructured way.
This will cause issues in the long run imho, as we need to update two places etc ...

nekrut:
I don't think we duplicate this info. We identity an option right by the UI widget, but then also provide a global list of parameters in help. This way a user can see all parameters at once. I often use this approach. I get lost if a tool section does not list all options.

tools/freebayes/macros.xml
```diff
@@ -44,9 +44,9 @@
         </conditional>
     </xml>
     <token name="@COVERAGE@">
-        --min-coverage ${options_type.min_coverage}
-        --skip-coverage ${options_type.skip_coverage}
-        --limit-coverage ${options_type.limit_coverage}
+        --min-coverage ${coverage_options.min_coverage}
```
bgruening: the test failed beacuse `--min-coverage 0 --skip-coverage 0 --limit-coverage 400` is not part of test 5 ... is that correct?

davebx: It should be passing, 0 is the default for those options when they're not specified, and test 5 is passing on my local system.

bgruening: The output is differes slightly ... maybe you need to commit your test-data? It is failing atm.

bgruening:
Merci!

Issue #2105, opened 2018-09-26
SRA-tools --table option


mmiladi:
Hello,

fastq-dump step for SRA-tools download has an option `--table` which is not exposed to the galaxy wrapper. This option is crucial for noisy long-reads, otherwise the output is empty with a message like in std-err:

>Rejected 101233 READS because READLEN < 1
Read 101233 spots for SRR6982805
Written 0 spots for SRR6982805

Further info can be found in this discussion forum: http://seqanswers.com/forums/archive/index.php/t-58456.html

A test of a sample SRA entry can be found here: https://usegalaxy.org/u/miladi/h/testsra-longreads

mmiladi:
ping @bgruening @shiltemann

lldelisle:
It is now added.

bernt-matthias:
closed via https://github.com/galaxyproject/tools-iuc/pull/2541

Issue #1099, opened 2017-01-03
Bug: SRA tools -fastqdump not splitting PE reads


MoHeydarian:
The SRA tools -> Extract reads (fastqdump) tool has an option to 'split spot by read pairs'. Regardless of what is specified for this setting the imported data is never split based on read pairs (in the case of PE data). 

This requires the user to have to run FastQ splitter on their imported data, which is an additional time and space consuming step.

yhoogstrate:
It could be that this only works when a text file with identifiers is used for input.

blankenberg:
@MoHeydarian issue described in https://github.com/galaxyproject/tools-iuc/issues/970 still occurs.

lldelisle:
I was suprised to see that the output of sra with PE reads was interleaved.
I just need to wrap mind. Let's do it within 1 month. If it is no up. Please remind me.

lldelisle:
Maybe the best solution:
Fasterq-dump
A faster fastq-dump.

lldelisle:
Hi,
To get 2 separated files from a PE sra accession number, you need to generate a txt file with your accession number and select list of SRA accession, one per line. This will create a collection with a pair of fastq inside.

lldelisle:
Hi @MoHeydarian,
I wrapped the fasterq-dump which should be quicker to download fastq from sra. In this wrapper, the read pairs will always be in 2 different datasets, however, I found it quite convenient to have them in collection. If your user want to have them as datasets, he just need to unhide them.
I hope it helps,

Lucille

PR #2541, opened 2019-08-22
sra-tools: new wrapper for fasterq-dump and update to v2.10

lldelisle:
FOR CONTRIBUTOR:

Hi,
I decided to wrap fasterq-dump and update to v2.10.
fasterq-dump works faster than fastq-dump.

My wrapper will make output in collections even if  an accession number is given or a sra file is provided. This will solve #1099.
In addition, I noticed as described in #970 that by default the technical reads are dumped and the fastq identity is defined independently of this option (forward reverse).
Which means that:
- if you have one biological read and 3 technical reads as described in #970, the dataset will not appear as there will be more than a forward and reverse file so it will not be integrated in the collection.

- if you have one biological read and 1 technical read like in SRR002702. One would be consider as forward and the other one as reverse whereas the first one is technical and the second one is biological.

In the wrapper I wrote, there is a third collection which is other and in which I put the fastq when I am not sure it is a true pair or where there are more than 2 fastq. I also use this collection to put singleton reads in paired dataset so the collection with paired-end reads can be used by bowtie2.

I included the --table option (#2105).

@mvdbeek,
Do you think I should change the fastq-dump wrapper to make it like the fasterq-dump wrapper, or should I change my wrapper to be closer to yours, or we keep them like this?
Do you still have features you would like to include?


tools/sra-tools/fasterq_dump.xml
```diff
@@ -0,0 +1,257 @@
...
+    done
+    #end if
+    ]]>
+    </command>
+    <inputs>
+        <expand macro="input_conditional"/>
+        <param name="outputformat" type="select" display="radio" label="Select output format" help="Compression will greatly reduce the amount of space occupied by downloaded data. Downstream applications such as a short-read mappers will accept compressed data as input. Consider this example: an uncoimpressed 400 Mb fastq datasets compresses to 100 Mb or 80 Mb by gzip or bzip2, respectively.">
+            <option value="fastqsanger.gz">gzip compressed fastq</option>
+            <option value="fastqsanger">Uncompressed fastq</option>
+            <option value="fastqsanger.bz2">bzip2 compressed fastq</option>
```
mvdbeek: That's not a widely used option, and a lot of wrappers would trigger unpacking to fastq first, if this isn't too important to you I'd skip this

tools/sra-tools/fasterq_dump.xml
```diff
@@ -0,0 +1,257 @@
+<tool id="fasterq_dump" name="Faster Download and Extract Reads in FASTQ" version="@VERSION@.4">
```
mvdbeek:
```suggestion
<tool id="fasterq_dump" name="Faster Download and Extract Reads in FASTQ" version="@VERSION@.4" profile="18.09">
```
this will give us an isolated $HOME, which fastq-dump used to use to write temporary files, I imagine that hasn't changed.

tools/sra-tools/fasterq_dump.xml
lldelisle: You mean we gzip by default? Because the tool by default would not compress.

tools/sra-tools/fasterq_dump.xml
lldelisle: Yes it changed. It will create a folder in the current working directory. No need to change the home so I would prefer to keep it working with lower version of galaxy if you don't mind.

tools/sra-tools/fasterq_dump.xml
mvdbeek: No, I meant the bzip2 option. gzip by default makes sense.

lldelisle:
The `maxSportId` will probably never work because it will download with multi thread the whole sra... But for the tests I found small sra files thanks to SRAdb...

tools/sra-tools/fasterq_dump.xml
mvdbeek: Are you certain that there is no ~/.ncbi folder on your system?

Please add a profile version, it can be lower than that, but without this we use some old and bad defaults.

lldelisle:
For `-e`, what should I use? `\${GALAXY_SLOTS:-4} `?

tools/sra-tools/fasterq_dump.xml
lldelisle: Yes sure. And in the wiki:"The tool will also create a directory named 'fast.tmp' ".
OK I will put 16.01

mvdbeek:
> For `-e`, what should I use? `\${GALAXY_SLOTS:-4} `?

Sure, I'd probably go for  `\${GALAXY_SLOTS:-1}`, but in practice the fallback is never going to be used.

tools/sra-tools/fasterq_dump.xml
mvdbeek: OK, thanks for checking.

tools/sra-tools/fasterq_dump.xml
```diff
@@ -0,0 +1,238 @@
...
+    #end if
+
+    fasterq-dump "\$acc" -e \${GALAXY_SLOTS:-4}
+    $adv.split
+    #if str( $adv.minlen ) != "":
+        --min-read-len "$adv.minlen"
+    #end if
+    $adv.skip_technical >> $log 2>&1 
+    &&
+    mkdir -p output;
```
mvdbeek: please use `&&` for all commands (https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#command-formatting).

lldelisle:
@mvdbeek, do you see other changes I should do?

mvdbeek:
Nope, looks very good, thanks a lot!

lldelisle:
Thanks

PR #2547, opened 2019-08-26
pygenometracks: use labels in the config file when off

lldelisle:
Hi,
Sorry, I just noticed that the option plot labels for bed and genes was not written in the config file when set to off whereas the default is on...


bgruening:
Danke!

lldelisle:
Thank you for the reactivity.

lldelisle:
It looks like it did not updated in usegalaxy.eu (I was expecting to see version 3.1.2)...
image
Do you know why?

bgruening:
The automatic update only happens on Saturdays, I will do a semi-manual round today and get @bebatut's new tools also in.

lldelisle:
Ah I did not know. Thanks.

bgruening:
Should be installed by now.

lldelisle:
Yes many thanks

PR #2548, opened 2019-08-26
Fix to Table Compute tool

mtekman:
FOR CONTRIBUTOR:

+ [x] Fix to version command and import statements
+ [ ] TODO Resolve statement `table.sub(table.mean(1), 0).div(table.std(1),0)`


bgruening:
Put on WIP until @bebatut issues are also solved.

mtekman:
**Updates**

The issue right now is that a statement like `(table.T - table.mean(1)).T` is parsed by our _Safety.py_ to give `['(', 'table.T', '-', 'table.mean', '(', '1', ')', ')', '.T']` and is then cleaned to give `['table.T', 'table.mean', '1', '.T']`

The problem was mentioned by @bebatut in the meeting that `.T` is unbound, and so the parser does not know which class it should check.

*****

So I've been playing around with the idea that we could solve this custom expression problem by instead of relying on our own tokenizing of statements, which is dependent on the order that the user provides, we standardize everything using `ast` syntax tree module.

I have a small mockup below

```python
def representThis(expr):
    import ast
    
    def str_node(node):
        """Print Node Contents, and children rec. """
        if isinstance(node, ast.AST):
            fields = [(name, str_node(val)) for name, val in ast.iter_fields(node) if name not in ('left', 'right')]
            fields = [f for f in fields if f[0] in ("attr","id", "n")]
            #fields = [f for f in fields if f[0]]
            #for f in fields:
            #    print(f[0], f[1])
            #rv = '%s(%s' % (node.__class__.__name__, ', '.join('%s' % y for x,y in fields))
            rv = ",".join(f[1] for f in fields)
            return rv + ')'
        else:
            return repr(node)
    
    def ast_visit(node, level=0):
        """Visit all iterable nodes"""
        print((" " * level) + str_node(node))
        for field, value in ast.iter_fields(node):
            if isinstance(value, list):
                for item in value:
                    if isinstance(item, ast.AST):
                        ast_visit(item, level=level+1)
            elif isinstance(value, ast.AST):
                ast_visit(value, level=level+1)
    
    ast_visit(ast.parse(expr))
```
The function call

`representThis("(table.T - table.mean(1)).T")`

gives

```
)
 )
  'T')
   )
    'T')
     'table')
      )
     )
    )
    )
     'mean')
      'table')
       )
      )
     1)
   )
```

which could be a neater way of trying to standardize the tree

*****
After a talk with @bgruening, it now seems that the AST route is superfluous and that for unbounded functions like `.T` we could just infer the correct class from the context given in the Galaxy wrapper, where the custom functions exist in either the `elem`, `vec`, or `table` contexts.

This would be a much quicker fix, and I will try to work on this soon

mtekman:
@bebatut give this one a try ;-)

tools/table_compute/table_compute.xml
```diff
@@ -53,10 +63,16 @@
...
             <param name="input" type="data" format="tsv,tabular" label="Table" />
             <param name="col_row_names" type="select" display="checkboxes" multiple="true" optional="true"
-            label="This input data has">
+                   label="Input data has">
                 <option value="has_col_names" selected="true">Column names on the first row</option>
-                <option value="has_row_names" selected="true">Row names on the first column"</option>
+                <option value="has_row_names" selected="true">Row names on the first column</option>
             </param>
+            <section name="adv" title="Advanced File Options " expanded="false" >
+                <param name="header" type="integer" min="0" optional="true" label="Header begins at line N" help="All lines before line N will be skipped. If a value is set, this will override the above 'Column names on the first row' parameter." />
```
bgruening: It seems to me this is really complicated and references other parameters, overwrites them add additional code ... instead, people could easily use tail and head to archive the same.

mtekman: Head and tail will be the main defaults for tasks like this, but I can understand the need for a user to quickly parse a table without trimming it first. 

I placed the options for it in the collapsed section, so these options will likely not be used very much, but they are native functions of `pd.read_csv` so I figured why not just include them?

mtekman:
+ [X] Operation on a specific column or row
  + [X] adding text instead of replacing
    * [X] "chr" + elem not allowed in custom manipulate
    * [X] table[0] = "chr" + table[0].astype(str), brackets or strings not allowed
  + [X] Add arbitrary column with value
  + [X] Fix sanitisers so that 1:10 are allowed
  + [X] Allow negative indexes
  + [X] Add option to skip the first N lines
  + [X] Add option to skip the last N lines
  + [X] Add option to read only N lines

So I added the ability to manipulate elements by selecting rows and column

tools/table_compute/table_compute.xml
```diff
@@ -1595,6 +1668,65 @@ In Galaxy we would select the following:
...
+ * *Type of table operation* →  **Manipulate selected table elements**
+
+   * *Operation to perform* → **Custom**
+
+     * *Custom Expression* → ::
+
+         "chr%.f" % elem
+
+     * *Operate on elements* → **Specific Rows and/or Columns**
+     * *List of columns to select* → "2"
```
bgruening: What happend when a user puts in 2,3 and 2,4?

mtekman: yep it still works - if the user sets the inclusive flag to true, then all elements in row 2+4 and columns 2+3 will be modified. if set to false, then only the elements [2,2], [2,3],[4,2],[4,3] will be modified

bgruening:
Test was green, just github was not updated. Thanks @mtekman!

PR #2550, opened 2019-08-26
Add calculate_contrast_threshold tool

PrashantKuntala:
FOR CONTRIBUTOR:

FYI, based on the recommended IUC standards for python tools, I checked for pep8 and flake8 standards, below are the commands showing the errors I am ignoring.

`pycodestyle --show-pep8 --show-source --ignore=E501 mytool.py`
`flake8 --ignore=E201,E202,E501 --count .`


tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
```diff
@@ -0,0 +1,104 @@
...
+    <inputs>
+        <param name="input_file" argument="-i" type="data" format="txt" label="Pileup Data Matrix "/>
+        <param name="header" argument="-d" type="text" value="T" label="Header (T/F)" help="T: First line is header; F: Header is absent"/>
+        <param name="start_col" argument="-s" type="integer" value="2" label="Start Column"/>
+        <param name="col_num" argument="-l" type="integer" value="300" label="Column Number" help="This would be the width of the plot you will be generating"/>
+        <param name="row_num" argument="-r" type="integer" value="600" label="Row Number" help="This would be the height of the plot you will be generating"/>
+
+        <conditional name="quantile_type">
+            <param name="quantile_type_selector" type="select" display="radio" label="Choose the Contrast paramter">
+                <option value="b_option" selected="true">- b</option>
```
Slugger70: Could you perhaps include a bit of information of what the -b option and -t option are? Some short text here and more detailed explanation of the options in the `<help>` section would be good.

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
```diff
@@ -0,0 +1,104 @@
+<tool id="calculate_contrast_threshold" name="Calculate Contrast threshold" version="1.0.0">
+    <description>from tag pileup CDT
+    </description>
+    <requirements>
+        <requirement type="package" version="1.14.2">numpy</requirement>
+        <requirement type="package" version="2.7.14">python</requirement>
+    </requirements>
```
Slugger70: Does the python script have a version command?

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
```diff
@@ -0,0 +1,117 @@
+<tool id="calculate_contrast_threshold" name="Calculate Contrast threshold" version="1.0.0">
+    <description>from tag pileup CDT
+    </description>
+    <requirements>
+        <requirement type="package" version="1.14.2">numpy</requirement>
+        <requirement type="package" version="2.7.14">python</requirement>
```
mvdbeek: Could you make the script python 3 compatible ? On first glance it looks like you'll only have to replace the print statements with print function calls.

tools/calculate_contrast_threshold/calculate_contrast_threshold.py
```diff
@@ -0,0 +1,187 @@
...
+
+    M, N = a.shape
+
+    # compare the heatmap matrix
+    a_compress = a.reshape((m, M / m, n, N / n)).mean(3).mean(1)
+    return np.array(a_compress)
+
+
+def load_Data(input_file, quantile, absolute, header, start_col, row_num, col_num, min_upper_lim):
+    data = open(input_file, 'r')
```
mvdbeek: please use `with open(input_file, 'r') as data:` or equivalent when you open files.

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
```diff
@@ -0,0 +1,117 @@
...
+
+        <conditional name="quantile_type">
+            <param name="quantile_type_selector" type="select" display="radio" label="Choose the Contrast paramter">
+                <option value="b_option" selected="true"> Calculate thresholds from data (-b)</option>
+                <option value="t_option"> Enforce absolute thresholds (-t)
+                </option>
+            </param>
+
+            <when value="b_option">
+                <param name="quantile" argument="-b" type="float" value="95" label="Quantile" help="Enter your quantile value above. can take values between (1 to 100)"/>
```
mvdbeek: If this is limited to a given range you should also use `min="1"` and `max="100"`

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
```diff
@@ -0,0 +1,117 @@
...
+                <param name="min_contrast" type="float" value="0" label="Minimum upper limit after Quantile calculation" help="set contrast to above value, if calculated quantile is less than or equal to above value. can take real values (>= 0)" argument="-m"/>
```
mvdbeek: in this case `min="0"`, right ?

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
```diff
@@ -0,0 +1,117 @@
...
+            </when>
+
+            <when value="t_option">
+                <param name="quantile2" argument="-t" type="float" value="0.0" label="Absolute tag threshold" help="Enter your custom tag threshold value above. can take real values (>= 0)"/>
```
mvdbeek: `min="0"`

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
```diff
@@ -0,0 +1,117 @@
...
+        -t '${quantile_type.quantile2}'
+      #end if
+
+      -d '$header' -s '$start_col' -r '$row_num' -l '$col_num'
+    ]]>
+    </command>
+
+    <inputs>
+        <param name="input_file" argument="-i" type="data" format="txt" label="Pileup Data Matrix "/>
+        <param name="header" argument="-d" type="text" value="T" label="Header (T/F)" help="T: First line is header; F: Header is absent"/>
```
mvdbeek: I don't quite understand this parameter. Should it be a `type="boolean"` parameter ?

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
PrashantKuntala: yes, it is a range. Will update the wrapper

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
PrashantKuntala: yes it is `min=0`

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
PrashantKuntala: Yeah makes sense to have a `type=boolean`, thanks

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
```diff
@@ -0,0 +1,123 @@
...
+       -d 'F'
+      #end if
+    
+      -s '$start_col' -r '$row_num' -l '$col_num'
+    ]]>
+    </command>
+
+    <inputs>
+        <param name="input_file" argument="-i" type="data" format="txt" label="Pileup Data Matrix "/>
+        <param name="header" argument="-d" type="boolean" truevalue="true" falsevalue="false" checked="true" value="T" label="Does the input file have a Header ?" help="standard CDT file have headers."/>
```
mvdbeek: You can simplify this 
```suggestion
        <param name="header" argument="-d" type="boolean" truevalue="T" falsevalue="F" checked="true"  label="Does the input file have a Header ?" help="standard CDT file have headers."/>
```

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
```diff
@@ -0,0 +1,123 @@
...
+
+      ## Setting the quantile values properly.
+      #if str($quantile_type.quantile_type_selector) == "b_option":
+        -q '${quantile_type.quantile}'
+        -m '${quantile_type.min_contrast}'
+      #else if str($quantile_type.quantile_type_selector) == "t_option":
+        -t '${quantile_type.quantile2}'
+      #end if
+
+      #if str($header) == "true":
```
mvdbeek:
```suggestion
```

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
```diff
@@ -0,0 +1,123 @@
...
+       -d 'T'
+      #else:
```
mvdbeek:
```suggestion
```

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
```diff
@@ -0,0 +1,123 @@
...
+       -d 'F'
```
mvdbeek:
```suggestion
```

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
```diff
@@ -0,0 +1,123 @@
...
+      #end if
```
mvdbeek:
```suggestion
```

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
mvdbeek:
```suggestion
```

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
mvdbeek:
```suggestion
```

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
```diff
@@ -0,0 +1,123 @@
...
+      ## Setting the quantile values properly.
+      #if str($quantile_type.quantile_type_selector) == "b_option":
+        -q '${quantile_type.quantile}'
+        -m '${quantile_type.min_contrast}'
+      #else if str($quantile_type.quantile_type_selector) == "t_option":
+        -t '${quantile_type.quantile2}'
+      #end if
+
+      #if str($header) == "true":
+       -d 'T'
```
mvdbeek:
```suggestion
```

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
mvdbeek: this will take on the truevalue or falsevalue that you define in the boolean parameter

PrashantKuntala: thanks, its more cleaner the way you suggest

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
```diff
@@ -0,0 +1,117 @@
...
+        <conditional name="quantile_type">
+            <param name="quantile_type_selector" type="select" display="radio" label="Choose the Contrast paramter">
+                <option value="b_option" selected="true"> Calculate thresholds from data (-b)</option>
+                <option value="t_option"> Enforce absolute thresholds (-t)
+                </option>
+            </param>
+
+            <when value="b_option">
+                <param name="quantile" argument="-b" type="float" min="0" max="100" value="95" label="Quantile" help="Enter your quantile value above."/>
+                <param name="min_contrast" type="float" min="0" value="0" label="Minimum upper limit after Quantile calculation" help="sets contrast to above value, if calculated quantile is less than or equal to above value. takes real values (>= 0)" argument="-m"/>
```
mvdbeek: Again, I think the "above" makes it confusing, it should be obvious that this help text refers to this parameter (and not to quantile, for instance, which is actually above this parameter in the tool form). So maybe phrase it as `This value will be used as the upper limit if the calculated quantile is below this value` ?

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
```diff
@@ -0,0 +1,117 @@
...
+      #end if
+      
+      -d '$header' -s '$start_col' -r '$row_num' -l '$col_num'
+    ]]>
+    </command>
+
+    <inputs>
+        <param name="input_file" argument="-i" type="data" format="txt" label="Pileup Data Matrix "/>
+        <param name="header" argument="-d" type="boolean" truevalue="T" falsevalue="F" checked="true" label="Does the input file have a Header ?" help="standard CDT file have headers."/>
+        <param name="start_col" argument="-s" type="integer" value="2" label="Start Column"/>
```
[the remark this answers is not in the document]
PrashantKuntala: sure, will extend the help section describing all parameters and how to choose them.

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
```diff
@@ -0,0 +1,117 @@
...
+
+        <conditional name="quantile_type">
+            <param name="quantile_type_selector" type="select" display="radio" label="Choose the Contrast paramter">
+                <option value="b_option" selected="true"> Calculate thresholds from data (-b)</option>
+                <option value="t_option"> Enforce absolute thresholds (-t)
+                </option>
+            </param>
+
+            <when value="b_option">
+                <param name="quantile" argument="-b" type="float" min="0" max="100" value="95" label="Quantile" help="Enter your quantile value above."/>
```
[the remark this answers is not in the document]
PrashantKuntala: Will cleanup `param`'s help attributes.

PrashantKuntala:
The `test-data` is about 329kb. I can't make it smaller.

mvdbeek:
> The `test-data` is about 329kb. I can't make it smaller.

If I just take the first twenty columns with `cut -f1-23 sample.tabular` the test passes fine. Is that expected ?  If yes, is there any reason to not use the first 20 columns only ?

PrashantKuntala:
It is expected to pass. Well, I don't see any reason to have first 20 columns or even 10 columns in the `sample.tabular`, only that I don't expect anyone to work with few data columns. I guess for the purposes of testing it should be fine.

mvdbeek:
> Well, I don't see any reason to have first 20 columns or even 10 columns in the `sample.tabular`, only that I don't expect anyone to work with few data columns

The reason is that we're limited in size for this repo, if you could upload a smaller file that would be very helpful.

PrashantKuntala:
> The reason is that we're limited in size for this repo, if you could upload a smaller file that would be very helpful.

Sure, makes sense to limit the file sizes. I modified `sample.tabular` to contain a smaller data matrix.

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
```diff
@@ -0,0 +1,122 @@
...
+      #end if
+      
+      -d '$header' -s '$start_col' -r '$row_num' -l '$col_num'
+    ]]>
+    </command>
+
+    <inputs>
+        <param name="input_file" argument="-i" type="data" format="txt" label="Pileup Data Matrix "/>
+        <param name="header" argument="-d" type="boolean" truevalue="T" falsevalue="F" checked="true" label="Does the input file have a Header ?" help="standard CDT file have headers."/>
+        <param name="start_col" argument="-s" type="integer" value="2" label="Valid Data Start Column" help="valid data values start from this column"/>
```
mvdbeek:
```suggestion
        <param name="start_col" argument="-s" type="integer" value="2" label="Valid Data Start Column" help="Valid data values start from this column"/>
```

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
```diff
@@ -0,0 +1,122 @@
...
+        <param name="col_num" argument="-l" type="integer" value="300" label="Plot Width in pixels" help="equal to the heatmap width you plan to create"/>
```
mvdbeek: Sorry for the piecemeal review here, but why do you need to set plotting options here ? I assume there is another tool for plotting, why doesn't the user set this value there ? It doesn't look like the calculated threshold depends on this ?

PrashantKuntala: Yes, `calculate threshold` does depend on the plot dimensions, it uses these values to re-bin the data matrix to reduce sparsity in the data before calculating quantile value. There is also a plotting script that can use these values from the output of `calculate threshold` as an option, or use custom plot dimensions if needed.  `calculate threshold` and plotting script use similar re-bining algorithm before generating plots.

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
PrashantKuntala: Is this really necessary ?  I don't understand the suggested change.

mvdbeek: It looks better to start the help tag in uppercase (as you do most other help tags), please also do this for the col_num, row_num and  header parameter.

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
mvdbeek: ok, that's fine then.

tools/calculate_contrast_threshold/calculate_contrast_threshold.xml
PrashantKuntala: alright, will make them uppercase.

mvdbeek:
Thanks a lot @PrashantKuntala!

PrashantKuntala:
Thank you for the review and merge !

PR #2579, opened 2019-09-03
samtools stats: remove plot_bamstats call

bernt-matthias:
To get rid of the gnuplot and perl dependency

advantage: no additional mulled environment needed

FOR CONTRIBUTOR:

### DIFF
--- a/tool_collections/samtools/samtools_flagstat/samtools_flagstat.xml
+++ b/tool_collections/samtools/samtools_flagstat/samtools_flagstat.xml
@@ -1,4 +1,4 @@
-<tool id="samtools_flagstat" name="Samtools flagstat" version="2.0.2">
+<tool id="samtools_flagstat" name="Samtools flagstat" version="2.0.3">
     <description>tabulate descriptive stats for BAM datset</description>
 
     <macros>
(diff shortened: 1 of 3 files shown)


bgruening:
Is the additional mulled env problematic? I think bioconda will not like it. There was recently a discussion about adding R to Picard and this was shut down. https://github.com/bioconda/bioconda-recipes/issues/3859

What you can do is create a samtools-full packages or similar, but then again you have an additional environment.

bernt-matthias:
No real problem, the extra env just takes more memory which is not really necessary.

Interesting discussion. I'm pro adding the extra dependency to the package. Lets see what the community thinks.

After finding out that multiqc can plot the stats I kind of regret that I added the plot_bamstats call... we might as well remove it .. it would not remove any functionality from Galaxy.

bgruening:
Thank remove it :)

> No real problem, the extra env just takes more memory which is not really necessary.

If you mean storage here, think about it the other way, everyone that just needs samtools, including all our other tools that include samtools, will get now R and gnuplot as well - in their mulled-env.

bernt-matthias:
Would it be OK to keep the version. For me it would be fine if the change goes to the TS with the next mayor improvement of the tools.

bgruening:
Not sure how? Do you mean not merging this PR, or merge and kill travis to not upload the recipes?

bernt-matthias:
Ahh .. lets just bump :) .. 

Maybe it would be an option not to upload to the TS if the version is already present in the TS anyway?

mvdbeek:
Can you bump the version ? I don't see the point of holding back releases ? The smaller the release the better IMO.

bernt-matthias:
sure. done

PR #2585, opened 2019-09-08
update Cutadapt to the latest 1.x series and test Galaxy dev

bgruening:
FOR CONTRIBUTOR:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,7 +73,7 @@ jobs:
 
 before_install:
   - export GALAXY_REPO=https://github.com/galaxyproject/galaxy
-  - export GALAXY_RELEASE=release_19.05
+  - export GALAXY_RELEASE=dev
   # - export PLANEMO_CONDA_PREFIX="$HOME/conda"
   - unset JAVA_HOME
 
--- a/tools/cutadapt/cutadapt.xml
+++ b/tools/cutadapt/cutadapt.xml
@@ -1,10 +1,10 @@
-<tool id="cutadapt" name="Cutadapt" version="1.16.6" profile="17.09">
+<tool id="cutadapt" name="Cutadapt" version="1.18" profile="17.09">
     <description>Remove adapter sequences from Fastq/Fasta</description>
     <macros>
         <import>macros.xml</import>
     </macros>
     <requirements>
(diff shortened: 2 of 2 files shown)


bernt-matthias:
Cool. 

I guess we should check if the modifications in the changelog have been reflected.

```
v1.18 (2018-09-07)
------------------

Features
~~~~~~~~

* Close :issue:`327`: Maximum and minimum lengths can now be specified
  separately for R1 and R2 with ``-m LENGTH1:LENGTH2``. One of the
  lengths can be omitted, in which case only the length of the other
  read is checked (as in ``-m 17:`` or ``-m :17``).
* Close :issue:`322`: Use ``-j 0`` to auto-detect how many cores to run on.
  This should even work correctly on cluster systems when Cutadapt runs as
  a batch job to which fewer cores than exist on the machine have been
  assigned. Note that the number of threads used by ``pigz`` cannot be
  controlled at the moment, see :issue:`290`.
* Close :issue:`225`: Allow setting the maximum error rate and minimum overlap
  length per adapter. A new :ref:`syntax for adapter-specific
  parameters <trimming-parameters>` was added for this. Example:
  ``-a "ADAPTER;min_overlap=5"``.
* Close :issue:`152`: Using the new syntax for adapter-specific parameters,
  it is now possible to allow partial matches of a 3' adapter at the 5' end
  (and partial matches of a 5' adapter at the 3' end) by specifying the
  ``anywhere`` parameter (as in ``-a "ADAPTER;anywhere"``).
* Allow ``--pair-filter=first`` in addition to ``both`` and ``any``. If
  used, a read pair is discarded if the filtering criterion applies to R1;
  and R2 is ignored.
* Close :issue:`112`: Implement a ``--report=minimal`` option for printing
  a succinct two-line report in tab-separated value (tsv) format. Thanks
  to :user:`jvolkening` for coming up with an initial patch!
```

Also seems that there is already v2.5. But this seems like a larger step.

bernt-matthias:
For reference it would be great to get this https://github.com/bioconda/bioconda-recipes/pull/18158 in as well (however the actual fix will turn out).

gallardoalba:
It can also be closed I think @bernt-matthias.

PR #2592, opened 2019-09-11
Repeatmasker in maker: rmlib does not exists

bgruening:
We got the following error:

`NotFound: cannot find 'repeatmasker' while searching for 'repeat_masking.repeat_source.repeatmasker.rmlib'`

xref: https://help.galaxyproject.org/t/maker-failed-to-communicate-with-remote-job-server/2064/2

ping @abretaud; does that look good for you?
fyi @gmauro


nsoranzo:
I don't know why, but the last test is failing.

abretaud:
Oh sorry, this got lost in the post-vacation flow of notifications
I think I'll push some commit to fix the problem another way, but the failing test comes from something strange in the conda package, I'm working on it

PR #2595, opened 2019-09-12
Table compute fixes

wm75:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/table_compute/scripts/table_compute.py
+++ b/tools/table_compute/scripts/table_compute.py
@@ -4,7 +4,7 @@ Table Compute tool - a wrapper around pandas with parameter input validation.
 """
 
 
-__version__ = "0.9.1"
+__version__ = "0.9.2"
 
 import csv
 import math
--- a/tools/table_compute/scripts/table_compute.py
+++ b/tools/table_compute/scripts/table_compute.py
@@ -265,7 +265,12 @@ if user_mode == "single":
         mode = params["element_mode"]
         if mode == "replace":
             replacement_val = params["element_replace"]
-            out_table = data.mask(bool_mat, replacement_val)
+            out_table = data.mask(
+                bool_mat,
+                data.where(bool_mat).applymap(
+                    lambda x: replacement_val.format(elem=x)
+                )
+            )
         elif mode == "modify":
             mod_op = Utils.getOneValueMathOp(params["element_modify_op"])
             out_table = data.mask(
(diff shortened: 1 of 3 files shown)


tools/table_compute/scripts/table_compute.py
```diff
@@ -265,7 +265,12 @@ def fun(vec):
         mode = params["element_mode"]
         if mode == "replace":
             replacement_val = params["element_replace"]
-            out_table = data.mask(bool_mat, replacement_val)
+            out_table = data.mask(
+                bool_mat,
+                data.where(bool_mat).applymap(
+                    lambda x: replacement_val.format(elem=x)
```
mtekman: does this work when the replacement value is not a string. Like 5?

tools/table_compute/table_compute.xml
```diff
@@ -424,10 +430,12 @@ Data = {
                     <when value="fulltable">
                         <conditional name="general" >
                             <param name="use" type="select" label="Operation" help="See Examples 5, 7, and 8 for usage" >
-                                <option value="melt" >Melt</option>
-                                <option value="pivot" >Pivot</option>
-                                <option value="custom" >Custom</option>
+                                <option value="transpose">Transpose</option>
```
mtekman: nice

tools/table_compute/table_compute.xml
```diff
@@ -220,24 +226,24 @@ Data = {
           #end if
         "element_inclusive": $singtabop.user.elem_val.inclusive_selection,
       #else:
-        "element_op": '$singtabop.user.elem_val.element_op.value',
+        "element_op": '$filter_type',
         ## Value is string or float, parsed in code later
         "element_value" : '$singtabop.user.elem_val.element_value',
       #end if
-        "element_mode": '$singtabop.user.element.mode.value',
```
mtekman: thanks for clearing up my crazy `singtabop` var names :D

tools/table_compute/scripts/table_compute.py
wm75: It works when *elem* (the previous content of the table) is not a string.  replacement_val, OTOH, is always a string.

In fact, your comment just made me realize the parameter help for the replacement value is wrong. Only the **filter value**, but not the **replacement value** is converted to numeric if possible, and that makes sense.
Going to fix the help ...

tools/table_compute/table_compute.xml
wm75: I put it at the very top of the list because it's a much simpler and possibly more common operation than Melt and Pivot. Are you saying you would prefer to put it below those two, or have you just misread the diff?

mtekman: Ah yes, it was at the top -- thanks

PR #2601, opened 2019-09-18
Add merge_heatmap_png

PrashantKuntala:
FOR CONTRIBUTOR:


astrovsky01:
This tools tests to see that the plots are of the same image dimensions, but is there an error if the axes are scaled differently for some reason?

PrashantKuntala:
Yes, 
If the scales are different for some reason, the final image has errors. But the final merged heat map would still be made if the dimensions are right.

bernt-matthias:
Just one first question: Why not merge based on the original data?

tools/merge_heatmap_png/merge_heatmap_png.xml
```diff
@@ -0,0 +1,62 @@
+<tool id="merge_heatmap_png" name="Merge Heatmap PNGs" version="1.0.0">
+    <description>from tag pileup datasets. </description>
+    <requirements>
+        <requirement type="package" version="3.7.3">python</requirement>
+        <requirement type="package" version="6.0.0">Pillow</requirement>
```
bgruening: lower case

tools/merge_heatmap_png/merge_heatmap_png.xml
```diff
@@ -0,0 +1,62 @@
+<tool id="merge_heatmap_png" name="Merge Heatmap PNGs" version="1.0.0">
+    <description>from tag pileup datasets. </description>
```
bgruening: please remove the dot and the space

PrashantKuntala:
@bgruening  I tried using imagemagick, but the results were not what we expect it to be.

bgruening:
> @bgruening I tried using imagemagick, but the results were not what we expect it to be.

Can you give more details? Thanks!

PrashantKuntala:
These are the test images I tried to merge and Below are some of the commands I tried :
inputs

- `composite -blend 100x100 -gravity center anti.png sense.png -alpha Set blend100x100.png`

[command reference](http://www.imagemagick.org/Usage/compose/#blend)
> Didn't work, white + anycolor = white; color1 + color2 = Blended color
compositeBlend

- `convert sense.png anti.png -evaluate-sequence mean mean_magick.png`

> Closest to the expected result I was able to achieve, unfortunately adds a 'ghostly blur' to the final image.
mean_output

- your suggested example didn't work either. It outputs one of the input images, guess its overlaying one image completely over the other.
`convert anti.png sense.png  -gravity center -composite -format jpg -quality 90 output.jpg`

My concern with the second command that almost works is that it would be harder to visualize in those cases where the input images have lower color intensities to begin with.

Am I missing anything ? Ideally, It would be great to be able to merge more than two. I'm open to suggestions. Thank you.

bernt-matthias:
Maybe something along the lines of: https://halberesford.com/how-to-make-an-images-background-transparent-using-imagemagick/ does the trick?

PR #2609, opened 2019-09-26
Rename output with intended ascii char

erasche:
FOR CONTRIBUTOR:

Noticed this when a processing script choked on a stacks_population output, and while yes said script should support unicode chars, I'm guessing this wasn't intended, and anyone who will in the future write manual workflows with yaml probably doesn't want to have to copy and paste an `ĥ`


peterjc:
Does this need a version bump?

nsoranzo:
@peterjc It's a bugfix, but modifies the tool outputs, so maybe yes.

erasche:
Only populations uses it, so bumped that.

PR #2615, opened 2019-09-27
STARsolo fix to consume paired list

mtekman:
Added ability to consume collection pairs
Removed strict barcode size restriction
Added collection test
Added test if R1 and R2 are same type for all pairs

FOR CONTRIBUTOR:

### DIFF
--- a/tools/rgrnastar/rg_rnaStarSolo.xml
+++ b/tools/rgrnastar/rg_rnaStarSolo.xml
@@ -2,7 +2,7 @@
     <description>mapping, demultiplexing and gene quantification for single cell RNA-seq</description>
     <macros>
         <import>macros.xml</import>
-        <token name="@WRAPPER@"></token>
+        <token name="@WRAPPER@">1</token>
     </macros>
     <expand macro="requirements"/>
     <expand macro="stdio" >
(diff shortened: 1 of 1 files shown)


tools/rgrnastar/rg_rnaStarSolo.xml
```diff
@@ -14,10 +14,40 @@
...
+        ## cDNA sequence(s) always go first, then barcode(s)
+        ## see: Section 3.1 of STAR manual for multiple inputs, and Section 13 for STARsolo inputs
+        #set $reads2 = ','.join([ '%s' % $x.input2 for $i,$x in enumerate($input_types.input_repeats)])
+        #set $reads1 = ','.join([ '%s' % $x.input1 for $i,$x in enumerate($input_types.input_repeats)])
+
+    #else if str($input_types.use) == "list:paired"
+
+        ## Again, check that the input pairs are of the same type
+        ## otherwise STARsolo will run for a long time and then error out
+        #for $i,$x in enumerate($input_types.input_repeats)
```
bgruening: nifty! But I'm not sure this is needed. This logic should be in Galaxy. So that such a collection can not be created.
ping @jmchilton to double-check, maybe I talk bs

jmchilton: It is a strong suggestion to not mix types but one can definitely at least use the API to create collections that don't have matching type - probably the GUI also in some cases.

bgruening: Should we prevent this in the UI instead of adding such a check here? Or do you think this check is ok?

mtekman: I could remove this altogether if I allow only paired or paired list input, but I want to keep the inputs as single dataset inputs within repeats because otherwise it will break my workflow :'(

PR #2637, opened 2019-10-15
Fix issues with STAR data manager index path and versioning

wm75:
FOR CONTRIBUTOR:

### DIFF
--- a/data_managers/data_manager_star_index_builder/data_manager_conf.xml
+++ b/data_managers/data_manager_star_index_builder/data_manager_conf.xml
@@ -12,7 +12,7 @@
                             out_file.extra_files_path is used as base by default
                             if no source, eg for type=directory, then refers to base 
                         -->
-                        <target base="${GALAXY_DATA_MANAGER_DATA_PATH}">rnastar/${version}/${dbkey}/${value}</target>
+                        <target base="${GALAXY_DATA_MANAGER_DATA_PATH}">rnastar/${version}/${dbkey}/${value}/${path}</target>
                     </move>
                     <value_translation>${GALAXY_DATA_MANAGER_DATA_PATH}/rnastar/${version}/${dbkey}/${value}/${path}</value_translation>
                     <value_translation type="function">abspath</value_translation>
--- a/data_managers/data_manager_star_index_builder/data_manager_conf.xml
+++ b/data_managers/data_manager_star_index_builder/data_manager_conf.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <data_managers>
-    <data_manager tool_file="data_manager/rna_star_index_builder.xml" id="rna_star_index_builder">
+    <data_manager tool_file="data_manager/rna_star_index_builder.xml" id="rna_star_index_builder" version="0.0.6">
         <data_table name="rnastar_index2_versioned">
             <output>
                 <column name="value" />
@@ -14,7 +14,7 @@
                         -->
                         <target base="${GALAXY_DATA_MANAGER_DATA_PATH}">rnastar/${version}/${dbkey}/${value}</target>
                     </move>
-                    <value_translation>${GALAXY_DATA_MANAGER_DATA_PATH}/rnastar/${version}/${dbkey}/${value}</value_translation>
+                    <value_translation>${GALAXY_DATA_MANAGER_DATA_PATH}/rnastar/${version}/${dbkey}/${value}/${path}</value_translation>
                     <value_translation type="function">abspath</value_translation>
                 </column>
                 <column name="with_gene_model" />
--- a/data_managers/data_manager_star_index_builder/data_manager/rna_star_index_builder.py
+++ b/data_managers/data_manager_star_index_builder/data_manager/rna_star_index_builder.py
@@ -18,13 +18,29 @@ def main():
     args = parser.parse_args()
 
     if args.dbkey in [ None, '', '?' ]:
-        raise Exception( '"%s" is not a valid dbkey. You must specify a valid dbkey.' % ( args.dbkey ) )
+        raise Exception(
+            '"%s" is not a valid dbkey. You must specify a valid dbkey.'
+            % ( args.dbkey )
+        )
 
     with_gene_model = "0"
     if args.with_gene_model:
         with_gene_model = "1"
 
-    data_manager_dict = {'data_tables': {args.data_table: [dict({"value": args.value, "dbkey": args.dbkey, "name": args.name, "path": args.subdir, "with_gene_model": with_gene_model, "version": args.index_version} )]}}
+    data_manager_dict = {
+        'data_tables': {
+            args.data_table: [
+                {
+                    "value": args.value,
+                    "dbkey": args.dbkey,
+                    "name": args.name,
+                    "path": args.subdir,
+                    "with_gene_model": with_gene_model,
+                    "version": args.index_version
+                }
+            ]
+        }
+    }
     open( args.config_file, 'w' ).write( json.dumps( data_manager_dict ) )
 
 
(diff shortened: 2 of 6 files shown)


wm75:
So, the previous version of the STAR data manager wrote a path that ended up one directory too high to the data table -> STAR couldn't discover the indices built for it :(

I've also fixed a few other things, like making it clearer which version strings need to be bumped when and the sample data table shipping with STAR itself.

wm75:
I don't think we need to bump the STAR version. Although some of the changes here are in the STAR tool itself, not the DM, but these are only inside comments or cosmetic (like the altered column names in the sample data table).
But please have a look yourself and tell me if you have a different opinion.

wm75:
Since I'm currently having problems testing this against Python 3:
@nsoranzo @mvdbeek will this, specifically:
https://github.com/galaxyproject/tools-iuc/blob/93138516106c74b1ccc70d2d946e6eaa29c7c1a3/data_managers/data_manager_star_index_builder/data_manager/rna_star_index_builder.xml#L19-L22

work in Python3? How does cheetah3 expand bytes parameters like `$target_dir` and `$subdir`?

nsoranzo:
I don't see the point of the `encode()` call. I also don't understand the point of creating `${target_directory}/${subdir}`, which duplicates the subdir part. (I'm aware you didn't write this code).

wm75:
the `subdir` part is used later to move stuff over to a final directory of the same name (the point is that using different settings you can generate different STAR indices for the same ref genome so that directory name was chosen to distinguish them).
As for the encoding thing, I don't see the point either. I was just reluctant to change it without need.

wm75:
Ok, I think the encoding/replace thing is not just weird, but actually wrong: it makes no sense to ever replace any characters in the `extra_files_path` determined by Galaxy since we do not want to write data to some arbitrary location.
I've removed this questionable part now and also let STAR create its index in `extra_files_path` directly. The necessary subdirectory is now created only in the final data destination.

wm75:
Unless someone spots yet another problem, I think this is ready.

data_managers/data_manager_star_index_builder/data_manager_conf.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <data_managers>
-    <data_manager tool_file="data_manager/rna_star_index_builder.xml" id="rna_star_index_builder">
+    <data_manager tool_file="data_manager/rna_star_index_builder.xml" id="rna_star_index_builder" version="0.0.6">
```
nsoranzo: Do you actually need a version here? I find it confusing.

wm75: ah, my memory. I had the same thought initially and had removed it when I updated the DM to STAR version 2.7 in .
Then I found that this version is actually used somewhere and causes an issue when it's missing, so I put it back in.
Just give me a moment to recall what that issue was ... :thinking:

wm75: I think I remember more or less accurately now:
This version gets incorporated into the `shed_data_manager_conf.xml` when you install the DM - seemingly without consequence when you do this for the first time.
However, when you have an existing install of an older version of that same DM, then what happens is that you end up with *two* `<data_manager>` records in the xml. Now if you don't provide that version, Galaxy seems to default to 0.0.1, which appears older than what it was (0.0.3) when it got bumped last.
This then seems to prompt Galaxy to keep using the old data table generation instructions, i.e., it won't use the new versioned table, but the old one.
This is totally confusing, IMO, but that's why I decided to just keep that version.
Oh, and 0.0.6 because the last version of the *wrapper* (before I changed it to reflect the IDX_VERSION it is building) was 0.0.5.

Executive summary of how I *think* the different versions should be used:
- bump VERSION to require a more up to date version of STAR
- bump IDX_VERSION only when that version of STAR will build a new kind of index structure
- bump the data_manager version only when the .loc file, its structure or the way of moving the data to its final destination changes

nsoranzo: Wow, thanks for digging this up, interesting!
We should document this, probably on https://galaxyproject.org/admin/tools/data-managers/data-manager-xml-syntax/ , do you want to open a PR?

nsoranzo:
@wm75 It seems the Tool Shed did not like the change in data table column names: https://travis-ci.org/galaxyproject/tools-iuc/jobs/598606040

wm75:
Argh. Strange that planemo let me update my own test repo on the testtoolshed.
I'll think about it later today.

wm75:
@nsoranzo I'm a bit lost with this error:
- First of all, I do not fully understand the error message because the only data table layouts changed here are in `tools/rgrnastar/tool_data_table_conf.xml.sample` and `tools/rgrnastar/tool_data_table_conf.xml.test`, yet the error seems to be about the `data_manager_star_index_builder` repository.
- Second, I tested `planemo shed_update --shed_target testtoolshed` of a simplified rgrnastar_test repo trying to change only those two data tables - and that worked without a problem (see https://testtoolshed.g2.bx.psu.edu/view/wolma/rgrnastar_test/77d511dd305d).
- Maybe things would work if I separated the changes to the DM and to the rgrnastar wrapper into two PRs, but I'm not sure. And I'm not sure how to do this now anyway since this PR here is already merged.
Should I open a new PR that reverts the DM changes, then maybe the upload of the data table changes would succeed, and I could then open another PR that reintroduces the DM changes?
Thanks for your help!

mvdbeek:
So what happens when you start Galaxy (or reload a data manager, which happens on planemo upload) is that all tool data tables for the same table get merged, provided that they contain the same columns. You'll have to use a new table or rename `with_gene_model ` back to `with-gtf`. This worked on the testtoolshed because either it didn't contain the conflicting table, or it did and it got cleaned.

wm75:
Just feels weird that I would have to do that just to satisfy planemo/Galaxy.

natefoo:
I think this column rename indeed has broken the data table when we upgraded on usegalaxy.org. I can change our column name in `tool_data_table_conf.xml`, but this will break old versions of the tool, correct?

wm75:
which *old* versions exactly? Versions prior to 2.7 should use an entirely separate table anyway.

natefoo:
My mistake, I didn't catch the table name change. Both versions on usegalaxy.org are pre-2.7.

wm75:
Ah, maybe you're affected by what I explained above in https://github.com/galaxyproject/tools-iuc/pull/2637#discussion_r335401505

Sorry, that this is causing you trouble. Tried to fix it when we realized the issues, but the column name thing prevented me from doing it.

natefoo:
We just updated to 2.7.2b. Should I wait to build new indexes?

wm75:
Well, you should be able to get things to work manually, but as a major issue, the DM writes an incomplete path to the data table (ending one level higher than where it puts the data) so you'd have to fix everyone of these paths.
Also, for 2.7.3 I'll have to introduce a new data table gain (because of the messed up column names). The 2.7.2 ref genomes will be reuseable, but you'll have to add them to the new table manually again.

So, if you don't have pressure from users that are asking for the latest STAR version, I would wait a little longer, then move to 2.7.3 directly and experience a proper ref genome install.
#2648 is waiting for some STARsolo fixes, but should be ready soon.

wm75:
@natefoo the newly merged #3151 bumps STAR to the latest version (2.7.5b) and fixes the data manager and the data table.
@bgruening has already built a first test index for .eu and everything looks good this time.
2.7.5b uses a new index format (2.7.4a was the version introducing it) so you'll have to rebuild all indexes no matter what, but at least it should just work now.

Sorry that this took so long, but STAR is a complicated beast.

natefoo:
Thank you @wm75 and @mtekman.

@davebx is getting the process started in galaxyproject/usegalaxy-tools#215.

PR #2638, opened 2019-10-15
add option to load gff from loc file

ieguinoa:
It's just wise to, when providing the genomes to map to, also provide access to quality annotation files that go with them.

### DIFF
--- a/tools/htseq_count/htseq-count.xml
+++ b/tools/htseq_count/htseq-count.xml
@@ -232,10 +232,14 @@
                 <param name="gff_source" value="history" />
             </conditional>
             <param name="gff_history" value="htseq-test.gff" />
-            <param name="advanced_options_selector" value="advanced" />
-            <param name="samout" value="Yes" />
-	        <conditional name="reference_source">
-                <param name="reference_source_selector" value="history" />
+            <conditional name="advanced_options">
+                <param name="advanced_options_selector" value="advanced" />
+                <conditional name="samout_conditional">
+                    <param name="samout" value="Yes" />
+	                <conditional name="reference_source">
+                        <param name="reference_source_selector" value="history" />
+                    </conditional>
+                </conditional>
             </conditional>
             <param name="ref_file" value="htseq-test_reference.fasta" />
             <output name="counts" file="htseq-test_counts.tsv" />
@@ -248,10 +252,12 @@
                 <param name="gff_source" value="history" />
                 <param name="gff_history" value="htseq-test.gff" />
             </conditional>
-            <param name="advanced_options_selector" value="advanced" />
-            <param name="nonunique" value="all" />
-            <param name="secondary_alignments" value="ignore" />
-            <param name="supplementary_alignments" value="ignore" />
+            <conditional name="advanced_options">
+                <param name="advanced_options_selector" value="advanced" />
+                <param name="nonunique" value="all" />
+                <param name="secondary_alignments" value="ignore" />
+                <param name="supplementary_alignments" value="ignore" />
+            </conditional>
             <output name="counts" file="htseq-test_counts_nonunique.tsv" />
             <output name="othercounts" file="htseq-test_othercounts.tsv" />
         </test>
--- a/tools/htseq_count/htseq-count.xml
+++ b/tools/htseq_count/htseq-count.xml
@@ -57,7 +57,13 @@
     --format=sam
     '$name_sorted_alignment_filename'
 
-    "$gfffile" | awk '{if ($1 ~ "no_feature|ambiguous|too_low_aQual|not_aligned|alignment_not_unique") print $0 | "cat 1>&2"; else print $0}'
+    #if $gff_source_selector.gff_source=="standard":
+        $gff_source_selector.gff_cache.fields.path
+    #else if $gff_source_selector.gff_source=="history":
+        $gff_source_selector.gff_history
+    #end if
+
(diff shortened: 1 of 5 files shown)


ieguinoa:
done, but also upgraded the tool version since there is a new one that just tested and it passed ok.

tools/htseq_count/htseq-count.xml
```diff
@@ -74,7 +80,22 @@
 
     <inputs>
         <param format="sam,bam" name="samfile" type="data" label="Aligned SAM/BAM File"/>
-        <param format="gff" name="gfffile" type="data" label="GFF File"/>
+        <conditional name="gff_source_selector">
+            <param name="gff_source" type="select" label="Will you select a reference gff from your history or use a standard gff file?" help="Standard gff files are available for model organisms">
+              <option value="standard" selected="true">Use a standard gff file</option>
```
bgruening: Can we call this cached GFF file?
Can we make history still default?

tools/htseq_count/htseq-count.xml
```diff
@@ -161,42 +182,48 @@
             <filter>advanced_options['advanced_options_selector'] == 'advanced' and  advanced_options['samout_conditional']['samout'] == "Yes"</filter>
         </data>
     </outputs>
-
     <tests>
         <test expect_num_outputs="2">
             <param name="samfile" value="htseq-test.sam" />
-            <param name="gfffile" value="htseq-test.gff" />
+            <param name="reference_source_selector" value="history" />
```
bgruening: this line is wrong ... better to use the more explicit syntax.

```
<conditional name="gff_source_selector">
</conditional>
```

tools/htseq_count/htseq-count.xml
```diff
@@ -207,14 +234,23 @@
...
             <param name="advanced_options_selector" value="advanced" />
             <param name="nonunique" value="all" />
             <param name="secondary_alignments" value="ignore" />
             <param name="supplementary_alignments" value="ignore" />
             <output name="counts" file="htseq-test_counts_nonunique.tsv" />
             <output name="othercounts" file="htseq-test_othercounts.tsv" />
         </test>
+        <test expect_num_outputs="2">
+            <param name="samfile" value="htseq-test.sam" />
+            <param name="reference_source_selector" value="history" />
```
bgruening: that line is not needed? Can you please use the more explicit syntax?

tools/htseq_count/tool_data_table_conf.xml.sample
```diff
@@ -1,8 +1,7 @@
-<!-- Use the file tool_data_table_conf.xml.oldlocstyle if you don't want to update your loc files as changed in revision 4550:535d276c92bc-->
 <tables>
-    <!-- Location of SAMTools indexes and other files -->
```
bgruening: the sam_fa_indexes index is still part of this tool ... however, I do not know if this is working....

bgruening:
@ieguinoa ping

tools/htseq_count/tool_data_table_conf.xml.sample
bgruening: ping @ieguinoa

bgruening:
@lparsons can you please look at this one?

tools/htseq_count/htseq-count.xml
```diff
@@ -1,7 +1,7 @@
-<tool id="htseq_count" name="htseq-count" version="0.9.1" profile="16.04">
+<tool id="htseq_count" name="htseq-count" version="0.11.1" profile="16.04">
     <description> - Count aligned reads in a BAM file that overlap features in a GFF file</description>
     <requirements>
-        <requirement type="package" version="0.9.1">htseq</requirement>
+        <requirement type="package" version="0.11.1">htseq</requirement>
```
lparsons: Might want to bump to version 0.11.2 now: https://htseq.readthedocs.io/en/latest/history.html#version-0-11-2

tools/htseq_count/tool_data_table_conf.xml.sample
```diff
@@ -1,8 +1,7 @@
...
-    <table name="sam_fa_indexes" comment_char="#">
-        <columns>line_type, value, path</columns>
-        <file path="tool-data/sam_fa_indices.loc" />
+    <!-- Locations of GFF files -->
+    <table name="all_gff" comment_char="#" allow_duplicate_entries="False">
+        <columns>value, dbkey, name, path</columns>
+        <file path="tool-data/all_gff.loc" />
```
lparsons: Also, this probably isn't something to address in this PR, but would be nice to unify these annotation files and those used by `featurecounts` (at least where possible). https://github.com/galaxyproject/tools-iuc/blob/master/tools/featurecounts/tool_data_table_conf.xml.sample

tools/htseq_count/tool_data_table_conf.xml.sample
lparsons: Yes, at a minimum, this needs to stay here. Would be good to add a test that uses it.

tools/htseq_count/tool_data_table_conf.xml.sample
ieguinoa: definitely, there are several tools that use/can make use of cached annotation files and I've been trying to come up with a way to add an unified data table for this in Galaxy + a data manager that goes with it.

PR #2645, opened 2019-10-16
Guarantee header line in snpeff databases output

wm75:
FOR CONTRIBUTOR:


wm75:
The current version of the tool fails to grep the header line for include patterns.
When the search pattern is not found, grep returns with exit code 1, which causes the job run to be detected as failed, which is not an ideal situation for the end user.

PR #2649, opened 2019-10-17
Add tool to dynamically downsample BAM files

davebx:
FOR CONTRIBUTOR:


bgruening:
Dave do you think we can integrate this functionality into https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/peterjc/sample_seqs/sample_seqs/0.2.5

ping @peterjc

davebx:
I don't know how feasible that would be, since @peterjc's tool operates on FASTQ files, whereas this takes SAM or BAM inputs.

pvanheus:
I agree with @davebx : the difference between this and @peterjc 's tool is a bit like filtering on read quality versus filtering on alignment quality. In the incoming fastq file you do not know how many reads will map to the reference sequence. Filtering on BAM to produce a lower coverage *after* mapping then makes sense. Merging the two tools would perhaps confuse users, perhaps not. It would, however, mean that two sets of logic are merged and also that two sets of dependencies are merged.

BTW another tool somewhat similar (but not with all the functionality) to @peterjc 's tool is [seqtk sample](https://github.com/galaxyproject/tools-iuc/blob/master/tools/seqtk/seqtk_sample.xml).

pvanheus:
Ps thanks for flagging with `reviewer needed` - I had not seen that on tools-iuc PRs before and it helped focus my eyes.

peterjc:
I could see an argument for extending https://github.com/peterjc/pico_galaxy/blob/master/tools/sample_seqs/sample_seqs.py to handle (unmapped) BAM files.

Probably would need a pysam dependency, not trivial, but it would certainly be possible. Still, an existing SAM/BAM sampling tool would likely be a better fit?

bgruening:
My argument is that from a user point of view a general downsampling tool for sequences would be nice. If sequences are in BAM, FASTQ, FASTA, SAM ... a user should not care.

pvanheus:
@bgruening I still feel there are different kinds of downsampling. :)

bgruening:
Any reasons? In the end its @peterjc and @davebx call I guess. I would like to keep the number of tools for us (to maintain) and for the users low.

pvanheus:
Let's think through this from a user interface perspective. This is the interface of @peterjc 's sub-sample tool:

image

So if this PR was merged into that, the help text would need extension. Would the existing UI options - Nth, Percentage, N - apply for mapped reads? I think the main aim of @davebx 's tool is to implement a depth-N approach. This is somewhat similar to to the 3rd option of the existing tool? I think the issue here is @bgruening 's concern with adding yet another thing to the tool menu vs adding complexity to the UI of the sub sample tool.

@peterjc suggests that there might be another tool that is better to build on. Peter did you have anything specific in mind?

bernt-matthias:
My preferred solution would be to use the available samtools view wrapper (Parameter `Specify a fraction to keep`). I think it should be possible to input the required parameter from the results of of the samtools depth wrapper..?

bernt-matthias:
Seems to be a mini 3 step workflow .. if one can express the computations done in the awk call of the PR in Galaxy.

bgruening:
I feel sorry to block this PR. @peterjc would you accept a PR that add this functionality to your tool? If not I think we should merge this. I do not consider `awk` as a standard tool that every user can and should use. It's too complicated imho.

nekrut:
As it turns out, this can nicely be done with workflow expression tools

bernt-matthias:
wonderful @nekrut : could you share the workflow on a public instance? I never found the time to work with workflow expression tools and would like to see this is action. 

Question is if a separate tool is then needed?

bernt-matthias:
One downside is that such a workflow can't be distributed as easily as a tool.

Would be nice to be able to have automatically tested workflows also here in the repo and tested by CI.

wm75:
I do understand that this is meant to downsample coverage of small genomes, but how does the implementation scale to larger genomes?
Since you're using the median you have to store all the per-base coverages, then sort them. Would that still work with reasonable memory requirements for, say, yeast, C. elegans, human? Just to have a rough idea how generally this tool could be used.

If it's restricted to just viral/bacterial genomes then maybe that's an argument *against* including it in a more generic tool?
Similarly, a workflow like @nekrut suggested may be better then because it would leave it up to the user to calculate the arithmetic, geometric or whatever mean as a scaling factor?

As a small side note: samtools depth without the `-a` option skips zero-coverage positions. This would not affect the tool much for its intended usecase (highly covered small genomes), but could introduce a bias with low or partial coverage. Maybe an option like `Skip zero-coverage positions in Factor calculation` would be nice?

wm75:
Maybe an implementation making use of a dictionary of counts or sth like a Python collections.Counter structure could boost efficiency for larger genomes (since coverages are a limited number of integers)?

mvdbeek:
> One downside is that such a workflow can't be distributed as easily as a tool.

I agree, though at least users can share them without needing the admin. In principle workflows can also be loaded up in the tool panel and look like tools. But for sure a single tool is easier to use.

pvanheus:
@nekrut I have tried to create a workflow like the one you proposed but I am stuck on a few items:

1. The "compute" step fails for me because it doesn't see the datamash output as having columns.
2. More generally I can't figure out how to feed parameters as inputs in workflows - two parameters are needed, firstly the target depth and secondly the computed parameter that fees into the downsample tool.
3. Of the usegalaxy.* servers only usegalaxy.eu has all the tools I'd want to use for this (samtools depth, datamash, compute, picard downsample).

bernt-matthias:
@pvanheus note that also `samtools view` allows to downsample .. so no need to switch toolbox

pvanheus:
@bernt-matthias noted. I don't see `samtools view` on any of the usegalaxy.* servers but it looks like a good option.

bgruening:
Is this one still needed?

bgruening:
@davebx ping

PR #2651, opened 2019-10-18
bcftools_consensus: add 'rename' parameter

pvanheus:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/bcftools/bcftools_consensus.xml
+++ b/tools/bcftools/bcftools_consensus.xml
@@ -1,5 +1,5 @@
 <?xml version='1.0' encoding='utf-8'?>
-<tool name="bcftools @EXECUTABLE@" id="bcftools_@EXECUTABLE@" version="@TOOL_VERSION@+galaxy1">
+<tool name="bcftools @EXECUTABLE@" id="bcftools_@EXECUTABLE@" version="@TOOL_VERSION@+galaxy2">
     <description>Create consensus sequence by applying VCF variants to a reference fasta file</description>
     <macros>
         <token name="@EXECUTABLE@">consensus</token>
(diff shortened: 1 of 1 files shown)


pvanheus:
This tool adds the ability to rename the consensus created by bcftools consensus. The common use case for this is taking a collection (list) of VCF (e.g. one VCF for each bacterial sample), and creating a consensus from each member of the list (combined with a reference sequence). In this case the new consensuses (consensi?) should not have the name of the reference sequence consensus but rather inherit a name from the VCF used to generate them. This works in an elegant way for list collections with items named after sample names.

Longer-term TODO: motivate for this option to be added to the bcftools code so that the use of awk is no longer necessary. For now the pipe to awk handles the renaming.

tools/bcftools/bcftools_consensus.xml
```diff
@@ -38,8 +39,14 @@ ${section.iupac_codes}
 #end if
 
 ## Primary Input/Outputs
---output "$output_file"
+#if str($rename) == "no"
+    --output "$output_file"
```
bgruening: please use singe-quotes

pvanheus: done.

tools/bcftools/bcftools_consensus.xml
```diff
@@ -25,21 +26,27 @@ bcftools @EXECUTABLE@
...
 #end if
 
 ## Primary Input/Outputs
---output "$output_file"
+#if str($rename) == "no"
+    --output '$output_file'
+#end if
 @INPUT_FILE@
+#if str($rename) == "yes":
+    #set basename=$input_file.name
```
mvdbeek:
```suggestion
    #set basename=$input_file.element_identifier
```

pvanheus: done. what exactly is the distinction here? I thought that historically 'element_identifier' was only for list elements?

PR #2655, opened 2019-10-21
Added breseq and accompanying gdtools utility

hepcat72:
See: https://github.com/barricklab/breseq/issues/209

FOR CONTRIBUTOR:


tools/breseq/.shed.yml
```diff
@@ -0,0 +1,10 @@
+name: breseq
+owner: hepcat72
```
bgruening: iuc

tools/breseq/breseq.xml
```diff
@@ -0,0 +1,171 @@
...
+                #end for
+            #else:
+	        #for $i, $ref in enumerate( $reference.fixed_genome )
+                    #if $ref
+                        --reference $ref.fields.path
+                    #end if
+                #end for
+            #end if
+
+            #for $i, $s in enumerate( $fastqs )
```
bgruening: the enumerate is not needed here?

tools/breseq/breseq.xml
```diff
@@ -0,0 +1,171 @@
...
+    <outputs>
+        <data format="html" name="report" label="${tool.name} on ${on_string}: Report (Webpage)" />
+        <data format="txt" name="log" label="${tool.name} on ${on_string}: Log" />
+        <data format="txt" name="output" label="${tool.name} on ${on_string}: gdtools Output" />
+        <data format="zip" name="zip_output" label="${tool.name} on ${on_string}: Gzipped Output" />
+    </outputs>
+
+    <tests>
+        <test>
+            <param name="source" value="history" />
```
bgruening: can you please add a test for the genbank data table as well?

tools/breseq/breseq.xml
```diff
@@ -0,0 +1,171 @@
...
+        </param>
+
+        <param name="name" argument="--name" type="text" value="" label="Analysis Name" help="Human-readable name of the analysis run for output (DEFAULT=none)." />
+
+        <param name="predict_junctions" type="boolean" truevalue="" falsevalue="--no-junction-prediction" checked="true" label="Predict Junctions" help="Predict new sequence junctions (default).  --no-junction-prediction is supplied if 'No' is selected.  Otherwise, there is no flag." />
+    </inputs>
+
+    <outputs>
+        <data format="html" name="report" label="${tool.name} on ${on_string}: Report (Webpage)" />
+        <data format="txt" name="log" label="${tool.name} on ${on_string}: Log" />
```
bgruening: can we make log and zip optional by default?

tools/breseq/breseq.xml
```diff
@@ -0,0 +1,171 @@
...
+            <output name="output" file="gdout.txt" lines_diff="8" />
+            <output name="zip_output" file="gzipout.zip" compare="sim_size" delta="100" />
+        </test>
+    </tests>
+        
+    <help>
+        <![CDATA[
+            breseq (pronounced: \\brēz-ˈsēk\\ or breeze-seq) is a computational pipeline for the analysis of short-read re-sequencing data (e.g. Illumina, 454, IonTorrent, etc.). It uses reference-based alignment approaches to predict mutations in a sample relative to an already sequenced genome. breseq is intended for microbial genomes (<10 Mb) and re-sequenced samples that are only slightly diverged from the reference sequence (<1 mutation per 1000 bp).
+
+**IMPORTANT:** *In order to properly view breseq's html output in Galaxy, breseq needs to be whitelisted by your galaxy admin.  Otherwise, the results will not be displayed properly and interactive controls will not function.*
```
bgruening: this should not be added to the tool, but rather to a readme file next to the tool ... but in general, the Galaxy Admin should know about this and it is not needed.

tools/breseq/breseq.xml
```diff
@@ -0,0 +1,171 @@
...
+                </assert_contents>
+            </output>
+            <output name="output" file="gdout.txt" lines_diff="8" />
+            <output name="zip_output" file="gzipout.zip" compare="sim_size" delta="100" />
+        </test>
+    </tests>
+        
+    <help>
+        <![CDATA[
+            breseq (pronounced: \\brēz-ˈsēk\\ or breeze-seq) is a computational pipeline for the analysis of short-read re-sequencing data (e.g. Illumina, 454, IonTorrent, etc.). It uses reference-based alignment approaches to predict mutations in a sample relative to an already sequenced genome. breseq is intended for microbial genomes (<10 Mb) and re-sequenced samples that are only slightly diverged from the reference sequence (<1 mutation per 1000 bp).
```
bgruening: can you break the lines so reading/reviewing is easier?

tools/breseq/gdtools_annotate.xml
```diff
@@ -0,0 +1,155 @@
...
+
+            <output name="report" file="gdtoolsout.html" compare="sim_size" delta="100" />
+        </test>
+    </tests>
+        
+    <help>
+        <![CDATA[
+             Annotate a GenomeDiff file (generated by breseq) with information about mutations (what genes they affect, amino acid substitutions, etc.) If multiple input files are provided, then also COMPARE the frequencies for identical mutations across samples.
+
+**IMPORTANT:** *In order to properly view gdtools_annotate's html output in Galaxy, gdtools_annotate needs to be whitelisted by your galaxy admin.  Otherwise, the results will not be displayed properly.*
```
bgruening: pleaes remove

tools/breseq/gdtools_annotate.xml
```diff
@@ -0,0 +1,155 @@
...
+                </param>
+            </when>
+            <when value="history">
+                <param name="own_genome" argument="--reference" type="data" format="fasta,genbank" multiple="true" label="Fasta or Genbank Reference(s)" />
+            </when>
+        </conditional>
+
+        <param name="gds" type="data" format="txt" multiple="true" label="GenomeDiff (gd) Files" help="Files as produced by breseq" />
+
+        <param argument="--format" name="html" type="boolean" checked="true" label="Output an HTML File" help="Note, this tool will run as many times (serially) as the number of output formats selected." />
```
bgruening: this could be a multple=true select box

tools/breseq/breseq.xml
```diff
@@ -0,0 +1,171 @@
...
+
+- ``data/reference.fasta`` (file with reference genome: can be used in eg. IGV browser)
+- ``data/reference.gff`` (file with genomic annotations: can be used in eg. IGV browser)
+- ``data/areference.bam`` (file with read alignments: can be used in eg. IGV browser)
+- ``data/unmatched.*`` (files with read that failed to align: can be used to build an assembly or to eg. blast against NCBI)
+        ]]>
+    </help>
+
+    <citations>
+        <citation type="doi">10.1007/978-1-4939-0554-6_12</citation>
```
bgruening: this could be moved in a macro file and shared acroos the 2 tools, same goes for the requirements

hepcat72:
> @hepcat72 would it make sense to merge both tools into one?
> Small comments inline. Thanks!

Good idea. Will do.

tools/breseq/gdtools_annotate.xml
hepcat72: You know, I tried a multiple select box first and since I discovered I couldn't get back a list of selected options (because I wanted to loop on them), I went with a series of booleans. I'll give it another go.

tools/breseq/breseq.xml
hepcat72: I'll wrap both in 1 tool.

tools/breseq/breseq.xml
hepcat72: Will be in next push.

tools/breseq/breseq.xml
hepcat72: Will be in next push.

tools/breseq/breseq.xml
hepcat72: Sure.  Will be in next push.

tools/breseq/gdtools_annotate.xml
hepcat72: Will be reflected in next push.

tools/breseq/breseq.xml
hepcat72: Will be reflected in next push.

tools/breseq/gdtools_annotate.xml
bgruening: I think something like that should work `for $elem in str($foo_select).split(','):`

hepcat72: Oh cool, thanks!  I'll try that out.

tools/breseq/breseq.xml
hepcat72: Somewhat stuck on the genbank data table test.  I haven't messed with the builtin reference stuff before.  Do you have any hints you can give me on what this error in the tool's output dataset means?: `'SelectToolParameterWrapper' object is not iterable`

bgruening: I guess you need to cast it to str() or list(). Probably str() and then split(',')

hepcat72: I'll try that.  I have a separate question, if you don't mind...

I'm trying to use the multiple select list for output formats like you suggested in another comment, but I'm having trouble with the filters for the optional output formats.  I'm trying this:

```
<data format="tabular" name="tabdelim" label="${tool.name} on ${on_string}: Tab-delimited Output">
    <filter>len(list(filter(lambda x: str(x) == "tsv", str(formats).split(',')))) > 0</filter>
</data>
```

but the history item never shows when I select (for example) the tsv list item.  As a sanity check, I added this at the end of the command:

```
#set $testone = len(list(filter(lambda x: str(x) == "tsv", str($formats).split(','))))
#set $testtwo = len(list( str($formats).split(',') ))
&& echo $testone $testtwo
```

and the command on the info page has `echo 1 2` at the end (when I select tsv and another format), so the code passes proof of concept, but I don't know why it's not working in the filter.

bgruening: would this work? `<filter>'tsv' in str(formats).split(',')</filter>`

hepcat72: Ugh.  Still not getting output datasets.  Let me push what I've got and change the PR to WIP.  I haven't merged the tools yet.  I'm working on this output format list, and have yet to merge the 2 tools and finish the builtin reference test.

bgruening: Oh its easier: https://github.com/bgruening/galaxytools/blob/master/tools/antismash/antismash.xml#L103

hepcat72: Oh gee.  Yeah, that did it.

tools/breseq/gdtools_annotate.xml
```diff
@@ -0,0 +1,142 @@
...
+                </param>
+            </when>
+            <when value="history">
+                <param name="own_genome" argument="--reference" type="data" format="fasta,genbank" multiple="true" label="Fasta or Genbank Reference(s)" />
+            </when>
+        </conditional>
+
+        <param name="gds" type="data" format="txt" multiple="true" label="GenomeDiff (gd) Files" help="Files as produced by breseq" />
+
+        <param argument="--format" name="formats" type="select" label="Output Formats" multiple="true" help="Note, this tool will run as many times (serially) as the number of output formats selected.">
```
nsoranzo: A `<param type="select" multiple="true"/>` is optional by default, so a user can select no option. Is that what you want?

Also, I'd suggest to use checkboxes:
```suggestion
        <param argument="--format" name="formats" type="select" multiple="true" display="checkboxes" label="Output formats" help="Note, this tool will run as many times (serially) as the number of output formats selected.">
```

hepcat72: I want at least 1 selected, so I have a filter that essentially defaults to html if no option is selected.  Is there a minimum attribute?

nsoranzo: Just use `optional="false"`:

```suggestion
        <param argument="--format" name="formats" type="select" multiple="true" optional="false" display="checkboxes" label="Output formats" help="Note, this tool will run as many times (serially) as the number of output formats selected.">
```

tools/breseq/breseq.xml
hepcat72: How do I retrieve the path from the data table when multiple selections are allowed?  When I do this:

```
#for $ref in str($reference.fixed_genome).split(','):
    #if $ref
        --reference str($ref.fields.path)
    #end if
#end for
```

I get the error:

```
cannot find 'fields' while searching for 'ref.fields.path'
```

hepcat72: OK.  Never mind.  I figured out a solution, though it's ugly and I'm hoping someone will see it and supply a simpler solution after I push it.

tools/breseq/gdtools_annotate.xml
hepcat72: This will be in the next push.

hepcat72:
I have the 2 tools merged into one and an nearing a push, but I have a quick question.  The gdtools tool takes their own format (".gd") files.  I have a data input of type text.  Can I limit what pops up in the select list to gd files or would a separate file type called "gd" have to be created elsewhere in galaxy to accomplish that?

hepcat72:
I also discovered that the phylip output format option requires a minimum number of input gd files of 2.  Is there a way to require that before the tool can execute?  And is there a way to gray out that output format option if fewer than 2 files are supplied?

hepcat72:
OK.  All the suggestions have been implemented and I fixed an issue I found where phylip output format requires a minimum of 2 input GD files.  I also changed the gd format to tabular.

I merged the 2 tools into one and made all output formats optional (but at least 1 required) and displayed as a select list using checkboxes.  I added a test for the builtin reference.

Request: I would like feedback on the way I handle multiple built-in references in the command section.  It's a pretty ugly filter/lambda, but I didn't know how else to do it.  I got the code from:

https://galaxyproject.org/admin/tools/data-tables/

tools/breseq/breseq.xml
```diff
@@ -0,0 +1,362 @@
...
+                </section>
+
+            </when>
+        </conditional>
+
+    </inputs>
+
+    <outputs>
+        <data format="tabular" name="output" label="${tool.name} on ${on_string}: Variants (GenomeDiff)">
+            <filter>run['mode'] == 'detect' and 'gd' in run['output_options']['formats']</filter>
```
nsoranzo: All the filters check that `run['mode'] == 'detect'`, is that correct?

tools/breseq/breseq.xml
```diff
@@ -0,0 +1,362 @@
...
+                        <option value="log">Log (Text)</option>
+                    </param>
+                </section>
+
+            </when>
+            <when value="annotate">
+
+                <param name="gds" type="data" format="tabular" multiple="true" optional="false" label="GenomeDiff (gd) Files" help="Files as produced by breseq" />
+
+                <section name="output_options" title="Output Options" expanded="false">
```
nsoranzo: You can use macros with `<yield/>` to reduce code duplication.

tools/breseq/breseq.xml
hepcat72: Hmm, that may be unnecessary now, now that I think about it. I’ll look tomorrow.

tools/breseq/breseq.xml
hepcat72: Ok. Cool. I’ll familiarize myself with that and try it out.

tools/breseq/breseq.xml
hepcat72: Alright.  Some of the `run['mode'] == 'detect'` and `run['mode'] != 'detect'` checks are unnecessary, although there are 2 in each case that are, only in order to differentiate between *annotated* and *unannotated* variant output by the `gdtools` and `breseq` executables respectively.  The only ones that need to check `run['mode']` are the ones where the `run['output_options']['formats']` are the same (`html` and `gd`).  I could change the values in the formats select list for the `breseq` interface to make them unique and thus not need to check `run['mode']` in any of the filters.  I'll go with whatever you recommend.

nsoranzo: I think just removing the unnecessary checks should be fine.

hepcat72: OK. This will be reflected in the next push. Thanks.

hepcat72:
I have a question regarding this error in the tests:

```
JobOutputsError: History item  different than expected, difference (using sim_size):
1624( /tmp/tmpu5x1czgzipout.zip v. /tmp/tmpwwo9Y9gzipout.zip )
1625Files /tmp/tmpwwo9Y9gzipout.zip=689004b but /tmp/tmpu5x1czgzipout.zip=747716b - compare by size (delta=100) failed
```

This test passes on my system, but fails in the remote test.  The size difference in the remote test is 58kb.  I assume that the `gzip` default compression options somehow differ or something.  What is the best way to deal with testing gzip output that avoids this issue?

tools/breseq/breseq.xml
hepcat72: I have this working now.  It will be in the next push.

nsoranzo:
> What is the best way to deal with testing gzip output that avoids this issue?

Increase the `delta`?

bgruening:
With the next versoin of Galaxy (20.01) you can use:
```
         <assert_contents>
            <has_size value="20000" delta="50"/>
         </assert_contents>
```

hepcat72:
> > What is the best way to deal with testing gzip output that avoids this issue?
> 
> Increase the `delta`?

I was worried that the output was not correct since the delta was so big, so I tried changing the size test to a has_archive_member assertion, but it yields the following error.  At first, I thought it was my handling of the section tag I had added, but I just tried removing the section tags temporarily and I still get the same error:

```
  File "/private/var/folders/8d/z84mms353qb2n0r3n2jhww0m0000gq/T/tmphcg1ync8/galaxy-dev/test/functional/test_toolbox.py", line 99, in test_tool
    self.do_it(tool_version=tool_version, test_index=test_index)
  File "/private/var/folders/8d/z84mms353qb2n0r3n2jhww0m0000gq/T/tmphcg1ync8/galaxy-dev/test/functional/test_toolbox.py", line 36, in do_it
    verify_tool(tool_id, self.galaxy_interactor, resource_parameters=resource_parameters, test_index=test_index, tool_version=tool_version, register_job_data=register_job_data)
  File "/private/var/folders/8d/z84mms353qb2n0r3n2jhww0m0000gq/T/tmphcg1ync8/galaxy-dev/lib/galaxy/tools/verify/interactor.py", line 789, in verify_tool
    raise e
JobOutputsError: 'NoneType' object has no attribute 'read'
```

I've tried a dozen or so different ways to define the test, but each one has yielded an error suggesting that the previous test xml structure was correct.  I figure I must have something wrong in here though:

```
             <output name="zip_output" file="out.tar.gz">
                <assert_contents>
                    <has_archive_member path="results/output/output.gd" />
                </assert_contents>
            </output>
```

(NOTE: I tried the above both with & without ` file="out.tar.gz"`.)

I started supplying `--job_output_files testdir` and it appears that the `.tar.gz` files is not getting copied to the testdir (although I don't see an error from the tar command).

So I changed back to the compare_size/delta method and now the testdir gets an output `.tar.gz` file and the test succeeds, so there has to be something wrong with my has_archive_member item.  I cannot find any working examples of the has_archive_member usage, so I can't tell if I'm doing it wrong.

The problem is that with such a big file size difference on the remote tests, I would like a more conclusive test (such as what has_archive_member should provide) instead of just create a huge delta.  I can tell it works on my system, but I'm just not certain about the remote system.

*UPDATE*: I finally got `has_archive_member ` to be included in the test **without error**, however, I had to use *BOTH* sim_size/delta and has_archive_member:

```
            <output name="zip_output" file="out.tar.gz" compare="sim_size" delta="100000">
                <assert_contents>
                    <has_archive_member path="results/output/output.gd" />
                </assert_contents>
            </output>
```

Can anyone explain why that works?

hepcat72:
So, I pushed out the changes which all pass now, however I've tested a few things in order to sanity-check the test that was failing before.  Here are the things I tested WRT the test of the zip_output output item in the first test and what happens:

1. Test using has_archive_member with correct file name (alone) = FAILs with error `'NoneType' object has no attribute 'read'`
2. Test using size check (alone) = PASS
3. Test using has_archive_member with **correct** file name **and** size check = PASS
4. Test using has_archive_member with **incorrect** file name **and** size check = FAILs with error `'NoneType' object has no attribute 'read'`

Here is the test in each of those cases:

1:
```
<output name="zip_output">
    <assert_contents>
        <has_archive_member path="results/output/output.gd" />
    </assert_contents>
</output>
```

2:
```
<output name="zip_output" file="out.tar.gz" compare="sim_size" delta="100000" />
```

3:
```
<output name="zip_output" file="out.tar.gz" compare="sim_size" delta="100000">
    <assert_contents>
        <has_archive_member path="results/output/output.gd" />
    </assert_contents>
</output>
```

4:
```
<output name="zip_output" file="out.tar.gz" compare="sim_size" delta="100000">
    <assert_contents>
        <has_archive_member path="results/output/badfilename.gd" />
    </assert_contents>
</output>
```

I would like to remove these from the zip_output test: `file="out.tar.gz" compare="sim_size" delta="100000"`, but if I do (like in case 1, I get that error `'NoneType' object has no attribute 'read'`).  So what is wrong in case 1?

tools/breseq/breseq.xml
```diff
@@ -0,0 +1,353 @@
+<tool id="breseq" name="breseq" version="0.1.0">
```
lparsons: Make version a macro token. Break into BreSeq version and Galaxy version (see https://github.com/galaxyproject/tools-iuc/blob/a4cee4cac5097029188d836d5b904b605dbb943d/tools/dropletutils/dropletutils.xml#L2-L6)
```
<tool id="breseq" name="breseq" version="@PACKAGE_VERSION@+@GALAXY_VERSION@" >
...
    <macros>
        <token name="@PACKAGE_VERSION@" >0.34.0</token>
        <token name="@GALAXY_VERSION@" >1</token>
```

tools/breseq/breseq.xml
```diff
@@ -0,0 +1,353 @@
...
+
+    <description>find mutations in haploid microbial genomes</description>
+
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+
+    <requirements>
+        <requirement type="package" version="0.34.0">breseq</requirement>
```
lparsons: Use macro token for package version:
```
         <requirement type="package" version="@PACKAGE_VERSION@">breseq</requirement>
```

tools/breseq/breseq.xml
```diff
@@ -0,0 +1,353 @@
...
+                    $ref_opts
+
+                    #for $s in $run.gds:
+                        ${s}
+                    #end for
+
+                    && cp
+
+                    #if $o == 'html':
+                        annotated.html '$annreport'
```
lparsons: Generating the output to one name, and they copying to the final name was a bit confusing to me. Perhaps moving the conditional to top of the loop to find the output name and then using that directly in the command would be more readable.
```
#if $o == 'html':
     output_file = $annreport
```
...
```
- o '$output_file'
```

tools/breseq/breseq.xml
```diff
@@ -0,0 +1,353 @@
...
+            #set $ref_opts = ""
+            #if str($reference.source) == "history":
+                #for $ref in $reference.own_genome:
+                    #if $ref
+                        #set $ref_opts = $ref_opts + " --reference '" + str($ref) + "'"
+                    #end if
+                #end for
+            #else:
+                #for $rowid in str($reference.fixed_genome).split(','):
+                    #set $reflist = [filter( lambda x: str( x[0] ) == rowid, $__app__.tool_data_tables['genbank_files'].get_fields() )[0][-1]]
```
lparsons: Phew! Nice job sorting this out!

hepcat72: I lifted the code directly from https://galaxyproject.org/admin/tools/data-tables/

tools/breseq/breseq.xml
hepcat72: I put the conditional under the -o for the next push.  Let me know if you think it's better at the top.

tools/breseq/breseq.xml
hepcat72: Sure.  This will be in the next push.

tools/breseq/breseq.xml
hepcat72: Got it.  Next push.

tools/breseq/breseq.xml
bgruening: Sorry, we can not do this. Tools should not rely on Galaxy internals. Why can you not use $genbank_files.fields.path and operate on this?

hepcat72: So I ended up with this because the builtin reference select list is a multiple select list. I couldn’t find an example and what I tried yielded an error. I commented that I was getting an error and I asked how to loop on selected items in the data table select list. I was told I could cast to a string and split on comma, but I could not get .fields.path from the result, so I searched the documentation and found this code on the data tables page.

If I understand your suggestion, then it looks like it will get all (?) paths from the data table. Are you suggesting I put that inside the filter call? I will definitely make that change. Unless there’s an even more straightforward to get all the paths of the selected table IDs?

hepcat72: Hmmm, now that I'm looking at this, (and if I understand) I can't use `$genbank_files.fields.path` because I can't filter on the selected items.  But maybe I can use `$genbank_files.fields` to get back the full list inside the filter and then set $ref to `$reflist[0].path`?

hepcat72: Well, I tried your suggestion, but I get:

```
  File "/private/var/folders/8d/z84mms353qb2n0r3n2jhww0m0000gq/T/tmpctof0hum/galaxy-dev/lib/galaxy/jobs/runners/__init__.py", line 224, in prepare_job
    job_wrapper.prepare()
  File "/private/var/folders/8d/z84mms353qb2n0r3n2jhww0m0000gq/T/tmpctof0hum/galaxy-dev/lib/galaxy/jobs/__init__.py", line 871, in prepare
  File "/private/var/folders/8d/z84mms353qb2n0r3n2jhww0m0000gq/T/tmpctof0hum/galaxy-dev/lib/galaxy/tools/evaluation.py", line 451, in build
    raise e
NotFound: cannot find 'genbank_files'
FAIL
...
```

hepcat72: Is there a newer version of:

https://galaxyproject.org/admin/tools/data-tables/

which explains how to do this without relying on Galaxy internals?

I just reviewed the doc again to see if I misinterpreted the suggested way to access the paths of the selected items, but it says to do what I did:

> In the tool XML, there would need to be a parameter such as:
> 
> ```
> <param name="index" type="select">
>     <options from_data_table="bowtie_indexes" />
>   </param>
> ```
> 
> This would give the user a dropdown listing all of the values in the name column supplied in the loc file (in our example, "User-friendly phiX174 Description"). The value stored as the value of the parameter index would be "phiX_ID". The script or binary will need the actual path in the command, and since the ID is the value, some work will need to be done to extract the path. **The following line would do the trick: --ref="${ filter( lambda x: str( x[0] ) == str( $index ), $ __app__.tool_data_tables['bowtie_indexes'].get_fields() )[0][-1] }" In this line, str( x[0] ) refers to the unique ID in column 0 and the -1 in [0][-1] gets the path in the last column.**

so I'm guessing this is outdated.

I know that a single selection select list returns an object on which I can use `.fields.path`, but what is returned when `multiple=true` does not appear to be iterable nor an object on which I can use `.fields.path` and I cannot find how to turn the list of IDs from `str($reference.fixed_genome).split(',')` into their corresponding paths otherwise (unless I did something wrong, which is quite possible).

It seems as though this might be the first tool that takes multiple references as input for a single run and that there either is not code or documentation to support this as a multiple select list?  There is (I am told) a defunct old version of this tool on the test toolshed that used a repeat to supply the multiple references, but I wanted to make the interface nicer and require fewer clicks with a multiple select list.  If a multiple selection select list for data tables is not in fact supported (yet), I could implement the repeat version?

hepcat72: Still though, it would be nice to be able to make the builtin select list a multiple select list...

hepcat72:
Just FYI: I think I accidentally "requested" a review by @bgruening with a stray click the other day. I don't know if that delays things?

bgruening:
Thanks!

PR #2669, opened 2019-11-04
Variant Analyzer

monikaheinzl:
FOR CONTRIBUTOR:


tools/VA/mut2read.py
```diff
@@ -0,0 +1,160 @@
...
+                dcs_median = np.median(np.array(dcs_len))
+                cvrg_dict[chrom_stop_pos] = (count_ref, count_alt, dcs_median)
+
+                print("coverage at pos %s = %s, ref = %s, alt = %s, other bases = %s, N = %s, indel = %s, low quality = %s, median length of DCS = %s\n" %
+                      (pileupcolumn.pos, count_ref + count_alt, count_ref, count_alt, count_other, count_n,
+                       count_indel, count_lowq, dcs_median))
+    bam.close()
+
+    with open(json_file, "wb") as f:
+        json.dump((tag_dict, cvrg_dict), f)
```
wm75: This pattern won't work in Python 3. For compatibility, just open the file in text mode.

wm75: In general, wouldn't it be worth the effort to port the Python code to Python 3 at this occasion?
At first glance it doesn't look like too much work.

monikaheinzl: I will also ping @gpovysil who is the author of the tools. I just opened the pull request for her.

tools/VA/mut2read.py
```diff
@@ -0,0 +1,160 @@
...
+        print("Error: Could not find '{}'".format(file1))
+        exit(0)
+
+    if os.path.isfile(file2) is False:
+        print("Error: Could not find '{}'".format(file2))
+        exit(1)
+
+    if os.path.isfile(file3) is False:
+        print("Error: Could not find '{}'".format(file3))
+        exit(2)
```
wm75:
```
if not os.path.isfile(file1):
     # and so on
```

In addition, I think you should use `sys.exit` instead of the built-in `exit`, and not exit with a status code of `0` (indicating success to the shell) in an error condition.
In fact, I'd rather just do `sys.exit('Your specific message')`, which will result in an exit code of `1` and the message being printed to stderr (no to stdout as with the current `print`).

tools/VA/mut2read.py
gpovysil: We will change the file opening, I missed that when changing over from pickle format.
I'm reluctant to move to Python 3 since these tools will be part of the Du Novo pipeline, which is in Python 2.7.

tools/variant_analyzer/mut2read.py
```diff
@@ -0,0 +1,157 @@
...
+    # read mut file
+    with open(file1, 'r') as mut:
+        mut_array = np.genfromtxt(mut, skip_header=1, delimiter='\t', comments='#', dtype='string')
+
+    # read dcs bam file
+    pysam.index(file2)
+    bam = pysam.AlignmentFile(file2, "rb")
+
+    # get tags
+    mut_tags = []
```
wm75: I may be mistaken, but it seems to me that `mut_tags` is completely dispensable. Looks like all its information is in `tag_dict` anyway, so maintaining it is just a waste of memory and perfomance (because a dict lookup will be much faster than a list one)?

wm75: Maybe just rename `tag_dict` to `mut_tags`?

tools/variant_analyzer/mut2read.xml
```diff
@@ -0,0 +1,66 @@
...
+    </macros>
+	<requirements>
+        <requirement type="package" version="2.7">python</requirement>
+        <requirement type="package" version="1.4.0">matplotlib</requirement>
+        <requirement type="package" version="0.15">pysam</requirement>
+    </requirements>
+    <command>
+python '$__tool_directory__/mut2read.py' 
+--mutFile '$file1'
+--bamFile '$file2'
```
wm75: Here and wherever you are using BAM datasets, you could save substantial runtime by reusing Galaxy's BAM indices.
Just symlink the BAM and its index to the job working directory, then pass the symlink to the corresponding python script and pysam can reuse the index instead of building it itself.
Something like:
```
        ln -s '$file2' bam_input.bam &&
        ln -s '${file2.metadata.bam_index}' bam_input.bam.bai &&
        python ....
```
should work.

tools/variant_analyzer/mut2read.py
monikaheinzl: you are right. we will change it.

bgruening:
ping @wm75

monikaheinzl:
I have added an update for one of the tools. Could somebody review them again?

wm75:
Well, essentially nearly everything I pointed out previously:

- binary file access
- unnecessary pysam indexing
- unconventional exiting strategies

still exist in at least one occasion (check out read2mut as an example, but possibly also other places).
I was only pointing out examples of these before.

gpovysil:
> 
> 
> Well, essentially nearly everything I pointed out previously:
> 
>     * binary file access
> 
>     * unnecessary pysam indexing
> 
>     * unconventional exiting strategies
> 
> 
> still exist in at least one occasion (check out read2mut as an example, but possibly also other places).
> I was only pointing out examples of these before.

I just checked the files again, all these things are fixed in the latest version. e.g. pysam indexing is commented out in all 3 files, sys.exit is used in all 3, ....
Is it possible that you did not get the last commit?

wm75:
Hmm, what I have is:
```
commit 00d2cda9645b447795cb3c69f11bd1a2a86eb21c (HEAD -> variantAnalyzer_iuc, origin/variantAnalyzer_iuc)
Author: Monika Heinzl <k1556322@students.jku.at>
Date:   Tue Nov 19 11:28:43 2019 +0100

    fixed linting3
```

wm75:
Ah, but wait, there are two folders - one called VA, one variant_analyzer - in your fork. I guess I was looking at the first one. You did your changes in the second?
Could you remove the wrong one then?

gpovysil:
Sorry about that. Monika renamed it and didn't realize that it would create a new folder. She will remove the wrong one tomorrow.

wm75:
Looking at all changed files, it seems things in tools/fsd are also part of this PR and probably shouldn't?

gpovysil:
Correct, they should not be part of the pull request, but somehow ended up in it. fsd has already been accepted in a separate pull request.

monikaheinzl:
Sorry for all the confusion. I have figured now out how to remove the folders.

wm75:
Much better now, thanks!

PR #2671, opened 2019-11-06
Add ncbi_acc_download

dfornika:
FOR CONTRIBUTOR:


dfornika:
There are a few issues I'd like to address before this is merged:

1. **How to input the accession number(s):** It currently takes a file containing a list of accessions as input, but should probably also accept a single accession from a text input box
2. **How to download multiple accessions:** I've initially written the wrapper to iterate over the input file of accessions and run `ncbi-acc-download` once on each accession. But `ncbi-acc-download` can accept multiple accessions directly so it probably makes more sense to pass them all at once(?)
3. **Retry on failure**: During testing I've noticed that downloads will occasionally fail, so some files will be missing from the output. I'd like to have some way to retry a failed download, with some reasonable limit on the number of retries.

wm75:
I may have missed the story behind this, but what is your tool offering that you are not supposed to be able to do with [ncbi_eutils_efetch](https://toolshed.g2.bx.psu.edu/repository?repository_id=e4d17fb432d1ef7a)?
That tool, available for example at https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/iuc/ncbi_eutils_efetch/ncbi_eutils_efetch/1.2, seems to have issues, but maybe it could be fixed and enhanced?

dfornika:
That's a good suggestion @wm75. 

To provide a bit of context: I'm working on a workflow for plasmid analysis, and one step of the pipeline involves downloading a reference plasmid to align against based on an NCBI accession. I wasn't sure what the simplest/most reliable method for doing that was, so I did a bit of searching and found the `ncbi-acc-download` tool that seems to do exactly what I want. I developed a preliminary wrapper and plugged it into my workflow and it (mostly) seems to fit my needs.

I'm generally aware of the NCBI eutils but haven't used the galaxy tools. I'll try plugging the `ncbi_eutils_efetch` tool into my workflow and see if that will work.

peterjc:
I see a role for ``ncbi-acc-download`` which uses the NCBI API internally, but adds a bunch of error checking etc on top, and makes this particular (common) task much easier.

dfornika:
I've briefly tried the `ncbi_eutils_efetch` tool in my workflow, and although it seems to provide the functionality that I want, it failed on my test run. I noticed that there is some documentation about adjusting the `job_conf.xml` to accomodate the NCBI API rate limit.

I've added a simple `sleep 5` command and enabled validation on this tool, and on my quick test run it successfully downloaded all 6 accession files that I need (both genbank and fasta).

wm75:
Yeah, was hoping to get you motivated to polish/fix the existing tool instead ;-), but if it seems easier for you to get the dedicated new tool to work that's ok with me, too.

wm75:
So, to discuss your proposal, specifically: does anything else than `fasta` format actually work for proteins? I think  `ncbi-acc-download` ignores the format parameter if the molecule type is not nucleotide.

dfornika:
@wm75 You're right, protein downloads only give fasta files, even if another format is requested. I've added a test (that currently fails) that demonstrates that behavior.

I'll try to modify the wrapper to match the actual behavior of `ncbi-acc-download`, so the user shouldn't be able to select other file formats besides `fasta` if they want to download protein records.

dfornika:
I think I've addressed the issues I described above and also the question raised by @wm75 about available formats for protein sequence downloads.

I haven't implemented any logic to retry failed downloads, but I have enabled `ncbi-acc-download`'s validation features. If multiple accessions are provided, then downloads are performed in series, with a 5-second pause in between. This is done to avoid triggering any rate-limiting done by the NCBI API.

tools/ncbi_acc_download/ncbi_acc_download.xml
```diff
@@ -0,0 +1,154 @@
...
+                <element name="NP_003192">
+                    <assert_contents>
+                        <has_line line=">NP_003192.1 transcription factor A, mitochondrial isoform 1 precursor [Homo sapiens]" />
+                    </assert_contents>
+                </element>
+            </output_collection>
+        </test>
+    </tests>
+    <help><![CDATA[
+Note: only fasta format is supported for protein sequence downloads
```
bgruening: Can we extend the help section a little bit? What does this tool do in general.

dfornika: I've added a basic statement of what the tool does, and also added a warning that records are downloaded sequentially so the tool is probably not appropriate for downloading a large number of records.

bgruening:
@wm75 can you please merge this one if you are happy.

wm75:
There are still some issues with this:
* contrary to what it claims, the wrapper doesn't accept newlines as record separators (Galaxy sanitizes `\n` to `__cn__` by default and the wrapper is not handling this
* the resulting collection elements don't have their proper datatypes assigned
  The file format of generated collection elements should also be verified by the wrapper tests
* *Output* isn't a desirable name for the generated collection

In addition, but that's a general problem with tools that only produce a collection as their output:
If anything goes wrong, the user does not have access to error messages, tool output (there is no bug icon. etc. attached to the collection itself).
@bgruening Do you know if there is ongoing work addressing this problem?

@dfornika Since this is particularly relevant for download tools like yours (where the user would really like to know whether, e.g., the download failed because of a wrong accession number), I would consider producing a simple log file as an additional dataset *outside* the collection - giving the user an entry point for debugging.

bgruening:
> In addition, but that's a general problem with tools that only produce a collection as their output:
If anything goes wrong, the user does not have access to error messages, tool output (there is no bug icon. etc. attached to the collection itself).
@bgruening Do you know if there is ongoing work addressing this problem?

Kindof ... the new history from @nerdinacan will pave the way to make this happen. But this should not stop this PR imho.

dfornika:
@wm75 Thanks for these suggestions. I'll address as many as possible as soon as I can.

dfornika:
@wm75 I think I've addressed your suggestions. Can this be merged?

wm75:
@dfornika great improvements!
One remaining issue is that the tool doesn't fail on failed downloads unless the last accession number causes the failure. Seems the for loop just continues on failures and it's only the last loops exit code that is seen by Galaxy.

wm75:
@dfornika Sorry, was in a hurry earlier today and missed a small mistake in my patch. All that's really necessary to make it work is to add a terminal `$` to the first grep pattern.
The main purpose off the patch was to catch empty lines so your regex adjustment is not really correct I think.

dfornika:
@wm75 I've had some difficulty causing the job to fail if one or more of the downloads fails (except the last one as you described).

But could I suggest an alternative behaviour? We could write the accessions of any failed downloads to an output file. Then the user or pipeline designer could use that info to decide how to proceed (they might even pass it to another copy of this tool to retry the download).

wm75:
> But could I suggest an alternative behaviour? We could write the accessions of any failed downloads to an output file. Then the user or pipeline designer could use that info to decide how to proceed (they might even pass it to another copy of this tool to retry the download).

@dfornika That is a very good idea. I enhanced it a bit with a new parameter that gives the user the choice between just logging failures and failing on the first failed download. The latter may be useful for workflows where it does not make sense to continue with partial data.
Give it a try, then if you don't find any problems, I don't think I have anything more to complain about :smile: and we could merge this.

dfornika:
That's looking really nice. Thanks for your help!

dfornika:
@wm75 can we merge? Thanks again.

bgruening:
@dfornika thanks a lot for all your contribution!

wm75:
@dfornika A big thank you also from me for bearing with me and all my requests!

dfornika:
@wm75 Your suggestions made the tool much better. I know it's important for iuc tools to be high-quality.

PR #2676, opened 2019-11-14
Simplify implicit snippy dependencies.

jmchilton:
It's biocontainer doesn't work for this tool because tar doesn't support ``-z`` in the container.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/snippy/snippy.xml
+++ b/tools/snippy/snippy.xml
@@ -54,7 +54,7 @@
         #end if
 
         mkdir -p ${dir_name} && cp -r out/reference out/snps.tab out/snps.aligned.fa out/snps.vcf ${dir_name}/ &&
-        tar -czf out.tgz ${dir_name}
+        tar -cf - ${dir_name} | gzip > out.tgz
         #if "outcon" in str($outputs) and $adv.rename_cons
           && sed -i 's/>.*/>${dir_name}/' out/snps.consensus.fa
         #end if


tools/snippy/snippy.xml
```diff
@@ -54,7 +54,7 @@
         #end if
 
         mkdir -p ${dir_name} && cp -r out/reference out/snps.tab out/snps.aligned.fa out/snps.vcf ${dir_name}/ &&
-        tar -czf out.tgz ${dir_name}
+        gzip -dc out.tgz | tar xf - ${dir_name}
```
nsoranzo: That's replacing a compress action with a decompress, I think.

jmchilton: Hopefully the tests would have caught that, let me give this another shot.

nsoranzo: Yes, tests failed: https://travis-ci.org/galaxyproject/tools-iuc/jobs/611915772 , but I'm faster :laughing:

nsoranzo:
@jmchilton Thanks! Before we merge this, can you explain why the `-z` option of tar does not work inside the container? Is it a non-GNU tar or is gzip missing or something else??
Since there are other IUC tools using `tar`, it may be worth fixing them all e.g. by adding a dependency on the [GNU tar package](https://anaconda.org/conda-forge/tar/), if that is an alternative solution.

jmchilton:
```
docker run -it --rm 'quay.io/biocontainers/snippy:4.4.5--0' bash
bash-4.2# tar --version
tar: unrecognized option '--version'
BusyBox v1.22.1 (2014-05-23 01:24:27 UTC) multi-call binary.

Usage: tar -[cxthvO] [-X FILE] [-T FILE] [-f TARFILE] [-C DIR] [FILE]...

Create, extract, or list files from a tar file

Operation:
	c	Create
	x	Extract
	t	List
	f	Name of TARFILE ('-' for stdin/out)
	C	Change to DIR before operation
	v	Verbose
	O	Extract to stdout
	h	Follow symlinks
	exclude	File to exclude
	X	File with names to exclude
	T	File with names to include
```

xref sort of... https://github.com/BioContainers/mulled/issues/148

I'm fine with annotating the tar dependency and bumping the version, this fix would fix the issue without requiring a new tool version or image build so it was my first attempt but whatever the IUC thinks is best is fine with me.

mvdbeek:
I think either is fine, adding another dependency is not a big deal either. I think the important things is that we switch to using mulled_build testing, which would have caught this issue before merging.

bgruening:
In the sake to get this in fast, let's merge it, but I think we should fix the annotation in the future, probably when we have the new testing in place. We also can/need to migrate to CCI or GitHub pipeline during this move.

nsoranzo:
Thanks @jmchilton , I have tried to track down this issue and it seems that Biocontainers are built using galaxy-lib, which uses `bgruening/busybox-bash:0.1` as base image: https://github.com/galaxyproject/galaxy-lib/blob/master/galaxy/tools/deps/mulled/invfile.lua#L63

In turn, [bgruening/busybox-bash:0.1](https://github.com/bgruening/docker-busybox-bash/releases/tag/0.1) is based on top of `progrium/busybox:latest` , which is where the issue is also displayed.

The new `bgruening/busybox-bash:0.2` is instead based on top of `alpine:3.9` , where gzip support in tar works fine.

Ping @bgruening

nsoranzo:
And another instance of this issue is https://github.com/bioconda/bioconda-recipes/issues/12243

jmchilton:
The Travis CI has ``sudo: required`` enabled - I'd think that is the hardest part to getting mulled testing working. I'll open a PR and see what happens?

mvdbeek:
💯

PR #2677, opened 2019-11-15
[WIP] Dockerized Testing.

jmchilton:
FOR CONTRIBUTOR:

The goal of this PR is to enable testing inside of containers to catch more tool problems in container environments as quickly as possible.

Added a test I expect to fail to debug this actually working.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@
 sudo: required
 language: python
 cache: pip
+services:
+  - docker
 
 python: 2.7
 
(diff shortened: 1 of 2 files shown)


jmchilton:
Shockingly easy! It even somehow found another issue with another tool.

nsoranzo:
> Shockingly easy! It even somehow found another issue with another tool.

1) For the `snippy` tool, the Travis logs for the snippy tool test contains this line:

```
/tmp/tmpvlqBmk/job_working_directory/000/17/tool_script.sh: line 9: /tmp/tmpvlqBmk/job_working_directory/000/17/COMMAND_VERSION: Read-only file system
```

which is caused by `<version_command><![CDATA[snippy --version]]></version_command>` being translated by Galaxy into `snippy --version > /tmp/tmpvlqBmk/job_working_directory/000/17/COMMAND_VERSION 2>&1;` in the `tool_script.sh` : this looks like a (non-fatal) Galaxy bug.

2) For the `snippy_core` tool, the error is:

```
tar: unrecognized option '--strip-components=1'
```

which is due to this part of the `<command/>`:

```
tar -xf '$indir' -C '$sample_name' --strip-components=1
```

I think this can only be fixed by adding a requirement for GNU tar.

jmchilton:
The version problem should be fixed with something like: https://github.com/galaxyproject/galaxy/pull/8997. I'm not sure we should fix it in 19.09.

nsoranzo:
We are doing this for a while.

PR #2684, opened 2019-11-18
Fix data params format options

wm75:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/spades/macros.xml
+++ b/tools/spades/macros.xml
@@ -1,6 +1,4 @@
 <macros>
     <token name="@TOOL_VERSION@">3.12.0</token>
-    <token name="@INTYPES@">
-        fasta,fasta.gz,fastq,fastq.gz,fastqsanger,fastqsanger.gz
-    </token>
+    <token name="@INTYPES@">fasta,fasta.gz,fastq,fastq.gz,fastqsanger,fastqsanger.gz</token>
 </macros>
--- a/tools/spades/macros.xml
+++ b/tools/spades/macros.xml
@@ -1,4 +1,4 @@
 <macros>
     <token name="@TOOL_VERSION@">3.12.0</token>
-    <token name="@INTYPES@">fasta,fasta.gz,fastq,fastq.gz,fastqsanger,fastqsanger.gz</token>
+    <token name="@INTYPES@">fasta,fasta.gz,fastq,fastq.gz</token>
 </macros>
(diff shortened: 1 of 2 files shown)


wm75:
The line break and the indentation in the token got expanded into the params format identifier string.
While Galaxy seems to cope with that when building the tool interface, it spoils the json you get under
/api/tools/spades?io_details=True and the association with an edam format identifier.

tools/spades/macros.xml
```diff
@@ -1,6 +1,4 @@
 <macros>
     <token name="@TOOL_VERSION@">3.12.0</token>
-    <token name="@INTYPES@">
-        fasta,fasta.gz,fastq,fastq.gz,fastqsanger,fastqsanger.gz
-    </token>
+    <token name="@INTYPES@">fasta,fasta.gz,fastq,fastq.gz,fastqsanger,fastqsanger.gz</token>
```
mvdbeek: while we're at it let's also drop the redundant formats:
```suggestion
    <token name="@INTYPES@">fasta,fasta.gz,fastq,fastq.gz</token>
```

bgruening:
@wm75 @mvdbeek shouldn't we fix this in Galaxy and strip the string? I have seen also `xxx, yyy` not working. Maybe this is related.

mvdbeek:
Sure, but there's no harm merging this IMO.

wm75:
> I have seen also `xxx, yyy` not working. Maybe this is related.
@bgruening not working in the sense of preventing edam format association?

wm75:
@jmchilton This is the whitespace inside a parameter attribute issue I was mentioning briefly to you.
In your opinion, should planemo flag this as an error, or should Galaxy just sanitize the xml silently everywhere?

bernt-matthias:
Are the formats really redundant?

wm75:
Why not? Isn't fastq including fastqsanger? Then there is no need to accept fastqsanger explicitly if fastq is accepted.

bernt-matthias:
Not sure. I always forget if this works. Can you try? Maybe with a test or try manually with `planemo server`.

wm75:
spades actually already has tests for all the different input formats, e.g., https://github.com/galaxyproject/tools-iuc/blob/137c64bd71994d77c8b90c5d6504111e037d94b4/tools/spades/spades.xml#L300
:smile:

mvdbeek:
I don't think we need to bump as it doesn't change the output and is more of a cosmetic fix (except when you want to parse formats from the /api/tools/<tools_id>?io_details=true)

wm75:
Yes, that was my thought, too.

bernt-matthias:
I guess we can merge then. Thanks a lot.

wm75:
underlying issue is addressed in https://github.com/galaxyproject/galaxy/pull/9022

PR #2691, opened 2019-11-25
samtools view fixes and enhancements

wm75:
FOR CONTRIBUTOR:

Upon a closer look at the samtools_view wrapper I discovered a few bugs and opportunities to enhance it, which I'd like to address in this PR.


wm75:
@bernt-matthias @bwlang while the above changeset fixes some special cases of the "subsampling to target number of reads" code, there is another issue with it that is much harder to fix:
The -s option of samtools view can be combined with all filters that the tool has to offer. The way the fractional selection works it will always return the specified fraction of the reads passing the filters.

In contrast, with the target number approach in its current form the user, effectively, **specifies the number of reads that should go into filtering**, and the final number will then be reduced further.

(Quick example: assume 10,000 fw and 10,000 rv reads from PE data and a filter for the *read is first in a pair* flag. With `-s 0.5`, you will keep around 5000 fw reads as expected, but with a target # of 5,000 reads you will not get those 5000 fw reads, but only ~ 2500 of them.)


Fixing this is rather complicated and expensive, too. Basically, one would have to turn all tool wrapper filter options into their command line counterparts, then use these in a `samtools view -c` subcommand (similar to what's done now for SAM format input).
Is this worth it? Or should I just document the confusing status quo in the tool help?

tool_collections/samtools/samtools_view/samtools_view.xml
```diff
@@ -13,17 +13,20 @@
...
         #if $cond_subsample.select_subsample == 'target':
             ##this must be done before the main command because we don't know the total # reads until execution time
-            #if str($cond_subsample.seed) == '':
-                #set seed_expr="int(rand() * 32768)"
-            #else
-                #set seed_expr=$cond_subsample.seed
-            #end if
             #if $input.is_of_type('sam'):
-                sample_fragment=`samtools view -c infile | awk '{s=\$1} END {frac=$cond_subsample.target/s; print(frac < 1 ? "-s "  $seed_expr+frac : "")}'` &&
+                sample_fragment=`samtools view -c infile | awk '{s=\$1} END {frac=$cond_subsample.target/s; print(frac < 1 ? $seed+frac : ".0")}'` &&
```
bernt-matthias: I don't understand why seed and frac are added?

wm75: I did not really change this part, but I think they are added inside awk to produce the correct printed form. `$seed` is an integer at that point, `frac` a float calculated by awk. Adding them gives the `seed.frac` format expected by samtools view.

bernt-matthias: Ah - my bad :) 

What happens in the case when frac >= 1? then no reads will be in the output?

wm75: No, the .0 frac means *all* reads (of course, the seed becomes irrelevant then).

wm75: The old way of trying to just pass an empty string instead of the whole -s option did not work because samtools then takes "" as the file name.

tool_collections/samtools/samtools_view/samtools_view.xml
```diff
@@ -13,17 +13,20 @@
         ## prepare reference data
         @PREPARE_FASTA_IDX@
         @PREPARE_IDX@
+        #if $cond_subsample.select_subsample != 'no':
+            #import random
+            #if str($cond_subsample.seed):
+                #set $seed = $cond_subsample.seed
+            #else:
+                #set $seed = random.randrange(32768)
```
bernt-matthias: I guess that we should call random.seed() before that.

wm75: It's one of the advantages of using Python here that it takes care of these mundane things for you :smile:
The random module seeds its Random instance at import time:
https://github.com/python/cpython/blob/f4a21d3b239bf4f4e4e2a8a5936b9b040645b246/Lib/random.py#L100

bernt-matthias: nice :)

wm75:
@bernt-mathias I would still like to address the target-based downsampling issue above and introduce one or two cosmetic changes, but if you want to check out this current version, I'd be happy to get feedback on it.
It was actually really hard to improve the user interface over your initial version, but I think I finally enhanced it a little.
Lets see what you and others are thinking ...

tool_collections/samtools/samtools_view/samtools_view.xml
```diff
@@ -459,29 +745,50 @@
...
-But it is also possible to use CRAM without a reference with the disadvantage that the reference is stored explicitely (as in SAM and BAM). 
+The tool allows you to specify a reference sequence. This is required for SAM input with missing @SQ headers (which include sequence names, length, md5, etc) and useful (and sometimes necessary) for CRAM input and output. In the following the use of reference sequence in the CRAM format is detailed. 
+CRAM is (intended as a primarily) a reference-based compressed format, i.e. only differences between the stored sequences and the reference are stored. As a consequence, the reference that was used to generate the alignemnts is always needed in order to interpret the alignments (a checksum stored in the CRAM file is used to verify that the only the correct sequence can be used). This allows for more space efficient storage than with BAM format, but a CRAM file like this is not useful without its reference.
```
nsoranzo:
```suggestion
CRAM is (primarily) a reference-based compressed format, i.e. only differences between the stored sequences and the reference are stored. As a consequence, the reference that was used to align the sequences is needed in order to interpret the alignments (a checksum stored in the CRAM file is used to verify that only the correct sequence can be used). This allows for more space-efficient storage than with BAM format, but such a CRAM file is not usable without its reference.
```

tool_collections/samtools/samtools_view/samtools_view.xml
```diff
@@ -459,29 +745,50 @@
...
+If you ask for *A filtered/subsampled selection of reads*, the tool will allow
+you to specify filter conditions and/or to choose a subsampling strategy, and
+the output will contain one of the following depending on your choice under
+*What would you like to have reported?*:
 
+- All reads retained after filtering and subsampling
+- Reads dropped during filtering and subsampling
+
+If instead you want to *split* the input reads based on your criteria and
+obtain *two* datasets, one with the retained, one with the dropped reads, check
```
nsoranzo:
```suggestion
obtain *two* datasets, one with the retained and one with the dropped reads, check
```

tool_collections/samtools/samtools_view/samtools_view.xml
```diff
@@ -490,7 +797,9 @@ You may specify one or more space-separated region specifications after the inpu
 
 Regions can be specified as: RNAME[:STARTPOS[-ENDPOS]] and all position coordinates are 1-based.
 
-Important note: when multiple regions are given, some alignments may be output multiple times if they overlap more than one of the specified regions.
+.. class:: Warning mark
+
+   when multiple regions are given, some alignments may be output multiple times if they overlap more than one of the specified regions.
```
nsoranzo:
```suggestion
When multiple regions are given, some alignments may be output multiple times if they overlap more than one of the specified regions.
```

wm75:
@nsoranzo thanks for the review. Any idea about the failing tests?
My guess is that the culprit is https://github.com/galaxyproject/tools-iuc/blob/122e574d58f15b851afb550011719cef85c56e68/tool_collections/samtools/samtools_view/samtools_view.xml#L310
which specifies the empty string as a default for an integer param, but I don't understand how https://github.com/galaxyproject/tools-iuc/pull/2691/commits/573d21c2abb210151d3ee8e6e9a64840d386e123 triggered it. Is this something that has changed recently in Galaxy? Locally (on 19.05) all tests are passing.

bgruening:
```
2019-12-18 09:55:44,625 ERROR [galaxy.jobs.runners] (248) Failure preparing job
  File "/tmp/tmpbKbTPj/galaxy-dev/lib/galaxy/jobs/runners/__init__.py", line 230, in prepare_job
    job_wrapper.prepare()
  File "/tmp/tmpbKbTPj/galaxy-dev/lib/galaxy/jobs/__init__.py", line 1066, in prepare
  File "/tmp/tmpbKbTPj/galaxy-dev/lib/galaxy/tools/evaluation.py", line 483, in build
    raise e
ValueError: invalid literal for int() with base 10: ''
FAIL
```

@gmauro and me are seeing that for floats as well. If I remember correctly @bernt-matthias and @mvdbeek worked on some parameter related things. Can you confirm Marius and Matthias?

bgruening:
This is what we are seeing every now and then on EU:

```
   File "/opt/galaxy/server/lib/galaxy/jobs/runners/__init__.py", line 230, in prepare_job
     job_wrapper.prepare()
   File "/opt/galaxy/server/lib/galaxy/jobs/__init__.py", line 1066, in prepare
   File "/opt/galaxy/server/lib/galaxy/tools/evaluation.py", line 483, in build
     raise e
ValueError: could not convert string to float: 
```

bernt-matthias:
Ha. On dev I get 

```
2019-12-18 12:53:07,639 ERROR [galaxy.jobs.runners] (2) Failure preparing job
  File "/home/berntm/projects/galaxy/lib/galaxy/util/__init__.py", line 1039, in unicodify
    value = text_type(value)
  File "/home/berntm/projects/galaxy/.venv/lib/python3.6/site-packages/Cheetah/Template.py", line 1053, in __unicode__
  File "cheetah_DynamicallyCompiledCheetahTemplate_1576669987_6300395_22088.py", line 166, in respond
ValueError: invalid literal for int() with base 10: ''
```

What a nice collection of tracebacks

bernt-matthias:
Change 

```
#if str($inclusive_filter) != 'None'
```

to 

```
#if str($inclusive_filter) != ''
```

same for the other filters. Maybe check the other tools for comparisons with 'None'

Apparently I developed this during the time when the comparison did not work.

The problem was that `<token name="@FLAGS@">#set $flags = sum(map(int, str($filter).split(',')))</token>` failed because `''.split(',')` is `['']` and `int('')` raises the exception that we saw (which does not cover the exception of @bgruening )

wm75:
@bernt-matthias Now I'm confused. Are you saying that an optional multiselect param is no longer `None`, but `''` now when the user does not select an option?

bernt-matthias:
Seems to be the case at the moment. I guess this is unwanted -- aka a bug...?

wm75:
Hmm, I know the [docs](https://docs.galaxyproject.org/en/latest/dev/schema.html#select) always said it would be `''`, when in fact it was `None`. Wonder whether the behavior was changed to match the docs? But that will break quite some tools then (I think I relied on `None` in several other tools)

bernt-matthias:
A quick check of the tools in this repo verifies this.

bernt-matthias:
Might one of these be the cause?

https://github.com/galaxyproject/galaxy/pull/9059

https://github.com/galaxyproject/galaxy/pull/9054

bgruening:
@bernt-matthias could you maybe add a test tool to Galaxy exposing those bugs?

wm75:
> Might one of these be the cause?
> 
> [galaxyproject/galaxy#9059](https://github.com/galaxyproject/galaxy/pull/9059)
> 
> [galaxyproject/galaxy#9054](https://github.com/galaxyproject/galaxy/pull/9054)

Yeah, that's the one I was just looking at.

bernt-matthias:
https://github.com/galaxyproject/galaxy/pull/9142 lets see .. not sure if I got everything right :)

nsoranzo:
> > Might one of these be the cause?
> > [galaxyproject/galaxy#9059](https://github.com/galaxyproject/galaxy/pull/9059)
> > [galaxyproject/galaxy#9054](https://github.com/galaxyproject/galaxy/pull/9054)
> 
> Yeah, that's the one I was just looking at.

I think these are unrelated, they deal with `InputValueWrapper`, not `SelectToolParameterWrapper`. But https://github.com/galaxyproject/galaxy/pull/9067 could be the one though.

nsoranzo:
> I think these are unrelated, they deal with `InputValueWrapper`, not `SelectToolParameterWrapper`. But [galaxyproject/galaxy#9067](https://github.com/galaxyproject/galaxy/pull/9067) could be the one though.

Confirmed with `git bisect`.

nsoranzo:
Should be fixed by https://github.com/galaxyproject/galaxy/pull/9144 .

wm75:
There is nothing more I'd like to change here - so do you think this is good enough for merging?

bernt-matthias:
Just have seem that the samtools suite in the toolshed is not uptodate. Can we fix this?

wm75:
What exactly do you mean by not up to date?
At least for samtools_view it seems to depend on the latest toolshed version, but I haven't checked the other tools.

tool_collections/samtools/samtools_view/samtools_view.xml
```diff
@@ -1,7 +1,42 @@
-<tool id="samtools_view" name="Samtools view" version="@TOOL_VERSION@">
-    <description>reformat, filter, or subsample</description>
+<tool id="samtools_view" name="Samtools view" version="@TOOL_VERSION@+galaxy1">
+    <description>- reformat, filter, or subsample SAM, BAM or CRAM</description>
```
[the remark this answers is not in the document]
wm75: Hmm. I kind of like it because with *view* in the name the *reformat* reads strange when it's directly adjacent. Willing to change this though if you have a better suggestion.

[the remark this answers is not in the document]
bernt-matthias: OK. You think that users might read it as list. But it will be shown already as 

**Samtools view** reformat, filter, or subsample SAM, BAM or CRAM

i.e. Samtools view in bold letters. Sufficient in my opinion. 

Furthermore, I like consistency .. hence we would need to change it in all tools.

bernt-matthias:
idxstats is an example

The suite lists 3:811fc2d75ad4, but the latest version is 4:7a6034296ae9

bgruening:
@bernt-matthias the suites are strange. I wouldn't use them, we should drop them from the TS or hide them.

Can this PR go in?

bernt-matthias:
Jep. I can live with the dash.

bernt-matthias:
Still don't see an alternative to suites in particular for large tool collections like openms or mothur...

bgruening:
> Still don't see an alternative to suites in particular for large tool collections like openms or mothur...

External yaml files, we have those if you are interested.

PR #2694, opened 2019-11-27
Add explicit dependency on python for data managers

nuwang:
Most of the data managers did not have a python version defined, and as a result, the containers built for them would fail with "missing python". This PR adds an explicit dependency on python 3.7. `data_manager_fetch_refseq` had a python dependency defined but not an explicit version, so I specified 3.7.
`data_manager_fetch_mothur_reference_data` has an explicit dependency on 2.7 since the code still uses old style print statements. This will need to be updated in a separate PR.

FOR CONTRIBUTOR:

### DIFF
--- a/data_managers/data_manager_bowtie2_index_builder/data_manager/bowtie2_index_builder.xml
+++ b/data_managers/data_manager_bowtie2_index_builder/data_manager/bowtie2_index_builder.xml
@@ -1,6 +1,7 @@
 <tool id="bowtie2_index_builder_data_manager" name="Bowtie2 index" tool_type="manage_data" version="@WRAPPER_VERSION@" profile="18.09">
     <description>builder</description>
     <requirements>
+        <requirement type="package" version="3.7">python</requirement>
         <requirement type="package" version="@WRAPPER_VERSION@">bowtie2</requirement>
     </requirements>
     <macros>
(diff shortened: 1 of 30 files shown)


nsoranzo:
Can you bump the version for all the tools, please?

mvdbeek:
I have a branch that needs some history re-writing that will add dependencies and tests for all data managers (actually everything in tools-iuc) and that runs tests against biocontainers at https://github.com/mvdbeek/tools-iuc/tree/gh_actions. I'll pick out the data manager changes, cause I don't think we should dependencies without tests.

nuwang:
@mvdbeek That's great! If the fixes are already in, I'll close this PR. Let me know if you'd like me to reissue the PR against a different branch or something.

mvdbeek:
I'll cherry-pick and resolve, thanks @nuwang

wm75:
At least the gemini DM is still not Python 3 compatible. The single responsible line for the incompatibility is this one:
https://github.com/galaxyproject/tools-iuc/blob/39695220be0005518c7f600e382b4918c71632ff/data_managers/data_manager_gemini_database_downloader/data_manager/data_manager_gemini_download.py#L86

where the binary mode 'wb' would need to be changed to 'w'. @nuwang It would be great if you could add this fix to the PR or the DM will be left in a broken state.

@mvdbeek are all other DMs here Python3 compatible by now?

mvdbeek:
I'm working on it, I'll not have the time to fix everything but I'm setting up the environment so that we can work on this, give me a day or so and I'll have a list of what needs to be done.

PR #2703, opened 2019-11-29
Fix qiime reading/writing from $HOME

mvdbeek:
by adding profile="18.01" to qiime tools, which gives tools an isolated
home directory in the job working directory.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/qiime/qiime_core/align_seqs.xml
+++ b/tools/qiime/qiime_core/align_seqs.xml
@@ -1,4 +1,4 @@
-<tool id="qiime_align_seqs" name="Align sequences" version="@WRAPPER_VERSION@.0">
+<tool id="qiime_align_seqs" name="Align sequences" version="@WRAPPER_VERSION@.0" profile="@PROFILE@">
     <description> using a variety of alignment methods (align_seqs)</description>
     <macros>
         <import>macros.xml</import>
--- a/tools/qiime/qiime_core/alpha_diversity.xml
+++ b/tools/qiime/qiime_core/alpha_diversity.xml
@@ -1,4 +1,4 @@
-<tool id="qiime_alpha_diversity" name="Calculate alpha diversity" version="@WRAPPER_VERSION@.0">
+<tool id="qiime_alpha_diversity" name="Calculate alpha diversity" version="@WRAPPER_VERSION@.0" profile="@PROFILE@">
     <description> on each sample in an otu table, using a variety of alpha diversity metrics (alpha_diversity)</description>
     <macros>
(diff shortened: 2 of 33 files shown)


bgruening:
How do we proceed in general here? Do we merge those PRs but cancel the TS upload and after all the testing and other modifications are done, upload everyone at once? I would like to avoid as many unneeded TS updates as possible. Or in other words, I would like to pull as many updates together as possible before we hit the TS.

Thanks @mvdbeek!

mvdbeek:
> I would like to avoid as many unneeded TS updates as possible. Or in other words, I would like to pull as many updates together as possible before we hit the TS.

Is that because of the automatic deployment on usegalaxy.eu ? These should be very cheap updates as there is no version bump.

bgruening:
I could deactivate those updates if needed - but yes, also this. Hitting the TS less often is in general probably good and pulling changes together if possible. Just consider how big an entire checkout of qiime + test-data is on disc - assuming the updates do work now automatically.

But I'm also talking in general, we will touch many tools probably multiple times in the next week.
Maybe it's wise if we do not upload those PRs. The only downside I can see is that we diverge TS with github/master for a few weeks.

mvdbeek:
I don't think this would be worth it. There will be no additional space taken on the TS hg repo (except for a couple of bytes for the changed code), and if there is no version bump we update the repo in place on the Galaxy side, which doesn't lead to big changes either. But I'm also not strongly opposed to cancelling the upload.

nsoranzo:
@bgruening How to keep track of all these PRs we need to push manually? Wouldn't it be simpler to make sure an entire suite of tools is fixed in a single PR instead?

mvdbeek:
Also the cloud team is testing things atm in k8s, so every update would be good to roll out timely.

mvdbeek:
@nsoranzo we could have a dev-branch ?

nsoranzo:
A dev branch could work to stage all the changes without uploading and for tracking, but when merging to master tests will very probably fail and we'll have to push manually to the TSs.

mvdbeek:
Yeah, it's not that I would be favorable towards this, I think our infrastructure simply must handle frequent updates. If it doesn't we need to work on it.

bgruening:
I like the idea of a dev branch, just this time for the huge migration.

mvdbeek:
I've backout out the changes to `qiime_add_on`, otherwise the suite never finishes tests.

mvdbeek:
This is ready for review ... it doesn't pass all the container tests, for that we need to update the mulled-build tool to respect the extended_base_image

Issue #2701, opened 2019-11-28
List of tools that fail containerized tool tests


mvdbeek:
Hello IUC,

**Tl;DR:**

Help us fix tool tests!
---------------------

we would like to make sure that our tools are well tested and working in the widest range of deployment scenarios possible. And as you may have noticed, tool dependency installation from Conda may break (or be fixed!) as time passes due to dependency-of-a-dependency upgrades, newer/older Conda versions etc.

To make sure that our tool wrappers will work on a wide range of deployments it is important that we include **all** dependencies that a tool has (i.e include coreutils if a wrapper uses `paste`, include sed if a wrapper uses `sed`, include python if a wrapper calls a python script). One way to make sure this is the case is to run tool tests in a minimal container that does not contain any software except the one listed in our tool wrappers `<requirements/>` tag set.

We also currently build containers for all IUC tools as well as other tool repositories listed in https://github.com/galaxyproject/planemo-monitor, but so far we are not testing that tool runs actually succeed in these containers. For dependencies (and dependency combinations) that are not available as a downloadable container Galaxy can build containers when needed. This functionality is also available via `planemo test --biocontainers`.

We have set up a github workflow that tests all IUC tools using `planemo test` and the `--biocontainers` flag. The test run can be seen here: https://github.com/mvdbeek/tools-iuc/commit/529a26943046ca80b18bccae31021db5c317eb7e/checks?check_suite_id=331921644
And a html report of only the failing tests is here: https://mvdbeek.github.io/iuc-tool-test-results/
Note also that the tool tests run against python 3.7, so this could be another class of errors we may need to fix (should concern only the cheetah templating).

The list of tools with failing tests is the following:

- [x] multigps @gregvonkuster - https://github.com/galaxyproject/tools-iuc/pull/2754
- [x] ucsc-twobittofa @mvdbeek https://github.com/galaxyproject/tools-iuc/pull/2731
- [x] umi_tools (pysam ImportError: libcrypto.so.1.0.0: cannot open shared object file: No such file or directory) @mvdbeek https://github.com/galaxyproject/tools-iuc/pull/2729 and container rebuild
  - [x] umi_tools_group
  - [x] umi_tools_dedup
- [x] scanpy: @mvdbeek https://github.com/galaxyproject/tools-iuc/pull/2704
  - [x] scanpy_filter
  - [x] scanpy_normalize
  - [x] scanpy_cluster_reduce_dimension
  - [x] scanpy_plot
[...]

PR #2705, opened 2019-11-30
dada2 fixes

bernt-matthias:
detected here https://github.com/galaxyproject/tools-iuc/issues/2701

- data manager: real bug fix
- bug fix in makeSequencTable: cheetah for testing `plot` variable was wrong
- allow larger delta for tests on Rdata and pdf (mostly 1/2 file size, but anyway less then the file size .. essentially now a test for file presence)

FOR CONTRIBUTOR:


data_managers/data_manager_dada2/data_manager/data_manager.py
```diff
@@ -120,7 +120,7 @@ def remote_dataset(dataset, outjson):
         data_manager_entry['path'] = dataset + ".species"
         data_manager_json["data_tables"]["dada2_species"] = data_manager_entry
 
-    with file(outjson, 'w') as jf:
+    with open(outjson, 'w') as jf:
```
mvdbeek: To make this more consistent for python versions that don't guarantee dictionary sort order (python < 3.6) I'd change the line below to `jf.write(json.dumps(data_manager_json, sort_keys=True))`

data_managers/data_manager_dada2/data_manager/dada2_fetcher.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<tool id="dada2_fetcher" name="dada2 data manager" tool_type="manage_data" version="0.0.7">
+<tool id="dada2_fetcher" name="dada2 data manager" tool_type="manage_data" version="0.0.8">
```
mvdbeek: Can you add a python requirement (I add python 3.7 for the data managers I already worked on) here and bump the `profile` to 18.09, which should be the first one that will use containers for data managers ?

tools/dada2/dada2_makeSequenceTable.xml
```diff
@@ -15,7 +15,7 @@
 @WRITE_FOO@
 
 library(dada2, quietly=T)
-#if $plot == "yes"
+#if $plot
```
mvdbeek: That should have been fixed by https://github.com/galaxyproject/galaxy/pull/9052/ ... can you check if that is still necessary ? It's fine the way you're doing it here but better to be aware if we're breaking Galaxy tools in unforeseen ways.

data_managers/data_manager_dada2/data_manager/data_manager.py
mvdbeek: (that may also require updating any test output data if the order changes)

data_managers/data_manager_dada2/data_manager/dada2_fetcher.xml
bernt-matthias: done

bgruening:
@bernt-matthias can you maybe use https://github.com/galaxyproject/galaxy/pull/8852 ... and just check for size. This way we also remove the Rdata from the repo.

bernt-matthias:
Not sure if file size is the same on all systems since the files can be compressed with different methods or even uncompressed. 

We could fix to uncompressed. Which would be faster, but require a bit more space. My guess would be that the file size might be the same on all systems.

data_managers/data_manager_dada2/data_manager/data_manager.py
bernt-matthias: consider it done

tools/dada2/dada2_makeSequenceTable.xml
bernt-matthias: seems to work with `planemo t --galaxy_branch release_19.09`, but not with `planemo t`. I guess the commits are not yet in master?

Anyway, I will restore it.

bgruening:
You have a delta in there as well: https://github.com/galaxyproject/galaxy/pull/8852/files#diff-4191170733b24a9e99ba70b64c8b8787R30

mvdbeek:
Oh yeah, it's not in master yet, I'll do that. There's also https://github.com/galaxyproject/galaxy/pull/9054 to look out for.

mvdbeek:
master is updated now

bernt-matthias:
@bgruening sounds good. just was wondering if --update_test_data updates the numbers ..? Probably not... :)

bernt-matthias:
Seems that also https://github.com/galaxyproject/galaxy/pull/8852 has not found its way to master and 19.09 which would be necessary.

bernt-matthias:
Anyway, I just updated the test and do not see the advantage here: We can't remove the test outputs, since subsequent steps of the dada2 pipeline use them as input. 

My last commit could be easily undone if you agree.

bgruening:
Imho it makes a lot of sense for PDF ... that tools read Rdata is not good, security wise. But I guess this is unrelaed to this PR.

bernt-matthias:
> that tools read Rdata is not good, security wise. But I guess this is unrelaed to this PR.

still did not get this point. as far as I see it nothing is executed on import.

bernt-matthias:
@mvdbeek to make this work we need https://github.com/galaxyproject/galaxy/pull/8852 in master and 19.09 .. and I think also the xsd of planemo needs an update

mvdbeek:
It isn't a bug, I don't think we can do a backport. But it might not be required if we start using using the proper galaxy packages instead of galaxy-lib in planemo. @nsoranzo was working on this.
Can we just work with the delta for now ?

bernt-matthias:
yep. will reset to https://github.com/galaxyproject/tools-iuc/pull/2705/commits/770ea3840e37f02963dd3ca480846e6445f5bfc6

PR #2710, opened 2019-11-30
newick display: add imagemagick dependency

bernt-matthias:
imagemagick requirement was missing anyway. since for some reason
neither imagemagick not graphicsmagick can convert the image I
suggest to stick to svg only for now.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/newick_utils/newick_display.xml
+++ b/tools/newick_utils/newick_display.xml
@@ -1,4 +1,4 @@
-<tool id="newick_display" name="Newick Display" version="1.6">
+<tool id="newick_display" name="Newick Display" version="1.6+galaxy1">
     <description>visualize a phylogenetic tree</description>
     <edam_operations>
         <edam_operation>operation_0567</edam_operation>
@@ -36,15 +36,7 @@ nw_display
     -w  $width
     $radial
 
-    '$fileNewick' > output.svg
-
-#if $outformat == 'png':
-    && convert output.svg output.png
-    && mv output.png '$output'
-#else:
-    && mv output.svg '$output'
-#end if
-
+    '$fileNewick' > '$output'
     ]]></command>
     <inputs>
         <param name="fileNewick" format="txt,newick,nw,nwk,nhx,mothur.tre" type="data" label="Newick file"/>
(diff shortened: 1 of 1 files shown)


tools/newick_utils/newick_display.xml
```diff
@@ -36,15 +36,7 @@ nw_display
     -w  $width
     $radial
 
-    '$fileNewick' > output.svg
-
-#if $outformat == 'png':
```
mvdbeek: Is that maybe also due to the bug fixed by https://github.com/galaxyproject/galaxy/pull/9052 ? Because this would just move output.svg '$output' instead of running the conversion ?

bernt-matthias: nope. `convert` was just not installed. The container test output was `convert: command not found`. 

I tried on my local system to use `convert` and `gm convert` and both were not able to create a png from the svg.

mvdbeek: I was able to add imagemagick from conda-forge and got a png with this patch:
```diff
diff --git a/tools/newick_utils/newick_display.xml b/tools/newick_utils/newick_display.xml
index fbe029b5e..5dcc7b17a 100644
--- a/tools/newick_utils/newick_display.xml
+++ b/tools/newick_utils/newick_display.xml
@@ -5,6 +5,7 @@
     </edam_operations>
     <requirements>
         <requirement type="package" version="1.6">newick_utils</requirement>
+        <requirement type="package" version="7.0.9_6">imagemagick</requirement>
     </requirements>
     <command detect_errors="aggressive"><![CDATA[
 nw_display
```

But the png seems to miss fonts  ?
tree

mvdbeek: Cool, we can add a minimal fonts package like this to fix it:
```
diff --git a/tools/newick_utils/newick_display.xml b/tools/newick_utils/newick_display.xml
index fbe029b5e..7629f5458 100644
--- a/tools/newick_utils/newick_display.xml
+++ b/tools/newick_utils/newick_display.xml
@@ -5,6 +5,8 @@
     </edam_operations>
     <requirements>
         <requirement type="package" version="1.6">newick_utils</requirement>
+        <requirement type="package" version="7.0.9_6">imagemagick</requirement>
+        <requirement type="package" version="1">fonts-conda-forge</requirement>
     </requirements>
 nw_display
```

bernt-matthias: great. woudn't it be better to add the fonts package as a requirement to the imagemagick recipe?

mvdbeek: Possibly, I'm not sure how that decision is made. It seems imagemagick will use whatever is available.

bernt-matthias: lets see https://github.com/conda-forge/imagemagick-feedstock/pull/101 .. I will restore the original code.

bgruening: out of interest ... does graphicsmagick work?

bernt-matthias: gm gives the following error `gm convert: invalid primitive argument (small).`

PR #2714, opened 2019-12-01
circos: add tar requirement

bernt-matthias:
seems that another problem is that the generated configfiles are moved to working/... which seems not to work in the container tests. 

`mv: can't remove '/tmp/tmpzvg40cuz/job_working_directory/000/2/tmpnuxq8_9j': Read-only file system`

We could try to cp or symlink? Or should this be fixes in another way

FOR CONTRIBUTOR:

### DIFF
--- a/tools/circos/circos.xml
+++ b/tools/circos/circos.xml
@@ -62,13 +62,13 @@ touch circos/conf/karyotype-colors.conf &&
     #end if
 #end if
 
-cp -s '$circos_conf' circos/conf/circos.conf &&
-cp -s '$ticks_conf' circos/conf/ticks.conf &&
-cp -s '$ideogram_conf' circos/conf/ideogram.conf &&
-cp -s '$data_conf'  circos/conf/data.conf &&
-cp -s '$highlight_conf'  circos/conf/highlight.conf &&
-cp -s '$links_conf'  circos/conf/links.conf &&
-cp -s '$test_case_conf' circos/conf/galaxy_test_case.json &&
+cp '$circos_conf' circos/conf/circos.conf &&
+cp '$ticks_conf' circos/conf/ticks.conf &&
+cp '$ideogram_conf' circos/conf/ideogram.conf &&
+cp '$data_conf'  circos/conf/data.conf &&
+cp '$highlight_conf'  circos/conf/highlight.conf &&
+cp '$links_conf'  circos/conf/links.conf &&
+cp '$test_case_conf' circos/conf/galaxy_test_case.json &&
 
 ## Highlights
 #for $hi, $data in enumerate($sec_highlight.data):
--- a/tools/circos/circos.xml
+++ b/tools/circos/circos.xml
@@ -64,7 +64,7 @@ touch circos/conf/karyotype-colors.conf &&
 
 ln -s '$circos_conf' circos/conf/circos.conf &&
 ln -s '$ticks_conf' circos/conf/ticks.conf &&
-ls -s '$ideogram_conf' circos/conf/ideogram.conf &&
+ln -s '$ideogram_conf' circos/conf/ideogram.conf &&
 ln -s '$data_conf'  circos/conf/data.conf &&
 ln -s '$highlight_conf'  circos/conf/highlight.conf &&
 ln -s '$links_conf'  circos/conf/links.conf &&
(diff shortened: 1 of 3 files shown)


tools/circos/circos.xml
```diff
@@ -86,7 +86,7 @@ cp '${data.data_source}' circos/data/highlight-${hi}.txt &&
 #end for
 
 #if $outputs.tar == "yes"
-    tar cvfz circos.tar.gz circos >/dev/null
+	tar c circos | gzip - > circos.tar.gz
```
mvdbeek: Has this ever worked ? Shouldn't this be `-c`? Also I think we should just add gnu-tar in all tools that explicitly use tar ... that seems to be better for reproducibility.

mvdbeek:
> We could try to cp or symlink? Or should this be fixes in another way

I think symlink should be fine for now, but we may want to place the config files into a separate, writable directory.

tools/circos/circos.xml
bernt-matthias: probably yes. Apparently the dash is not strictly required by tar

tools/circos/circos.xml
```diff
@@ -62,13 +62,13 @@ touch circos/conf/karyotype-colors.conf &&
...
 #end if
 
-mv '$circos_conf' circos/conf/circos.conf &&
-mv '$ticks_conf' circos/conf/ticks.conf &&
-mv '$ideogram_conf' circos/conf/ideogram.conf &&
-mv '$data_conf'  circos/conf/data.conf &&
-mv '$highlight_conf'  circos/conf/highlight.conf &&
-mv '$links_conf'  circos/conf/links.conf &&
-mv '$test_case_conf' circos/conf/galaxy_test_case.json &&
+ln -s '$circos_conf' circos/conf/circos.conf &&
```
mvdbeek: Ah, I think this might get packaged up for use by the tool ... maybe we need to cp here. @erasche probably knows ?

tools/circos/circos.xml
erasche: works for me at least? but yes, adding gnu-tar sounds better. No the dash very much isn't required.

```
[hxr@mk:/tmp/tmp.XNdae8u3zj]$ mkdir asdf
[hxr@mk:/tmp/tmp.XNdae8u3zj]$ touch asdf/a
[hxr@mk:/tmp/tmp.XNdae8u3zj]$ tar c asdf | gzip - > circos.tgz
[hxr@mk:/tmp/tmp.XNdae8u3zj]$ tar tvfz circos.tgz
drwxrwxr-x hxr/hxr           0 2019-12-01 19:28 asdf/
-rw-rw-r-- hxr/hxr           0 2019-12-01 19:28 asdf/a
```

tools/circos/circos.xml
erasche: Yes, definitely, `cp` would be preferred here if `mv` isn't an option.

tools/circos/circos.xml
```diff
@@ -62,13 +62,13 @@ touch circos/conf/karyotype-colors.conf &&
...
 #end if
 
-mv '$circos_conf' circos/conf/circos.conf &&
-mv '$ticks_conf' circos/conf/ticks.conf &&
-mv '$ideogram_conf' circos/conf/ideogram.conf &&
-mv '$data_conf'  circos/conf/data.conf &&
-mv '$highlight_conf'  circos/conf/highlight.conf &&
-mv '$links_conf'  circos/conf/links.conf &&
-mv '$test_case_conf' circos/conf/galaxy_test_case.json &&
+cp -s '$circos_conf' circos/conf/circos.conf &&
```
erasche: the man page for `-s` reads `make symbolic links instead of copying`, isn't that explicitly what we're trying to avoid by switching to cp from ln?

tools/circos/macros.xml
```diff
@@ -12,6 +12,7 @@
         <requirement type="package" version="1.70">biopython</requirement>
         <requirement type="package" version="0.3.13">pybigwig</requirement>
         <requirement type="package" version="0.23">circos-tools</requirement>
+        <requirement type="package" version="1.32">tar</requirement>
```
nsoranzo: The latest version available on conda for macOS is 1.29, see https://github.com/conda-forge/tar-feedstock/pull/8

bernt-matthias: Hmm. Should be still stick to 1.32? I have used this version in quite a few tools. Probably 1.32 will be available for osx in the near future anyway?

mvdbeek: I don't think we'll need to fix the already merged ones, but maybe let's go with 1.29 for now ... the fixed version is probably not going to be release as 1.32 but 1.33 ?

PR #2717, opened 2019-12-01
prokka: more flexible testing

bernt-matthias:
Testdata differences were due the installed version is nowadays 1.13.7 and the version used to create the test data was 1.13. We could fix this now.

Probably does not cover the original error:

```
[tbl2asn] This copy of tbl2asn is more than a year old.  Please download the current version.
[22:27:37] Could not run command: tbl2asn -V b -a r10k -l paired-ends -M n -N 1 -y 'Annotated using prokka 1.13 from https://github.com/tseemann/prokka' -Z outdir\/prokka\.err -i outdir\/prokka\.fsa 2> /dev/null
```

which I don't get for local test. 

FOR CONTRIBUTOR:

### DIFF
--- a/tools/prokka/prokka.xml
+++ b/tools/prokka/prokka.xml
@@ -183,15 +183,25 @@ $input
             <param name="input" ftype="fasta" value="phiX174.fasta" />
             <param name="outputs" value="gff,gbk,fna,faa,ffn,sqn,fsa,tbl,tsv,err,txt" />
             <output name="out_gff" file="out.gff" />
-            <output name="out_gbk" file="out.gbk" lines_diff="14" />
+            <output name="out_gbk" >
+                <assert_contents>
+                    <has_text_matching expression="LOCUS" />
+                    <has_text_matching expression="//" />
+                </assert_contents>
+            </output>
             <output name="out_fna" file="out.fna" />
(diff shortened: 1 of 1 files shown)


mvdbeek:
That's quite annoying building in a check like that, even if I understand you don't want to receive bug reports for old versions ... 😠(xref https://www.biostars.org/p/301365/). I think we probably just need to rebuild the image, not much we can do on the Galaxy / IUC side. We could perhaps use libfaketime, but that wouldn't work on other architectures.

bernt-matthias:
Checking the tbl2asn recipe: the used URL is not stable. So I guess we can't rebuild the image.

bgruening:
@bernt-matthias if we rebuild the tbl2ans pkg with the latest version and bump the build-number of prokka, we should be save for one more year isn't it?

bernt-matthias:
Jep, but we can't rebuild the current version, because the download link is not stable. Guess we should build the current version (maybe with a detour via the depot .. even if this seems useless given that the tool stops working in 1yr).

bgruening:
Yeah abondend all our hopes for tbl2asn and decided to care only about the latest verion and not about reproducibility in bioconda. We once discssued even to patch this behaviour out, but decided against this I think.

PR #2720, opened 2019-12-01
hmmer3: update and nhmmer test fixes

bernt-matthias:
remaining problems seem due to the fact that nowadays v3.2.1 is used instead of 3.2. Could we fix this? Or better update test data?

since so far only tests changed I did not bump the tool version .. just kidding (I forgot it) .. 

FOR CONTRIBUTOR:


bernt-matthias:
I think the tests fail because now 3.2.1 is used instead of 3.2. 

`--update_test_data` ?

bernt-matthias:
Also upgraded to 3.3. .. according to the changelog there seem to be no changes in the CLI.

aliscore output in nhmmer scan creates segfault (reported here https://github.com/EddyRivasLab/hmmer/issues/190). So I enabled it for testing in nhmmer search... still keeping the test file for scan for when it works again. 

Needed to update quite a bit of the test data since numeric values changed.

bernt-matthias:
Also added some `expect_num_outputs` for tools with filters ... I was a bit lazy and just added tests instead of introducing macros for the outputs and testing filters just once... :(

bgruening:
Still failing ... :(

PR #2718, opened 2019-12-01
fsd+repmatch_gff3: use recent matplotlib

bernt-matthias:
old versions wer unsatisfiable by conda

Galaxy actually installed an `_uv_` environment in my local test. 

python (3?) should come automatically

FOR CONTRIBUTOR:
* [x] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [x] - License permits unrestricted use (educational + commercial)
* [ ] - This PR adds a new tool or tool collection
* [x] - This PR updates an existing tool or tool collection
* [ ] - This PR does something else (explain below)

PR #2727, opened 2019-12-02
tag_pileup_frequency test fixes

gregvonkuster:
FOR CONTRIBUTOR:


tools/tag_pileup_frequency/tag_pileup_frequency.xml
```diff
@@ -64,10 +64,10 @@ java -jar '$__tool_directory__/TagPileup.jar'
             <param name="combine_strands" value="0" />
             <param name="composite_smoothing_window" value="0" />
             <param name="output_heatmap" value="true" />
-            <output name="output1" file="output1.tabular" ftype="tabular"/>
+            <output name="output1" file="output1.tabular" ftype="tabular" compare="contains"/>
             <output_collection name="heatmaps" type="list">
-                <element name="dataset_2_dataset_1_read1_anti.tabular" file="input2_input1_read1_anti.tabular" ftype="tabular" compare="contains"/>
-                <element name="dataset_2_dataset_1_read1_sense.tabular" file="input2_input1_read1_sense.tabular" ftype="tabular" compare="contains"/>
+                <element designation="anti.tabular" file="input2_input1_read1_anti.tabular" ftype="tabular" compare="contains"/>
```
mvdbeek: I think `name` was correct here, designation is just for discovered output files

gregvonkuster: The jar file reads the base name of the inputs on disk and uses them to generate the name of the output.  For example, if the inputs are named dataset1 and dataset2, the output will be named dataset_2_dataset_1_read1_anti.tabular.

But planemo test generates temporary files for inputs, so the collection element names set in the tool config test section are not found during testing.  This is why I changed the name attribute to designation - I thought that was the proper way to define the test - and the test was successful in my environment.

Is there a way to change this back to using the name attribute and have planemo test work under these conditions?

mvdbeek: I think you'll need to symlink the inputs using their element_identifier, e.g `ln -s '$input1' '$input1.element_identifer' &&` and so on (you'll also need to escape the element_identifer, for instance using @ESCAPE_IDENTIFER@ as in https://github.com/galaxyproject/tools-iuc/pull/2718/commits/e227a13e968aaf9f5d26240c4e8bedcd9e3857bf)

gregvonkuster: Ah, ok, thanks!

tools/tag_pileup_frequency/tag_pileup_frequency.xml
```diff
@@ -64,12 +64,8 @@ java -jar '$__tool_directory__/TagPileup.jar'
             <param name="proper_pe" value="false" />
             <param name="combine_strands" value="0" />
             <param name="composite_smoothing_window" value="0" />
-            <param name="output_heatmap" value="true" />
+            <param name="output_heatmap" value="false" />
```
mvdbeek: Should I give this a go ? I assume this test used to work prior to isolating the upload paths.

gregvonkuster: Yes, please give it a go - I'd love to see how this should work!  ;)  Thanks for the help.

gregvonkuster: The output naming convention for heatmap ouitput collection elements is something like <input2 file name base>_<input1 file name base>_read1_anti.tabular

bernt-matthias: We should add a test that also outputs the heatmaps in order to see if the updated requirements work.

mvdbeek:
@bernt-matthias I've restored the test, but it does look like this tool doesn't do any plotting (only tabullar outputs). I'll see if it passes without the fonts as well.

mvdbeek:
Yep, works fine without the fonts.

tools/tag_pileup_frequency/tag_pileup_frequency.xml
gregvonkuster: Thanks so much for the help on this one!  It's good to see how this works.

mvdbeek: No worries, thanks for getting this started 👍

PR #2745, opened 2019-12-10
Fix resize_coordinate_window for python 3

gregvonkuster:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/resize_coordinate_window/resize_coordinate_window.py
+++ b/tools/resize_coordinate_window/resize_coordinate_window.py
@@ -1,4 +1,4 @@
-from __future__ import print_function
+
 
 import argparse
 import fileinput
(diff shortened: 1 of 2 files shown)


tools/resize_coordinate_window/resize_coordinate_window.py
```diff
@@ -25,7 +25,7 @@ def stop_err(msg):
 args = parser.parse_args()
 
 extend_existing = args.extend_existing == 'existing'
-out = open(args.output, 'wb')
+out = open(args.output, 'w')
```
mvdbeek: Can you also use `sys.exit(msg)` on line 13 ? I suspect otherwise msg would need to be bytestring.

gregvonkuster: Sure, thanks @mvdbeek for pointing this out.  I'm still getting familiar with some of these Python 3 twists - the ```sys.exit([arg])``` description here https://docs.python.org/3/library/sys.html, states that integers are ok, so I left it alone.  However, it also states that anything except an integer or None is printed to stderr and the exit code is set to 1.  So should I also delete line 12 as part of this fix?

mvdbeek: Yes, that's more concise and that way we don't need to care about how we write to sys.stderr.

gregvonkuster: Done - thanks! ;)

PR #2752, opened 2019-12-16
xpath: try to make output type depend on expression

bernt-matthias:
seems not to work as intended

FOR CONTRIBUTOR:


bernt-matthias:
- with `format` in the `<data>` tag the format is always xml
- without `format` the format is not set (i.e. it becomes data)

bernt-matthias:
Yeah, but does not work yet :) .. and I don't understand why :(

bernt-matthias:
With this https://github.com/galaxyproject/galaxy/pull/9127 it would work

bernt-matthias:
Alternatively `auto_format="true"` would work. But then we would require a xml declaration (`<?xml ` in the first line) which is currently the basis of Galaxy's xml sniffer.

erasche:
Xpath won't necessarily return a valid XML document, so they don't add the header either. There will always be edge cases, but, maybe it's better than nothing?

looking at `config/datatypes_conf.xml.sample` with `xpath -q -e '...'`

query | sniffs as XML | valid XML
-- | -- | --
`//datatype` | yes | no
`//datatypes` | yes | yes
`//datatype/@type | no | no

So I think the galaxy sniffer is too restrictive here. What if we checked with `xmllint` or so? (e.g.` xmllint output 2&>1 >/dev/null; if $? > 0; then mv output output.txt`)

I'd propose patching the datatype but it would add another dependency + be significantly slower (unnecessarily) for most xml files on upload?

bernt-matthias:
The `xmllint` approach for the tool sounds good. Still the problem is to set the datatype dynamically. Lets see if there will be some discussion here https://github.com/galaxyproject/galaxy/pull/9127

bernt-matthias:
Tried xmllint (from the libxml2 package). It errors on all files created by the tool, i.e., none is valid xml. 

On the pro side one may use `xmlint --xpath '$expression'` which gives the same result as the perl script.

erasche:
Huh, surprising it errors on everything. Even this one?

```
</datatypes>
$ cat config/datatypes_conf.xml.sample | xpath -q -e '//datatypes' | xmllint - > /dev/null
$ cat config/datatypes_conf.xml.sample | xpath -q -e '//datatype' | xmllint - > /dev/null
-:2: parser error : Extra content at the end of the document
<datatype extension="afg" type="galaxy.datatypes.assembly:Amos" display_in_uploa
^
```

for me the first query works fine, querying a single node should return valid xml.

bernt-matthias:
@hexylena `xmlint` seems to work. Was just asking myself if it would be better to transform the output to valid xml in case it isn't .. by adding `<xml>...</xml>` or something similar. 

Maybe the tool would be easier to use in workflows.

hexylena:
In the case where it sniffs as xml and is returning a set of nodes, then, definitely would be useful!

bernt-matthias:
Seems that xmllint differentiates the two cases (list of nodes and text) with the stderr messages:

```
test-data/2.xml:2: parser error : Extra content at the end of the document
test-data/3.xml:1: parser error : Start tag expected, '<' not found
```

do you think its a good idea to base on this? Btw. `xmllint --xpath "QUERY"` can achieve the same as the perl script :)

hexylena:
> do you think its a good idea to base on this? 

I don't know, all of this is a bit scary. But, it sounds good? maybe it's the most user-friendly thing. hopefully.

> Btw. xmllint --xpath "QUERY" can achieve the same as the perl script :)

oh, what?! awesome.

bgruening (approved):
It is an improvement over the old version, so we can ship it and rewrite it later if needed.

PR #2561, opened 2019-08-30
reference data table columns consistently by name

bernt-matthias:
just for consistency .. none of these were wrong as far a I could tell .. except for gffcompare where the asserts are now actually considered (since we now use this https://github.com/galaxyproject/galaxy/pull/8395).

FOR CONTRIBUTOR:
* [x] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [x] - License permits unrestricted use (educational + commercial)
* [ ] - This PR adds a new tool or tool collection
* [x] - This PR updates an existing tool or tool collection
* [ ] - This PR does something else (explain below)

PR #2754, opened 2019-12-17
Fix MultiGPS tool

gregvonkuster:
FOR CONTRIBUTOR:


abretaud:
@gregvonkuster last time we made changes to multigps (https://github.com/galaxyproject/tools-iuc/pull/2561) there were permission problems to upload it to the toolshed, did you change them?

gregvonkuster:
> @gregvonkuster last time we made changes to multigps (#2561) there were permission problems to upload it to the toolshed, did you change them?

Ah, sorry I missed that.  The MultiGPS tools in both TTS and MTS owned by iuc were both marked deprecated for some time since tests were not working.  I've "de-deprecated" the tools in both tool sheds now, so hopefully that will resolve permissions problems that you saw.  If not, then I'm not quite sure what could be causing them, but we can take a look if they continue.

tools/multigps/multigps.xml
```diff
@@ -282,15 +338,15 @@
...
+                                        <!-- <filter type="data_meta" key="dbkey" ref="expt" column="1"/> does not yet work in a repeat...-->
                                     </options>
                                     <validator type="no_options" message="A built-in reference genome is not available for the build associated with the selected input file"/>
                                 </param>
                             </when>
                             <when value="history">
                                 <param name="reference_genome" type="data" format="fasta" label="Using reference genome">
                                     <options>
-                                        <filter type="data_meta" key="dbkey" ref="expt"/>
+                                        <!-- <filter type="data_meta" key="dbkey" ref="expt"/> does not yet work in a repeat...-->
```
bgruening: Mh, this is wired. Can you create a Galaxy functional tool test that exposes this?

gregvonkuster: @bgruening Ah, yes, this tag reflects an old version of the tool, and should have been updated to something like:

```<filter type="data_meta" key="dbkey" ref="signal" column="1"/>```

I haven't worked with this tool for quite some time now, but IIRC, the ```<filter>``` tag will not work because the ```ref``` attribute refers to a param buried withing ```<repeat>``` blocks:

```
<repeat name="condition_repeat"...
    <repeat name="signal_repeat"...
        <param name="signal"...
```

The tool will still function correctly, but the dynamically generated select list for the reference genome (when performing motif-finding or using motif-priors) will not be filtered to the genome associated with the selected inputs for signal and control, so the user has to manually select the appropriate genome.

I believe this is a limitation of the underlying Galaxy tool parameter framework.  Un-commenting the ```<filter>``` tags, and correcting them by setting the ```ref``` attribute to be ```signal``` produces this exception when the tool is loaded in the tool panel:

```Current on-disk tool is not valid```.

Should I open a ticket on this?

bgruening: Adding a new Galaxy test tool under https://github.com/galaxyproject/galaxy/tree/dev/test/functional/tools that exposes this error would help a lot I think. Keep it minimal if possible.

Thanks @gregvonkuster and happy new year.

gregvonkuster: @bgruening I added a test in this https://github.com/galaxyproject/galaxy/pull/9196.  Thanks for the help on this!  :)

PR #2763, opened 2019-12-22
Don't let bracken write to kraken index dir

mvdbeek:
Writing to another index dir isn't good for reproducibility and doesn't work in containers.
I am not sure this works correctly, it would be good if someone who
knows bracken could check if this still works (maybe @dfornika ?)
If it doesn't I think we should copy the kraken databases instead
of symlinking them.

FOR CONTRIBUTOR:


dfornika:
I was uncertain about this when I wrote the bracken data manager but I'm open to alternative solutions.

mvdbeek:
I'm symlinking the kraken index files into the working directory and then deleting the symlinks. I guess what I don't know is if this would be sufficient for tools that use the bracken indexes to work ?

dfornika:
Are you able to do a test-run of this bracken tool using the new database? 

https://github.com/galaxyproject/tools-iuc/blob/master/tools/bracken/est-abundance.xml

It's the only tool that I'm aware of that uses the bracken index.

mvdbeek:
Thanks @dfornika, I've removed the kraken database from est-abundance and the tool still passes testing, so I think this works fine.

data_managers/data_manager_build_bracken_database/data_manager/bracken_build_database.xml
```diff
@@ -4,16 +4,22 @@
     <requirements>
         <requirement type="package" version="2.5">bracken</requirement>
         <requirement type="package" version="2.0.8_beta">kraken2</requirement>
+        <requirement type="package" version="4.6.0">findutils</requirement>
     </requirements>
     <command>
     <![CDATA[
+        #import os##set db_dir = os.path.basename($kraken_db.fields.path)
+        mkdir '$db_dir' &&
+        ln -s '${kraken_db.fields.path}'/* '$db_dir/' &&
```
bgruening: I'm not sure this will work in production.
Indices are often mounted read-only on the cluster nodes. Can we copy the kraken indices?

mvdbeek: Right, so the original approach won't work for sure. The indices are also read-only on docker, that's why this is necessary and works (keep in mind the `/*`, we're not writing to the original, read-only location).

bgruening: Ah, I see db_dir is in the cwd ... but then why are you removing the symlinks again?

mvdbeek: To avoid dangling symlinks / giving the impression that kraken indices are available

PR #2772, opened 2020-01-06
tetyper: Optionally read variant profiles from tool data table

dfornika:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/tetyper/tetyper.xml
+++ b/tools/tetyper/tetyper.xml
@@ -23,16 +23,16 @@
             --min_qual '${min_qual}'
             #if $snp_profiles_source.snp_profiles_source_selector:
                 #if str( $snp_profiles_source.snp_profiles_source_selector ) == "tool_data_table":
-                    --snp_profiles = '$snp_profiles_source.snp_profiles.fields.path'
-                #else:
-                    --snp_profiles = '$snp_profiles_source.snp_profiles'
+                    --snp_profiles '$snp_profiles_source.snp_profiles.fields.path'
+                #elif str( $snp_profiles_source.snp_profiles_source_selector ) == 'history':
+                    --snp_profiles '$snp_profiles_source.snp_profiles'
                 #end if
             #end if
             #if $struct_profiles_source.struct_profiles_source_selector:
                 #if str( $struct_profiles_source.struct_profiles_source_selector ) == "tool_data_table":
-                    --struct_profiles = '$struct_profiles_source.struct_profiles.fields.path'
-                #else:
-                    --struct_profiles = '$struct_profiles_source.struct_profiles'
+                    --struct_profiles '$struct_profiles_source.struct_profiles.fields.path'
+                #elif str( $struct_profiles_source.struct_profiles_source_selector ) == 'history':
+                    --struct_profiles '$struct_profiles_source.struct_profiles'
                 #end if
             #end if
             --outprefix output
--- a/tools/tetyper/tetyper.xml
+++ b/tools/tetyper/tetyper.xml
@@ -60,7 +60,7 @@
                 <option value="history" selected="true">From history</option>
             </param>
             <when value="tool_data_table">
-                <param name="snp_profiles" type="select" format="tabular" optional="true">
+                <param name="snp_profiles" type="select" format="tabular" label="SNP Profile" optional="true">
                     <options from_data_table="tetyper_snp_profiles">
                         <validator type="no_options" message="No TETyper SNP profiles are available" />
                     </options>
@@ -76,7 +76,7 @@
                 <option value="history" selected="true">From history</option>
             </param>
             <when value="tool_data_table">
-                <param name="struct_profiles" type="select" format="tabular" optional="true">
+                <param name="struct_profiles" type="select" format="tabular" label="Structural Variant Profile" optional="true">
                     <options from_data_table="tetyper_struct_profiles">
                         <validator type="no_options" message="No TETyper structural variant profiles are available" />
                     </options>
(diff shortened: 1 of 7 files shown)


tools/tetyper/tetyper.xml
```diff
@@ -44,14 +54,46 @@
                 <param format="fastq,fastq.gz" name="reverse_input" type="data" label="Reverse strand" help="FASTQ dataset"/>
             </when>
         </conditional>
+        <conditional name="snp_profiles_source">
+            <param name="snp_profiles_source_selector" type="select" label="Select a SNP profile from your history or use one from a tool data table?"
+                   help="Screening files must be stored in the 'tetyper_snp_profiles' tool data table">
```
bgruening: the help seems to be confusing to a user ... maybe something like, if you don't see any data please contact your galaxy admin, or just remove this sentence.

tools/tetyper/tetyper.xml
```diff
@@ -44,14 +54,46 @@
...
+            </param>
+            <when value="tool_data_table">
+                <param name="snp_profiles" type="select" format="tabular">
+	            <options from_data_table="tetyper_snp_profiles">
+	                <validator type="no_options" message="No TETyper SNP profiles are available" />
+                    </options>
+	        </param>
+            </when>
+            <when value="history">
+                <param name="snp_profiles" type="data" format="tabular" label="SNP Profile" help="A three-column tabular file with SNP profile definitions" optional="true"/>
```
bgruening: why should that be optional? Should not the upper conditional then optional?

tools/tetyper/tetyper.xml
```diff
@@ -44,14 +54,46 @@
...
+            </param>
+            <when value="tool_data_table">
+                <param name="struct_profiles" type="select" format="tabular">
+	            <options from_data_table="tetyper_struct_profiles">
+	                <validator type="no_options" message="No TETyper structural variant profiles are available" />
+                    </options>
+	        </param>
+            </when>
+            <when value="history">
+                <param name="struct_profiles" type="data" format="tabular" label="Structural Variant Profile" help="A two-column tabular file with structural variant profile definitions" optional="true"/>
```
bgruening: optional? See above.

tools/tetyper/tetyper.xml
```diff
@@ -80,8 +122,31 @@
...
+                <param name="snp_profiles_source_selector" value="history" />
+                <param name="snp_profiles" value="Tn4401b_snp_profiles.txt" />
+            </conditional>
+            <conditional name="struct_profiles_source">
+                <param name="struct_profiles_source_selector" value="history" />
+                <param name="struct_profiles" value="Tn4401b_struct_profiles.txt" />
+            </conditional>
+            <output name="summary" file="output_summary_2.txt" />
+        </test>
+                <test>
```
bgruening: indentation ...

tools/tetyper/tetyper.xml
```diff
@@ -80,8 +122,31 @@
                 <param name="forward_input" value="ERR1911133_Tn4401b-1_mapped_subsampled_1.fastq" />
                 <param name="reverse_input" value="ERR1911133_Tn4401b-1_mapped_subsampled_2.fastq" />
             </conditional>
-            <param name="snp_profiles" value="Tn4401b_snp_profiles.txt" />
-            <param name="struct_profiles" value="Tn4401b_struct_profiles.txt" />
+            <conditional name="snp_profiles_source">
+                <param name="snp_profiles_source_selector" value="history" />
```
bgruening: can you include a test for the data_manager?

tools/tetyper/tetyper.xml
dfornika: You're right, I've removed it.

tools/tetyper/tetyper.xml
dfornika: The tool will work without these variant profile files, so I wanted to make sure the user could still launch the tool without them. I've made all variant profile params optional.

tools/tetyper/tetyper.xml
dfornika: I'm not sure what you mean here. I've included tests that read from both the history and the tool data tables. There is no dedicated data manager tool for this. I was thinking that admins could either install manually or use 'data_manager_manual' since they're just plaintext tabular files.

tools/tetyper/tetyper.xml
bgruening: My thinking is if the user has already chosen "from history" the intention is to provide a files and it should not be optional here. The option "history/loc" should maybe be optional?

bgruening: Does this make sense @dfornika?

dfornika: Thanks for checking. I'm still a bit confused. I just want to make sure that the user is able to run the tool without providing either of these variant profile files.

There are only two options for each of the 'source_selector' params (`history` or `tool_data_table`). `history` is selected by default. If the user wants to run the tool without a variant profile, they can go ahead without entering a value there.

If the params aren't optional when the selections are set to `history`, then how would a user launch the tool without a variant profile?

Note that `history` is selected by default, so the user hasn't necessarily signalled any intention when that is selected.

bgruening: Would my commit work for you?

> Note that history is selected by default, so the user hasn't necessarily signalled any intention when that is selected.

Thats what I want to change and make it optional.

dfornika: Sorry I haven't responded more quickly. I see that the first test is failing. I've tried running the current version of the tool through the Galaxy web interface and wasn't able to run the tool without the variant profile files. I think this still needs some tweaking. Will give it some more attention as soon as I can.

bgruening:
Thanks @dfornika!

PR #2777, opened 2020-01-12
Update Multiqc: handle plots outputs and --flat option.

loraine-gueguen:
FOR CONTRIBUTOR:

When there are 100 or more samples (or when using the --flat option), the plots are non-interactive (flat) images within the report, and the plot tabular data cannot be downloaded through the report as it is the case usually, and are only available as supplementary output files. 
To be able to get these plot tabular data, I have added them as output collection, and I have added the --flat option to be able to test the use case.


tools/multiqc/multiqc.xml
```diff
@@ -809,7 +811,7 @@ sp:
         </repeat>
         <param name="title" type="text" value="" optional="true" label="Report title" help="It is printed as page header"/>
         <param name="comment" type="text" value="" optional="true" label="Custom comment" help="It will be printed at the top of the report"/>
-        <param name="saveLog" type="boolean" truevalue="True" falsevalue="False" checked="False" label="Output the multiQC log file?" help="This is mostly useful for debugging purposes"/>
```
bgruening: Is there a reason why you removed this option @loraine-gueguen?

loraine-gueguen: Because this line was duplicated

loraine-gueguen:
Is it ok to be merged ?

PR #2779, opened 2020-01-13
Fix varscan_somatic test

mvdbeek:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/varscan/varscan_somatic.xml
+++ b/tools/varscan/varscan_somatic.xml
@@ -410,7 +410,7 @@
             </conditional>
             <assert_stderr>
                 <has_line_matching
-                expression="Min coverage:&#09;2x for Normal, 2x for Tumor" />
+                expression=" *Min coverage:&#09;2x for Normal, 2x for Tumor" />
                 <has_line_matching
                 expression="Min reads2:&#09;1" />
                 <has_line_matching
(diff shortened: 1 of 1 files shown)


tools/varscan/varscan_somatic.xml
```diff
@@ -410,7 +410,7 @@
             </conditional>
             <assert_stderr>
                 <has_line_matching
-                expression=" Min coverage:&#09;2x for Normal, 2x for Tumor" />
+                expression="Min coverage:&#09;2x for Normal, 2x for Tumor" />
```
wm75: What about:
```suggestion
                expression=" *Min coverage:&#09;2x for Normal, 2x for Tumor" />
```
?
That space isn't a typo, but is really part of the output. It just ends up on a different line in the containerized test because of a setlocale warning. Simply removing the space from the pattern breaks local, non-containerized testing.

mvdbeek: Sounds good. While we're at it, should we annotate varscan to require the extended base container with the locale ?

bgruening: Let's use the smaller container for the moment.

bgruening:
Thanks @mvdbeek!

PR #2800, opened 2020-01-21
Adding `findutils` requirement to JBrowse

almahmoud:
`jbrowse.py` uses subprocess module to call `find` with `-xtype` flag, but the `find` command in the biocontainer does not support that flag. Adding the `findutils` dependency guarantees the existence of the proper `find`, although might not be necessary in all environments (eg: works without the requirement when using Ubuntu base image, but doesn't work in the container which I believe uses CentOS base). Tested by running `planemo mull` after adding the requirement and it worked (temporary container with the added dependency at `cloudve/jbrowse:1.16.5`)

FOR CONTRIBUTOR:

### DIFF
--- a/tools/jbrowse/macros.xml
+++ b/tools/jbrowse/macros.xml
@@ -15,7 +15,7 @@
     </requirements>
   </xml>
   <token name="@DATA_DIR@">\$GALAXY_JBROWSE_SHARED_DIR</token>
-  <token name="@WRAPPER_VERSION@">galaxy6</token>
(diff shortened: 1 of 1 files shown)


afgane:
Thanks @hexylena.

Where can we find the container that got built as a result of this PR being merged?

mvdbeek:
That'll automatically happen tonight. I'll open a PR now and we'll get it in half an hour.

mvdbeek:
https://github.com/BioContainers/multi-package-containers/pull/962

PR #2814, opened 2020-01-23
Creation of the Bio-tradis tool suite for Transposon insertion sequencing analyses

Delphine-L:
This PR contains 3 tools : 
- **bacteria tradis**, for read trimming, mapping, and read counts
- **tradis gene insert sites**, provides read and insertion counts per gene
- **tradis essentiality**, that predicts essential genes.

I have encountered some troubles for wrapping these tools : 
- **tradis gene insert sites** needs an Embl file as input, but planemo does not recognise the format. (I also don't see in on Galaxy main, although it is present in the datatypes_com.xml.sample) 
- **tradis essentiality**. I tried both the _"discover_dataset"_ method (as shown here : https://github.com/jmchilton/galaxy/blob/2909e74642180bd818019ebdcb62e62f12e56e69/test/functional/tools/tool_provided_metadata_3.xml#L15 ) and the _"from_work_dir"_ method, but I couldn't get the tool to recognize the correct outputs. I made it work by changing the output names in the command line.

I would be grateful if someone could give me ideas on how to make these work properly.

FOR CONTRIBUTOR:


bgruening:
@Delphine-L the test data is way to big :-(

Delphine-L:
@bgruening 

I reduced the size of the files, I hope it is small enough.
I am not sure I can reduce it much more.

Delphine-L:
Reduced the test data even more. This is the absolute smallest size that will allow the tests to work

mvdbeek:
Thanks @Delphine-L, looks good, but this is missing a .shed.yml file, can you add that please ?

Delphine-L:
I put the "Transposon Insertion Sequencing" section in the shed.yml . I think we should create it and put Tradis tools in there as well, what do you think?

mvdbeek:
That is a fairly specific category. Maybe genome annotation and variant analysis are broader categories that fit ? Not opposed to more specialised categories either, in that case someone (@martenson ?) needs to create that category.

Delphine-L:
I don't think it is more specific than other categories we have. 
The TIS category would contain tools equivalent to the tools contained in the RNA-seq section (even if less numerous) :  preprocessing, alignment, counts, comparisons... 

If we put them in genome annotation, we either put everything and then we have mapping tools in there, or we separate the suite into several sections but then It gets confusing for the users.

Delphine-L:
We can put the tools in Genome Annotation for now, and discuss that for later...

mvdbeek:
Thanks @Delphine-L!

PR #2819, opened 2020-01-24
Add support for vcf_bgzip input to gemini

wm75:
FOR CONTRIBUTOR:


mvdbeek:
Could you also take a look at the failing gemini tests (annotate and burden)  in https://mvdbeek.github.io/iuc-tool-test-results/#functional.test_toolbox.TestForTool_gemini_annotate.test_tool_000001 ?

tools/gemini/gemini_annotate.xml
```diff
@@ -19,9 +19,19 @@
 <![CDATA[
 
     ## For GEMINI to work correctly, tabixed file must have form [name].[bed|vcf].gz
-    #set $tabixed_file = "tabixed.%s.gz" % $annotate_source.ext
-    bgzip -c '$annotate_source' > '$tabixed_file' &&
-    tabix -p '$annotate_source.ext' '$tabixed_file' &&
+    #if $annotate_source.ext == 'vcf_bgzip':
+        ## File is bgzipped already -> just tabix it
+        #set $tabixed_file = 'tabixed.vcf.gz'
```
mvdbeek: vcf_bgzip should already be indexed and available as `$tabixed_file.tabix_index`

wm75: Ah, of course! Let me fix this.

wm75:
> Could you also take a look at the failing gemini tests (annotate and burden) in https://mvdbeek.github.io/iuc-tool-test-results/#functional.test_toolbox.TestForTool_gemini_annotate.test_tool_000001 ?

Good point! Had missed those, but will fix here (though currently I have no idea why you're getting these errors)

mvdbeek:
The `-e +` assertion is nonsense, but I couldn't figure out what's wrong with burden. This works if you bump to 0.30.2, but then gemini load fails.

wm75:
Yeah, a version upgrade is on my TO DO list, but likely requires quite some work.

wm75:
@mvdbeek I have a speculative idea about gemini_burden, but to verify this: which pandas version is in the container?

wm75:
Ah, the build date of the container is revealing enough I think. So here's my hypothesis:

The container is from June 2018 and that means it should still have a pandas version affected by this bug: https://github.com/pandas-dev/pandas/issues/21561
The relevant source code for gemini burden is:
https://github.com/arq5x/gemini/blob/cc3563fc50275d12f5af30550bf693853efbb41c/gemini/tool_burden_tests.py#L268

When I test locally after downgrading the pandas in my gemini env to pandas-0.23.1 I'm getting empty output indeed.

Question is: how do I fix this/get the container updated? Is requiring pandas at a suitable version going to help?

mvdbeek:
> Question is: how do I fix this/get the container updated?

We could "cheat" and list an explicit pandas version for gemini burden. Else you'd have to trigger a new conda build for gemini 0.20.1.

mvdbeek:
So 
```
diff --git a/tools/gemini/gemini_burden.xml b/tools/gemini/gemini_burden.xml
index 4300a7655..dcdfd7102 100755
--- a/tools/gemini/gemini_burden.xml
+++ b/tools/gemini/gemini_burden.xml
@@ -5,7 +5,9 @@
         <token name="@BINARY@">burden</token>
         <token name="@GEMINI_BUG_WARNING@">Caveat: due to a bug in this version of GEMINI, filtering on alternate allele frequency will get disabled when you select 'All non-synonymous variants' as the types of variants to be used for the burden calculations above.</token>
     </macros>
-    <expand macro="requirements" />
+        <requirement type="package" version="0.24.2">pandas</requirement>
+    </expand>
     <expand macro="stdio" />
     <command>
```
works fine, great detective work @wm75!
I can try bumping the old build and we can see how much of a hassle that would be.

wm75:
Exactly the diff I just made locally :smile: Just had no time to test it yet.
Just let me know if that's the preferred solution and I'll push it.

mvdbeek:
Fingers crossed for https://github.com/bioconda/bioconda-recipes/pull/19953. If we can get a new build that'd be better for other people possibly using this container and deployers not needing both images.

bgruening:
We need to bump the version number, right?

wm75:
Of course, thanks! Glad if that's all that's left to do here :smile:

PR #2820, opened 2020-01-25
Add 10X H5 inputs to ScanPy

mtekman:
FOR CONTRIBUTOR:
* [X] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [X] - License permits unrestricted use (educational + commercial)
* [ ] - This PR adds a new tool or tool collection
* [X] - This PR updates an existing tool or tool collection
* [ ] - This PR does something else (explain below)

PR #2824, opened 2020-01-26
Add 10X H5 inputs to Anndata

mtekman:
FOR CONTRIBUTOR:

This adds 10x H5 compatibility for the datasets produced by DropletUtils, and also for the example datasets given on the 10X website.

It makes the following PR redundant https://github.com/galaxyproject/tools-iuc/pull/2820, which I will now close.


tools/anndata/import.xml
```diff
@@ -124,6 +128,9 @@ adata.write('anndata.h5ad')
                         <param name="input" type="data" format="tabular,csv,tsv" label="Annotated data matrix"/>
                         <param name="first_column_names" type="boolean" truevalue="True" falsevalue="False" checked="true" label="Does the first column store the row names?"/>
                     </when>
+                    <when value="10x_h5" >
+                        <param name="input" type="data" format="h5,h5ad" label="Data matrix"/>
```
bgruening: Do we need a new Galaxy filetype for this one? Why is the anndata filetype still here?

mtekman: dropletutils outputs an h5ad in galaxy, but it is actually an h5 object. I just changed this now in this PR too

mtekman:
I changed dropletutils in this PR to output H5, and then I changed this AnnData wrapper to import 10X H5 datasets via ScanPy (since AnnData has no native method for it).

tools/anndata/macros.xml
```diff
@@ -4,6 +4,7 @@
     <xml name="requirements">
         <requirements>
             <requirement type="package" version="@VERSION@">anndata</requirement>
+            <requirement type="package" version="1.4.4.post1">scanpy</requirement>
```
bebatut: Scanpy is already a requirement for the `import.xml` wrapper

mtekman: removed, thanks

mtekman:
@bebatut I can't seem to replicate the travis test error locally on my machine. For some reason the loom test fails (but is green on travis), and the 10x_h5 test works fine on my machine (but is red on travis).

Is there something I am missing? As far as I can tell, the tool is not targetting the 19.01 release, right?

mtekman:
I seperated the  dropletutils and anndata into seperate PRs. It got messy.

mtekman:
I thought the error was related to the output not being parsed properly, which is why I added a large test output dataset `dropletutils_output.h5ad` instead of just asserting the contents of it as before.... but now it seems like the error is something to do with scanpy.

I am unable to replicate this at all in a conda environment on any local machine.

mtekman:
So the error in travis is related to a function `_downsample` not being cached properly. 

This led to a bunch of SO posts where it is suggested to set the `NUMBA_CACHE_DIR` to something else. Here, I have set it it to `/tmp`...

mvdbeek:
> So the error in travis is related to a function `_downsample` not being cached properly.

Yeah, because it tries to cache in the not-wrieable home dir. Add `profile="18.01"` to the <tool/> tag and you get a clean, writable home dir as part of the job

mvdbeek:
You should always add a relatively recent `profile` anyway, those are always better defaults.

mtekman:
Hey thanks! I will try this now

mvdbeek:
Sure thing. I'm contemplating setting this for containerized jobs by default, given that as a deployer you'd really need to go out of your way to provide a writable home dir in a container.

mtekman:
@mvdbeek it looks like it's working beautifully -- thanks, I can progress with PR more easily now

mtekman:
force pushed to get rid of the large output test data

PR #2828, opened 2020-01-27
Fix ncbi_eutils tests

mvdbeek:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/ncbi_entrez_eutils/epost.xml
+++ b/tools/ncbi_entrez_eutils/epost.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<tool id="ncbi_eutils_epost" name="NCBI EPost" version="@WRAPPER_VERSION@" profile="18.01">
+<tool id="ncbi_eutils_epost" name="NCBI EPost" version="@WRAPPER_VERSION@" profile="@PROFILE@">
   <description>post UIDs to NCBI History Server</description>
   <macros>
     <import>macros.xml</import>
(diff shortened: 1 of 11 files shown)


tools/ncbi_entrez_eutils/epost.xml
```diff
@@ -1,12 +1,13 @@
 <?xml version="1.0"?>
-<tool id="ncbi_eutils_epost" name="NCBI EPost" version="@WRAPPER_VERSION@">
+<tool id="ncbi_eutils_epost" name="NCBI EPost" version="@WRAPPER_VERSION@" profile="18.01">
```
wm75: Any specific reason why you'd want to require 18.01 specifically here?

mvdbeek: For the isolated home directory introduced with the 18.01 defaults. Otherwise the biopython entrez handler attempts to write to the real home directory, which fails with permission errors.

wm75: That's definitely good to know! What I actually meant though was why you use the explicit profile version here and not the token as elsewhere.

mvdbeek: oversight

mvdbeek:
```suggestion
<tool id="ncbi_eutils_epost" name="NCBI EPost" version="@WRAPPER_VERSION@" profile="@PROFILE@">
```

PR #2835, opened 2020-01-28
iqtree: handle '+' character in custom models

### LINKED ISSUE #2834: iqtree: '+' char not handled in custom models
When running `iqtree`, models are combined using a `+` character, such as `GTR+G`:

http://www.iqtree.org/doc/Substitution-Models

...But when a model such as `GTR+G` is entered via the galaxy tool, it results in a command-line like this:
```
iqtree [parameters omitted for brevity] -m 'GTRXG'
```

...and the following message on stderr:

```
ERROR: Invalid model name GTRXG
```

dfornika:
FOR CONTRIBUTOR:

Fixes #2834

### DIFF
--- a/tools/iqtree/iqtree.xml
+++ b/tools/iqtree/iqtree.xml
@@ -632,6 +632,34 @@ IQ-TREE also works for codon, binary and morphogical data.
                 </assert_contents>
             </output>
         </test>
+        <test>
+            <param name='s' value='example.phy' />
+            <output name='iqtree'>
+                <assert_contents>
+                    <has_text_matching expression=".*Human.*Whale.*" />
+                </assert_contents>
+            </output>
+            <output name='treefile'>
+                <assert_contents>
+                    <has_text_matching expression="\(LngfishAu:(\d|\..)+,\(LngfishSA:(\d.)+,.*" />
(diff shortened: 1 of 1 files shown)


dfornika:
@wm75, ah that might be a cleaner solution (although some of those characters aren't used in the model specifications). I did notice that the model param seems to be the only one that doesn't use the `sanitize_query` macro.

dfornika:
I think this might be ready to merge(?)

mvdbeek:
Isn't it a problem that you're adding `'` as an allowed character ?

dfornika:
@mvdbeek I believe that's removed in the `sanitize_query` macro? Unless I'm misunderstanding how it works.

https://github.com/galaxyproject/tools-iuc/blob/ea9a67b5711b53fb911a0c50b6b49eb9db2366a7/tools/iqtree/iqtree_macros.xml#L10-L18

wm75:
Almost - this macro quotes the single quotes to turn them into literal quotes for the shell :smile:

wm75:
Thanks @dfornika!

mvdbeek:
The macro looks so close to the sanitizer I thought I was looking at the sanitizer!

PR #2838, opened 2020-01-29
Hicexplorer: update to version 3.4

joachimwolff:
FOR CONTRIBUTOR:

@bgruening: Please review and merge this PR. Needs to be installed on our server until the paper hand-in deadline.


tools/hicexplorer/hicAverageRegions.xml
```diff
@@ -41,6 +42,12 @@
                  <param name="downstreamRange" type="integer" value="" label="Downstream range"/>
             </when>
         </conditional>
+        <param name="coordinatesToBinMapping_selector" type="select" label="Define start value of range" help="If the region contains start and end coordinates, define if the start, center (start + (end-start) / 2) or end bin should be used as start for range.">
+                <option value="start">Start</option>
+                <option value="center" selected="True">Center</option>
+                <option value="end" selected="True">End</option>
+
```
bgruening: remove line here

tools/hicexplorer/hicCompartmentalization.xml
```diff
@@ -1,7 +1,7 @@
-<tool id="hicexplorer_hiccompartmentspolarization" name="@BINARY@" version="@WRAPPER_VERSION@.0">
+<tool id="hicexplorer_hiccompartmentalization" name="@BINARY@" version="@WRAPPER_VERSION@.0">
```
bgruening: If you change the ID, you get a new Galaxy tool. This mean no update in workflows is possible and we need to deprecate/hide the old one. If you keep the ID a smooth update process would be possible.

tools/hicexplorer/hicCompartmentalization.xml
```diff
@@ -17,10 +17,16 @@
             --obsexp_matrices $m
 
             --pca '$pca'
-            ##--quantile  #echo int($quantile) # 
-
+            --quantile  #echo int($quantile) #
```
bgruening: the casting with int() and the echo should not be needed

tools/hicexplorer/hicCompartmentalization.xml
```diff
@@ -36,8 +42,19 @@
...
-        <param name='outliers'  type="float"  label="Outliers" help='The number of outliers to be removed.' value="0.0"/> -->
+        <param name='quantile'  type="integer" label="Quantile" help='The number of to be used quantiles.' value="30"/>
+        <param name='outliers'  type="float"  label="Outliers" help='The number of outliers to be removed.' value="0.0"/>
+         <conditional name="set_offset_conditional">
+            <param name='set_offset_selector' type='select' label='Set an offset for the main diagonal'>
+                <option value='' selected="True">No offset</option>
+                <option value='set_offset'>Set offset</option>
+            </param>
+            <when value=''/>
+
```
bgruening: remove line here

bgruening: This entire conditional is not needed, isn't it? You could simply make the set_offset_value optional="true"

tools/hicexplorer/hicCompartmentalization.xml
```diff
@@ -62,54 +79,28 @@
             <param name="pca" value="hicCompartmentsPolarization/pca1.bedgraph"/>
             <output name="outFileName" file="hicCompartmentsPolarization/compartmentsPolarizationRatio.png" ftype="png" compare="sim_size"/>
         </test>
+        <!-- <test>
```
bgruening: I guess this tool needs a few more test, this one above it not enough imho.

tools/hicexplorer/hicNormalize.xml
```diff
@@ -16,26 +16,56 @@
...
+        
+        <conditional name="zero_threshold_conditional">
+            <param name='zero_threshold' type='select' label='Set values below this threshold to zero'>
+                <option value=''>No threshold</option>
+                <option value='set_threshold'>Set threshold</option>
+            </param>
+            <when value=''/>
+            
+            <when value='set_threshold'>
+                <param name="zero_threshold_value" type="float" value="1.0" label="Threshold value"/>
```
bgruening: can we make this simply optional="true" and avoid the conditional?

tools/hicexplorer/hicCompartmentalization.xml
joachimwolff: We decided to rename it in HiCExplorer therefore we need to rename it here too. Anyhow, it would be enough to change the name because this is what is displayed to the user and the id is just internal, right?

mvdbeek: That's right.

joachimwolff:
How much memory is provided on travis? I am surprised it runs out of memory for `hicTransform` test case. 

`MemoryError: Unable to allocate array with shape (33754, 33754) and data type int64` Doing the maths tells me these should be ~67 GB. I understand that travis does not provide that much memory, at the same time this test case passes on my computer with 16 GB memory and planemo 0.70 without any problems. Running it locally I cannot see any increment in memory usage during the whole planemo run.

joachimwolff:
@bgruening ping.

joachimwolff:
Please wait with merging until the bugfixed version 3.4.1 is available via bioconda.

PR #2844, opened 2020-01-31
Fix lofreq viterbi, UI of viterbi and filter

wm75:
* failure of lofreq viterbi now result in failure of the tool run
* fixed missing falsevalue for --keepflags option
* moved dangerous lofreq viterbi params into advanced section and added
  help
* force values between 0 and 1 for AF filter settings in lofreq filter

FOR CONTRIBUTOR:


wm75:
Next step would be updating the tools to use lofreq 2.1.4 released earlier this month.
For now, we have a solid 2.1.3 version to build on.

tools/lofreq/lofreq_viterbi.xml
```diff
@@ -8,13 +8,41 @@
...
     <command detect_errors="exit_code"><![CDATA[
         @PREPARE_REF@
-        lofreq viterbi --ref '$reference_fasta_fn' --out - $keepflags --defqual $defqual '$reads' | samtools sort - -O BAM -o '$realigned'
+        lofreq viterbi
+        --ref '$reference_fasta_fn'
+        ${adv_options.delflags}
+        --defqual ${adv_options.bq2_handling.defqual}
+        --out tmp.bam '$reads' &&
+
+        samtools sort -O BAM -o '$realigned' tmp.bam
```
mvdbeek:
```suggestion
        samtools sort -T "\${TMPDIR:-.}" -O BAM -o '$realigned' tmp.bam
```
otherwise sort creates intermediate files in `'$realigned.tmp.##########.bam` ... which might be a problem if not using working dir isolation and the samtools job dying for some reason.

wm75: I think sort would generate the prefix based on the input, not the output file, which would be less of a problem. Anyway, it's a good idea to make use of $TMPDIR if set. Thanks!

PR #2858, opened 2020-02-09
Scanpy and Anndata minor fixes to labels

mtekman:
FOR CONTRIBUTOR:

Mirrors fixes to the PBMC 3K tutorial here: https://github.com/galaxyproject/training-material/pull/1751

### DIFF
--- a/tools/scanpy/cluster_reduce_dimension.xml
+++ b/tools/scanpy/cluster_reduce_dimension.xml
@@ -1,5 +1,5 @@
-<tool id="scanpy_cluster_reduce_dimension" name="Cluster," version="@galaxy_version@" profile="@profile@">
-    <description>infer trajectories and embed with scanpy</description>
+<tool id="scanpy_cluster_reduce_dimension" name="Cluster, infer trajectories and embed" version="@galaxy_version@" profile="@profile@">
+    <description>with scanpy</description>
     <macros>
         <import>macros.xml</import>
         <xml name="pca_inputs">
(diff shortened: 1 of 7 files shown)


bgruening:
Can you please also fix the linting of the failing URL?

tools/scanpy/cluster_reduce_dimension.xml
```diff
@@ -1,5 +1,5 @@
-<tool id="scanpy_cluster_reduce_dimension" name="Cluster," version="@galaxy_version@" profile="@profile@">
-    <description>infer trajectories and embed with scanpy</description>
+<tool id="scanpy_cluster_reduce_dimension" name="Cluster, infer trajectories and embed" version="@galaxy_version@" profile="@profile@">
```
wm75: I guess a change of the tool's name warrants a wrapper version bump?

bgruening: I don't think so. The user will either way only see one tool and get the new name in both cases.

mtekman:
thanks guys, I didn't bump because it didn't look like a feature upgrade

tools/scanpy/cluster_reduce_dimension.xml
wm75: That's true for any given server. On *different* servers, however, you can now have the same tool version under two different names, which may leave users wondering whether these are equivalent.

Anyway, this is not particularly critical though :)

PR #2861, opened 2020-02-10
add two more chunks

bgruening:
@mvdbeek I see timounts with these new tests, maybe this helps?


mvdbeek:
That wouldn't help, it's only a single tool that times out (probably when downloading stuff from somewhere). The last few runs worked ok though: https://github.com/galaxyproject/tools-iuc/actions

nsoranzo:
Outdated, we are doing this dynamically now.

PR #2868, opened 2020-02-12
Fix ScanPy default stdout

mtekman:
FOR CONTRIBUTOR:

The motivation here is to reduce the training time of the *PBMC 3K with ScanPy* training material, where there is a large time bottleneck in performing inspection on the generated H5AD sets. 

Instead of asking users to query their H5AD sets for general information, this PR aims to just print this nicely within the stdout window within the H5AD dataset in the history. The stdout at the moment contains debugging information, so this will be redirected to a seperate hidden output (so it can be still tested against, but not visible to the user).

### DIFF
--- a/tools/scanpy/macros.xml
+++ b/tools/scanpy/macros.xml
@@ -22,13 +22,16 @@
     <token name="@CMD@"><![CDATA[
 cp '$adata' 'anndata.h5ad' &&
 cat '$script_file' > '$hidden_output' &&
-python '$script_file' >> '$hidden_output' 2>&1 &&
+python '$script_file' >> '$hidden_output' &&
 ls . >> '$hidden_output' &&
 touch 'anndata_info.txt' &&
 cat 'anndata_info.txt' @CMD_prettify_stdout@
     ]]>
     </token>
     <token name="@CMD_imports@"><![CDATA[
+import sys
+sys.stderr = open('$hidden_output', 'a')
+
 import scanpy as sc
 import pandas as pd
 import numpy as np
--- a/tools/scanpy/macros.xml
+++ b/tools/scanpy/macros.xml
@@ -24,6 +24,7 @@ cp '$adata' 'anndata.h5ad' &&
 cat '$script_file' > '$hidden_output' &&
 python '$script_file' >> '$hidden_output' 2>&1 &&
 ls . >> '$hidden_output' &&
+touch 'anndata_info.txt' &&
 cat 'anndata_info.txt' @CMD_prettify_stdout@
     ]]>
     </token>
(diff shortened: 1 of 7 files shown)


tools/scanpy/macros.xml
```diff
@@ -21,9 +21,10 @@
     </xml>
     <token name="@CMD@"><![CDATA[
 cp '$adata' 'anndata.h5ad' &&
-cat '$script_file' &&
-python '$script_file' &&
-ls .
+cat '$script_file' > '$hidden_output' &&
+python '$script_file' >> '$hidden_output' 2>&1 &&
```
mvdbeek: Redirecting the stderr as you do here will lead to a tool will be impossible to debug.Do you really need to redirect stderr as well ?

mtekman: sadly yes, but I will try to find a way to do that within python

mtekman: @mvdbeek A bit cleaner now https://github.com/galaxyproject/tools-iuc/pull/2868/commits/2ad7f8783b1ecfc35998260d2848b696a025af0d

bgruening:
@mtekman it is failing in Docker. If you run your tests with `planemo t --docker` they will fail. Without Docker it works.

bgruening:
How much time do you spend here? Is the debug output that bad? Can you not add the information that you need ontop of the current output?

mtekman:
Ah I forgot docker, thanks I will try with that. The debug output throws future depreciation function output taking up 6 lines in the stdout window. That stdout window could be better put to use giving actual h5ad metrics instead. It would also align it with the RaceId tos which give info about the binary RDS object.

mtekman:
Found the problem. Fixing now

mtekman:
@mtekman Future Reference:

* Test against the docker image if you can't reproduce Travis bugs
* Don't redirect stderr somewhere, otherwise its impossible to debug (as mentioned)
* Isolate just one test case and incrementally simplify the CMD until bug is found

In this case the problem was that Python in the Docker image had trouble writing out to `utf-8`, and was set to `ascii` by default. I'm guessing on my system this is different.

tools/scanpy/macros.xml
```diff
@@ -41,11 +46,21 @@ adata = sc.read('anndata.h5ad')
     </token>
     <xml name="anndata_outputs">
         <data name="anndata_out" format="h5ad" from_work_dir="anndata.h5ad" label="${tool.name} (${method.method}) on ${on_string}: Annotated data matrix"/>
+        <data name="hidden_output" format="txt" label="Log file" hidden="true" >
+            <!-- The purpose of this is to consolidate all unnecessary logging information into a hidden
+                 output file that:
```
bgruening: Why do we not make this output a normal output but optional as we do usually.

mtekman: because the debug info is only useful to the developer and the user will never want to see it IMO

mtekman: but it will mess with the history numbering....
Hmm.... okay give me an hour or so to patch it your way ;-)

bgruening:
Sorry, for the delay ....

PR #2879, opened 2020-02-17
Update to FastANI version and small bugfix

DarianHole:
FOR CONTRIBUTOR:

I was asked to push to this repo. Initial galaxy wrapper found here: https://github.com/quadram-institute-bioscience/galaxy-tools/tree/master/tools/fastani


tools/fastani/.shed.yml
```diff
@@ -0,0 +1,12 @@
+name: fastani
+owner: thanhlv
```
Takadonet: Change owner to 'iuc'

DarianHole: Done

tools/fastani/fastani.xml
```diff
@@ -0,0 +1,102 @@
...
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements" />
+    <version_command>fastANI --version</version_command>
+    <command detect_errors="exit_code">
+    <![CDATA[
+    ### SET UP INPUTS ###
+    #for $input_q in $query:
+        ln -fs '$input_q' ${input_q.element_identifier}_query &&
```
Takadonet: single quotes around **${input_q.element_identifier}_query** and  **${input_r.element_identifier}_ref** .Sometime identifier may have weird characters in them not  friendly to cmd line.

DarianHole: Done!

tools/fastani/fastani.xml
```diff
@@ -0,0 +1,102 @@
...
+    <description> fast alignment-free computation of whole-genome Average Nucleotide Identity</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements" />
+    <version_command>fastANI --version</version_command>
+    <command detect_errors="exit_code">
+    <![CDATA[
+    ### SET UP INPUTS ###
+    #for $input_q in $query:
```
bgruening: do we need to use element_identifier here or can we simply use 1,2,3,4,5 ... etc ... if the latter is possible we can use enumerate($query)

bgruening: background is that we need to avoid element_identifier ... if we need to use it, we need to strip all strange chars away like we do in other tools

DarianHole: I think you are correct. Give me a second to try it out

tools/fastani/fastani.xml
```diff
@@ -0,0 +1,102 @@
+<tool id="fastani" name="FastANI" version="1.2">
```
bgruening: This version could be a macro share with the requirement version ... this way both versions do not diverge (as it is currently.)

DarianHole: Woops I missed that. Should be good now!

tools/fastani/fastani.xml
DarianHole: Doing it that way removes the file names from the output so I'll have to strip the characters I suppose

tools/fastani/macros.xml
```diff
@@ -0,0 +1,21 @@
+<macros>
+    <xml name="requirements">
+        <requirements>
+        <requirement type="package" version="1.3">fastani</requirement>
+            <yield/>
+        </requirements>
+    </xml>
+    <xml name="citations">
+        <citations>
+            <citation type="bibtex">@ARTICLE{a1,
```
bgruening: here it would be better to use the DOI

tools/fastani/fastani.xml
bgruening: e.g. see here: https://github.com/galaxyproject/tools-iuc/blob/02413e02c9484cdb604bda2c09931599f9259792/tools/snippy/snippy.xml#L17

there are also examples with sed in the code base

DarianHole: Ok I'll add that change in with the other ones and make sure it works

tools/fastani/fastani.xml
DarianHole: Should be good

tools/fastani/macros.xml
DarianHole: Done. That's a lot easier than how I had done it.

tools/fastani/fastani.xml
DarianHole: Let me know if the change looks good here.

thanhleviet:
> @DarianHole great work!
> 
> @thanhleviet do you want to look at this PR?

Looks good, thanks for making it neater :)

PR #2881, opened 2020-02-18
Upgrade HyPhy tools to 2.5.4

davebx:
FOR CONTRIBUTOR:


tools/hyphy/macros.xml
```diff
@@ -68,11 +76,27 @@
     </xml>
     <xml name="citations">
         <citations>
-            <citation type="doi">10.1093/bioinformatics/bti079</citation>
+         <citation type="bibtex">
```
bgruening: why are we not using the DOI here as well?

davebx: From @spond on slack:

> If you pull the doi citation, there are some weird character issues
> Hyphens don't render correctly
> Which is why I put the explicit citation in place

mvdbeek: We fixed that very recently in 20.01, can you check it out again? doi is much preferable

bgruening:
> NameMapper.NotFound: cannot find 'mcmc'

One error that I say.

PR #2888, opened 2020-02-22
[WIP] Raceid Fixes and Version Update

mtekman:
FOR CONTRIBUTOR:

* [X] Fix User report
     The mkgenelist function fails and the log does not catch it because it writes to a different file. Surrounded inside of main loop with a tryCatch so that the file is still written, but the contents might not be if an error is thrown.
* [ ] Update to newer version
     * [ ] Include UMAP
     * [ ] Include VarID
     * [ ] Other features

### DIFF
--- a/tools/raceid/scripts/cluster.R
+++ b/tools/raceid/scripts/cluster.R
@@ -1,5 +1,5 @@
 #!/usr/bin/env R
-VERSION = "0.5"
+VERSION = "0.6"
 
 args = commandArgs(trailingOnly = T)
 
--- a/tools/raceid/scripts/cluster.R
+++ b/tools/raceid/scripts/cluster.R
@@ -1,9 +1,9 @@
 #!/usr/bin/env R
-VERSION = "0.6"
+VERSION <- "0.6" # nolint
 
-args = commandArgs(trailingOnly = T)
+args <- commandArgs(trailingOnly = T)
 
-if (length(args) != 1){
+if (length(args) != 1) {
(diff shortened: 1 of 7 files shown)


mtekman:
I'm seeing some really funny behaviour:

> `conda create -n test raceid=0.1.1 bioconductor-scran=1.6.3`

This fails, when it used to work fine....
If I use the latest `bioconductor-scran=1.14.1` some really strange behaviour occurs (it looks like it tries to parse all packages individually or something).

TODO: I will need to play around with this more. At the moment with the current requirements, everything segfauls on some mutex error.

tools/raceid/scripts/cluster.R
```diff
@@ -13,31 +13,34 @@ suppressWarnings(suppressPackageStartupMessages(require(scran)))
 source(args[1])
 
 
-do.filter <- function(sc){
-    if (!is.null(filt.lbatch.regexes)){
-        lar <- filt.lbatch.regexes
+do.filter <- function(sc) { # nolint
+    if (!is.null(filt.lbatch.regexes)) {
+        lar <- filt.lbatch.regexes # nolint
```
bgruening, line 18: why is this a nolint and many others here?

mtekman: (Top) do.filter is not snake_case
(Bottom) filt.lbatch.regex not snake_case, or exceeds 30 characters, or other style errors

Also lintr cannot see the sourced params, so it thinks many variables are undeclared

PR #2892, opened 2020-02-24
fix binascii.hexlify argument

meugr:
fix error TypeError: argument should be bytes, buffer or ASCII string, not 'int'

FOR CONTRIBUTOR:

### DIFF
--- a/tool_collections/galaxy_sequence_utils/fastq_manipulation/fastq_manipulation.xml
+++ b/tool_collections/galaxy_sequence_utils/fastq_manipulation/fastq_manipulation.xml
@@ -40,7 +40,7 @@ def match_read(fastq_read):
         #else:
             #continue
         #end if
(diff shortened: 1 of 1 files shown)


meugr:
Tests say, what 
```
if not re.search(binascii.unhexlify('b'2e2a'').decode(), search_target):
SyntaxError: invalid syntax
```
but xml in local run galaxy only substitutes raw value in python code.
I think, tests wrong.

mvdbeek:
No, the tests run on python 3, and you're probably using python 2. I'll have a look at maintaining python 2 compatibility, but I would suggest updating your Galaxy setup to python 3.

meugr:
This bug available in online version galaxy, if use tool in workflow(as i was told). I fix quotes in my local machine - and it's fix this bug.

PR #2893, opened 2020-02-24
Bump sra toolkit version

mvdbeek:
There are occasional connection issues when downloading datasets,
so this an easy first try.

FOR CONTRIBUTOR:


tools/sra-tools/fasterq_dump.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="fasterq_dump" name="Faster Download and Extract Reads in FASTQ" version="@VERSION@.4" profile="16.01">
+<tool id="fasterq_dump" name="Faster Download and Extract Reads in FASTQ" version="@VERSION@.4+galaxy1" profile="16.01">
```
wm75: I guess the `.4` should be dropped here, and in `fastq_dump`and `sam_dump`?

mvdbeek: No, then the tool wouldn't be newer in the tool panel.

wm75: Damn! But it won't be either way since 2.10.3.4 < 2.10.4 ?

mvdbeek: Damn, you're right.

mvdbeek: I guess we'll need to keep version at 2.10.4 and add +galaxy1

wm75: Yeah, this is really unfortunate, but I can't think of another way for now either. Lets hope that the package will arrive at 2.10.4 soon :)

mvdbeek:
This is now stuck on getting sra-tools bioconda recipe updated (https://github.com/bioconda/bioconda-recipes/pull/20503)

tools/sra-tools/sra_macros.xml
```diff
@@ -1,8 +1,11 @@
 <macros>
-    <token name="@VERSION@">2.10</token>
+    <token name="@VERSION@">2.10.4+galaxy1</token>
```
nsoranzo:
```suggestion
    <token name="@VERSION@">2.10.4</token>
```

tools/sra-tools/sam_dump.xml
```diff
@@ -1,9 +1,11 @@
-<tool id="sam_dump" name="Download and Extract Reads in BAM" version="@VERSION@.4">
+<tool id="sam_dump" name="Download and Extract Reads in BAM" version="@VERSION@">
```
nsoranzo:
```suggestion
<tool id="sam_dump" name="Download and Extract Reads in BAM" version="@VERSION@+galaxy1">
```

Issue #2906, opened 2020-03-02
allow for pre-test-shell-script


bernt-matthias:
Use case: A tool where test data can be generated (reproducibly) or downloaded from somewhere (versioned)
Advantage(s): 
- test data does not need to be stored in the repo or at all
- I often thought that an non `--update_test_data` way to generate/update test-data might be better
Implementation: If a shell script with a given name exists then execute it prior to tool testing

PR #2908, opened 2020-03-03
create test-data with shell script

bernt-matthias:
Note: the dada related commit is not supposed to be in this PR in the end -- it is only here for demonstration purposes. 

Implements https://github.com/galaxyproject/tools-iuc/issues/2906: 

If there is a file `test-data.sh` in the tool dir then execute it prior to testing. Idea is that the script generates the test outputs and/or inputs, e.g. by downloading it or by calling the corresponding programs.

Goals: 

- allow to reduce the amount of test data stored in this repo by approx 50% - 100%
- have test data created independent of planemo (`--update_test_data`) which is kind of circular anyway -- but one might also call `planemo t --update_test_data` in the script (but I would not encourage this) 

There are already some tools where this could be used (at least parts this list may be considered as a TODO list): 

```
./tools/multiqc/generate_test_data.sh
./tools/pygenometracks/test-data/testpyGT.sh
./tools/metaphlan2/generate_test_data.sh
./tools/dada2/test-data/gentest.sh
./tools/qiime/qiime_core/generate_test_data.sh
./tools/qiime/qiime_add_on/generate_test_data.sh
./tools/gemini/test-data/util/build-gemini-testdata.sh
./tools/bracken/test-data/reproduce_test_data.sh
./tools/kma/test-data/generate_test_data.sh
./tools/stacks2/test-data/gentest.sh
```

For discussion:

- would it be better to `source test-data.sh`

Disadvantage: 

- adds computation time to the test

Alternatives: 

- implement directly in `planemo test`

FOR CONTRIBUTOR:


mvdbeek:
It's a nice idea, but this is going to break if any one of the urls go away or if the script breaks (for numerous reasons, shell syntax changes, requirement of requirement changes ...) and I don't like this execution of a random script approach.
I'd much rather see a URI/url approach with a checksum ... if we want to preserve these URIs we can archive a copy somewhere by just following the url. Have a look at https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/verify/test_data.py#L59, this shouldn't be too hard to implement.

bernt-matthias:
> It's a nice idea,

:) I understand your concerns, but I guess I need a bit more discussion / details.

> but this is going to break if any one of the urls go away

That's why the URLs need to be stable. 

> or if the script breaks (for numerous reasons, shell syntax changes, requirement of requirement changes ...)

Same holds for the tool_script itself. I think the chance of such problems could be reduced by adding bash and all tools used in the script to the conda environment that is used for the execution. Appart from this the requirements used in the script should be the same as thos in the tools.

> and I don't like this execution of a random script approach.

What do you mean with _random_? 

In the end the test data generation would need to undergo a review as usual for the tools and test data.

> I'd much rather see a URI/url approach with a checksum ... if we want to preserve these URIs we can archive a copy somewhere by just following the url. Have a look at https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/verify/test_data.py#L59, this shouldn't be too hard to implement.

Indeed, this might be a better solution. But I need to understand better what it going on there. When is this invoked and how. Is there an example using these resolvers?

mvdbeek:
> _random_?

random as in you can write whatever into this script. Difficult to control what gets executed there.

> That's why the URLs need to be stable.

Which is much easier to achieve programmatically if the URL is declared in the test instead of referencing a test data file that will be generated by executing the script.



> Indeed, this might be a better solution. But I need to understand better what it going on there. When is this invoked and how. Is there an example using these resolvers?

Whenever the list of test data resolvers is built, which get used whenever Galaxy is asked for the path to test data. You can trace this along the code, but the most important usage is in verify.py (via https://github.com/mvdbeek/galaxy/blob/f96c725646b24d97498c8c7c37b5486953286471/lib/galaxy/tool_util/verify/__init__.py#L27)

PR #2931, opened 2020-03-23
change delimiter

beatrizserrano:
This change avoids conflicts with plate names that already contain `_`

FOR CONTRIBUTOR:

### DIFF
--- a/tools/idr_download/idr_download_by_ids.py
+++ b/tools/idr_download/idr_download_by_ids.py
@@ -189,7 +189,7 @@ def download_image_data(
                     )
 
             # download and save the region as TIFF
-            fname = '#'.join(
+            fname = '__'.join(
                 [image_name, str(image_id)] + [str(x) for x in tile]
             )
             download_plane_as_tiff(image, tile, z_stack, channel_index, frame, fname)
--- a/tools/idr_download/idr_download_by_ids.py
+++ b/tools/idr_download/idr_download_by_ids.py
@@ -189,7 +189,7 @@ def download_image_data(
                     )
 
             # download and save the region as TIFF
-            fname = '_'.join(
+            fname = '#'.join(
                 [image_name, str(image_id)] + [str(x) for x in tile]
             )
             download_plane_as_tiff(image, tile, z_stack, channel_index, frame, fname)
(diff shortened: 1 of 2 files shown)


bernt-matthias:
Found it https://github.com/beatrizserrano/tools-iuc/blob/8d6996e34f1aa51195b67fe8288649ae952d7fb7/tools/idr_download/idr_download_by_ids.xml#L93

bernt-matthias:
For the failing test you would need to change the collection element identifier to `Centrin_PCNT_Cep215_20110506_Fri-1545_0_SIR_PRJ#1884807#3#3#5#5`

Nevertheless another delimiter would be good in my opinion. What about underscore?

beatrizserrano:
> For the failing test you would need to change the collection element identifier to `Centrin_PCNT_Cep215_20110506_Fri-1545_0_SIR_PRJ#1884807#3#3#5#5`

yes, thanks!

> Nevertheless another delimiter would be good in my opinion. What about underscore?

Underscore would be better, yes, that was the one used before. The problem is that in the strings that are being joined, I can find all kind of characters already... like `_`, `-`, `--`, ` `, ... any other suggestion that avoids those?

bernt-matthias:
How about removing `_` from the plate names instead?

wm75:
> How about removing `_` from the plate names instead?

Yes, either this, or you could use double underscores `__` as the element separator.

beatrizserrano:
I'm thinking now to directly remove the plate name... was there a reason to keep it that I don't remember @wm75?

wm75:
Well, the query gives us only a name and an ID. I don't know if there is a reliable pattern to detect the plate name part within the image name so probably this is all or nothing - remove the whole image name or keep it?

wm75:
What @bernt-matthias suggests is simply:
```
            fname = '_'.join(
                [image_name.replace('_', ''), str(image_id)] + [str(x) for x in tile]
            )
```

The alternative is:
```
            fname = '__'.join(
                [image_name, str(image_id)] + [str(x) for x in tile]
            )
```

Both require also a corresponding test adjustment.

beatrizserrano:
Thank you @bernt-matthias and @wm75! I think it makes sense to keep the plate name as it is. That would help users that need to visually inspect images when they want to find them in the IDR. I've checked all the published datasets, and the double underscore doesn't seem to be used to name the plates so far, so I've gone for the second option. I've changed the tests accordingly, hope I didn't miss anything :)

bernt-matthias:
Thanks a lot @beatrizserrano

PR #2939, opened 2020-03-28
GitHub actions for push and PR

bernt-matthias:
Did this for the galaxyproteomics repo and guess this could be a nice addition over here. 

Mostly based on the ci workflow that is already running here. Added `planemo shed_lint` and pycodestyle. 

~Currently tests are not containerized .. did not work in the other repo. Would be nice if we get ths running.~ Could also be splitted into multiple workflows, maybe also the caching job is not necessary here.

Maybe the very last step here could be nice for the ci workflow, i.e. the check if the merged test outputs indicate an error.

Open for more suggestions. I was thinking in particular about how to avoid code duplication between the different tasks and also the different workflows. Maybe we could provide our own actions?  

FOR CONTRIBUTOR:


.github/workflows/pr.yaml
```diff
@@ -0,0 +1,247 @@
...
+    # TODO --no_conda_auto_init
+    - name: fake a planemo run to update cache
+      run: 'touch tool.xml && PIP_QUIET=1 planemo test --galaxy_python_version ${{ matrix.python-version }} --galaxy_source $GALAXY_REPO --galaxy_branch $GALAXY_RELEASE'
+    - name: Cache .cache/pip
+      uses: actions/cache@v1
+      id: cache-pip
+      with:
+        path: ~/.cache/pip
+        key: pip_cache_${{ matrix.python-version }}_${{ env.GALAXY_RELEASE }}
+     # TODO Cache .planemo?
```
mvdbeek: That exceeds the cache size.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,247 @@
...
+    - name: fake a planemo run to update cache
+      run: 'touch tool.xml && PIP_QUIET=1 planemo test --galaxy_python_version ${{ matrix.python-version }} --galaxy_source $GALAXY_REPO --galaxy_branch $GALAXY_RELEASE'
+    - name: Cache .cache/pip
+      uses: actions/cache@v1
+      id: cache-pip
+      with:
+        path: ~/.cache/pip
+        key: pip_cache_${{ matrix.python-version }}_${{ env.GALAXY_RELEASE }}
+     # TODO Cache .planemo? 
+     # TODO generate and cache changed_repositories.list?
```
mvdbeek: 👍

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,247 @@
...
+    strategy:
+      matrix:
+        python-version: [3.7]
+    steps:
+    - uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+        architecture: 'x64' # (x64 or x86)
+    - name: setup planemo
+      run: 'git clone --recursive --depth 1 https://github.com/galaxyproject/planemo && python -m pip install --pre planemo/ && rm -rf planemo'
```
mvdbeek: You can just pip install planemo now, this was a temporary workaround.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,247 @@
...
+    - name: set tool list
+      id: tools
+      run: echo "::set-output name=tool_list::$(tr '\n' ' ' < changed_repositories.list)"
+    - name: show repo list
+      run: 'cat changed_repositories.list'
+    - name: show changed tools/repositories chunk list
+      run: 'cat changed_tools_chunk.list changed_repositories_chunk.list'
+    # TODO without this planemo task aborts with `System.IO.IOException: No space left on device` 
+    - name: free disk space
+      run: |
```
mvdbeek: `docker system prune -a -f` seems like a better idea. If necessary you can also `rm -Rf /usr/share/anaconda`. It's odd though that the weekly testing doesn't require this.

bernt-matthias: Maybe it is because I still run my test-data.sh script in the galaxyproteomics repo which creates a lot of data... but I'm not sure. Seems not to hurt.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,247 @@
...
+        sudo swapoff -a
+        sudo rm -f /swapfile
+        sudo apt clean
+    # TODO --biocontainers --no_conda_auto_init
```
mvdbeek: We're doing `--biocontainers --no_conda_auto_init` in .travis.yml, was there any reason to not do it here ?

bernt-matthias: I just temporarily disabled it because it failed for some other reason (probably the disk space issue). Seems to work now. Will add it again.

.github/workflows/pr.yaml
bernt-matthias: lets see if it works without

.github/workflows/pr.yaml
bernt-matthias: Not sure anymore if this is a good idea .. apart from the fact that I could not get it working.

Also wonder why caching .cache works in the CI workflow since according to the docs the cache in only accessibe for workflows triggered by push/ pull https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache

Evenmore since the cache is removed after 7days. https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy

Also according to the last link the cache can be up to 5GB, so maybe `.planemo` would fit?

mvdbeek: > Not sure anymore if caching `changed_repositories.list` is a good idea .. apart from the fact that I could not get it working.

You probably need to publish this as an artefact (like we do for the test results)

> Also wonder why caching .cache works in the CI workflow

It doesn't work anymore, if you look at the last run in https://github.com/galaxyproject/tools-iuc/runs/528997353?check_suite_focus=true you see the warning when restoring the cache.

We definitely can't cache all of `.planemo`, as we can't reliably cache the virtualenv (we did that on the Galaxy repo, but this caused problems when the host python interpreter changed).

bernt-matthias: > You probably need to publish this as an artefact (like we do for the test results)

I would like to leave it as it is -- I think saving ~1min run time is not worth it. Until today I thought that this caching is for reuse between different jobs of the same workflow. After studying the docs a bit more I think that it also caches between different runs of the workflow. In this case caching of the changed repos would be wrong.

mvdbeek: I think it is worth it (surely for the weekly runs), but I can also implement this.

bernt-matthias: For the weekly runs .. maybe. But here I think it would be wrong, because the file differs for different PRs.

mvdbeek: > You probably need to publish this as an artefact (like we do for the test results)

bernt-matthias:
Cleaned the commits. We must not forget to remove the mock tool changes. These ones https://github.com/galaxyproject/tools-iuc/pull/2939/commits/e9b7b95c09e9e33deb868a9648544417524c11f9 could go to a different PR if you prefer. 

I guess with the parallel travis checks this is good to go.

mvdbeek:
Do you want to add the deployment too ?

bernt-matthias:
Sounds like a good idea. Would this be `in: deployment`?

I guess this would go to a different workflow?

mvdbeek:
For `deployment` we'd have to create tags for tool releases, we can't do that with this repo structure.

I'd put the TS upload either in a new job in the same workflow or in the `combine_outputs` step, with the same logic as in travis. We'll want to upload only on passed tests on the master branch.  The deployment job (or step) can be guarded with `if: github.ref == 'refs/heads/master'`. If you want to use a separate job you'd also have to either upload an artifact with the final job state or redo the final job state calculation.

bernt-matthias:
Added a first draft. Could you check if the commit range in ci_find_repos is useful. Also from where do I get  SHED_EMAIL ...?

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,278 @@
...
+      with:
+        path: ~/.cache/pip
+        key: pip_cache_${{ matrix.python-version }}_${{ env.GALAXY_RELEASE }}
+    - name: setup planemo
+      run: 'pip install planemo'
+    - name: planemo ci_find_repos
+      run: |
+           mkdir changed_repositories
+           planemo ci_find_repos --changed_in_commit_range origin/master... --exclude packages --exclude deprecated --output changed_repositories.list
+    - name: deploy on testtoolshed
```
mvdbeek:
```suggestion
    - name: deploy on testtoolshed
        env:
          SHED_EMAIL: ${{ secrets.shed_email }}
          SHED_PASSWORD: ${{ secrets.shed_password }}
```
This is described in https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets.

bernt-matthias:
I just moved it the the job and fixed indentation? Or are the secrets different for MTS and TTS?

mvdbeek:
They're the same, but I think I prefer having them available only in the steps that need them. We may also want to use API keys, which wouldn't be the same. I'm not sure why don't do that now ...

bernt-matthias:
How can we check if deployment works? 

Also I guess travis deployment should be disabled?

mvdbeek:
Yep, can you delete `.travis.yml` ? I guess we can disable toolshed upload for a quick test and just test that testtoolshed upload works.

bernt-matthias:
OK. Have cleaned the commits a bit. The pycodestyle fixes should be enough to trigger a deployment (version bump should not be necessary)

Did you check the commit range in the deployment step `planemo ci_find_repos --changed_in_commit_range origin/master... ` I'm unsure.

mvdbeek:
Probably need to get those commits via something like https://github.com/actions/runner/issues/342#issuecomment-590670059

mvdbeek:
Although `--changed_in_commit_range origin/master...` seems to work, doesn't it ?

bernt-matthias:
In the other jobs it seems to work. Question that I had in mind is if the workflow runs before / after the merge.

bernt-matthias:
If before it should be fine

.github/workflows/pr.yaml
nsoranzo: That could also help with the verbosity of GH workflows!

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,281 @@
...
+        python-version: [3.7]
+    steps:
+    - uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+        architecture: 'x64' # (x64 or x86)
+    - name: setup planemo
+      run: 'pip install planemo'
+    - name: fake a planemo run to update cache
+      run: 'touch tool.xml && PIP_QUIET=1 planemo test --galaxy_python_version ${{ matrix.python-version }} --no_conda_auto_init --galaxy_source $GALAXY_REPO --galaxy_branch $GALAXY_RELEASE'
```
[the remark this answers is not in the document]
bernt-matthias: As far as I know this should be equivalent since this runs with `bash --noprofile --norc -eo pipefail {0}` https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,281 @@
...
+      with:
+        path: ~/.cache/pip
+        key: pip_cache_${{ matrix.python-version }}_${{ env.GALAXY_RELEASE }}
+    - name: setup planemo
+      run: 'pip install planemo'
+    - name: planemo ci_find_repos
+      run: |
+           mkdir changed_repositories
+           planemo ci_find_repos --changed_in_commit_range origin/master... --exclude packages --exclude deprecated --output changed_repositories.list
+    - name: setup pycodestyle
```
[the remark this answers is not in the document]
bernt-matthias: hah. I was looking here: https://github.com/galaxyproject/galaxy/blob/dev/.ci/autopep8.sh which uses something else. Lets switch to flake8

.github/workflows/pr.yaml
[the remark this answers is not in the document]
bernt-matthias: my bad.. this is exactly what you meant. sorry.

bernt-matthias:
Anything left to do here?

mvdbeek:
That depends, are you comfortable with testing the deployment by merging ?

I think you could try that on your own fork, this should be the same logic, and it'd abort due to lack of the secrets.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,274 @@
...
+
+  test:
+    name: test tools
+    # This job runs on Linux
+    runs-on: ubuntu-latest
+    needs: [lint,flake8]
+    strategy:
+      fail-fast: false
+      matrix:
+        chunk: [0,1,2,3]
```
nsoranzo:
```suggestion
        chunk: [0, 1, 2, 3]
```

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,274 @@
...
+    - uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install Planemo
+      run: pip install planemo
+    - name: fake a planemo run to update cache
+      run: |
+        touch tool.xml
+        PIP_QUIET=1 planemo test --galaxy_python_version ${{ matrix.python-version }} --no_conda_auto_init --galaxy_source $GALAXY_REPO --galaxy_branch $GALAXY_RELEASE
+    - name: Cache .cache/pip
```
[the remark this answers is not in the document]
bernt-matthias: Done. And I also will check if there was a cache miss. Only in this case the following two steps need to run.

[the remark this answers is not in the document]
nsoranzo: I think you may be under the impression that the cache is shared across jobs as in Travis, but this is not the case with GitHub workflows.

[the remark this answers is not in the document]
bernt-matthias: No. Each the jobs are completely independent. The cache is for sharing between different workflow runs. I guess I have the comparison the wrong way around. I meant: If there is a cache hit it won't be updated, so we don't need to call the command lines that fill `.cache/pip`.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,274 @@
...
+      uses: actions/cache@v1
+      id: cache-pip
+      with:
+        path: ~/.cache/pip
+        key: pip_cache_${{ matrix.python-version }}_${{ env.GALAXY_RELEASE }}
```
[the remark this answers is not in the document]
bernt-matthias: I'm thinking about this a bit. 

A potential problem  of caching `changed_repositories.list` etc is that if they change between different workflow runs the information in the cache is not valid anymore. If we assume that this must never happen (I remember pull requests where I included tool by tool of a tool collection into a PR) then I could change it. Certainly I do not like the current code duplication.

Caching `.planemo` is not possible according to @mvdbeek:

> We definitely can't cache all of .planemo, as we can't reliably cache the virtualenv (we did that on the Galaxy repo, but this caused problems when the host python interpreter changed).

On using artifacts to persist the results of `planemo ci_find_repos` and `planemo ci_find_tools`:
- I agree that this would make the workflow more verbose which would be nice
- For the first ci_find_repos call it should be easy
- For the calls that determine changed_repositories_chunk.list changed_tools_chunk.list we would need also a chunk matrix. I guess we don't want this...?

bernt-matthias:
> I think you could try that on your own fork, this should be the same logic, and it'd abort due to lack of the secrets.

This means that I would manually merge on my fork right?

.github/workflows/pr.yaml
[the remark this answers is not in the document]
nsoranzo: We definitely don't want to **cache** `changed_repositories.list`, this will change for each PR. But I think we should pass it on from this first job to the following ones via an artifact.

I also agree the output of `planemo ci_find_tools` should not be cached or passed on as artifact, that was my mistake.

For the `~/.planemo` directory, I think that can also be passed on as artifact, so there is no problem with Python or Galaxy versions.

.github/workflows/pr.yaml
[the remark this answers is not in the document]
nsoranzo: The fake planemo test is used to fill `~/.planemo` , not `~/.cache/pip` as far as I understand.

.github/workflows/pr.yaml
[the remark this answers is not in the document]
mvdbeek: > We definitely don't want to **cache** `changed_repositories.list`, this will change for each PR. But I think we should pass it on from this first job to the following ones via an **artifact**.

👍 

> For the `~/.planemo` directory, I think that can also be passed on as artifact, so there is no problem with Python or Galaxy versions.

I don't think we want multi-gigabyte artefacts. It might work, but I doubt this is the intended use.

mvdbeek:
Yes

.github/workflows/pr.yaml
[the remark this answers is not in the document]
bernt-matthias: Its both, the packages are installed to `~/.planemo/gx_venv...` using pip's cache in `~/.cache/pip`, hence the download happens only in this job and only in case of cache miss (otherwise they are just taken from the cache)

For testing:

```
virtualenv -p python3 .venv
. ./.venv/bin/activate
pip install planemo

rm -rf ~/.cache/pip ~/.planemo 
planemo test --galaxy_python_version 3.7 --no_conda_auto_init
```

`du -hs ~/.cache/pip/` shows me 142M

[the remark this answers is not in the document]
mvdbeek: > The fake planemo test is used to fill `~/.planemo` , not `~/.cache/pip` as far as I understand.

I did this so to startup Galaxy and install all packages. Prior to us figuring out the problems with caching the virtualenv we'd also cache `gx_venv` between jobs. So yes, now we only do this to populate `~/.cache/pip/`.

.github/workflows/pr.yaml
```diff
@@ -20,15 +20,22 @@ jobs:
...
+    - name: Install Planemo and flake8
+      run: pip install planemo flake8
     - name: fake a planemo run to update cache
-      if: steps.cache-pip.outputs.cache-hit != 'true'
+      if: steps.cache-pip.outputs.cache-hit == 'false'
       run: |
         touch tool.xml
         PIP_QUIET=1 planemo test --galaxy_python_version ${{ matrix.python-version }} --no_conda_auto_init --galaxy_source $GALAXY_REPO --galaxy_branch $GALAXY_RELEASE
 
+    - name: planemo ci_find_repos
```
mvdbeek, line 31: You need to check out the code :)

.github/workflows/pr.yaml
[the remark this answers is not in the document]
nsoranzo: There's no point in populating `~/.cache/pip/` with the Galaxy dependencies, given this cache is not passed to other jobs.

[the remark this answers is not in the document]
mvdbeek: It does for subsequent runs, but that's not obvious and admittedly not what I had in mind.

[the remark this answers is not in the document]
bernt-matthias: Still not clear on this point. If you check the output of the cache step, e.g. for the lint job, then it says:

```
Cache Size: ~141 MB (147586134 B)
/bin/tar -xz -f /home/runner/work/_temp/36792418-d766-472e-ab97-6ad58f648fba/cache.tgz -C /home/runner/.cache/pip
Cache restored from key: pip_cache_3.7_release_20.01
```

So the data is restored from the first step. This also happens in the situation of a cache miss (for testing I renamed the cache). ... but the check if it is a miss/hit seems wron at the moment.

Also still wondering if caching `.planemo` is possible, given that a change of the python interpreter (version) should result in the use of a different cache.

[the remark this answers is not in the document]
nsoranzo: > So the data is restored from the first step.

No, the data are coming from a previous run of the **lint** job!

> Also still wondering if caching `.planemo` is possible, given that a change of the python interpreter (version) should result in the use of a different cache.

That's why we add the Python version to the cache key:
```yaml
key: pip_cache_python${{ matrix.python-version }}_${{ env.GALAXY_RELEASE }}
```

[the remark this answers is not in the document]
bernt-matthias: > No, the data are coming from a previous run of the lint job!

I don't think so. I just renamed the cache: 

- the setup job had a cache miss and filled the cache 
- the lint job had a cache hit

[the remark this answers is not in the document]
mvdbeek: > There's no point in populating `~/.cache/pip/` with the Galaxy dependencies, given this cache is not passed to other jobs.

What makes you think that? The only description (https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache) I could find that concerns cache isolation says

> Access restrictions provide cache isolation and security by creating a logical boundary between different workflows and branches. For example, a cache created for the branch feature-a (with the base master) would not be accessible to a pull request for the branch feature-b (with the base master).

From this I think jobs in the same workflow might share a cache if they run sequentially.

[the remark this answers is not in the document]
bernt-matthias: > Caching is not appropriate because we want to use the latest version of a Galaxy branch.

Then we could use the commit sha. Should be possible to determine this and use it in the cache ID...?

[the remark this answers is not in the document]
nsoranzo: That's a bit of a catch 22, since the Galaxy git repo is inside `~/.planemo`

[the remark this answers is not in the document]
nsoranzo: > I don't think so. I just renamed the cache:
> 
>     * the setup job had a cache miss and filled the cache
> 
>     * the lint job had a cache hit


> From this I think subsequent jobs in the same workflow might pull a cache that has been created in a previous job. If that's the case you want to establish a full cache in the first job that caches something.

Uh, then it must be a misconception on my side, happy if that works!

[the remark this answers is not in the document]
bernt-matthias: > That's a bit of a catch 22, since the Galaxy git repo is inside ~/.planemo

True. `git ls-remote https://github.com/galaxyproject/galaxy.git refs/heads/release_20.01` would be a possibility to get the latest commit.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,329 @@
...
+        path: ~/.cache/pip
+        key: pip_cache_python_${{ matrix.python-version }}_${{ env.GALAXY_RELEASE }}
+    - name: determine latest galaxy commit
+      run: echo ::set-env name=GALAXY_HEAD_SHA::$(git ls-remote ${{ env.GALAXY_REPO }} refs/heads/${{ env.GALAXY_RELEASE }} | cut -f1)
+    - name: Cache .planemo
+      uses: actions/cache@v1
+      id: cache-planemo
+      with:
+        path: ~/.planemo
+        key: planemo_cache_py${{ matrix.python-version }}_${{ env.GALAXY_HEAD_SHA }}
```
mvdbeek, line 79:
```suggestion
```
linting doesn't use anything from ~/.planemo

.github/workflows/pr.yaml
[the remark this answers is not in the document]
bernt-matthias: Seems that caching .planemo is working now. The numbers: 

size 770MB, needs 1:15min for creation (i.e. for the post cache step), 36sec for loading the cache. Without the cache `planemo t` needs 4:00min and with the cache 1:10min.

Also passing changed_repositories.list as artifact seems functional. 

Guess its time for testing the deployment job in my branch. I will also test deploying to the TTS by setting a secret in my repo.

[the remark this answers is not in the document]
bernt-matthias: One remaining concern with caching `.planemo`. We need to make sure that the same cache is used throughout the workflow. Which is not the case if the SHA changes during the execution of the workflow. Otherwise also the containers would be stored in the cache. 

Ideas: 
- remove containers after `planemo test` 
- make the SHA persistent as artifact

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,316 @@
...
+      id: cache-pip
+      with:
+        path: ~/.cache/pip
+        key: pip_cache_python_${{ matrix.python-version }}_${{ env.GALAXY_RELEASE }}
+    - name: setup planemo
+      run: pip install planemo
+    - uses: actions/download-artifact@v1
+      with:
+        name: Changed repositories
+        path: changed_repositories.list
```
nsoranzo, line 77: Would it work with `.` (or `~`)? That seems less confusing then creating a directory with the same name as the file.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,316 @@
...
+                               --output changed_tools_chunk.list \
+                               $(cat changed_repositories.list/changed_repositories.list)
+            else
+                planemo ci_find_repos --chunk_count 4 --chunk ${{ matrix.chunk }} \
+                               --output changed_repositories_chunk.list \
+                               $(cat changed_repositories.list/changed_repositories.list)
+            fi
+        fi
+    - name: show repo list
+      run: cat changed_repositories.list/changed_repositories.list
```
nsoranzo, line 181: I'd move this step to the `setup` job, before the upload as an artifact.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,316 @@
...
+    - name: Cache .planemo
+      uses: actions/cache@v1
+      id: cache-planemo
+      with:
+        path: ~/.planemo
+        key: planemo_cache_py${{ matrix.python-version }}_${{ env.GALAXY_HEAD_SHA }}
+    - name: install Planemo and flake8
+      run: pip install planemo flake8
+    - name: fake a planemo run to update cache
+      if: steps.cache-pip.outputs.cache-hit != 'true' && steps.cache-planemo.outputs.cache-hit != 'true'
```
nsoranzo, line 34: Shouldn't this be an OR?

```suggestion
      if: steps.cache-pip.outputs.cache-hit != 'true' || steps.cache-planemo.outputs.cache-hit != 'true'
```

bernt-matthias:
First tests with merging were not successful. The first question is how to determine the commit range. `master...` does not work since it is empty. Maybe use some infos from the github context https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context

Do you know if merging a PR always triggers a 3way merge or if ff merges are possible.

I'm also wondering if we need the `on: push`. Somehow docs and infos from the internet suggest that `pull_request` might be sufficient. Need to construct a PR against my master for testing...

.github/workflows/pr.yaml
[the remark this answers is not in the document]
nsoranzo: > make the SHA persistent as artifact

This seems a good solution to me.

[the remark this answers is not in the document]
mvdbeek: Are we talking about docker containers? these are stored through the docker daemon, usually in /var/lib/docker, they are definitely not stored in ~/. planemo

.github/workflows/pr.yaml
bernt-matthias: Upload of single files will come with v2 of this action. So long I will put the SHA and this file in a single artifact, or?

bernt-matthias:
Still no solution for the selection of the commit range. 

- I guess I need to determine the range in two steps checking for `if: github.ref == 'refs/heads/master'
`, The current scheme fails for merge commits to the feature branch.
- Seems that checkout depth of 1 was insufficient with `checkout@v2` which also only gets the feature branch by default?? 
- with `master..` and `master...` I often see new commits on master .. is there a better way?

bernt-matthias:
Heureka: https://testtoolshed.g2.bx.psu.edu/view/matthias/data_manager_bowtie2_index_builder/68fd790f3916

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,327 @@
...
+        PIP_QUIET=1 planemo test --galaxy_python_version ${{ matrix.python-version }} --no_conda_auto_init --galaxy_source $GALAXY_REPO --galaxy_branch $GALAXY_RELEASE
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: get all branches
+      run: |
+        git fetch --no-tags --prune --depth=1 origin +refs/heads/*:refs/remotes/origin/*
+    - name: Determine commit range for feature branch
+      if: github.ref != 'refs/heads/master'
+      run: echo ::set-env name=COMMIT_RANGE::"origin/master.."
```
nsoranzo, line 50: I think this should be (as in `.travis.yml`):

```suggestion
      run: echo ::set-env name=COMMIT_RANGE::"origin/master..."
```

Otherwise the commit range may contain changes from `origin/master` that are not in the feature branch.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,327 @@
...
+      with:
+        name: Changed repositories
+        path: changed_repositories.list
+    - name: Planemo lint
+      run: |
+           set -e
+           while read -r DIR
+           do
+               planemo shed_lint --tools --ensure_metadata --urls --report_level warn --fail_level error --recursive "$DIR";
+           done < changed_repositories.list/changed_repositories.list
```
nsoranzo, line 101:
```suggestion
      run: |
        set -e
        while read -r DIR; do
            planemo shed_lint --tools --ensure_metadata --urls --report_level warn --fail_level error --recursive "$DIR";
        done < changed_repositories.list/changed_repositories.list
```

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,327 @@
...
+                       TESTPATH=$(planemo ci_find_tools "$DIR")
+                   else
+                       TESTPATH="$DIR"
+                   fi
+                   PIP_QUIET=1 planemo test --database_connection postgresql://postgres:postgres@localhost:5432/galaxy --biocontainers --no_conda_auto_init --galaxy_source $GALAXY_REPO --galaxy_branch $GALAXY_RELEASE --galaxy_python_version ${{ matrix.python-version }} --test_output_json "$DIR"/tool_test_output.json "$TESTPATH" || true
+                   docker system prune --all --force --volumes || true
+               done < changed_repositories_chunk.list
+           else
+               echo '{"tests":[]}' > tool_test_output.json
+           fi
```
nsoranzo, line 216:
```suggestion
      run: |
        if [ -s changed_tools_chunk.list ]; then
            PIP_QUIET=1 planemo test --database_connection postgresql://postgres:postgres@localhost:5432/galaxy --biocontainers --no_conda_auto_init --galaxy_source $GALAXY_REPO --galaxy_branch $GALAXY_RELEASE --galaxy_python_version ${{ matrix.python-version }} --test_output_json tool_test_output.json $(cat changed_tools_chunk.list) || true
            docker system prune --all --force --volumes || true
        elif [ -s changed_repositories_chunk.list ]; then
            while read -r DIR; do
                if [[ "$DIR" =~ ^data_managers.* ]]; then
                    TESTPATH=$(planemo ci_find_tools "$DIR")
                else
                    TESTPATH="$DIR"
                fi
                PIP_QUIET=1 planemo test --database_connection postgresql://postgres:postgres@localhost:5432/galaxy --biocontainers --no_conda_auto_init --galaxy_source $GALAXY_REPO --galaxy_branch $GALAXY_RELEASE --galaxy_python_version ${{ matrix.python-version }} --test_output_json "$DIR"/tool_test_output.json "$TESTPATH" || true
                docker system prune --all --force --volumes || true
            done < changed_repositories_chunk.list
        else
            echo '{"tests":[]}' > tool_test_output.json
        fi
```

.github/workflows/pr.yaml
bernt-matthias: I think its the other way around: 

>  triple-dot syntax, which specifies all the commits that are reachable by either of two references but not by both of them

https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection

But I was also wondering, because in the travis config the tripple dot was used

bernt-matthias: The odd thing is that with `git log origin/master.. --oneline` I see exactly the commits that I want to (while `git log origin/master... --oneline ` I also see new commits from the master branch), but in `planemo ci_find_repos` `origin/master..` gives me nothing in some cases .. don't get it yet.

bernt-matthias: Got it: I was reading the wrong docs

> However, "diff" is about comparing two endpoints, not ranges, and the range notations

nsoranzo:
@bernt-matthias The last commit 234c6f9 is undoing the changes in the previous 2 commits by mistake.

bernt-matthias:
> @bernt-matthias The last commit [234c6f9](https://github.com/galaxyproject/tools-iuc/commit/234c6f96961011f18e8c5a44e0593c6573e040a1) is undoing the changes in the previous 2 commits by mistake.

Ups. Thanks for noting. All this merging tests are confusing my gitting. 

Will try now if I can get the ranges correct for the 3 cases of merging 
- Create a merge commit
- Squash and merge 
- Rebase and merge

The first was alŕeady working. I guess the later two create a fast forward merge. Squash and merge should be easy. Bit worried about the last one. 

Any good idea to determine the diff commit range easily would be helpful.

bernt-matthias:
The last one is tricky. Since it does not create a merge commit -- which is odd since if you click `view command line instructions` in the github UI it shows `git merge --no-ff ` which should create a merge commit. 

At the moment I have no idea how to determine the diff range for this case.

bernt-matthias:
> Probably need to get those commits via something like actions/runner#342 (comment)

Damn missed this one. That seems to solve it.

nsoranzo:
@bernt-matthias I was also experimenting a bit on https://github.com/nsoranzo/tools-iuc/tree/topic/github-actions , do you want me to merge my changes here?

bernt-matthias:
Just wanted to write to you :) Nearly done (hopefully): 

I will merge the changes in the master of my fork. Then it would be nice if someone could test this a bit over there, maybe even using the different merge possibilities.

@mvdbeek @nsoranzo I added you as collaborators, so you can merge on your own.

Sometimes I have seen that opening a PR triggers the same action twice. I guess this is because we have this PR. Therefore I will close this PR for the moment. We can reopen after testing.

bernt-matthias:
Has been merged. Would be nice to test the full workflow of branching / creating a PR / pushing more commits / merging at least one.

nsoranzo:
I've added my changes to your branch (so here too), if you want to take a look.

bernt-matthias:
Those workflow runs are odd: they are created when I create a new branch, just before I open the PR: 

```
event: push
sha: 7765bb704deda6fe9d92e15f476596232a4b3bec
ref: refs/heads/topic/stricter-flake8
head_ref: 
base_ref: 
event.before: 0000000000000000000000000000000000000000
event.after: 7765bb704deda6fe9d92e15f476596232a4b3bec
```

the workflow run fails because determining the commit range fails. Which results in flake8 running over the complete repo (which fails).


Also changes that do not address tools (like a change to setup.cfg), result in an empty changed_repos.list ... which makes flake8 fail.

Do we want to do something against it?

bernt-matthias:
Seems that the `on: pull_request` is not necessary for our workflow. 

The `push` event runs the workflow for the first push to a new branch. When the pull request is created then the workflow run for the HEAD of the branch is shown.

For pushes that don't affect tools or data managers I added filters for paths.

bernt-matthias:
Wondering why the workflow does not run for this PR even if I reorder the commits such that the commit changing the data manager is the last. 

On my fork it seemed to work well.

bernt-matthias:
Quite cool. When I reword a commit of an open PR then the workflow is not triggered, but it is if the force pushed commit actually changes something. 

Getting more and more confident that this is ready to go.

nsoranzo:
> Seems that the on: pull_request is not necessary for our workflow.

That's not correct, nothing is shown now in the "Checks" tab. Can you revert the last commit, please?

> The workflow run fails because determining the commit range fails. Which results in flake8 running over the complete repo (which fails).

Can you guard the `flake8 $(cat ../workflow_artifacts/changed_repositories.list)` with `if [ -s ../workflow_artifacts/changed_repositories.list ]` ?

bernt-matthias:
> That's not correct, nothing is shown now in the "Checks" tab.

Really? I checked it on my own fork... If you like I can bring my master again in the state of this branch. 

> Can you guard the flake8 ... 

I could -- probably also for planemo test? But is this necessary now that I have restricted the path to those were tools are affected? May this be the reason my the checks did not show up in your experiments?

nsoranzo:
> > That's not correct, nothing is shown now in the "Checks" tab.
> 
> Really? I checked it on my own fork... If you like I can bring my master again in the state of this branch.

I mean there is nothing in the "Checks" tab of this PR (if you see any, try to view this page when not logged in).

> > Can you guard the flake8 ...
> 
> I could -- probably also for planemo test?

No need for that, the code copied from Travis already handles that.

> But is this necessary now that I have restricted the path to those were tools are affected? May this be the reason my the checks did not show up in your experiments?

Not sure I like the restriction, seems a bit too ad hoc, but maybe it's just me. Any way I did no further experiment, the issue is with removing `on: pull_request`.

bernt-matthias:
OK - undone.

> I mean there is nothing in the "Checks"

True .. But I do not understand it. In the experiments on my fork I had `on: push` triggered when I pushed. Here its `pull_request`. 

Also on my fork I see a push event when I push a new branch with `event_before: 0000000000000000` (many 0s .. not the exact number here) which fails. And If I open the PR for the branch I get another (`pull_request`) event triggered which succeeds. The result is that each job appears 2x in the checks, where one fails. 

Thought, with the guarded flake8 it shouldn't fail anymore, but still the workflow is triggered 2x.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,373 @@
...
+    #   (note that this does not work for feature branch events since we want all
+    #   commits on the feature branch and not just the commits of the last event)
+    # - for pull requests we compare against the 1st ancestor, given the current
+    #   HEAD is the merge between the PR branch and the base branch
+    - name: Set commit range (push to the feature branch)
+      if: github.ref != 'refs/heads/master' && github.event_name == 'push'
+      run: |
+        git fetch origin master
+        echo ::set-env name=COMMIT_RANGE::"origin/master..."
+    - name: Set commit range (push to the master branch, i.e. merge)
```
nsoranzo, line 73:
```suggestion
    - name: Set commit range (push to the master branch)
```

nsoranzo:
> True .. But I do not understand it. In the experiments on my fork I had on: push triggered when I pushed. Here its pull_request.

That's normal, the workflow is run twice, one on your feature branch under your account (as is also run before you open a PR), and one for the PR (where the feature branch is merged with the upstream code).

Regarding `event_before`, I'm not sure about the zeroes, but be careful that when you force-push this is set to the previous pushed HEAD commit.

Any way this is good to go from my side!

@mvdbeek?

bernt-matthias:
Cool. I have nothing to add at the moment.

bernt-matthias:
Should we keep 549172ce19cb6d257c1c30201ed576ce4296a0c5 in this PR?

nsoranzo:
> Should we keep [549172c](https://github.com/galaxyproject/tools-iuc/commit/549172ce19cb6d257c1c30201ed576ce4296a0c5) in this PR?

Sure, it's a good test.

mvdbeek:
This is awesome, thanks for much for all the work @bernt-matthias and @nsoranzo!

PR #2940, opened 2020-03-29
ci: check final success, test 20.05 and cache@v1.2

bernt-matthias:
Get the status of the merged test results, i.e. check if there is a job with status != success.

FOR CONTRIBUTOR:


.github/workflows/ci.yaml
```diff
@@ -1,239 +1,248 @@
...
+      - uses: actions/upload-artifact@v1
+        with:
+          name: 'All tool test results'
+          path: upload
+      - name: check status of combined status
+        run: |
+             stat=$(jq '.["tests"][]["data"]["status"]' upload/tool_test_output.json | grep -v "success" || true)
+             if [[ "$(echo -n $stat | wc -l)" -gt "0" ]]; then
+                 echo "unsuccessful tests found $(sort <<<$stat | uniq -c)"
+                 exit 1
```
mvdbeek: I intentionally did not do that, this will  prevent uploading the test artifacts.
If you want to do this you can download the published results in another step.

bernt-matthias: For my pull/pr action it seems to work: https://github.com/bernt-matthias/tools-galaxyp/actions/runs/65362695

mvdbeek: Alright then, either I didn't remember this correctly (probably) or it has changed.

mvdbeek: Do you have any idea why github thinks lines 1 to 228 have changed ?

bernt-matthias: For some reason the file seemingly had Win style line endings

mvdbeek: I think that's the other way round, `cat -v` on the workflow in your fork shows dos line endings

bernt-matthias: You are right

bernt-matthias:
How do you thing about adding a linter step/job?

mvdbeek:
I think that's not something that impacts long-term executability of jobs, which I think is what the purpose of the weekly jobs is. Repo-wide I also think it generates too much noise. That makes more sense for the PR tests.

I think you've also reverted the intended changes in the last commit.

bernt-matthias:
One more idea: is it feasible to switch from testing per repo to testing per tool? Advantage might be a more even load on the chunks. Would be to the cost of an additional call to planemo ci_find_tools.

mvdbeek:
Starting one instance per tool test seems too much overhead to me. I think it might make more sense to sort the repos by number of tools or tool tests included.

bernt-matthias:
Ah. OK. Then I will revert :( .. just have seen the chunk with the timeout in the weekly run and was thinking how to reduce runtime.

bernt-matthias:
Or might my space saving tricks (swapoff ...) might work?

mvdbeek:
> Or might my space saving tricks (swapoff ...) might work?

I really don't think turning swap off is a good idea. It also is unlikely to be sufficient, we have 2800 tests in 40 chunks, that's up to 70 tests per chunk, meaning up to 70 containers to download. In practice it is going to be less than that, but I really don't see this contributing to test time in a major way.

mvdbeek:
> just have seen the chunk with the timeout

we need to limit test run time, these are individual tests timing out.

bernt-matthias:
Took over some of the changes from the other workflow. How to proceed with caching? Seems that currently there is only a warning that caching does not work and it might be added in the future https://github.com/actions/cache/issues/63. So we could

- leave it as is
- remove caching
- implement as in the other PR (with caching of `.planemo`

What would be your preference.

bernt-matthias:
> Starting one instance per tool test seems too much overhead to me.

What I meant is to start a single planemo test run per chunk which just gets the list of tools that is determined with `planemo ci_find_tools`. (data managers would need another call to `planemo test`)

mvdbeek:
> What I meant is to start a single planemo test run per chunk which just gets the list of tools that is determined with `planemo ci_find_tools`. (data managers would need another call to `planemo test`)

That seems like it would quickly run out of space due to not cleaning up the docker containers ?

bernt-matthias:
True

mvdbeek:
Can you resolve the conflict ?

bernt-matthias:
Thanks @mvdbeek .. I was in offline land.

mvdbeek:
No worries. If you're happy with this feel free to merge.

PR #2954, opened 2020-04-06
be more strict with flake8

bernt-matthias:
since flake8 wasn't used before it should not have any side effects

plus some flake improvements that accumulated on my repo for testing

FOR CONTRIBUTOR:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [flake8]
-ignore = D,E201,E202,E501,W503,W504
+ignore = D,E501
 # For flake8-import-order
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -1,5 +1,5 @@
 [flake8]
-ignore = D,E201,E202,E501,W503,W504
```
wm75, line 2: Does it make sense to enable W503 **and** W504?
I'm not entirely sure, but wouldn't this emit warnings for binary operators *before* **and** *after* line breaks?
Wouldn't we want to allow one of them, i.e., ignore the outdated W503?

bernt-matthias:
Will do.

PR #2973, opened 2020-04-17
Data manager for SARS-CoV-2 primers from ARTIC project

pvanheus:
FOR CONTRIBUTOR:


pvanheus:
The [ARTIC project](https://artic.network/) has produced primers for amplicon sequencing of the SARS-CoV-2 virus. These primers are widely used. Galaxy workflows already exist to work with SARS-CoV-2 ARTIC amplicon data. The [IRIDA](https://www.irida.ca/) project is a user interface to routine bioinformatic analyses in a public health setting, using Galaxy as a workflow system. It only allows FASTQ input and a single reference dataset. 

This data manager allows a Galaxy admin to install the ARTIC primer files in their Galaxy server and thus produce Galaxy workflows for SARS-CoV-2 sequencing that are suitable for use with IRIDA.

bgruening:
@pvanheus there are  some flake8 errors.

pvanheus:
Ok this looks read-ish for merge. @dfornika care to have a look?

dfornika:
This looks great @pvanheus. The only change I might propose is to consider using a more generic name for the the tool data table. I've been using the name `primer_scheme_bedfiles`, because in theory there could be other research groups or protocols that create similar primer schemes and it might make sense to consolidate them under a single tool data table.

But I could imagine one could make the opposite argument that it would be better to have separate tool data tables, so I'll leave that up to you.

pvanheus:
Yes I was thinking about that @dfornika - for example [this](https://www.protocols.io/view/sars-cov-2-enrichment-sequencing-by-spiked-primer-beshjeb6) protocol. To make it more generic, the tool should have an option to select files from history. I'll put some work into that and make it more generic. So let's call this a WIP for now.

dfornika:
Personally I don't think you necessarily have to allow users to select files from history for this tool. If you did do that then you might consider renaming the tool.

For tools that consume these primer scheme files, I think it would make sense to allow users to either load from the tool data table or from their history.

pvanheus:
@dfornika I changed the name to `data_manager_primer_scheme_bedfiles` and added the ability to add a BED file to the primer schemes from history. The intention here is to allow Galaxy admins to configure an installation so that the resulting data table can be used in an IRIDA workflow.

I think this DM is now looking good and would appreciate a review and merge.

dfornika:
Looks good to me @pvanheus. I don't have access to review or merge on this repo. 

@bgruening: cant his be merged?

bgruening:
Of course! Do you need this on EU?

PR #2981, opened 2020-04-21
IDR download - Add error handling

beatrizserrano:
FOR CONTRIBUTOR:

Some errors fixed by @bgruening

### DIFF
--- a/tools/idr_download/idr_download_by_ids.py
+++ b/tools/idr_download/idr_download_by_ids.py
@@ -7,10 +7,22 @@ from omero.gateway import BlitzGateway  # noqa
 from omero.constants.namespaces import NSBULKANNOTATIONS  # noqa
 
 
-def warn(message, image_identifier):
+def warn(message, image_identifier, warn_skip=False):
+    message = message.rstrip()
+    if warn_skip:
+        if message[-1] in ['.', '!', '?']:
+            skip_msg = ' Skipping download!'
+        else:
+            skip_msg = '. Skipping download!'
+    else:
+        skip_msg = ''
     print(
-        'ImageSpecWarning for {0}: {1}'
-        .format(image_identifier, message),
+        'ImageSpecWarning for {0}: {1}{2}'
+        .format(
+            image_identifier,
+            message,
+            skip_msg
+        ),
         file=sys.stderr
     )
 
@@ -106,6 +118,19 @@ def download_image_data(
     coord=(0, 0), width=0, height=0, region_spec='rectangle',
     skip_failed=False
 ):
+    # basic argument sanity checks and adjustments
+    prefix = 'image-'
+    # normalize image ids by stripping off prefix if it exists
+    image_ids = [
+        intiid[len(prefix):] if iid[:len(prefix)] == prefix else iid
+        for iid in image_ids
+    ]
+
+    if region_spec not in ['rectangle', 'center']:
+        raise ValueError(
+            'Got unknown value "{0}" as region_spec argument'
+            .format(region_spec)
+        )
 
     # connect to idr
     conn = BlitzGateway('public', 'public',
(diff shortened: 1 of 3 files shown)


wm75:
The added error handling is not clear to me. The way it's implemented, it will swallow the inner ValueErrors with their more descriptive message, wouldn't it?
So which specific situation is this supposed to guard against? An exception thrown by `conn.getObject("Image", image_id)` or `image.getName()`?
If so, then only that specific code should be wrapped in its own `try: ... except:`.

tools/idr_download/idr_download_by_ids.py
```diff
@@ -115,6 +116,7 @@ def download_image_data(
     try:
         prefix = 'image-'
         for image_id in image_ids:
+            image_warning_id = 'No image ID could be retrieved/converted (%s)' % (image_id)
```
bgruening, line 119: Afaik this one was needed when the image_id was a well_id or something like that.

tools/idr_download/idr_download_by_ids.py
```diff
@@ -139,6 +141,7 @@ def download_image_data(
             image_warning_id = '{0} (ID: {1})'.format(
                 image_name, image_id
             )
+            warn('Download', image_warning_id)
```
bgruening, line 144: This could be removed, its was just added to see which images we actaully tried to download.

tools/idr_download/idr_download_by_ids.py
```diff
@@ -203,6 +206,8 @@ def download_image_data(
                 [image_name, str(image_id)] + [str(x) for x in tile]
             )
             download_plane_as_tiff(image, tile, z_stack, channel_index, frame, fname)
+    except Exception:
+        warn('Some random error', image_warning_id)
```
bgruening, line 210: This one is to capture all wired errors, that have not been caught before.
@wm75 all the other more explicit `warn()` will still be collected - nothing is overwritten. This one is only triggered if a new error that we have not seen before is raised - and finally, this one is making the tool at least useable. So we finally get 5000 images out of 200.000 ;)

wm75: Alas, this *does* interfere with the rest of the warnings and errors.
I've created https://gist.github.com/wm75/9580ada3926e195fa6383dd681d266b4 to demonstrate the issues.
In that gist, test 1 shows that the "random error" handling does not respect `skip_failed`.
Test 3 shows that it does swallow error messages.

wm75: Test 1 also reveals a malformatted warning message.

wm75:
An attempt to handle errors correctly in pretty much any circumstance. Mostly untested from my side so please try it out and fix errors before merging.

beatrizserrano:
Thanks so much for your help, @wm75! @sunyi000 has tested it in our local Galaxy instance and seems to work fine. I will restore here the original tests, that should work :)

beatrizserrano:
There's another issue that I've seen but not sure if it's me not using it well.
According to the help section:

> paste the URL into the input field of this tool, for example:
> https://idr.openmicroscopy.org/webclient/?show=image-9036708|image-9036710|image-9036711

But when I go to the IDR website, select a few images and copy the link the format is:
https://idr.openmicroscopy.org/webclient/?show=well-119032|well-119049|well-119088

As you can see, there are well ids there, not image ids. Maybe I'm selecting the wrong link, but that can lead either to exceptions or to downloading the wrong image (if by chance a well id corresponds to an image id too). @wm75 Do you remember if we discuss this at some point? Maybe I forgot :) Any hints @jburel?

wm75:
Weird. No idea how you got those well IDs. I select images in the webclient, click on that chain icon (`Link to this`) in the upper right corner, and it gives me a link with image-IDs in it.

beatrizserrano:
OK, I see now! If you select the thumbnails you get the well ids:

Screenshot 2020-04-22 at 15 26 10

Only selecting the image at the bottom you get the image ids in the URL:
Screenshot 2020-04-22 at 15 27 52

Maybe it's worth clarifying that in the help, can be confusing I think... I can rewrite it if you agree :)

jburel:
A well might have multiple images so it is required to select the image(s) you want in the bottom section

jburel:
An option could be to also add support for well ID

beatrizserrano:
I think that neither the image id nor the well id are informative for the users. So I would rather prefer to query by a metadata field (gene, reagent...). So supporting the well might add complexity to the tool and not improve the usability. Maybe we need to think of another tool that takes a list of genes and gives the image ids?

jburel:
@beatrizserrano that will certainly more natural. A query by gene could be doable that's what the OMERO.web mapr does, but it could take time to search across all the studies

beatrizserrano:
I think that time doesn't matter much in this case, it won't be a waiting time, just a tool included in the workflow which output will be plug to the IDR download. That shouldn't bother users much :)

beatrizserrano:
If all this green means that you also are happy with the changes, I'm ready to merge :)

wm75:
Well, the tests pass again and, if you're telling me that my changes did not seem to break anything, then we can merge this.
I'll test it again once it has landed on the EU server.

wm75:
@bgruening You're ok with this?

bgruening:
Sure!

beatrizserrano:
Thank you all!! :)

PR #2984, opened 2020-04-22
Updated Abricate to v1.0.1

Slugger70:
FOR CONTRIBUTOR:


bernt-matthias:
The release notes mention another new option an database: https://github.com/tseemann/abricate/releases

Do you intend to add those as well?

Tests covering the new options would also be very welcome.

Slugger70:
Done! Thanks for pointing it out Matthias!

bernt-matthias:
How about `--fofn` ?

Slugger70:
Could --fofn just be a collection or multiple selection of inputs? I'm not sure how having a text file of file names would work in the Galaxy context.

bernt-matthias:
If I get it right then `--fofn` is just another way to specify the samples that are processed. Question is if the program processed the files listed in the text file one after the other or jointly. 

If the former is the case, then we do not need the option since batch processing is built it in Galaxy. 

Otherwise one could add `multiple="true"` to the input param and do something along those lines 

```
#for f in $file_input
    #set $sample_name = re.sub('[^\w\-_\.]', '_', $f.element_identifier)
    ln -sf '$f' `${sample_name}` &&
    echo $sample_name >> fofn.txt
#end for

abricate --fofn fofn.txt
```

bernt-matthias:
But to me it seems that the tool is just doing batch processing .. so we can skip it. 

One could also specify multiple file names on the CLI. 

But lets skip this.

Slugger70:
It just runs them one at a time. (It's for batching purposes.) There is another tool that aggregates different abricate reports I think.

PR #2990, opened 2020-04-24
Update multiqc with qualimap

DarianHole:
FOR CONTRIBUTOR:

Includes:
- Changes to qualimap bamqc to have the input name be correct on the reports it generates for both itself and multiqc
- New multiqc modules available in 1.8 in the options list but commented out
- qualimap bamqc module

I'm not too sure on the way that the current way multiple inputs are put in so feedback is welcome


bgruening:
> Help making the program correctly accept collection lists and collection list:lists as I was having trouble figuring that out

Without looking at this tool specifically, any tool with `multiple="true"` on a "data" parameter, should accept list and nested lists.

bgruening:
Thanks @DarianHole for working on this!

DarianHole:
My issue was trying to create the needed file structure for the tool to detect the input when using both a normal collection or `multiple=true` and the nested list that is generated if you run 2+ samples at once.

File structure needs to be: `multiqc_WDir/qualimapX/sample_name_stats/...`

If i had a nested list it was making 2 outputs instead of 1 and I was having trouble combining them

DarianHole:
But also thanks for the quick response @bgruening , if you have a any suggestions let me know, or I'll keep playing with it till I find something

bgruening:
My ... maybe you can use `collection_type` see here: https://docs.galaxyproject.org/en/latest/dev/schema.html

But I wonder if we not better encourage the user to use the collection-flatten tool before to use this tool here. See here: https://usegalaxy.eu/root?tool_id=__FLATTEN__

DarianHole:
Ok I'll give each of those a shot

DarianHole:
@bgruening I've found something that works and allows both a normal collection and a flattened collection but I'm still not too happy with it. Thoughts?

tools/multiqc/multiqc.xml
```diff
@@ -663,6 +696,9 @@ sp:
                 <when value="prokka">
                     <param name="input" type="data" format="txt" multiple="true" label="Output of Prokka" help="It should contain 'contigs:'"/>
                 </when>
+                <when value="qualimap">
+                    <param name="input" type="data_collection" collection_type="list" format="txt,tabular,tsv,csv" multiple="true" label="Output of Qualimap BamQC" help="If a single output is wanted from a nested collection, run through the 'flatten collection tool first with the colon (:) delimiter" />
```
bgruening, line 700: can collection_type="list" now be removed? multiple=true should have the same effct

tools/multiqc/multiqc.xml
```diff
@@ -272,6 +272,33 @@ mkdir multiqc_WDir &&
     #elif str($repeat.software_cond.software) == "prokka"
         #set $pattern = "contigs:"
         @LN_FILES@
+    #elif str($repeat.software_cond.software) == "qualimap"
+        #for $file in $repeat.software_cond.input
+            #if re.search('(flattened)', str($repeat.software_cond.input.name))
+                #if re.search(':', str($file.element_identifier))
```
bgruening, line 278: can you not ask the user to provide this input ... if the file is a genome-result or raw-data?

We do similar  things for FASTQC, where we ask if the input is html or plain txt.

DarianHole: genome results is part of the raw data collection, it just ends up in a different directory to be recognized

tools/multiqc/multiqc.xml
DarianHole: I will attempt to change it around to that

DarianHole:
Issues currently:
- Qualimap bamqc outputs a collection of 10 elements of which 3 are used in the report
- folder name determines the sample name in multiqc
    - Ex. multiqc_Wdir/qualimap0/sample1 would be sample1 in the qualimapqc report
    - Have to have a way to try to correctly name the folder to the sample name, especially if more than 1 sample is put in
- The three files used in the report have to have the exact names and be in the correct folder:
    - "genomic_coverage.txt"
    - "coverage_histogram.txt"
    - "mapped_reads_gc-content_distribution.txt"

Anything else seems to break it or the samples are clumped together when they shouldn't be

DarianHole:
@bgruening Thoughts on this method to create sample directories and organize data for input

tools/multiqc/multiqc.xml
```diff
@@ -272,6 +272,31 @@ mkdir multiqc_WDir &&
     #elif str($repeat.software_cond.software) == "prokka"
         #set $pattern = "contigs:"
         @LN_FILES@
+    #elif str($repeat.software_cond.software) == "qualimap"
+        #set nested_dir_name = "unknown_sample"
+        #for $file in $repeat.software_cond.input
+            #if re.search("genome_results", str($file.element_identifier))
+                #set command = "grep 'bam file = ' {} | sed '1,$s/bam file = //g'".format(str($file))
```
bgruening, line 279: why are you not using pure bash here, why cheetah?

bgruening:
@DarianHole I like it very much! Thanks, I have just one comment.

tools/multiqc/multiqc.xml
DarianHole: No particular reason, I just though it would be easier to pass the variable in cheetah. If you believe bash is better I can switch it up. Should be an easy change

bgruening: The difference is that pure bash is executed on the cluster, everything in cheetah is executed on the headnode. I think its better to  push as much as possible to the cluster.

DarianHole: Got it! I'll quickly test and change that

DarianHole: Both ways to create a bash variable from a bash output that I know don't seem to be working, both the $( ) or \``.  I've tried the following:
- command=$(grep 'bam file = ' genome_results.txt | sed '1,$s/bam file = //g')
    - Fails bash syntax at the $ sign
- command=\`grep 'bam file = ' genome_results.txt | sed '1,$s/bam file = //g'\` 
    - cannot find $s

Any suggestions? Both work on the cl alone

bgruening: If you want to use `$` on bash, you need to write `\$` as `$` is the cheetah placeholder.

DarianHole: Sorry for all the questions, adding the `\` allows the command to run, but I cannot find the output variable with any combination of `/$sample_value` or `$sample_value` or `${sample_value}`

Also tried it with bash export command

DarianHole:
All bash now. Seems to work but I am not as confident with bash as I am with cheetah or python

bgruening:
Cool, thanks a lot @DanFaria

bgruening:
Sorry, thanks @DarianHole ... :) Also to @DanFaria and to everyone else that makes those tools here awesome!

PR #2999, opened 2020-04-28
permit anonymous usage of jbrowse

hexylena:
FOR CONTRIBUTOR:

In response to your comment @jennaj,

> We (@dannon & @mvdbeek) were wondering if there was a way to adjust Jbrowse to be able run in anonymous (not logged in) sessions. 

this should address your issue.

(Though it would be nicer if galaxy returned a fake 'anonymous' user when trying to access that in the tool form, but I guess very few tools use this "feature".)

@abretaud if you have a chance to review?

Edit: Screenshot of how it looks when anonymous:

image

### DIFF
--- a/tools/jbrowse/jbrowse.xml
+++ b/tools/jbrowse/jbrowse.xml
@@ -80,22 +80,16 @@ $trackxml &&
                       size="${reference_genome.genome.get_size(nice_size=True)}"
                       edam_format="${reference_genome.genome.datatype.edam_format}"
                       file_ext="${reference_genome.genome.ext}" />
-                  #if $reference_genome.genome.history:
                   <history id="${__app__.security.encode_id($reference_genome.genome.history_id)}"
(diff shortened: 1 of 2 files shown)


tools/jbrowse/jbrowse.xml
```diff
@@ -80,10 +80,22 @@ $trackxml &&
                       size="${reference_genome.genome.get_size(nice_size=True)}"
                       edam_format="${reference_genome.genome.datatype.edam_format}"
                       file_ext="${reference_genome.genome.ext}" />
+                  #if $reference_genome.genome.history:
```
mvdbeek, line 83: Is the outer if necessary ? I don't think we can have historyless datasets at this point.

hexylena: Sure doesn't need it! Good suggestion, rewrote a bit :)

PR #3005, opened 2020-04-30
[WIP]: Fixed some issues with iqtree

Slugger70:
FOR CONTRIBUTOR:

We have had trouble running iq-tree over Pulsar. iq-tree runs on the remote host just fine but pulsar cannot find the tool outputs.

**Problem 1: Wildcards**

In the tool xml, the outputs are collected from the working directory with a wildcard. Example:
```
<data name="bionj" format="nhx" from_work_dir="*.bionj" label="${tool.name} on ${on_string}: BIONJ Tree" />
```
But Pulsar is unable to look for files with a wildcard. :(

However, the output file prefix is set in the command thus:
```
iqtree
    -pre PREF
```
Therefore the outputs can be set to:
```
<data name="bionj" format="nhx" from_work_dir="PREF.bionj" label="${tool.name} on ${on_string}: BIONJ Tree" />
```
etc.

**Problem 2: Output filtering**

Two of the output files searched for in each run only exist when certain parameters are used.

* The output "contree" only exists when using bootstrapping (either fast or non-parametric)
* The output "splits.nex" only exists when using fast bootstrapping.

This doesn't cause a problem when running locally on Galaxy as it just returns an empty file for each, but on Pulsar it looks for a non-existant file and fails returning an error.

Therefore, these two outputs need to be filtered to suit. But I'm having trouble getting the filters to work! Help please!

**Problem 3: non-parametric bootstrapping parameters ill-formed**

The parameters `-bc` and `-bo` are not boolean parameters as set in the .xml file. They both expect an integer value. They are also mutually exclusive with each other and `-b`.

Therefore, the parameters need to be refactored into an option box and a integer input to represent the type of bootstrapping and the number of replicates respectively.

### DIFF
--- a/tools/iqtree/iqtree.xml
+++ b/tools/iqtree/iqtree.xml
@@ -262,12 +262,9 @@ $bootstrap_parameters.ultrafast_bootstrap.wbtl
 
 $bootstrap_parameters.ultrafast_bootstrap.bnni
 
-#if str($bootstrap_parameters.nonparametric_bootstrap.b) != ''
-    -b '$bootstrap_parameters.nonparametric_bootstrap.b'
+#if str($bootstrap_parameters.nonparametric_bootstrap.bootstrap_replicates) != ''
+    $bootstrap_parameters.nonparametric_bootstrap.bootstrap_type '$bootstrap_parameters.nonparametric_bootstrap.bootstrap_replicates'
 #end if
-
-$bootstrap_parameters.nonparametric_bootstrap.bc
-$bootstrap_parameters.nonparametric_bootstrap.bo
 
 #if str($miscellaneous_options.fconst) != ''
     -fconst '$miscellaneous_options.fconst'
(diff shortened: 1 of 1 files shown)


bgruening:
@Slugger70 are you running into this maybe? https://github.com/galaxyproject/pulsar/issues/193

Slugger70:
I thought so, but never the less, some files are not created at all unless certain parameters are chosen. So Pulsar can never return them.

bernt-matthias:
Problem 3 has already been solved. 

I included the changes wrt Problem 1 here:  https://github.com/galaxyproject/tools-iuc/pull/4724

Regarding Problem 2 I'm afraid that there are to many ways to enter bootstrap replicates in the tool in order to formulate reliable filters (without knowing the program in more detail). So I would skip this for now. Maybe non-existing files should be fixed on the pulsar side? Or should we just create empty files (via `touch`)? 

Closing this here. One more valuable contribution from @Slugger70 ..

PR #3016, opened 2020-05-05
PR workflow fix to early checkout

bernt-matthias:
@bgruening : lets see if this fixes it.

FOR CONTRIBUTOR:

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -27,9 +27,6 @@ jobs:
         echo 'base_ref: ${{ github.base_ref }}'
         echo 'event.before: ${{ github.event.before }}'
         echo 'event.after: ${{ github.event.after }}'
-    - uses: actions/checkout@v2
-      with:
-        fetch-depth: 0
     - uses: actions/setup-python@v1
       with:
         python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


bernt-matthias:
Seems to work now. Successful run with mock tool change is here: https://github.com/galaxyproject/tools-iuc/actions/runs/96007320 

I cleaned the git history already removed the mock tool change.

mvdbeek:
Why do you think this makes a difference ? Did you notice that the entire workflow only runs on the merge commit?

bernt-matthias:
> Did you notice that the entire workflow only runs on the merge commit?

No. And not exactly sure what you mean. Which workflow run?

> Why do you think this makes a difference?

Before: the checkout action seems to change directory, then the fake planemo test run tries to consume the complete repo, e.g. here: https://github.com/galaxyproject/tools-iuc/actions/runs/95653778. 

Also the checkout seems to do more than `git clone ... && cd repo_dir` (the working dir after checkout seems to be not the working dir before the checkout / repo_dir). This is why the upload of galaxy.sha was moved up.

mvdbeek:
> No. And not exactly sure what you mean. Which workflow run?
> 
> >

That was just a correlation, if you look at the action tab it's mostly the PRs on the master branch that run for 6 hours until they got killed. But there's also the medaka PR that ran for 6 hours.


> Before: the checkout action seems to change directory, then the fake planemo test run tries to consume the complete repo, e.g. here: https://github.com/galaxyproject/tools-iuc/actions/runs/95653778.

thanks, that's what I was wondering about.

PR #3027, opened 2020-05-09
Add vsnp_add_zero_coverage tool

gregvonkuster:
FOR CONTRIBUTOR:


tools/vsnp/vsnp_add_zero_coverage.xml
```diff
@@ -0,0 +1,172 @@
...
+#set input_vcf_dir = 'input_vcf_dir'
+#set output_vcf_dir = 'output_vcf_dir'
+#set output_metrics_dir = 'output_metrics_dir'
+mkdir -p $input_bam_dir &&
+mkdir -p $input_vcf_dir &&
+mkdir -p $output_vcf_dir &&
+mkdir -p $output_metrics_dir &&
+#if str($input_type) == "single":
+    #set bam_input = $input_type_cond.bam_input
+    #set file_name = $bam_input.file_name
```
bgruening, line 22: instead of file_name you should use element_identifier I think ..

tools/vsnp/vsnp_add_zero_coverage.xml
```diff
@@ -0,0 +1,172 @@
...
+    #set file_name_base = $os.path.basename($file_name)
```
bgruening, line 23: Do you assume that the dataset name is a path here?

tools/vsnp/vsnp_add_zero_coverage.xml
```diff
@@ -0,0 +1,172 @@
...
+            <output_collection name="output_vcf_collection" type="list">
+                <element name="vcf_input.vcf" file="output_vcf.vcf" ftype="vcf" compare="contains"/>
+                <element name="vcf_input2.vcf" file="output_vcf.vcf" ftype="vcf" compare="contains"/>
+            </output_collection>
+            <output_collection name="output_metrics_collection" type="list">
+                <element name="vcf_input.tabular" file="output_metrics.tabular" ftype="tabular" compare="contains"/>
+                <element name="vcf_input2.tabular" file="output_metrics.tabular" ftype="tabular" compare="contains"/>
+            </output_collection>
+        </test>
+        <test>
```
bgruening, line 137: can you please add a test for the cached case?

tools/vsnp/test-data/output_metrics.tabular
```diff
@@ -0,0 +1,2 @@
+# File	Number of Good SNPs	Average Coverage	Genome Coverage
+	0
```
bgruening, line 2: is this 0 an expected output?

tools/vsnp/vsnp_add_zero_coverage.xml
```diff
@@ -0,0 +1,172 @@
...
+
+A metrics file is produced for each combination which provides the number of good SNPs, the average coverage and the genome
+coverage percentage.
+
+**Required Options**
+
+ * **Choose the category of the files to be analyzed** - select "Single files" or "Collections of files", then select the appropriate history items (single BAM and VCF files or collections of BAM and VCF files) based on the selected option.
+ * **Choose the source for the reference genome** - select "locally cached" if the reference associated with the BAM and VCF files is available within the Galaxy environment or "from history" to select the reference from the current history.
+    </help>
+    <citations>
```
bgruening, line 161: @gregvonkuster no that everything is in one folder you can share many things in a common macro file. E.g. this citation or maybe even the requirements.

gregvonkuster: Ah, yes, sorry about this.  Macros implemented here https://github.com/galaxyproject/tools-iuc/pull/3027/commits/f795fe5bf821cc4376f65923842a0220fe121668

tools/vsnp/test-data/output_metrics.tabular
gregvonkuster: Yes, the primary goal was small data files for testing, so 0 good SNPs is correct for the input data.

tools/vsnp/vsnp_add_zero_coverage.xml
gregvonkuster: Do you mean a "cached" reference installed via a data manager tool?  I didn't know these kinds of tests are now available for tools - can you point me to an example of a tool test that does this?

tools/vsnp/vsnp_add_zero_coverage.xml
gregvonkuster: Yes, $bam_input.file_name returns the full path to the dataset on disk.

tools/vsnp/vsnp_add_zero_coverage.xml
gregvonkuster: In this case, I'm using $bam_input.file_name instead of $element_identifier because this and other vSNP tools in the workflow depend heavily on the disk file name for the dataset.  This becomes a bit complex due to the differences in the way that the Galaxy framework handles single file inputs verses input dataset collections, and this tool accepts both kinds of inputs.

For dataset collections, the $element_identifier approach will work well, and it will return the file name on disk (e.g., BCG_human_R6000.bam) for each element in the collection.  However, for single file inputs, the $element identifier will be a sting like ```vSNP: add zero coverage on 15```, whereas the $bam_input.file_name would be something like ```dataset_2342.dat```.  Although the ```dataset``` name is not so useful, it is much better throughout the entire workflow than the history item string label for these single file inputs.

tools/vsnp/vsnp_add_zero_coverage.xml
gregvonkuster: Hmm...looks like the requirements macro doesn't work for pandas.  I don't see a typo :(

```job standard error:
  File "/home/runner/.planemo/planemo_tmp_tys1ykrr/vsnp_build_tables.py", line 6, in <module>
    import pandas
ModuleNotFoundError: No module named 'pandas'```

File: tools/vsnp/macros.xml [hunk 6]
```diff
@@ -0,0 +1,48 @@
...
+            <option value="collection">Collections of files</option>
+        </param>
+    </xml>
+    <xml name="param_reference_source">
+        <param name="reference_source" type="select" label="Choose the source for the reference genome">
+            <option value="cached" selected="true">locally cached</option>
+            <option value="history">from history</option>
+        </param>
+    </xml>
+    <token name="@REQUIREMENT_BIOPYTHON@"><requirement type="package" version="1.76">biopython</requirement></token>
```
bgruening, line 17: its not a token, its a `<xml>`

gregvonkuster: Ah, ok, can't use tokens for xml tag sets within the macros.xml file itself, so I think moving the requirements tag sets to macros.xml creates more lines of code.  :)

tools/vsnp/vsnp_add_zero_coverage.xml
gregvonkuster: I found a tool that has a cached reference test like this and added the test for this tool here https://github.com/galaxyproject/tools-iuc/pull/3027/commits/ea05ce148f47169fc1c12dd50098fa2385da0d6e

tools/vsnp/macros.xml
bgruening: You can use that fancy stuff https://github.com/galaxyproject/tools-iuc/blob/24f33bda62a7b6771ad42a39fe8c683e09f6d8b8/tools/freebayes/macros.xml#L7 :)

gregvonkuster: Uh, what does the ```<yield />``` do for requirements?  :)

gregvonkuster: I reverted the requirements tags back to the tool config here https://github.com/galaxyproject/tools-iuc/pull/3027/commits/c4915451641c3a72e35fae1c26aa3fa890193e5c.  I hope this is ok.  

The reason is that I don't see much benefit to making them macros since the combinations of requirements across tools does not easily allow for patterns.

Here are the tools and combinations of requirements, if you see a pattern that makes sense to move to a macro, please let me know.  I suppose the biopython-pandas combination may be one, but I don't think it brings much benefit.
```
vsnp_statistics - numpy, pandas, xlrd, xlsxwriter
vsnp_add_zero_coverage - biopython, pandas, pysam
vsnp_build_tables - biopython, pandas, xlsxwriter
vsnp_get_snps - pandas, pyvcf, xlrd
vsnp_determine_ref_from_data - biopython, pyyaml
```

tools/vsnp/vsnp_add_zero_coverage.xml
martenson: @gregvonkuster Did you resolve this? I am seeing the same problem with pandas in a different tool.

gregvonkuster: @martenson I believe this was resolved and all tests are passing.  I don't remember the cause for sure since its been quite a while, but possibly a version conflict?

bgruening (approved):

bgruening:
Sorry, I somehow missed this :(

PR #3032, opened 2020-05-13
download idr images as tarball

sunyi000:
FOR CONTRIBUTOR:


sunyi000:
@beatrizserrano @bgruening 
code works but may need review, both .py and .xml file
@bgruening I was trying to use has_archive_member in test case, but haven't figured out the proper way to do it yet..any pointers are much appreciated

tools/idr_download/idr_download_by_ids.xml
```diff
@@ -86,6 +87,9 @@
         <param name="skip_failed" type="boolean" checked="false" truevalue="--skip-failed" falsevalue=""
         label="Skip failed retrievals?"
         help="By default the tool will fail with an error on the first non-retrievable image ID. Here, you can choose to skip non-retrievable image IDs and continue downloading the available ones instead. The error log will contain warnings about failed IDs in this case." />
+        <param name="download_tar" type="boolean" checked="false" truevalue="--download-tar" falsevalue=""
+        label="Download images in a tarball?"
+        help="By default the tool will download individual image into Galaxy history. With this option set to Yes, you can package all images into a single tarball." />
```
beatrizserrano, line 92:
```suggestion
        help="By default, the tool will download individual images into your Galaxy history. With this option set to Yes, you can package all images into a single tarball." />
```

tools/idr_download/idr_download_by_ids.py
```diff
@@ -139,6 +135,11 @@ def download_image_data(
     conn.connect()
 
     try:
+        if download_tar:
+            archive = tarfile.open("images.tar", mode='w')
```
beatrizserrano, line 139: should we make it mandatory when the number of images to download exceeds a certain limit?

sunyi000: yes we can change it.. if that's the case, we don't need the parameter in xml wrapper..
let's discuss this and then make decision..

wm75: I would not recommend it. Magically changing a tool's output format is almost always a bad design decision that will cause you trouble at some point. For example, how would you use such a tool in a workflow if you cannot know which downstream tools will be able to handle the output.
The current version is doing it just right :)

beatrizserrano: @wm75 so you propose to download just as a tarball?

wm75: Ah no, I meant leave it as an explicit choice of the user. Sorry if that wasn't clear.

tools/idr_download/idr_download_by_ids.py
```diff
@@ -139,6 +135,12 @@ def download_image_data(
     conn.connect()
 
     try:
+        # create an archive file for write operation
+        if download_tar:
+            archive = tarfile.open("images.tar", mode='w')
+        else:
+            archive = None
```
wm75, line 142: This flag is redundant with `download_tar`, isn't it? Simplify by relying only on the latter.

tools/idr_download/idr_download_by_ids.py
```diff
@@ -263,14 +265,44 @@ def download_image_data(
...
+
+                im_array = get_image_array(image, tile, z_stack, channel_index, frame)
+
+                # pack images into tarball.
+                # I have tried to convert array to BytesIO file object and add it to tarball,
+                # but after untar the tarball, the image headers are empty so I'm using TemporaryFile
+                # instead. This may need to revised
+                if download_tar and archive is not None:
+                    tar_img = Image.fromarray(im_array)
+                    buf = TemporaryFile()
```
wm75, line 281: This might be a use case for `tempfile.SpooledTemporaryFile`, if you want to get sophisticated :)
Could save some IO and could be used in a with statement.

tools/idr_download/idr_download_by_ids.py
```diff
@@ -263,14 +265,44 @@ def download_image_data(
...
             except Exception as e:
                 if skip_failed:
                     # respect skip_failed on unexpected errors
                     warn(str(e), image_warning_id, warn_skip=True)
                     continue
                 else:
                     raise
+
+        # close the archive file,if it is created
+        if archive is not None:
```
wm75, line 303: Move to the `finally` clause below and check `download_tar` directly.

sunyi000:
> @sunyi000 in the case of the tarball you only want to output one tarball and not a collection with one tarball, isn't it? In this case you also need to change your output section.

yes I agree. I thought about this before.. what's the best way to do this in <outputs>,i tried to use 'conditional' in <outputs>but seems it is not allowed....thanks

bgruening:
@sunyi000 you can use filters, like here: https://github.com/galaxyproject/tools-iuc/blob/57493fdc88e6128bed127993464f2a6480e2774d/tools/colibread/commet.xml#L53

wm75:
> > @sunyi000 in the case of the tarball you only want to output one tarball and not a collection with one tarball, isn't it? In this case you also need to change your output section.
> 
> yes I agree. I thought about this before.. what's the best way to do this in 'outputs',i tried to use 'conditional' in 'outputs' but seems it is not allowed....thanks @bgruening

You need to define two, mutually exclusive outputs and use filters to decide, which one to produce based on user inputs. See here: https://docs.galaxyproject.org/en/master/dev/schema.html#tool-outputs-collection-filter

tools/idr_download/idr_download_by_ids.py
sunyi000: thanks, sorry im a bit confused here :) do you mean just checking archive =None, or give it a default value or a one liner if-else? thx

tools/idr_download/idr_download_by_ids.py
sunyi000: I will try that SpooledTemporayFile and with statement

tools/idr_download/idr_download_by_ids.py
wm75: My suggestion is to use only `download_tar` and check it everywhere you are currently using `archive`.

sunyi000: ah OK..got it

tools/idr_download/idr_download_by_ids.xml
```diff
@@ -86,6 +87,9 @@
         <param name="skip_failed" type="boolean" checked="false" truevalue="--skip-failed" falsevalue=""
         label="Skip failed retrievals?"
         help="By default the tool will fail with an error on the first non-retrievable image ID. Here, you can choose to skip non-retrievable image IDs and continue downloading the available ones instead. The error log will contain warnings about failed IDs in this case." />
+        <param name="download_tar" type="boolean" checked="false" truevalue="--download-tar" falsevalue=""
```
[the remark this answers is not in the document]
sunyi000: changed default to true in this commit, and also the output section.

tools/idr_download/idr_download_by_ids.xml
```diff
@@ -203,6 +207,30 @@
...
+                <param name="width" value="671" />
+                <param name="height" value="511" />
+            </conditional>
+            <param name="frame" value="0" />
+             <param name="download_tar" value="true" />
+            <output_collection name="output_file" type="list">
+                <element name="images" ftype="tar">
+                    <assert_contents>
+                        <has_size value="1382400" />
+<!--                         <has_archive_member path="HT26 [Well A4, Field 1]__1828167__0__0__671__511.tiff" />
```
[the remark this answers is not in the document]
sunyi000: i believe yes spaces and brackets are the causes of the problem after testing. 
tried to escape them but don't seem to work, so I have used * in the first part of the filename, if this is ok

tools/idr_download/idr_download_by_ids.xml
```diff
@@ -86,12 +87,18 @@
...
         label="Skip failed retrievals?"
         help="By default the tool will fail with an error on the first non-retrievable image ID. Here, you can choose to skip non-retrievable image IDs and continue downloading the available ones instead. The error log will contain warnings about failed IDs in this case." />
+        <param name="download_tar" type="boolean" checked="true" truevalue="--download-tar" falsevalue=""
+        label="Download images in a tarball?"
+        help="By default, the tool will download a tarball containing individual images into your Galaxy history. With this option set to No, you can download a collection of images" />
     </inputs>
     <outputs>
         <data name="out_log" format="txt" label="${tool.name} error log" />
         <collection name="output_file" type="list">
             <discover_datasets pattern="__name_and_ext__" directory="downloads" />
```
bgruening, line 97: here you should add a ` <filter>not download_tar</filter>` left?

tools/idr_download/idr_download_by_ids.py
sunyi000: could you review the last commit again and see if thats the proper way to use spooledtemporryfile. thx

tools/idr_download/idr_download_by_ids.xml
```diff
@@ -203,6 +216,31 @@
...
+            </conditional>
+            <param name="frame" value="0" />
+            <output name="output_tar">
+                    <assert_contents>
+                        <has_size value="1382400" />
+                        <has_archive_member path=".*/*__1828167__0__0__671__511.tiff" >
+                            <has_size value="685762" />
+                        </has_archive_member>
+                        <has_archive_member path=".*/*__1828658__0__0__671__511.tiff" >
+                            <has_size value="685762" />
```
bgruening, line 239: Very cool!

tools/idr_download/idr_download_by_ids.xml
sunyi000: ah yes.. i will add it so the two outputs are exclusive to each other

tools/idr_download/idr_download_by_ids.py
```diff
@@ -263,15 +263,40 @@ def download_image_data(
...
+                im_array = get_image_array(image, tile, z_stack, channel_index, frame)
+
+                # pack images into tarball
+                if download_tar:
+                    tar_img = Image.fromarray(im_array)
+                    with SpooledTemporaryFile() as buf:
+                        tar_img.save(buf, format='TIFF')
+                        buf.seek(0)
+                        tarinfo = tarfile.TarInfo(name=fname)
+                        tarinfo.size = len(tar_img.tobytes())
```
wm75, line 280: One final optimization suggestion:
Instead of loading the `tar_img.tobytes()` representation of the image into memory, couldn't you just use `buf.tell()` to know how many bytes have been written?

sunyi000: I actually tried buf.tell() but I get a different size of the tar, also the images in the tar has 0 bytes.
not sure what I have missed..

wm75: Hmm, not sure either, but you could force buf's cursor position to the end of the stream by doing `buf.seek(0, 2)` before doing `buf.tell()` just in case it does not get advanced automatically.
As for the different sizes, I was wondering about that anyway. Why is `tar_img.tobytes()` supposed to be the same as the TIFF's size? Shouldn't the TIFF be a bit larger than the pure image data? In which case the tell() value would be accurate and not the tobytes() length?

tools/idr_download/idr_download_by_ids.py
wm75: This looks really good already. What you should still do is provide a max_size to `SpooledTemporaryFile()` so that the whole thing gets written to disc if it becomes too large.
What you will have to test is whether `tar_img.save` gives `buf` a chance to write its content to disc in between. If not, you would end up with the whole thing in memory anyway, in which case you'd have to revert to a regular `TemporaryFile`. Guess you could try downloading a rather large image and monitor memory consumption on your machine to find out.

tools/idr_download/idr_download_by_ids.py
sunyi000: You are right.. need to move the cursor back to 0 after buf.tell() then it will work. Also this time the TIFF is a bit bigger...
I will update shortly, thanks very much

tools/idr_download/idr_download_by_ids.py
sunyi000: I will try find a large image and test it then...cheers

tools/idr_download/idr_download_by_ids.xml
```diff
@@ -86,12 +87,20 @@
         <param name="skip_failed" type="boolean" checked="false" truevalue="--skip-failed" falsevalue=""
         label="Skip failed retrievals?"
         help="By default the tool will fail with an error on the first non-retrievable image ID. Here, you can choose to skip non-retrievable image IDs and continue downloading the available ones instead. The error log will contain warnings about failed IDs in this case." />
+        <param name="download_tar" type="boolean" checked="true" truevalue="--download-tar" falsevalue=""
+        label="Download images in a tarball?"
+        help="By default, the tool will download a tarball containing individual images into your Galaxy history. With this option set to No, you can download a collection of images" >
```
beatrizserrano, line 92:
```suggestion
        help="By default, the tool will download a tarball containing individual images into your Galaxy history. This option is strongly recommended, especially for a large number of files. With this option set to No, you will download a collection of images, only suggested for a small number of files." >
```

beatrizserrano:
@wm75 is this ready to merge or is there something else to do?

wm75:
I think @sunyi000 still needs to investigate if the use of SpooledTemporaryFile is a good idea here or not. What you'd want to avoid as much as possible is having potentially large images held in memory unnecessarily. If we go with SpooledTemporaryFile, we still need to figure out a reasonable size threshold beyond which image data should get written to disk instead of being kept in memory. Alternatively, he could just switch back to TemporaryFile as initially. It's a simpler, but not too bad solution.

sunyi000:
> I think @sunyi000 still needs to investigate if the use of SpooledTemporaryFile is a good idea here or not. What you'd want to avoid as much as possible is having potentially large images held in memory unnecessarily. If we go with SpooledTemporaryFile, we still need to figure out a reasonable size threshold beyond which image data should get written to disk instead of being kept in memory. Alternatively, he could just switch back to TemporaryFile as initially. It's a simpler, but not too bad solution.

How about I switch back to TemporaryFile for now, so this can get merged and @beatrizserrano can start doing some of her work..then I will see how to run proper tests for SpooledTemporaryFile ?

Does this sound ok?
Cheers

wm75:
Sure, absolutely fine with me:)

PR #3039, opened 2020-05-18
Plot with Scanpy: Changes way in which coloring defaults are set

pcm32:
FOR CONTRIBUTOR:

In the current version using the default colors and palette in tool the leads to:

```
sc.pl.scatter(
    adata,
    save='.png',
...
    palette='',
    color_map='',
...
    frameon=True)
```

which has the unintended effect of actually not using any color or palette:

image

In contrast, with the additions in this PR, when selecting Default you actually allow scanpy to use its default palette and colours, producing the following code:

```
sc.pl.scatter(
    adata,
    save='.png',
    show=False,
    x='n_counts',
    y='n_genes',
    color='individual',
    use_raw=False,
    sort_order=False,
    projection='2d',
    legend_loc='right margin',
    legend_fontsize=10,
    legend_fontweight='normal',
    frameon=True)
```

which doesn't set color and palette to `''`.

This also adds the ability to color by a field in `obs` or `var`, producing:

image

### DIFF
--- a/tools/scanpy/plot.xml
+++ b/tools/scanpy/plot.xml
@@ -25,6 +25,9 @@ sc.pl.scatter(
     #if $method.type.type == 'xy'
     x='$method.type.x',
     y='$method.type.y',
+    #if str($method.type.color) != ''
+    color='$method.type.color',
+	  #end if
         #if str('$method.type.layers.use_layers') == 'true'
     layers=('$method.type.layers.layer_x', '$method.type.layers.layer_y', '$method.type.layers.layer_color'),
         #end if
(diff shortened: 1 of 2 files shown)


tools/scanpy/plot.xml
```diff
@@ -414,6 +419,7 @@ sc.pl.rank_genes_groups_stacked_violin(
                     <when value="xy">
                         <param argument="x" type="text" value="" label="x coordinate" help="Index or key from either '.obs' or '.var'"/>
                         <param argument="y" type="text" value="" label="y coordinate" help="Index or key from either '.obs' or '.var'"/>
+                        <param argument="color" type="text" value="" label="Color by" help="Color points by single variable in `.obs` or `.var`"/>
```
mtekman, line 422: I was thinking maybe an `optional` here, but it's the same user experience both ways so this is fine :-)

mtekman:
Looks like the output summary function is being applied to the plot, let me see if I can fix this

pcm32:
Thanks for the prompt review @mtekman !

PR #3071, opened 2020-06-10
starting with ena_upload_cli wrapper

ieguinoa:
- Still missing the handling of a password for submissions
- Need to extend to other cases: modify only some metadata? modify the datasets ? not sure if that's even possible


ieguinoa:
Still missing some test, working on that now. But I think it's ready for review by IUC

ieguinoa:
@bgruening can you check on the general structure? will add the tests now, but don't think it's possible to test submissions, just the preprocessing.

bgruening:
> @bgruening can you check on the general structure? will add the tests now, but don't think it's possible to test submissions, just the preprocessing.

I think that is fine. You can also check the generated commandline in a test.

tools/ena_upload/.shed.yml
```diff
@@ -0,0 +1,12 @@
+categories:
+    - TODO
```
bgruening, line 2: `Data Export`

tools/ena_upload/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+To enable users to set their credentials for this tool,
+make sure the file `config/user_preferences_extra_conf.yml` has the following section:
+
+```
+    ena_account:
+        description: Your ENA Brokering account details
+        inputs:
+            - name: webin_id
+              label: webin_id
```
bgruening, line 16: ENA Webin ID

tools/ena_upload/ena_upload.xml
```diff
@@ -0,0 +1,203 @@
+<tool id="ena_upload" name="ENA Upload tool" version="0.2" python_template_version="3.5">
```
bgruening, line 1: remove `python_template_version="3.5"`

tools/ena_upload/ena_upload.xml
```diff
@@ -0,0 +1,203 @@
...
+  <requirements>
+    <requirement type="package" version="@VERSION@">ena-upload-cli</requirement>
+    <requirement type="package" version="1.2.0">xlrd</requirement>
+  </requirements>
+  <command detect_errors="exit_code"><![CDATA[
+cwd=\$(pwd);
+#set $use_secret = "False"
+#set $webin_id = $__user__.extra_preferences.get('ena_account|webin_id', "").strip()
+#set $password = $__user__.extra_preferences.get('ena_account|password', "").strip()
+echo $webin_id;
```
bgruening, line 15: remove ...

tools/ena_upload/ena_upload.xml
```diff
@@ -0,0 +1,203 @@
...
+#if $dry_run_option == "False":
+ena-upload-cli
+    --tool 'ena-upload-cli v@VERSION@ @ Galaxy'
+    --action '$action_options.action'
+    --center '$action_options.center'
+    --webin_id '$webin_id'
+    #if $use_secret == "False":
+        --password $password
+    #else:
+        --secret $webin_secret
```
bgruening, line 98: we can not do this ... as this would expose the credentials onto the command line.
Can you use this trick? https://github.com/galaxyproject/tools-iuc/blob/master/tools/pyega3/pyega3.xml#L37

So the credentials are ONLY written to a config file. This way they are not exposed on the CMD and also not in an ENV. Environment variables are also exposed in some Galaxy systems as part of provenance. So the file solution seems to be the best one.

tools/ena_upload/ena_upload.xml
```diff
@@ -0,0 +1,203 @@
...
+    #else:
+        #if $action_options.input_format_conditional.viral_submission == "true":
+          --vir
+        #end if
+    #end if
+#end if
+#if $action_options.submit_dev == "true":
+    -d
+#end if
+  > $output
```
bgruening, line 130: please quote all out and inputs

tools/ena_upload/ena_upload.xml
```diff
@@ -0,0 +1,203 @@
...
+    --experiment $action_options.input_format_conditional.experiments_users_table
+    --study $action_options.input_format_conditional.studies_users_table
+    --run $action_options.input_format_conditional.runs_users_table
+    --sample $action_options.input_format_conditional.samples_users_table
+    #if $action_options.input_format_conditional.viral_submission == "true":
+        --vir
+    #end if
+#else:
+    --experiment $experiments_table_path
+    --study $studies_table_path
```
bgruening, line 114: quote

bgruening (commented):
That tool is awesome! @ieguinoa the biggest issue is in how we handle secrets here. I left a comment. Do you think this is possible?

ieguinoa:
Upgraded cli version to the new one that handles username+pass via file.
Added a small trick to check the contents of preferences without revealing anything to command line or environment.
other small fixes
Still missing the tests.

tools/ena_upload/README.md
```diff
@@ -0,0 +1,23 @@
+# ena-upload-wrapper
+Galaxy wrapper for ena-cli-upload
+
+This tool is shipped in a ready to use Galaxy container found [here](https://github.com/ELIXIR-Belgium/ena-upload-container).
```
bgruening, line 4: I guess we should mention the global configuration file as well.

tools/ena_upload/ena_upload.xml
```diff
@@ -0,0 +1,208 @@
...
+    <import>samples_macros.xml</import>
+</macros>
+  <requirements>
+    <requirement type="package" version="@VERSION@">ena-upload-cli</requirement>
+    <requirement type="package" >pyyaml</requirement>
+    <requirement type="package" version="1.2.0">xlrd</requirement>
+  </requirements>
+  <command detect_errors="exit_code"><![CDATA[
+cwd=\$(pwd);
+webin_id=\$( python -c "import yaml; print( yaml.load(open('$credentials').read(), Loader=yaml.SafeLoader).get('username','') )" );
```
bgruening, line 12: if this is too complicated you could simply grep for it. This way you do not need to start a python process and grep can just return 0 or 1.

tools/ena_upload/ena_upload.xml
```diff
@@ -0,0 +1,208 @@
...
+      
+if [ "\$webin_id" = "" ]; then
+  ## No credentials in user defined preferences    
+  ## Fallback to global defined credentials (if exist)   
+  #import os
+  #if os.path.isfile(os.environ.get('ENA_SECRETS', '')):
+      credentials_path=\${ENA_SECRETS};     
+      webin_id=\$( python -c "import os, yaml; print( yaml.load(open(os.environ.get('ENA_SECRETS','')).read(), Loader=yaml.SafeLoader).get('username','') )" );
```
bgruening, line 20: same here, simply grep?

tools/ena_upload/ena_upload.xml
```diff
@@ -0,0 +1,208 @@
...
+  </requirements>
+  <command detect_errors="exit_code"><![CDATA[
+cwd=\$(pwd);
+webin_id=\$( python -c "import yaml; print( yaml.load(open('$credentials').read(), Loader=yaml.SafeLoader).get('username','') )" );
+      
+if [ "\$webin_id" = "" ]; then
+  ## No credentials in user defined preferences    
+  ## Fallback to global defined credentials (if exist)   
+  #import os
+  #if os.path.isfile(os.environ.get('ENA_SECRETS', '')):
```
bgruening, line 18: should we call it GALAXY_ENA_SECRETS?

tools/ena_upload/ena_upload.xml
```diff
@@ -0,0 +1,208 @@
...
+  #end if
+else
+  credentials_path='$credentials';
+fi;
+      
+#set working_dir = os.getcwd()
+#set $dry_run_option = "False"
+#set viral_submission = "False"
+#if $action_options.input_format_conditional.input_format == "build_tables":
+  python $__tool_directory__/extract_tables.py --action $action_options.action --out_dir \$cwd --studies $studies_json;
```
bgruening: You could use here `./` as you do below.

tools/ena_upload/ena_upload.xml
```diff
@@ -0,0 +1,208 @@
...
+  #set $studies_table_path = "$cwd/studies.tsv"
```
bgruening: this is all not needed I think. You could just use `./studies.tsv`

tools/ena_upload/ena_upload.xml
```diff
@@ -0,0 +1,207 @@
...
+  #end if
+else
+  credentials_path='$credentials';
+fi;
+      
+#set working_dir = os.getcwd()
+#set $dry_run_option = "False"
+#set viral_submission = "False"
+#if $action_options.input_format_conditional.input_format == "build_tables":
+  python '$__tool_directory__/extract_tables.py' --action $action_options.action --out_dir \$cwd --studies $studies_json;
```
bgruening, line 37: I would remove cwd here and use `./` or be cleaner and create a `./submission_files/` dir.

tools/ena_upload/ena_upload.xml
```diff
@@ -0,0 +1,207 @@
...
+
+## create the list of files to upload and make the symlinks 
+#set $files_to_upload = list()
+#if $action_options.input_format_conditional.input_format == "build_tables":
+    #for $study in $action_options.input_format_conditional.conditional_viral_metadata.rep_study:
+      #for $sample in $study.rep_sample:
+        #for $experiment in $sample.rep_experiment:
+          #for $run in $experiment.rep_runs:
+            #for $file in $run.upload_files:
+                ln -s $file $file.element_identifier;
```
bgruening: @ieguinoa do we need this element_identifier or can we simply give them unique names? The reason is that element_identifier can be changed freely by the user and can potentially include strange symbols, whitespaces etc ...

bgruening (commented):
Nice, very close I guess. Can we check the commandline and preprocessing as text? Or any other idea?

tools/ena_upload/ena_upload.xml
ieguinoa: The thing is: the users can upload all the metadata about a file through a table or an excel file. The only way I could find to make the link between the filename they define for a run in these tables and the dataset (inluded in a multiple dataset entry field) was to ask them to use the same identifier that is in the metadata table. 
alternatively, we can ask the users to, one by one, write down the filename associated with each dataset selected, but this is a bit of a hassle if there are a lot of files.
But of course I'm more than open for ideas, that part was also not easy to code :P

bgruening: What about adding some help text and say the files should only be named with alphanumeric characters, no white-spaces etc... and than use https://github.com/galaxyproject/tools-iuc/blob/0b2de0d35b6ef4c0ef0dd77e8415115a4a3b415b/tools/fargene/fargene.xml#L11 do keep this tool save. Would that work? 

(we could also apply the same regex to the excel table in theory ...)

ieguinoa: just pushed the change applying that rule to the element_identifier. Not sure if applying it to the excel table entries for the file names is needed as that is just written in the tabular file and sent to the cli tool, but I guess won't hurt doing it. Most important is to let users know what could cause an error here, mainly whitespaces.

bgruening: Awesome! Are you planning more tests, or should I look at it tm and get it in if possible?

ieguinoa: just added the nonviral version to test the parsing of the excel file, still all is dry-run tests just to check the parsing.
Will add the test for submissions using the wrapper form instead of everything coming from a file.
Then will add the the submission step test: command line parsing and checking that the ena_cli tool returns that the credentials are not correct, that's as far as we can go I guess.

ieguinoa:
@bgruening can you check on the error its showing for the last test? 
says "parameter 'upload_files': specify a dataset of the required format / build for parameter", it's the same input definition used in the other tests, so not sure why it's now complaining about the format, tried with the other file 'sample.fq' but same, any idea?

bgruening:
@ieguinoa please have a look, this now seems to work for me but I get:

```
galaxy.tool_util.verify.interactor.JobOutputsError: Output experiments_table_out:  different than expected
Expected 5 lines in output, found 2 lines
Output samples_table_out:  different than expected
Expected 5 lines in output, found 2 lines
Output runs_table_out:  different than expected
Expected 5 lines in output, found 2 lines
```

The interesting part in my commit is just the different test-syntax for such hugely nested forms.

ieguinoa:
the tests were using the assertions of the previous taste cases, didn't change it because it never worked for me.
this should fix it, at least  it works for me locally.
I think I should start adding a test on the command line, or something to test the not-dry-run case. The problem is if I add a test for it and there are no credentials defined it will always fail, is there a way to specify a negative case, where the tool should fail?

bgruening:
> will always fail, is there a way to specify a negative case, where the tool should fail?

https://github.com/galaxyproject/tools-iuc/blob/master/tools/idr_download/idr_download_by_ids.xml#L343

ieguinoa:
> > will always fail, is there a way to specify a negative case, where the tool should fail?
> 
> Yes this is possible :)
> https://github.com/galaxyproject/tools-iuc/blob/master/tools/idr_download/idr_download_by_ids.xml#L343

ok, I've used that to check for the failing case and also check the stdout for the report of it.
The way the tool is set now it will first check if the credentials are set in any way (preferences or global var to the secrets file), if not then it will fail before calling the cli tool. So, for the non-dry-case I can't go past the credentials check unless I actually set some credentials, do you know a way to create a file and set the ENA_GALAXY_SECRETS var in order to pass that check? if so then we can check that the command line generated is corect and also that the cli fails correctly.

bgruening:
>  So, for the non-dry-case I can't go past the credentials check unless I actually set some credentials, do you know a way to create a file and set the ENA_GALAXY_SECRETS var in order to pass that check? if so then we can check that the command line generated is corect and also that the cli fails correctly.

We can add a `<param type="hidden" value="false" />` parameter and set it to true in a test. When its set to true you can add some code to create the file. Does that sound worth it?

ieguinoa:
this should do the trick, I will add a test for the --vir case and a few details to further test the current cases .

bgruening:
@ieguinoa this is perfect thanks. Something went wrong with the branch. Can you please rebase against master? I think then we are ready for merge!

bgruening (approved):
Fantastic work! Let's get this to work!

bgruening (approved):

PR #3072, opened 2020-06-10
add lintr step to PR workflow

bernt-matthias:
also integrate latest changes (from https://github.com/galaxyproject/tools-devteam/pull/560) from devteam repo.

FOR CONTRIBUTOR:

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -172,6 +172,9 @@ jobs:
     name: Lint R scripts
     needs: setup
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        r-version: [4.0.1]
     steps:
     # checkout the repository to master
     # and use it as the current working directory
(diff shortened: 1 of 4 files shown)


.github/workflows/pr.yaml
```diff
@@ -172,13 +172,50 @@ jobs:
...
+        path: ../workflow_artifacts/
+    - uses: goanpeca/setup-miniconda@v1
+      with:
+        auto-update-conda: true
+        python-version: ${{ matrix.python-version }}
+        channels: conda-forge
+        use-only-tar-bz2: true
+    - name: Install lintr
+      shell: bash -l {0}
+      run: conda install r-lintr
```
nsoranzo, line 201: Possible alternative: https://github.com/r-lib/actions/tree/master/examples

bernt-matthias:
Seems to work -- also the caching runs if the lint step is successful. I think the last question is if the caching of the R packages is fine as it is. Maybe it should be versioned?

.github/workflows/pr.yaml
```diff
@@ -172,13 +168,63 @@ jobs:
...
+    - uses: actions/download-artifact@v2
+      with:
+        name: Workflow artifacts
+        path: ../workflow_artifacts/
+    - uses: r-lib/actions/setup-r@master
+    - name: Cache R packages
+      uses: actions/cache@v2
+      with:
+        path: ${{ env.R_LIBS_USER }}
+        key: lintr-rpackages
```
nsoranzo, line 190:
```suggestion
        key: r_cache_${{ matrix.r-version }}
```

bernt-matthias:
OK. Cool. Not sure about R package management. What happens if a new version of lintr or one of its requirements is available?

nsoranzo:
> OK. Cool. Not sure about R package management. What happens if a new version of lintr or one of its requirements is available?

As it is now, the present version of lintr will be used from the cache until we change the R version in the matrix. I don't think it is a big deal, we can change that later if needed.

PR #3077, opened 2020-06-15
bump lofreq

bgruening:
FOR CONTRIBUTOR:


tools/lofreq/lofreq_call.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="lofreq_call" name="Call variants" version="@TOOL_VERSION@+galaxy2">
+<tool id="lofreq_call" name="Call variants" version="@WRAPPER_VERSION@0">
```
mvdbeek, line 1:
```suggestion
<tool id="lofreq_call" name="Call variants" version="@WRAPPER_VERSION@+galaxy0">
```

tools/lofreq/lofreq_filter.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="lofreq_filter" name="Lofreq filter" version="@TOOL_VERSION@+galaxy1">
+<tool id="lofreq_filter" name="Lofreq filter" version="@WRAPPER_VERSION@0">
```
mvdbeek, line 1:
```suggestion
<tool id="lofreq_filter" name="Lofreq filter" version="@WRAPPER_VERSION@+galaxy0">
```

tools/lofreq/lofreq_viterbi.xml
```diff
@@ -1,11 +1,9 @@
-<tool id="lofreq_viterbi" name="Realign reads" version="@TOOL_VERSION@+galaxy1">
+<tool id="lofreq_viterbi" name="Realign reads" version="@WRAPPER_VERSION@0">
```
mvdbeek, line 1:
```suggestion
<tool id="lofreq_viterbi" name="Realign reads" version="@WRAPPER_VERSION@+galaxy0">
```

tools/lofreq/lofreq_viterbi.xml
```diff
@@ -1,11 +1,9 @@
...
     <description>with LoFreq viterbi</description>
     <macros>
         <import>macros.xml</import>
     </macros>
-    <expand macro="requirements">
-        <requirement type="package" version="1.9">samtools</requirement>
```
wm75, line 7: are your sure? I remember removing the requirement during a previous iteration. Had to restore it, but don't remember why right now.

mvdbeek:
I think you'll want to do a `lines_diff=2` on the bam output of viterbi.

tools/lofreq/lofreq_viterbi.xml
wm75: Ah, samtools got added to lofreq's bioconda recipe, I see.

tools/lofreq/lofreq_filter.xml
bgruening: Wrapper version is defined as: `<token name="@WRAPPER_VERSION@">@TOOL_VERSION@+galaxy</token>`

tools/lofreq/lofreq_viterbi.xml
bgruening: Mh, samtools is run-dep in the conda package. Tests are passing so I hope thisis ok?

tools/lofreq/test-data/indel-call-out.vcf
```diff
@@ -10,12 +10,12 @@
...
-pBR322	725	.	G	GN	0	.	;HRUN=1
-pBR322	746	.	A	AN	0	.	;HRUN=1
-pBR322	747	.	T	TN	0	.	;HRUN=1
-pBR322	751	.	G	GN	0	.	;HRUN=1
-pBR322	807	.	T	TN	0	.	;HRUN=3
-pBR322	943	.	C	CCG	0	.	;HRUN=1
-pBR322	1152	.	C	CN	0	.	;HRUN=2
-pBR322	1271	.	A	AN	0	.	;HRUN=2
-pBR322	1376	.	A	ANN	0	.	;HRUN=1
+pBR322	725	.	G	GN	0	.	DP=46;AF=0.021739;SB=0;DP4=22,23,0,1;INDEL;HRUN=1
```
bgruening, line 13: this is the bug that was fixed.

tools/lofreq/lofreq_filter.xml
mvdbeek: OK, thanks. Do you think that is better than `@TOOL_VERSION@+galaxyN` ? Now I'd have to look up the macro to understand the change, the previous convention is understandable on its own.

bgruening: I don't really mind. But we should have a standard way at some point - at least when we will have the update-bot all tools should follow one standard. I do not remember who has created this macro here, I just used it to stay consistent inside the tool.

mvdbeek: Having `+galaxy0` in there seems like a better handle than just an integer in that regard as well. I don't think changing this is a good idea before having established that we want to do that. https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tool-versions recommends the previous approach.

tools/lofreq/lofreq_viterbi.xml
wm75: I guess so, yes.

bgruening:
> I think you'll want to do a lines_diff="2" on the bam output of viterbi.

It's not even working with 4 lines diff :(

mvdbeek:
But then it does a line-by-line comparison.

mvdbeek:
Hmm, the BAM output is actually different, it looks like some (all?) sam tags are being reset.

Old output:
```
ACTTCCTCCCACTTCGGCCAACTT	163	pBR322	718	60	284M	=	1036	601	CGTCGCCGCACTTATGACTGTCTTCTTTATCATGCAACTCGTAGGACAGGTGCCGGCAGCGCTCTGGGTCATTTTCGGCGAGGACCGCTTTCGCTGGAGCGCGACGATGATCGGCCTGTCGCTTGCGGTATTCGGAATCTTGCACGCCCTCGCTCAAGCCTTCGTCACTGGTCCCGCCACCAAACGTTTCGGCGAGAAGCAGGCCATTATCGCCGGCATGGCGGCCGACGCGCTGGGCTACGTCTTGCTGGCGTTCGCGACGCGAGGCTGGATGGCCTTCCCCC	~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~	NM:i:1	MD:Z:283A0	MC:Z:283M	AS:i:283	XS:i:0	RG:Z:s0
```
New output:
```
ACTTCCTCCCACTTCGGCCAACTT	163	pBR322	718	60	284M	=	1036	601	CGTCGCCGCACTTATGACTGTCTTCTTTATCATGCAACTCGTAGGACAGGTGCCGGCAGCGCTCTGGGTCATTTTCGGCGAGGACCGCTTTCGCTGGAGCGCGACGATGATCGGCCTGTCGCTTGCGGTATTCGGAATCTTGCACGCCCTCGCTCAAGCCTTCGTCACTGGTCCCGCCACCAAACGTTTCGGCGAGAAGCAGGCCATTATCGCCGGCATGGCGGCCGACGCGCTGGGCTACGTCTTGCTGGCGTTCGCGACGCGAGGCTGGATGGCCTTCCCCC	~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~	XS:i:0	RG:Z:s0
```

mvdbeek:
That could be a problem for markdup ... worth reporting upstream ?

tools/lofreq/lofreq_filter.xml
nsoranzo: I don't particularly like the `<token name="@WRAPPER_VERSION@">@TOOL_VERSION@+galaxy</token>` macro, I think it's used only here and in chira.
The galaxy-iuc-standards leaves a choice between `@TOOL_VERSION@` and `@TOOL_VERSION@+galaxy0` for the first wrapper release of each version of the underlying tool. I normally use the former, which would rule out this macro, but I'm not sure if we want to update the standard to enforce one or the other.
One thing I'd recommend is to not have the `+galaxyN` part in a macro, but add it only on specific wrappers, so tools in a suite get bumped only when actually needed.

bgruening:
@mvdbeek sorry I have limited time atm to look at this. But please note that there is a "keepflags" option that strip/not-strips the flags.

tools/lofreq/lofreq_filter.xml
bgruening: Not defending here things that I have not done. But to be fair only `+galaxy` was part of the macro, the bump was happening per tool basis with the single integer. I guess the intention was to simply DRY and do not need to write `+galaxy` all the time when you go from the first version to a second.

nsoranzo: Yeah, I noticed, but I find it harder to spot without looking at the macro.

tools/lofreq/lofreq_viterbi.xml
```diff
@@ -20,7 +18,7 @@
         <param name="reads" type="data" format="bam" label="Reads to realign" />
         <expand macro="reference_interface" />
         <section name="adv_options" title="Advanced options" expanded="false">
-            <param argument="--keepflags" name="delflags" type="boolean" truevalue="" falsevalue="--keepflags" checked="true"
+            <param argument="--keepflags" type="boolean" truevalue="" falsevalue="--keepflags" checked="true"
```
mvdbeek, line 21: Since this will change the parameter name (`name` takes precedence over `argument` when persisting tool state) I think we should invert this. Because now you have the funny testcase where `keepflags=false` means that we set `--keepflags`, which is weird IMO.
```suggestion
            <param argument="--keepflags" type="boolean" truevalue="--keepflags" falsevalue="" checked="false"
```

tools/lofreq/lofreq_viterbi.xml
```diff
@@ -20,7 +18,7 @@
...
             label="Delete flags MC, MD, NM, and A?"
```
mvdbeek, line 22:
```suggestion
            label="Keep MC, MD, NM, and A flags?"
```

tools/lofreq/lofreq_viterbi.xml
```diff
@@ -20,7 +18,7 @@
...
             help="These flags are all prone to getting invalidated during realignment. Keep them only if you know what you are doing." />
```
mvdbeek, line 23:
```suggestion
            help="Re-alignment resets the MC, MD, NM, and A flags by default. The values of these tags depend on the exact alignment. Use this option only if you know what you are doing." />
```

wm75:
AFAICT, the viterbi tests are failing only because the bioconda package is using samtools 1.10 now (not 1.9 as previously). Apparently samtools sort 1.10 adds itself on a header `@PG` line, but has grown a `--no-PG` option if you want the old behavior.
So the simplest solution for making the tests pass again would be to add `--no-PG` to the `samtools sort` part of lofreq_viterbi.

tools/lofreq/lofreq_viterbi.xml
```diff
@@ -51,17 +49,18 @@
...
             <param name="ref" ftype="fasta" value="pBR322.fa" />
-            <output name="realigned" file="viterbi-out1.bam" />
+            <output name="realigned" file="viterbi-out1.bam" lines_diff="2" />
         </test>
         <test>
             <param name="reads" ftype="bam" value="lofreq-in1.bam" />
             <param name="ref_selector" value="history" />
             <section name="adv_options">
-                <param name="delflags" value="false" />
+                <param name="keepflags" value="false" />
```
mvdbeek, line 60:
```suggestion
                <param name="keepflags" value="true" />
```

mvdbeek:
> So the simplest solution for making the tests pass again would be to add `--no-PG` to the `samtools sort` part of lofreq_viterbi.

If that is the issue it should have been taken care of by `lines_diff`.

wm75:
You would think so, yes. What exactly is required to make `lines_diff` work with `bam`? The tests don't specify the output file type, for example. Would that be required for doing a line-based diff on a bam?

mvdbeek:
Ohh ... https://github.com/mvdbeek/galaxy/blob/f566c87b3cebfae314a79488ceb0b6f3e4d2924e/lib/galaxy/tool_util/verify/__init__.py#L125 we need to add ftype="bam" on the output for that to work

wm75:
The `--no-PG` option might still be preferable since the added @PG line is not helpful/misleading anyway.
It does not mention lofreq, but states that samtools sort was used after alignment.

mvdbeek:
We should probably also look at the output parameter type there ... .
`ftype="bam"` fixes it

What do you think about inverting the `--keepflags` parameter ?

> The `--no-PG` option might still be preferable since the added @pg line is not helpful/misleading anyway.
> It does not mention lofreq, but states that samtools sort was used after alignment.

With my job caching hat on I fully agree ... file system metadata (like paths) should not be part of the actual file. That would make comparing datasets by hash way more useful.

wm75:
> What do you think about inverting the `--keepflags` parameter ?

It's a good suggestion and I should probably have written it that way to begin with :)
What is a bit annoying now is that this would force users to revisit the parameters when upgrading an existing workflow to the new version.

mvdbeek:
> What is a bit annoying now is that this would force users to revisit the parameters when upgrading an existing workflow to the new version.

It would use the defaults though, since the names don't match up anymore. So I think that's OK ?

wm75:
One last thing: would you mind adding a `version_command`?
Not high-priority so we could also do that in a separate PR. (`lofreq version | grep version | cut -d ' ' -f 2` should work though I haven't tested it with 2.1.5.)

bgruening:
Cool, thanks.

PR #3082, opened 2020-06-17
Humann2 join

pravs3683:
FOR CONTRIBUTOR:


bgruening:
@pravs3683 can you explain this change and please also update the tool version.

pravs3683:
> @pravs3683 can you explain this change and please also update the tool version.

Join function in HUMAnN2 joins data from multiple samples (files). It uses filenames to represent those columns (from multiple samples). This update will help retain those original filenames instead of using arbitrary names generated by Galaxy.

In the existing version, the column names are reflected as Galaxy generated filename (eg. "dataset_21372519	dataset_21372520	dataset_21372521	dataset_21372522	dataset_21372524").

In this update, the joined output file will reflect column names as the original filenames.

pravs3683:
Hi Bjorn @bgruening , Can you please look into the PR. All checks are successful. Please see if it can be merged. Thanks!

tools/humann2/humann2_join_tables.xml
```diff
@@ -9,7 +9,7 @@
 mkdir tmp_dir
 &&
 #for $table in $input_table
-    cp '$table' 'tmp_dir' &&
+    cp '$table' 'tmp_dir/${table.name}' &&
```
bgruening, line 12: I think you should then use `table.element_identifier` here and probably need to replace whitespaces and special characters with `sed`, we do this in other tools.

pravs3683: Thank you @bgruening . I have modified it based on your suggestion. Couldn't figure out how to do that using sed but have used split and join to achieve this. If you still prefer me to use sed, can you please provide any example tool where this was used. I can take a look and try it accordingly. Thanks!

bgruening:
Thanks @pravs3683!

pravs3683:
@bgruening . Thank you for merging the changes.

humann2_join_tables tool is part of humann2 suite. Would this update automatically get installed on the usegalaxy.eu server.

I checked the iuc tools yaml file (https://github.com/usegalaxy-eu/usegalaxy-eu-tools/blob/master/tools_iuc.yaml), but didn't find humann2 listed in this file. Should I 
add humann2 here? If so, adding humann2 would include all the other tools in the suite or will I have to list all the tools within this suite.

Thanks!

bgruening:
@pravs3683 yes please add it and all other tools from this suite as well. Thanks.

PR #3089, opened 2020-06-18
snpEFF COVID-19 start ... this needs cleanups, but should work

bgruening:
FOR CONTRIBUTOR:


bgruening:
This tool is a fork of the real tool, just stripped down to only offer COVID-19 databases. This tool only supports Genomes shipped with the package, not *.loc file support, no history support.
@mvdbeek snpeff uses `NC_045512__.2__`. EU also uses this ID for the build-in genome which we have been built. Maybe it's wise to switch the ID in the workflow as well?

Todo:

* cleanup needed, many test files are not needed
* check help
* check if all params are still meaningful

wm75:
I've removed a bunch of snpeff options that don't make sense for viral genomes and I adjusted the selectable upstream/downstream region lengths to small genome sizes.
One could probably remove a couple of other options, but I tried to catch the obvious ones.

mvdbeek:
I'm a bit confused, do you mean the contig name with ID ? We've gone back and forth between `NC_045512` and `NC_045512.2`. `NC_045512.2` seemed more accurate but wasn't suported by snpsift in 4.3 :(

wm75:
Yes, it's about the version ID in the contig name. This version of SnpEff ships with a *NC_045512.2* genome (with version ID). So for annotations to succeed, input VCFs need to use this contig name, too.
Didn't know about the SnpSift problem until now.

mvdbeek:
I guess this was because the interal database was still generated as `NC_045512 `, so I assume it was fixed now ?

mvdbeek:
```
--------------------------------------
| 2 - SnpEff eff: on data 1 (HID - NAME) 
| Dataset State:
|  error
| Dataset Blurb:
|  tool error
| Dataset Info:
|  cannot find 'spliceSiteSize'
| Peek:
|  *Peek unavilable.*
| Dataset Job Standard Output:
|  *Standard output was empty.*
| Dataset Job Standard Error:
|  *Standard error was empty.*
|
```

wm75:
Ups, too courageous :)

wm75:
> I guess this was because the internal database was still generated as `NC_045512 `, so I assume it was fixed now ?

So you're talking about Snp*Eff*, not Snp*Sift* failing?

mvdbeek:
Possibly ...

wm75:
Yes, then, this got fixed in SnpEff as part of the covid19 release, too.

wm75:
I was just a bit worried that changing back to the contig name with version ID now, would create huge diffs somewhere in the covid19 project variant lists?

mvdbeek:
It would ... and it means we need to redo everything starting with the mapping step. The alternative would be to just add a step where we add the `.2` to the VCF files.

bgruening:
Is the tool id and toolshed name `snpeff_sars_cov_2` ok?

wm75:
LGTM :)

bgruening:
Thanks @wm75 and @mvdbeek!

PR #3098, opened 2020-06-27
Update shovill to set memory using GALAXY_MEMORY_MB

apetkau:
FOR CONTRIBUTOR:

This update sets the memory passed to shovill based on the `GALAXY_MEMORY_MB` environment variable (as described in the [planemo docs](https://planemo.readthedocs.io/en/latest/writing_advanced.html#cluster-usage)).

Previously this was being set via the `SHOVILL_RAM` environment variable. The changes I've made will still support this functionality. Specifically, the memory will be set in the following order:

1. If `SHOVILL_RAM` exists, use this value.
2. Else, derive memory from `GALAXY_MEMORY_MB`
3. Finally, default to 4 GB if `GALAXY_MEMORY_MB` is not set.


bgruening:
Merci!

PR #3103, opened 2020-07-01
Expose the --no-PG option of samtools reheader

wm75:
FOR CONTRIBUTOR:

### DIFF
--- a/tool_collections/samtools/samtools_reheader/samtools_reheader.xml
+++ b/tool_collections/samtools/samtools_reheader/samtools_reheader.xml
@@ -1,4 +1,4 @@
-<tool id="samtools_reheader" name="Reheader" version="@TOOL_VERSION@">
+<tool id="samtools_reheader" name="Reheader" version="2.0.1+galaxy1">
     <description>copy SAM/BAM header between datasets</description>
     <macros>
         <import>macros.xml</import>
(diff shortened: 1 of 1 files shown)


wm75:
The most controversial thing about this PR should be the proposed tool version change.
It's very unfortunate that this would *decrease* the version number :(
OTOH, with the upstream decision to move to 1.10 it may take a very long time until we can move to a sane versioning scheme again without sacrificing something.

tool_collections/samtools/samtools_reheader/samtools_reheader.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="samtools_reheader" name="Reheader" version="2.0.1">
+<tool id="samtools_reheader" name="Reheader" version="@TOOL_VERSION@">
```
mvdbeek, line 1: I'm afraid we'll have to stay with `2.0.1+galaxy1` ... otherwise `2.0.1`will stay the newest version.

wm75: Well, if we stick to the current scheme, then I would rather bump it to 2.0.2. It's not that this version scheme has a meaning outside Galaxy, is it?

mvdbeek: This at least gives us a theoretic chance that samtools catches up with our versioning scheme.

bgruening: :+1: for 2.0.1+galaxy1 ...

wm75: Alright, thanks for the feedback everyone :)

mvdbeek:
uploaded manually since the TTS is down.

PR #3113, opened 2020-07-07
PR deploy dependencies

bernt-matthias:
I also disabled testing for enasearch which made the CI fail for the last weeks.

Wondering if tools in `.tt_skip` and `.biocontainers_skip` really should be excluded from tests in the PR workflow. 

FOR CONTRIBUTOR:

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -374,7 +374,7 @@ jobs:
   # deploy the tools to the toolsheds (first TTS for testing)
   deploy:
     name: Deploy
-    needs: combine_outputs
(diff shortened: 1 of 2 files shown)


bernt-matthias:
Do you think that something like this might be nice https://github.com/marketplace/actions/gitter-notify .. currently it seems that the weekly CI failures are ignored by the community

peterjc:
I don't follow the logic behind your question. Tools in the skip lists have known failures, so most likely any PR changing them would fail too. If you are fixing a tool then the PR can also remove it from the skip list.

bgruening:
Thanks @bernt-matthias!

PR #3121, opened 2020-07-13
Update data manager with new columns, rename table due to collisions.

davebx:
FOR CONTRIBUTOR:

On usegalaxy.org and in `config/tool_data_table_conf.xml.sample`, the `indexed_maf_files` data table is defined with the standard four columns, but the `maf_index.loc` entries to which it points all should have five columns. This adds a new data table `maf_indexes` with the column definitions matching the expected file contents.


bernt-matthias:
Looks good. Would it be possible to add a test?

data_managers/data_manager_fetch_index_maf/data_manager/data_manager_fetch_and_index_maf.py
```diff
@@ -188,7 +188,7 @@ def main():
 
     data_manager_entry = {
         'data_tables': {
-            'indexed_maf_files': {
```
bernt-matthias, line 191: Why is renaming of the data table necessary? I'm not sure about side effects.

davebx: Because the existing data table on usegalaxy.org is defined with 4 columns, and when I used the same name for the new data table, it crashed the server.

PR #3141, opened 2020-07-23
Fastqe: escape element identifier

mblue9:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/fastqe/fastqe.xml
+++ b/tools/fastqe/fastqe.xml
@@ -1,13 +1,17 @@
-<tool id="fastqe" name="FASTQE" version="0.2.6">
+<tool id="fastqe" name="FASTQE" version="0.2.6+galaxy0">
     <description>visualize fastq files with emoji's 🧬😎</description>
     <requirements>
         <requirement type="package" version="0.2.6">fastqe</requirement>
+        <requirement type="package" version="4.7">sed</requirement>
     </requirements>
     <command detect_errors="exit_code">
     <![CDATA[
 
-    #for i in $input:
-        ln -s '$i' '$i.element_identifier' &&
+    #import re
+
+    #set $filenames = [re.sub('[^\w\-\s.]', '_', $i.element_identifier) for $i in $input]
+    #for $i, $filename in zip($input, $filenames):
+        ln -s '$i' '$filename' &&
     #end for
 
     echo "<html><head><title>FASTQE Report 🤔</title></head><body><h1>FASTQE Report 🤔</h1>" > '$output' &&
@@ -18,8 +22,8 @@
         --min
         --long $long
         --max
-        #for i in $input:
-            '$i.element_identifier'
+        #for $filename in $filenames:
+            '$filename'
         #end for
         | sed 's/$/<br>/;s/^\([^\t]*\)\t\([^\t]*\)\t/<h2>\1: \2<\/h2>\n/' >> '$output' &&
 
@@ -50,8 +54,6 @@ Scores can also be binned:
 +-------+-------+
 | Bin   | Emoji |
 +=======+=======+
-| N     | 🚫    |
-+-------+-------+
 | N     | 🚫    |
 +-------+-------+
 | 2-9   | 💀    |


hexylena:
Can you add a `+galaxy0` to the version since the behaviour is changing here?

mblue9:
Thanks @hexylena! I didn't update the wrapper version as I had thought of it as a bug fix but yes it does alter the output. Thanks for making the change @bgruening!
Do you know why the test is failing here or how I can see the what the error is?
In the log I just see `Unsuccessful tests found, inspect the 'All tool test results' artifact for details.`

bgruening:
You can see the error here https://github.com/galaxyproject/tools-iuc/pull/3141/checks?check_run_id=903659685 in the "planemo tool test". Seems that the test-data changed?

mblue9:
ah ok I think I understand, the output now contains text like _fq instead of .fq but I couldn't update the test output well as sed OSX works differently, @hexylena or someone are you able to update the test data using Linux or do you know what I need to do to make that sed command also OSX compatible & I can update the test data?

bgruening:
One thing to make it compatible with OSX is to include sed as a requirement for the tool.

hexylena:
image

we could permit `.` to fix the failing test.

If the sed command doesn't work on OSX, I'm sorry. I don't have OSX and can't test what sed commands are platform independent :( But bjoern is right, adding sed should help.

tools/fastqe/fastqe.xml
```diff
@@ -1,13 +1,17 @@
...
+        <requirement type="package" version="4.7">sed</requirement>
     </requirements>
     <command detect_errors="exit_code">
     <![CDATA[
 
+    #import re
+
     #for i in $input:
-        ln -s '$i' '$i.element_identifier' &&
+        #set $filename = re.sub('[^\w\-\s]', '_', str($i.element_identifier))
```
hexylena, line 13:
```suggestion
        #set $filename = re.sub('[^\w\-\s.]', '_', str($i.element_identifier))
```

tools/fastqe/fastqe.xml
```diff
@@ -19,7 +23,8 @@
         --long $long
         --max
         #for i in $input:
-            '$i.element_identifier'
+            #set $filename = re.sub('[^\w\-\s]', '_', str($i.element_identifier))
```
hexylena, line 26:
```suggestion
            #set $filename = re.sub('[^\w\-\s.]', '_', str($i.element_identifier))
```

mblue9:
Thanks for the sed tip @bgruening and looking into it @hexylena ! Yep adding sed as requirement seems to work great 👍

hexylena:
ah you've updated the test data that works too, no need to accept my changes then.

tools/fastqe/fastqe.xml
mblue9: Oh just saw this, should I allow the "." or let it change to underscore?

hexylena: Doesn't really matter to me. would `.`.be nice in the identifiers forusers?

mblue9: Maybe a bit nicer to see `.fastq.gz` than `_fastq_gz` but not sure if allowing "." is considered unsafe (ping @nsoranzo)

hexylena: it's safe, they're single quoted everywhere so no risk of directory traversal.

mblue9: Ok great, in that case I've allowed the `.`.

nsoranzo: Dots seem fine to me.

tools/fastqe/fastqe.xml
hexylena: this should match the other one, right? I honestly don't know how this passed tests, if it did `ln -s '...' 'test.fq'` here and then later `fastqe ... 'test_fq'`

nsoranzo: In a bit I am going to rewrite this to save the filenames to a list and not repeat ourselvers.

mblue9: Thanks @nsoranzo ! I meant to accept both @hexylena and I don't know how that passed tests either.

mblue9: Ah actually think I do know how it passed `.` needs to be escaped here?

nsoranzo:
The tests were working when they should have failed and viceversa, but a rebase fixed it (weird!)

tools/fastqe/fastqe.xml
nsoranzo: > Ah actually think I do know how it passed `.` needs to be escaped here?

No, according to https://docs.python.org/3/library/re.html#regular-expression-syntax , special characters lose their special meaning inside sets of characters.

mblue9:
Ok thanks for the info & merge @nsoranzo !

PR #3149, opened 2020-07-30
tradis essentiality fix?

bernt-matthias:
added `echo $?` .. so the tool should not fail anymore .. but it is .. don't get it

Isn't failing for non-containerized tests ... even without the `echo $?`. Maybe just a faulty container?

FOR CONTRIBUTOR:


bernt-matthias:
Problem was that R prints something on stderr which is interpretet as error if no stdio... is given. So I switched to agressive error checking.

PR #3150, opened 2020-08-03
IDBA: fix idba_tran and make testing more rigorous

pvanheus:
FOR CONTRIBUTOR:


pvanheus:
Two files are changed here:

1) idba_tran wrapper: this was looking for a `scaffold.fa` output whereas idba tran produces a `contig.fa` output. The result was a zero-length output dataset and this was not picked up by the test because it uses `sim_size` which, by default, means within 10000 bytes. The test dataset is < 1000 bytes so the difference between success and failure was not being noticed.

2) since the idba_ud wrapper also uses sim_size in its tests, so this was also made more rigorous. The tool version was not bumped because no change to the actual tool was performed.

tools/idba_ud/idba_ud.xml
```diff
@@ -99,7 +99,7 @@
                 <has_text text="--min_support 2" />
                 <has_text text="--seed_kmer 29" />
             </assert_command>
-            <output name="output" file="out/scaffold.fa" compare="sim_size"/>
+            <output name="output" file="out/scaffold.fa" compare="sim_size" delta="10"/>
```
bernt-matthias, line 102: Do we need a delta at all? Maybe even compare="diff" works?

pvanheus: Yes it does, and switching to "diff" highlighted that the saved output was wrong.

bernt-matthias: What is the content of the differing line?

PR #3154, opened 2020-08-04
do not skip tools in .tt_biocontainer_skip but test un-containerized

bernt-matthias:
found it a good idea that tools in `.tt_biocontainer_skip` are tested at least using conda env instead of no test at all. .. not sure if implemented correctly. otherwise it makes not much sense to have two .tt_.. files.

FOR CONTRIBUTOR:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -81,6 +81,11 @@ jobs:
                else
                    TESTPATH="$DIR"
                fi
+               if grep -lqf .tt_biocontainer_skip <(echo $DIR); then
+                   PLANEMO_OPTIONS=""
+               else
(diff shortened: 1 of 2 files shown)


tools/jq/jq.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<tool id="jq" name="JQ" version="1.0">
+<tool id="jq" name="JQ" version="1.0+galaxy1">
```
bgruening, line 2: @bernt-matthias why is this bumped and maker as well?

bernt-matthias: Just for demonstration purposes. Just drop the commit before merging...

bernt-matthias: Dropped.

bgruening:
Merci!

PR #3158, opened 2020-08-11
Busco 4.1.2

abretaud:
FOR CONTRIBUTOR:

So here's an update of the busco tool, from 3.0.2 to 4.1.2
The main change is that busco now downloads the lineage dataset at runtime (from https://busco-data.ezlab.org/v4/data/lineages/). I've removed the possibility to select datasets from the busco data table: it would be quite a mess to maintain it, with little added value I think:
 - the lineage dataset format changed between busco v3 and v4, but the data table is currently not versioned
 - the download urls changed with the v4 release, breaking the current busco data manager


pvanheus:
This is very useful! There is a minor issue of support for people whose compute nodes don't have Internet access (because `--lineage_dataset` downloads from the Internet), but I don't think that that should block that PR. I think the correct solution for that case is to update the data manager that fetches BUSCO files and add a DM-supported offline mode to this tool.

Slugger70:
It looks as though busco will download the lineage data to `./busco_downloads` (according to it's [config file](https://gitlab.com/ezlab/busco/-/blob/master/config/config.ini) - line 46) and for Galaxy this would be in the job working directory. This also means that it would download it every time it ran. Would it be better to modify the `config.ini` file in Busco to point to `galaxy_data_manager_data_path` so that it only gets downloaded once?

abretaud:
Why not, but do we have access to `galaxy_data_manager_data_path` from the tool command line?
I wonder what happens if multiple jobs download the same lineage to the same dir

bgruening:
> Why not, but do we have access to galaxy_data_manager_data_path from the tool command line?

No, this can not be assumed. Are the files that large that we need to worry?

Slugger70:
@bgruening I don't think so. I reckon we should approve and merge the PR in the short term for people who need it now and work on an updated DM to help with Peter's issue.

bgruening:
Ok, than let's get this in.

PR #3177, opened 2020-08-24
Updating medaka wrappers (and tests) to include annotation script in variant caller and allow for older versions of bash

NatPRoach:
FOR CONTRIBUTOR:

Updating medaka to include a script I wrote that annotates called variants with allele frequency and strand bias etc annotated in the output VCF. Modifying tests to account for this and reduce size so that these tests can fit in the test-data directory.
Also removing `|&` uses in the wrappers as this fails for older versions of bash, replaced with backwards compatible equivalent `2>&1 |`.


tools/medaka/medaka_consensus.xml
```diff
@@ -96,24 +98,29 @@ medaka_consensus
...
                     <has_line line=">4 length=3792 depth=114.52x circular=true"/>
-                    <has_line line="CATCTCTTT"/>
+                    <has_line line="ATCTCTTT"/>
                 </assert_contents>
             </output>
             <output name="out_probs">
                 <assert_contents>
-                    <has_size value="192710" delta="100"/>
+                    <has_size value="166251" delta="100"/>
+                    <!-- <has_size value="192710" delta="100"/> -->
```
bernt-matthias, line 107: You may remove the comment.

tools/medaka/variant.xml
```diff
@@ -25,62 +25,85 @@ reference.fa
...
         <param argument="inputs" type="data" format="h5" multiple="true" label="Select consensus file(s)"/>
         <expand macro="reference"/>
-        <param argument="--region" type="text" value="" optional="true" label="Set reference names to limit variant calling" help="Separated by ','."/>
-        <param argument="--verbose" type="boolean" truevalue="--verbose" falsevalue="" label="Populate VCF info fields?"/>
-        <param name="out" type="select" multiple="true" optional="false" label="Select output file(s)">
-            <option value="result" selected="true">Result</option>
-            <option value="log">Log</option>
+        <param argument="--region" type="text" value="" optional="true" label="Set reference names to limit variant calling" help="Separated by ','.">
+            <sanitizer invalid_char="">
+                <valid initial="string.ascii_letters,string.digits">
```
bernt-matthias, line 39: Maybe you should remove `'` and `"` otherwise you may end up in quoting hell :)

tools/medaka/variant.xml
```diff
@@ -25,62 +25,85 @@ reference.fa
...
+            <param name="output_annotated_select" type="select" label = "Output annotated VCF?">
+                <option value="true" selected="true">Output annotated VCF</option>
+                <option value="false">Don't output annotated VCF</option>
+            </param>
+            <when value="true">
+                <param name="in_mpileup" type = "data" format="pileup" optional="false" label="Mpileup to annotate the VCF"/>
+            </when>
+        </conditional>
+
+        <conditional name="output_log">
```
bernt-matthias, line 58: no need for the conditional here.

tools/medaka/variant.xml
NatPRoach: Do you mean in the sanitizer? I thought I already had? Unless you mean in the help statement?

bernt-matthias: Yep.

NatPRoach: I think the sanitizer already does remove `'` and `"`? The initial setup only includes ascii letters and digits, not `'` and `"`.

bernt-matthias: I see. Then its fine

tools/medaka/variant.xml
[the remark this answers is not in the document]
NatPRoach: Woops, yeah the conditional was a mistake. I will fold into conditional above if you think I should. I left it as a boolean for now but I can change that if you think it'd be better with just a single selector for output.

NatPRoach: Fixed in latest commit. Thanks for pointing that out.

tools/medaka/medaka_consensus.xml
NatPRoach: Done in latest commit.

bgruening (approved):

bgruening:
@NatPRoach thanks!

PR #3186, opened 2020-09-07
set the format of chain file to txt instead of csv

drosofff:
FOR CONTRIBUTOR:

I cannot figure out why the asked format for the LiftOver chain file is `cvs` nor why the tests are indeed passing. I don't see either the rational since the LiftOver chain file are most often kind of tabular-like and for sure not cvs.
Anyway this tools in its present version crashes with the test data on several artbio instances under 20.01, whereas it works with the input format `txt`, and importantly this does not affect the current tests (not surprisingly)

### DIFF
--- a/tools/crossmap/macros.xml
+++ b/tools/crossmap/macros.xml
@@ -7,7 +7,7 @@
         </requirements>
     </xml>
     <token name="@TOOL_VERSION@">0.3.7</token>
-    <token name="@WRAPPER_VERSION@">@TOOL_VERSION@</token>
+    <token name="@WRAPPER_VERSION@">@TOOL_VERSION@_galaxy_1</token>
     <xml name="stdio">
         <stdio>
             <regex match="Aborted (core dumped)" source="stdout" level="fatal"/>
@@ -34,7 +34,7 @@ CrossMap.py 2>&1 | head -n 1 | grep -E --only-matching 'CrossMap.*'
                 </param>
             </when>
             <when value="history">
-                <param name="input_chain" type="data" format="csv" label="LiftOver chain file"/>
+                <param name="input_chain" type="data" format="txt" label="LiftOver chain file"/>
             </when>
         </conditional>
     </xml>
--- a/tools/crossmap/macros.xml
+++ b/tools/crossmap/macros.xml
@@ -7,7 +7,7 @@
         </requirements>
     </xml>
     <token name="@TOOL_VERSION@">0.3.7</token>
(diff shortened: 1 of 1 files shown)


tools/crossmap/macros.xml
```diff
@@ -7,7 +7,7 @@
         </requirements>
     </xml>
     <token name="@TOOL_VERSION@">0.3.7</token>
-    <token name="@WRAPPER_VERSION@">@TOOL_VERSION@</token>
+    <token name="@WRAPPER_VERSION@">@TOOL_VERSION@_galaxy_1</token>
```
bgruening, line 10: Its is just `galaxy1` ...

tools/crossmap/macros.xml
```diff
@@ -34,7 +34,7 @@ CrossMap.py 2>&1 | head -n 1 | grep -E --only-matching 'CrossMap.*'
                 </param>
             </when>
             <when value="history">
-                <param name="input_chain" type="data" format="csv" label="LiftOver chain file"/>
+                <param name="input_chain" type="data" format="txt" label="LiftOver chain file"/>
```
bgruening, line 37: mh, those files are usually `tabular` isn't it?

drosofff: Well, if you upload this files from UCSC, they are sniffed by Galaxy as tabular. However, strictly speaking, a [chain file](https://genome.ucsc.edu/goldenPath/help/chain.html) is a succession of "header" lines without tabulation followed by blocks of lines with two tabulations and finally an empty line. Thus, I would not say "tabular" that fast.

I bet it would also work with `tabular` as input. But in the doubt, I put `txt`. In any case `CrossMap.py` does not seem to check the format.

drosofff: So, did I get the point @bgruening or not ?

bgruening: Yes, ok. But the above is blocking and I can not push to your branch.

drosofff: hoops sorry

bgruening:
merci!

drosofff:
@bgruening it seems that the upgrade was not pushed to the toolshed https://github.com/galaxyproject/tools-iuc/runs/1159991614. Is there a way to fix this ?

nsoranzo:
@drosofff I've restarted the build (https://github.com/galaxyproject/tools-iuc/runs/1188074938 actually) and this time it was uploaded fine, thanks for noticing!

drosofff:
Thank a lot @nsoranzo

PR #3193, opened 2020-09-16
skip containerized tests for bwameth

bernt-matthias:
failing currently with `samtools: error while loading shared libraries: libcrypto.so.1.0.0: cannot open shared object file: No such file or directory`

maybe the proper solution would be to rebuild the container? 

FOR CONTRIBUTOR:

### DIFF
--- a/tools/bwameth/bwameth.xml
+++ b/tools/bwameth/bwameth.xml
@@ -1,4 +1,4 @@
-<tool id="bwameth" name="bwameth" version="@TOOL_VERSION@" profile="17.01">
+<tool id="bwameth" name="bwameth" version="@TOOL_VERSION@+galaxy1" profile="17.01">
     <description>Fast and accurate aligner of BS-Seq reads.</description>
     <macros>
         <token name="@TOOL_VERSION@">0.2.2</token>
--- a/tools/lofreq/lofreq_viterbi.xml
+++ b/tools/lofreq/lofreq_viterbi.xml
@@ -1,4 +1,4 @@
-<tool id="lofreq_viterbi" name="Realign reads" version="@TOOL_VERSION@+galaxy0">
+<tool id="lofreq_viterbi" name="Realign reads" version="@TOOL_VERSION@+galaxy1">
     <description>with LoFreq viterbi</description>
     <macros>
         <import>macros.xml</import>


mvdbeek:
Correct, I just wonder what happened there ? We must have had a container that worked previously?

mvdbeek:
(And I've seen the lofreq error too, and I'm sure we had a previously working version of the container)

bernt-matthias:
Seems that libcrypto causes some problems with samtools and other packages https://github.com/bioconda/bioconda-recipes/issues/17212

In another case I have seen that an older version of samtools was used that hasn't been bumped for a while.

I'm surprised to see that containers break over time.

bgruening:
Its probably not the container that breaks, but new broken-builds that have been uploaded in the meantime?

mvdbeek:
Yeah, and the Galaxy will use the most recently built container.

mvdbeek:
fwiw, `quay.io/biocontainers/lofreq:2.1.5--py38h1bd3507_2` is definitely broken

bgruening:
> fwiw, quay.io/biocontainers/lofreq:2.1.5--py38h1bd3507_2 is definitely broken

I can take it offline ... but the conda package still needs to be fixed I guess. However, I do not know why the conda tests have not catched that during the build.

mvdbeek:
I guess they didn't call `samtools` at all.
quay.io/biocontainers/lofreq:2.1.5--py37h3a01142_1 is working fine, which should be the next container in the resolution order.

bgruening:
renamed the container, should work again ...

bernt-matthias:
Is bioconda calling the tests only for the main package or also those of depending packages (during container building)?

mvdbeek:
Probably just the main package, descending into all dependencies seems like a difficult-ish task?

quay.io/biocontainers/bwameth:0.2.2--pyh864c0ab_2 and quay.io/biocontainers/bwameth:0.2.2--pyh864c0ab_3 are also broken in the same way, if we rename/remove these containers I suppose bwameth is going to work too

bernt-matthias:
Currently bwameth still fails

mvdbeek:
Yep, the containers are still there: https://quay.io/repository/biocontainers/bwameth?tag=0.2.2&tab=tags

bernt-matthias:
> quay.io/biocontainers/bwameth:0.2.2--pyh864c0ab_2 and quay.io/biocontainers/bwameth:0.2.2--pyh864c0ab_3 are also broken in the same way, if we rename/remove these containers I suppose bwameth is going to work too

@bgruening could you remove these containers? I guess then we are done here.

bernt-matthias:
I guess also unicycler is a case: 

In the output I see: `samtools        ?                 too old` and `Error: Unspecified error with Unicycler dependencies`

bgruening:
> > quay.io/biocontainers/bwameth:0.2.2--pyh864c0ab_2 and quay.io/biocontainers/bwameth:0.2.2--pyh864c0ab_3 are also broken in the same way, if we rename/remove these containers I suppose bwameth is going to work too
> 
> @bgruening could you remove these containers? I guess then we are done here.

I removed them. But can you please trigger a new build on Bioconda. The changes to _1 and _2 are maybe useful.

bernt-matthias:
Yes, I can do this. Tests are working now. 

Wondering if bumping the recipes is really helping? If I get it right there was some broken package (samtools or one of its dependencies) installed in the container. If this is still broken then it won't help. How could we test this / avoid this...

Maybe I can start a discussion at bioconda if there is a way to trigger the tests of all packages that are installed.

bgruening:
@bernt-matthias we can and should for samtools add a simple test to those recipes `- samtools --help`

bernt-matthias:
So I guess we can close this one? Or should I rerun the tests again to check the new containers.

bgruening:
Let's check the new ones again. Danke Matthias, richtig cool!

bernt-matthias:
If I get it right the new containers are working.

PR #3204, opened 2020-09-23
do not use >1 threads for velveth in velveloptimizer

bernt-matthias:
velvetoptimizer runs with `-t` already multithreaded (one for each k).
with `OMP_NUM_THREADS` > 1 also each velveth run will be multithreaded.

so overall CPUs are overutilized.

removed `export OMP_THREAD_LIMIT=1` because otherwise the tool gets
stuck.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/velvet_optimiser/velvetoptimiser.xml
+++ b/tools/velvet_optimiser/velvetoptimiser.xml
@@ -6,7 +6,8 @@
     </requirements>
     <version_command>VelvetOptimiser.pl --version</version_command>
     <command detect_errors="exit_code"><![CDATA[
-        export OMP_NUM_THREADS=$advanced.velveth_threads &&
+        export OMP_NUM_THREADS=1 &&
+        ## export OMP_THREAD_LIMIT=1 &&
         VelvetOptimiser.pl
             -t "\${GALAXY_SLOTS:-1}"
             -s $start_kmer
@@ -110,7 +111,6 @@
             <param name="velvetg_opts" type="text" value="" label="Other velvetg options" help="Add any other required velvetg options from the advanced set"/>
             <param name="minCutoff" type="integer" value="0" label="Minimum coverage cutoff" help="The minimum coverage cutoff to consider in the optimisation"/>
             <param name="maxCutoff" type="float" value="0.8" label="Maximum coverage cutoff" help="The maximum coverage cutoff to consider expressed as a fraction of the calculated expected coverage."/>
-            <param name="velveth_threads" type="integer" value="1" min="1" label="Number of threads used by each velvet run" help="vetvetoptimizer is running multiple velvet calls in parallel. By using values greater one for this option also the separate velvet runs will run multithreaded. Note that this will lead to overutilization of the available CPUs."/>
         </section>
     </inputs>
 
--- a/tools/velvet_optimiser/velvetoptimiser.xml
+++ b/tools/velvet_optimiser/velvetoptimiser.xml
@@ -6,8 +6,7 @@
     </requirements>
     <version_command>VelvetOptimiser.pl --version</version_command>
     <command detect_errors="exit_code"><![CDATA[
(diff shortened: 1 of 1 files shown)


tools/velvet_optimiser/velvetoptimiser.xml
```diff
@@ -1,13 +1,13 @@
...
     <description>Automatically optimize Velvet assemblies</description>
     <requirements>
         <requirement type="package" version="1.2.10">velvet</requirement>
         <requirement type="package" version="2.2.6">perl-velvetoptimiser</requirement>
     </requirements>
     <version_command>VelvetOptimiser.pl --version</version_command>
     <command detect_errors="exit_code"><![CDATA[
-        export OMP_NUM_THREADS=2 &&
-        export OMP_THREAD_LIMIT=2 &&
+        export OMP_NUM_THREADS=1 &&
```
mvdbeek, line 9: Shouldn't this be `"\${GALAXY_SLOTS:-1}" or "\${OMP_NUM_THREADS:-1}" ? Don't know if GALAXY_SLOTS is a good proxy, but ideally admins can override the default.

bernt-matthias: There are two levels of parallelization here. 

1. velveth (which is called internally): the number of threads is controlled via environment variable  `OMP_NUM_THREADS` that needs to be set (I recall that not setting it leads to using max number of CPUs)
2. VelvetOptimiser starts multiple threads (one per value of k) which is controlled via `-t`

Previously both levels of parallelization were used. Leading to overutilization by a factor of 2.

So we can either 

- use `-t "${GALAXY_SLOTS:-1}"` or and set `OMP_NUM_THREADS` to 1 OR
- the opposite (i.e. `-t 1` and `OMP_NUM_THREADS="${GALAXY_SLOTS:-1}"`)

mvdbeek: I see. I guess in this case it makes more sense to do what you're doing, but maybe make the `OMP_NUM_THREADS ` / `OMP_THREAD_LIMIT ` configurable with a default of 1 ? And maybe add this as a comment, that may come in handy in the future.

bernt-matthias: Comment is no problem.

Configurable by the user? Or do you have something else in mind?

Slugger70: Hey @bernt-matthias, I actually use the OMP_NUM_THREADS when we use VelvetOptimiser (though we only use it for training as it is a very old assembler with much better and newer ones available). So I agree with Marius, I'd prefer it to be configurable by the user.

bernt-matthias: OK. Did so as an advanced option.

bgruening:
Offering this to a user is not a good idea imho. No one knows what to do with that, isn't it?

bernt-matthias:
> Offering this to a user is not a good idea imho.

I also think so. I think we should decide for one of the two or we need to balance, e.g. using two threads for each velvet call and use `-t GALAXY_SLOTS / 2`

tools/velvet_optimiser/velvetoptimiser.xml
bernt-matthias: @Slugger70 would it be fine with you to fix `OMP_NUM_THREADS` to one. You would still be able to use multiple threads (i.e. multiple parallel runs of velvet). 

Alternatively I would suggest that I add a check if OMP_NUM_THREADS is set and if not set it to one. Thereby the admin would be able to set it via an environment variable.

bernt-matthias:
Anyone wants to decide on this one?

bgruening (approved):

bernt-matthias:
I'm going to merge this if nobody objects

PR #3215, opened 2020-09-28
Add vsnp_determine_ref_from_data

gregvonkuster:
FOR CONTRIBUTOR:


gregvonkuster:
I know this test file ```tools/vsnp/test-data/Mcap_Deer_DE_SRR650221.fastq.gz ``` is larger than what is generally accepted in this repo.  This tool sniffs input datasets for "dna footprints" that are contained in reference file like this, and I've filtered out as much data as possible from the original file, leaving just enough data for the footprints to be determined.

gregvonkuster:
Hi all, I'm hoping this can be reviewed and merged more quickly.  There are several more tools in this suite that I will submit as each tool is merged.  We have a paper coming out by the end of the year (hopefully) and we'd like to have the entire suite merged and available in preparation for the publication.  Please let me know if there are any changes needed to this PR and I'll commit them quickly.  Thanks!

bgruening:
Its on my todo list, sorry ... :(

@galaxyproject/iuc can anyone maybe help here?

gregvonkuster:
Thanks @bgruening - it's no problem - I know how busy you are!  I've enhanced the tool to now properly handle collections of paired reads in addition to single and paired reads and collections of single reads.  So the tool is ready for a review when possible.

tools/vsnp/vsnp_determine_ref_from_data.xml
```diff
@@ -0,0 +1,213 @@
...
+    </tests>
+    <help>
+**What it does**
+
+Accepts a single fastqsanger read, a set of paired reads, or a collection of reads and inspects the data to discover the
+best reference genome for aligning the reads.  This tool is, in essence, a DNA sniffer, and is the first Galaxy tool to
+perform this task.  While inspecting the data, a string of 0's and 1's is compiled based on the data contents, and we call
+the complete string a "DNA print".  All of the "DNA prints" files installed by the complementary **vSNP DNAprints data
+manager** tool are then inspected to find a match for the compiled "DNA print" string.  These files are each associated
+with a Galaxy "dbkey" (i.e., genome build), so when a metach is found, the associated "dbkey" is passed to a mapper (e.g.,
```
bernt-matthias, line 171: what's a metach?

tools/vsnp/vsnp_determine_ref_from_data.xml
```diff
@@ -0,0 +1,213 @@
...
+**Required Options**
+
+ * **Choose the category of the files to be analyzed** - select "Single files" or "Collection of files", then select the appropriate history items (single or paired fastqsanger reads or a collection of fastqsanger reads) based on the selected option.
+ * **Number of processes for job splitting** - Select the number of processes for splitting the job to shorten execution time.
+    </help>
+    <citations>
+        <citation type="bibtex">
+            @misc{None,
+            journal = {None},
+            author = {1. Stuber T},
```
bernt-matthias, line 206: Author list seems odd. If in preparation there should be a working title? Maybe bioarxive?

tools/vsnp/vsnp_determine_ref_from_data.xml
```diff
@@ -0,0 +1,213 @@
...
+                <element name="forward.txt" file="forward_metrics.txt" ftype="txt"/>
+                <element name="reverse.txt" file="reverse_metrics.txt" ftype="txt"/>
+            </output_collection>
+        </test>
+    </tests>
+    <help>
+**What it does**
+
+Accepts a single fastqsanger read, a set of paired reads, or a collection of reads and inspects the data to discover the
+best reference genome for aligning the reads.  This tool is, in essence, a DNA sniffer, and is the first Galaxy tool to
```
bernt-matthias, line 167: I don't understand the description. What are the positions in the strings corresponding to and what are 0 and 1 corresponding to?

tools/vsnp/vsnp_determine_ref_from_data.xml
```diff
@@ -0,0 +1,213 @@
...
+                    <when value="single_reads">
+                        <param name="reads_collection" type="data_collection" format="fastqsanger,fastqsanger.gz" collection_type="list" label="Collection of fastqsanger files"/>
+                    </when>
+                    <when value="paired_reads">
+                        <param name="reads_collection" type="data_collection" format="fastqsanger,fastqsanger.gz" collection_type="paired" label="Collection of fastqsanger paired read files"/>
+                    </when>
+                </conditional>
+            </when>
+        </conditional>
+        <param name="processes" type="integer" min="1" max="20" value="8" label="Number of processes for job splitting"/>
```
bernt-matthias, line 115: This should not be controllable by the user! Use `${GALAXY_SLOTS:-1}` in the command section.

tools/vsnp/vsnp_determine_ref_from_data.xml
```diff
@@ -0,0 +1,213 @@
...
+    <requirements>
+        <requirement type="package" version="1.76">biopython</requirement>
+        <requirement type="package" version="5.3">pyyaml</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+#import os
+#import re
+#set gzipped = 'false'
+#set input_type = $input_type_cond.input_type
+#set input_reads_dir = 'input_reads'
```
bernt-matthias, line 12: I think the variables  for the dir names are superfluous, you can use the string directly.

tools/vsnp/vsnp_determine_ref_from_data.xml
```diff
@@ -0,0 +1,213 @@
...
+#set output_dbkey_dir = 'output_dbkey'
+#set output_metrics_dir = 'output_metrics'
+mkdir -p $input_reads_dir &&
+mkdir -p $output_dbkey_dir &&
+mkdir -p $output_metrics_dir &&
+#if str($input_type) == "single":
```
bernt-matthias, line 18: Should be `$input_type_cond.input_type`

tools/vsnp/vsnp_determine_ref_from_data.xml
```diff
@@ -0,0 +1,213 @@
...
+        #if str($collection_type) == 'single_reads':
+            #set identifier = re.sub('[^\s\w\-]', '_', str($i.element_identifier))
+        #else:
+            ## Galaxy builds lists of pairs as nested lists with elements
+            ## named forward and reverse.  When flattened, these lists
+            ## will work as inputs to the Parse parameter value expression
+            ## tool in workflows.  However, the output list created by the
+            ## expression tool will not function correctly with the bwa_mem
+            ## mapper.  Naming the identifier as follows is a solution.
+            #set identifier = re.sub('[^\s\w\-]', '_', str($i.name))
```
bernt-matthias, line 52: shouldn't there be two symlinks in case paired data?

bernt-matthias: Do not understand the comment. Why does it work for gzipped data?

tools/vsnp/vsnp_determine_ref_from_data.xml
```diff
@@ -0,0 +1,213 @@
...
+    #else:
+        --read1 '${read1_identifier}'
+        --read2 '${read2_identifier}'
+    #end if
+    --output_dbkey '$output_dbkey'
+    --output_metrics '$output_metrics'
+#end if
+--gzipped $gzipped
+--processes $processes
+#if str($in_test_mode) == "false":
```
bernt-matthias, line 70: Do not like the test mode. If I get it right you can avoid it by using a test data table?

tools/vsnp/vsnp_determine_ref_from_data.py
```diff
@@ -0,0 +1,364 @@
...
+            tup = task_queue.get(block=True, timeout=timeout)
+        except queue.Empty:
+            break
+        fastq_file, count_list, brucella_string, brucella_sum, bovis_string, bovis_sum, para_string, para_sum = tup
+        group, dbkey = get_group_and_dbkey(dnaprints_dict, brucella_string, brucella_sum, bovis_string, bovis_sum, para_string, para_sum)
+        finished_queue.put((fastq_file, count_list, group, dbkey))
+        task_queue.task_done()
+
+
+def get_oligo_dict():
```
bernt-matthias, line 106: Why not just define: 

```
OLIGO_DICT = dict(("01_ab1", "AATTGTCGGATAGCCTGGCGATAACGACGC"), ....)
```

tools/vsnp/vsnp_determine_ref_from_data.py
```diff
@@ -0,0 +1,364 @@
...
+        num_files = len(fastq_list)
+        cpus = set_num_cpus(num_files, args.processes)
+        # Set a timeout for get()s in the queue.
+        timeout = 0.05
+
+        for fastq_file in fastq_list:
+            queue1.put(fastq_file)
+
+        # Complete the get_species_counts task.
+        processes = [multiprocessing.Process(target=get_species_counts_for_collection, args=(queue1, queue2, args.gzipped, timeout, )) for _ in range(cpus)]
```
bernt-matthias, line 322: Having these code blocks in a function would increase readability,

tools/vsnp/vsnp_determine_ref_from_data.py
```diff
@@ -0,0 +1,364 @@
...
+
+        # Complete the output_files task.
+        processes = [multiprocessing.Process(target=output_files_for_collection, args=(queue2, timeout, )) for _ in range(cpus)]
+        for p in processes:
+            p.start()
+        for p in processes:
+            p.join()
+        queue2.join()
+
+        if queue1.empty() and queue2.empty():
```
bernt-matthias, line 353: what happens else?

tools/vsnp/vsnp_determine_ref_from_data.py
```diff
@@ -0,0 +1,364 @@
...
+            fh.write("%d," % i)
+        fh.write("\nGroup: %s" % group)
+        fh.write("\ndbkey: %s\n" % dbkey)
+
+
+def set_num_cpus(num_files, processes):
+    num_cpus = int(multiprocessing.cpu_count())
+    if num_files < num_cpus and num_files < processes:
+        return num_files
+    if num_cpus < processes:
```
bernt-matthias, line 262: Doesn't this potentially lead to CPU overutilization?

tools/vsnp/vsnp_determine_ref_from_data.xml
```diff
@@ -0,0 +1,213 @@
...
+                <conditional name="collection_type_cond">
+                    <param name="collection_type" type="select" label="Collection of single reads or paired reads?">
+                        <option value="single_reads" selected="true">Single reads</option>
+                        <option value="paired_reads">Paired reads</option>
+                    </param>
+                    <when value="single_reads">
+                        <param name="reads_collection" type="data_collection" format="fastqsanger,fastqsanger.gz" collection_type="list" label="Collection of fastqsanger files"/>
+                    </when>
+                    <when value="paired_reads">
+                        <param name="reads_collection" type="data_collection" format="fastqsanger,fastqsanger.gz" collection_type="paired" label="Collection of fastqsanger paired read files"/>
```
bernt-matthias, line 110: Doesn't this only allow a single pair? I guess you want `list:paired`, or?

tools/vsnp/vsnp_determine_ref_from_data.xml
```diff
@@ -0,0 +1,213 @@
...
+            <filter>input_type_cond['input_type'] == 'collection'</filter>
+        </collection>
+        <collection name="output_metrics_collection" type="list" label="${tool.name} (metrics) on ${on_string}">
+            <discover_datasets pattern="__name__" directory="output_metrics" format="txt"/>
+            <filter>input_type_cond['input_type'] == 'collection'</filter>
+        </collection>
+    </outputs>
+    <tests>
+        <test>
+            <param name="in_test_mode" value="true"/>
```
bernt-matthias, line 137: Would like to see tests for all four cases.

tools/vsnp/vsnp_determine_ref_from_data.xml
```diff
@@ -0,0 +1,213 @@
...
+            <discover_datasets pattern="__name__" directory="output_dbkey" format="txt"/>
+            <filter>input_type_cond['input_type'] == 'collection'</filter>
+        </collection>
+        <collection name="output_metrics_collection" type="list" label="${tool.name} (metrics) on ${on_string}">
+            <discover_datasets pattern="__name__" directory="output_metrics" format="txt"/>
+            <filter>input_type_cond['input_type'] == 'collection'</filter>
+        </collection>
+    </outputs>
+    <tests>
+        <test>
```
bernt-matthias, line 136: Please add `axpect_num_outputs` to all tests.

tools/vsnp/vsnp_build_tables.py
```diff
@@ -3,10 +3,11 @@
 import argparse
```
bernt-matthias, line 3: I don't understand why these tools do not contain one of the scripts in https://github.com/USDA-VS/vSNP .. or the other way round why the python scripts here are not in https://github.com/USDA-VS/vSNP?

bernt-matthias: Seems quite a bit of work to review the tool itself .. timewise .. but as you suggest to add it to IUC we should. Can you make its easier for us? Maybe by adding comments and doc strings.

bernt-matthias (changes requested):

tools/vsnp/vsnp_build_tables.py
gregvonkuster: Thanks so much for the review @bernt-matthias.  I totally agree that the iuc process for reviewing and approving tools incurs significant overhead - not only for the reviewer, but for the tool developer was well.  I set aside time to respond to each of your comments, make spelling corrections and add more tests.

tools/vsnp/vsnp_determine_ref_from_data.xml
gregvonkuster: Typo - fixed here https://github.com/galaxyproject/tools-iuc/pull/3215/commits/d4ab188539becd44af8e840bb3eba41aeef9c159

tools/vsnp/vsnp_determine_ref_from_data.xml
gregvonkuster: It is definitely in preparation, although I have not yet been involved, so I have no working title, etc.  The initial author has not released anything for myself and others to review and add sections.  Is it ok to leave this citation as is until I have a doi?  I've fixed it to use the macro like these associated tools that have been accepted - https://github.com/galaxyproject/tools-iuc/tree/master/tools/vsnp.

tools/vsnp/vsnp_determine_ref_from_data.xml
gregvonkuster: I've enhanced the description here https://github.com/galaxyproject/tools-iuc/pull/3215/commits/8dc3332569ed31f53fefc5f1e604061fce190342.   Please let me know if it is still unclear.

tools/vsnp/vsnp_determine_ref_from_data.xml
gregvonkuster: Fixed here https://github.com/galaxyproject/tools-iuc/pull/3215/commits/d4ab188539becd44af8e840bb3eba41aeef9c159.

tools/vsnp/vsnp_determine_ref_from_data.xml
gregvonkuster: I respectfully disagree with you on this one.  Like constants in programming, I think it is better to associate the string with a variable and then refer to the variable throughout the code.  This eliminates potential issues if the string is inadvertently tweaked somewhere in the code.  For example,
```
...
mkdir -p 'input_reads_dir' &&
...
ln -s '$filename' 'input_reads_directory/$identifier' &&
...
```
would cause problems.

I've used this approach in many tools that have been accepted into this repo (e.g., https://github.com/galaxyproject/tools-iuc/blob/master/tools/vsnp/vsnp_add_zero_coverage.xml#L16).

tools/vsnp/vsnp_determine_ref_from_data.py
gregvonkuster: No, this ensures that the number of simultaneous processes is no more than 1/2 the total number of processors currently available.

tools/vsnp/vsnp_determine_ref_from_data.xml
gregvonkuster: The difference between "paired" and "list:paired" collection types has always confused me.  Looking at other tools, it seems that "list:paired" is generally defined for outputs, although not always.  I do know "paired" works here for input collections costing of more than 1 pair.

tools/vsnp/vsnp_determine_ref_from_data.py
bernt-matthias: Why would one do this? If GALAXY_SLOTS is set then you are guaranteed to have the number of processors for your process. So you should use them (if there are enough files).

tools/vsnp/vsnp_determine_ref_from_data.xml
```diff
@@ -0,0 +1,213 @@
...
+            ## tool in workflows.  However, the output list created by the
+            ## expression tool will not function correctly with the bwa_mem
+            ## mapper.  Naming the identifier as follows is a solution.
+            #set identifier = re.sub('[^\s\w\-]', '_', str($i.name))
+        #end if
+        ln -s '$filename' '$input_reads_dir/$identifier' &&
+    #end for
+#end if
+python '$__tool_directory__/vsnp_determine_ref_from_data.py'
+#if str($input_type) == "single":
```
bernt-matthias, line 61: should be input_type_cond.input_type

tools/vsnp/vsnp_determine_ref_from_data.py
gregvonkuster: This tool uses Python's multiprocessing https://github.com/galaxyproject/tools-iuc/pull/3215/files#diff-b009b15196c062aada354e898011aa6e2a384a249086c6a44a1923410b818e2eR5, which most tools do not use.  

This function will determine the safe number of cpus to use - https://github.com/galaxyproject/tools-iuc/pull/3215/files#diff-b009b15196c062aada354e898011aa6e2a384a249086c6a44a1923410b818e2eR314 when splitting processes across them - https://github.com/galaxyproject/tools-iuc/pull/3215/files#diff-b009b15196c062aada354e898011aa6e2a384a249086c6a44a1923410b818e2eR322.

I've used this approach in all associated tools that have been accepted and that use this feature - e.g., https://github.com/galaxyproject/tools-iuc/blob/master/tools/vsnp/vsnp_add_zero_coverage.py#L128.

tools/vsnp/vsnp_determine_ref_from_data.py
gregvonkuster: That's another approach, but both approaches are functionally correct and follow Python's best-practices.  So its more amateur of coding style.

tools/vsnp/vsnp_determine_ref_from_data.py
gregvonkuster: Hmm...I disagree.

tools/vsnp/vsnp_determine_ref_from_data.py
gregvonkuster: There is no need for it.  This is a check to ensure the multiprocessing queues are empty before unblocking and proceeding to whatever process follows after they are.

tools/vsnp/vsnp_determine_ref_from_data.xml
gregvonkuster: That is defined on line 14 - https://github.com/galaxyproject/tools-iuc/pull/3215/files#diff-d7fd26a4eec1fe38ea999f2c29bd394dbcf14b5ecdb816a3839aa92e8b39b67aR14

tools/vsnp/vsnp_determine_ref_from_data.py
bernt-matthias: OK. processes will be `1` anyway  (when `GALAXY_SLOTS` is not set) and  `num_cpus >= processes` will hold otherwise (When `GALAXY_SLOTS` is set this happens by job runners like drmaa / slurm .. i.e. the ressources are guaranteed).

tools/vsnp/vsnp_determine_ref_from_data.xml
gregvonkuster: 2 additional tests added here  https://github.com/galaxyproject/tools-iuc/pull/3215/commits/10eb011f23d038b927fea61ad0b3aff9d7645d62

tools/vsnp/vsnp_determine_ref_from_data.xml
gregvonkuster: Added here https://github.com/galaxyproject/tools-iuc/pull/3215/commits/923ab72ab8995cd12e08192c665322f2bd01ae8e

tools/vsnp/vsnp_determine_ref_from_data.xml
gregvonkuster: Whether the data is gzipped or not has nothing to do with the work-around here.

I struggled with this, trying to figure out why it is possible to hook a list:paired (automatically generated by the Galaxy tool framework if the input is a "paired" list) output from this tool to the input for the Parse parameter value expression tool in a workflow and then hook the output of the Parse parameter value expression tool to the input for the reference in the bwa-mem tool.  

This all works just fine in the workflow editor.  However, the workflow will stop execution with the output of the Parse parameter value expression tool, and the bwa-mem tool is not executed.  This may be a current weakness in the list produced by the Parse parameter value expression tools for paired reads, which does not seem to be functionally correct as an input to the reference setting for the bwa-mem tool.

I finally discovered this work-around where I produce a single text file that contains the Galaxy dbkey for both pairs, which is really what is needed anyway.

tools/vsnp/vsnp_determine_ref_from_data.xml
gregvonkuster: Unless I've missed it, it is not currently possible to test a regular tool (i.e., not a data manager tool) in such a way that it will test a cached data table value.  Please let me know if I'm incorrect on this.

I originally introduced this functional test hack in this tool https://github.com/galaxyproject/tools-iuc/blob/master/tools/multigps/multigps.xml#L21 so that a chromosome length file could be used.

tools/vsnp/vsnp_determine_ref_from_data.xml
bernt-matthias: I remember that there was some inconsistency that you may use `element_identifier` not in all cases .. and sometimes `name` is needed:

Maybe you can get some inspiration on looping over nested collections here:

https://github.com/galaxyproject/galaxy/blob/f9addab2260aa38027f80af736c397a89656887d/test/functional/tools/collection_nested_test.xml#L4

I guess you want to process all pairs at once? Currently you process each pair in a separate job, or?

tools/vsnp/vsnp_build_tables.py
gregvonkuster: @bernt-matthias I believe I have accommodated all of your comments, but please let me know if I missed any.  Thanks again for all of your help with this complex tool!

tools/vsnp/vsnp_determine_ref_from_data.xml
bernt-matthias: See here https://planemo.readthedocs.io/en/latest/writing_how_do_i.html#test-index-loc-data .. there are also a few tools in the IUC repo that do this, e.g. mitos

tools/vsnp/vsnp_build_tables.py
gregvonkuster: @bernt-matthias The scripts here https://github.com/USDA-VS/vSNP were built to be run in batch / single-user mode on the command line.  Python was used to tightly couple many distinct tool executions using a single python call on the command line.  

The Galaxy workflow for this analysis consist of about 13 tools, all of which are executed by the 2 Python scripts here https://github.com/USDA-VS/vSNP/blob/master/bin/vSNP_step1.py and here https://github.com/USDA-VS/vSNP/blob/master/bin/vSNP_step2.py.  This tight-cooling of tools is not consistent with the Galaxy approach for complex analyses in which tools are only coupled via the workflow itself.

tools/vsnp/vsnp_determine_ref_from_data.xml
gregvonkuster: Regarding the use of element_identifier, using it does not typically work for datasets outside of collections since the identifier in these cases is the history item text (e.g., Filter on data 1 and 2).  With collection elements, it preserves the useful file names (e.g., CMMF_R1.fastq), which in many cases is essential throughout a workflow.

Regarding how the pairs are processed, I have no preference and so I've left it to the Galaxy job framework.

Do you see a specific problem with how I've done this?

gregvonkuster:
@bernt-matthias Its unclear to me whether you require more changes to this PR in order for it to be merged - sorry, the review has become quite convoluted.  :(. Can you please specify any additional changes that are required?

tools/vsnp/vsnp_determine_ref_from_data.xml
gregvonkuster: @bernt-matthias Ah, yes, now I remember that this is now possible for regular tools (thanks for the pointer).  IIRC, I believe that I even have some tests using this approach in some of the associated tools I'll be submitting when this PR is merged.  

I started working through the tests and remembered why I used this approach.  Sorry, I've been sidetracked with a lot of other things over the past few months since I initially submitted this PR, so I had forgotten.

I'm passing the entries for the vsnp_dnaprints data table to the Python script ( https://github.com/galaxyproject/tools-iuc/pull/3215/files#diff-d7fd26a4eec1fe38ea999f2c29bd394dbcf14b5ecdb816a3839aa92e8b39b67aR74), and this value is received by the script as None when using planemo for testing.  So it seems that planemo doesn't load the data tables the same way that Galaxy does.  

Do you know if there is a way to have planemo accommodate these lines in the tool config?
```
#set $dnaprint_fields = $__app__.tool_data_tables['vsnp_dnaprints'].get_fields()
#for $i in $dnaprint_fields:
    --dnaprint_fields '${i[0]}' '${i[2]}'
#end for
```
If so, I'll fix the tests to use that approach.  I don't believe there are any other tools that are doing this.

bernt-matthias: If I get it right you use a data table without the need for the user to choose anything .. you just take the complete thing (two columns if I get it right). 

I would hope that you just add the files `vsnp_dnaprints.loc.sample` `tool_data_table_conf.xml.sample` `tool_data_table_conf.xml.test`, and a loc file in the test-data directory (which is referred by `tool_data_table_conf.xml.test`) .. maybe you just try?

tools/vsnp/vsnp_determine_ref_from_data.py
bernt-matthias: Its 3x nearly the same code, just the function and the args change... but up to you.

tools/vsnp/vsnp_determine_ref_from_data.py
bernt-matthias: So if you are sure that they are empty, then you don't need the if?

tools/vsnp/vsnp_determine_ref_from_data.xml
bernt-matthias: 1. If you use `paired` and pass a list of pairs then each pair in the list in processed by a separate job.
2. If you use `list:paired` then all pairs in the list are processed in a single job. 

I guess you want the later here. At least this would be consistent with what you are doing in the collection single_reads case.

gregvonkuster: Thanks for pointing out the difference - fixed here https://github.com/galaxyproject/tools-iuc/pull/3215/commits/410d60485eabbefe1652ce42ced7735876a4c86b.  I so appreciate now knowing the difference!

tools/vsnp/vsnp_determine_ref_from_data.py
gregvonkuster: The if is most definitely necessary.  There are 2 queues here working in complementary fashion, and both queues must be empty before either is closed.  Is there something specific here that is concerning you?

tools/vsnp/vsnp_determine_ref_from_data.py
gregvonkuster: Duplicate lines eliminated here https://github.com/galaxyproject/tools-iuc/pull/3215/commits/711d9ad40b9429a7b8e93c737c76a601210e096e

tools/vsnp/vsnp_determine_ref_from_data.xml
gregvonkuster: I am building an in-memory dictionary of each dbkey and associated dnaprint strings.  The user cannot choose a data table selection since the data tables for this tool are being used only to map donkeys to dnaprint strings, and the entire dictionary must be loaded into memory (it is extremely small, < 1kb) so that the dnaprint strings deciphered in the input data can be matched to each of the dnaprint strings associated with each dbkey in the dictionary.

 I actually did change the code to attempt to use the planemo approach for testing data tables - here are the diff'd files in my repo.  I added a .greg extension to them.
```
$ git status
On branch vsnp_drfd
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	test-data/NC_002945v4.yml
	tool-data/vsnp_dnaprints.loc.sample
	tool_data_table_conf.xml.sample.greg
	tool_data_table_conf.xml.test.greg
	vsnp_determine_ref_from_data.py.greg
	vsnp_determine_ref_from_data.xml.greg
```
I also eliminated the use of the "in_test_mode" flag throughout the code and ran planemo test.  All tests failed due to the dna_prints value being passed as None on the command line to the script.  Seeing this was what caused me to remember why I had used this approach.

tools/vsnp/vsnp_determine_ref_from_data.xml
bernt-matthias: I'm really wondering why this works for `list:paired` .. if I'm not wrong you create only a single link for each. 

Also the test for this case essentially only gives a single pair, one would need a nested element tags as in 
https://github.com/galaxyproject/galaxy/blob/dev/test/functional/tools/collection_creates_dynamic_nested.xml

But then the question is if the command block needs to accommodate for both cases. 

But maybe this is obsolete with the other comment ...

tools/vsnp/vsnp_determine_ref_from_data.py
```diff
@@ -0,0 +1,355 @@
...
+        # dataset collections will be produced as outputs.
+        multiprocessing.set_start_method('spawn')
+        queue1 = multiprocessing.JoinableQueue()
+        queue2 = multiprocessing.JoinableQueue()
+        num_files = len(fastq_list)
+        cpus = set_num_cpus(num_files, args.processes)
+        # Set a timeout for get()s in the queue.
+        timeout = 0.05
+
+        for fastq_file in fastq_list:
```
bernt-matthias, line 325: Late question from my side (sorry): 

Am I right that you use data parallelization, i.e. the fastq files (or fastq pairs) are processed independently. If so I would strongly suggest to kick all the parallelization in the python script and to use Galaxy's parallelization power (using batch mode inputs). 

If I'm wrong you do not need to continue reading :)

With the current implementation you are limited to the number of cores on a single node while Galaxy could roll out the fastq files to multiple nodes.

This would also simplify the tool form significantly because you only need to allow for 

- single reads 
- a paired read a paired collection 
- and maybe a paired read in separate data sets 

You can still give a list (resp. list:paired) to the tool form, but now Galaxy will roll out independent jobs for each element and create a list (resp. list:paired) as a result.

gregvonkuster: Will this PR be rejected if the tool remains in its current form?

bernt-matthias: I just see an easy way for a massive simplification of the tool wrapper, that should be easy to implement (if you like you do not need to touch the python script at all .. but also this change should be simple), and also would be much more performant on clusters (as long as the overhead to start a Galaxy job is smaller than the typical runtime for a fastq file).

Also I'm not convinced that the tool is correctly implemented (see my other comment). Maybe you can add a test that uses two pairs as input to prove me wrong.

gregvonkuster: Regarding eliminating the use of Python multiprocessing, @davebx and I had a long discussion on this back in early April when the tool (at that time) looped over an input collection in this way https://github.com/gregvonkuster/galaxy_tools/blob/dc2eb46c922d117238e0c8567f251a0697336d48/tools/sequence_analysis/vsnp/vsnp_determine_ref_from_data/vsnp_determine_ref_from_data.py#L226.  

I was trying to find a way to split the job over multiple processors, and defined this configuration for the tool in the job config https://github.com/gregvonkuster/galaxy_tools/blob/master/galaxy/config/vsnp/job_conf.xml#L64, which still restricted the job to a single processor since the above loop is executed after the job is started.  This is the reason that I went with the multiprocessing implementation.

So are you saying that Galaxy will automatically split the job across multiple processors (or more likely create multiple jobs) if I simply change the input types?   I know that Galaxy does this for tools that don't take input collections (e.g., bwa_mem) when an input collection is selected for the tool in the tool form.  But I need the ability to do this inside of a workflow because the vsnp tool suite is most meaningful in a defined workflow.

The tool will need to accept any of the following inputs:

- a single read
- a set of paired reads
- a collection of single reads
- a collection of paired reads

So how would I accommodate that in a simpler way in the tool config, ensuring that Galaxy will load balance across available processors and that single collections will be produced as output for the tool if Galaxy creates multiple jobs for load balancing? 

Also, this tool will typically take input collections with 100's (and sometimes 1000's) of elements for the 3rd and 4th input type above.  So it would be good to ensure whatever load balancing approach is taken is optimal.

gregvonkuster: Regarding your request for a test that include 2 sets of paired reads as input, they would have to be in a collection, correct?  Or are you asking for 2 sets of pairs outside of a collection?  If so, the tool currently does not support this as input.

bernt-matthias: Thanks for the clarifications.

> So are you saying that Galaxy will automatically split the job across multiple processors (or more likely create multiple jobs) if I simply change the input types? I know that Galaxy does this for tools that don't take input collections (e.g., bwa_mem) when an input collection is selected for the tool in the tool form. 

Exactly.

> But I need the ability to do this inside of a workflow because the vsnp tool suite is most meaningful in a defined workflow.

Works the same. 

> The tool will need to accept any of the following inputs:
> 
>     * a single read
> 
>     * a set of paired reads
> 
>     * a collection of single reads
> 
>     * a collection of paired reads

This works. But an implementation would be needed only for the first two. The others are provided automagically by Galaxy.

> So how would I accommodate that in a simpler way in the tool config, ensuring that Galaxy will load balance across available processors and that single collections will be produced as output for the tool if Galaxy creates multiple jobs for load balancing?

input will be

- conditional single / paired
  - when single -> param accepting a single fastq
  - when paired -> param accepting a pair

the output will be

- a single fastq + a filter that ensures that it is produced when the conditional is single
- a pair of fastq + a filter that ensures that it is produced when the conditional is paired

For collections nothing else is needed. The tool will take a collection (of pairs) and produce a collection (of pairs).

> Also, this tool will typically take input collections with 100's (and sometimes 1000's) of elements for the 3rd and 4th input type above. So it would be good to ensure whatever load balancing approach is taken is optimal.

That is actually the critical question. How long will one of these jobs take. If they are short (i.e. (much) shorter than the overhead to roll out a job) then your approach would probably more efficient (using the terminology for parallel computing).

bernt-matthias: > Regarding your request for a test that include 2 sets of paired reads as input, they would have to be in a collection, correct? Or are you asking for 2 sets of pairs outside of a collection? If so, the tool currently does not support this as input.

The former two pairs in a collection.

gregvonkuster: What is the Galaxy magic that splits the for loop that is executed after the job starts over multiple processors?  When I executed the tool back in April, watching top demonstrated that it was restricted to a single processor, no matter the setting for the tool in the job config.  Here are the inputs I had at that time https://github.com/gregvonkuster/galaxy_tools/blob/dc2eb46c922d117238e0c8567f251a0697336d48/tools/sequence_analysis/vsnp/vsnp_determine_ref_from_data/vsnp_determine_ref_from_data.xml#L64.

From your response, it seems that tools should never be written to accept collections on input.  Is this correct?

I don't have a lot of statistics on job time for large input collections.  A collection of 15 samples (~100 mb each) takes around 12 - 13 minutes to complete when splitting the job across 8 processors using multiprocessing.

I'll add another test that includes 2 pairs in an input collection.

tools/vsnp/vsnp_determine_ref_from_data.xml
gregvonkuster: @bernt-matthias I spent time looking more closely at the differences between paired and list:paired collection types, which are significant, but I haven't found how Galaxy splits the job differently based on this 2 collection types.  

In any case, I don't want an input list:paired collection type for this tool, I want a paired collection type, so I have reverted the setting back to what I had originally here https://github.com/galaxyproject/tools-iuc/pull/3215/commits/e2ef12484caf681127b08abe589d487c701141ef.

This issue provides a good explanation https://github.com/galaxyproject/galaxy/issues/3617.  This tool is performing a similar task as those discussed in the issue - accepting a list of pairs on input and providing a list of single outputs per pair.

tools/vsnp/vsnp_determine_ref_from_data.py
gregvonkuster: @bernt-matthias I'm very confused about your feedback on the input types for this tool and tools in general.  I understand how tools that are not written to accept input collections will automatically create multiple jobs (I assume 1 per input) when a collection is provided as input to the tool.  An example is the bwa_mem tool.

In writing this tool, I discovered that Galaxy will not automatically perform job load balancing for an input collection, something I didn't know before I wrote the tool to accept collections on input.  I assume the Galaxy framework currently treats the collection as a single input, and so doesn't split the job based on the collection elements - I may be completely wrong on this, it is simply what I've seen while watching top during tool execution.  This is why I enhanced the tool to use Python's multiprocessing.

I know that not many tools use multiprocessing to load balance jobs with input collections in this way.  But many tools are written to accept input collections (e.g., fargene, fastp, tb-profiler, etc).  In the tools-idc repo there are dozens of tools that are written to accept input collections.  I wonder if these tools are restricted from job splitting in those cases where the input is a collection.  Do you know if the is the case?

Another reason that I use collections on input and output for the vsnp tool suite is because preserving file names is critical throughout the vsnp pipeline, and these file names are lost when outside of collections.  The sample files have very descriptive names like Mcap_Deer_DE_SRR650221.fastq.gz, and the vsnp tools must have the Mcap_Deer_DE_SRR650221 sample name identifier preserved throughout the complex analysis chain.  If collections are not used, the file name becomes something like dataset_453.dat (and the identifier becomes something like "vSNP: get SNPs on data 456 and data 479"), which at best is meaningless at the end of the vsnp analysis pipeline, and at worst, throws exceptions during the workflow execution.

So I'm struggling with the changes that you're requesting for this tool.  I'll gladly eliminate the input collections and the Python multiprocessing if we can be assured that file names will be preserved and tool executions will be optimally load-balanced.

tools/vsnp/vsnp_determine_ref_from_data.xml
bernt-matthias: Hi @gregvonkuster .. just working on it .. nearly finished and I just need to fix the use of data tables :) Could you be so nice to add 

- `tool-data/vsnp_dnaprints.loc.sample`
- and the modified `tool_data_table_conf.xml.sample` and `tool_data_table_conf.xml.test`

Those are needed anyway.

gregvonkuster: Added here https://github.com/galaxyproject/tools-iuc/pull/3215/commits/908ef3c12e88c6e88b51f7af73baea6e2264b0b5

bernt-matthias: Excellent. Thanks. Should I then push my commits to this branch (if I'm allowed to) or create a PR on your branch.

tools/vsnp/vsnp_determine_ref_from_data.py
gregvonkuster: @bernt-matthias can you please provide a response to my comments above?  I don't understand why the IUC has allowed other tools into this repo that are written to accept collections on input, but you are not allowing this one.  Are you going to make the same changes to these associated tools that have already been merged?  https://github.com/galaxyproject/tools-iuc/tree/master/tools/vsnp.  I have additional associated tools that I was planning to submit in separate PRs after this one was merged, but I'm no longer sure I should.

bernt-matthias: To keep it simple I will only discuss the difference between the handling of parameters that are of type `data` vs `list`. The same arguments hold for `paired` vs `list:paired`.

> @bernt-matthias I'm very confused about your feedback on the input types for this tool and tools in general. I understand how tools that are not written to accept input collections will automatically create multiple jobs (I assume 1 per input) when a collection is provided as input to the tool. An example is the bwa_mem tool.

Exactly. If you have a `data` input then

- if you input a single data set 1 job is created
- if you input a collection 1 job is created for each data set in the collection. All of these jobs may run in parallel if Galaxy's configuration allows it. 

> In writing this tool, I discovered that Galaxy will not automatically perform job load balancing for an input collection, something I didn't know before I wrote the tool to accept collections on input. I assume the Galaxy framework currently treats the collection as a single input, and so doesn't split the job based on the collection elements - I may be completely wrong on this, it is simply what I've seen while watching top during tool execution. This is why I enhanced the tool to use Python's multiprocessing.

Indeed. If you have an input of type `list` then a single job is created that processes all the files in the collection. This is needed you can not analyze the data sets independently (trivial example would be a tool that merges data sets). Sometimes it may also be desired to do this even if the data sets could be processed independently. One example would be if the processing time per data set is very short because then the overhead due to creating separate jobs would be to much. 

> I know that not many tools use multiprocessing to load balance jobs with input collections in this way. But many tools are written to accept input collections (e.g., fargene, fastp, tb-profiler, etc). In the tools-idc repo there are dozens of tools that are written to accept input collections. I wonder if these tools are restricted from job splitting in those cases where the input is a collection. Do you know if the is the case?

If the input is a collection a single job is created and no job splitting happens. I don't know any of the tools. Let me give another example https://github.com/galaxyproject/tools-iuc/blob/f2a33fe115fef9d711112b53136cf7619f1b19be/tools/dada2/dada2_plotQualityProfile.xml 

Here I allow the user to select between the two cases. Here the point was that the underlying R function assumed (unnecessarily .. in most cases) that inputs are treated jointly. So I left the choice to the user. But the output is identical in both cases .. which seems not to be the case for your tests involving a pair: https://github.com/galaxyproject/tools-iuc/blob/908ef3c12e88c6e88b51f7af73baea6e2264b0b5/tools/vsnp/vsnp_determine_ref_from_data.xml#L169 vs https://github.com/galaxyproject/tools-iuc/blob/908ef3c12e88c6e88b51f7af73baea6e2264b0b5/tools/vsnp/vsnp_determine_ref_from_data.xml#L188

> Another reason that I use collections on input and output for the vsnp tool suite is because preserving file names is critical throughout the vsnp pipeline, and these file names are lost when outside of collections. The sample files have very descriptive names like Mcap_Deer_DE_SRR650221.fastq.gz, and the vsnp tools must have the Mcap_Deer_DE_SRR650221 sample name identifier preserved throughout the complex analysis chain. If collections are not used, the file name becomes something like dataset_453.dat (and the identifier becomes something like "vSNP: get SNPs on data 456 and data 479"), which at best is meaningless at the end of the vsnp analysis pipeline, and at worst, throws exceptions during the workflow execution.

That's true iff you apply the tool to a single data set. But if you apply it to a collection the identifiers in the output collection will be the same as those in the input collection. Admittedly the input of a single data set is possible (and can't be forbidden as far as I know) and creates unusable data set names, but this might be solved by docs in the help section.

> So I'm struggling with the changes that you're requesting for this tool. I'll gladly eliminate the input collections and the Python multiprocessing if we can be assured that file names will be preserved and tool executions will be optimally load-balanced.

You may simply checkout the latest commit and run a `planemo serve` and try the tool with an input collection. You will see that multiple jobs are created (probably only 4 .. due to planemos config) and that ids of input and output collection match.

bernt-matthias: > I don't understand why the IUC has allowed other tools into this repo that are written to accept collections on input, but you are not allowing this one. 

Please take my commits as a suggestion. Maybe you come to the conclusion that this approach is worth considering or maybe you convince me that collections are really needed here.

As I wrote earlier input collections have their use cases. I'm just not yet convinced that they apply here. Also I was still not convinced that the implementation was correct (different output for pairs as written above).

In the end I (and many others) only try to make sure the the tools in this repository are as good as possible (in terms of usability as well as efficiency).

> Are you going to make the same changes to these associated tools that have already been merged? https://github.com/galaxyproject/tools-iuc/tree/master/tools/vsnp. 

If you agree on my suggestions it seems consequential to do analogous changes. But I think that my time does not allow me to do so. So far its only two that might be improved, or?

> I have additional associated tools that I was planning to submit in separate PRs after this one was merged, 

Please be ensured that IUC is really happy about every contribution. 

> but I'm no longer sure I should.

I really hope that this is not due to my review.

gregvonkuster: @bernt-matthias Thanks very much for all of your help on this and especially your re-write to eliminate the use of input collections and the resultant need for multiprocessing.  Results in your version are as expected in the tests that I've run.  I'll make the changes to all of the associated tools.  Thanks also for figuring out how to eliminate the functional testing hack that I used as well as for confirming that input collections cannot be load balanced by the Galaxy framework.  I'll make sure to stay away from them for input types on future tools.  Your significant help on this PR is very much appreciated.

bernt-matthias (approved):

tools/vsnp/vsnp_determine_ref_from_data.py
bernt-matthias: Thanks for your contribution. Anything else you would like to change? 

Any other comments @bgruening ?

bgruening: Now, this looks great. The only nitpick is that Mcap_Deer_DE_SRR650221.fastq.gz is large? Can this be reduced?

gregvonkuster: Yeah, I know this is an issue, but it really cannot be further reduced.  I spent a lot of time reducing the original sample down to this size.  Any additional data that is removed (believe me, I've tried) will eliminate the sequences that are needed for the tool to find the "DNA prints" needed for the functional tests to pass.

gregvonkuster: To clarify, my lack of expertise on the biology side caused me to remove data randomly.  I suppose additional unnecessary sequences could be removed if a data expert took a look.

bgruening: Ok, if Matthias is fine with it, lets get it in. @gregvonkuster hopefully you can reuse this file for your other tools as well.

bernt-matthias: I'm fine with it. But I also have 2 ideas to reduce the size:

- if the tools do not use the qualities we could use the same value for all positions. This should make the fastq file better compressible (btw. has max compression level been used)
- instead of a random subset one could try
  - remove short sequences
  - throw out very similar sequences .. maybe this has a smaller effect on the results.

gregvonkuster: I'm removing the short sequences and have had success so far.  I'll see how far I can get.

gregvonkuster: I made it smaller here https://github.com/galaxyproject/tools-iuc/pull/3215/commits/1deab04caf858d3a487218ac0655ada68931e266, although it is still pretty large.  I made sure max compression is used.  We can make it a bit smaller if we eliminate the first functional test for the single read.  I think the 2 paired read tests provide sufficient coverage if you'd like me to do that, but I'm fine either way.  I assume this PR will be "squished" before merging?

bernt-matthias: In this case I prefer the test. 

Yes, we should "squash" it :)

bgruening: @bernt-matthias please feel free to squeeze during merge and get it in.

gregvonkuster:
Thanks so very much for all that you've done on this @bernt-matthias !!

bernt-matthias:
Welcome. Thank you to.

PR #3217, opened 2020-09-29
increase speed by not converting as a separate command

astrovsky01:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/plink/plink.xml
+++ b/tools/plink/plink.xml
@@ -46,18 +46,7 @@
     ## Create Plink folder for any inputs
     mkdir ./plink_output
     && mkdir ./plink_input
-    #if $inputs.inputs.filetype == 'bfile':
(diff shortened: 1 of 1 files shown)


bgruening:
@astrovsky01 such big changes warrant a version bump I think - next time.

astrovsky01:
@bgruening the only thing I did was to remove the double run on vcf. Outputs are identical. Imo, it should be more considered  more of a bug fix than a version update.

wm75:
@astrovsky01 I guess @bgruening is right. Outputs may be identical or not, but you are executing a different command line. These kinds of "small" changes without version bumps are causing issues more regularly than one would think.

astrovsky01:
@wm75 That's fair. Will version bump next time. If you'd prefer, we could roll back the version, then pr this with a version bump?

PR #3232, opened 2020-10-05
Add rnaquast

lehmann-ju:
FOR CONTRIBUTOR:
* [x] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [x] - License permits unrestricted use (educational + commercial)
* [x] - This PR adds a new tool or tool collection
* [ ] - This PR updates an existing tool or tool collection
* [ ] - This PR does something else (explain below)

Github: [https://github.com/ablab/rnaquast](url)
(can also be added to the quast folder)

PR #3239, opened 2020-10-08
balance chunks in CI workflow

bernt-matthias:
fixes: https://github.com/galaxyproject/tools-iuc/issues/2959

FOR CONTRIBUTOR:
* [ ] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [ ] - License permits unrestricted use (educational + commercial)
* [ ] - This PR adds a new tool or tool collection
* [ ] - This PR updates an existing tool or tool collection
* [ ] - This PR does something else (explain below)

PR #3248, opened 2020-10-13
fail PR workflow if no test was executed

bernt-matthias:
e.g. in case planemo itself fails as here https://github.com/galaxyproject/tools-iuc/pull/3232/checks?check_run_id=1246607831

- not sure why the linked example does not fail at linting stage as well (it fails, but for a different reason).
- also not entirely sure if its the right thing to do

FOR CONTRIBUTOR:

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -371,7 +371,10 @@ jobs:
         path: upload
     - name: Check status of combined status
       run: |
-        if jq '.["tests"][]["data"]["status"]' upload/tool_test_output.json | grep -v "success"; then
+        if [[ $(jq '.["tests"]' upload/tool_test_output.json) == "[]" ]]; then
+            echo "No tests were executed inspect the output of the chunks"
+            exit 1
+        else if jq '.["tests"][]["data"]["status"]' upload/tool_test_output.json | grep -v "success"; then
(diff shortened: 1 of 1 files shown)


nsoranzo:
I think it's fine as it is? The linked build still fails because the linting step fails.

Also, we don't want to fail a build which just modifies `README.md` .

The only annoying issue for me is that a new tool added without a `.shed.yml` is silently not linted/tested, but not sure how to solve that.

bernt-matthias:
> The only annoying issue for me is that a new tool added without a .shed.yml is silently not linted/tested

If I got it right the change to the CI workflow introduced here https://github.com/galaxyproject/tools-iuc/pull/3239 would help. If I get it right `ci_find_repos` enumerates tool dirs by checking for shed.yml files (and therefore tested samtools 2x) and `ci_find_tools` enumerates tool files (but I did not explore this to the end).

If I'm right we can do the same change in the PR workflow. Even if balancing workload is not that relevant, it would simplify the logic (no different handling of data managers and tools) in addition.

nsoranzo:
It seems so, we could:
- use `ci_find_repos` to determine on which directories to run `planemo shed_lint`
- use `ci_find_tools` to determine on which files to run `planemo test`
- check that every tool path found by `ci_find_tools` is a substring of a directory find by `ci_find_repos`, if not a `.shed.yml` file is missing

PR #3255, opened 2020-10-16
Various scpipe fixes

mvdbeek:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/scpipe/scpipe.xml
+++ b/tools/scpipe/scpipe.xml
@@ -163,7 +163,7 @@ Rscript '$__tool_directory__/scpipe.R'
                         </param>
                     </when>
                     <when value="history">
-                        <param name="ref_fa_hist" type="data" format="fasta,fasta.gz" label="Select a history FASTA" />
+                        <param name="ref_fa_hist" type="data" format="fasta" label="Select a history FASTA" />
                     </when>
                 </conditional>
                 <conditional name="paired_format">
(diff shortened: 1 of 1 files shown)


tools/scpipe/scpipe.xml
```diff
@@ -74,7 +74,7 @@ ln -s '${exons}' '$anno_name' &&
 TAB=\$(printf '\t') &&
 
 #if $samples.barcodes:
-    sed -i.bak -e "s/\${TAB}/,/g" '$samples.barcodes' &&
+    sed -e "s/\${TAB}/,/g" '$samples.barcodes' > samples.barcodes &&
```
bernt-matthias, line 77: Nasty one. Good catch.

tools/scpipe/scpipe.xml
```diff
@@ -244,8 +244,8 @@ Rscript '$__tool_directory__/scpipe.R'
         <test>
             <param name="format_select" value="fastq" />
             <param name="fasta_source" value="history"/>
-            <param name="ref_fa_hist" ftype="fasta" value="mm10_MT19.fa.gz"/>
-            <param name="exons" ftype="gff3" value="mm10_MT19.gff3.gz"/>
+            <param name="ref_fa_hist" ftype="fasta.gz" value="mm10_MT19.fa.gz"/>
```
bernt-matthias, line 247: Question: we do not need to list xyz.gz in the format of the param, but here we need to?

bernt-matthias (approved):

tools/scpipe/scpipe.xml
mvdbeek: Right, if you don't add an ftype the format guessed at upload will be chosen. IF we do specify it it need to be correct though, so we shouldn't use `ftype="fasta"` when the file is gzipped.

bernt-matthias: thanks for the explanation

PR #3256, opened 2020-10-16
fixing bug with annotateVCF.py

NatPRoach:
FOR CONTRIBUTOR:

Had an issue with annotateVCF.py not handling insertions when quantifying SNPs correctly.

### DIFF
--- a/tools/medaka/annotateVCF.py
+++ b/tools/medaka/annotateVCF.py
@@ -75,6 +75,7 @@ def parseSimpleSNPpileup(fields, ref_base, alt_base):
                 ins_str += base
             else:
                 ins_len = int(ins_str) - 1
+                ins_str = ""
                 insertion = base
                 ins_flag = False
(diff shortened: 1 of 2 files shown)


wm75:
Seems that these same issues exist in the `parseIndelPileup` function. Is that not an issue?

NatPRoach:
> Seems that these same issues exist in the `parseIndelPileup` function. Is that not an issue?

Oh shoot, looks like you're right RE: AF. I don't think the ins_str and del_str has the same problem. There's a line that resets ins_str and del_str in parseIndelPileup in the following block:
```
        elif base == '+':
            ins_flag = True
            ins_str = ""
        elif base == '-':
            del_flag = True
            del_str = ""
```

wm75:
Ah, right, just in a different place that I haven't noticed.

wm75 (approved):

PR #3257, opened 2020-10-19
Use planemo-discover-action to reduce a bit of CI boilerplate

mvdbeek:
FOR CONTRIBUTOR:

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -17,6 +17,9 @@ jobs:
     strategy:
       matrix:
         python-version: [3.7]
+    outputs:
+      galaxy-commit: ${{ steps.galaxy-commit.outputs.galaxy-commit }}
+      changed-repositories: ${{ steps.discover-repositories.changed-repositories }}
     steps:
     - name: Print github context properties
       run: |
@@ -32,12 +35,9 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Determine latest galaxy commit
       run: echo "GALAXY_HEAD_SHA=$(git ls-remote ${{ env.GALAXY_REPO }} refs/heads/${{ env.GALAXY_RELEASE }} | cut -f1)" >> $GITHUB_ENV
-    - name: Save latest galaxy commit to artifact file
-      run: echo $GALAXY_HEAD_SHA > galaxy.sha
-    - uses: actions/upload-artifact@v2
-      with:
-        name: Workflow artifacts
-        path: galaxy.sha
+    - name: Set galaxy head commit
+      id: galaxy-commit
+      run: echo "::set-output name=galaxy-commit::$GALAXY_HEAD_SHA"
     - name: Cache .cache/pip
       uses: actions/cache@v2
       id: cache-pip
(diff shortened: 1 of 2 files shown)


mvdbeek:
@bernt-matthias @nsoranzo @bgruening what do you think about this ?
Ideally we'd also push the caching and artifact up/download into the composite action, but they don't support steps with `using:` yet. If we like this I would rename `mvdbeek/planemo-discover-action` into `galaxyproject/planemo-action` or something along the lines

.github/workflows/pr.yaml
```diff
@@ -109,31 +79,22 @@ jobs:
...
-      run: pip install planemo
-    - name: Planemo lint
-      run: |
-        set -e
-        while read -r DIR; do
-            planemo shed_lint --tools --ensure_metadata --urls --report_level warn --fail_level error --recursive "$DIR";
-        done < ../workflow_artifacts/changed_repositories.list
+        key: pip_cache_py_${{ matrix.python-version }}_gxy_${{ needs.setup.outputs.galaxy-commit }}
+    - uses: mvdbeek/planemo-discover-action@master
+      id: discover-repositories
```
bernt-matthias, line 89: I think here something like

```
with:
    changed-repositories: ${{ needs.setup.outputs.changed-repositories }}
    lint-tools: true
```

is missing?

.github/workflows/pr.yaml
```diff
@@ -345,26 +250,19 @@ jobs:
...
-      run: planemo test_reports tool_test_output.json --test_output tool_test_output.html
-    - name: Copy artifacts into place
-      run: |
-        mkdir upload
-        mv tool_test_output.json tool_test_output.html upload/
+    - uses: mvdbeek/planemo-discover-action@master
+      with:
+         changed-repositories: ${{ needs.setup.outputs.changed-repositories }}
+         combine-outputs: true
+         html-report: true
```
bernt-matthias, line 265: Can we set the path to the generated html report explicitly in the action?

.github/workflows/pr.yaml
```diff
@@ -50,51 +50,21 @@ jobs:
...
-      run: echo "COMMIT_RANGE=${{ github.event.before }}.." >> $GITHUB_ENV
-    - name: Set commit range (pull request)
-      if: github.event_name == 'pull_request'
-      run: echo "COMMIT_RANGE=HEAD~.." >> $GITHUB_ENV
-    - name: Planemo ci_find_repos
-      run: planemo ci_find_repos --changed_in_commit_range $COMMIT_RANGE --exclude packages --exclude deprecated --exclude_from .tt_skip --output changed_repositories.list
-    - name: Show repo list
-      run: cat changed_repositories.list
-    - uses: actions/upload-artifact@v2.0.1
+    - uses: mvdbeek/planemo-discover-action@master
```
bernt-matthias, line 56: Would like to have `name: ...` for those steps.

bernt-matthias (commented):
In the output there are some distracting outputs: 

```
+ '[' false '!=' false ']'
+ '[' tools/add_input_name_as_column == '' ']'
+ echo tools/add_input_name_as_column
+ '[' false == true ']'
+ '[' false == true ']'
+ '[' false == true ']'
+ '[' '' == true ']'
```

Also I'm not sure if caching works yet. The output suggests that pip downloads the packages in each job.

bernt-matthias:
I'm in favor of `galaxyproject/planemo-action`

bernt-matthias:
Wondering if this can be readily adapted for the CI workflow. There a few wee differences, e.g. the grouping of tools.

.github/workflows/pr.yaml
mvdbeek: Sure, it's just a bash script: https://github.com/mvdbeek/planemo-discover-action/blob/master/planemo_ci_actions.sh

mvdbeek:
> Wondering if this can be readily adapted for the CI workflow. There a few wee differences, e.g. the grouping of tools.

I'm sure we can add this, I just based the script on what was on the master branch at the time.

bernt-matthias:
closed https://github.com/galaxyproject/tools-iuc/pull/3437

PR #3259, opened 2020-10-20
Medaka update

NatPRoach:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/medaka/convert_VCF_info_fields.py
+++ b/tools/medaka/convert_VCF_info_fields.py
@@ -8,8 +8,8 @@
 # 10/21/2020 - Nathan P. Roach, natproach@gmail.com
 
 import sys
-from math import log10
 from collections import OrderedDict
+from math import log10
 
 from scipy.stats import fisher_exact
 
--- a/tools/medaka/convert_VCF_info_fields.py
+++ b/tools/medaka/convert_VCF_info_fields.py
@@ -20,10 +20,12 @@ def pval_to_phredqual(pval):
 def parseInfoField(info):
     info_fields = info.split(';')
     info_dict = {}
+    info_list = []  # Want to keep fields in order they appear in original VCF
     for info_field in info_fields:
         code, val = info_field.split('=')
         info_dict[code] = val
-    return info_dict
+        info_list.append(code)
+    return info_dict, info_list
 
 
 def annotateVCF(in_vcf_filepath, out_vcf_filepath):
@@ -52,7 +54,7 @@ def annotateVCF(in_vcf_filepath, out_vcf_filepath):
             out_vcf.write(line)
         else:
             fields = line.strip().split('\t')
-            info_dict = parseInfoField(fields[7])
+            info_dict, info_list = parseInfoField(fields[7])
             sr_list = [int(x) for x in info_dict["SR"].split(',')]
             sc_list = [int(x) for x in info_dict["SC"].split(',')]
             if len(sr_list) == len(sc_list):
@@ -75,9 +77,7 @@ def annotateVCF(in_vcf_filepath, out_vcf_filepath):
                     as_ = [sc_list[ref_fwd], sc_list[ref_rev], sc_list[var_fwd], sc_list[var_rev]]
 
                     info = []
-                    for code in info_dict:
-                        if code in to_skip:
-                            continue
+                    for code in info_list:
                         val = info_dict[code]
                         info.append("%s=%s" % (code, val))
 
@@ -96,9 +96,9 @@ def annotateVCF(in_vcf_filepath, out_vcf_filepath):
                     else:
                         raf = float(dp4[3]) / float(dpr)
                         info.append("RAF=%.6f" % (raf))
+                    info.append("SB=%d" % (sb))
                     info.append("DP4=%s" % (','.join([str(x) for x in dp4])))
                     info.append("AS=%s" % (','.join([str(x) for x in as_])))
-                    info.append("SB=%d" % (sb))
                     new_info = ';'.join(info)
                     fields[7] = new_info
                     out_vcf.write("%s\n" % ("\t".join(fields)))
--- a/tools/medaka/convert_VCF_info_fields.py
+++ b/tools/medaka/convert_VCF_info_fields.py
@@ -9,102 +9,101 @@
 
 import sys
 from math import log10
+from collections import OrderedDict
 
 from scipy.stats import fisher_exact
 
 
 def pval_to_phredqual(pval):
-    return int(round(-10. * log10(pval)))
+    return round(-10 * log10(pval))
 
 
 def parseInfoField(info):
     info_fields = info.split(';')
-    info_dict = {}
-    info_list = []  # Want to keep fields in order they appear in original VCF
+    info_dict = OrderedDict()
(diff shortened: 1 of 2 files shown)


bgruening:
@NatPRoach  1.1.3 is out https://anaconda.org/bioconda/medaka/

NatPRoach:
> @NatPRoach 1.1.3 is out https://anaconda.org/bioconda/medaka/

Yes I will try to get 1.1.3 wrapped shortly. Just trying to fix problems with the current version. I'll try to open a new PR for the new version shortly.

NatPRoach:
Commenting for record keeping purposes RE: @bgruening 's comment that medaka 1.1.3 is released. It would seem that medaka 1.1.3 makes the switch to tensorflow2 and as such many of it's old models are no longer functional in the new version and are not packaged in the conda package. We therefore decided to update v1.0.3 with the new approach to annotating variants first, then update the wrapper to 1.1.3, such that there is a valid version of medaka on tools-iuc that can do variant calling on reads basecalled with older versions of the basecallers.

tools/medaka/variant.xml
```diff
@@ -47,12 +47,16 @@
...
-    && grep -v "^#" '$out_result' > tmp.txt
-    && cut -f 1,2 tmp.txt > positions.txt
-    && samtools mpileup -d 0 --positions positions.txt -Q $output_annotated.min_basequal -q $output_annotated.min_mapqual -f reference.fa '$output_annotated.in_bam' > mpileup.txt
-    && python '$__tool_directory__/annotateVCF.py' '$out_result' mpileup.txt '$out_annotated'
+#if $out_annotated:
+##    && samtools faidx reference.fa
+##    && grep -v "^#" '$out_result' > tmp.txt
+##    && cut -f 1,2 tmp.txt > positions.txt
+##    && samtools mpileup -d 0 --positions positions.txt -Q $output_annotated.min_basequal -q $output_annotated.min_mapqual -f reference.fa '$output_annotated.in_bam' > mpileup.txt
+##    && python '$__tool_directory__/annotateVCF.py' '$out_result' mpileup.txt '$out_annotated'
```
wm75, line 55: No need to keep these old lines around - remove.

tools/medaka/variant.xml
```diff
@@ -84,10 +88,14 @@
                 <option value="true" selected="true">Output annotated VCF</option>
                 <option value="false">Don't output annotated VCF</option>
             </param>
-            <when value="true">
+            <!-- <when value="true">
                 <param name="in_bam" type="data" format="bam" optional="false" label="BAM to annotate the VCF" />
                 <param name="min_basequal" type="integer" value="0" min="0" max="50" label="Minimum base quality" help="The minimum base quality (default: 0) at a given variant position required to include a read in the calculation of variant statistics (samtools mpileup -Q)" />
                 <param name="min_mapqual" type="integer" value="0" min="0" max="60" label="Minimum mapping quality" help="The minimum mapping quality (default: 0) required for a read to be considered in calculation of variant statistics (samtools mpileup -q)" />
+                <param name="max_depth" type="integer" value="0" min="0" max="2147483647"/>
+            </when> -->
```
wm75, line 96: remove old code

tools/medaka/convert_VCF_info_fields.py
```diff
@@ -0,0 +1,113 @@
...
+# 10/21/2020 - Nathan P. Roach, natproach@gmail.com
+
+import sys
+from math import log10
+
+from scipy.stats import fisher_exact
+
+
+def pval_to_phredqual(pval):
+    return int(round(-10. * log10(pval)))
```
wm75, line 17:
```suggestion
    return round(-10 * log10(pval))
```

but is also used only once, so could be inlined.

wm75 (changes requested):
I think this PR still contains the now abandoned annotateVCF.py script, which can be removed.

Also note that the convert_VCF_info_fields script would emit now one VCF record per ALT allele, but without splitting the ALT field itself.
AFAIU, medaka will never emit multiallelic records so this won't be an issue then, but since the code tries to be able to deal with such records, it should do so correctly.

NatPRoach:
> Also note that the convert_VCF_info_fields script would emit now one VCF record per ALT allele, but without splitting the ALT field itself.
> AFAIU, medaka will never emit multiallelic records so this won't be an issue then, but since the code tries to be able to deal with such records, it should do so correctly.
You're absolutely right, I'll fix that thanks for pointing it out.

tools/medaka/convert_VCF_info_fields.py
```diff
@@ -0,0 +1,113 @@
...
+    for i, line in enumerate(in_vcf):
+        if i == 1:
+            out_vcf.write("##convert_VCF_info_fields=0.1\n")
+        if line[0:2] == "##":
+            if "Number=." in line:
+                assert line[0:11] == "##INFO=<ID="
+                line = line[11:]
+                id_ = line.split(',')[0]
+                to_skip.add(id_)
+                continue
```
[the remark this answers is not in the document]
NatPRoach: I suppose that makes sense. My logic was that "Number=." fields will probably be made in such a way that they will scale the number of entries based on number of variants, but I suppose that's not always true.

bernt-matthias:
Not sure yet what the problem is over here https://github.com/galaxyproject/tools-iuc/issues/3270 .. but it would be great if it would be solved within this PR :)

NatPRoach:
> Not sure yet what the problem is over here #3270 .. but it would be great if it would be solved within this PR :)

That's the current plan. Apologies will get to addressing #3270 soon, been slammed with other stuff but it's on the list.

NatPRoach:
Oh, the issue mentioned in #3270 , that read groups are ignored in medaka_variant pipeline is not an issue on our end. It's a limitation of medaka v1.0.3 : 
Relevant section of `medaka_variant --help`
```
-i  input bam of reads aligned to ref. Read groups are currently ignored,
        so the bam should only contain reads from a single sample.
```

bgruening:
ping @NatPRoach

NatPRoach:
> ping @NatPRoach

Hi @bgruening, I think the status on this is where this thread left it previously. The changes @wm75 requested have been addressed I believe but there was no movement on merging it because of this issue with the medaka annotate function https://github.com/nanoporetech/medaka/issues/192 .

I personally think we should move ahead with the merge anyway, both because our current approach to annotating variants fails with large genomes or high depth of coverage because of memory issues, and we should address that in a version of the wrapper with the old models based on tensorflow 1, since the version after this drops a bunch of models in the upgrade to tensorflow 2. The turn around to bumping a wrapper to the next version of medaka should be quite short after that.

bgruening (approved):

bgruening:
@wm75 are you ok if we proceed here?

davebx (approved):

NatPRoach:
> @wm75 are you ok if we proceed here?
Ping @wm75 . Would appreciate an up or down decision when you get the chance.

wm75 (commented):
Really sorry for not getting round to this again earlier :(

Just tiny comments this time, then everything looks ready to go.

wm75 (approved):

wm75:
Thank you @NatPRoach !

PR #3264, opened 2020-10-26
Remove filter from MAF data table references.

davebx:
FOR CONTRIBUTOR:

This PR removes the input dataset dbkey filter from the maf_indexes reference in the tools that use this table. This allows a user to select any MultiZ alignment that they believe includes the appropriate species, instead of potentially needing to guess which dbkey to use.

### DIFF
--- a/tools/genebed_maf_to_fasta/genebed_maf_to_fasta.xml
+++ b/tools/genebed_maf_to_fasta/genebed_maf_to_fasta.xml
@@ -1,4 +1,4 @@
-<tool id="GeneBed_Maf_Fasta2" name="Stitch Gene blocks" version="1.0.1">
+<tool id="GeneBed_Maf_Fasta2" name="Stitch Gene blocks" version="1.0.1+galaxy0">
     <description>given a set of coding exon intervals</description>
     <macros>
           <import>macros.xml</import>
--- a/tools/genebed_maf_to_fasta/macros.xml
+++ b/tools/genebed_maf_to_fasta/macros.xml
@@ -35,11 +35,9 @@
                         <column name="indexed_for" index="2"/>
                         <column name="exists_in_maf" index="3" />
                         <column name="path" index="4" />
-                        <filter type="data_meta" ref="input1" key="dbkey" column="2" multiple="True" separator=","/>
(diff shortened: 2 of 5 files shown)


davebx:
@bgruening do you think I should update the tool versions for this?

bgruening:
Yes bumping is needed I think.

>This PR removes the input dataset dbkey filter from the maf_indexes reference in the tools that use this table. This allows a user to select any MultiZ alignment that they believe includes the appropriate species, instead of potentially needing to guess which dbkey to use.

Why is this @davebx ... why do you need to guess?

davebx:
@bgruening according to reviewer 2, the main motivation here is that if the user hasn't set a dbkey on their input dataset, or they've mistakenly set the dbkey to a species that doesn't exist in the MAF, the tool form will claim that there are no multi-alignments available. If the filter is removed, they will see all available alignments and be able to specify an appropriate dbkey on their dataset.

bgruening:
> @bgruening according to reviewer 2, the main motivation here is that if the user hasn't set a dbkey on their input dataset, or they've mistakenly set the dbkey to a species that doesn't exist in the MAF, the tool form will claim that there are no multi-alignments available. If the filter is removed, they will see all available alignments and be able to specify an appropriate dbkey on their dataset.

I will not delay your review because of this, but this is a faeture not a bug. This is exactly what those filter should help with - prevent errors. Maybe the Galaxy warning should be better and point people to the dbkex etc ...?

bernt-matthias:
I think this should be fixed on galaxy side. I remember that I started here https://github.com/galaxyproject/galaxy/pull/8599

bernt-matthias:
Why has this been merged? In particular since the Galaxy fix https://github.com/galaxyproject/galaxy/pull/8599 has just been merged?

bernt-matthias:
>  the main motivation here is that if the user hasn't set a dbkey on their input dataset

which is impossible by https://github.com/davebx/tools-iuc/blob/298d46ab4af775100dfaae84c7e44151cd9c0b6e/tools/maf_stats/maf_stats.xml#L23 

.. but it seems reasonable to remove the validator from the input 

> or they've mistakenly set the dbkey to a species that doesn't exist in the MAF

then the Galaxy UI would tell the user and its easy to set the dbkey correctly.

PR #3269, opened 2020-10-30
EDAM terms for Circos, Jbrowse

hexylena:
FOR CONTRIBUTOR:

xref https://github.com/galaxyproject/galaxy/pull/10592

### DIFF
--- a/tools/circos/circos.xml
+++ b/tools/circos/circos.xml
@@ -3,7 +3,6 @@
     <description>visualizes data in a circular layout</description>
     <xrefs>
         <xref type="bio.tools">galactic_circos</xref>
-        <xref type="bio.tools">circos</xref>
         <!--<xref type="bioconda">circos</xref>-->
         <!--<xref type="rrid">RRID:SCR_018207</xref>-->
     </xrefs>
--- a/tools/circos/circos.xml
+++ b/tools/circos/circos.xml
@@ -3,6 +3,7 @@
     <description>visualizes data in a circular layout</description>
     <xrefs>
         <xref type="bio.tools">galactic_circos</xref>
+        <xref type="bio.tools">circos</xref>
         <!--<xref type="bioconda">circos</xref>-->
         <!--<xref type="rrid">RRID:SCR_018207</xref>-->
     </xrefs>
--- a/tools/circos/circos.xml
+++ b/tools/circos/circos.xml
@@ -13,6 +13,10 @@
         <import>macros_tests.xml</import>
     </macros>
     <expand macro="requirements"/>
+    <edam_topics>
+        <edam_topic>topic_0797</edam_topic>
+        <edam_topic>topic_0092</edam_topic>
+    </edam_topics>
     <edam_operations>
         <edam_operation>operation_0337</edam_operation>
     </edam_operations>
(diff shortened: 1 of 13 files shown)


bgruening (approved):
Can you maybe also add the bio.tools identifier? https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-xrefs


Thanks @hexylena!

hexylena:
@bgruening sure, added to jbrowse.

I've also added the "real" circos ID to the circos tool which only had our 'galactic circos' identifier. Is that correct to do? Or should we keep it with just the 'galactic circos' id?

tools/circos/circos.xml
```diff
@@ -3,6 +3,7 @@
     <description>visualizes data in a circular layout</description>
     <xrefs>
         <xref type="bio.tools">galactic_circos</xref>
+        <xref type="bio.tools">circos</xref>
```
bgruening, line 6: ha, good question. Not sure ... but better to have more annotation than less :)
But we should consult with @hmenager, he has integrated this feature if I remember corectly. Maybe move the main bio.tools ID at the top?

hexylena: The underlying tool is Circos, but our wrapper is a significant addition on top of it, hence having it's own ID, so, yea. Strange :)

bgruening: Changed the ordering, hope that is ok. Ready to merge.

hexylena: It is not time sensitive, happy to wait for @hmenager's comments :)

hmenager:
Thanks a lot for pushing this @hexylena !
Here are a few recommendations, up for discussion:
- `EDAM topics` I would have the same for all circos wrappers, since they designate the general subfield that the tool is concerned with. I would annotate circos with `Comparative genomics (topic_0797)`and `Data visualisation (topic_0092)` just like in bio.tools, unless there is a specificity in Galaxy circos which would have it diverge significantly from this. For jbrowse, the topic in bio.tools is `Genomics` and I would add `Data visualisation (topic_0092)`. 
- As a general rule, try to avoid topic_0003 (which is the Topic terms root) for tool annotation.
- `EDAM operations` look fine to me, nothing to say :smiley: 
- `xrefs` are fine, but I would only keep galactic_circos (and not circos) since from what you say @hexylena it is an overlay on circos.

Please tell me if there is anything else I can do to help out.

hmenager:
One thing that comes to my mind is that some of these remarks I made could be useful as documentation. I can help out with that as well.

hexylena:
Hey @hmenager, that's incredibly useful advice! Thank you !

> remarks I made could be useful as documentation

Yes, definitely!

hmenager:
> Yes, definitely!

@hexylena Would a PR for [this documentation page](https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#) the most appropriate location?

hexylena:
Yeah that's definitely a good place for it!

man. 2. nov. 2020, 17:36 skrev Hervé Ménager <notifications@github.com>:

bernt-matthias:
As a follow up: I just found this old one: https://github.com/galaxyproject/tools-iuc/pull/1897 .. could be interesting to get EDAM for more tools

bgruening:
@hexylena is that one ready?

hexylena:
Yep

bgruening:
Danki!

PR #3273, opened 2020-11-02
some awks seem to have strange precidence rules.  Test for this alrea…

bwlang:
…dy passed but did not represent the situation in the wild...maybe not worth adding an explicit awk dependency.

FOR CONTRIBUTOR:
* X ] - License permits unrestricted use (educational + commercial)

### DIFF
--- a/tool_collections/samtools/samtools_view/samtools_view.xml
+++ b/tool_collections/samtools/samtools_view/samtools_view.xml
@@ -153,11 +153,11 @@
                         ## not dealing with all of the reads in the indexed
                         ## file. We have to do an extra pass over the input to
                         ## count the reads to subsample.
-                        sample_fragment=`samtools view -c $std_filters infile $reg_filters | awk '{s=\$1} END {frac=s/${mode.subsample_config.subsampling_mode.target}; print(frac > 1 ? $seed+1/frac : ".0")}'` &&
+                        sample_fragment=`samtools view -c $std_filters infile $reg_filters | awk '{s=\$1} END {frac=s/${mode.subsample_config.subsampling_mode.target}; print(frac > 1 ? $seed+(1/frac) : ".0")}'` &&
                     #else:
                         ## We can get the count of reads to subsample using
                         ## an inexpensive call to idxstats.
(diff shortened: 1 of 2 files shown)


wm75 (approved):

wm75:
Hilarious finding.
So just for my understanding, why doesn't the `gawk` requirement protect against this? Which awks behave like this?

tool_collections/samtools/samtools_view/samtools_view.xml
```diff
@@ -153,11 +153,11 @@
                         ## not dealing with all of the reads in the indexed
                         ## file. We have to do an extra pass over the input to
                         ## count the reads to subsample.
-                        sample_fragment=`samtools view -c $std_filters infile $reg_filters | awk '{s=\$1} END {frac=s/${mode.subsample_config.subsampling_mode.target}; print(frac > 1 ? $seed+1/frac : ".0")}'` &&
+                        sample_fragment=`samtools view -c $std_filters infile $reg_filters | awk '{s=\$1} END {frac=s/${mode.subsample_config.subsampling_mode.target}; print(frac > 1 ? $seed+(1/frac) : ".0")}'` &&
```
bernt-matthias, line 156: Maybe we should use `gawk` .. given that we have the `gawk` requirement already? I would also suggest specifying an explicit version for gawk.

bernt-matthias (changes requested):

tool_collections/samtools/samtools_view/samtools_view.xml
bwlang: @bernt-matthias  i didn't see the gawk dependency in macros.xml... where are you finding that dependency?

wm75: Not 100% sure I like this. Potentially makes it harder to ignore the requirement (during development, testing) cause you now need gawk in your PATH (not just awk).

Conversely, since gawk comes with an `awk` symlink in its conda env I still don't understand how using gawk is supposed to protect against anything.

bernt-matthias: If `awk` is just a symlink it should be fine (i.e. make no difference). 

Is there a way to show that there is actually a difference? 

Anyway, having explicit precedence is fine, so I'm not objecting to the PR. Just want to understand what's going on.

wm75: Same here :)

PR #3284, opened 2020-11-09
Pear update

mthang:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/proteinortho/proteinortho_macros.xml
+++ b/tools/proteinortho/proteinortho_macros.xml
@@ -1,8 +1,8 @@
 <?xml version="1.0"?>
 <macros>
    <token name="@TOOL_VERSION@">6.3.1</token>
-   <token name="@WRAPPER_VERSION@">2</token>
-   <token name="@PROFILE@">20.09</token>
+   <token name="@WRAPPER_VERSION@">0</token>
+   <token name="@PROFILE@">22.05</token>
    <xml name="citations">
         <citations>
             <citation type="doi">10.1186/1471-2105-12-124</citation>
@@ -18,10 +18,10 @@
     <xml name="requirements">
         <requirements>
             <requirement type="package" version="@TOOL_VERSION@">proteinortho</requirement>
-            <requirement type="package" version="2.1.4">diamond</requirement>
-            <requirement type="package" version="2.13.0">blast</requirement>
-            <requirement type="package" version="377">ucsc-blat</requirement>
-            <requirement type="package" version="1422">last</requirement>
+            <requirement type="package" version="2.1.8">diamond</requirement>
+            <requirement type="package" version="2.15.0">blast</requirement>
+            <requirement type="package" version="445">ucsc-blat</requirement>
+            <requirement type="package" version="1519">last</requirement>
         </requirements>
     </xml>
     <xml name="version_command">


tools/pear/pear.xml
```diff
@@ -129,6 +130,20 @@
         <data format="input" name="discarded_reads" from_work_dir="pear.discarded.fastq" label="${tool.name} on ${on_string}: Discarded reads">
             <filter>'discarded' in outputs</filter>
         </data>
+
+        <data name="logout" format="txt" label="${tool.name} on ${on_string}.log">
```
bernt-matthias, line 134: Please change to the more consistent label: `"${tool.name} on ${on_string}: log"`

Please make the indentation consistent.

What exactly is the action doing? We also would need a test for this.

What are the

bernt-matthias (changes requested):

tools/pear/pear.xml
mthang: HI @bernt-matthias,
Thank you for looking after this! I just updated the pear.xml wrapper at my end and uploaded a file (F3D0.log) to the test-data folder. https://github.com/QFAB-Bioinformatics/tools-iuc/blob/pear_update/tools/pear/pear.xml 

To answer your question about "what exactly is the action doing"?, the additional code added to the pear.xml is used to capture the console output of the pear program and output it as a text/log file instead of just printing the console output in the history. You can see the example output in the F3D0.log as a physical file in the test-data folder in my repo.

I am not sure if I need to make the open pull request again for the latest changes.

Many thanks,
Mike

bernt-matthias: Now you redefine the `name` attribute of the output (see linter error). 

> to capture the console output of the pear program

I have to ask "capture it where"? Unfortunately I still do not get the intention of the action. 

Currently the stdout of pear is redirected to `logout` over here: https://github.com/galaxyproject/tools-iuc/blob/e8e64dbfd6cc5b73e37c5417be63777f505a5ce0/tools/pear/pear.xml#L50

If you like to you may use `| tee '$logout'` then the output goes to the data set in the history and you still see it in stdout. 

Btw. Currently the file input and output parameters are either unquoted or use double quotes. Could you change this to single quotes?

What I can suggest is to add the file that you are referring to to the `<test>` - this definitely makes sense.

> I am not sure if I need to make the open pull request again for the latest changes.

You can keep adding commits to this branch. They will appear in this PR.

mthang: Hi @bernt-matthias,
Thank you for your quick response !  I can update the pear.xml wrapper to use single quotes instead of the double quotes or unquoted. I understand that my request of the enhancement of this pear wrapper can be confusing. The main reason why I need that enhancement is to create an aggregate statistic of all paired-end file processed by pear. I have made another tool to do the aggregation after the main pear wrapper. 

The example output of my tool can be seen here https://raw.githubusercontent.com/QFAB-Bioinformatics/metaDEGalaxy/master/pear_stats/test-data/test_output.txt . As you see, my tool retrieves only few lines from the pear stdout, but the existing pear wrapper does not have the console output printed to a physical file. Therefore, my pear can't be used without the pear stdout in physical files. does it make sense ?

bernt-matthias: I understand. But the problem is that the log file is already added to the history, isn't it? So the action is not needed at al in my opinion.

Furthermore the action seems nonsensical to me. In case the conditional select is `library.type` is `"paired"` the filter checks again if `library.type=="paired" (which is always true in this case) .. same for `"paired_collection"`. 

So my suggestion would be:

- restore the original state for `logout`
- extend the test to check `logout` against the file `F3D0.log` ()
- improve the quoting of input and output files in the command block

mthang: Hi @bernt-matthias,
Thank you for quick response! You are correct. The log output of pear is added to the history (please see the attached screenshot), but I do not have a clue how to parse the data out of the history. Perhaps, you can give a pointer on how to parse data from the pear log in the history. I am not entirely sure how to create a tool to retrieve the data between "format fastsanger, database?" and "PEAR v0.9.6 [January 15,2015] section in the history. if there is already a tool to do that, please let me know. I am happy to use the official IUC pear wrapper and just update my wrapper to parse the pear log using your method.

PEAR_History

Many thanks,
Mike

bernt-matthias: I guess the following tools can do the task:

- Search in textfiles (grep) 
- Text transformation with sed

when applied to the `log file`. 

Still it would be great if you could finish the PR as I outlined above.

PR #3290, opened 2020-11-10
Fix for whitespace error on cloud instance

astrovsky01:
FOR CONTRIBUTOR:
On cloud instance, there's a single whitespace removed from end of second line which throws the test off. This just allows for that.

### DIFF
--- a/tools/column_order_header_sort/column_order_header_sort.xml
+++ b/tools/column_order_header_sort/column_order_header_sort.xml
@@ -25,7 +25,7 @@
         <test>
             <param name="input_tabular" value="in_1.tabular" ftype="tabular"/>
             <param name="key_column" value="1"/>
-            <output name="output_tabular" file="out_1.tabular" ftype="tabular"/>
+            <output name="output_tabular" file="out_1.tabular" ftype="tabular" compare="sim_size" delta="1"/>
         </test>
     </tests>
     <help>


bernt-matthias:
Do you have a link to the failing test (I would like to understand the problem)?

Do you know what the `gawk` requirement is good for?

astrovsky01:
Can't do the link, but here is the error:
```
Output output_tabular:  different than expected, difference (using diff):
( /tmp/tmpp56fu2lvout_1.tabular v. /tmp/tmpmfe_ftapout_1.tabular )
--- local_file
+++ history_data
@@ -1,4 +1,4 @@
 #KEY	a	b	c
 one	1-3	1-1	1-2
 two	1-6	1-4	1-5
-three	1-9	1-7	1-8+three	1-9	1-7	1-8
```

bernt-matthias:
Did you check that its really a space? Or might it be a different white space character?

I guess (wild guess) awk may treat the newline on the final line differently sometimes. So I suggest to switch to a python only implementation: https://github.com/astrovsky01/tools-iuc/pull/1

While coding I had the question in mind if the `#` character that is assumed by Galaxy  as header prefix for tabular data needs special treatment.

astrovsky01:
@bernt-matthias I said whitespace, but s0eciifically I did mean the newline, yeah. Trying your test now.

bernt-matthias:
> While coding I had the question in mind if the `#` character that is assumed by Galaxy as header prefix for tabular data needs special treatment.

Should we consider this? Since 

1. `#` is one of the first characters in the sort order the first column will remain the first if `key_column` is left empty.
2. If another `key_column != 1` is chosen then the header will be messed up and the sort order will be wrong.

bgruening:
ping @astrovsky01

PR #3296, opened 2020-11-11
fastp: fastqsanger datatype for tests

astrovsky01:
FOR CONTRIBUTOR:
Test modification


tools/fastp/fastp.xml
```diff
@@ -372,7 +372,7 @@ mv first${ext} '${out1}'
             <param name="in1" ftype="fastq" value="R1.fq"/>
             <param name="single_paired_selector" value="single"/>
             <param name="adapter_sequence1" value="ATCG"/>
-            <output name="out1" ftype="fastq" file="out_a.fq"/>
+            <output name="out1" ftype="fastqsanger" file="out_a.fq"/>
```
bernt-matthias, line 375: If I get the code right the output format should be equal to the input format. So why should this be the more specific fastqsanger?

astrovsky01: @bernt-matthias It looks like the problem is actually just with the same files being used as multiple filetypes when multiple tests are run simultaneously. I'm going to change the PR to reflect that by duplicating the files and using different names.

tools/fastp/fastp.xml
```diff
@@ -369,14 +369,14 @@ mv first${ext} '${out1}'
         </test>
         <!-- Ensure custom adapter works -->
         <test expect_num_outputs="2">
-            <param name="in1" ftype="fastq" value="R1.fq"/>
+            <param name="in1" ftype="fastq" value="R1.fastq"/>
             <param name="single_paired_selector" value="single"/>
             <param name="adapter_sequence1" value="ATCG"/>
-            <output name="out1" ftype="fastq" file="out_a.fq"/>
+            <output name="out1" ftype="fastqr" file="out_a.fq"/>
```
almahmoud, line 375: typo?

tools/fastp/fastp.xml
bernt-matthias: OK, I was not aware that the fastp tests are failing, but probably I'm wrong,

astrovsky01: It was only when tests are run in parallel

bernt-matthias: Huh. Tests running in parallel -- exciting. How?

astrovsky01: shed-tools test with ephemeris

bernt-matthias: Cool. Question is if this is the way to fix the problem. I imaginge that a lot of tools (if not most) would require changes.

astrovsky01: You may be right, but this also doesn't change anything about this tool while  we figure out how to solve it longer-term

tools/fastp/fastp.xml
```diff
@@ -398,16 +398,16 @@ mv first${ext} '${out1}'
         </test>
         <!-- Ensure paired-end fastq works -->
         <test expect_num_outputs="3">
-            <param name="in1" ftype="fastq" value="bwa-mem-fastq1.fq"/>
-            <param name="in2" ftype="fastq" value="bwa-mem-fastq2.fq"/>
+            <param name="in1" ftype="fastq" value="bwa-mem-fastq1.fastq"/>
```
bgruening, line 401: the old files are still in the repo?

astrovsky01: Yes, they're called as the original files names in other tests so that the same files are not called as both fastq and fastqsanger.

bernt-matthias: Btw. would symlinks be an option?

bgruening:
Ping @astrovsky01 and @bernt-matthias

bernt-matthias:
I labeled this as WIP and suggest to not merge this until we know if this is a problem of the testing approach or really a problem of the tool (which needs to be described .. currently I still don't get it).

bernt-matthias:
Going to close this. File extension should be irrelevant.

Issue #3272, opened 2020-11-02
dynamically set number of chunks


bernt-matthias:
https://github.com/TGAC/earlham-galaxytools/pull/132#discussion_r509393624

PR #3299, opened 2020-11-13
make number of chunks dynamic

bernt-matthias:
Computed as ~max~min(40, 4*log_2(ntools+1)) where ntools is the number of XML files in `tools/`, `tool_collections/`, and `data_managers` that do not contain `"macro"`.

Note that the formula gives 40 for the current number of tools

Fixes https://github.com/galaxyproject/tools-iuc/issues/3272

FOR CONTRIBUTOR:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -53,10 +53,10 @@ jobs:
       id: get-chunks
       run: |
         nchunks=$(find tools/ tool_collections/ data_managers/ -iname '*xml' | grep -v macro | wc -l)
-        if [[ $nchunks -gt $MAX_CHUNKS ]]; then
+        if [ "$nchunks" -gt "$MAX_CHUNKS" ]; then
             nchunks=$MAX_CHUNKS
         fi
-        chunk_list=$(echo "["$(seq -s ", " 1 $nchunks)"]")
+        chunk_list='['$(seq -s ", " 0 $(($nchunks - 1)))']'
         echo "::set-output name=chunks::$chunk_list"
   test:
     name: Test tools
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,6 +7,7 @@ on:
     types: [run-all-tool-tests-command]
 env:
   GALAXY_RELEASE: release_20.09
+  MAX_CHUNKS: 40
 jobs:
   setup:
     name: Setup cache
@@ -51,13 +52,11 @@ jobs:
     - name: Get approximate number of tools and compute number of chunks
       id: get-chunks
       run: |
-        ntools=$(find tools/ tool_collections/ data_managers/ -iname '*xml' | grep -v macro | wc -l)
-        log=$(printf "%.0f" $(echo "l($ntools+1)/l(2)" | bc -l))
-        nchunks=$(echo "4 * $log" | bc)
-        if [[ $nchunks -gt 39 ]]; then
-            nchunks="39"
+        nchunks=$(find tools/ tool_collections/ data_managers/ -iname '*xml' | grep -v macro | wc -l)
+        if [[ $nchunks -gt $MAX_CHUNKS ]]; then
+            nchunks=$MAX_CHUNKS
         fi
-        chunk_list=$(echo "["$(seq -s ", " 0 $nchunks)"]")
+        chunk_list=$(echo "["$(seq -s ", " 1 $nchunks)"]")
         echo "::set-output name=chunks::$chunk_list"
   test:
     name: Test tools
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,6 +12,8 @@ jobs:
     name: Setup cache
     if: github.repository_owner == 'galaxyproject'
     runs-on: ubuntu-latest
+    outputs:
+      chunks: ${{ steps.get-chunks.outputs.chunks }}
     strategy:
       matrix:
         python-version: [3.7]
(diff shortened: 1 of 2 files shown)


nsoranzo (commented):
That's fantastic, thanks Matthias!
What do you think of a simpler linear `min(ntools, 40)` function? And maybe `min(ntools, 4)` for PRs.
With your formula, there would be 4 chunks for 1 tool.

Edit: modified my suggested formulas from the wrong `max(ntools, N)` to `min(ntools, N)`

bernt-matthias:
> What do you think of a simpler linear max(ntools, 40) function?

I just did not want to waste resources. I was hoping that the setup overhead of the test job should take not more than the actual test. In other words I tried to avoid the "1 tool per chunk" case. 

> And maybe max(ntools, 4) for PRs.

I just had the galaxyproteomics openMS in tool suite in mind which has >100 tools .. i.e. the adaption should work for small and large tool sets.

bernt-matthias:
Note that I tested the CI workflow quite intensively. The changes in PR are more or less copy paste ..

nsoranzo:
> > What do you think of a simpler linear max(ntools, 40) function?
> 
> I just did not want to waste resources. I was hoping that the setup overhead of the test job should take not more than the actual test. In other words I tried to avoid the "1 tool per chunk" case.

Please note that the identity function generates less chunks than your formula till ntools=15 : http://fooplot.com/#W3sidHlwZSI6MCwiZXEiOiI0KmxvZyh4KzEpL2xvZygyKSIsImNvbG9yIjoiIzAwMDAwMCJ9LHsidHlwZSI6MCwiZXEiOiJ4IiwiY29sb3IiOiIjMDAwMDAwIn0seyJ0eXBlIjoxMDAwLCJ3aW5kb3ciOlsiMCIsIjQwIiwiMCIsIjQwIl19XQ--

> > And maybe max(ntools, 4) for PRs.
> 
> I just had the galaxyproteomics openMS in tool suite in mind which has >100 tools .. i.e. the adaption should work for small and large tool sets.

I think we want to use way less chunks for PR tests, since these are run much more frequently than the weekly CI tests (which is why we are using 4 chunks now).

bernt-matthias:
> I think we want to use way less chunks for PR tests, since these are run much more frequently than the weekly CI tests (which is why we are using 4 chunks now).

True, but if we find a good formula we will use more chunks only if needed. 

For instance the OpenMS tools execute ~600 tests needing 1.5h (approx 1/5 of the IUC tests which also need 1.5h). In my opinion some more power for such cases is certainly needed. 

Therefore I tried to come up with a more clever way to compute the number of chunks .. seems that I failed :)

Maybe just linear with a factor `min(40, ncores / X)` for PR?

nsoranzo:
> Therefore I tried to come up with a more clever way to compute the number of chunks .. seems that I failed :)

I also failed in writing what I had in mind, which was `min(ntools, 40)`, not max, sorry about that!
The Biohackathon has fried my neurons!

> Maybe just linear with a factor `min(40, ncores / X)` for PR?

Not sure what are `ncores` and `X` here?

bernt-matthias:
> Not sure what are ncores and X here?

LOL. I meant `min(40, ntools / X)` and x being some constant, e.g. 2 or 4, ...

nsoranzo:
> > Not sure what are ncores and X here?
> 
> LOL. I meant `min(40, ntools / X)` and x being some constant, e.g. 2 or 4, ...

OK, then we agree for X=1 :D , which works also for ntools=1. Any way, I suggest to start with something like that (with 40 replaced with something smaller for PRs), and see how it goes.

bernt-matthias:
OK. I introduced a variable `MAX_CHUNKS` which is 4 by default for PR and 40 for CI. So other repos may easily change it.

.github/workflows/ci.yaml
```diff
@@ -43,6 +46,18 @@ jobs:
...
         touch tool.xml
         PIP_QUIET=1 planemo test --galaxy_python_version ${{ matrix.python-version }} --no_conda_auto_init --galaxy_branch $GALAXY_RELEASE
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 1
+    - name: Get approximate number of tools and compute number of chunks
+      id: get-chunks
+      run: |
+        nchunks=$(find tools/ tool_collections/ data_managers/ -iname '*xml' | grep -v macro | wc -l)
+        if [[ $nchunks -gt $MAX_CHUNKS ]]; then
```
nsoranzo, line 56:
```suggestion
        if [ "$nchunks" -gt "$MAX_CHUNKS" ]; then
```

.github/workflows/pr.yaml
```diff
@@ -90,6 +93,16 @@ jobs:
       with:
         name: Workflow artifacts
         path: changed_repositories.list
+    - name: Get number of changed tools and compute number of chunks
+      id: get-chunks
+      run: |
+        planemo ci_find_tools --output changed_tools.list $(cat changed_repositories.list)
+        nchunks=$(wc -l changed_tools.list)
+        if [[ $nchunks -gt $MAX_CHUNKS ]]; then
```
nsoranzo, line 101:
```suggestion
        if [ "$nchunks" -gt "$MAX_CHUNKS" ]; then
```

.github/workflows/ci.yaml
```diff
@@ -43,6 +46,18 @@ jobs:
...
+            nchunks=$MAX_CHUNKS
+        fi
+        chunk_list=$(echo "["$(seq -s ", " 1 $nchunks)"]")
```
nsoranzo, line 59:
```suggestion
        chunk_list='['$(seq -s ", " 0 $(($nchunks - 1)))']'
```

.github/workflows/pr.yaml
```diff
@@ -90,6 +93,16 @@ jobs:
...
+            nchunks=$MAX_CHUNKS
+        fi
+        chunk_list=$(echo "["$(seq -s ", " 1 $nchunks)"]")
```
nsoranzo, line 104:
```suggestion
        chunk_list='['$(seq -s ", " 0 $(($nchunks - 1)))']'
```

nsoranzo (approved):

.github/workflows/pr.yaml
```diff
@@ -90,6 +93,16 @@ jobs:
       with:
         name: Workflow artifacts
         path: changed_repositories.list
+    - name: Get number of changed tools and compute number of chunks
+      id: get-chunks
+      run: |
+        planemo ci_find_tools --output changed_tools.list $(cat changed_repositories.list)
+        nchunks=$(wc -l changed_tools.list)
```
nsoranzo, line 100: We need to not have the filename in the wc output
```suggestion
        nchunks=$(wc -l < changed_tools.list)
```

nsoranzo:
Thanks @bernt-matthias , this is awesome!

PR #3300, opened 2020-11-13
fixing bugs and adding test to IDR download tool

sunyi000:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/idr_download/idr_download_by_ids.xml
+++ b/tools/idr_download/idr_download_by_ids.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
-<tool id="idr_download_by_ids" name="IDR Download" version="0.43" profile="18.09">
-    <description>- download images from the Image Data Resource using image IDs</description>
+<tool id="idr_download_by_ids" name="IDR/OMERO Download" version="0.43" profile="18.09">
+    <description>- download images from any OMERO instance using image IDs</description>
     <macros>
         <xml name="region_spec" token_pos="upper-left corner">
             <param name="x_coord" type="integer" value="0" min="0"
@@ -22,12 +22,14 @@
     <command detect_errors="exit_code"><![CDATA[
         
         #if $omero_instance_type.omero_instance == "priv":
+            ## the user wants to use a non-public OMERO instance
+            ## check if credentials are set in the user-preferences, if not warn the user and exit
             #set $username = $__user__.extra_preferences.get('omero_account|username', "")
             #set $password = $__user__.extra_preferences.get('omero_account|password', "")
 
-            #if $username == "" or $password =="":
-                echo "OMERO connection credentials are empty. Set your credentials via: User -> Preferences -> Manage Information"  &&
-                exit 1
+            #if $omero_instance_type.galaxy_test_param != 'true' and ($username == "" or $password ==""):
+                echo "OMERO connection credentials are empty. Set your credentials via: User -> Preferences -> Manage Information" 1>&2 &&
+                exit 1 &&
             #end if
         #end if
         
@@ -51,13 +53,28 @@
         $skip_failed
         $download_tar
         #if $omero_instance_type.omero_instance =='priv':
-            -h '$omero_host'
-            $omero_secured
-            -u '$username'
-            -p '$password'
+            -oh '$omero_instance_type.omero_host'
+            $omero_instance_type.omero_secured
+            -cf '$credentials'
         #end if
         2> >(tee -a $out_log >&2)
     ]]></command>
+    <configfiles>
+        <configfile name="credentials"><![CDATA[
+#if $omero_instance_type.omero_instance =='priv' and $omero_instance_type.galaxy_test_param == 'true':
+    ## as a test for a private instance we actually use a public instance, but with credentials
+    #set $username = 'public'
+    #set $password = 'public'
+#else:
+    #set $username = $__user__.extra_preferences.get('omero_account|username', "")
+    #set $password = $__user__.extra_preferences.get('omero_account|password', "")
+#end if
+{
+    "username": "$username",
+    "password": "$password"
+}
+        ]]></configfile>
+    </configfiles>
 <inputs>
         <conditional name="omero_instance_type">
             <param name="omero_instance" type="select" label="Which OMERO instance to connect?"
@@ -68,13 +85,13 @@
             </param>
             <when value="priv">
                 <param name="omero_host" type="text" label="OMERO host URL">
-                    <validator type="regex" message="Enter a valid host location, for example, your.omero.server">
-                        [.A-Za-z0-9_-]
-                    </validator>
+                    <validator type="regex" message="Enter a valid host location, for example, your.omero.server">^[a-zA-Z0-9._-]*$</validator>
+                    <validator type="expression" message="No two dots (..) allowed">'..' not in value</validator>
                 </param>
                 <param name="omero_secured" type="boolean" label="Seured connection?" checked="true" truevalue="--omero-secured" falsevalue=""
                         help="Select Yes if your OMERO instance is running with SSL, otherwise select No">
                 </param>
+                <param name="galaxy_test_param" type="hidden" value="false" />
             </when>
             <when value="idr" />
         </conditional>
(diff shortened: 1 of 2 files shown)


sunyi000:
@bgruening Could you have a look see if that's the correct way to use environment variables?
Also, not very sure how to assign values to env variables in test? I'm stiill searching for examples

Thanks very much

tools/idr_download/idr_download_by_ids.xml
```diff
@@ -20,15 +20,10 @@
...
     </requirements>
     <command detect_errors="exit_code"><![CDATA[
-        
-        #if $omero_instance_type.omero_instance == "priv":
-            #set $username = $__user__.extra_preferences.get('omero_account|username', "")
-            #set $password = $__user__.extra_preferences.get('omero_account|password', "")
-
-            #if $username == "" or $password =="":
+        #if $omero_instance_type.omero_instance == "priv":            
+            #if "\$GALAXY_SECURITY_USERNAME" == "" or "\$GALAXY_SECURITY_PASSWORD" =="":
```
bgruening, line 24: Thats not what I meant. Maybe this is the most easiest way, can you please do it like https://github.com/galaxyproject/tools-iuc/blob/aac23c58bf6095860b49de0bb117949f0184aa78/tools/pyega3/pyega3.xml#L37

E.g. keep the credentials in a file.

tools/idr_download/idr_download_by_ids.xml
```diff
@@ -300,6 +303,29 @@
                     </assert_contents>
             </output>
         </test>
+        <!-- test for other omero instance -->
+        <test>
+            <conditional name="omero_instance_type">
+                <param name="omero_instance" value="priv" />
+                <param name="omero_host" value="idr.openmicroscopy.org"/>
+                <param name="omero_secured" value="--omero-secured" />
```
bgruening, line 311: this should be `true`

sunyi000:
Thanks @bgruening . can you have a look again? 
I need some help with the last ''test'', how do we set username and password in a test? I looked at the ega tool test, but still unclear about how their tests are passed..

Thanks

tools/idr_download/idr_download_by_ids.xml
```diff
@@ -69,7 +79,7 @@
             <when value="priv">
                 <param name="omero_host" type="text" label="OMERO host URL">
                     <validator type="regex" message="Enter a valid host location, for example, your.omero.server">
-                        [.A-Za-z0-9_-]
+                        ^[.A-Za-z0-9_-]*$
                     </validator>
                 </param>
                 <param name="omero_secured" type="boolean" label="Seured connection?" checked="true" truevalue="--omero-secured" falsevalue=""
```
beatrizserrano, line 85:
```suggestion
                <param name="omero_secured" type="boolean" label="Secured connection?" checked="true" truevalue="--omero-secured" falsevalue=""
```

tools/idr_download/idr_download_by_ids.xml
```diff
@@ -51,13 +51,23 @@
...
             $omero_secured
-            -u '$username'
-            -p '$password'
+            -cf '$credentials'
         #end if
         2> >(tee -a $out_log >&2)
     ]]></command>
+    <configfiles>
+        <configfile name="credentials"><![CDATA[
+#set $username = $__user__.extra_preferences.get('omero_account|username', "")
```
bgruening, line 62: @sunyi000 the default username is here "" and the password as well. This is then used in the tests.

beatrizserrano:
Should we rename this tool to include `OMERO` in it? e.g. `IDR/OMERO download` or sth like that...

sunyi000:
all green @bgruening @beatrizserrano  :)
shall we change the name from the tool config file? We probably need to change the help text as well ?

THanks

tools/idr_download/idr_download_by_ids.xml
```diff
@@ -22,8 +22,8 @@
     <command detect_errors="exit_code"><![CDATA[
         
         #if $omero_instance_type.omero_instance == "priv":
-            #set $username = $__user__.extra_preferences.get('omero_account|username', "")
-            #set $password = $__user__.extra_preferences.get('omero_account|password', "")
+            #set $username = $__user__.extra_preferences.get('omero_account|username', "public")
+            #set $password = $__user__.extra_preferences.get('omero_account|password', "public")
```
bgruening, line 26: the if condition will never be triggered as the default value is now public.

bgruening: You could move the default value maybe in the python file, or change the if condition here.

tools/idr_download/idr_download_by_ids.xml
```diff
@@ -68,9 +78,9 @@
             </param>
             <when value="priv">
                 <param name="omero_host" type="text" label="OMERO host URL">
-                    <validator type="regex" message="Enter a valid host location, for example, your.omero.server">
-                        [.A-Za-z0-9_-]
-                    </validator>
+<!--                     <validator type="regex" message="Enter a valid host location, for example, your.omero.server">
```
bgruening, line 81: this validator need to be fixed I think.

bgruening (commented):
I think we can change `IDR Download` to OMERO/IDR download and

`    <description>- download images from the Image Data Resource using image IDs</description>` to

`    <description>download images from any OMERO instance using image IDs</description>`

tools/idr_download/idr_download_by_ids.xml
sunyi000: @bgruening this is where I'm a bit unclear. If I leave the default 'public' to empty, or change the if condition how do we set username password values in <test>

 so for the last test case, how do I set username=public, and password= public for that particular case?

the ega example, set default username password in the xml to some EBI public credentials..

Thanks,

bgruening: At some point you need to set the `public` to pass the IDR test, question is just where, left? So maybe in your python file. I don't have a real preference.

sunyi000: > the if condition will never be triggered as the default value is now public.

can you explain it a bit more, I thought the "if" is testing the param name="omero-instance-type" , so I think it will be triggered?

Thanks

bgruening: With the `if` you just want to check if the user knows about the user-preference trick. And you assume if nothing is checked that the user does not know and you want to give them an error message that indicates "go to user-preference, change settings". If you return as default `public` or if you check in this `if` for `public` the user is not allowed to set `public` in the user-preferences.

sunyi000: yes, it just need some intial values for the tests.. it can be anything.
but the credential file configfile is only passed to pythong script if user select param ="priv" via the interface.
and in the python I checked if the credential file is passed, if yes it means private omero instance, if not the python set connection to IDR by default...

tools/idr_download/idr_download_by_ids.xml
sunyi000: i need some help here.. is ^[.A-Za-z0-9_-]$ correct? i use some online regex tester, it says valid :.....

tools/idr_download/idr_download_by_ids.xml
bgruening: > and in the python I checked if the credential file is passed, if yes it means private omero instance

Can you check in this stage of password and username is empty and use the IDR defaults?

sunyi000: ye it is in line 120,121 of the python file...

tools/idr_download/idr_download_by_ids.xml
sunyi000: ok..spaces are the culprit....I left too many spaces before the regex...

sunyi000:
@bgruening I made another round of changes... all your comments are addressed I believe.... :) so around of review please

tools/idr_download/idr_download_by_ids.xml
```diff
@@ -26,8 +26,9 @@
             #set $password = $__user__.extra_preferences.get('omero_account|password', "")
 
             #if $username == "" or $password =="":
+                #set username = 'public'
```
bgruening, line 29: I would not do it and bail out here. If the user has chosen `priv` they really need to set those values in the user-settings. It's good to fail here explicitly I think.

sunyi000: If explicitly fail here putting exit 1.the code won't continue,so test will never succeed. Sorry honestly Im really out of idea and don't know how to proceed here....

sunyi000: I need help here @bgruening .. really out of idea now.. it wont continue if explicitly fail here, and the second  last tests will never succeed

bgruening (approved):

bgruening:
Thanks @sunyi000! That is a great feature!

PR #3307, opened 2020-11-16
VDM plot tool initial commit

fxce:
FOR CONTRIBUTOR:

Hi, 
I would like to submit this plotting tool that is part of a Variant Discovery Mapping (bulk segregant linkage analysis) approach/pipeline for mapping mutations from mutagenesis screens. It is part of a manuscript that is being written (in Dr. Maria Doitsidou group and in collaboration with Dr. Richard Poole). It shares much of the same functionality with the deprecated cloudmap tool and pipeline and may be considered a replacement (at least for the VDM application aspect). I have marked it as new tool for the manuscript to accompany it.

I am new to github and submitting tools, so please let me know if there are any issues so I can rectify them. Thanks. 

Feng


wm75:
Hi @fxce and welcome here.

A couple of general questions before I start going into an in-depth review:

1. You didn't check that item `License permits unrestricted use (educational + commercial)` in the FOR CONTRIBUTOR list.
    Maybe you intended to do so and simply forgot, but this checkbox is not an optional one.

2. Your Galaxy wrapper mentions in the help that there is a standalone R script available at https://github.com/fxce/vdm_plot.
    In that case, it may be better to have a bioconda package built from your standalone script, which you can then require in the Galaxy wrapper xml instead of duplicating your code. If you need help with this, let us know.

3. Just out of curiosity: how is your tool different from the MiModD tool suite, which also started as a replacement for CloudMap (disclaimer: I'm the author of that package). Is it just the Freebayes support?

fxce:
> Hi @fxce and welcome here.
> 
> A couple of general questions before I start going into an in-depth review:
> 
>     1. You didn't check that item `License permits unrestricted use (educational + commercial)` in the FOR CONTRIBUTOR list.
>        Maybe you intended to do so and simply forgot, but this checkbox is not an optional one.
> 
>     2. Your Galaxy wrapper mentions in the help that there is a standalone R script available at https://github.com/fxce/vdm_plot.
>        In that case, it may be better to have a bioconda package built from your standalone script, which you can then require in the Galaxy wrapper xml instead of duplicating your code. If you need help with this, let us know.
> 
>     3. Just out of curiosity: how is your tool different from the MiModD tool suite, which also started as a replacement for CloudMap (disclaimer: I'm the author of that package). Is it just the Freebayes support?

Hi @wm75,

Thanks for your quick reply. Apologies for my initial slow response, had some issues come up last week, but I should be back to be able respond promptly and make changes now.

1. I edited my comment to have the box for License Permits Restricted use option ticked

2. My idea for the standalone was for it to have some extra output options in contrast to the galaxy version was to have it as simple a user experience as possible (ie 1 datatable and 1 plot). Your suggestion of using bioconda is useful, if you have a direct link for instructions, I will check it out and integrate in a future update (I need to gain some familiarity with bioconda first).

3. So the main drive for this tool was to have a tool and pipeline that can be used on public galaxy servers so that users, who might be intimidated by setting up Galaxy locally, could run this whole workflow online and generate the relevant outputs directly for the specific purpose of variant discovery mapping from mutagenesis. The overall workflow will be like cloudmap's which I think is nicely granular so users might be able to manipulate additional intermediate files with other tools in Galaxy.

I was previously aware of the MiModD tool suite but was under the misconception that it was primarily standalone (as opposed to Galaxy-based). I spent the last few days exploring the suite (as Galaxy tools) with my own data more thoroughly and I definitely see the power and efficiency of it, especially for large datasets. However, I think there is a place for my simpler tool that serves as part of a tailored workflow for the kind of variant discovery mapping that we use.

In addition, for the MiMod NasceousMap tool, is there an option for Variant Allele Frequency mapping to do so without an outcross parent (Hawaii)? i.e. we analyze the allele frequency ratios for mutant candidate lines that have been backcrossed with the original background strain- pooled in bulk and sent for sequencing. I only seem to have gotten (No markers found! errors) when I used the VAF option and did not include an unrelated outcross parent.

Actually, my updated plan now is to actually integrate my simpler filtering and plotting tool to support VCFs from GATK (since it's back on public galaxy), Freebayes (default), and also your MiModD Extract Variants (ie BCF->VCF) output.


Feng

wm75:
> My idea for the standalone was for it to have some extra output options in contrast to the galaxy version was to have it as simple a user experience as possible (ie 1 datatable and 1 plot).

Such simplification should be handled by the Galaxy wrapper (the xml) then, which should call the command line tool with appropriate arguments and reasonable default values. You just need to build a command line interface for your script that makes it callable in a flexible way so that the same code becomes useful from the command line and from within Galaxy.

> Your suggestion of using bioconda is useful, if you have a direct link for instructions, I will check it out and integrate in a future update (I need to gain some familiarity with bioconda first).

Bioconda is really not that hard to learn and IMHO you should do so *before* we deploy something here. All you have to do is contribute a bioconda recipe to https://github.com/bioconda/bioconda-recipes, which typically consists of 1 meta.yaml and 1 tiny build.sh script. This recipe then takes over the role of your xml requirements, but not just for Galaxy, but also for command line users (they can then just do `conda install <your_package>` and need not worry about installing R. Documentation for bioconda lives at https://bioconda.github.io/, but it's equally helpful to just study some existing recipes.

PR #3314, opened 2020-11-18
migmap: demonstrate fix by adding java 8

bernt-matthias:
see https://github.com/mikessh/migmap/issues/55#issuecomment-729802798

I would suggest to fix this properly in bioconda, but then I guess we need to invalidate the old container, or?

Also switched to agreessive error detection, because the JAVA exception were not noted by Galaxy.

FOR CONTRIBUTOR:


bgruening:
> I would suggest to fix this properly in bioconda, but then I guess we need to invalidate the old container, or?

galaxy should pull the latest container, so the new container should be preferred

bernt-matthias:
> galaxy should pull the latest container, so the new container should be preferred

done

tools/migmap/migmap.xml
```diff
@@ -7,7 +7,7 @@
     <requirements>
         <requirement type="package" version="@TOOL_VERSION@">migmap</requirement>
     </requirements>
-    <command detect_errors='exit_code'><![CDATA[
+    <command detect_errors='aggressive'><![CDATA[
```
mvdbeek, line 10: Was that just for triggering a test or is this really necessary ?

bernt-matthias: Switched to agreessive error detection, because the JAVA exceptions (shown here https://github.com/mikessh/migmap/issues/55#issuecomment-729802798) were not noted by Galaxy.

If you don't like this we could switch stdio regex?

mvdbeek: No that's fine.

mvdbeek (approved):

bernt-matthias:
Odd that the PR workflow failed for the merge commit. Guess the commit range is wrong .. since unrelated (recently changed) tools fail.

mvdbeek:
Are you looking at your fork ? I don't see failed tests on the IUC

mvdbeek:
https://github.com/galaxyproject/tools-iuc/actions/runs/374515120 should be the run that corresponds to this merge commit.

PR #3234, opened 2020-10-06
Update to HyPhy 2.5.21

davebx:
FOR CONTRIBUTOR:
* [x] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [x] - License permits unrestricted use (educational + commercial)
* [x] - This PR updates an existing tool or tool collection

PR #3315, opened 2020-11-18
meme, vg_convert: fix openmpi/xg error by setting TMPDIR

bernt-matthias:
fixes this problem:

```
A call to mkdir was unable to create the desired directory:

  Directory: /ompi.530361e930c3.1001
  Error:     Permission denied

Please check to ensure you have adequate permissions to perform
the desired operation.
```

Not sure if we should also add `-p \${GALAXY_SLOTS:-1}`, how does MPI play on out Galaxy instances?

FOR CONTRIBUTOR:

### DIFF
--- a/tools/meme/meme.xml
+++ b/tools/meme/meme.xml
@@ -8,6 +8,7 @@
 export TMPDIR=/tmp;
 
 meme '$input1'
+-p \${GALAXY_SLOTS:-1}
 -o '${html_outfile.files_path}'
 -nostatus
 -maxsize 1000000
--- a/tools/meme/meme.xml
+++ b/tools/meme/meme.xml
@@ -5,6 +5,8 @@
     </macros>
     <expand macro="requirements" />
     <command detect_errors="exit_code"><![CDATA[
+export TMPDIR=/tmp;
+
 meme '$input1'
 -o '${html_outfile.files_path}'
 -nostatus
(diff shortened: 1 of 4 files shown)


tools/meme/meme.xml
```diff
@@ -5,7 +5,10 @@
     </macros>
     <expand macro="requirements" />
     <command detect_errors="exit_code"><![CDATA[
+export TMPDIR=/tmp;
```
nsoranzo, line 8:
```suggestion
export TMPDIR=\${TMPDIR:-.};
```

See https://github.com/galaxyproject/tools-iuc/pull/3234/files#diff-90bfa2f2d9fe3174844eaf5edc72bdd6418e85a2cf68ee69846b80244b8cd4b9

bernt-matthias: Hmm, but then this might write to the job working dir. Isnt there a variable set by galaxy that we can use? Or just `$TMP`. Then export TMPDIR=\${TMPDIR:-$TMP}; might be an option.

nsoranzo: Galaxy already sets TMP, TMPDIR and TEMP if instructed to do so in job_conf.xml, see https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/sample/job_conf.xml.sample_advanced#L613
The job working directory is the correct default because is the only directory the galaxy user is guaranteed to be able to write to.

bernt-matthias:
`vg_convert` had the "_same_" problem: `[xg]: couldn't create temp directory` so I added the same solution.

nsoranzo (approved):

nsoranzo:
You need to change something in `tools/raven` to have it tested and then deployed to the TSs on merge.

Issue #2169, opened 2018-11-08
fastx suite and tool names


bernt-matthias:
Is it possible to get a suite for the fastx tool collection. 

I think this one is deprecated. https://github.com/galaxyproject/tools-iuc/tree/master/suites/suite_fasta_tools

I guess this should be done with `.shed.yml` (auto_tool_repositories, suite, ...), but I'm not sure how to to this here. Given some documentation links I could do this. 

I would also suggest to change the name of the tools: "Name" -> "fastx toolkit Name". Otherwise users don't know what they are actually using. Of course I could also do this.

PR #3318, opened 2020-11-18
change names of fastx tools

bernt-matthias:
`find . -type f -name "*xml" -exec sed -i 's/\(<tool.*name="\)/\1fastx: /' {} \;`

fixes: https://github.com/galaxyproject/tools-iuc/issues/2169

FOR CONTRIBUTOR:

### DIFF
--- a/tool_collections/fastx_toolkit/fasta_nucleotide_changer/fasta_nucleotide_changer.xml
+++ b/tool_collections/fastx_toolkit/fasta_nucleotide_changer/fasta_nucleotide_changer.xml
@@ -1,4 +1,4 @@
-<tool id="cshl_fasta_nucleotides_changer" version="1.0.2" name="RNA/DNA" >
+<tool id="cshl_fasta_nucleotides_changer" version="1.0.2" name="fastx: RNA/DNA" >
     <description>converter</description>
     <macros>
         <import>macros.xml</import>
--- a/tool_collections/fastx_toolkit/fasta_nucleotide_changer/fasta_nucleotide_changer.xml
+++ b/tool_collections/fastx_toolkit/fasta_nucleotide_changer/fasta_nucleotide_changer.xml
@@ -1,4 +1,4 @@
-<tool id="cshl_fasta_nucleotides_changer" version="1.0.2" name="fastx: RNA/DNA converter" >
+<tool id="cshl_fasta_nucleotides_changer" version="1.0.2" name="FASTX-Toolkit: RNA/DNA converter" >
     <description></description>
     <macros>
         <import>macros.xml</import>
--- a/tool_collections/fastx_toolkit/fasta_nucleotide_changer/fasta_nucleotide_changer.xml
+++ b/tool_collections/fastx_toolkit/fasta_nucleotide_changer/fasta_nucleotide_changer.xml
@@ -1,5 +1,5 @@
-<tool id="cshl_fasta_nucleotides_changer" version="1.0.2" name="fastx: RNA/DNA" >
-    <description>converter</description>
+<tool id="cshl_fasta_nucleotides_changer" version="1.0.2" name="fastx: RNA/DNA converter" >
+    <description></description>
     <macros>
         <import>macros.xml</import>
     </macros>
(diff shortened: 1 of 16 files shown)


tool_collections/fastx_toolkit/fasta_nucleotide_changer/fasta_nucleotide_changer.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="cshl_fasta_nucleotides_changer" version="1.0.2" name="RNA/DNA" >
+<tool id="cshl_fasta_nucleotides_changer" version="1.0.2" name="fastx: RNA/DNA" >
     <description>converter</description>
```
nsoranzo, line 2: I'd move `converter` from the description to the tool name.

tool_collections/fastx_toolkit/fastq_quality_boxplot/fastq_quality_boxplot.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="cshl_fastq_quality_boxplot" name="Draw quality score boxplot" version="1.0.1">
+<tool id="cshl_fastq_quality_boxplot" name="fastx: Draw quality score boxplot" version="1.0.1">
```
nsoranzo, line 1:
```suggestion
<tool id="cshl_fastq_quality_boxplot" name="fastx: Quality score boxplot" version="1.0.1">
```

tool_collections/fastx_toolkit/fasta_nucleotide_changer/fasta_nucleotide_changer.xml
nsoranzo, line 2:
```suggestion
<tool id="cshl_fasta_nucleotides_changer" version="1.0.2" name="fastx: RNA/DNA converter" >
    <description></description>
```

tool_collections/fastx_toolkit/fastq_to_fasta/fastq_to_fasta.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="cshl_fastq_to_fasta" name="FASTQ to FASTA" version="1.0.2">
+<tool id="cshl_fastq_to_fasta" name="fastx: FASTQ to FASTA" version="1.0.2">
     <description>converter from FASTX-toolkit</description>
```
nsoranzo, line 2:
```suggestion
<tool id="cshl_fastq_to_fasta" name="fastx: FASTQ to FASTA converter" version="1.0.2">
    <description></description>
```

nsoranzo:
The original name of the upstream suite of tools is FASTX-Toolkit, maybe we should use that (or just `FASTX` uppercase) instead of `fastx`?

nsoranzo (approved):

bgruening:
Why is this really needed? Should a user really care about the underlying tool? And if so we have citations, bio.tools, help sections etc to make clear its a FASTX tools.

bernt-matthias:
> Why is this really needed? 

No. I just found it convenient to make it immediately recognizable for those quite general tools. For `fastq_to_fasta` we had Fastx-Toolkit already in the description.

But we can leave it as is if you prefer, users can just search for fastx and will see the tools of the suite. 

> Should a user really care about the underlying tool? 

Yes - in many cases. If you think of a task like adapter trimming I want my users to use a specific underlying tool (cutadapt) and not any other adapter trimmer (most of which are completely undocumented). 

> And if so we have citations, bio.tools, help sections etc to make clear its a FASTX tools.

True.

bgruening:
I'm not blocking this PR, but if its about hacking around the tool-search I would rather prefer to discuss ad fix tool accessibility and not adding more unstructured data to tools. E.g. we should consider adding `<tags>` to the XML, work further on EDAM and add here the bio.tools ID for fastx.

bernt-matthias:
Maybe we can do both. I still like the idea that users can see what tool they are using in the tool panel. In particular for cases where there are many similar tools.

So something like: 

```
<edam_topics>
    <edam_topic>topic_0157</edam_topic>
</edam_topics>
<edam_operations>
    <edam_operation>operation_0231</edam_operation>
    <edam_operation>operation_3219</edam_operation>
</edam_operations>
```

This should be operations: "sequence editing" and "Read pre-processing"; and topic "Sequence composition, complexity and repeats"

Possibibly and/or 

```
   <xrefs>
	<xref type="bio.tools">fastx-toolkit</xref>
   </xrefs>
```

PR #3319, opened 2020-11-19
Update busco to 4.1.4

cat-bro:
Busco uses biopython.BioAlphabet which has been removed in biopython 1.78, but installing busco currently installs the new biopython version in the environment, and busco jobs fail.  This is a patch for 4.1.2 only: busco versions >= 4.1.4 do not use BioAlphabet.

Is it OK to add a dependency without a version bump?  The tool is not changing at all.

### DIFF
--- a/tools/busco/test-data/genome_results/full_table
+++ b/tools/busco/test-data/genome_results/full_table
@@ -1,5 +1,5 @@
-# BUSCO version is: 4.1.2 
-# The lineage dataset is: arthropoda_odb10 (Creation date: 2019-11-20, number of species: 90, number of BUSCOs: 1013)
+# BUSCO version is: 4.1.4 
+# The lineage dataset is: arthropoda_odb10 (Creation date: 2020-09-10, number of species: 90, number of BUSCOs: 1013)
 # Busco id	Status	Sequence	Gene Start	Gene End	Score	Length	OrthoDB url	Description
 774at6656	Missing
 980at6656	Missing
--- a/tools/busco/busco.xml
+++ b/tools/busco/busco.xml
@@ -82,7 +82,11 @@ ${adv.long}
             <param name="mode" value="geno"/>
             <output name="busco_sum" file="genome_results/short_summary" compare="diff" lines_diff="4"/>
             <output name="busco_table" file="genome_results/full_table" compare="diff" lines_diff="4"/>
-            <output name="busco_missing" file="genome_results/missing_buscos_list" compare="diff" lines_diff="4"/>
+            <output name="busco_missing" file="genome_results/missing_buscos_list" compare="diff" lines_diff="4">
+                <assert_contents>
+                    <has_text text="# BUSCO version is: @TOOL_VERSION@" />
+                </assert_contents>
+            </output>
         </test>
         <test>
             <param name="input" value="proteome.fa"/>
--- a/tools/busco/macros.xml
+++ b/tools/busco/macros.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <macros>
-    <token name="@TOOL_VERSION@">4.1.2</token>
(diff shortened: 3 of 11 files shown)


bernt-matthias:
Wouldn't it be better to just bump to 4.1.4 (which seems to contain only minor improvements.. The changelog does not mention Bio.Alphabet https://gitlab.com/ezlab/busco) . And maybe drop biopython  from the conda recipe (or is it just the one module that was dropped).

bernt-matthias:
Test failed with  `No space left on device` .. I restarted them. Lets see if this is deterministic.

abretaud:
Bumping to 4.1.4 will not be enough, biopython.BioAlphabet was removed from biopython 1.78, but busco still need it and other parts of biopython.
I can make a PR on the bioconda recipe to pin the biopython requirement to <=1.77. And an issue on busco to report the problem and hopefully get a fix in future versions

abretaud:
https://github.com/bioconda/bioconda-recipes/pull/25438

abretaud:
Oh wait... just found https://gitlab.com/ezlab/busco/-/issues/340
Maybe 4.1.4 is enough after all!?

bernt-matthias:
Lets try.

cat-bro:
> Wouldn't it be better to just bump to 4.1.4 (which seems to contain only minor improvements.. The changelog does not mention Bio.Alphabet https://gitlab.com/ezlab/busco) . And maybe drop biopython from the conda recipe (or is it just the one module that was dropped).

Could we have both?  Fix version 4.1.2 because the one in the toolshed will no longer work and also make a 4.1.4 wrapper?  I like the idea of older toolshed versions still being usable.

bernt-matthias:
> Could we have both? Fix version 4.1.2

This should then be done in biopython as well, by creating a subdir for the 4.1.2 version

> and also make a 4.1.4 wrapper?

Lets bump it and see if it works.

tools/busco/test-data/genome_results/full_table
```diff
@@ -1,5 +1,5 @@
-# BUSCO version is: 4.1.2 
-# The lineage dataset is: arthropoda_odb10 (Creation date: 2019-11-20, number of species: 90, number of BUSCOs: 1013)
+# BUSCO version is: 4.1.4 
+# The lineage dataset is: arthropoda_odb10 (Creation date: 2020-09-10, number of species: 90, number of BUSCOs: 1013)
```
bernt-matthias, line 2: It might be slightly better to allow a diff of 4/6 lines instead of updating test data.

cat-bro: The line that reports the busco version is important.  If that line is different it might be an indicator that _uv_ environment has been installed.  I can update the files so that only that line is updated.

bernt-matthias: I see, how about adding:

```
<assert_contents>
    <has_text text="# BUSCO version is: @TOOL_VERSION@" />
</assert_contents>
```

to the test output?

bgruening: ping @cat-bro

cat-bro: If this passes the tests it is good to go I think?  Asserting the version is a good idea (thanks @bernt-matthias) since 4 lines of diff are allowed for each file and could probably hide a discrepancy.  I'm in favour of updating the test output files, since otherwise they would still say 4.1.2 which would be confusing if there were a significant difference in output.

cat-bro: How would I go about updating a recipe for an earlier version of tool in bioconda such as busco?  They don't seem to be versioned in the github repository.

bgruening: You can add the old version back by creating a subfolder under the main recipes/busco/ folder.

abretaud (approved):

PR #3331, opened 2020-11-21
goseq: update to 1.44.0

mblue9:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/goseq/goseq.r
+++ b/tools/goseq/goseq.r
@@ -94,15 +94,15 @@ results <- list()
 
 run_goseq <- function(pwf, genome, gene_id, goseq_method, use_genes_without_cat, repcnt, gene2cat, p_adj_method, out_fp) {
   out <- goseq(pwf, genome = genome, id = gene_id, method = goseq_method, use_genes_without_cat = use_genes_without_cat, gene2cat = go_map)
-  out$p.adjust.over_represented <- p.adjust(out$over_represented_pvalue, method = p_adj_method)
-  out$p.adjust.under_represented <- p.adjust(out$under_represented_pvalue, method = p_adj_method)
+  out$p_adjust_over_represented <- p.adjust(out$over_represented_pvalue, method = p_adj_method)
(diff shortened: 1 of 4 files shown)


mblue9:
Have fixed the linting issues here except 2, 
`no visible binding for global variable ‘go_map’`
`no visible global function definition for ‘goseq’`

Other R tools here have similiar issue e.g deseq2

I wonder if the lintr here is too strict as this issue was reported here
https://github.com/jimhester/lintr/issues/27

Seems because (as says here https://github.com/jimhester/lintr/issues/382)
>lintr is currently mostly geared towards package development rather than scripts unfortunately

Workaround in first issue
>You can disable the object_usage_linter to avoid it.

e.g. 
`linters=subset(lintr::default_linters,!lintr::default_linters %in% lintr::default_linters["object_usage_linter"])`

Is that what needs to be done or something else?

bgruening:
@mblue9 can you use `#nolint`? https://rdrr.io/cran/lintr/man/exclude.html

If you think this is more problematic than useful we need to discuss the linting step of R code again.

mblue9:
I'll have a look, will be good to use the linting if possible.

By the way these R jobs are also failing here due to timeouts during the planemo testing, taking > 6hrs
`The job running on runner GitHub Actions 9 has exceeded the maximum execution time of 360 minutes.`
They don't take very long to test locally so is something wrong here?

bgruening:
@mblue9 please try to add r-base as dependency. r-base=4.0 should hopefully work.

bgruening:
@mblue9 fixed!

The failing tests now are real:

grafik

We also included linting of R code, do you have time to fix that?

We also included xref and EDAM tags. It would be nice if those could be added too:

https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-xrefs-xref
https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-edam-topics

You can find that information here: https://bio.tools/goseq

Thanks!

bgruening (approved):

bgruening:
Gracias! :)

PR #3333, opened 2020-11-21
Add SNP Frequency Plot

mtekman:
FOR CONTRIBUTOR:

TODO:
* [x] Cheetah
* [x] Tests
* [x] PNG and SVG outputs
* [x] Shed YML
* [x] Add VCF parsing script
* [x] VCF parsing incorporated into code
* [x] VCF test input cases
* [x] Permit inidividual datasets
* [x] Clustering parameter should be less or equal to the number of samples
* [x] Label issue for variants


tools/snpfreqplot/heatmap_for_variants.R
```diff
@@ -0,0 +1,129 @@
...
+    color=c("white", "green", "orange", "red", "black", "grey", "blue"),
+    name = c("non-coding", "syn", "non-syn", "deletion", "frame shift", "stop gained", "insertion"))
+subset_colors <- subset(all_colors, color %in% colors)
+effect_color <- subset_colors$color
+names(effect_color) = subset_colors$name
+color_list <- list(gene_color = gene_color, effect_color = effect_color)
+names(color_list) <- c("gene", "effect")
+
+                                        # visualize heatmap
+pdf(out.filepdf,
```
bgruening, line 113: @mtekman can we extend this to also output svg, png etc?

mtekman: Yep, but we'd have to make the `width` and `height` options in the line below more generic. PNG takes pixel values for dimensions, whereas SVG use integers.

If it's just PDF + SVG + PNG that we're adding, then this is no problem really

bgruening: SVG I think would be super nice, PNG a bonus :)

mtekman:
@bgruening @wm75  I think it's more or less complete, except for the VCF parsing.

For that though we need some example VCF test data with the right annotations in the INFO field, I cannot find any in my histories or the few published histories I've checked

wm75:
@mtekman That's a good point! I shared a history with possible test data with you.
Really looking forward to having this tool available :)

tools/snpfreqplot/snpfreqplot.xml
```diff
@@ -0,0 +1,263 @@
...
+        <data name="outpng" format="png" label="SNP Frequency Plot on ${on_string}: PNG" >
+            <filter>advanced["output_type"] == "PNG"</filter>
+        </data>
+        <data name="outsvg" format="svg" label="SNP Frequency Plot on ${on_string}: SVG" >
+            <filter>advanced["output_type"] == "SVG"</filter>
+        </data>
+    </outputs>
+    <tests>
+        <test expect_num_outputs="1">
+            <!-- defaults -->
```
bgruening, line 174: can you please use here the <repeat> syntax. https://github.com/galaxyproject/tools-iuc/blob/f2a33fe115fef9d711112b53136cf7619f1b19be/tools/dada2/dada2_seqCounts.xml#L91

tools/snpfreqplot/snpfreqplot.xml
```diff
@@ -0,0 +1,263 @@
...
+out.file <- '$outsvg'
+#else
+#raise RuntimeError("No such output type")
+#end if
+
+]]>
+        </configfile>
+    </configfiles>
+    <inputs>
+        <repeat name="sinputs" min="2" default="2" title="Sample Inputs">
```
bgruening, line 65: We could offer here a collection (multiple=True) and take the sample-id out of the `element_identifier` and remove the repeat? Or offer the repeat as an optional alternative to the collection?

wm75: Collection would be great! I don't think a repeat is necessary since the user can always select multiple files.

tools/snpfreqplot/snpfreqplot.xml
```diff
@@ -0,0 +1,263 @@
...
+                <option value="SVG" />
+            </param>
+            <param name="multiplier_width" type="float" value="0.5"
+                   min="0.1" max="1" label="Width Multiplier" />
+            <param name="multiplier_height" type="float" value="0.6"
+                   min="0.1" max="1" label="Height Multiplier" />
+            <param name="multiplier_png" type="integer" min="1" value="100"
+                   label="PNG multiplier"
+                   help="General size multiplier, applies only to PNG images." />
+            <param name="color" type="select" label="Color Gene Annotation" >
```
wm75, line 101: Isn't there special tool syntax for picking colors?

wm75: `type="color"`

wm75: Ah, I see. This defines a palette of colors. Nevermind my comment then.

wm75:
Should this tool have Covid-19 in its name, or do you think it will be generically useful?

bgruening:
> Should this tool have Covid-19 in its name, or do you think it will be generically useful?

I had hoped so. But you are the expert :)

wm75:
I can see it becoming useful for *viral* variant calling, in general. (For any larger genomes the plots wouldn't be readable.)
However, we would have to come up with logic for handling overlapping genes, which are quite common in viral genomes.
For a SARS-CoV-2 specific tool, we only need to deal with the overlapping transcripts in ORF1ab, which can easily be hard-coded.
So maybe we start out with a SARS-CoV-2 specific tool (with a name or description indicating the fact) and try to make it more generic in following versions?

wm75:
@mtekman alright, can you add `for SARS-CoV-2` to the wrapper description then.
Changing the description later on is easier than changing the tool name.

mtekman:
thanks for the reviews!

I will try to get some EFF-annotated VCF input datasets working with this too

bgruening:
@mtekman can you please add a shed.yml file as well?

wm75:
My observations from first local planemo serve:

Plotting:
- Images (all output formats) are cut off to various extents depending on width/height settings
- I don't really understand the image multipliers and their effect on the image (needs help and maybe better labels)
- The PNG multiplier should only be offered for PNG format; you can use a macro with all image settings and a `yield` inside that only gets used in the PNG case.
- Maybe less is more here: The PNG multiplier, for example, is really just adding white background around the actual plot so it would be best from a user perspective to come up with a setting that doesn't cut off any content and just use that (assuming this can be coded in relative to the plot size)

Annotations:
- The tool doesn't handle the tricky annotations within ORF1ab correctly yet. The gene row is empty for ORF1ab variants except the most 5' ones, and for cleavage product duplicates of precursor and cleaved peptides are plotted.

Clustering:
- Clustering is doomed to fail if the number of cluster settings is larger than the number of samples (tool should detect that condition and cap the cluster number accordingly); in addition, since the hierarchy of samples is shown even with just a single cluster, I'd say 1 is the best default for the number of clusters

tools/snpfreqplot/snpfreqplot.xml
```diff
@@ -0,0 +1,269 @@
+<tool id="snpfreqplot" name="SNP Frequency Plot" version="@VERSION@+@GALAXY_VERSION@" profile="20.09">
```
wm75, line 1: `version="@VERSION@+galaxy@GALAXY_VERSION@"`

tools/snpfreqplot/snpfreqplot.xml
```diff
@@ -0,0 +1,269 @@
...
+    <description>Plots a heatmap of allele frequencies grouped by variant type for SARS-CoV-2 data</description>
+    <macros>
+        <token name="@VERSION@">1.0</token>
+        <token name="@GALAXY_VERSION@">1</token>
```
wm75, line 5: start at zero

tools/snpfreqplot/snpfreqplot.xml
```diff
@@ -0,0 +1,269 @@
...
+
+## 4. Generate Heatmap (external script)
+##    ----------------------------------
+source('$__tool_directory__/heatmap_for_variants.R')
+
+]]>
+        </configfile>
+    </configfiles>
+    <inputs>
+        <param name="sinputs" format="tabular,vcf" type="data_collection"
```
wm75, line 80: Not sure about the radical approach of allowing *only* collections here.
Obviously it simplifies the naming of the samples in the plot, but makes usage less flexible.
Just multiple datasets would be naturally valid input, too, and the dataset names could be used as the sample names.

tools/snpfreqplot/snpfreqplot.xml
```diff
@@ -0,0 +1,269 @@
...
+               collection_type="list" label="Sample Inputs"
+               help="The dataset names within the collection will be used as sample identifiers." />
+        <param name="varfreq" type="float" min="0" max="1"
+               value="0.1" label="Variant Frequency" />
```
wm75, line 84: label="Variant frequency threshold"

tools/snpfreqplot/snpfreqplot.xml
```diff
@@ -0,0 +1,269 @@
...
+        <token name="@GALAXY_VERSION@">1</token>
+    </macros>
+    <requirements>
+        <requirement type="package" version="1.0.12">r-pheatmap</requirement>
+        <requirement type="package" version="1.1_2">r-rcolorbrewer</requirement>
+        <requirement type="package" version="1.3.0">r-tidyverse</requirement>
+        <requirement type="package" version="1.36.0">bioconductor-variantannotation</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+Rscript '$configscript' >&2
```
wm75, line 14: I assume you'd have to touch or otherwise mention the R files somewhere in the command section to have them available in pulsar jobs (we had this issue with table compute).

tools/snpfreqplot/snpfreqplot.xml
bgruening: > dataset names could be used as the sample names.

`element_identifier` falls back to "dataset name" when not in a collection. So changing this to multiple="true" should hopefully just work.

tools/snpfreqplot/snpfreqplot.xml
wm75: Seems also like a good use case for `license="https://spdx.org/licenses/GPL-3.0-or-later.html"` or similar?

tools/snpfreqplot/snpfreqplot.xml
```diff
@@ -0,0 +1,269 @@
+<tool id="snpfreqplot" name="SNP Frequency Plot" version="@VERSION@+@GALAXY_VERSION@" profile="20.09">
+    <description>Plots a heatmap of allele frequencies grouped by variant type for SARS-CoV-2 data</description>
```
wm75, line 2: We could also shamelessly copy some of the edam terms from #3269:

```
<edam_topics>
    <edam_topic>topic_0797</edam_topic>
    <edam_topic>topic_0092</edam_topic>
</edam_topics>
<edam_operations>
    <edam_operation>operation_3436</edam_operation>
</edam_operations>
```
seems to fit

tools/snpfreqplot/snpfreqplot.xml
mtekman: So should I add the `repeat` functionality again? I guess the Cheetah won't need to change

mtekman:
> Plotting:
> 
>     * Images (all output formats) are cut off to various extents depending on width/height settings
> 

Yeah, so I found this strange too because normally the plot scales to fit the dimensions you give it -- but I recall using pheatmap before and having issues with it related to the above.

It could just be that we replace all image dimension parameters with a DPI scale. That might work.

> Annotations:
> 
>     * The tool doesn't handle the tricky annotations within ORF1ab correctly yet. The gene row is empty for ORF1ab variants except the most 5' ones, and for cleavage product duplicates of precursor and cleaved peptides are plotted.
> 

Hmm. Is this using the 3 VCF test data? Or using your own?

> Clustering:
> 
>     * Clustering is doomed to fail if the number of cluster settings is larger than the number of samples (tool should detect that condition and cap the cluster number accordingly); in addition, since the hierarchy of samples is shown even with just a single cluster, I'd say 1 is the best default for the number of clusters

Yeah, I saw the same. I'm not too versed with how to autofill this. I'll dig around.

tools/snpfreqplot/snpfreqplot.xml
mtekman: I am unsure about where to put the license tag -- I've added it to `tool`, but I have not seen this done anywhere else in the repo

tools/snpfreqplot/snpfreqplot.xml
wm75: If you switch the `type` back to `data`, but add `multiple="true"` that should be all that's needed to work with both multiple datasets and collections.
Repeats only have the advantage of letting you specify the order of the datasets independent of their order in the history, but I don't think that's worth it for this tool.

tools/snpfreqplot/snpfreqplot.xml
wm75: In the tool element should be the right place (at least that's how it's documented). It's quite new functionality and meant to specify the license of the *wrapper* and other stuff you upload to the TS as opposed to a command line tool. So not surprising you don't find other uses of it yet, but I think it fits very nicely here.

wm75:
For the number of clusters, I don't think you can take care of this at the level of the tool interface (cause that would require dynamic knowledge of the number of input datasets).
What I was thinking of is that the script should use min(user_specified_number_of_clusters, number_of_samples) to avoid an outright error.

wm75:
> Yeah, so I found this strange too because normally the plot scales to fit the dimensions you give it -- but I recall using pheatmap before and having issues with it related to the above.
>
> It could just be that we replace all image dimension parameters with a DPI scale. That might work.

Yeah, I'm not familiar with the way you best specify those heatmap dimensions either. Think I'm going to play around with various options and see.

wm75:
For the annotations issue: I've seen it with the three VCFs I shared with you. Haven't actually checked whether you're using them in the tests unmodified now.

tools/snpfreqplot/snpfreqplot.xml
wm75: Just gathering experience with that feature. Apparently, it should be `license="GPL-3.0-or-later"` (i.e., just the license identifier, not the link to it). Galaxy will then create a link to the license document for you just as it does for DOIs in citations.

wm75: You can see what it gets formatted like if you planemo serve against a Galaxy dev version. Should become generally available with release_21.01 then.

mtekman:
> For the annotations issue: I've seen it with the three VCFs I shared with you. Haven't actually checked whether you're using them in the tests unmodified now.

Is this more accurate?

Galaxy21- SNP_Frequency_Plot_on_data_3,_data_2,_and_data_1__PNG

tools/snpfreqplot/snpfreqplot.xml
```diff
@@ -175,6 +187,15 @@ source('$__tool_directory__/heatmap_for_variants.R')
         <data name="outsvg" format="svg" label="SNP Frequency Plot on ${on_string}: SVG" >
             <filter>advanced["output_type"] == "SVG"</filter>
         </data>
+        <data name="outtiff" format="tiff" label="SNP Frequency Plot on ${on_string}: TIFF" >
```
bgruening, line 190: @mtekman see here: https://github.com/galaxyproject/tools-iuc/blob/7565dcc0d2019e53121fce73313e7fc16bfa80c8/tools/last/maf-convert.xml#L87

gregvonkuster:
Will this tool be complementary to this one?
 https://github.com/galaxyproject/tools-iuc/blob/master/tools/vsnp/vsnp_build_tables.xml

Here's an example table that looks similar.  We should clearly differentiate the tools.
https://github.com/galaxyproject/tools-iuc/blob/master/tools/vsnp/static/images/table_description.png

nekrut:
Thanks for adding this!

wm75:
> Will this tool be complementary to this one?
> https://github.com/galaxyproject/tools-iuc/blob/master/tools/vsnp/vsnp_build_tables.xml
> 
> Here's an example table that looks similar. We should clearly differentiate the tools.
> https://github.com/galaxyproject/tools-iuc/blob/master/tools/vsnp/static/images/table_description.png

@gregvonkuster That image does suggest substantial overlap between the outputs of the tools. So this is what's inside the excel spreadsheets produced by your tool? Hadn't looked into them beofre.
The scope of this tool here is kind of simpler though. It takes VCFs of SnpEff-annotated variants and generates a comparison plot of the samples, *i.e.*, it's a reporting tool for some of the Galaxy WFs for covid-19 variant analysis, whcih we hope to be able to generalize for other viral variants sets in the future.

wm75:
> > For the annotations issue: I've seen it with the three VCFs I shared with you. Haven't actually checked whether you're using them in the tests unmodified now.
> 
> Is this more accurate?
> 
> Galaxy21- SNP_Frequency_Plot_on_data_3,_data_2,_and_data_1__PNG

@mtekman Your plot illustrates the duplicate variants issue very nicely still!
Only the first two variants are annotated as ORF1ab, while all the white ones should be ORF1ab, too.
These white ones are all coming in duplicates of variants annotated as affecting the precursor protein (those with the higher AA numbers) *and* as affecting the cleavage product (lower AA numbers).
IMO, what the tool *should* do is, if it sees a variant effect on gene ORF1ab, it should keep it only if the affected transcript ID is `GU280_gp01` (the ID of the precursor in the relevant genbank files and in the covid-19 edition of SnpEff). It feels bad to hard-code this into the tool, but the quickest solution now. We can think of letting users specify a file with transcript IDs they want to have plotted or something similar once the first version is out.

wm75:
@mtekman I generated an SVG with the tool without problems yesterday. What is not supported about it?

mtekman:
> @mtekman I generated an SVG with the tool without problems yesterday. What is not supported about it?

To fix the boundary/cropping issue, `pheatmap` is no longer manually supplied a plotting device (of any format), but is asked to write the plot straight to file using its own native method. This native method fixes the cropping, but lacks SVG

> Only the first two variants are annotated as ORF1ab, while all the white ones should be ORF1ab, too.

Ah of course. I see the issue, I'll have a look at what is going on.

mtekman:
@wm75  I think both issues (one failing VCF file in a collection, and the gene highlighting issue) are now fixed:

new

mtekman:
One issue I could not resolve is that I cannot seem to be able to feed the tool a collection list from the `<tests>` section, despite this working perfectly fine in the interface. I have left my attempts commented in the XML

mtekman:
Screenshot_2020-11-24 Tool Test Results (powered by Planemo)

Works fine locally

bgruening:
> One issue I could not resolve is that I cannot seem to be able to feed the tool a collection list from the <tests> section, despite this working perfectly fine in the interface. I have left my attempts commented in the XML

What is your issue here. The comma separated list is totally fine.

bgruening:
You touch a file in your tool_directory, but your tool_directory is read-only in production systems. Cat them to /dev/null or simply use one R script :)

mtekman:
> > One issue I could not resolve is that I cannot seem to be able to feed the tool a collection list from the  section, despite this working perfectly fine in the interface. I have left my attempts commented in the XML
> 
> What is your issue here. The comma separated list is totally fine.

Well the elements take the filenames from the test-data folder, instead of the ones I give them. I guess it's not a huge issue, but I was wondering if I was doing something wrong

bgruening:
@mtekman you don't need to compare PDFs anymore you can use `has_size` so just check the filesize.

> Expected 13 pieces. Additional pieces discarded in 34 rows

Is this warning expected?

tools/snpfreqplot/heatmap_for_variants.R
```diff
@@ -0,0 +1,142 @@
...
+                 pheat_number_of_clusters))
+}
+
+## image trickery needed here:
+## - the heatmap file type is determined by the extension, but
+##   galaxy only serves .dat files, so we need to create a
+##   a new output file with that extension, and move to the
+##   the galaxy output dat file.
+
+out_tmpfile <- tempfile(fileext = paste0(".", tolower(out_type)))
```
bgruening, line 127: you can use here a file in the current woring dir `./foo.png` etc ... and let Galaxy move the file with `from_work_dir`

bgruening (commented):
I don't know what those crashes are, I suspect some cairo problems in the containers. But the `memory-map` error is strange.

If its to hard to convince R to plot reliable to PNG and JPEG, I think its fine if we just support PDF and SVG?

tools/snpfreqplot/snpfreqplot.xml
```diff
@@ -178,25 +164,14 @@ source('$__tool_directory__/heatmap_for_variants.R')
...
-            <filter>advanced["output_type"] == "TIFF"</filter>
-        </data>
-        <data name="outbmp" format="bmp" label="SNP Frequency Plot on ${on_string}: BMP" >
-            <filter>advanced["output_type"] == "BMP"</filter>
-        </data>
-        <data name="outjpeg" format="jpeg" label="SNP Frequency Plot on ${on_string}: JPEG" >
-            <filter>advanced["output_type"] == "JPEG"</filter>
+        <data name="outfile" label="SNP Frequency Plot on ${on_string}: ${advanced.output_type}" >
+            <change_format>
+                <when input="advanced.output_type" value="PDF" format="pdf" />
```
wm75, line 169: You need to decide for a default output type here (pdf would make sense cause it's the default in the UI).
<change_format> changes the format to something *else* when certain conditions apply.
So, add `format="pdf"` to the `data` element and remove the first `when` here.

tools/snpfreqplot/heatmap_for_variants.R
mtekman: Ah, true -- I thought it would be cleaner to do from R, but I guess it's causing issue

tools/snpfreqplot/snpfreqplot.xml
mtekman: are you okay with the lack of native SVG?

we could implement it ourselves, whilst harnessing the cellheight and cellwidth parameters

tools/snpfreqplot/heatmap_for_variants.R
mtekman: is it possible to do `from_work_dir="${file_with_dynamic_ext}"` ?

tools/snpfreqplot/snpfreqplot.xml
```diff
@@ -0,0 +1,259 @@
...
+    </requirements>
+    <edam_topics>
+        <edam_topic>topic_0797</edam_topic>
+        <edam_topic>topic_0092</edam_topic>
+    </edam_topics>
+    <edam_operations>
+        <edam_operation>operation_3436</edam_operation>
+    </edam_operations>
+    <command detect_errors="exit_code"><![CDATA[
+#set $outfile_tmp = str($outfile) + "." + str($advanced.output_type).lower()
```
wm75, line 22: You can make the logic of the file renaming clearer by reusing this line here at the top of the configfile instead of solving the same task in R there.

wm75: ... and you could simplify things further by:
1. using a hard-coded base name for the tmp file
2. defining lowercase values for $advanced.output_type (I don't think these values need to be in uppercase for any particular reason?)

With that you could write the above as:
`#set $outfile_tmp = 'tmp_output.' + str($advanced.output_type)`

wm75: In addition, it seems that you can use wildcards in the `from_work_dir` value so you could also make this the final output with e.g. `#set $outfile = 'plot_output.' + str($advanced.output_type)` and have that discovered in the outputs section as `from_work_dir='plot_output.*` (I think).

tools/snpfreqplot/snpfreqplot.xml
```diff
@@ -0,0 +1,259 @@
...
+
+cat '$__tool_directory__/helperFunctions.R' > /dev/null
+&& cat '$__tool_directory__/snpEffExtract.R' > /dev/null
+&& cat '$__tool_directory__/heatmap_for_variants.R' > /dev/null
+&& echo "output file: $outfile_tmp"
+&& Rscript '$configscript'
+&& mv '$outfile_tmp' '$outfile'
```
wm75, line 29: This line could be removed if from_work_dir wildcards work as I think they do.

tools/snpfreqplot/snpfreqplot.xml
```diff
@@ -0,0 +1,259 @@
...
+pheat_clustering_method <- "ward.D2"
+pheat_number_of_clusters <- 5
+#end if
+
+cell_width = as.integer('$advanced.cellwidth')
+cell_height = as.integer('$advanced.cellheight')
+if (cell_width == 0 ){ cell_width = NA}
+if (cell_height == 0 ){ cell_height = NA}
+
+out_filetmp = paste0('$outfile', ".", tolower('$advanced.output_type'))
```
wm75, line 79: Replace with cheetah

tools/snpfreqplot/snpfreqplot.xml
```diff
@@ -0,0 +1,259 @@
...
+            <param name="output_type" type="select" label="Output Format" >
+                <option value="PDF" selected="true" />
+                <option value="PNG" />
+                <!-- <option value="SVG" /> -->
+                <option value="TIFF" />
+                <option value="BMP" />
+                <option value="JPEG" />
+            </param>
+            <param name="cellwidth" label="Cell Width" type="integer"
+                   min="0" value="0"
```
wm75, line 104: Can you make "20" the default here?

tools/snpfreqplot/snpfreqplot.xml
```diff
@@ -0,0 +1,259 @@
...
+                   help="Individual Cell Width in points. If set to 0, then the values depend on the size of plotting window." />
+            <param name="cellheight" label="Cell Height" type="integer"
+                   min="0" value="0"
```
wm75, line 107: Also here: value="20"

tools/snpfreqplot/snpfreqplot.xml
```diff
@@ -0,0 +1,259 @@
...
+            </param>
+        </section>
+        <conditional name="clustering">
+            <param name="do" type="select" label="Perform Clustering?" >
+                <option value="TRUE">Yes</option>
+                <option value="FALSE" selected="true">No</option>
+            </param>
+            <when value="TRUE" >
+                <param name="nclust" type="integer"
+                       min="1" value="5" label="Number of Cluster" />
```
wm75, line 154: I still think "1" makes a better default than "5" here.

tools/snpfreqplot/snpfreqplot.xml
mtekman: fantastic stuff, thanks

bgruening (commented):
* help is missing
* the other output formats are offered but not tested

wm75:
@bgruening Yes, been postponing help section. Going to add it now.

mtekman:
> the other output formats are offered but not tested

The tool technically supports these formats, but the Cairo library on the  remote build segfaults if the output is anything but PDF.

I'm not sure how to fix this, but maybe we can serve the tool as it is (in the hopes that this behaviour is unique only to the remote build here), with PDF as a default stable output, and the other output types as available but not guaranteed to not error out...?

mtekman (commented):
Good shout with the minification. Did you mean to add the column indexes here though?

wm75:
Oops, you're right! Let me fix that ...

wm75:
@mtekman Maybe you can restore a test for SVG output format. Who knows, maybe even the other formats would start working again if using the R plotting device again? It seems scaling and cut-offs are solved with the new logic.

mtekman:
`lintr` is the bane of my existence, but yes I can try. I guess we can switch back to normal plotting again, though I do have my doubts it will fix the cropping issue entirely

wm75:
@mtekman We're back in business! Tests are failing now because tests are failing :)
Crashes seem to be caused by R v4 in combination with the container, downgrading to R 3.6 and cleaning up the dependency jungle solved it.

mtekman:
The issue caused by the variant frequency being less than 0.05 was due to the sudden detection of a `SPLICE_SITE_REGION+SYNONYMOUS_CODING` effect in snpeff.789.vcf 

At the moment I simply added this new effect to the color list as a hardcoded value to expect, but I'm not sure how robust this is given that `+` can now likely show other combinations of effects.

How do we represent this? I see two options:

 1. We subselect for only the effects that we want to see, namely: non-coding, syn, non-syn, deletion, frame shift, stop gained, insertion, and discard all others.
 2. We show all effects in the heatmap, but then we cannot hardcode color values for specific effects, and that we do not make name substitutions (e.g. `"CODON_CHANGE_PLUS_CODON_DELETION" <- "deletion"`)

wm75:
I happen to have a bit of experience with translating SnpEff effects. Let me fix that properly ...

wm75:
@mtekman I think I've addressed the variant effects issue. Haven't checked if tests are failing because of slightly different output, nor if lintr likes my changes (probably not), but my manual testing indicates that the new functionality works as expected.

mtekman:
looks good, I can happily iron out any kinks that may appear

mtekman:
Nice updates! Is this now ready?

wm75 (approved):
I'm nearly sure we will keep discovering additional issues for a while, but we may speed up this process by merging, then testing on a public server against our giant collections of SARS-CoV-2 variants there.

bgruening:
wow, thanks for all your work on it @mtekman @wm75

bgruening:
uh PR 3333!

PR #3336, opened 2020-11-22
Missing comma in make.contigs.xml

pmpcosta:
FOR CONTRIBUTOR:

Bugfix. Actual excerpt from make.contigs logfile:
mothur > 
<-1,gapopen=-2,gapextend=-1,rename=Trueprocessors=4)

Using 1 processors.

After correcting missing comma in "rename" parameter:
mothur > 
<1,gapopen=-2,gapextend=-1,rename=True,processors=4)

Using 4 processors.

Now the tool is able to use multiple cores. No impact in the result itself so much as I could see.

### DIFF
--- a/tools/mothur/make.contigs.xml
+++ b/tools/mothur/make.contigs.xml
@@ -54,7 +54,7 @@ echo 'make.contigs(
     mismatch=$mismatch,
     gapopen=$gapopen,
     gapextend=$gapextend,
-    rename=$rename
+    rename=$rename,
     processors='\${GALAXY_SLOTS:-8}'
 )'
 | sed 's/ //g'  ## mothur trips over whitespace
--- a/tools/mothur/make.contigs.xml
+++ b/tools/mothur/make.contigs.xml
@@ -169,9 +169,21 @@ echo 'make.contigs(
             <param name="mismatch" value="-2"/>
             <param name="gapopen" value="-3"/>
             <param name="gapextend" value="-2"/>
-            <output name="fasta" md5="48e32c65bd9f064c5c0b4ea7695cabe9" ftype="fasta"/>
-            <output name="qual" md5="1e7778cee0d86bfa2759a07bb4356165" ftype="qual"/>
-            <output name="report" md5="5274725ef45890fd6da4650d5d536173" ftype="txt"/>
+            <output name="fasta" md5="48e32c65bd9f064c5c0b4ea7695cabe9" ftype="fasta">
+                <assert_contents>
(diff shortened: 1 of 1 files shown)


pmpcosta:
Ouch! "rename sequences" now works, which means all tests will have to be redone...

bernt-matthias (changes requested):
Good find. Should be ready to go with updated test data.

tools/mothur/make.contigs.xml
```diff
@@ -171,6 +173,25 @@ echo 'make.contigs(
...
+            <conditional name="input_type">
+                <param name="type" value="simple_collection"/>
+                <param name="paired_collection">
+                    <collection type="paired">
+                        <element name="forward" value="Mock_S280_L001_R1_001_small.fastq" />
+                        <element name="reverse" value="Mock_S280_L001_R2_001_small.fastq" />
+                    </collection>
+                </param>
+            </conditional>
+            <output name="fasta" md5="addcc2f4233df75ab9e5feacadb6850f" ftype="fasta"/>
```
bernt-matthias, line 190: md5 is fine for me .. but I prefer asserts on the contents. I leave this decision to you,

bernt-matthias (approved):

tools/mothur/make.contigs.xml
pmpcosta: There are already a lot of closely related test files. Since this test only checks for change of sequence names, it seems reasonable to me.

bernt-matthias: I meant to add something like:

```
<assert_contents>
    <has_text text="...." />
</assert_contents>
```

to the output tag .. you may have this in addition to the md5 if you like. Then the intention of the test maybe easier to understand compared to the md5. I agree completely that we do not need more test files.

pmpcosta: Ah, ok..

bernt-matthias: Ah. I guess tabs are not possible in the text, maybe:

```
<assert_contents>
    <has_text_matching expression="...." />
</assert_contents>

```

pmpcosta: From the xml docs it seems you can use the xml escape sequence for tabs, lets try..

bernt-matthias:
One more thing: we need to bump the tool version to `@WRAPPER_VERSION@.1`

bgruening (approved):

bgruening:
Cool, thanks a lot!

PR #3340, opened 2020-11-23
CI fix matrix calculation

bernt-matthias:
For the workflows we could require a min of 1 chunk to make the workflow pass? When no tools changed

FOR CONTRIBUTOR:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -56,7 +56,7 @@ jobs:
         nchunks=$(find tools/ tool_collections/ data_managers/ -iname '*xml' | grep -v macro | wc -l)
         if [ "$nchunks" -gt "$MAX_CHUNKS" ]; then
             nchunks=$MAX_CHUNKS
-        elif [ "$nchunks" -eq "0" ]
+        elif [ "$nchunks" -eq 0 ]; then
             nchunks=1
         fi
         echo "::set-output name=nchunks::$nchunks"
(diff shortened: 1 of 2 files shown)


nsoranzo:
> For the workflows we could require a min of 1 chunk to make the workflow pass? When no tools changed

I was also thinking that!

nsoranzo (approved):

bernt-matthias:
> > For the workflows we could require a min of 1 chunk to make the workflow pass? When no tools changed
> 
> I was also thinking that!

Then lets do it :)

.github/workflows/ci.yaml
```diff
@@ -56,9 +56,11 @@ jobs:
         nchunks=$(find tools/ tool_collections/ data_managers/ -iname '*xml' | grep -v macro | wc -l)
         if [ "$nchunks" -gt "$MAX_CHUNKS" ]; then
             nchunks=$MAX_CHUNKS
+        elif [ "$nchunks" -eq "0" ]
```
nsoranzo, line 59:
```suggestion
        elif [ "$nchunks" -eq 0 ]; then
```

.github/workflows/pr.yaml
```diff
@@ -106,6 +106,8 @@ jobs:
         nchunks=$(wc -l < changed_tools.list)
         if [ "$nchunks" -gt "$MAX_CHUNKS" ]; then
             nchunks=$MAX_CHUNKS
+        elif [ "$nchunks" -eq "0" ]
```
nsoranzo, line 109:
```suggestion
        elif [ "$nchunks" -eq 0 ]; then
```

PR #3350, opened 2020-11-25
Adding the tool GECKO

estebanpw:
This PR adds the tool GECKO (more [here](https://github.com/otorreno/gecko)
The tool is already merged in bioconda. I think everything is correct, but I not completely sure in regards to:

- Paths, i.e. the two executables are `workflow.sh` and `frags2align.sh` which are usually in the `bin` folder of the tool repository. However, due to the `${PREFIX}` path in the `build.sh` of the bioconda repo, I believe I can/should run them without any path (or should I include `${__tool_directory__}`?). Can somebody take a look on this?
- The `cp` command. I was having issues when using soft links (I think because of further softlinks that are done afterwards). Is it okay to use `cp` instead of softlinks?

FOR CONTRIBUTOR:


estebanpw:
OK I realised the paths are OK

tools/gecko/gecko.xml
```diff
@@ -0,0 +1,162 @@
...
+    <command><![CDATA[
+    #import re
+
+    #set $queryName=re.sub('[^\w\-_.]', '_', str($query.name))
+    #set $dbName=re.sub('[^\w\-_.]', '_', str($db.name))
+    #set $queryNoFormat=$queryName.replace(".fasta","")
+    #set $dbNoFormat=$dbName.replace(".fasta","")
+
+    cp $query '$queryName' &&
+    cp $db '$dbName' &&
```
bernt-matthias, line 18:
```suggestion
    cp $query query &&
    cp $db db &&
```
and simply use `query` and `db` from here. Rationale. The Filenames seem not to be used anywhere. So we can chose whatever we like.

bernt-matthias (commented):
Looks good. But the tests are failing with something like `Expected 11118 lines in output, found 1 lines`

tools/gecko/gecko.xml
```diff
@@ -0,0 +1,162 @@
...
+            <param name="minlen" value="20"/>
+            <param name="similarity" value="60"/>
+            <param name="selection_mode" value="noalignments"/>
+
+            <output name="frags_output0" file="mycoplasma-232-mycoplasma-7422.frags"/>
+            <output name="csv_output1" file="mycoplasma-232-mycoplasma-7422.csv"/>
+            <assert_command>
+                <!-- <has_text text="All by-Identity Ungapped Fragments (Hits based approach)"/> -->
+                <has_n_lines n="11118"/>
+            </assert_command>
```
bernt-matthias, line 85: I guess you want:

```suggestion
            <output name="csv_output1" file="mycoplasma-232-mycoplasma-7422.csv">
                <assert_content>
                <!-- <has_text text="All by-Identity Ungapped Fragments (Hits based approach)"/> -->
                    <has_n_lines n="11118"/>
                </assert_content>
        </output>
```

tools/gecko/gecko.xml
```diff
@@ -0,0 +1,162 @@
...
+            <option value="32" selected="true">32</option>
+            <option value="28">28</option>
+            <option value="24">24</option>
+            <option value="20">20</option>
+            <option value="16">16</option>
+            <option value="12">12</option>
+            <option value="8">8</option>
+        </param>
+        <param name="minlen" type="integer" value="40" min="8" label="Minimum length" help="The minimum length a fragment must achieve to be reported"/>
+        <param name="similarity" type="integer" value="60" min="1" max="99" label="Similarity value" help="Percentage of similarity (calculated as the attained score divided by the maximum possible score)"/>
```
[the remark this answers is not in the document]
estebanpw: yes, thats more appropriate

tools/gecko/gecko.xml
estebanpw: In the CSV generated in the output there is metadata (rows `seqX filename` and `seqY filename`) which contains the names of the sequences, see below:
```
All by-Identity Ungapped Fragments (Hits based approach)
[Abr.98/Apr.2010/Dec.2011 -- <ortrelles@uma.es>
SeqX filename        : mycoplasma-232.fasta
SeqY filename        : mycoplasma-7422.fasta
SeqX name            : gi|54019969|ref|NC_006360.1|
SeqY name            : gi|525903163|ref|NC_021831.1|
SeqX length          : 892758
SeqY length          : 898495
Min.fragment.length  : 20
Min.Identity         : 60
Tot Hits (seeds)     : 61704
Tot Hits (seeds) used: 38301
Total fragments      : 12415
``` 
If the files are not renamed then it will contain the internal galaxy dataset, right? Thats the reason I wanted to do all the naming, I should have added a comment on it :p
Can we keep this way?

estebanpw:
btw, before merge, can we add some warning or maybe limit the input size (gecko is disk intensive and since it can run concurrently here I think it would be wise to do so). @bernt-matthias

estebanpw:
Planemo is saying that `assert_contents` is not expected:
`.. ERROR: Invalid XML found in file: tmp_y76817j. Errors [/tmp/tmp_y76817j:98:0:ERROR:SCHEMASV:SCHEMAV_ELEMENT_CONTENT: Element 'assert_contents': This element is not expected.]`

But according to [here](https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-tests-test-output-assert-contents) it should be OK. Any idea?

estebanpw:
Tests are passing now but its failing on `Error: read ECONNRESET`
Is this related to the gecko tool?

tools/gecko/gecko.xml
bernt-matthias: Yes. But then you should use `element_identifier` instead of `name`. Maybe also remove the `".fasta"` only if its at the end (by using regex of os.path.basename) .. or just leave. I guess you will see .fasta only in the tests anyway. So I tend to suggest the leave the extension.

bernt-matthias:
> Tests are passing now but its failing on `Error: read ECONNRESET`
> Is this related to the gecko tool?

I restarted the tests and they seem to pass now.

tools/gecko/gecko.xml
estebanpw: I changed to `element_identifier` and removed only the ending `.fasta`. For this I used [os.path.splitext](https://docs.python.org/3/library/os.path.html#os.path.splitext) which splits between root and extension (only the last one).

bernt-matthias: Hmm. Thought again: splitext will just split at the last `.` .. if there is a dot in the identifier it will "fail". 

Somehow I favor to skip this part and use the identifier as is.

tools/gecko/gecko.xml
```diff
@@ -1,35 +1,35 @@
...
     <command><![CDATA[
     #import re
-    #import os
 
-    #set $queryName=re.sub('[^\w\-_.]', '_', str($query.element_identifier))
-    #set $dbName=re.sub('[^\w\-_.]', '_', str($db.element_identifier))
-    #set $queryNoFormat=os.path.splitext($queryName)[0]
-    #set $dbNoFormat=os.path.splitext($dbName)[0]
+    #set $queryName=re.sub('/', '_', str($query.element_identifier))
+    #set $dbName=re.sub('/', '_', str($db.element_identifier))
```
bernt-matthias, line 13:
```suggestion
    #set $queryName=re.sub('/', '\\/', str($query.element_identifier))
    #set $dbName=re.sub('/', '\\/', str($db.element_identifier))
```

tools/gecko/gecko.xml
```diff
@@ -0,0 +1,165 @@
...
+                <!-- <has_text text="All by-Identity Ungapped Fragments (Hits based approach)"/> -->
+                    <has_n_lines n="11118"/>
+                </assert_contents>
+        </output>
+        </test>
+
+        <!-- test run w non defaults (alignment included) -->
+        <test expect_num_outputs="3">
+            <param name="query" value="mycoplasma-232.fasta"/>
+            <param name="db" value="mycoplasma-7422.fasta"/>
```
bernt-matthias, line 93: Can you add `ftype` to all in and outputs? In the tests.

tools/gecko/gecko.xml
```diff
@@ -0,0 +1,165 @@
...
+            <param name="kmer" value="16"/>
+            <param name="minlen" value="20"/>
+            <param name="similarity" value="60"/>
+            <param name="selection_mode" value="alignments"/>
+
+            <output name="frags_output0" file="mycoplasma-232-mycoplasma-7422.frags"/>
+            <output name="csv_output1" file="mycoplasma-232-mycoplasma-7422.csv"/>
+            <output name="alignments2" file="mycoplasma-232-mycoplasma-7422.txt">
```
bernt-matthias, line 101: I think some of the outputs are to big to keep them in the repo. Can you use some assertions instead. And / or reduce input size?

tools/gecko/gecko.xml
bernt-matthias, line 93: Can you add `ftype` to all test in and outputs?

tools/gecko/gecko.xml
```diff
@@ -0,0 +1,165 @@
...
+            <param name="db" value="mycoplasma-7422.fasta"/>
+            <param name="kmer" value="16"/> 
+            <param name="minlen" value="20"/>
+            <param name="similarity" value="60"/>
+            <param name="selection_mode" value="noalignments"/>
+
+            <output name="frags_output0" file="mycoplasma-232-mycoplasma-7422.frags"/>
+            <output name="csv_output1" file="mycoplasma-232-mycoplasma-7422.csv">
+                <assert_contents>
+                <!-- <has_text text="All by-Identity Ungapped Fragments (Hits based approach)"/> -->
```
bernt-matthias, line 84: Comments are fine, but commented code should be removed.

tools/gecko/gecko.xml
```diff
@@ -0,0 +1,165 @@
+<tool id="gecko" name="Gecko" version="@TOOL_VERSION@">
```
bernt-matthias, line 1: Please add `profile`, eg 20.01

bernt-matthias (commented):
Nearly there 👍

Can you also remove some of the empty lines?

tools/gecko/gecko.xml
estebanpw: Done, also removed empty lines

tools/gecko/gecko.xml
estebanpw: Done, reduced to <30 KB

tools/gecko/gecko.xml
estebanpw: Done

tools/gecko/gecko.xml
bernt-matthias: Having `file="hyopneumoniae_16S-flocculare_16S.csv"` and assertions is a bit redundant (but fine). But you could remove the `file` attribute if your assertions are good enough.

bernt-matthias: Would reduce the test data size even more :)

estebanpw: Right now is less than 10 KB (since I also removed the frags)

bernt-matthias (approved):

bgruening (approved):

bgruening:
Thanks @estebanpw!

estebanpw:
Thanks to you guys for all the reviewing and guidance 👍

PR #3363, opened 2020-12-02
Update kraken2 data manager for new MiniKraken2 releases

davebx:
FOR CONTRIBUTOR:

### DIFF
--- a/data_managers/data_manager_build_kraken2_database/data_manager/kraken2_build_database.py
+++ b/data_managers/data_manager_build_kraken2_database/data_manager/kraken2_build_database.py
@@ -49,6 +49,13 @@ class Minikraken2Versions(Enum):
     def __str__(self):
         return self.value
 
+class Minikraken2Releases(Enum):
+    March_2020 = 'March_2020'
+    April_2019 = 'April_2019'
+
+    def __str__(self):
+        return self.value
+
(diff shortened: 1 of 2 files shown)


bernt-matthias (commented):
Hey @davebx I'm new to kraken2 and the data manager. 

- What is the meaning of the releases and versions?
- What is the relation of kraken and minikraken?
- Seems that the data is now stored on aws, see https://ccb.jhu.edu/software/kraken2/ and https://benlangmead.github.io/aws-indexes/. Maybe we should adapt?

Tests for the new cases would certainly be good. 

But maybe I'm not the best for the review. But I will see what I can do.

davebx:
@bernt-matthias, 

1. As far as I can tell, the new release is a database generated in March, alongside the two databases (v1 and v2) generated in April of last year
2. As I understand it, minikraken is one of the smaller pre-generated databases for kraken.
3. I agree, that could be a good feature for the `+galaxy1` version

pvanheus:
@davebx I presume the aim here is to download the pre-built Minikraken DBs? In that regard, this looks good. I don't want to block your PR but these builds (hosted on AWS) are probably also worth adding to the data manager: https://benlangmead.github.io/aws-indexes/k2

PR #3371, opened 2020-12-04
Eliminate use of input collections from vSNP tools

gregvonkuster:
FOR CONTRIBUTOR:


gregvonkuster:
These changes are per instruction from @bernt-matthias

tools/vsnp/vsnp_add_zero_coverage.xml
```diff
@@ -9,70 +9,33 @@
...
-                <param name="bam_input_collection" type="data_collection" format="bam" collection_type="list" label="Collection of BAM files">
-                    <validator type="unspecified_build"/>
-                </param>
-                <param name="vcf_input_collection" type="data_collection" format="vcf" collection_type="list" label="Collection of VCF files">
-                    <validator type="unspecified_build"/>
-                </param>
-            </when>
-        </conditional>
+        <param name="bam_input" type="data" format="bam" label="BAM file">
+            <validator type="unspecified_build"/>
```
bernt-matthias, line 34: Do you use the build somewhere? Would be useful in the reference, or?

But the it may be better not to have the validator. If the user selects a build only matching reference will be shown. If no build is specified then all options are shown.

tools/vsnp/vsnp_add_zero_coverage.xml
```diff
@@ -90,61 +53,17 @@ python '$__tool_directory__/vsnp_add_zero_coverage.py'
...
-            <filter>input_type_cond['input_type'] == 'collection'</filter>
-        </collection>
-        <data name="output_metrics" format="tabular"  label="${tool.name} (metrics) on ${on_string}">
-            <filter>input_type_cond['input_type'] == 'single'</filter>
-        </data>
-        <collection name="output_metrics_collection" type="list" label="${tool.name} (metrics) on ${on_string}">
-            <discover_datasets pattern="__name__" directory="output_metrics_dir" format="tabular" />
-            <filter>input_type_cond['input_type'] == 'collection'</filter>
-        </collection>
+        <data name="output_vcf" format="vcf"  label="${tool.name} (filtered VCF) on ${on_string}"/>
```
bernt-matthias, line 56: Can you change the labels to `${tool.name} on ${on_string}: (filtered VCF) ` .. this is more common.

tools/vsnp/vsnp_add_zero_coverage.xml
```diff
@@ -90,61 +53,17 @@ python '$__tool_directory__/vsnp_add_zero_coverage.py'
...
-            <output name="output_metrics" file="output_metrics.tabular" ftype="tabular" compare="contains"/>
-        </test>
-        <test>
-            <param name="bam_input" value="bam_input.bam" ftype="bam" dbkey="89"/>
-            <param name="vcf_input" value="vcf_input.vcf" ftype="vcf" dbkey="89"/>
-            <param name="reference_source" value="cached"/>
-            <output name="output_vcf" value="output_vcf.vcf" ftype="vcf" compare="contains"/>
-            <output name="output_metrics" file="output_metrics.tabular" ftype="tabular" compare="contains"/>
+            <output name="output_metrics" file="output_metrics.tabular" ftype="tabular"/>
         </test>
```
bernt-matthias, line 67: Can you add a test with the cached reference?

tools/vsnp/vsnp_build_tables.xml
```diff
@@ -109,8 +75,8 @@ python '$__tool_directory__/vsnp_build_tables.py'
         </conditional>
     </inputs>
     <outputs>
-        <collection name="excel" type="list">
-            <discover_datasets pattern="__name__" directory="output_excel_dir" format="xlsx" />
+        <collection name="excel" type="list" label="${tool.name} (tables) on ${on_string}">
+            <discover_datasets pattern="__name__" directory="output_excel_dir" format="xlsx"/>
```
bernt-matthias, line 79: If you use `__name_and_ext__` then the xlsx is gone in the identifiers. Also format would not be needed here.

tools/vsnp/vsnp_build_tables.xml
```diff
@@ -120,43 +86,40 @@ python '$__tool_directory__/vsnp_build_tables.py'
             <param name="input_avg_mq_json" value="input_avg_mq_json.json" ftype="json" dbkey="89"/>
             <param name="gbk_param" value="no"/>
             <output_collection name="excel" type="list">
```
bernt-matthias, line 88: Are those all collection elements? If not then specify the count attribute. Otherwise its fine.

tools/vsnp/vsnp_add_zero_coverage.py
```diff
@@ -1,189 +1,138 @@
...
+def get_sample_name(file_path):
     base_file_name = os.path.basename(file_path)
-    if base_file_name.find(".") > 0:
+    if base_file_name.find("_fastqsanger"):
+        return base_file_name.split("_fastqsanger")[0]
+    elif base_file_name.find("_fastq"):
+        return base_file_name.split("_fastq")[0]
+    elif base_file_name.find("_fq"):
+        return base_file_name.split("_fq")[0]
+    elif base_file_name.find(".") > 0:
```
bernt-matthias, line 21:
```suggestion
```

I think this part is not necessary. In practice the filenames will be SAMPLENAME.extension. Having this `_fq`, ... suffix is just because we can not define the identifiers for the data sets in the tests, instead in tests the identifier of uploaded data will be simply the filename (and the underscore stems from the re.sub). 

It might even be "dangerous". Just think of the case when someone has a sample containing "_fq"

bernt-matthias (commented):
Looks pretty good. 

I added comments to only one tool. But most apply to all.

tools/vsnp/vsnp_add_zero_coverage.xml
gregvonkuster: When included in a workflow, this tool is just downstream from the Picard MarkDuplicates tool (which provides the BAM inputs here), and the Picard tool receives its input from the bwa-mem tool which does the mapping in one of 2 ways: via the output of the vsnp_determine_ref_from_input_data that you just reviewed or by the pre-configured setting in the workflow which the user can override.  So the bam inputs are expected to be mapped at this point in the analysis.  

If executed directly by a user outside of a workflow, they should be able to choose the reference from a select list containing all possibilities.  

The selected reference is used by the associated script here https://github.com/galaxyproject/tools-iuc/pull/3371/files#diff-b54d89b657fd585021413c5008917f3a43a8cf72893de0969c818319368a8954R42

tools/vsnp/vsnp_add_zero_coverage.xml
gregvonkuster: Done here https://github.com/galaxyproject/tools-iuc/pull/3371/commits/d95f5d7b65a6dd248a0db5a498d6f49b49dfc251.

tools/vsnp/vsnp_add_zero_coverage.xml
gregvonkuster: Hmmm...the original version of the tool had this test, so I must have inadvertently deleted it.  Re-added here https://github.com/galaxyproject/tools-iuc/pull/3371/commits/d95f5d7b65a6dd248a0db5a498d6f49b49dfc251

tools/vsnp/vsnp_build_tables.xml
gregvonkuster: Ah, good to know - thanks!  Fixed here https://github.com/galaxyproject/tools-iuc/pull/3371/commits/d95f5d7b65a6dd248a0db5a498d6f49b49dfc251

tools/vsnp/vsnp_build_tables.xml
gregvonkuster: The first test produces a couple of temporary json files in the output directory, so added the count even though they are not part of the collection.

tools/vsnp/vsnp_add_zero_coverage.py
gregvonkuster: Yeah, I went back-and-forth on this quite a bit.  I've seen these "extension underscore strings" (e.g., _fastq, etc) in real tools outputs in workflows, etc, but IIRC they only happen when inputs are not collections.  So I've cleaned things up here https://github.com/galaxyproject/tools-iuc/pull/3371/commits/d95f5d7b65a6dd248a0db5a498d6f49b49dfc251

tools/vsnp/vsnp_add_zero_coverage.xml
bernt-matthias: The link just shows me the diff. Can you send me the link to the code line in the latest commit.

What I meant is if the cached reference (and maybe the refernce from the history) input should be filtered by this dbkey?

tools/vsnp/vsnp_build_tables.xml
bernt-matthias: Ok. This means that the to json files are also part of the collection or are there 4 xlsx members? If the latter its OK. Otherwise we need to improve the discovery pattern to something like `(?P&lt;designation&gt;.+)\.(?P&lt;ext&gt;xlsx)`

gregvonkuster: These are temporary json files, not xlsx files, and they are not part of the resulting collection.  The json files are used by the Excel formatter to produce the final Excel files - https://github.com/galaxyproject/tools-iuc/blob/master/tools/vsnp/vsnp_build_tables.py#L168

tools/vsnp/vsnp_add_zero_coverage.xml
gregvonkuster: Ah, sorry - here is the link in the latest commit https://github.com/galaxyproject/tools-iuc/blob/master/tools/vsnp/vsnp_add_zero_coverage.py#L49.  If the build select list is filtered, then the user will not be able to change it in the workflow run form, correct?  I think it would be better to allow for this if it is ok.

tools/vsnp/vsnp_build_tables.xml
bernt-matthias: Yes, but the current `discover_datasets`  `pattern` picks up `*.*`. So the json files are included. Either you use a more strict pattern or you remove the json files in the command block.

tools/vsnp/vsnp_add_zero_coverage.xml
bernt-matthias: > Ah, sorry - here is the link in the latest commit https://github.com/galaxyproject/tools-iuc/blob/master/tools/vsnp/vsnp_add_zero_coverage.py#L49. 

Can't see that the dbkey is used in this line.

> If the build select list is filtered, then the user will not be able to change it in the workflow run form, correct? I think it would be better to allow for this if it is ok.

That's a good point. I'm not sure -- but I would expect it. Could you try or ask on gitter?

tools/vsnp/vsnp_build_tables.xml
gregvonkuster: Sorry for the confusion on my end - fixed here https://github.com/galaxyproject/tools-iuc/pull/3371/commits/53042314cc3573913309e6316496e00ebffad7af

tools/vsnp/vsnp_add_zero_coverage.xml
gregvonkuster: I'm not using the Galaxy dbkey itself, but the genome index file associated with the dbkey.  I pass the path to the index file on the command line here https://github.com/galaxyproject/tools-iuc/blob/master/tools/vsnp/vsnp_add_zero_coverage.xml#L46-L50, and then it is parsed in the script here https://github.com/galaxyproject/tools-iuc/blob/master/tools/vsnp/vsnp_add_zero_coverage.py#L49-L53.

In thinking about this, I think its good to include the filter after all - the only place in the workflow run form that a user should have to change the dbkey is in the mapping step which is upstream from these tools.  I've fixed things up here https://github.com/galaxyproject/tools-iuc/pull/3371/commits/5ebb3403cc50d47f5dc5b89548a82671548060d0.

bernt-matthias: Great. Last thing to think about: do you want to have the validator? 

- with the validator the tool will only work with the dbkey specified
- without the cached reference will be filtered if the dbkey is specified, if no dbkey is specified then all options are shown for the cached reference.

tools/vsnp/vsnp_build_tables.xml
```diff
@@ -109,8 +75,8 @@ python '$__tool_directory__/vsnp_build_tables.py'
         </conditional>
     </inputs>
     <outputs>
-        <collection name="excel" type="list">
-            <discover_datasets pattern="__name__" directory="output_excel_dir" format="xlsx" />
+        <collection name="excel" type="list" label="${tool.name} on ${on_string} (tables)">
```
bernt-matthias, line 78: If there is only a single output `${tool.name} on ${on_string}` might be sufficient

tools/vsnp/vsnp_add_zero_coverage.xml
gregvonkuster: Thanks for pointing this out - fixed here https://github.com/galaxyproject/tools-iuc/pull/3371/commits/4e08c1cc982d4c30ad845558bd8e1fb9cb35e529.

tools/vsnp/vsnp_build_tables.xml
gregvonkuster: Ah, yes, good point!  Fixed here https://github.com/galaxyproject/tools-iuc/pull/3371/commits/4e08c1cc982d4c30ad845558bd8e1fb9cb35e529

bernt-matthias (approved):

bernt-matthias:
Ready from your side?

gregvonkuster:
I'm ready - just let me know if there is something I should do.

tools/vsnp/test-data/paired_metrics.txt
```diff
@@ -1,4 +1,4 @@
-Sample: forward
+Sample: forward_fastq_gz
```
bernt-matthias, line 1: Wondering if one needs to derive the sample name from the paired collection?

tools/vsnp/vsnp_determine_ref_from_data.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="vsnp_determine_ref_from_data" name="vSNP: determine reference" version="1.0.0">
+<tool id="vsnp_determine_ref_from_data" name="vSNP: determine reference" version="@WRAPPER_VERSION@.0" profile="@PROFILE@">
```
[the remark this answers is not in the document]
gregvonkuster: Done here https://github.com/galaxyproject/tools-iuc/pull/3371/commits/a153f0fe3a733687519347d18837ead9e64b4450

tools/vsnp/test-data/paired_metrics.txt
gregvonkuster: This would certainly be an improvement, is it possible to do this with an identifier associated with the dataset?  

Similarly, I have a tool that I'm using in the vSNP workflow that gets all sample names from the input files (via the file name) and outputs a list of the file names that can be accepted by the Parse parameter value expression tool.  These sample names are then used as values for the Read Group Identifier parameter in the bwa-mem tool.  Do you know of a better approach for this?

bernt-matthias: You take the data sets identifiers as names the links. 

https://github.com/galaxyproject/tools-iuc/blob/a153f0fe3a733687519347d18837ead9e64b4450/tools/vsnp/vsnp_determine_ref_from_data.xml#L27

If you determine the sample name from read1 only then I guess it is sufficient to get the name from  `input_type_cond.reads_collection.element_identifier` / `input_type_cond.reads_collection.name` (unsure).

Otherwise I would suggest to add sample_name parameter to the python scripts and pass it directly. Maybe you can even skip all the linking in these cases.

gregvonkuster: Ok, thanks!  I'll take a look at this.

gregvonkuster: Improvements here https://github.com/galaxyproject/tools-iuc/pull/3371/commits/4aab5771c15736cce00dee59f2fbaf529c63860a

gregvonkuster:
@bernt-matthias Do you think this can be merged or is there something that you'd like fixed?

bernt-matthias:
Excellent

gregvonkuster:
Thanks so much for everything @bernt-matthias !

PR #3397, opened 2020-12-23
Added AEGeAn toolkit

gallardoalba:
FOR CONTRIBUTOR:

Four tools belonging to the AEGEaN toolkit have been added:

- ParsEval is a program for comparing distinct sets of gene structure annotations for the same sequence(s). 
- CanonGFF3 is a tool for preprocessing GFF3 data.
- LocusPocus is a program for computing gene loci from one or more gene prediction sets.
- GAEVAL is a program for computing coverage and integrity scores for gene models using transcript alignments.


tools/aegean/gaeval.xml
```diff
@@ -0,0 +1,113 @@
+<tool id="aegean_gaeval" name="GAEVAL" version="0.1.0" python_template_version="3.5">
+  <description>- compute coverage and integrity scores for gene models using transcript alignments.</description>
+  <requirements>
```
bgruening, line 3: can you please use macros for the:

* requirements
* citations
* the version (please see other tools with multiple tools)

tools/aegean/gaeval.xml
```diff
@@ -0,0 +1,113 @@
...
+	<!-- Weight options -->
+	<section name="weights-for-calculating-integrity-score"
+		 title="Weights for calculating integrity score"
+		 expanded="False">
+	<param type="float" name="alpha"
+	       label="Introns confirmed, or % expected CDS lenght for single-exon genes (-a)"
+	       value="0.6" min="0" max="1"
+	       help="The percentage of introns confirmed by an alignment gap; for single-exon gene predictions lacking introns, it represents the ratio of the observed CDS length to the expected CDS length. "/>
+	<param type="float" name="beta"
+	       label="Exon coverage (-b)"
```
bgruening, line 28: If you use argument="" You do not need to specify the `(-b)` ... and I would never put this into the label, if you need to put it put it into the help.

tools/aegean/gaeval.xml
```diff
@@ -0,0 +1,113 @@
...
+	-b $beta
+	-g $gamma
+	-e $epsilon
+	-c $expcds
+	-5 $exp5putr
+	-3 $exp3putr > '$output1'
+    ]]></command>
+    <inputs>
+      <param type="data" name="alignmentGFF3file" label="Alignment GFF3 file" format="gff3" />
+        <param type="data" name="genesGFF3file" label="Genes GFF3 file" format="gff3" />
```
bgruening, line 18: indentation

tools/aegean/locuspocus.xml
```diff
@@ -0,0 +1,137 @@
...
+	    -m $refine_options.minoverlap
+	    #if $ilens
+	    "$ilens" "$output2"
+	    #end if
+	    #if $genemap
+	        "$genemap" "$output3"
+	    #end if
+	    $retainids
+	    #if $transmap
+	        "$transmap" "$output4"
```
bgruening, line 23: please use always single-quotes

tools/aegean/locuspocus.xml
```diff
@@ -0,0 +1,137 @@
...
+	    -o '$output1'
+    ]]></command>
+    <inputs>      
+      <param type="data" name="input1" format="gff3" />
+      <param type="integer" name="delta" argument="-l"
+	     label="Gene loci extension"
+	     value="500" min="0" max="1000"
+	     help="When parsing interval loci, use the following delta to extend gene loci and include potential regulatory regions" />
+      <param name="mode" type="select" label="Annotation mode">
+	<option value="" selected="true">Default mode</option>
```
bgruening, line 35: indentation

tools/aegean/locuspocus.xml
```diff
@@ -0,0 +1,137 @@
...
+	     label="Create a mapping from each transcript annotation to its corresponding locus (-t)"
+	     truevalue="-t $output4" falsevalue="" />
+      <param type="text" name="filter"
+	     label="Comma-separated list of feature types to use (-f)"
+	     value="gene" />
+    </inputs>
+    <outputs>
+      <data name="output1" format="txt" />
+      <data name="output2" format="txt">
+	<filter>ilens</filter>
```
bgruening, line 82: indentation

tools/aegean/locuspocus.xml
```diff
@@ -0,0 +1,137 @@
...
+      </data>
+      <data name="output3" format="txt">
+	<filter>genemap</filter>
+      </data>
+      <data name="output4" format="txt">
+	<filter>transmap</filter>
+      </data>
+    </outputs>
+    <tests>
+        <test>
```
bgruening, line 92: please use `num_expect_outputs` as well here in the test and maybe add more tests with more parameters, especially the conditional

tools/aegean/canongff3.xml
```diff
@@ -0,0 +1,72 @@
+<tool id="aegean_canongff3" name="CanonGFF3" version="0.1.0" python_template_version="3.5">
```
bgruening, line 1: python_template_version="3.5" can be removed, but a more recent `profile` could be added

bgruening (commented):
Is this the tool? https://bio.tools/gaeval If so can you add a bio.tools ID and edam ontologies to this tool here (probably as macros)

Please use 4 spaces indentation

tools/aegean/locuspocus.xml
gallardoalba: Could you provide me more details about how to use ```num_expect_outputs```?

tools/aegean/.shed.yml
```diff
@@ -0,0 +1,14 @@
+name: suite_aegean
```
bgruening, line 1: @gallardoalba if you have multiple tools in such a suite, please use a suite-shed.yml file ... example here: https://github.com/galaxyproject/tools-iuc/blob/master/tools/hyphy/.shed.yml

tools/aegean/canongff3.xml
```diff
@@ -0,0 +1,73 @@
...
+            $infer
+            #if $source
+                '$source'
+            #end if
+            -o $output1
+            ]]>
+    </command>
+    <inputs>
+        <param type='data' name='input' label='GFF3 File' format='gff3'/>
+        <param name='infer' type='boolean'
```
bgruening, line 20: This is really nitpicking, but we also have an ordering for the arguments: https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#coding-style

tools/aegean/canongff3.xml
```diff
@@ -0,0 +1,73 @@
...
+    <expand macro='edam_ontology'/>
+    <expand macro='requeriments'/>
+    <command detect_errors='exit_code'>
+        <![CDATA[
+            canon-gff3 '$input'
+            $infer
+            #if $source
+                '$source'
+            #end if
+            -o $output1
```
bgruening, line 15: all inputs, outputs and text-parameter need to be single-quoted

tools/aegean/canongff3.xml
```diff
@@ -0,0 +1,73 @@
...
+    </outputs>
+    <tests>
+        <test>
+            <param name='input' value='TAIR10_GFF3_genes.gff'/>
+            <output name='output1' file='canon-gff3-output_test1.txt'/>
+        </test>
+        <test>
+            <param name='input' value='TAIR10_GFF3_genes.gff'/>
+            <param name='infer' value='True'/>
+            <output name='output1' file='canon-gff3-output_test2.txt'/>
```
bgruening, line 40: add a test for the sourde parameter?

tools/aegean/canongff3.xml
```diff
@@ -0,0 +1,73 @@
...
+            -o $output1
+            ]]>
+    </command>
+    <inputs>
+        <param type='data' name='input' label='GFF3 File' format='gff3'/>
+        <param name='infer' type='boolean'
+            argument='-i' label='Infer gene feature for transcripts lacking it'
+            help='For  transcript  features  lacking an explicitly declared gene feature as a parent, create this feature on-they-fly.'
+            truevalue='--infer' falsevalue=''/>
+        <param name='source' type='text' value=''
```
bgruening, line 24: Do you think we should add a validator here and explude a few character as input

tools/aegean/gaeval.xml
```diff
@@ -0,0 +1,127 @@
+<tool id="aegean_gaeval" name="GAEVAL" version="@TOOL_VERSION@" profile="20.09">
+    <description>- compute coverage and integrity scores for gene models using transcript alignments.</description>
+    <xrefs>
+        <xref type='bio.tools'>gaeval</xref>
+    </xrefs>
+    <macros>                                                                                           
+        <import>macros.xml</import>
+    </macros>
```
bgruening, line 8: should xref be moved into the macros? Oh I see, so the other subtools don't have this bio.tools ID? I would have assumed that they all have the same.

tools/aegean/gaeval.xml
```diff
@@ -0,0 +1,127 @@
+<tool id="aegean_gaeval" name="GAEVAL" version="@TOOL_VERSION@" profile="20.09">
```
bgruening, line 1: profile=20.01 for all tools might be better, so slightly older Galaxy instance can install them as well

tools/aegean/gaeval.xml
```diff
@@ -0,0 +1,127 @@
...
+            -b $weights.beta
+            -g $weights.gamma
+            -e $weights.epsilon
+            -c $expected.expcds
+            -5 $expected.exp5putr
+            -3 $expected.exp3putr > '$output1'
+            ]]>
+    </command>
+    <inputs>
+        <param type='data' name='input1' label='Alignment GFF3 file' format='gff3' />
```
bgruening, line 24: maybe the names can be more meaningful ... eg. `alignment` and `genes`?

tools/aegean/gaeval.xml
```diff
@@ -0,0 +1,127 @@
...
+    <tests>
+        <test expect_num_outputs="1">
+            <param name='input1' value='TAIR10_GFF3_alignment.gff'/>
+            <param name='input2' value='TAIR10_GFF3_genes.gff'/>
+            <output name='output1' file='gaeval_output_test1.txt'/>
+        </test>
+        <test expect_num_outputs="1">
+            <param name='input1' value='TAIR10_GFF3_alignment.gff'/>
+            <param name='input2' value='TAIR10_GFF3_genes.gff'/>
+            <param name='alpha' value='0.5'/>
```
bgruening, line 74: I'm not sure why this works, but usually we need to open here a section as well.

tools/aegean/locuspocus.xml
```diff
@@ -0,0 +1,171 @@
...
+            $retainids
+            #if $transmap
+                '$transmap' '$output4'
+            #end if
+            -f $filter
+            -o '$output1'
+            ]]>
+    </command>
+    <inputs>
+        <param type='data' name='input1' format='gff3' />
```
bgruening, line 33: better param name?

bgruening: label is missing and optionally help as well

tools/aegean/locuspocus.xml
```diff
@@ -0,0 +1,171 @@
...
+        <param name='mode' type='select' label='Annotation mode'>
+            <option value='' selected='true'>Default mode</option>
+            <option value='--skipends'> Exclude unannotated iLoci at either end of the sequence</option>
+            <option value='--endsonly'> Report only incomplete iLocus fragments at the unannotated ends of sequences</option>
+        </param>
+        <param type='boolean' name='skipiloci'
+            label='Do not report intergenic iLoci (-y)'
+            truevalue='---skipiiloci' falsevalue='' />
+        <section name='refine_options' title='Refine options' expanded='False'>
+            <conditional name='refine_parameters'>
```
bgruening, line 47: This is interesting ... you have a conditional with only one parameter. Do you think the conditional is really needed than?

tools/aegean/locuspocus.xml
```diff
@@ -0,0 +1,171 @@
...
+                        truevalue='-c' falsevalue=''
+                        help='(-c)' />
+                </when>
+            </conditional>
+            <param type='integer' name='minoverlap' argument='(-m)'
+                label='Minimum number of overlapping nucleotides'
+                value='1' max='20' min='1'
+                help='The minimum number of nucleotides two genes must overlap to be grouped in the same iLocus' />
+        </section>
+        <param type='boolean' name='ilens'
```
bgruening, line 67: can this be a select list for output files? Maybe multiple="true" and optional="true"?

tools/aegean/parseval.xml
```diff
@@ -0,0 +1,90 @@
...
+        <![CDATA[
+            parseval '$input1' '$input2' 
+            -l $delta
+            -t $maxtrans
+            -w
+            -o '$output'
+            ]]>
+    </command>
+    <inputs>
+        <param type='data' name='input1' format='gff3' />
```
bgruening, line 18: better names?

bgruening (commented):
Thanks @gallardoalba

tools/aegean/canongff3.xml
```diff
@@ -0,0 +1,73 @@
+<tool id='aegean_canongff3' name='CanonGFF3' version='@TOOL_VERSION@' profile='20.09'>
+    <description>- pre-process GFF3 files, removing all features not directly related to protein-coding genes</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro='edam_ontology'/>
+    <expand macro='requeriments'/>
```
bernt-matthias, line 7: It might be worth fixing the typo in the macro name.

tools/aegean/canongff3.xml
```diff
@@ -0,0 +1,73 @@
...
+            #if $source
+                '$source'
+            #end if
+            -o $output1
+            ]]>
+    </command>
+    <inputs>
+        <param type='data' name='input' label='GFF3 File' format='gff3'/>
+        <param name='infer' type='boolean'
+            argument='-i' label='Infer gene feature for transcripts lacking it'
```
bernt-matthias, line 21: What is "it" refering to?

tools/aegean/canongff3.xml
```diff
@@ -0,0 +1,73 @@
...
+            help='For  transcript  features  lacking an explicitly declared gene feature as a parent, create this feature on-they-fly.'
+            truevalue='--infer' falsevalue=''/>
+        <param name='source' type='text' value=''
+            label='Reset the source of each feature'
+            optional='true' argument='-s'
+            help='Reset the source of each feature to the given value'/>
+    </inputs>
+    <outputs>
+        <data name='output1' format='txt' />
```
bernt-matthias, line 30: I think the output type should be gff3

tools/aegean/canongff3.xml
```diff
@@ -0,0 +1,73 @@
...
+    </outputs>
+    <tests>
+        <test>
+            <param name='input' value='TAIR10_GFF3_genes.gff'/>
+            <output name='output1' file='canon-gff3-output_test1.txt'/>
+        </test>
+        <test>
+            <param name='input' value='TAIR10_GFF3_genes.gff'/>
+            <param name='infer' value='True'/>
```
bernt-matthias, line 39: Use lower case `true` here and everywhere else.

tools/aegean/canongff3.xml
```diff
@@ -0,0 +1,73 @@
+<tool id='aegean_canongff3' name='CanonGFF3' version='@TOOL_VERSION@' profile='20.09'>
```
bernt-matthias, line 1: I often like to have the suite name in the tool name as well, e.g. `name="AEGeAn CanonGFF3"`. Up to you. 

Also do you think that the dash in the beginning of the description is necessary?

tools/aegean/gaeval.xml
bernt-matthias: @bgruening this is not necessary at the moment - but may be clearer. May change in future profiles: https://github.com/galaxyproject/galaxy/pull/9493

tools/aegean/locuspocus.xml
```diff
@@ -0,0 +1,171 @@
...
+            truevalue='-T' falsevalue='' />
+        <param type='boolean' name='transmap'
+            label='Create a mapping from each transcript annotation to its corresponding locus (-t)'
+            truevalue='-t $output4' falsevalue='' />
+        <param type='text' name='filter'
+            label='Comma-separated list of feature types to use (-f)'
+            value='gene' />
+    </inputs>
+    <outputs>
+        <data name='output1' format='txt' />
```
bernt-matthias, line 84: Are those all txt files or are some of them tabular or even gff?

tools/aegean/gaeval.xml
gallardoalba: Should I include the same xref for all of them then?

bgruening: I think so, they belong to the same tool suite.

tools/aegean/locuspocus.xml
gallardoalba: Well, I think that it is necessary because the cds parameter should only be available when one of the conditions is satisfied (value == --refine), What do you think?

tools/aegean/canongff3.xml
gallardoalba: I have seen it in some tools, but I can remove it.

bernt-matthias: Up to you. Just my personal preference. Check how it looks best to you with planemo serve.

tools/aegean/locuspocus.xml
gallardoalba: Indeed those were tabular, and also the outputs generated by the ParsEval and GAEVAL tools.

tools/aegean/.shed.yml
gallardoalba: Could you give me more details about which part should I modify? Should I just need to add it? 
`auto_tool_repositories:
  name_template: "{{ tool_id }}"
  description_template: "Wrapper for the HyPhy batch operation: {{ tool_name }}"`

bgruening: Please see here: https://planemo.readthedocs.io/en/latest/standards/docs/best_practices/shed_yml.html#advanced-parameters

Also there are a lot of examples in this repo.

tools/aegean/canongff3.xml
gallardoalba: I think that it is difficult to establish a validator because there is not a defined set of potential sources. But I think that adequate sources can be suggested in the help parameter.

bgruening: But we know that it needs to be alphanumeric characters, isn't it?

tools/aegean/locuspocus.xml
bgruening: Why not offering only this boolean parameter and call the label "Handle overlapping genes by using CDS rather than UTRs". And set it to false by default.

Would this not cover all cases?

tools/aegean/canongff3.xml
gallardoalba: Sure

tools/aegean/.shed.yml
gallardoalba: I guess that now is correct.

bgruening:
Test file canon-output_test3.gff3 is missing and one other test is failing with `    Please only use letters, numbers or underscores in sample label`

tools/aegean/locuspocus.xml
gallardoalba: But it would allow to enable the cds parameter when the refine parameter is set as Default (which is supposed not to be allowed). wouldn't it?

tools/aegean/canongff3.xml
gallardoalba: I renamed it, and I guess that it would be clear now with the info provided with the help parameter.

tools/aegean/locuspocus.xml
bernt-matthias: You probably want to include `$refine_options.refine_parameters.refine` somewhere in the command section... then it should be fine.

tools/aegean/locuspocus.xml
gallardoalba: I implemented it in that way, but it doesn't work properly. Could you check it?

tools/aegean/locuspocus.xml
```diff
@@ -0,0 +1,164 @@
...
+            <option value='--endsonly'> Report only incomplete iLocus fragments at the unannotated ends of sequences</option>
+        </param>
+        <param name='skipiloci' type='boolean'            
+            truevalue='---skipiiloci' falsevalue=''
+            label='Do not report intergenic iLoci (-y)' />
+        <param name='retainids' argument='-T' type='boolean'
+            truevalue='--retainids' falsevalue=''
+            label='Retain original feature IDs'
+            help='Retain original feature IDs from input files; conflicts will arise if input contains duplicated ID values'/>
+        <param name='outputfiles' type='select' display='checkboxes' label='Output files'
```
bernt-matthias, line 45:
```suggestion
      <param name="outputfiles" type="select" display="checkboxes" label="Output files" multiple="true" optional="true">
            <option value="ilens"'>Create file with the lenghts of each intergenic iLocus</option>
            <option value="genemap">Create a mapping from each gene annotation to its correspondig locus</option>
            <option value="transmap">Create a mapping from each transcript annotation to its correspondent locus</option>
        </param>
```

Good idea to use `value='--ilens "$output_ilens"'`, but I think this does not work.

wondering why the linter does not stumble over `multiple="true"`?

tools/aegean/locuspocus.xml
```diff
@@ -0,0 +1,164 @@
...
+            -l $delta
+            $mode
+            $skipiloci
+            #if $refine_options.refine_parameters.refine
+                $refine_options.refine_parameters.cds
+            #end if
+            -m $refine_options.minoverlap
+	        #if $outputfiles
+                $outputfiles
+            #end if
```
bernt-matthias, line 21: should be something like 

```
            #if $outputfiles
                #if "ilens" in outputfiles
                    --ilens `$output_ilens`
                #end if
                ...
            #end if
```

tools/aegean/locuspocus.xml
```diff
@@ -0,0 +1,164 @@
...
+            <param name='minoverlap' argument='(-m)' type='integer' 
+                max='20' min='1' value='1' 
+                label='Minimum number of overlapping nucleotides'
+                help='The minimum number of nucleotides two genes must overlap to be grouped in the same iLocus' />
+        </section>
+    </inputs>
+    <outputs>
+        <data name='output' format='tabular' />
+        <data name='output_ilens' format='tabular'>
+            <filter>'--ilens "$output_ilens"' in  outputfiles</filter>
```
bernt-matthias, line 79:
```
<filter>outputfiles and "ilens" in outputfiles</filter>
```

gallardoalba: Thanks!

gallardoalba: Is there any difference between using single or double quotes in strings?

tools/aegean/.shed.yml
bernt-matthias: Please see here: https://github.com/galaxyproject/tools-iuc/blob/master/tools/dada2/.shed.yml (note that the duplicate category entry is not necessary).

gallardoalba: Thanks @bernt-matthias, when should I use auto_tool_repositories in the .shed.yml file? I read about it but I'm confused because it does not appear in many cases.

bernt-matthias: I think it's the way to go if you want to add a set of tools in a single directory.

gallardoalba: I think that now it fits with that structure.

tools/aegean/gaeval.xml
gallardoalba: Done!

tools/aegean/gaeval.xml
```diff
@@ -0,0 +1,125 @@
...
+            <param name='genesgff3' value='TAIR10_GFF3_genes.gff'/>
+            <output name='output' file='gaeval_output_test1.txt'/>
+        </test>
+        <test expect_num_outputs="1">
+            <param name='alignmentgff3' value='TAIR10_GFF3_alignment.gff'/>
+            <param name='genesgff3' value='TAIR10_GFF3_genes.gff'/>
+            <param name='$weights.alpha' value='0.5'/>
+            <param name='$weights.beta' value='0.4'/>
+            <param name='$weights.gamma' value='0.05'/>
+            <param name='$weights.epsion' value='0.05'/>
```
bernt-matthias, line 75:
```suggestion
            <param name='weights|alpha' value='0.5'/>
            <param name='weights|beta' value='0.4'/>
            <param name='weights|gamma' value='0.05'/>
            <param name='weights|epsion' value='0.05'/>
```

tools/aegean/locuspocus.xml
bgruening: @gallardoalba this filter is a python expression. So the version Matthias suggested is correct and in his version double and single quotes do not matter. In cheetah they do matter and you should use single-quotes.

tools/aegean/gaeval.xml
```diff
@@ -0,0 +1,125 @@
...
+            <output name='output' file='gaeval_output_test2.txt'/>
+        </test>
+        <test expect_num_outputs="1">
+            <param name='alignmentgff3' value='TAIR10_GFF3_alignment.gff'/>
+            <param name='genesgff3' value='TAIR10_GFF3_genes.gff'/>
+            <param name='$expected.expcds' value='20'/>
+            <param name='$expected.exp5putr' value='150'/>
+            <param name='$expected.exp3putr' value='120'/>
```
bernt-matthias, line 83:
```suggestion
            <param name='expected|expcds' value='20'/>
            <param name='expected|exp5putr' value='150'/>
            <param name='expected|exp3putr' value='120'/>
```

tools/aegean/gaeval.xml
bgruening: Can we please not do this and do this instead: https://github.com/galaxyproject/tools-iuc/blob/3da28c7772346e1872b6d768b904305be0c61db7/tools/delly/call.xml#L135

tools/aegean/test-data/gaeval_output_test3.txt
```diff
@@ -46,7 +46,7 @@ Chr1	TAIR10	exon	8417	8464	.	-	.	Parent=AT1G01020.1
 Chr1	TAIR10	exon	8571	8737	.	-	.	Parent=AT1G01020.1
 Chr1	AEGeAn::GAEVAL	start_codon	8664	8666	.	-	.	Parent=AT1G01020.1
 Chr1	TAIR10	five_prime_UTR	8667	8737	.	-	.	Parent=AT1G01020.1
-Chr1	TAIR10	mRNA	6790	8737	.	-	.	ID=AT1G01020.2;Parent=AT1G01020;Name=AT1G01020.2;index=1;gaeval_coverage=0.000;gaeval_integrity=0.068
+Chr1	TAIR10	mRNA	6790	8737	.	-	.	ID=AT1G01020.2;Parent=AT1G01020;Name=AT1G01020.2;index=1;gaeval_coverage=0.000;gaeval_integrity=0.074
```
bgruening, line 49: @gallardoalba it can be that those numbers are a little bit random. Numerical instability etc ... in those cases, you can use diff_lines or some asserts in the test. Even regex in those files are allowed.

bgruening (approved):
Thanks @gallardoalba!

tools/aegean/locuspocus.xml
```diff
@@ -0,0 +1,173 @@
+<tool id='aegean_locuspocus' name='AEGeAn LocusPocus' version='@TOOL_VERSION@' profile='20.01'>
+    <description> calculate locus coordinates for the given gene annotation</description>
+    <macros>                                                                                           
+        <import>macros.xml</import>
+    </macros>                                                                                          
+    <expand macro='xrefs'/>
+    <expand macro='edam_ontology'/>
+    <expand macro='requirements'/>
+    <command detect_errors='exit_code'>
```
bernt-matthias, line 9: A `version_command` tag would also be nice https://docs.galaxyproject.org/en/master/dev/schema.html#tool-version-command

tools/aegean/locuspocus.xml
```diff
@@ -0,0 +1,173 @@
...
+            -o '$output'
+            ]]>
+    </command>
+    <inputs>
+        <param name='genesgff3' type='data' format='gff3' label="GFF3 File" />
+        <param name='delta' argument='-l' type='integer' 
+            min='0' max='1000' value='500'
+            label='Gene loci extension'
+            help='When parsing interval loci, use the following delta to extend gene loci and include potential regulatory regions' />
+        <param name='mode' type='select' label='Annotation mode'>
```
bernt-matthias, line 41: Can't `--skipends` and `--endsonly` be combined: 

```
  iLocus parsing:
    -l|--delta: INT        when parsing interval loci, use the following
                           delta to extend gene loci and include potential
                           regulatory regions; default is 500
    -s|--skipends          when enumerating interval loci, exclude
                           unannotated (and presumably incomplete) iLoci at
                           either end of the sequence
    -e|--endsonly          report only incomplete iLocus fragments at the
                           unannotated ends of sequences (complement of
                           --skipends)
    -y|--skipiiloci        do not report intergenic iLoci
```

If so I suggest to either: 
- make the select `multiple="true" optional="true"`, remove the "empty option" (Default), and include  `--skipiiloci` as additional option.
- transform this into two booleans for `--skipends` and `--endsonly`
`--skipiloci`

tools/aegean/locuspocus.xml
```diff
@@ -0,0 +1,173 @@
...
+                </param>
+                <when value=''/>
+                <when value='--refine'>
+                    <param name='cds' argument='-c' type='boolean' 
+                        truevalue='--cds' falsevalue=''
+                        label='Use CDS rather than UTRs'
+                        help='(-c)' />
+                </when>
+            </conditional>
+            <param name='minoverlap' argument='(-m)' type='integer'
```
bernt-matthias, line 79: does `--minoverlap` make sense without `--refine`, or should it be included in the conditional? If so one could also think about removing the section and only leave the conditional.

tools/aegean/locuspocus.xml
```diff
@@ -0,0 +1,173 @@
...
+                #if "transmap" in $outputfiles
+                    --transmap '$output_transmap'
+                #end if
+            #end if
+            $retainids
+            -f '$filter'
+            -o '$output'
+            ]]>
+    </command>
+    <inputs>
```
bernt-matthias, line 35: I kind of like the grouping of parameters in the `--help` output. Maybe we should do the exact same thing with sections here. This might also help command line users to orientate: 

iLocus parsing
Refinement options:
Output options:
Input options:

Also some of the options are missing. Are there good reasons? (I certainly accept time as a good reason :) .. your work is already more than a good start .. also please consider my comments as suggestions and not critic)?

tools/aegean/macros.xml
```diff
@@ -0,0 +1,36 @@
...
+        <edam_topics>
+            <edam_topic>topic_3511</edam_topic>
+        </edam_topics>
+        <edam_operations>
+            <edam_operation>operation_0362</edam_operation>
+        </edam_operations>
+    </xml>
+    <xml name="xrefs">
+      <xrefs>
+        <xref type='bio.tools'>gaeval</xref>
```
bernt-matthias, line 12: Seems that bio.tools is a bit outdated and inprecise here. It lists gaeval as perl tool. Also the others are missing. Maybe this can be updated?

tools/aegean/locuspocus.xml
```diff
@@ -0,0 +1,173 @@
...
+        <param name='genesgff3' type='data' format='gff3' label="GFF3 File" />
```
bernt-matthias, line 36: Seems that locuspocus also accepts multiple gff3 files. Would this be useful?

tools/aegean/parseval.xml
```diff
@@ -0,0 +1,91 @@
...
+    <description> compare two sets of gene annotations for the same sequence.</description>
+    <macros>                                                                                           
+        <import>macros.xml</import>
+    </macros>                                                                                          
+    <expand macro='xrefs'/>
+    <expand macro='edam_ontology'/>
+    <expand macro='requirements'/>
+    <command detect_errors='exit_code'>
+        <![CDATA[
+            parseval '$referencegff3' '$predictiongff3'
```
bernt-matthias, line 11: I have just noticed that parseval has a html output which could be quite fancy in the Galaxy environment...

bernt-matthias (approved):
Great work @gallardoalba 

A few more comments from my side .. I consider all of them as optional. The tools are already good to go. 

Just let us know if you want to implement any of them or not.

gallardoalba:
> Great work @gallardoalba
> 
> A few more comments from my side .. I consider all of them as optional. The tools are already good to go.
> 
> Just let us know if you want to implement any of them or not.

Sure, I'll include it.

tools/aegean/locuspocus.xml
gallardoalba: Currently is not possible to combine the --skipends and the --endonly parameter (both are incompatible).

tools/aegean/locuspocus.xml
gallardoalba: According the documentation --minoverlap doesn't require the --refine parameter, but because both are included under the _Refinement options_ section in the documentation, I considered that should be included under the same section.

tools/aegean/parseval.xml
gallardoalba: Sure, I'll include it.

tools/aegean/macros.xml
gallardoalba: Do you mean to remove it?

tools/aegean/locuspocus.xml
gallardoalba: Yes, sure, it has sense.

gallardoalba: I included the sections and the _parents_ and _pseudo_ parameters, but now I'm not sure how to modify the filter tags, because ` <filter>output_options.outputfiles and "ilens" in output_options.outputfiles</filter>` doesn't work.

tools/aegean/locuspocus.xml
gallardoalba: Sure, also gaeval and canongff3 accept multiple inputs. It has been fixed.

tools/aegean/parseval.xml
bernt-matthias: docs are here https://planemo.readthedocs.io/en/latest/writing_advanced.html#single-html-output

tools/aegean/macros.xml
bernt-matthias: No. Was just wondering if this can be improved there.

tools/aegean/locuspocus.xml
bernt-matthias: This should work as you suggested

gallardoalba: Something should be wrong because it doesn't work.

tools/aegean/parseval.xml
gallardoalba: I included the html and csv outputs, and also some additional parameters and tests. But I have the same problem with the filter tag.

tools/aegean/macros.xml
gallardoalba: I think that the only alternative is to remove it because there are not additional entries in bio.tools.

bernt-matthias:
Hi @gallardoalba I fixed the outputtype problem here: https://github.com/galaxyproject/tools-iuc/pull/3397/commits/3859099e9ee9fafbaf4e349a11591d734136eccb

Could you also use `argument` instead of `name` as in https://github.com/galaxyproject/tools-iuc/pull/3397/commits/37fed22b26404010bc5c4dbe9295df0b11e542a1 in the other tools. The name is generated automatically and the advantage is that the user will see the argument in the parameter help.

bernt-matthias:
Btw. is it intended that sections are not expanded?

gallardoalba:
> Btw. is it intended that sections are not expanded?

HI!, do you think that it is better to expand them by default? Also,  I'm trying to solve the html issue (link the html file with the css which is generated in the same folder), but not sure what it doesn't work.

bernt-matthias:
> > Btw. is it intended that sections are not expanded?
> 
> HI!, do you think that it is better to expand them by default? 

Depends. I would expand them if the intention of sections is to structure the options and hide them if they contain only advanced options. 

> Also, I'm trying to solve the html issue (link the html file with the css which is generated in the same folder), but not sure what it doesn't work.

What exactly is the problem. I could have a look tomorrow.

gallardoalba:
> > > Btw. is it intended that sections are not expanded?
> > 
> > 
> > HI!, do you think that it is better to expand them by default?
> 
> Depends. I would expand them if the intention of sections is to structure the options and hide them if they contain only advanced options.
> 
> > Also, I'm trying to solve the html issue (link the html file with the css which is generated in the same folder), but not sure what it doesn't work.
> 
> What exactly is the problem. I could have a look tomorrow.

Thanks! Finally Bjorn has solved the problem with the css (it was due to configuration parameter in the local Galaxy test instance). About the sections, I have expanded those that I consider basics.

bernt-matthias:
Excellent. Just noticed that the html output is incomplete (reported upstream here: https://github.com/BrendelGroup/AEGeAn/issues/257). Since the browser still renders this nicely we can go ahead and merge this anyway .. or wait for the feedback of the author.

bgruening (approved):

bgruening:
@bernt-matthias your call :)

bernt-matthias:
Lets wait for the fix. Seems that @gallardoalba is on it.

gallardoalba:
> Lets wait for the fix. Seems that @gallardoalba is on it.

Fixed. Let me know if any additional modification is necessary.

bernt-matthias:
Great work @gallardoalba

gallardoalba:
> Great work @gallardoalba

Thank you and @bgruening for your help.

Issue #3392, opened 2020-12-20
Tools request - chloroplast and mitochondrial assemblies


dougsd3:
One thing thet really called my attention was the lack or chloroplast/mitochondrial assembly tools in Galaxy. Recently, Novoplasty was included, but I feel others could be added.

I would recommend:

1- Getorganelle: https://github.com/Kinggerm/GetOrganelle (assembles chloroplast and mitochondria)
2- Organelle_PBA: https://github.com/aubombarely/Organelle_PBA (devoted to PacBio reads)
3- Chloro: https://github.com/sestaton/Chloro (an abandoned project who might be adopted by Galaxy)
4- MITObin (devoted for mitochondria, but I've heard it works well with chloroplasts): https://github.com/chrishah/MITObim https://github.com/galaxyproject/tools-iuc/pull/3400
4- OGA (Organelle Genome assembler): https://github.com/quxiaojian/OGA

These requests are quite in order of priority I would recommend.

best regards
Douglas

bernt-matthias:
Is there an order in the list? Do you have any favorites?

bernt-matthias:
Could you check which are allready on bioconda?

dougsd3:
I put them pretty much in the order.

I'll if they're available at bioconda.

Thanks
Douglas


Em dom, 20 de dez de 2020 13:47, M Bernt <notifications@github.com>
escreveu:

dougsd3:
Getorganelle is the only one avaliable in bioconda, but it is the most up-to-date one. I would strongly recommend going further with it, because it fits to mitochondrial and chloroplast genomes

dougsd3:
https://bioconda.github.io/recipes/getorganelle/README.html

bernt-matthias:
@dougsd3 even if not your favourite you may want to have a look here https://github.com/galaxyproject/tools-iuc/pull/3400 .. comments welcome. 

For GetOrganelle some changes might be necessary https://github.com/Kinggerm/GetOrganelle/issues/64

bernt-matthias:
Maybe the list should be extended by 

- https://github.com/ndierckx/NOVOPlasty
- https://github.com/ibest/ARC
- https://git.metabarcoding.org/org-asm/org-asm

dougsd3:
that's good
-- 
Douglas Silva Domingues, PhD
Research Professor
Group of Genomics and Transcriptomes in Plants
Institute of Biosciences at Rio Claro
[...]

PR #3400, opened 2020-12-29
Add MITObim 

bernt-matthias:
One of the tools requested here: https://github.com/galaxyproject/tools-iuc/issues/3392

FOR CONTRIBUTOR:


tools/mitobim/mitobim.xml
```diff
@@ -0,0 +1,137 @@
...
+    #if str($mismatch) != ""
+        --mismatch $mismatch
+    #end if
+    #if str($min_cov) != ""
+        --min_cov $min_cov
+    #end if
+    #if str($min_len) != ""
+        --min_len $min_len
+    #end if
+## TODO    --redirect_tmp    redirect temporary output to this location (useful in case you are running MITObim on an NFS mount)
```
bgruening, line 44: You could send this to $TEMP

tools/mitobim/mitobim.xml
```diff
@@ -0,0 +1,137 @@
...
+                <option value="--maf">extracts reference from maf file created by previous MITObim iteration/MIRA assembly</option>
+            </param>
+            <when value="--quick">
+                <param name="reference" type="data" format="fasta" label="Reference fasta file" argument="--quick"/>
+            </when>
+            <when value="--maf">
+                <param name="reference" type="data" format="maf" label="Reference MAF file" argument="--maf"/>
+            </when>
+        </conditional>
+        <param argument="--ref" type="text" value="" label="Reference ID" help="If resuming, use the same as in previous iteration/initial MIRA assembly"/>
```
bgruening, line 67: should we add a sanitizer here ...

tools/mitobim/mitobim.xml
```diff
@@ -0,0 +1,137 @@
...
+        <param argument="--readpool" type="data" format="fastq,fastq.gz" label="readpool" help="read pairs need to be interleaved for full functionality of the '-pair' option below"/>
+        <param argument="--sample" type="text" value="" label="Sample ID" help="If resuming, the sampleID needs to be identical to that of the previous iteration / MIRA assembly"/>
+        <param argument="--start" type="integer" value="" optional="true" label="iteration to start with" help="defaults to 0 when using '--quick' reference"/>
+        <param argument="--end" type="integer" value="" optional="true" label="iteration to end with" help="default=startiteration, i.e. if not specified otherwise stop after 1 iteration"/>
+        <param argument="--kbait" type="integer" value="31" label="kmer for baiting stringency" help=""/>
```
bgruening, line 72: is it also true here that uneven kmers are preferable? Should we add a validator here to filter out even numbers?

tools/mitobim/mitobim.xml
```diff
@@ -0,0 +1,137 @@
...
+        <param argument="--platform" type="select" label="sequencing platform" help="">
+            <option value="solexa">Solexa</option>
+            <option value="iontor">ION Torrent</option>
+            <option value="454">445</option>
+            <option value="pacbio">PacBIO</option>
+        </param>
+        <param argument="--denovo" type="boolean" checked="false" truevalue="--denovo" falsevalue="" label="run MIRA in denovo mode" help=""/>
```
bgruening, line 79: Should all labels start with a capital letter?

tools/mitobim/mitobim.xml
```diff
@@ -0,0 +1,137 @@
...
+    <macros>
+        <token name="@TOOL_VERSION@">1.9.1</token>
+    </macros>
+    <requirements>
+        <requirement type="package" version="@TOOL_VERSION@">mitobim</requirement>
+    </requirements>
+    <version_command>MITObim.pl --version</version_command>
+    <command detect_errors="exit_code"><![CDATA[
+    export LC_ALL=C &&
+    cp '$reftype.reference' reference.$reftype.reference.ext &&
```
bgruening, line 12: what does the $reftype do here ...

bernt-matthias: There are two ways to run mitobim: `--maf MAFFILE` or `--quick FASTAFILE`. The `reftype` conditional makes the distinction of the input formats. 

I also though about making this dependent on the ext of the input data, but decided to make this explicit. Maybe for workflows this would be better. What do you think?

tools/mitobim/mitobim.xml
bernt-matthias: I think this is massed directly to mira. I checked the manual and could not find such a restriction, even if the docs use odd numbers for all examples that I have seen. 

But I have found a max for the kmer length. So I added this.

tools/mitobim/mitobim.xml
bgruening: From a developer perspective, I'm always in favor of explicit, but this is user-world. So maybe implicit is better? And as you said you could use on the workflow for both inputs.
There are no downsides here from a user perspective, left?

I don't know your call :)

tools/mitobim/mitobim.xml
bernt-matthias: Maybe. The values are single quoted, so they should not create trouble on the command line. 

Still, as far as I know the ids are used internally in file names which **might** create problems internally in the tool. The help already asks not to use `.` in the ids (but I do not know why).

But, the ids are supposed to match the id in the MAF input (in the `--maf` case), so the corresponding error that this might create if we have a sanitizer here could also be confusing. 

Maybe a validator... this would create a warning on the tool form already.

tools/mitobim/mitobim.xml
bgruening: ok!

tools/mitobim/mitobim.xml
bgruening: Validator sounds good. excluding `'` and `"` and `.`.

tools/mitobim/mitobim.xml
bgruening: ping ...

bernt-matthias: I tried. But then the 3rd test fails (likely a bug). So lets leave it as is.

bgruening (approved):

bgruening:
Thanks @bernt-matthias!

bernt-matthias:
Was just about to add the validator :) .. Cancelled the workflow

PR #3404, opened 2020-12-31
Add the vsnp_statistics tool

gregvonkuster:
FOR CONTRIBUTOR:


tools/vsnp/vsnp_statistics.py
```diff
@@ -0,0 +1,201 @@
...
+
+import argparse
+import gzip
+import os
+import shutil
+
+import numpy
+import pandas
+
+QUALITYKEY = {'!': '0', '"': '1', '#': '2', '$': '3', '%': '4', '&': '5', "'": '6', '(': '7',
```
bernt-matthias, line 11: This should be computable from the ascii value?

Isn't there a library around that can help you with reading fastq? biopython maybe?

gregvonkuster: Possibly, but I'm not sure.  I just used the same approach taken in the USDA vSNP code here https://github.com/USDA-VS/vSNP/blob/master/bin/vsnp_fastq_quality.py#L56

gregvonkuster:
@bernt-matthias Can we resurrect this PR?

tools/vsnp/vsnp_statistics.py
```diff
@@ -0,0 +1,201 @@
...
+        ref_with_coverage, avg_depth_of_coverage, good_snp_count = process_metrics_file(metrics_file)
+        current_sample_df.at[file_name_base, 'Reference with Coverage'] = ref_with_coverage
+        # Average Depth of Coverage
+        current_sample_df.at[file_name_base, 'Average Depth of Coverage'] = avg_depth_of_coverage
+        # Good SNP Count
+        current_sample_df.at[file_name_base, 'Good SNP Count'] = good_snp_count
+        data_frames.append(current_sample_df)
+    excel_df = pandas.concat(data_frames)
+    excel_file_name = "output.xlsx"
+    writer = pandas.ExcelWriter(excel_file_name, engine='xlsxwriter')
```
bernt-matthias, line 116: Can I convince you to output a reasonable tabular format like tsv. xls is pretty much unusable (except for in Excel -- which nobody should be forced to use :)

bernt-matthias (changes requested):

tools/vsnp/vsnp_statistics.py
gregvonkuster: I actually strongly agree with you on this - my career started in the mid 80's in Silicon Valley, so I am averse to everything Microsoft!  :).   This tool is based on the vSNP pipeline from USDA, which for some reason chose to use Excel extensively as outputs https://github.com/USDA-VS/vSNP/blob/master/bin/vSNP_step1.py#L116/.  And the vsnp_build_tables tool produces Excel https://github.com/galaxyproject/tools-iuc/blob/master/tools/vsnp/vsnp_build_tables.py#L59.

But outputting a tabular file here is probably more ideal since column colors don't play a role, etc.  I'll make that change.

tools/vsnp/vsnp_statistics.py
gregvonkuster: I've changed the code to parse the input using SeqIO from biopython, but I'm now getting this exception on 3 of the 4 tests:
```
ValueError: End of file without quality information.
```
So I'll have to fix the test data.  Hopefully that will be the only problem.

gregvonkuster: @bernt-matthias For testing, I've made the following change for handling quality in order to use biopython instead of the hard-coded QUALITYKEY dict:
```
 import shutil
+from functools import partial
 
+from Bio import SeqIO
 import numpy
...
         # Read fastq_file into a data frame.
-        fastq_df = fastq_to_df(fastq_file, gzipped)
+        _open = partial(gzip.open, mode='rt') if gzipped else open
+        with _open(fastq_file) as fh:
+            identifiers = []
+            seqs = []
+            for seq_record in SeqIO.parse(fh, "fastq"):
+                identifiers.append(seq_record.id)
+                seqs.append(seq_record.seq)
+        # Convert lists to Pandas series.
+        s1 = pandas.Series(identifiers, name='id')
+        s2 = pandas.Series(seqs, name='seq')
+        # Gather Series into a data frame.
+        fastq_df = pandas.DataFrame(dict(id=s1, seq=s2)).set_index(['id'])
         total_reads = int(len(fastq_df.index) / 4)
```
The outputs differ, so I'm looking for your input here - with this change, this tool does not produce the same statistics as the USDA tool.
```
--- local_file
+++ history_data
@@ -1,3 +1,3 @@
 	Reference	File Size	Mean Read Length	Mean Read Quality	Reads Passing Q30	Total Reads	All Mapped Reads	Unmapped Reads	Unmapped Reads Percentage of Total	Reference with Coverage	Average Depth of Coverage	Good SNP Count
-Mcap_Deer_DE_SRR650221_fastq_gz	89	1.6 MB	121.0	38.0	      1.00	4317	17063	223	      0.05	8.27%
+Mcap_Deer_DE_SRR650221_fastq_gz	89	1.6 MB	120.6	29.6	      0.52	17268	17063	223	      0.01	8.27%
 "	0.439436	36
```

gregvonkuster: I know that the USDA invested heavily in their version of these tools, and the outputs were vetted.  The PI that I worked with on these Galaxy wrappers did not want outputs changed.  So even though the the code may not be ideal, I don't feel comfortable changing the outputs now.   Is it ok to leave the quality check as is?

tools/vsnp/vsnp_statistics.py
```diff
@@ -0,0 +1,201 @@
...
+
+QUALITYKEY = {'!': '0', '"': '1', '#': '2', '$': '3', '%': '4', '&': '5', "'": '6', '(': '7',
+              ')': '8', '*': '9', '+': '10', ',': '11', '-': '12', '.': '13', '/': '14', '0': '15',
+              '1': '16', '2': '17', '3': '18', '4': '19', '5': '20', '6': '21', '7': '22',
+              '8': '23', '9': '24', ':': '25', ';': '26', '<': '27', '=': '28', '>': '29',
+              '?': '30', '@': '31', 'A': '32', 'B': '33', 'C': '34', 'D': '35', 'E': '36',
+              'F': '37', 'G': '38', 'H': '39', 'I': '40', 'J': '41', 'K': '42', 'L': '43',
+              'M': '44', 'N': '45', 'O': '46', 'P': '47', 'Q': '48', 'R': '49', 'S': '50',
+              'T': '51', 'U': '52', 'V': '53', 'W': '54', 'X': '55', 'Y': '56', 'Z': '57',
+              '_': '1', ']': '1', '[': '1', '\\': '1', '\n': '1', '`': '1', 'a': '1', 'b': '1',
```
bernt-matthias, line 19: Hmm. I think mapping all those chars to `1` may be wrong. See https://en.wikipedia.org/wiki/FASTQ_format#Encoding 

Can you control the FASTQ encoding in biopython?

gregvonkuster: I'm outside of my comfort zone on this as the biology/data is not my area of expertise.  If the USDA approach here is producing invalid results, I'm happy to make the change to use biopython and let the USDA know about this issue.  Is this what you advise?

bernt-matthias: I would vote for biopython, but I guess the best thing is to check with USDA first .. maybe there is a reason

gregvonkuster: Here is the response from the USDA - I've told them that I'll change the Galaxy wrapper to use the biopython approach.
```
I wrote this looking for something that was more transparent.  Tools were giving me different values, so I did this to verify accuracy.  That being said it was only written with Illumina and Ion Torrent in mind.  There is a platform (I think nanopore) that uses lower case characters for poor quality.  I put a filler of 1 in for all lower case to just push these through, so I’m not surprised to hear reviewers question it.  I think it would be fine to change the FASTQ quality metric to the biopython tool.  Although I’m not convinced the higher values from biopython are necessarily accurate.  Maybe it’s a sampling issue, but 38 is suspiciously high.  I don’t like questioning biopython, but I wouldn’t assume it is more accurate in this case.
```

bernt-matthias: > There is a platform (I think nanopore) that uses lower case characters for poor quality.

This is wrong. in all encodings lower case characters are of high quality in all encodings that use them. 

> Maybe it’s a sampling issue, but 38 is suspiciously high. 

Also here I would not agree. If I read the data correctly it was 29.6 before. Ususaly we trim sequences with quality values of 20 or less. So 29.6 sounds suspiciously low to me. 

Maybe you can cross check the mean quality with an independent tool. I guess `fastx_quality_stats` could serve for this purpose.

gregvonkuster: Is it necessary to cross check the mean quality?  Since we've now eliminated the questionable code, can this be merged without doing that?

gregvonkuster:
@bernt-matthias Is this PR ready for merging? :)

bernt-matthias (approved):

gregvonkuster:
@bgruening or @bernt-matthias Can this be merged?  That would leave just 1 more vSNP tool here https://github.com/galaxyproject/tools-iuc/pull/3381, and I'll resurrect that PR by eliminating multiprocessing.

gregvonkuster:
@bernt-matthias since this one has been approved I'd merge it but it looks like a workflow check has to be completed.  I believe I've made some tweaks to this tool in my repo since March, so I'll take a look and add another commit if it turns out to be the case.

PR #3405, opened 2020-12-31
Add the vsnp_sample_names tool

gregvonkuster:
FOR CONTRIBUTOR:


bernt-matthias:
I have the feeling that there is a better way to do this. What will be the input in the workflow that you try to implement? A collection of fastq files .. each representing a sample?

gregvonkuster:
> I have the feeling that there is a better way to do this. What will be the input in the workflow that you try to implement? A collection of fastq files .. each representing a sample?

Yes. this would be the same inputs that we have for the vsnp_determine_ref_from_data tool, so either a single fastq file, a single set of paired fastq files, a collection of single fastq files or a collection of paired fastq files.  The file names are descriptive, something like BCG_Danish_Human_UK_SRR9596061.fastq.gz, and the sample name in this particular case would be BCG_Danish_Human_UK_SRR9596061.

gregvonkuster:
@bernt-matthias @mvdbeek    I'd like to revisit these remaining vSNP PRs, and I'm hoping for some advice - I know how busy you both are, so whenever it's convenient.  I'm starting with this one since it may be the easiest to resolve.  I understand it may not be ideal to create a tool simply to provide the dataset file names to the bwa-mem tool, but the only other option I could come up with was to enhance the bwa-mem tool itself.  And the bash commands to accommodate this would likely be complex and brittle.  Do you have an idea for another, simpler approach for this?  Or would you support enhancing the bash command in the bwa-mem tool?

bernt-matthias:
Thanks for keeping this alive.

Can you try the tool `Extract element identifiers of a list collection`. Maybe you need additional text processing tools (seems that you do some reformatting).

gregvonkuster:
@bernt-matthias Thanks for pointing out this tool ```Extract element identifiers of a list collection``` - I didn't know about it.  It almost does what I need here, but not quite, and the difference is a bit of a problem.

Using an input collection with these 2 elements:
```
Mbov_Capybara_BR_SRR9850824.fastq
Dassie_Dassie_ZA_SRR3745455.fastq
```
The vsnp_sample_names tool will produce a collection consisting of the 2 elements above, and the content of each of those elements is the name of the file.

This collection is then passed to the ```Parse parameter value``` tool which reads the content of each file and passes the file name string to the ```Map with bwa-mem``` tool as the value of both the ```Read group identifier (ID)``` and ```Read group sample name (SM)```.  

This approach produces a header line like this in the output of the ```Map with bwa-mem``` tool for the first sample above:
```
@RG ID:Mbov_Capybara_BR_SRR9850824 SM:Mbov_Capybara_BR_SRR9850824 PL:ILLUMINA PI:250
```
An a header line like this in the output for the second sample:
```
@RG ID:Dassie_Dassie_ZA_SRR3745455 SM:Dassie_Dassie_ZA_SRR3745455 PL:ILLUMINA PI:250
```
However, the ```Extract element identifiers of a list collection``` tool produces a single output, the content of which is all file names:
```
Mbov_Capybara_BR_SRR9850824.fastq
Dassie_Dassie_ZA_SRR3745455.fastq
```
When used as the input to the ```Parse parameter value``` tool, the output is then a single text string:
```
"Mbov_Capybara_BR_SRR9850824.fastq\nDassie_Dassie_ZA_SRR3745455.fastq\n"
```
And then the header line is the same in every output of the ```Map with bwa-mem``` tool:
```
@RG ID:Mbov_Capybara_BR_SRR9850824.fastq__cn__Dassie_Dassie_ZA_SRR3745455.fastq__cn__ SM:Mbov_Capybara_BR_SRR9850824.fastq__cn__Dassie_Dassie_ZA_SRR3745455.fastq__cn__ PL:ILLUMINA
```

Checking or unchecking the ```Remove newlines ?``` option in the ```Parse parameter value``` tool makes no difference here.

So I'll need a tool that reads each line of a text file (i.e., the output of the ```Extract element identifiers of a list collection``` tool) and produces a collection whose elements are named using the line and whose content is the line.

Are you aware of a simple way to get what I'm looking for here?

bernt-matthias:
In case the identifiers of the collection are not relevant but only the content then `Split file to dataset collection` would do the job (using a chunk size of 1). If they are important it should be easy to extend the tool. 

Also `Split file according to the values of a column` might be an option.

bernt-matthias:
The identifiers could also be fixed with `Relabel List Identifiers from contents of a file ` .. you may also want to have a look at `Apply Rule to Collection ` which is quite powerful ..

gregvonkuster:
The combination of ```Extract element identifiers of a list collection``` and ```Split file to collection``` produces what I need for input collections.  Thanks for directing me to these existing options - I'd love to know your process for keeping current with all of the existing tools.  I'm sure I can find one or more tools that will work for single inputs as well, so I'll close this PR to eliminate this unnecessary tool.  Thanks so much @bernt-matthias!

PR #3411, opened 2021-01-03
Add bio.tools IDs to tools

bgruening:
This will add bio.tools IDs to more than 600 Galaxy tools. This time it is manually curated and I think we can get this in soon.

This is a WIP. We will try to add some more ideas to our tools, please feel free to push to this branch.


bgruening:
Ok, done until `f`.

bernt-matthias:
Would it be possible to do this automatically from the bioconda annotations https://github.com/bioconda/bioconda-recipes/issues/26008?

bgruening:
Question: Do we want to merge this and stop the CI ... so that those annotations get added when the next tool version bump is happening or do we push it to the TS forcing a tool update (not upgrade) for all the tools?

bernt-matthias:
Hmm. 

Pro deployment: I expect that many of the tools won't get a PR for a long time. 

Con deployment: One would need to fix all the linter errors. It will need a while until this gets really useful within Galaxy.

bgruening:
> Con deployment: One would need to fix all the linter errors. It will need a while until this gets really useful within Galaxy.

I guess we can overule the linter here and ignore it.

bernt-matthias:
> I guess we can overule the linter here and ignore it.

Meaning manual deployment? I'm fine with this. 

I guess a few others from IUC should comment on this as well?

bgruening:
ping @galaxyproject/iuc

gregvonkuster:
I don't have the context needed for me to add benefit here.  Is there a ticket or details that provide info about the goal that is being accomplished here?

hexylena:
Normally I don't like in-place tool upgrades but maybe makes sense here. I'd love to see this so that we can automate the edam annotations, it starts to make https://github.com/galaxyproject/galaxy/pull/10592 which will be in the next release actually useful/interesting.

missing_biotools.txt
```diff
@@ -0,0 +1,124 @@
+annotatemyids
```
bernt-matthias, line 1: Do we intend to keep this up-to-date? Maybe we should then add a requirement to IUC best practices.

bgruening: Oh, no I will remove this. This is more for @hmenager and @hansioan. I will open an issue on https://github.com/bio-tools/content

bernt-matthias: Still we could add a requirement here https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tool-cross-references, e.g, if the tool is in bio.tools the it must be annotated in the wrapper?

bgruening:
I think that most of the tools have been annotated and we could proceed. We have two possibilities here:

1) we write a small script that extracts EDAM ontologies from bio.tools and adds them to the tools
2) we add plumping to Galaxy to actually fetch (and cache) EDAM ontologies from bio.tools on-demand __iff__ there is a bio.tools ID but __no__ EDAM annotation

Adding EDAM directly to tools has the advantage that we could add version-specific annotations. But what are we doing with that? Are different versions than listed in different sections etc ...

If we agree with 2) we could merge this PR, otherwise, I think we should add the EDAM terms before merging.

@mvdbeek @hexylena @galaxyproject/iuc

missing_biotools.txt
bgruening: Yes, we should add this and encourage people. We might also lint it and suggest potential IDs.

hexylena:
@bgruening you said you had a list of pros/cons for each approach?  The main con that comes to my mind is then required internet connectivity to check with bio.tools which doesn't give me a great feeling.

bgruening:
Ok, that one is ready for review @galaxyproject/iuc

My proposal is we merge that one but do not push to the TS, but let those changes propagate with new tool updates over time.

I settled on only annotating bio.tool IDs, because EDAM ontologies are changing over time and updating them in the tools is cumbersome, especially because they also needs to be updated (by the bio.tools community) in bio.tools.

So my idea would be:

* if a tool has a bio.tools ID __only__ annotate it with bio.tools ID
* if a tool has __no__ bio.tools ID annotate with EDAM ontology directly if possible
* implement in Galaxy a mechanism to retrieve and cache a mapping of bio.tools ID to EDAM ontology and populate tools with annotation on the fly (we could do that even for DOIs and more annotations if we like)

What do you all think about that?

tools/bedtools/macros.xml
```diff
@@ -77,19 +82,6 @@
...
-                label="Specify a genome file that defines the expected chromosome order in the input files." >
-                <option value="" selected="true">No</option>
-                <option value="-g">Yes</option>
-            </param>
-            <when value="-g">
-                <expand macro="input_conditional_genome_file" />
-            </when>
-            <when value="" />
-        </conditional>
-    </xml>
```
hexylena, line 92: is this intentional?

tools/metaphlan2/metaphlan_hclust_heatmap.xml
```diff
@@ -0,0 +1,275 @@
...
+            <param name="font_size" value="7"/>
+            <param name="clust_line_w" value="1"/>
+            <param name="c" value="pink"/>
+            <output name="out_pdf" file="heatmap.pdf" compare="sim_size"/>
+        </test>
+        <test>
+            <param name="in" value="merged_community_profile.tabular"/>
+            <param name="output_format" value="svg"/>
+            <param name="m" value="complete"/>
+            <param name="d" value="hamming"/>
```
hexylena, line 247: new tool?

hexylena (approved):
Love the annotations! Let's do it. This is way easier than doing edam annotations natively

bgruening:
I removed the missing bio.tools files and put it on gist: https://gist.github.com/bgruening/08e60450db1a83792b83c6212ea7db47

tools/metaphlan2/metaphlan_hclust_heatmap.xml
bgruening: yeah, that was a wrong rebase on my side, thanks for spotting, fixed all the comments :)

hexylena:
```
+Linting tool /home/runner/work/tools-iuc/tools-iuc/tools/deseq2/deseq2.xml
Failed linting
Applying linter tool_xsd... FAIL
.. ERROR: Invalid XML found in file: tmp8zw0dlmp. Errors [/tmp/tmp8zw0dlmp:24:0:ERROR:SCHEMASV:SCHEMAV_ELEMENT_CONTENT: Element 'xrefs': This element is not expected.]
```

I think that might be real?

bgruening:
@hexylena can we make sure those IDs are exposed via the GRT? That would enable a nice mapping for openebench and others.

hexylena:
It would be nice for sure! If we wanted that information it would require loading the complete galaxy toolbox in the GRT process to get access to that data which we haven't been doing as much (it's slow and complicated.) We could?

I really wish the TS had a better API for this (or even that just a mapping table existed somewhere) which had tool ID → biotools ID, that would be much easier to work with. I don't suppose that's something planemo monitor to scrape + build?

hmenager (changes requested):
Absolutely awesome to see this, thanks a lot to all contributors!

Only one minor problem with deseq2, where bio.tools ref is included in both the tool XML and as an expanded macro

hmenager:
> I removed the missing bio.tools files and put it on gist: https://gist.github.com/bgruening/08e60450db1a83792b83c6212ea7db47

It looks like a few of these tools are now available in bio.tools:

```sh
cat missing_bio.tools.txt | xargs -I {} -n 1 wget --spider https://bio.tools/api/t/{} 2>&1 | egrep 'https://bio.tools|response...'
```

It gets me **21** 200 responses, and **86** 404s. Out of these 21, there are probably some false positives (e.g. tool named **cat** in bio.tools), but many seem to be recent additions.

bgruening:
@galaxyproject/iuc is it ok that we merge this PR. Cancel the build and let the annotation populate over time into the TS and our Galaxy instances?

mvdbeek:
IMO just update without version bump. What's the downside of not cancelling the upload?

bgruening:
> IMO just update without version bump. What's the downside of not cancelling the upload?

We create a lot of new tool updates without a new version. I'm always scared when we do that. And it means a lot of tool installation downstream.

mvdbeek:
> We create a lot of new tool updates without a new version. I'm always scared when we do that. And it means a lot of tool installation downstream.

These are minor bumps, the tools are updated in place on disk using hg update, there will be no new dependency installation and the update will be very quick. I don't think that is a good reason to not deploy.

bgruening:
I know but in practice :)
If you think its ok to merge and deploy let's do it. The sooner we get these IDs in production the sooner we can play with it in a meaningful way.

mvdbeek (approved):

hexylena:
We're getting the new tool panel, let's merge this so they'll be useful! :)

hexylena (approved):

mvdbeek:
We have some minor conflicts here, and I expect we'll continue to have lint errors, do we merge anyway ?

bgruening:
I fixed the merge conflicts I hope, would be great to get this in before I need to rebase again :(

tools/exonerate/exonerate.xml
```diff
@@ -1,7 +1,10 @@
 <tool id="exonerate" name="Exonerate" profile="16.04" version="@VERSION@+galaxy2">
     <description>pairwise sequence comparison</description>
+    <xrefs>
+        <xref type="bio.tools">exonerate</xref>
+    </xrefs>
     <macros>
-	<import>macros.xml</import>
+	    <import>macros.xml</import>
```
mvdbeek, line 7: @bgruening  There seems to be a tab here ... do you want to fix or ignore this ?

tools/scanpy/macros.xml
```diff
@@ -10,6 +10,11 @@
             <yield />
         </requirements>
     </xml>
+    <xml name="bio_tools">
+        <xrefs>
+            <xref type="bio.tools">obitools</xref>
```
mvdbeek, line 15: Oh, that looks wrong, doesn't it ?

mvdbeek:
```suggestion
            <xref type="bio.tools">scanpy</xref>
```
?

tools/scater/macros.xml
```diff
@@ -9,6 +9,11 @@
             <yield />
         </requirements>
     </xml>
+    <xml name="bio_tools">
+        <xrefs>
+            <xref type="bio.tools">obitools</xref>
```
mvdbeek, line 14:
```suggestion
            <xref type="bio.tools">scater</xref>
```

tools/poretools/macros.xml
```diff
@@ -6,6 +6,11 @@
             <yield />
         </requirements>
     </xml>
+    <xml name="bio_tools">
+        <xrefs>
+            <xref type="bio.tools">obitools</xref>
```
mvdbeek, line 11:
```suggestion
            <xref type="bio.tools">poretools</xref>
```

tools/exonerate/exonerate.xml
bgruening: fixing! :) Latest commit. Thanks!

hexylena:
:tada: !

bgruening:
Wuhu! Thanks a lot for all the reviews! @jmchilton party time!

PR #2612, opened 2019-09-27
SPAdes update to 3.15.2; merge existing SPAdes, rnaSPAdes, metaSPAdes; add biosyntheticSPAdes, coronaSPAdes, metaplasmidSPAdes, metaSPAdes, metaviralSPAdes, plasmidSPAdes, rnaviralSPAdes, truSPAdes

stephanflemming:
- merge SPAdes, rnaSPAdes, metaSPAdes
- update to 3.15.2
- add biosyntheticSPAdes, coronaSPAdes, metaplasmidSPAde, metaSPAdes, metaviralSPAdes, plasmidSPAdes, rnaviralSPAdes, truSPAdes 

parameter overview can be found here https://docs.google.com/spreadsheets/d/1WvAyUy7gVuo0lRG0wuJ0IbjxGJmWgbhdfk__SFHGISw/edit?usp=sharing

PR #3417, opened 2021-01-07
Add memory option to rnaSPAdes

natefoo:
FOR CONTRIBUTOR:

Taken from the regular SPAdes wrapper.

### DIFF
--- a/tools/rnaspades/rnaspades.xml
+++ b/tools/rnaspades/rnaspades.xml
@@ -8,10 +8,15 @@
     </stdio>
     <command>
         <![CDATA[
+
+        if [[ -n \$GALAXY_MEMORY_MB ]]; then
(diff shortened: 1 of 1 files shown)


bernt-matthias (approved):
Its a pitty that spades and rnaspades aren't in the same folder .. then using a macro/token would be easily possible. Or could we relocate one of the tools?

bgruening:
> Its a pitty that spades and rnaspades aren't in the same folder .. then using a macro/token would be easily possible. Or could we relocate one of the tools?

If it is possible to keep the long tool IDs, then yes we could do this.

nsoranzo:
@bernt-matthias @bgruening There is actually https://github.com/galaxyproject/tools-iuc/pull/2612 that is trying this merge.

PR #3420, opened 2021-01-08
Ensure that secondary DESeq2 R script is in <command> for Pulsar

natefoo:
FOR CONTRIBUTOR:

Per @wm75 this is a trick (from @Slugger70) already used in some [other tools](https://github.com/galaxyproject/tools-iuc/blob/master/tools/snpfreqplot/snpfreqplot.xml) for getting Pulsar to transfer required files from the tool directory that don't appear directly in the command line. In this case, `get_deseq_dataset.R` is an include in `deseq2.R`.

From some discussion with @jmchilton on the admins gitter channel we'd like a way to annotate these sorts of things rather than a hack, but it's not on the list of priorities currently.

### DIFF
--- a/tools/deseq2/deseq2.xml
+++ b/tools/deseq2/deseq2.xml
@@ -41,6 +41,7 @@ echo $(R --version | grep version | grep -v GNU)", DESeq2 version" $(R --vanilla
     #end if
 #end if
 
+## This is needed for Pulsar to transfer the file
 cat '$__tool_directory__/get_deseq_dataset.R' > /dev/null &&
 
 #import json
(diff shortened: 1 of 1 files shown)


wm75 (approved):
LGTM, in the sense of should fix your issue.
Very much agree with you though that there should be a proper mechanism for this.

nsoranzo:
@natefoo Can you add a comment mentioning Pulsar, so we can grep the repo and remove all the workarounds when this is fixed?

jmchilton:
I can't believe that works, that is a terrible hack... but could be so much worse. Thanks @natefoo

nsoranzo (approved):
Thanks Nate!

PR #3421, opened 2021-01-11
use planemo ci_find_repos+ci_find_tools

bernt-matthias:
for counting tools and computing chunks.

- [ ] we should think about removing `-exclude packages --exclude deprecated` in favor of including these directories in `.tt_skip` (here and in the PR workflow). Advantage would be that the workflow would also run if those dirs do not exist in other repos. So all exclusions would be controlled in `.tt_skip`

FOR CONTRIBUTOR:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -66,21 +66,17 @@ jobs:
       with:
         fetch-depth: 1
     - name: Planemo ci_find_repos
-      run: planemo ci_find_repos --exclude packages --exclude deprecated --exclude_from .tt_skip --output changed_repositories.list
-    - name: Show repo list
-      run: cat changed_repositories.list
-    - name: Planemo ci_find_tools for the changed repos
-      run: |
-        touch changed_tools.list
-        if [ -s changed_repositories.list ]; then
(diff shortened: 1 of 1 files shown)


nsoranzo:
> we should think about removing -exclude packages --exclude deprecated in favor of including these directories in .tt_skip (here and in the PR workflow). Advantage would be that the workflow would also run if those dirs do not exist in other repos.

That's already the case, planemo works fine if the directories passed explicitly with `--exclude` do not exist. I'd drop this point.

nsoranzo (approved):
I haven't tested the changes (or my follow-up commit), but they look good to me.

bernt-matthias:
> I haven't tested the changes (or my follow-up commit), but they look good to me.

Me neither. I guess we should. How is the easiest way? What happens if we call the slash command in this PR?

bernt-matthias:
Seems to run https://github.com/bernt-matthias/tools-iuc/runs/1687636152

nsoranzo:
> Seems to run https://github.com/bernt-matthias/tools-iuc/runs/1687636152

Yes, I also just looked at it and it looks fine. I can merge if you rebase out the spurious changes.

nsoranzo:
Thanks! I'll open a PR with file renaming to pr.yaml for consistency.

PR #3427, opened 2021-01-13
fix ruvseq

bernt-matthias:
CI failure was because the parameter name `designFormula` was changed to
`design_formula` in the function `get_deseq_dataset` which is imported
from `get_deseq_dataset.R` (which is a symlink into `../deseq2`).

don't think that we need to bump the version here. question is if the symlink is a good idea?

FOR CONTRIBUTOR:


bgruening:
> don't think that we need to bump the version here. question is if the symlink is a good idea?

If our tests catches this and the actually tool in the TS is not broken, I think that is ok.

We will have soon auto-bumps and should see those things earlier.

bgruening:
Can you please bump the tool version?

bgruening (approved):

PR #3428, opened 2021-01-14
Add Salsa2

astrovsky01:
FOR CONTRIBUTOR:


tools/salsa2/test-data/.python-version
```diff
@@ -0,0 +1 @@
+anaconda3-5.3.0
```
bgruening, line 1: that file should be removed

tools/salsa2/salsa2.xml
```diff
@@ -0,0 +1,56 @@
...
+            -g $gfa_file
+        #end if
+        #if $iter:
+            -i $iter
+        #end if
+        -o out
+    ]]></command>
+    <inputs>
+        <param name="fasta_in" type="data" format="fasta" label="Initial assembly file" help="Headers must not contain ':'."/>
+        <param name="length" type="data" format="fai,tsv,tabular" label="Length file" help="Fasta index file with lenght of contigs at start."/>
```
bgruening, line 25: I think we should use the created indices from the data tables: https://github.com/galaxyproject/tools-iuc/blob/master/tools/freec/tool_data_table_conf.xml.test

tools/salsa2/salsa2.xml
```diff
@@ -0,0 +1,56 @@
...
+        <param name="bed_file" type="data" format="bed" label="Bed alignment" help="Sorted by read names"/>
+        <param name="cutoff" type="integer" label="Cutoff"  optional="true" help="Minimum contig length to scaffold."/>
+        <param name="gfa_file" type="data" format="gfa1, gfa2" label="GFA file" optional="true"/>
+        <param name="enzyme" type="text" label="Restriction enzyme" help="Restriction enzyme sequence. If multiple were used, include all as a comma separated list without spaces (ex. 'GATC,AAGCTT')."/>
```
bgruening, line 29: "restriction enzym sequence" as a label.

tools/salsa2/salsa2.xml
```diff
@@ -0,0 +1,56 @@
...
+        #if $iter:
+            -i $iter
+        #end if
+        -o out
+    ]]></command>
+    <inputs>
+        <param name="fasta_in" type="data" format="fasta" label="Initial assembly file" help="Headers must not contain ':'."/>
+        <param name="length" type="data" format="fai,tsv,tabular" label="Length file" help="Fasta index file with lenght of contigs at start."/>
+        <param name="bed_file" type="data" format="bed" label="Bed alignment" help="Sorted by read names"/>
+        <param name="cutoff" type="integer" label="Cutoff"  optional="true" help="Minimum contig length to scaffold."/>
```
bgruening, line 27: can we add `argument=""` as well?

bgruening (commented):
Small review attached

tools/salsa2/salsa2.xml
astrovsky01: Sure, but that is used for the values inside, rather than the file itself which is how Salsa uses it:
https://github.com/galaxyproject/tools-iuc/blob/master/tools/freec/macros.xml#L11

I can run samtools faidx in the code instead?
https://github.com/galaxyproject/tools-iuc/blob/master/tools/freec/control_freec.xml#L13

tools/salsa2/salsa2.xml
```diff
@@ -0,0 +1,56 @@
+<tool id="salsa2" name="salsa2" version="0.1.0" python_template_version="2.7">
```
bernt-matthias, line 1: Please use a recent profile, at least 20.01

tools/salsa2/salsa2.xml
```diff
@@ -0,0 +1,56 @@
...
+    <requirements>
+        <requirement type="package" version="2.7">python</requirement>
```
bernt-matthias, line 3: python is not required. comes automatically with salsa2

tools/salsa2/salsa2.xml
```diff
@@ -0,0 +1,56 @@
...
+        <param name="fasta_in" type="data" format="fasta" label="Initial assembly file" help="Headers must not contain ':'."/>
+        <param name="length" type="data" format="fai,tsv,tabular" label="Length file" help="Fasta index file with lenght of contigs at start."/>
+        <param name="bed_file" type="data" format="bed" label="Bed alignment" help="Sorted by read names"/>
+        <param name="cutoff" argument="-c" type="integer" label="Cutoff"  optional="true" help="Minimum contig length to scaffold."/>
+        <param name="gfa_file" argument="-g" type="data" format="gfa1, gfa2" label="GFA file" optional="true"/>
+        <param name="enzyme" type="text" label="Restriction enzyme sequence(s)" help="Restriction enzyme sequence. If multiple were used, include all as a comma separated list without spaces (ex. 'GATC,AAGCTT')."/>
+        <param name="iter" argument="-i" type="integer" label="Iterations" optional="true" help="Number of iterations to run"/>
+    </inputs>
+    <outputs>
+        <data name="scaffolds_fasta" format="fasta" from_work_dir="out/scaffolds_FINAL.fasta" label="Salsa fasta assembly"/>
```
bernt-matthias, line 33: Please use `label="${tool.name} on ${on_string}: fasta assembly"`

tools/salsa2/salsa2.xml
astrovsky01: When I didn't include it, the tool ran in a 3.7 env, but it needs to run in 2.7

mvdbeek: This is weird, the conda package pins python 2.7. Does that also happen if you run the test with `--biocontainers` ?

bernt-matthias: > Does that also happen if you run the test with --biocontainers ?

And if not: do you have any conda environment enabled (base)?

tools/salsa2/salsa2.xml
mvdbeek: Maybe make this input optional and run samtools faidx if the input is not provided ?
```suggestion
        <param name="length" type="data" format="fai,tsv,tabular" optional="true" label="Length file" help="Fasta index file with length of contigs of the initial assembly. Will be generated if not provided"/>
```
Or just always do samtools faidx internally, both is fine.

tools/salsa2/salsa2.xml
astrovsky01: No idea why, and I changed nothing about my environment, but now it's working properly locally without the python package, so I removed it.

tools/salsa2/salsa2.xml
astrovsky01: Left the option just in case

bgruening:
Should we also include `-e DNASE`?

tools/salsa2/salsa2.xml
```diff
@@ -22,16 +28,15 @@
     ]]></command>
     <inputs>
         <param name="fasta_in" type="data" format="fasta" label="Initial assembly file" help="Headers must not contain ':'."/>
-        <param name="length" type="data" format="fai,tsv,tabular" label="Length file" help="Fasta index file with lenght of contigs at start."/>
-        <param name="bed_file" type="data" format="bed" label="Bed alignment" help="Sorted by read names"/>
```
bgruening: bed_file was removed here, any reason why, is it needed?

astrovsky01:
> Should we also include `-e DNASE`?

No, according to the history, it was added after the most recent release, so it not available in the tarball from the bioconda package

bgruening (approved):

bgruening:
Thanks @astrovsky01

Issue #2569, opened 2019-08-31
hisat2 missing options


bernt-matthias:
While thinking about https://github.com/galaxyproject/tools-iuc/issues/1740 I observed that some options are missing that could be handy, e.g. setting read groups (to fixed text or element identifier).

```
--no-unal | Suppress SAM records for reads that failed to align.
--no-hd | Suppress SAM header lines (starting with @).
--no-sq | Suppress @SQ SAM header lines.
--rg-id <text> | Set the read group ID to <text>. This causes the SAM @RG header line to be printed, with <text> as the value associated with the ID: tag. It also causes the RG:Z: extra field to be attached to each SAM output record, with value set to <text>.
--rg <text> | Add <text> (usually of the form TAG:VAL, e.g. SM:Pool1) as a field on the @RG header line. Note: in order for the @RG line to appear, --rg-id must also be specified. This is because the ID tag is required by the SAM Spec. Specify --rg multiple times to set multiple fields. See the SAM Spec for details about what fields are legal.
--omit-sec-seq | When printing secondary alignments, HISAT by default will write out the SEQ and QUAL strings. Specifying this option causes HISAT to print an asterix in those fields instead.
```

bgruening:
@astrovsky01 is looking at this one.

astrovsky01:
Since we run samtools immediately after, the no-hd and no-sq options create opportunities for errors where samtools doesn't recognize it as a properly formatted file, and returns errored bam to the history. Thoughts on either not including those options, or turning off the auto-samtools run afterwards if those flags are present?

bernt-matthias:
> Since we run samtools immediately after, the no-hd and no-sq options create opportunities for errors where samtools doesn't recognize it as a properly formatted file, and returns errored bam to the history. Thoughts on either not including those options, or turning off the auto-samtools run afterwards if those flags are present?

Unsure. Checked the samtools man pages a bit. Seems that `samtools view` does not print the header by default. So maybe its not a problem...

PR #3440, opened 2021-01-21
hisat 2 update: add sam output flags

astrovsky01:
FOR CONTRIBUTOR:

Adds advanced sam flag options


astrovsky01:
Based on https://github.com/galaxyproject/tools-iuc/issues/2569

tools/hisat2/hisat2.xml
```diff
@@ -534,6 +554,36 @@ hisat2
...
             </conditional>
 
+            <conditional name="sam_options">
+                <param name="sam_options_selector" type="select" label="SAM options">
+                    <option value="defaults">Use default values</option>
+                    <option value="advanced">Specify output options</option>
+                </param>
+                <when value="defaults" />
+                <when value="advanced">
+                    <param name="no_unal" argument="--no-unal" type="boolean" truevalue="true" falsevalue="false" checked="False" label="Suppress SAM records for reads that failed to align." help="Default: false"/>
```
bernt-matthias, line 564: Can you use lower case false?

bgruening:
```
Output output_alignments:  different than expected, difference (using diff):
( /tmp/tmpvcdwekf5hisat_output_5.bam v. /tmp/tmpo7kfq79fhisat_output_5.bam )
Binary data detected, not displaying diff
```

I haven't had time to look closely at it, but it seems to be a problem with the BAM file. BAM files are unpacked automatically by Galaxy and can diffe'd, so this should be fine. So I would assume that maybe the lines_diff needs to be increased?

astrovsky01:
Even at 40 lines different (more than twice the length of the whole converted sam file...), it's failing. When I run a samtools view + diff on the test and the output files, they only have one line different

bernt-matthias:
Can you convert the bam file created during planemo test and the bam file in test-data, convert them manually to sam and compare the sam files? Maybe this sheds some light on the differences..

bernt-matthias:
I did a bit more research on this. When I call my system installed version of `samtools view ` I get an error 

```
[E::sam_hrecs_error] Malformed key:value pair at line 3: "@RG	ID:read_group	TAG:test1	TAG:test2"
samtools view: failed to add PG line to the header
```

but if I use the `samtools` from the conda env it works. So I guess that older versions of `samtools` (and htslib?) might have some problems that have been fixed in later versions. 

Galaxy uses pysam to convert the bam to sam for calculating the diff. In the output of `planemo test` one can find: 

```
2021-01-23 13:53:45,189 WARNI [galaxy.tool_util.verify] Converting local (test-data) BAM to SAM failed: 'samtools returned with error 1: stdout=, stderr=samtools view: failed to add PG line to the header\n'. Will compare BAM files
```

So my guess is that we need a newer version of pysam for Galaxy...?

bernt-matthias:
Think I found it, actually the read group data was wrongly formatted since only certain tags are allowed, see https://samtools.github.io/hts-specs/SAMv1.pdf page 4.

I would suggest to implement a validator to check for correct tags, or to add a select.

bgruening:
Looks good to me. Hisat2 2.2.1 is out? Do we want to give it a try?

astrovsky01:
Let's do that as a different PR -- this was just supposed to be the papercut

bernt-matthias:
> I would suggest to implement a validator to check for correct tags, or to add a select.

Do you think that this would fit in the paper cut. At least we should add a short note in the help.

astrovsky01:
Let's add that next so these flags can be in asap -- it hurts nothing to do these as separate PRs

bernt-matthias (approved):

PR #3450, opened 2021-01-29
Add --include filter to bcftools consensus

pvanheus:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/bcftools/macros.xml
+++ b/tools/bcftools/macros.xml
@@ -436,10 +436,34 @@ $vcfs_list_file
 #end if
   </token>
 
+  <xml name="macro_include_exclude_validate_sanitize">
+    <validator type="expression" message="Single quote or trailing backslash not allowed">"'" not in value and value[-1] != "\\"</validator>
+    <sanitizer>
+      <valid initial="string.ascii_letters,string.digits,string.whitespace">
+        <add value="." />
+        <add value='"' />
+        <add value="+" />
+        <add value="-" />
+        <add value="*" />
+        <add value="/" />
+        <add value="&lt;" />
+        <add value="&gt;" />
+        <add value="=" />
+        <add value="!" />
+        <add value="~" />
+        <add value="(" />
+        <add value=")" />
+        <add value="[" />
+        <add value="]" />
+        <add value="&amp;" />
+        <add value="|" />
+      </valid>
+    </sanitizer>
+  </xml>
 
   <xml name="macro_include">
     <param argument="--include" type="text" optional="true" label="Include" help="Select sites for which the expression is true">
-        <sanitizer sanitize="False"/>
+      <expand macro="macro_include_exclude_validate_sanitize" />
     </param>
   </xml>
   <token name="@INCLUDE@">
@@ -447,11 +471,10 @@ $vcfs_list_file
   --include '${section.include}'
 #end if
   </token>
-
+  
   <xml name="macro_exclude">
     <param argument="--exclude" type="text" optional="true" label="Exclude" help="Exclude sites for which the expression is true">
-        <validator type="expression" message="Single quote or trailing backslash not allowed">"'" not in value and value[-1] != "\\"</validator>
-        <sanitizer sanitize="False"/>
+      <expand macro="macro_include_exclude_validate_sanitize" />
     </param>
   </xml>
   <token name="@EXCLUDE@">
--- a/tools/bcftools/macros.xml
+++ b/tools/bcftools/macros.xml
@@ -439,7 +439,6 @@ $vcfs_list_file
 
   <xml name="macro_include">
     <param argument="--include" type="text" optional="true" label="Include" help="Select sites for which the expression is true">
-        <validator type="regex" message="Single quote not allowed">^[^']*$</validator>
         <sanitizer sanitize="False"/>
     </param>
   </xml>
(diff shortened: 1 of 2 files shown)


pvanheus:
The bcftools consensus subcommand allows for `--include` which allows an expression to be provided to filter variants applied when consensus is constructed. So for example `--include TYPE="snp"` includes only SNPs. The filters are useful if a consensus is going to be used as part of a multiple sequence alignment and then phylogeny.

bernt-matthias (changes requested):
I guess the tool version should be bumped.

pvanheus:
> I guess the tool version should be bumped.

good point. I've added a `+galaxy0` to the tool version because it was not there before, and this is a change on the side of the Galaxy wrapper, not on the side of the tool itself.

wm75:
@pvanheus `+galaxy0` and a missing `+galaxy` string are considered equivalent so IMO you should add `+galaxy1` here.

Although it's so far not part of your PR I would appreciate it if you could improve the `validator` at
https://github.com/galaxyproject/tools-iuc/blob/65bc2116f5e588ef91d8818546a040ecbdc73448/tools/bcftools/macros.xml#L442
a little bit.
I think `<validator type="expression" message="Single quote or trailing backslash not allowed">"'" not in value and value[-1] != "\\"</validator>` would be both easier to read and would cover an attempt to escape the closing single quote of the wrapper (not a safety issue as far as I can see, but would lead to a weird error message).
Same goes for the *exclude* macro at https://github.com/galaxyproject/tools-iuc/blob/65bc2116f5e588ef91d8818546a040ecbdc73448/tools/bcftools/macros.xml#L454

Thanks a lot for the extra effort!

wm75:
Also, why don't you enable the `--exclude` functionality, too, here?

pvanheus:
> Also, why don't you enable the `--exclude` functionality, too, here?

Because `bcftools consensus` doesn't seem to have an [--exclude flag](http://samtools.github.io/bcftools/bcftools.html#consensus)

pvanheus:
@wm75 in terms of the validator, I have added what you suggest, but I wonder if it can get even stricter - from [the documentation](http://samtools.github.io/bcftools/bcftools.html#expressions) it looks like a whitelist might be: `[A-Za-z0-9.+*-/=&gt;&lt;<!~()\[\]&amp;|"]` (not sure if & escapes are necessary here). Presumably this could be done using a `<sanitizer>` ?

I've added the `--exclude`. I didn't spot it at first.

pvanheus:
thinking more on the sanitizer, I think:

```
<sanitizer>
  <valid initial="string.ascii_letters,string.digits,string,string.whitespace">
    <add value="." />
    <add value="\"" />
    <add value="+" />
    <add value="-" />
    <add value="*" />
    <add value="/" />
    <add value="&lt;" />
    <add value="&gt;" />
    <add value="=" />
    <add value="!" />
    <add value="~" />
    <add value="(" />
    <add value=")" />
    <add value="[" />
    <add value="]" />
    <add value="&" />
    <add value="|" />
  <valid>
</sanitizer>
```
would work. I know its verbose, but does it seem reasonable?

(btw reading the [code](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/parameters/sanitize.py) suggests that `add` actually ads a substring to check against, not a character.

tools/bcftools/macros.xml
```diff
@@ -436,23 +436,45 @@ $vcfs_list_file
...
+        <add value="&gt;" />
+        <add value="=" />
+        <add value="!" />
+        <add value="~" />
+        <add value="(" />
+        <add value=")" />
+        <add value="[" />
+        <add value="]" />
+        <add value="&amp;" />
+        <add value="|" />
```
bernt-matthias, line 459: I'm not sure but I guess `_`, `,`, `:` also do not create any harm. Maybe its easier to initialize with `string.printable` and remove what we certainly do not want.

file references would be fancy `@samples.txt` .. but no idea how to implement this in a tool :)

pvanheus: I think we should not have file references. The others I left out because they were not explicitly included.

I thought about including punctuation and removing from there, but was a little worried that we would not know what was in punctuation over time. Definitely `'` can be excluded.

bernt-matthias: `_` for instance can be found in the docs https://samtools.github.io/bcftools/bcftools.html#expressions

wm75:
Sorry for not finding the time to reply earlier.
I think a `validator` is in fact better than a `sanitizer` for this particular purpose. A `sanitizer` would *modify* the user input before passing it on, but that will hardly do any good because it will not result in a valid expression (or if so not in the expression the user intended).
A validator that will refuse to submit the job is more helpful.

wm75:
You can still play with what to allow in that validator. My minimal suggestion can cerainly be improved - just didn't want to make things more complicated than necessary.

pvanheus:
Thinking through the combination of validator and sanitizer: the validator is there because the sanitizer is disabled. The validator catches attacks that would otherwise be caught by the default behaviour, which sanitizes input.

My aim with putting an allow list in place was to shift from "forbid known dangerous input" towards "only allow safe input". Obviously the range of what is safe is wider than what would normally be allowed, thus the need for custom rules.

With @bernt-matthias's ideas in mind, I have switched to using a set of punctuation that only excludes known-bad characters: `'` and `@` (this because file inputs should not be allowed). Perhaps `\` should also be excluded but I am not sure. While this is not strictly necessary (if we assume that the validator rules are good enough), having both in place doesn't particularly hurt and offers a little bit more security.

bernt-matthias (approved):

PR #3469, opened 2021-02-09
Add wgscoverageplotter from jvarkit

pvanheus:
FOR CONTRIBUTOR:


mvdbeek:
Can you add a `.shed.yaml` file @pvanheus ?

wm75:
@pvanheus do you really need picard here just to generate the sequence dict or would something like `samtools view -H` | `grep` work, too?

pvanheus:
> @pvanheus do you really need picard here just to generate the sequence dict or would something like `samtools view -H` | `grep` work, too?
You mean something like `samtools -H '$alignment_file' | grep -E '^(@HD|@SQ)' >reference.dict` ? I.e. extract the dictionary from the BAM header, not the FASTA? Does that do the same as `picard` ?

tools/jvarkit_wgscoverageplotter/jvarkit_wgscoverageplotter.xml
```diff
@@ -0,0 +1,236 @@
...
+        samtools view -H '$alignment_file' |grep -E '^(@HD|@SQ)' > reference.dict &&
+        ln -s '$alignment_file' '$alignment_name' &&
+        ln -s '${alignment_file.metadata.bam_index}' '$alignment_name'.bai &&
+        wgscoverageplotter.sh 
+        --reference reference.fasta
+        --dimension '$dimension'
+        --output plot.svg
+        #if str($adv.adv_choice) == "true"
+            $adv.disable_paired_overlap
+            #if len($include_regex) != 0
```
wm75, line 35:
```suggestion
            #if $include_regex
```

tools/jvarkit_wgscoverageplotter/jvarkit_wgscoverageplotter.xml
```diff
@@ -0,0 +1,236 @@
...
+                -I '$include_regex'
+            #end if
+            #if len($exclude_regex) != 0
```
wm75, line 38:
```suggestion
            #if $exclude_regex
```

tools/jvarkit_wgscoverageplotter/jvarkit_wgscoverageplotter.xml
```diff
@@ -0,0 +1,236 @@
...
+                -X '$exclude_regex'
+            #end if
+            --mapq $adv.min_mapq
+            --max-depth $adv.max_depth
+            $adv.clip
+            --min-contig-length $adv.min_contig_length
+            --percentile $adv.percentile
+            $adv.points
+            #if len(str($adv.samples).strip()) != 0
```
wm75, line 47:
```suggestion
            #if str($adv.samples).strip()
```

wm75: Actually, this one will turn into a validator on the param if you pick up my suggestion of a conditional block around --samples and --partition.

tools/jvarkit_wgscoverageplotter/jvarkit_wgscoverageplotter.xml
```diff
@@ -0,0 +1,236 @@
...
+            </when>
+        </conditional>
+        <param name="alignment_file" type="data" format="bam" label="BAM alignment" />
+        <param argument="--dimension" value="1000x500" type="text" label="Image dimensions" />
+        <param name="format" type="select" label="Output format" help="Note that SVG might not display correctly on all Galaxy servers">
+            <option value="PNG" selected="true">PNG</option>
+            <option value="SVG">SVG</option>
+        </param>
+        <conditional name="adv">
+            <param type="select" name="adv_choice" label="Show advanced options">
```
wm75, line 90: I guess a `<section>` is the modern style here instead of a `conditional`.

tools/jvarkit_wgscoverageplotter/jvarkit_wgscoverageplotter.xml
```diff
@@ -0,0 +1,236 @@
...
+                />
+                <param argument="--skip-contig-regex"
+                    name="skip_contig_regex"
+                    type="text"
+                    label="Skip chromosomes matching this regular expression" />
+                <param argument="--samples"
+                    type="text"
+                    label="Limit to these read groups" 
+                    help="See also --partition. Multiple read groups can be separated with commas." />
+                <param argument="--partition" type="select" label="Partition data using SAM Read Groups (requires --samples)">
```
wm75, line 138: Since this option requires `--samples` I would move things into a conditional that asks whether to limit to certain read groups -> only <when "yes"> -> bring up --samples and --partition

wm75 (commented):
Looks pretty good already. Just some minor improvement suggestions.

bernt-matthias:
What do you think about renaming the folder to jvarkit.. Just in case someone wants to add more tools from this suite someday? Maybe also use a suite style shed.yml..

tools/jvarkit_wgscoverageplotter/jvarkit_wgscoverageplotter.xml
pvanheus: I have been trying to figure out how `--samples` and `--partition` work, but failed, so I have opened an [issue](https://github.com/lindenb/jvarkit/issues/177) on the upstream project and have disabled these options in the wrapper for now.

wm75: Seems like a good decision!

pvanheus: I got a speedy reply from Pierre Lindenbaum clarifying the use of `--samples` and `--partition` and have added these options back in the wrapper as per your suggestion @wm75.

pvanheus:
> What do you think about renaming the folder to jvarkit.. Just in case someone wants to add more tools from this suite someday? Maybe also use a suite style shed.yml..

Good idea, I've switched to a suite style `shed.yml` and renamed the directory to `jvarkit`.

tools/jvarkit/jvarkit_wgscoverageplotter.xml
```diff
@@ -0,0 +1,227 @@
...
+        #if $exclude_regex
+            -X '$exclude_regex'
+        #end if
+        --mapq $adv.min_mapq
+        --max-depth $adv.max_depth
+        $adv.clip
+        --min-contig-length $adv.min_contig_length
+        --percentile $adv.percentile
+        $adv.points
+        #if str($adv.sample_filter.filter_by_sample) == "true" and str($adv.sample_filter.samples).strip()
```
wm75, line 45: I wouldn't do the check for an empty --samples param here, but rather have a `<validator>` on the param itself.

If a user wants filtering by sample, but doesn't enter a sample identifier, then either that's a typo or they didn't understand what they are doing. In either case this shouldn't be passed over silently, but should be caught and brought to the user's attention as early as possible.

tools/jvarkit/jvarkit_wgscoverageplotter.xml
```diff
@@ -0,0 +1,227 @@
...
+                name="skip_contig_regex"
+                type="text"
+                label="Skip chromosomes matching this regular expression" />
+            <conditional name="sample_filter">
+                <param name="filter_by_sample" type="select" label="Filter by sample">
+                    <option value="true">Yes</option>
+                    <option value="false" selected="true">No</option>
+                </param>
+                <when value="true">
+                    <param argument="--samples"
```
wm75, line 132: This should be validated and not allowed to be empty after stripping whitespace.

tools/jvarkit/jvarkit_wgscoverageplotter.xml
```diff
@@ -0,0 +1,227 @@
...
+                type="boolean"
+                label="Don't show coverage to be greater than 'max-depth' in the output image"
+                truevalue="--clip"
+                falsevalue="" />
+            <param argument="--min-contig-length" 
+                name="min_contig_length" 
+                type="integer" 
+                label="Skip chromosomes shorter than"
+                value="0"
+                help="A distance specified as a positive integer. Commas are removed. The following suffixes are interpreted: b,bp,k,kb,m,mb" />
```
wm75, line 109:
```suggestion
                min="0"
                help="A distance specified as a positive integer." />
```

More is not possible with an integer param.

tools/jvarkit/jvarkit_wgscoverageplotter.xml
```diff
@@ -0,0 +1,227 @@
...
+                <param name="filter_by_sample" type="select" label="Filter by sample">
+                    <option value="true">Yes</option>
+                    <option value="false" selected="true">No</option>
+                </param>
+                <when value="true">
+                    <param argument="--samples"
+                        type="text"
+                        label="Limit to these read groups" 
+                        help="See also --partition. Multiple read groups can be separated with commas." />
+                    <param argument="--partition" type="select" label="How to interpret identifiers in --samples" help="Partitions are modelled on the GATK login as described here: https://gatk.broadinstitute.org/hc/en-us/articles/360051307491-DepthOfCoverage-BETA-#--partition-type">
```
wm75, line 136:
```suggestion
                    <param argument="--partition" type="select" label="How to interpret identifiers in --samples" help="Partitions are modelled on the GATK logic as described here: https://gatk.broadinstitute.org/hc/en-us/articles/360051307491-DepthOfCoverage-BETA-#--partition-type">
```

tools/jvarkit/jvarkit_wgscoverageplotter.xml
```diff
@@ -0,0 +1,227 @@
...
+            <param argument="--disable-paired-overlap" 
+                name="disable_paired_overlap" 
+                type="boolean" truevalue="--disable-paired-overlap" 
+                falsevalue="" label="Count overlapping bases with mate for paired-end" />
+            <param argument="--include-contig-regex" name="include_contig_regex" type="text" label="Only keep chromosomes matching this regular expression." />
+            <param argument="--mapq" name="min_mapq" type="integer" value="1" label="Minimum mapping quality" />
+            <param argument="--max-depth" 
+                name="max_depth" 
+                type="integer" 
+                label="Max depth to display."
```
wm75, line 96: I know that's the help of the command line tool, but maybe 'y-axis (depth) limit of the plot' or similar would be clearer?

tools/jvarkit/jvarkit_wgscoverageplotter.xml
```diff
@@ -0,0 +1,227 @@
...
+            <param argument="--mapq" name="min_mapq" type="integer" value="1" label="Minimum mapping quality" />
+            <param argument="--max-depth" 
+                name="max_depth" 
+                type="integer" 
+                label="Max depth to display." 
+                value="-1"
+                help="The special value '-1' will first compute the average depth and the set the max depth to 2*average" />
+            <param argument="--clip"
+                type="boolean"
+                label="Don't show coverage to be greater than 'max-depth' in the output image"
```
wm75, line 101: Similar to the above: maybe sth like label="Clip the plot at the y-axis limit" help="Default is to let the coverage plot exceed the y-axis limit"?

This would also avoid the double-negation in the default situation, i.e.:
*Don't show coverage to be greater than 'max-depth' in the output image*: `No`
which always makes my head spin :)

tools/jvarkit/.shed.yml
```diff
@@ -0,0 +1,17 @@
...
+  Jvarkit,  a set of more than java-based tools for bioinformatics written by Pierre Lindenbaum 
+name: jvarkit
+owner: iuc
+remote_repository_url: https://github.com/galaxyproject/iuc/tree/master/tools/jvarkit
+type: unrestricted
+auto_tool_repositories:
+  name_template: "{{ tool_id }}"
+  description_template: "Wrapper for {{ tool_name }}"
+suite:
+  name: "suite_varscan"
```
bgruening, line 15:
```suggestion
  name: "suite_jvarkit"
```

tools/jvarkit/jvarkit_wgscoverageplotter.xml
```diff
@@ -0,0 +1,232 @@
...
+        $adv.points
+        #if str($adv.sample_filter.filter_by_sample) == "true"
+            --samples '$adv.sample_filter.samples'
+            --partition '$adv.sample_filter.partition'
+        #end if
+        '$alignment_name' &&
+        #if str($format) == "SVG"
+            mv plot.svg '$plot_output'
+        #else
+            convert plot.svg '$format:$plot_output'
```
bgruening, line 53: @pvanheus do we need this complexity inside this tool? We have an ImageMagick tool that can do this in a separate step. This is not a -1 just a question.

tools/jvarkit/jvarkit_wgscoverageplotter.xml
```diff
@@ -0,0 +1,232 @@
...
+        #else
+            #set ref_path = str($reference.cached_genome.fields.path)
+            #set ref_index_path = str($reference.cached_genome.fields.path) + '.fai'
+            ln -s '$ref_path' reference.fasta &&
+            ln -s '$ref_index_path' reference.fasta.fai &&
+        #end if
+        samtools view -H '$alignment_file' |grep -E '^(@HD|@SQ)' >reference.dict &&
+        cat reference.dict &&
+        ln -s '$alignment_file' '$alignment_name' &&
+        ln -s '${alignment_file.metadata.bam_index}' '$alignment_name'.bai &&
```
bgruening, line 27:
```suggestion
        ln -s '${alignment_file.metadata.bam_index}' '$alignment_name.bai' &&
```

bgruening (approved):

tools/jvarkit/jvarkit_wgscoverageplotter.xml
pvanheus: If we leave the PNG support out here then, for Galaxy servers without SVG display support (I think a lot of them at this point?) it means always having to run 2 tools to get an image. So the cost of simplifying this tool comes with a cost of making its use more complex, and thus I'd prefer to leave this in.

wm75 (approved):

wm75:
Thanks a lot @pvanheus for this very useful wrapper!

PR #3492, opened 2021-02-21
[WIP] Add stress-ng

mtekman:
FOR CONTRIBUTOR:

This adds [stress-ng](https://github.com/ColinIanKing/stress-ng), a tool for stress testing a system. Most of the inputs were autogenerated with macros.

Currently the conda recipe is being developed, which also needs the dependency `libbsd` to run

TODO:
* Conda Recipe
    * [X] stress-ng - https://github.com/conda-forge/staged-recipes/pull/14034
    * [X] ~~libbsd - https://github.com/conda-forge/staged-recipes/pull/14035~~ not needed
* Galaxy wrapper
   * [X] Inputs
   * [X] Cheetah
   * [x] Outputs
   * [X] Tests
   * [x] Tests passing
   * [x] Help
* Decide whether all inputs are necessary
   * [X] just expose a good subset of parameters


bgruening:
conda package is ready. @mtekman I don't think there is **no** need for all parameters. Only a few memory parameters should be enough.

mvdbeek:
Do we really want this in the IUC ? I tend to think no, since that means we need to add exceptions to not install this on public servers. If we decide to merge this we should probably upload this under a different owner (iuc-dev or similar).

bgruening:
fair point, we can move this to our repo ... it more for maintenance tasks.

mvdbeek:
I think we can maintain it here, but upload under a different name ? I think that's a good compomise

mtekman:
> conda package is ready. @mtekman I don't think there is **no** need for all parameters. Only a few memory parameters should be enough.

Yeah, I planned to have a "Common Options" conditional and a "Advanced Options" (i.e. everything) dropdown. Many of the advanced options will likely not have tests, but given that this is more for admins, user experience seemed optional.

I can base the common options off the examples, unless you want more a few more?

> I think we can maintain it here, but upload under a different name ? I think that's a good compomise

if it's private, should we go for `_stress-ng`?

mtekman:
@bgruening I've disabled most of the options, but they are there and can be uncommented if we later wish to incorporate more features. For now I expose only the most commonly used ones.

One thing to note is that the cpu usage cannot be disabled by default, but you can reduce the load to something negligible like `--cpu-load 1%`

I *do* have a shed file, but I have not added it to the commits because I do not wish to trigger remote tests just now. Locally, the tests make my laptop run super slow, so I don't know how that will scale on a remote. 

So I will commit the shed file tomorrow morning when there are more admins around who can watch it/kill it.

bgruening:
@mtekman I guess this is ok, just commit it. But can you please create this PR against https://github.com/bgruening/galaxytools

mtekman:
sure, but  should I complete the shed yml here first to start the tests and then create a PR against your repo?

 or should I close this immediately and resume committing in a PR against your repo?

bgruening:
move over to the other repo, the testing is everywhere the same

mtekman:
oh right, I see

PR #2534, opened 2019-08-16
Add: Strelka2

stephanflemming:
Initial pull request for strelka wrapper

PR #3507, opened 2021-03-01
fix test of strelka germline

bernt-matthias:
xref galaxyproject/galaxy#11313

FOR CONTRIBUTOR:

### DIFF
--- a/tools/strelka/strelka_germline.xml
+++ b/tools/strelka/strelka_germline.xml
@@ -179,7 +179,7 @@ minMapq = $strelka.minMapq
             </section>
             <output name="out_variants" ftype="vcf_bgzip" file="variants_test2.vcf" decompress="true" compare="diff" lines_diff="8" />
         </test>
-        <!-- #3; input bam, no defaults -->
+        <!-- #3; input bam, no defaults (note without s_e_e (genome) output seem not deterministic -->
         <test expect_num_outputs="2">
             <param name="bam" value="sample1.cram,sample2.cram"/>
             <conditional name="ref_cond">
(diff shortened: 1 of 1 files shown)


bernt-matthias:
Seems that the content of `out_genome` is non-deterministic. Does one of the contributors know if this is expected? Ping @bgruening @stephanflemming @wm75

wm75:
Does using
```
<section name="expert_settings">
    <param name="s_e_e" value="--disableSequenceErrorEstimation" />
</section>
```

in that test help? That's the trick I used in other tests to get reproducible results, but apparently I forgot to add it to this one (cause it never triggered an error before you fixed the bug with collection testing).

bernt-matthias:
@wm75 unfortunately this does not help. reason seems to be that the assertions work on the compressed output. https://github.com/galaxyproject/galaxy/issues/11521

wm75:
Ah, of course! That's why here: https://github.com/galaxyproject/tools-iuc/pull/2534/commits/d1dec9494a64d73ff264b5a387a35fd3e996bf9b
I changed the first two tests to use *diff* and the `--disableSequenceErrorEstimation` option.
Maybe the later wouldn't even have been necessary - don't remember anymore.
Let me check whether comparing to any of the existing test datasets would work.

bernt-matthias:
I disabled the assertion for now. I guess for the tool its sufficient to know that there are two elements in the collection. 

Feel free to extend the test or merge.

wm75:
Was about to suggest the same. Seems totally fine for now.

wm75 (approved):

PR #3510, opened 2021-03-02
stacks2: fix tests, output format and filter, and bump

bernt-matthias:
xref https://github.com/galaxyproject/galaxy/pull/11313
fixes https://github.com/galaxyproject/tools-iuc/issues/3526

also improving some tests by using `diff` instead of `sim_size`

FOR CONTRIBUTOR:


bgruening (commented):
There seem to be empty files in this PR and the files ending with gz are visible? I suspect they are not really compressed?

bernt-matthias:
> There seem to be empty files in this PR and the files ending with gz are visible?

Output type was wrong (should have been fastq). But they are empty -- could be replaced with an assertion if you like. 

Wondering why the tool linting step already worked. I expected it to pass only with https://github.com/galaxyproject/galaxy/pull/11536 .. and I do not understand how this change affects planemo that fast. 

But maybe it just fine and I should be happy :)

bgruening (approved):

bgruening:
> Wondering why the tool linting step already worked. I expected it to pass only with galaxyproject/galaxy#11536 .. and I do not understand how this change affects planemo that fast.

That is indeed strange. Does the lint fail locally?
Feel free to merge, that seems to be an unrelated problem.

bernt-matthias:
Problem seems to be that the repo_list file has no new line at the end: https://github.com/galaxyproject/planemo-ci-action/blob/a7fb2018f8283c28232f794c80f4c1e38c3c5a0a/planemo_ci_actions.sh#L96

I remember that I did this on purpose .. but I can not remember. 

Ultimately this leads to 

```
while read DIR
do
...
done
```
skipping the last line. The only thing I would so far is to check if $DIR is not empty after the loop and execute the loop body again if necessary. 

Or we remove the `-n` from the echo.


maybe @mvdbeek or @nsoranzo have an idea?

bernt-matthias:
tata .. looks good for me

bgruening (approved):

PR #3519, opened 2021-03-06
add pangolin

jpalmer37:
FOR CONTRIBUTOR:
* [ x ] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [ x ] - License permits unrestricted use (educational + commercial)
* [ x ] - This PR adds a new tool or tool collection


tools/pangolin/pangolin.xml
```diff
@@ -0,0 +1,68 @@
...
+    ]]></command>
+    <inputs>
+      <param type="data" name="input1" format="fasta" label="Input FASTA File(s)" /> 
+      <param name="alignment" type="boolean" label="Generate output alignment" 
+          truevalue="--alignment" falsevalue="" help="(--alignment)" /> 
+      <param name="decompress" type="boolean" label="Decompress model for greater speed" 
+          truevalue="--decompress-model" falsevalue="" help="(--decompress-model)" /> 
+      <param name="max_ambig" type="float" label="Maximum proportion of Ns allowed" 
+          value="0.5" min="0" max="1" help="(--max-ambig)" />
+      <param name="min_length" type="integer" label="Minimum query length allowed"
```
bgruening, line 22: You could actually use `argument="--min-length"` this way you can remove `name` and the brakets in help as well.

tools/pangolin/pangolin.xml
```diff
@@ -0,0 +1,68 @@
...
+          truevalue="--alignment" falsevalue="" help="(--alignment)" /> 
+      <param name="decompress" type="boolean" label="Decompress model for greater speed" 
+          truevalue="--decompress-model" falsevalue="" help="(--decompress-model)" /> 
+      <param name="max_ambig" type="float" label="Maximum proportion of Ns allowed" 
+          value="0.5" min="0" max="1" help="(--max-ambig)" />
+      <param name="min_length" type="integer" label="Minimum query length allowed" 
+          value="10000" min="0" help="(--min-length)" />
+    </inputs>
+    <outputs>
+      <data name="output1" format="csv" />
```
bgruening, line 26: Can we output `tabular` please. We can convert csv to tabular on the fly in this tool. Galaxy has many tools to work with tabular and that makes it easier for the user.

tools/pangolin/pangolin.xml
```diff
@@ -0,0 +1,68 @@
...
+          <output name="output1" file="result2.csv"/>
+      </test>
+
+    </tests>
+    <help><![CDATA[
+        usage: pangolin <query> [options]
+
+        pangolin: Phylogenetic Assignment of Named Global Outbreak LINeages
+
+        positional arguments:
```
bgruening, line 45: In a tool help we should describe the tool but not the parameters, those are described above already.

bgruening (commented):
@jpalmer37 thanks a lot.

To trigger a test, please include a .shed.yml file.

tools/pangolin/pangolin.xml
```diff
@@ -0,0 +1,68 @@
+<tool id="pangolin" name="Phylogenetic Assignment of Outbreak Lineages (Pangolin)" version="0.1.0" python_template_version="3.5">
```
bernt-matthias, line 1: Can you make the tool version equal to the pangolin version?

Please remove the python template version and add profile.. E. G. 20.01

tools/pangolin/pangolin.xml
```diff
@@ -0,0 +1,68 @@
...
+        --outfile '$output1' 
+        --max-ambig $max_ambig
+        --min-length $min_length
+        '$input1' 
+    ]]></command>
+    <inputs>
+      <param type="data" name="input1" format="fasta" label="Input FASTA File(s)" /> 
+      <param name="alignment" type="boolean" label="Generate output alignment" 
+          truevalue="--alignment" falsevalue="" help="(--alignment)" /> 
+      <param name="decompress" type="boolean" label="Decompress model for greater speed"
```
bernt-matthias, line 18: Is there a downside of enabling this feature? If yes, then it might be good to document it. If not we could just hardcode it.

tools/pangolin/pangolin.xml
```diff
@@ -0,0 +1,68 @@
...
+      <param name="decompress" type="boolean" label="Decompress model for greater speed" 
+          truevalue="--decompress-model" falsevalue="" help="(--decompress-model)" /> 
+      <param name="max_ambig" type="float" label="Maximum proportion of Ns allowed" 
+          value="0.5" min="0" max="1" help="(--max-ambig)" />
+      <param name="min_length" type="integer" label="Minimum query length allowed" 
+          value="10000" min="0" help="(--min-length)" />
+    </inputs>
+    <outputs>
+      <data name="output1" format="csv" />
+      <data name="sequences.aln.fasta" format="fasta" />
```
bernt-matthias, line 27: This seems to be unused in the command block. Also dots in the variable name are not a good idea.

tools/pangolin/pangolin.xml
```diff
@@ -0,0 +1,68 @@
...
+    </outputs>
+    <tests>
```
bernt-matthias, line 29: tools/pangolin/test-data/aln1.fasta seems unused in the test

tools/pangolin/pangolin.xml
```diff
@@ -0,0 +1,68 @@
...
+          --alignment           Optional alignment output.
+          --decompress-model    Permanently decompress the model file to save time
+                                running pangolin.
+          --max-ambig MAXAMBIG  Maximum proportion of Ns allowed for pangolin to
+                                attempt assignment. Default: 0.5
+          --min-length MINLEN   Minimum query length allowed for pangolin to attempt
+                                assignment. Default: 10000
+    ]]></help>
+    <citations>
+      <citation type="bibtex">
```
bernt-matthias, line 58: Is it possible to cite with a doi

jpalmer37: According to their repository, they have a paper in progress for Pangolin; they said to just cite their Github repository in the meantime. Thoughts on this?

bgruening: If there is no paper, then this is fine, but please replace the TODO :)

tools/pangolin/pangolin.xml
jpalmer37: Sounds good! Addressed this in 79b9cc272ad64b66f655fd252451eb878faa1de2

tools/pangolin/pangolin.xml
jpalmer37: Sounds good. Thanks for the tip! Addressed in 79b9cc272ad64b66f655fd252451eb878faa1de2

tools/pangolin/pangolin.xml
jpalmer37: For sure. Do you know if this method of version encoding would work here? 
```xml
<tool id="pangolin" name="Phylogenetic Assignment of Outbreak Lineages (Pangolin)" version="@TOOL_VERSION@" profile="20.01" >
```

tools/pangolin/pangolin.xml
jpalmer37: Did some more digging, but I'm still not really sure. This feature is only mentioned on the package's help menu, not on their repository. 

I tested it locally. When the flag is on, it doesn't produce normal output. It appears to save the model to a local folder for faster processing on the next run. Do you think it's something we should potentially skip for now?

tools/pangolin/pangolin.xml
jpalmer37: Totally missed that haha! Replaced author with the repository owner here 79b9cc2

tools/pangolin/pangolin.xml
jpalmer37: Sounds good. I fixed the variable name. I realize I made this output because I was trying to include an additional test.  I explain a bit more under your other comment: 
> This seems to be unused in the command block. Also dots in the variable name are not a good idea.

tools/pangolin/pangolin.xml
jpalmer37: I accidentally left this in because I was trying to test the alignment output. By using the `--alignment` flag, a FASTA alignment with the predefined output name `sequences.aln.fasta` is generated. It's name is fixed and can't be specified like the main output file. `aln1.fasta` is a test output for this case, but I'm not quite sure how to test this secondary output. 

Below is my attempt to test this (found here 79b9cc2) but unfortunately the test (test #3) fails. Would you have any tips on how to fix this?

```xml
<test>
    <param name="alignment" value="--alignment"/>
    <param name="input1" value="test1.fasta"/>
    <output name="output1" file="result1.csv"/>
    <output name="align1" file="aln1.fasta" ftype="fasta"/>
</test>
```

bgruening: @jpalmer37 you can add this to your `outputs`:

```xml
    <data name="alignment" format="fasta" from_work_fir="sequences.aln.fasta" />
```

tools/pangolin/pangolin.xml
bgruening: Yes, this looks mostly good. I would keep the name short and add a `<description>`.
For the version, you might consider following those rules: https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tool-versions

@simonbray is currently working on a fancy bot that could update this tool automatically.

tools/pangolin/pangolin.xml
bgruening: Yes, let's remove it.

tools/pangolin/pangolin.xml
```diff
@@ -0,0 +1,61 @@
...
+            truevalue="--alignment" falsevalue="" help="(--alignment)" /> 
+      <param name="decompress" type="boolean" label="Decompress model for greater speed" 
+          truevalue="--decompress-model" falsevalue="" help="(--decompress-model)" /> 
+      <param argument="--max-ambig" type="float" label="Maximum proportion of Ns allowed" 
+          value="0.5" min="0" max="1" help="--max-ambig" />
+      <param argument="--min-length" type="integer" label="Minimum query length allowed" 
+          value="10000" min="0" help="--min-length"/>
+    </inputs>
+    <outputs>
+      <data name="output1" format="tabular" />
```
bgruening, line 26: You have missed adding the output here which you are using in the test.

tools/pangolin/pangolin.xml
bgruening: ping

tools/pangolin/pangolin.xml
```diff
@@ -0,0 +1,61 @@
+<tool id="pangolin" name="Phylogenetic Assignment of Outbreak Lineages (Pangolin)" version="@TOOL_VERSION@" profile="20.01" >
+    <requirements>
+        <requirement type="package" version="2.3.2">pangolin</requirement>
```
bgruening, line 3: 2.3.5 is available already :)

bgruening: Oh 2.3.8!

pvanheus:
The test data provided here is not appropriate. `pangolin` takes assembled SARS-CoV-2 genomes as input, so there should be 1 FASTA record per genome.

bgruening:
@jpalmer37 thanks a lot! We needed this tool for some urgent COVID work and @pvanheus finished it. I hope that is ok. You will find the tool in a few minutes in the ToolShed.

Thanks a lot for your contributions.

jpalmer37:
Hey @bgruening! Sincerest apologies on my part. Had a very hectic few weeks. It is absolutely okay and I'm glad it got finished! A big thanks for picking up where I left off and for providing great feedback!

bgruening:
No worries, that was great work. The community was just eager to get a hands on it :) Which is good  :)

PR #3523, opened 2021-03-12
Comment change to intervene pairwise

heylf:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/intervene/intervene_pairwise.xml
+++ b/tools/intervene/intervene_pairwise.xml
@@ -89,6 +89,8 @@ mv results/Intervene_pairwise_*.${figtype} results/Intervene_pairwise.plot
 
 Plots the pairwise intersections of multiple interval files.
 
+Make sure your file names have no spaces, e.g., change "SortedBED Test1.bed" to "Test1.bed".
+
 .. image:: $PATH_TO_IMAGES/pairwise.png
 
     ]]></help>


heylf:
This comment is relevant, because the tool runs into an error if the filenames have spaces. The error is a bit unclear and users are confused about it.

bgruening:
@heylf do you have a reproducible case of that? Can you look at the commandline? In theory, we remove whitespaces with https://github.com/heylf/tools-iuc/blob/intervene_update/tools/intervene/macros.xml#L34 from the filenames.

heylf:
@bgruening yes here is a history: https://usegalaxy.eu/u/heylf/h/intervene-test
You have two runs with intervene, once files names without space, and once with spaces. The call with spaces produces an error.

bgruening:
Upps, I do not have access to the history. I think the proper way to address that is to add a test with a space in the element identifier and see how this fails and ultimately fix https://github.com/heylf/tools-iuc/blob/intervene_update/tools/intervene/macros.xml#L34 to make the test pass.

bernt-matthias (commented):
I can reproduce the error with spaces .. seems to be a downstream problem. 

One should do the following:

1. rename one of the test files such that it contains a space. 
2. Add the following after computing the labels

```
#for $label, $file in zip($labels, $input)
    ln -s '$file' `$label` &&
#end for
```

and replace

`--input #echo " ".join(["'%s'" % file for file in $input])` by `--input #echo " ".join(["'%s'" % label for label in $labels])`

PR #3535, opened 2021-03-17
Minimap2 update

NatPRoach:
FOR CONTRIBUTOR:


bernt-matthias:
But ... I'm wondering why the test data does not change ... ?

NatPRoach:
> But ... I'm wondering why the test data does not change ... ?

I think the previous xml won't actually mess up the output on every machine. It seems like it may be causing a problem on GalaxyPro instances in some cases, and even then it's a bit unclear that this is the problem, because sometimes it runs fine. We weren't sure what the problem was, but we noticed this so we figured we'd update it and see if it fixes things.

bgruening:
@bernt-matthias please feel free to merge.

PR #3536, opened 2021-03-18
Adding scoary wrapper and test-data

heylf:
FOR CONTRIBUTOR:


heylf:
@bebatut there are a few things where I wasn't sure about the best solution. 
1. I cannot set the names for the output files and scoary also generates a set of csv's based on the number of traits in the trait file that is why I had to create specific outpout files for the wrapper. 
2. The p-value cutoffs and corrections methods are defined by the user as a string. The problem is that you can define as many p-value cutoffs and corrections as you have traits. 
3. I did not include the option -w (--write_reduced) that states: Use with -r if you want Scoary to create a new gene presence absence file from your reduced set of isolates. Note: Columns 1-14 (No. sequences, Avg group size nuc etc) in this file do not reflect the reduced dataset. These are taken from the full dataset. I was not sure if this is needed?
4. I did not include the option -s (--start_col) that states: On which column in the gene presence/absence file do individual strain info start.I thought the best way is to define a standard input for the tool, else people get more confused why they should decide where the start_col is. 
5. I did not include the option --delimiter that states: The delimiter between cells in the gene presence/absence and trait files, as well as the output file. Such as point 4, I thought it is best if we have a standard input. 
6. There is the possbility to define the number of threads with --threads THREADS. The default is 1. I have not included this options. 
7. Scoary has an extra modul to convert vcf files to scoary input. However, it is in beta stage and it would only be necessary if the user want to analyze variants instead of traits.

tools/scoary/scoary.xml
```diff
@@ -0,0 +1,181 @@
...
+            <param name="permute" argument="-e" type="integer" min="0" value="0" label="Permutations" help="Perform N number of permutations of the significant results post-analysis. (Default = 0 = None)" />
+            <param name="maxhits" argument="-m" type="integer" min="0" value="0" label="Maximal number of hits to report" help="SCOARY will only report the top max hits results per trait. (Default = 0 = All)" />
+            <param name="collapse" argument="--collapse" type="boolean" checked="false" truevalue="--collapse" falsevalue="" label="Collapse correlated genes" help="Collapse correlated genes (genes that have identical distribution patterns in the sample) into merged units. (Default=false)"/>
+            <param name="input_newicktree" optional="true" argument="-n" type="data" format="txt" label="Supply a custom tree (Newick format) for phylogenetic analyses instead instead of calculating it internally." />
+        </section>
+
+        <!-- Output Options -->
+        <section name="output_options" title="Output Options">
```
bgruening, line 73: If its just one parameter, can we add `expanded="true"`

tools/scoary/scoary.xml
```diff
@@ -0,0 +1,181 @@
...
+
+    ]]></command>
+    <inputs>
+        <param name="input_traits" argument="-t" type="data" format="interval" label="Trait table or VCF file"/>
+        <param name="input_genes" argument="-g" type="data" format="interval" label="Gene table from ROARY"/>
+        <param name="input_restricts" optional="true" argument="-r" type="data" format="interval" label="Table to analyze a subset of strains" />
+
+        <!-- Additional Options -->
+        <section name="additional_options" title="Additional Options">
+            <param name="pvalue" argument="-p" type="text" value="0.05" label="P-value cutoff(s)" help="SCOARY will not report genes with higher p-values than this (Default=0.05). Provide a single value (applied to all) or exactly as many values as correction criteria and in corresponding order (e.g., 0.05 0.05 or two traits)."/>
```
bgruening, line 64: What about a repeat here?

tools/scoary/scoary.xml
```diff
@@ -0,0 +1,181 @@
...
+
+--no-time
+
+&&
+tail -n +1 *.csv > scoary_output.csv &&
+mv *.nwk scoary_output.nwk
+
+    ]]></command>
+    <inputs>
+        <param name="input_traits" argument="-t" type="data" format="interval" label="Trait table or VCF file"/>
```
bgruening, line 58: should the format be vcf and not interval?

tools/scoary/scoary.xml
```diff
@@ -0,0 +1,181 @@
...
+        <param name="input_genes" argument="-g" type="data" format="interval" label="Gene table from ROARY"/>
```
bgruening, line 59: is interval really the correct datatype here?

tools/scoary/.shed.yml
```diff
@@ -0,0 +1,8 @@
+name: scoary
+owner: iuc
+description: Scoary calculates the assocations between all genes in the accessory genome and the traits.
+homepage_url: https://github.com/BrooksLabUCSC/flair
```
bgruening, line 4: c&p error :)

tools/scoary/scoary.xml
```diff
@@ -0,0 +1,181 @@
...
+            <param name="permute" argument="-e" type="integer" min="0" value="0" label="Permutations" help="Perform N number of permutations of the significant results post-analysis. (Default = 0 = None)" />
+            <param name="maxhits" argument="-m" type="integer" min="0" value="0" label="Maximal number of hits to report" help="SCOARY will only report the top max hits results per trait. (Default = 0 = All)" />
+            <param name="collapse" argument="--collapse" type="boolean" checked="false" truevalue="--collapse" falsevalue="" label="Collapse correlated genes" help="Collapse correlated genes (genes that have identical distribution patterns in the sample) into merged units. (Default=false)"/>
+            <param name="input_newicktree" optional="true" argument="-n" type="data" format="txt" label="Supply a custom tree (Newick format) for phylogenetic analyses instead instead of calculating it internally." />
```
bgruening, line 69: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/sample/datatypes_conf.xml.sample#L446

we actually have that as input format.

tools/scoary/scoary.xml
```diff
@@ -0,0 +1,181 @@
...
+        <!-- Output Options -->
+        <section name="output_options" title="Output Options">
+            <param name="upgma" argument="-u" type="boolean" checked="false" truevalue="-u" falsevalue="" label="UPGMA tree" help="Calculate UPGMA tree to a newick file."/>
+        </section>
+
+    </inputs>
+
+    <outputs>
+        <data name="outfile1" format="interval" from_work_dir="scoary_output.csv" label="${tool.name} on ${on_string}: Table" />
+        <data name="outfile2" format="txt" from_work_dir="scoary_output.nwk" label="${tool.name} on ${on_string}: Tree">
```
bgruening, line 81: newick format?

tools/scoary/scoary.xml
```diff
@@ -0,0 +1,181 @@
...
+
+        <!-- Output Options -->
+        <section name="output_options" title="Output Options">
+            <param name="upgma" argument="-u" type="boolean" checked="false" truevalue="-u" falsevalue="" label="UPGMA tree" help="Calculate UPGMA tree to a newick file."/>
+        </section>
+
+    </inputs>
+
+    <outputs>
+        <data name="outfile1" format="interval" from_work_dir="scoary_output.csv" label="${tool.name} on ${on_string}: Table" />
```
bgruening, line 80: @heylf can you convert the csv to tsv on the fly ... this way we have more tools that can work with it natively.

tools/scoary/scoary.xml
heylf: You cannot really do. The user can choose different p-vlaues e.g., 0.05 0.01 0.02, so you cannot simply repeat the number.

bgruening: https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-inputs-repeat :)

Repeat a float element. We still can not count the number and match them against the samples, but we could at least have proper floats.

tools/scoary/scoary.xml
heylf: I put the conversion in

tools/scoary/scoary.xml
heylf: jupp

tools/scoary/scoary.xml
heylf: oh good to know

tools/scoary/.shed.yml
heylf: ups

tools/scoary/scoary.xml
heylf: changed!

tools/scoary/scoary.xml
heylf: I changed it to tabular. Its a csv file. The tools does not allow for tsv because it cannot read tabs xD

tools/scoary/scoary.xml
heylf: changed!

tools/scoary/scoary.xml
nsoranzo: If commas are required, Galaxy has a `csv` datatype.

bgruening: And the help states VCF as well.

tools/scoary/scoary.xml
heylf: I added this!

tools/scoary/scoary.xml
heylf: Changed to csv format. VCF files are not directly supported. You would need to convert vcf files into a Scoary readable csv file.

heylf:
I am not sure why the tests still fail. It works for me on my lacal machine.

bgruening:
This is now working. @heylf your tests fail because of rounding issues. I would recommend to use assert_content and check for has_n_columns or has_n_row plus a little bit of content.

tools/scoary/scoary.xml
```diff
@@ -0,0 +1,206 @@
...
+
+--no-time
+&&
+tail -n +1 *.csv | sed "s/\,/\\t/g" > scoary_output.tsv &&
+mv *.nwk scoary_output.nwk
+
+    ]]></command>
+    <inputs>
+        <param name="input_traits" argument="-t" type="data" format="csv" label="Trait table"/>
+        <param name="input_genes" argument="-g" type="data" format="csv" label="Gene table from ROARY"/>
```
bgruening, line 60: Can we maybe say here explicitly which roary output is meant? https://github.com/galaxyproject/tools-iuc/blob/master/tools/roary/roary.xml#L81

heylf: changed!

heylf:
> This is now working. @heylf your tests fail because of rounding issues. I would recommend to use assert_content and check for has_n_columns or has_n_row plus a little bit of content.

Its annoying, I think because of the rounding, the lines in the file also changes.

bgruening:
puh, than take n_columns and has_text? otherwise this tool looks great!

heylf:
> puh, than take n_columns and has_text? otherwise this tool looks great!

Sorry I meant not the number of lines I meant the order of the lines that changes.

bgruening:
https://docs.galaxyproject.org/en/latest/dev/schema.html#id93 --> can you try `sort`

bgruening (approved):

PR #3537, opened 2021-03-18
data_manager_fetch_ncbi_taxonomy: add an option to download accession2taxid data

abretaud:
FOR CONTRIBUTOR:

Added an option to download additional files that the [next version of data_manager_diamond_database_builder](https://github.com/bgruening/galaxytools/pull/1011) might want to use

I've synchronised [this version](https://github.com/abretaud/tools-iuc/blob/1819bbf1b5ac83608571400b685f9800335d5678/data_managers/data_manager_fetch_ncbi_taxonomy/data_manager/ncbi_taxonomy_fetcher.xml#L2) and [this one](https://github.com/abretaud/tools-iuc/blob/1819bbf1b5ac83608571400b685f9800335d5678/data_managers/data_manager_fetch_ncbi_taxonomy/data_manager_conf.xml#L3), not sure why it was not before


bernt-matthias (commented):
Cool. Wondering if we should indicate in the data table if the mapping have been downloaded .. which would require to add a column I guess?

abretaud:
Yes, I considered it, then as I was too lazy to change the table format, I thought the diamond data_manager could just look if the file is there and throw an explicit error if it's not. What do you think?

Anyway, wait before merging, I'm checking if the downloaded files need to unzipped

bernt-matthias:
The main question is if https://docs.galaxyproject.org/en/master/dev/schema.html#tool-inputs-param-options-filter works if some entries do not have a column (old entries will have one column less)

abretaud:
Is it even supported to add a column to an existing table? I have the feeling we usually change the table name when modifying it, but maybe I'm wrong

bernt-matthias:
> Is it even supported to add a column to an existing table? I have the feeling we usually change the table name when modifying it, but maybe I'm wrong

Probably not. Then how about managing two data tables with this DM.

abretaud:
Yeah, why not, you mean keeping the current one as is + adding a new one for accession2taxid only?

bernt-matthias:
> Yeah, why not, you mean keeping the current one as is + adding a new one for accession2taxid only?

Yep .. and they would point to the same data. But I'm not sure if Galaxy will then duplicate the data ...

abretaud:
So I tried to create a separate `ncbi_accession2taxid` table in 5f69520

bernt-matthias:
> So I tried to create a separate ncbi_accession2taxid table in 5f69520

Does this contain also the A2T data or the taxonomy as well (I would vote for containing both since the NCBI taxonomy is not versioned and we should therefor keep them paired such that we can ensure to use A2T and taxonomy from the time)

abretaud:
Only A2T at the moment, but I guess I could add taxonomy too, I'll take a look tomorrow

abretaud:
It should be ok now I think, the tests are passing \o/

bgruening (approved):

bgruening:
Thanks @abretaud

PR #3538, opened 2021-03-18
Fix R linting + lint dexseq

martenson:
GitHub workflows is being migrated to Ubuntu 20.04, which is breaking the R linting CI job with:
```
Error in dyn.load(file, DLLpath = DLLpath, ...) : 
  unable to load shared object '/home/runner/work/_temp/Library/stringi/libs/stringi.so':
  libicui18n.so.60: cannot open shared object file: No such file or directory
Calls: lint_dir ... namespaceImport -> loadNamespace -> library.dynam -> dyn.load
Execution halted
```

This is due to R package cache, its key should depend also on the Ubuntu version.

Fix https://github.com/galaxyproject/tools-iuc/issues/3539

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -23,7 +23,7 @@ jobs:
       chunk-list: ${{ steps.discover.outputs.chunk-list }}
     strategy:
       matrix:
-        python-version: [3.7]
+        python-version: ['3.7']
     steps:
     - name: Add reaction
       if: ${{ github.event.client_payload.slash_command.command == 'run-all-tool-tests' }}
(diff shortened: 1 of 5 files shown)


martenson:
@nsoranzo That seems to have fixed it.

nsoranzo:
OK, so I guess that the problem is the R package cache, its key should depend also on the ubuntu version. I'll change that here to test.

nsoranzo:
Sorry, I've hijacked this pull request! It now fixes the R linting issue on CI, and dexseq linting issues since I was using it to test :)
Green and ready to review.

bgruening (approved):

bgruening:
Cool, thanks!

martenson:
@nsoranzo thanks for the hijack!

PR #3547, opened 2021-03-23
Bedtools update/improve

lldelisle:
FOR CONTRIBUTOR:

This PR:
- allow to use -g in sortBed
- update the version of bedtools to 2.30
- update shuffleBed tests to really check that the arguments are used.
- in coverageBed, instead of using `-sorted` systematically when there is one bam file. Put it as a parameter as it is in intersectBed and provide tests where it is necessary to have it optional.


tools/bedtools/shuffleBed.xml
```diff
@@ -98,7 +101,7 @@ $allow_beyond
             <param name="genome" value="shuffleBed.len" ftype="bed" />
             <param name="allow_beyond" value="True" />
             <param name="seed_choose" value="True" />
-            <param name="seed" value="1" />
+            <param name="seed" value="3" />
```
bernt-matthias, line 104: > update shuffleBed tests to really check that the arguments are used.

Do I miss this part?

Why change the seed value. 

Wondering if there is a better way than comparison with the test files.

tools/bedtools/sortBed.xml
```diff
@@ -23,6 +24,17 @@ $option
             <option value="-chrThenScoreA">chromosome, then by score (asc).</option>
             <option value="-chrThenScoreD">chromosome, then by score (desc).</option>
         </param>
+        <conditional name="genome">
```
bernt-matthias, line 27: This part could be a macro.

tools/bedtools/shuffleBed.xml
lldelisle: I changed the seed value because when I put the seed value to 1 the shuffled interval would be the same with or without `allow_beyond = True`. I wanted to set a test that would raise an error if we change the wrapper  and do not set correctly this option.

tools/bedtools/sortBed.xml
lldelisle: But can I do a macro which uses a macro? Here I am using `<expand macro="input_conditional_genome_file" />`

bernt-matthias: > But can I do a macro which uses a macro?

Sure. 

I was assuming that this conditional corresponds to `@GENOME_FILE_MAPBED@` and therefore the same conditional exists somewhere else as a copy .. which would qualify it for a macro.

lldelisle: Sorry, I did not know we could use expand macro within a macro...

bgruening (approved):
Cool, thanks a lot @lldelisle!

lldelisle:
@bernt-matthias , is it ok for you?

bernt-matthias (approved):
Looks excellent. Thanks

PR #3549, opened 2021-03-23
minimap2: support gzip-ed input

bernt-matthias:
FOR CONTRIBUTOR:


tools/minimap2/minimap2.xml
```diff
@@ -23,10 +23,22 @@
...
+        #set reference=$reference_source.ref_file
+        #set ext=$reference.ext
     #else:
-        ln -f -s '$reference_source.ref_file.fields.path' reference.fa &&
+        #set reference=$reference_source.ref_file.fields.path
+        #set ext=$reference_source.ref_file.fields.path
     #end if
+
+    ## unzip if needed (minimap2 supports zipped input, but it needs
+    ## to be bgzip-ped which we can not ensure)
```
peterjc, line 34: That seems incredibly wasteful on larger files. For many use cases it would be better to leave it as is (especially if minimap2 gives a clear error message but that I have not checked).

There isn't any neat trick with ``bgzip`` for normal GZIP to blocked GZIP (with a no-op shortcut when already blocked), is there? https://www.htslib.org/doc/bgzip.html

bernt-matthias: > That seems incredibly wasteful on larger files. For many use cases it would be better to leave it as is (especially if minimap2 gives a clear error message but that I have not checked).

Jep. There is a kind of clear error message in the stderr (which might not be consulted by not so advanced users .. but we might highlight this with a stdio). My problem is then that this would be a dead end for Galaxy users, because Galaxy (currently) has no means to transform gz to bgz (as far as I know).

> There isn't any neat trick with bgzip for normal GZIP to blocked GZIP (with a no-op shortcut when already blocked), is there? https://www.htslib.org/doc/bgzip.html

Maybe ` --reindex`, but this still "wastes" CPU time (I guess in the same order of magnitude than just decompressing it) but wastes less temporary space (I do not care much about this). I guess the main question is if there is an easy way to detect with `bash`  if a file is bgziped? Maybe we even need Galaxy datatypes for this, i.e. `fasta.bgzip`? There seems to be some support already (`vcf.bgzip`). But then Galaxy would again do the implicit conversion (now gz->bgzip) which adds to the history.

bgruening: > because Galaxy (currently) has no means to transform gz to bgz (as far as I know).

If this is blocking we should fix that with a new converter, similar to https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/datatypes/converters/bz2_to_uncompressed.xml

> I guess the main question is if there is an easy way to detect with bash if a file is bgziped?

http://samtools.github.io/hts-specs/SAMv1.pdf you could use the first to bytes?

bernt-matthias: I think `file` can detect bgzf. Do BAM and bgzip have the same magix bytes? Then this would be painful to implement in Galaxy (sniffing etc).

bgruening: I don't know, but afaik BAM is a bgziped archive.

peterjc: BGZF is a specialised version of GZIP, the only way to be sure 100% it is all blocked gzip is to examine all the headers (it could for example be BGZF at the start followed by plain old GZIP, pathological but still valid GZIP). What we do in Biopython is try it and see:

https://github.com/biopython/biopython/blob/master/Bio/File.py#L90
https://github.com/biopython/biopython/blob/master/Bio/bgzf.py

Sniffing BAM is easier, you decompress the start of the BGZF (GZIP) file and look for the magic BAM string b"\x1f\x8b\x08\x04" aka 0x1F 0x8B 0x08 0x04 aka little endian BAM\1, see:

https://samtools.github.io/hts-specs/SAMv1.pdf

peterjc: You could make a pretty good BGZF sniffer if you checked for the GZIP magic and that the file ended with the special empty block EOF marker though... 

```python
b"\x1f\x8b\x08\x04\x00\x00\x00\x00\x00\xff\x06\x00BC\x02\x00\x1b\x00\x03\x00\x00\x00\x00\x00\x00\x00\x00\x00"
```

bernt-matthias: Thanks for the input @peterjc .. then we probably also need to check the gz sniffer for the absence of the EOF marker .. or sniff first for bgzip.

Or we extend the gz data type (actually I do not know if it is an extra data type) and add a metadata field `is_blocked_gz`.

peterjc: Probably the sniffing order should be BAM, other BGZF file, generic GZIP file.

I'm unclear that Galaxy actually generally treats compressed files as a distinct datatype through, but have not kept up with any changes here lately.

wm75: @bernt-matthias leaving the sniffer discussion aside for the moment:
What exactly happens when you pass an ordinary gzipped fasta?
I just tried simple indexing with `minimap2 -d index.mmi my_fasta.fa.gz` and that didn't complain?
Does that mean it's only at the alignment stage that an error gets thrown? Or is on-the-fly indexing different from indexing to a file?

bgruening:
related to that wrapper: https://github.com/galaxyproject/tools-iuc/pull/3590/files

tools/minimap2/minimap2.xml
bernt-matthias: @wm75 The error is `[E::fai_build3_core] Cannot index files compressed with gzip, please use bgzip` which actually comes from the `samtools sort` call.

bernt-matthias: ... which apparently needs the reference for creating cram

bernt-matthias: ... but then it seems also odd to use `--output-fmt-option no_ref`

wm75: Ah, but that should make detection of incompatible gzip simple and efficient.
If samtools sort needs to index the ref anyway, you could just as well attempt `samtools faidx` up front in the general ref handling section if the output type is cram, and decompress only if that command fails?

wm75: ... and why isn't this an issue in many more wrappers that need to index their ref?

bernt-matthias: I guess most of them do not consider zipped fasta.

PR #3564, opened 2021-03-31
pretext snapshot add

astrovsky01:
FOR CONTRIBUTOR:


tools/pretext/pretext_snapshot.xml
```diff
@@ -0,0 +1,170 @@
+<tool id="pretext_snapshot" name="Pretext Snapshot"  version="@VERSION@+@GALAXY_STRING@" python_template_version="3.5">
+    <macros>
+        <token name="@VERSION@">0.03</token>
+        <token name="@GALAXY_STRING@">galaxy0</token>
```
bgruening, line 4: can you use VERSION_SUFFIX here, please see: https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tool-versions

tools/pretext/pretext_snapshot.xml
```diff
@@ -0,0 +1,170 @@
+<tool id="pretext_snapshot" name="Pretext Snapshot"  version="@VERSION@+@GALAXY_STRING@" python_template_version="3.5">
```
bgruening, line 1: drop python_template and please add profile="20.01"

tools/pretext/pretext_snapshot.xml
```diff
@@ -0,0 +1,170 @@
...
+    <macros>
+        <token name="@VERSION@">0.03</token>
+        <token name="@GALAXY_STRING@">galaxy0</token>
+    </macros>
+    <requirements>
+        <requirement type="package" version="0.02">pretext-suite</requirement>
```
bgruening, line 7: @VERSION@ ?? 0.02 or 0.03

tools/pretext/pretext_snapshot.xml
```diff
@@ -0,0 +1,170 @@
...
+        <conditional name="formats">
+            <param name="outformat" argument="-f" type="select" label="Output image format" help="Format of output image">
+                <option value="png" selected="true"/>
+                <option value="bmp"/>
+                <option value="jpeg"/>
+            </param>
+            <when value="png"/>
+            <when value="bmp"/>
+            <when value="jpeg">
+                <param name="jpegQuality" argument="jpegQuality" type="integer" label="JPEG quality factor" value="80" help="An integer between 1 and 100, default 80. Larger values result in increased image quality and file size."/>
```
bgruening, line 41: min/max?

tools/pretext/pretext_snapshot.xml
astrovsky01: The suite is version 0.2 on bioconda, but the tool inside is 0.03

tools/pretext/pretext_snapshot.xml
```diff
@@ -0,0 +1,170 @@
...
+            <param name="sequences" value="=full, =all"/>
+            <param name="mintexels" value="30"/>
+            <conditional name="grid">
+                <param name="showGrid" value="yes"/>
+                <param name="gridsize" value="2"/>
+                <param name="gridcolor" value="red"/>
+            </conditional>
+            <param name="sequencenames" value="--printSequenceNames"/>
+            <output_collection name="pretext_snap_out">
+                <element name="pretext_snapshotFullMap" file="fullmap1.jpeg"/>
```
bgruening, line 81: can you add ftype here to check that Galaxy has recognised the correct filetype?

tools/pretext/pretext_snapshot.xml
```diff
@@ -0,0 +1,170 @@
...
+            <param name="sequences" value="=full, =all"/>
+            <param name="mintexels" value="30"/>
+            <conditional name="grid">
+                <param name="showGrid" value="yes"/>
+                <param name="gridsize" value="2"/>
+                <param name="gridcolor" value="red"/>
+            </conditional>
+            <param name="sequencenames" value="--printSequenceNames"/>
+            <output_collection name="pretext_snap_out">
+                <element name="pretext_snapshotFullMap" ftype="jpeg" file="fullmap1.jpeg"/>
```
bgruening, line 81: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/sample/datatypes_conf.xml.sample#L201

galaxy has `jpg` ... that might be a problem. I guess you need to rename the files in this case so that they end in '.jpg'

tools/pretext/.shed.yml
```diff
@@ -0,0 +1,12 @@
+categories:
```
bernt-matthias, line 1: Can you add `auto_tool_repositories` like in https://github.com/gregvonkuster/tools-iuc/blob/master/tools/bamutil/.shed.yml?

tools/pretext/pretext_snapshot.xml
```diff
@@ -0,0 +1,174 @@
+<tool id="pretext_snapshot" name="Pretext Snapshot"  version="@VERSION@+@VERSION_SUFFIX@" profile="20.01">
+    <macros>
+        <token name="@VERSION@">0.03</token>
```
bernt-matthias, line 3: Please change to `WRAPPER_VERSION`

tools/pretext/pretext_snapshot.xml
```diff
@@ -0,0 +1,174 @@
...
+            --gridSize $grid.gridsize
+            --gridColour $grid.gridcolor
+        #else:
+            --gridSize 0
+        #end if
+        '${sequencenames}'
+        -o output
+        --prefix pretext_snapshot
+        #if $formats.outformat == 'jpeg':
+            && find ./output/* | rename -s .jpeg .jpg
```
bernt-matthias, line 31:
```suggestion
            && find ./output/ -name "*.jped" -exec echo mv {} `basename {} .jpeg`.jpg \; && true
```

should work as well and you can drop the rename requirement (note `&& true` is necessary because Galaxy would remove the last `;`)

tools/pretext/pretext_snapshot.xml
```diff
@@ -0,0 +1,174 @@
+<tool id="pretext_snapshot" name="Pretext Snapshot"  version="@VERSION@+@VERSION_SUFFIX@" profile="20.01">
+    <macros>
+        <token name="@VERSION@">0.03</token>
+        <token name="@VERSION_SUFFIX@">galaxy0</token>
+    </macros>
+    <requirements>
+        <requirement type="package" version="0.02">pretext-suite</requirement>
```
bernt-matthias, line 7:
```suggestion
        <requirement type="package" version="@WRAPPER_VERSION@">pretext-suite</requirement>
```

tools/pretext/pretext_snapshot.xml
```diff
@@ -0,0 +1,174 @@
...
+        #end if
+        '${sequencenames}'
+        -o output
+        --prefix pretext_snapshot
+        #if $formats.outformat == 'jpeg':
+            && find ./output/* | rename -s .jpeg .jpg
+        #end if
+    ]]></command>
+    <inputs>
+        <param name="input" type="data" format="pretext" label="Input Pretext map file"/>
```
bernt-matthias, line 35: Please add the `argument` attribute. To all input parameters

tools/pretext/pretext_snapshot.xml
```diff
@@ -0,0 +1,174 @@
...
+        <conditional name="formats">
+            <param name="outformat" argument="-f" type="select" label="Output image format" help="Format of output image">
+                <option value="png" selected="true"/>
```
bernt-matthias, line 38: Shouldn't it be `<option value="png" selected="true">png</option>`? Or is Galaxy showing the value automatically?

tools/pretext/pretext_snapshot.xml
```diff
@@ -0,0 +1,174 @@
...
+            </param>
+            <when value="png"/>
+            <when value="bmp"/>
+            <when value="jpeg">
+                <param name="jpegQuality" argument="jpegQuality" type="integer" label="JPEG quality factor" value="80" min="0" max="100" help="An integer between 1 and 100, default 80. Larger values result in increased image quality and file size."/>
+            </when>
+        </conditional>
+        <param name="resolution" argument="-r" type="integer" label="Output image resolution" value="1000" help="Image resolution, a positive integer, default 1080. For non-square images this will be the resolution of the longest dimension."/>
+        <param name="colormap" argument='-c' type="integer" label="Color Map" value="5" min="0" max="30" help="Color map based on list available in the help section"/>
+        <param name="sequences" argument="--sequences" type="text" label="Sequence specification string" value="=full, =all" help="Each entry, except for '=all', corresponds to one output image. More information available in the help section."/>
```
bernt-matthias, line 50: I would suggest to make this a select if only `""`, `=full`, and `=full, =all` are possible. Otherwise we will need a validator or sanitizer.

tools/pretext/pretext_snapshot.xml
```diff
@@ -0,0 +1,174 @@
...
+        <param name="mintexels" argument="--minTexels" type="integer" label="Min Texels" value="64" help="Minimum map texels per image (along a single dimension), a positive integer, Output images over too small a range that violate this minimum will not be created."/>
```
bernt-matthias, line 51: Are there min and max values?

tools/pretext/pretext_snapshot.xml
```diff
@@ -0,0 +1,174 @@
...
+            <param name="colormap" value="1"/>
+            <param name="sequences" value="=full, =all"/>
+            <param name="mintexels" value="30"/>
+            <conditional name="grid">
+                <param name="showGrid" value="yes"/>
+                <param name="gridsize" value="2"/>
+                <param name="gridcolor" value="red"/>
+            </conditional>
+            <param name="sequencenames" value="--printSequenceNames"/>
+            <output_collection name="pretext_snap_out">
```
bernt-matthias, line 84: Can you add the `count` attribute to the output collections?

tools/pretext/pretext_snapshot.xml
astrovsky01: These are different values. The tool version != the version in bioconda. The bioconda package is the full suite, which is at v0.02, as opposed to PretextSnapshot inside the package,  which is 0.03

tools/pretext/pretext_snapshot.xml
astrovsky01: When not specified, the value is used as the label

tools/pretext/pretext_snapshot.xml
astrovsky01: More are possible, I will make a sanitizer. The details are in the help section

tools/pretext/pretext_snapshot.xml
astrovsky01: This isn't actually moving anything, even with the typo fixed

tools/pretext/pretext_snapshot.xml
bernt-matthias: Got it .. its a meta package.. Problem is that with the unspecified requirements in https://github.com/bioconda/bioconda-recipes/blob/master/recipes/pretext-suite/meta.yaml you will always get the latest version.

So you need to either fix the bioconda recipe or use the PretextSnapshot requirement directly

astrovsky01: I was hoping to be able to just use suite since we are adding the other commands as well, but that makes sense. I'll switch it to the individual requirement

bgruening (approved):

PR #3590, opened 2021-04-12
Move preset selector outside fastq input conditional.

davebx:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/minimap2/minimap2.xml
+++ b/tools/minimap2/minimap2.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<tool id="minimap2" name="Map with minimap2" version="@TOOL_VERSION@+galaxy4" profile="17.01">
+<tool id="minimap2" name="Map with minimap2" version="@TOOL_VERSION@+galaxy5" profile="17.01">
     <description>A fast pairwise aligner for genomic and spliced nucleotide sequences</description>
     <macros>
         <token name="@TOOL_VERSION@">2.17</token>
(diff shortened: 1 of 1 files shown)


bernt-matthias:
Can we merge this with https://github.com/galaxyproject/tools-iuc/pull/3549

wm75:
What's the motivation behind this change?
In other words, which of the presets would you like to apply to PE reads?

davebx:
@wm75 some users were experiencing difficulty finding the presets option at all, so I reasoned that putting it just below the input parameter would make it easier to find.

Is there a technical reason to only use the presets on single-end reads? Looking at the manual, I see that -x is occasionally used in examples that also have paired reads, so I would speculate it is at least supported by the software. Whether it's of any use would be a question for someone more experienced with sequencing.

wm75:
> some users were experiencing difficulty finding the presets option at all, so I reasoned that putting it just below the input parameter would make it easier to find.

Hmm, it was me who moved the presets *into* the conditional in a bigger revision of the wrapper last summer, and I did so to *reduce* confusion not to make things worse. Now it can be, of course, that it's just people who've used the previous version that have difficulties to adapt?

> Is there a technical reason to only use the presets on single-end reads? Looking at the manual, I see that -x is occasionally used in examples that also have paired reads, so I would speculate it is at least supported by the software. Whether it's of any use would be a question for someone more experienced with sequencing.

Technically, it is indeed  possible to provide any preset with any kind of input, but I concluded that this not by design, but rather to avoid too complicated command line parsing, or, at least, I thought I had considered what's useful for which scenario rather carefully last year. Now when I saw this PR I thought I had accidentally prevented some valid use of the tool and felt guilty. If it's just a matter of getting used to the interface I'm much less concerned.

In fact my change prevents users from selecting anything but -xsr as the preset for PE reads, and that's important because without that minimap2 treats the data as SE, i.e., doesn't set PE flag bits correctly.
See https://github.com/lh3/minimap2/issues/190#issuecomment-399615263 and tested by me last year.

wm75:
Ha, I even left a comment in the wrapper: https://github.com/galaxyproject/tools-iuc/blob/55e703f62a7aa03a02003056751a693fbb95ebf2/tools/minimap2/minimap2.xml#L217
about this last issue. I'm surprising myself :)

davebx:
@wm75 that is convincing logic, I'm closing this.

PR #3591, opened 2021-04-13
implement --genotype option, add test

FredericBGA:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/bcftools/bcftools_view.xml
+++ b/tools/bcftools/bcftools_view.xml
@@ -125,7 +125,7 @@ ${section.drop_genotypes}
                 <help>
                 Include only sites with one or more homozygous (hom), heterozygous (het) or missing (miss) genotypes.
                 When prefixed with ^, the logic is reversed.
-                Please notice taht if the input file doesn't have any genotype columns,
+                Please notice that if the input file doesn't have any genotype columns,
                 then this option is ignored altogether.
                 </help>
                 <option value="hom">hom</option>
--- a/tools/bcftools/bcftools_view.xml
+++ b/tools/bcftools/bcftools_view.xml
@@ -38,7 +38,7 @@ ${section.force_samples}
 #if str($section.max_ac)
   --max-ac ${section.max_ac}
 #end if
-#if str($section.select_genotype) != "__none__":
+#if str($section.select_genotype) != "None":
   --genotype "${section.select_genotype}"
 #end if
 ## known or novel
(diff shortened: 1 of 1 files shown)


tools/bcftools/bcftools_view.xml
```diff
@@ -121,13 +121,19 @@ ${section.drop_genotypes}
...
             <param name="max_ac" type="integer" label="Max Ac" optional="True"
                    help="(-C --max-ac) Maximum count for non-reference (nref), 1st alternate (alt1), least frequent (minor), most frequent (major) or sum of all but most frequent (nonmajor) alleles" />
-            <param name="select_genotype" type="select">
+            <param name="select_genotype" type="select" label="Select Genotype">
+                <help>
+                Include only sites with one or more homozygous (hom), heterozygous (het) or missing (miss) genotypes.
+                When prefixed with ^, the logic is reversed
+                </help>
                 <option value="__none__" selected="True">No selection</option>
-                <option value="require one or more hom/het/missing genotype or" />
```
wm75, line 126: Instead of using the `__none__` value, consider making the param *optional* (value would be `None` then if the user doesn't select anything.

tools/bcftools/bcftools_view.xml
```diff
@@ -121,13 +121,19 @@ ${section.drop_genotypes}
...
-                <option value="if prefixed with &quot;^&quot;" />
-                <option value="exclude sites with hom/het/missing genotypes" />
+                <option value="hom" >hom</option>
+                <option value="het" >het</option>
+                <option value="miss" >miss</option>
+                <option value="^hom" >^hom</option>
+                <option value="^het" >^het</option>
+                <option value="^miss" >^miss</option>
```
wm75, line 135: 👍 just remove the space between each value and the closing `>`.

wm75 (commented):
Nice! Thank you!

tools/bcftools/bcftools_view.xml
FredericBGA: ugly indeed!

tools/bcftools/bcftools_view.xml
FredericBGA: ok, I will do that, I've modified only what I needed

wm75 (approved):
Thanks, works nicely!

Just tested this and one thing I noticed is that when the input file doesn't have any genotype columns, then the option is ignored altogether.

Maybe that's worth mentioning in the param help.

Otherwise, this looks ready to me.

FredericBGA:
Yes, it makes sense, but this is not even mention in the official doc. Do we need to go further than this?
I'm not an end user here, but it makes sense too that without any genotype columns that option will be a kind of ghost.
It's easy to add, so let me know.

wm75:
Not sure. It surprised *me* at least. Would have assumed that this would count as missing genotype and any other filter than "miss" would return the header only.
In general, I'm always voting for improving on the original tool's documentation if we can :)

PR #3618, opened 2021-04-22
Ena upload update

ieguinoa:
Update ena_upload:
- Upgrade cli version
- Improved output to collect all metadata provided by the user (needed if they later want to do a consensus submission)
- Add support for paired list as input + test for it
- ..


tools/ena_upload/extract_tables.py
```diff
@@ -3,6 +3,11 @@
 import pathlib
 from datetime import datetime
 
+"""
+Parse the configfile generated by the Galaxy tool.
+This file is json formatted and should be converted to a set of tabular files.
```
nsoranzo, line 8:
```suggestion
This file is JSON-formatted and should be converted to a set of tabular files.
```

tools/ena_upload/samples_macros.xml
```diff
@@ -181,9 +198,11 @@
                         <column name="value" index="0"/>
                     </options>
                 </param>
+                <!--<conditional name="input_select">-->
```
nsoranzo, line 201:
```suggestion
```

tools/ena_upload/samples_macros.xml
```diff
@@ -181,9 +198,11 @@
...
                 <repeat name="rep_runs" title="Runs executed within this experiment" min="1" >
                     <param name="upload_files" type="data" format="fastqsanger.gz,fastqsanger.bz2,fastq.gz,fastq.bz2" multiple="true" label="File(s) associated with this run"/>
                 </repeat>
+                <!--</conditional>-->
```
nsoranzo, line 205:
```suggestion
```

tools/ena_upload/ena_upload.xml
```diff
@@ -1,60 +1,62 @@
-<tool id="ena_upload" name="ENA Upload tool" version="0.3" profile="20.01" license="MIT">
+<tool id="ena_upload" name="ENA Upload tool" version="0.4" profile="20.01" license="MIT">
```
nsoranzo, line 1: Not sure if there's a goal of syncing the tool version with the `ena-upload-cli`'s one. If yes, you can slow down the version here hoping the requirement version catches up.
```suggestion
<tool id="ena_upload" name="ENA Upload tool" version="0.3.1" profile="20.01" license="MIT">
```

ieguinoa: that sounds good

ieguinoa:
@bgruening can you check if the paired input feature fits the format used by the covid19 workflows?

tools/ena_upload/process_xlsx.py
```diff
@@ -36,15 +36,45 @@ def extract_data(xl_sheet, expected_columns):
...
+            tmp = data_dict[xl_sheet.cell(row_id, index_col).value]
+            data_dict[xl_sheet.cell(row_id, index_col).value] = [tmp]
+            data_dict[xl_sheet.cell(row_id, index_col).value].append(row_dict)
+        else:
+            data_dict[xl_sheet.cell(row_id, index_col).value] = row_dict
     return data_dict
 
 
+def paste_xls2yaml(xlsx_path):
+    print('YAML -------------')
```
bgruening, line 49: should this be removed?

tools/ena_upload/process_xlsx.py
```diff
@@ -36,15 +36,45 @@ def extract_data(xl_sheet, expected_columns):
...
+        # skip first 2 rows (column names and suggestions)
+        for row_id in range(2, xls_sheet.nrows):
+            row_dict = {}
+            for col_id in range(0, xls_sheet.ncols):
+                row_dict[colnames[col_id]] = xls_sheet.cell(row_id, col_id).value
+            # should check for duplicate alias/ids?
+            sheet_contents_dict[row_id] = row_dict
+        content_dict[sheet_name] = sheet_contents_dict
+    yaml.dump(content_dict, sys.stdout)
+    print('YAML -------------')
```
bgruening, line 69: remove?

bgruening:
@ieguinoa should we wait for https://github.com/usegalaxy-eu/ena-upload-cli/pull/10 or should this go into the next version. Estonia needs it as it seems!

tools/ena_upload/process_xlsx.py
ieguinoa: this is actually a feature. We use it to print in the same submission receipt all the metadata that was initially in the excel sheet or filled by the user in the UI. It's as a way of tracking provenance and also we use this receipt later as input for the consensus sequence submission, as it collects all data from user + accessions obtained from ena.

ieguinoa:
I haven't seen that PR, I'm willing to add it once it's ready in the new cli, but I would merge this version now.

wm75:
https://github.com/galaxyproject/tools-iuc/blob/c88b2e6ab83a055fc40295b14f7369b9f5c5a514/tools/ena_upload/samples_macros.xml#L122

maybe add a `selected="true"` here to make a common case the default.

wm75:
https://github.com/galaxyproject/tools-iuc/blob/c88b2e6ab83a055fc40295b14f7369b9f5c5a514/tools/ena_upload/samples_macros.xml#L137

There exists a **run_alias** field in ENA to let users specify their internal identifier/label of the run. I assume it's possible to specify this in the excel/user table branches? Would be good to offer here, too.
Estonian SARS-CoV-2 genome project samples specify this, for example.

wm75:
https://github.com/galaxyproject/tools-iuc/blob/c88b2e6ab83a055fc40295b14f7369b9f5c5a514/tools/ena_upload/samples_macros.xml#L71
https://github.com/galaxyproject/tools-iuc/blob/c88b2e6ab83a055fc40295b14f7369b9f5c5a514/tools/ena_upload/samples_macros.xml#L72

The help seems to be wrong for this param: according to the validator and from inspecting existing ENA metadata the date formats seems to use `-` as the separator, not `/`.
In addition, the validator has a typo in `not collected`!

Beyond that, it may be helpful to use *suggested options* (https://docs.galaxyproject.org/en/latest/dev/schema.html#id31) on the text param here.

wm75:
> https://github.com/galaxyproject/tools-iuc/blob/c88b2e6ab83a055fc40295b14f7369b9f5c5a514/tools/ena_upload/samples_macros.xml#L137
> 
> There exists a **run_alias** field in ENA to let users specify their internal identifier/label of the run. I assume it's possible to specify this in the excel/user table branches? Would be good to offer here, too.
> Estonian SARS-CoV-2 genome project samples specify this, for example.

Aha, I understood now that all aliases (at the study, samples, experiment *and* run level) are autogenerated in `extract_tables.py`. Maybe they could all have optional text fields, too?

tools/ena_upload/ena_upload.xml
```diff
@@ -91,15 +93,27 @@ mkdir ./submission_files;
...
 #else:
-    #for $file in $action_options.input_format_conditional.data:
-        #set $safename_reads_file = re.sub('[^\w\-_\.]', '_', $file.element_identifier)
-        ln -s '$file' $safename_reads_file &&
-        $files_to_upload.append(str($safename_reads_file))
-    #end for
+    #if $action_options.input_format_conditional.run_input_format_conditional.run_input_format == 'paired_list':
+        #for $pair in $action_options.input_format_conditional.run_input_format_conditional.paired_end_collection:
+            #set $safename_fwd_reads_file = re.sub('[^\w\-_\.]', '_', $pair.forward.name)
+            #set $safename_rev_reads_file = re.sub('[^\w\-_\.]', '_', $pair.reverse.name)
```
wm75, line 99: This won't work since `$pair.forward.name` and `$pair.reverse.name` will always be `forward` and `reverse`.

Something along the lines of:

```
            #set $safename_reads_file = re.sub('[^\w\-_\.]', '_', $pair.name)
            #set $safename_fwd_reads_file = $safename_reads_file + '_1'
            #set $safename_rev_reads_file = $safename_reads_file + '_2'
```

should work and catch the actual sample identifier.

wm75 (changes requested):

tools/ena_upload/ena_upload.xml
ieguinoa: I think that won't work in this case. I need the exact original identifier of each sample in the paired collection. The pair.name returns the name of the collection and we can't assume any relation with the element identifier, plus the user can just define a new one when creating the collection pair.
But as far as I could test, the $pair.forward.name gives the element identifier, while the $pair.forward.element_identifier always gives "forward" or "reverse" as you mentioned. As opposed to the single end reads (which is the other case of this conditional) that gives the real element identifier when using $file.element_identifier.
This was actually the core problem for me to add the paired end option, that's why I added a specific test for this, which creates a paired end collection and checks that the --data parameter in the command line (iterating over $files_to_upload and printing the elements) correctly lists the element identifiers.

tools/ena_upload/ena_upload.xml
```diff
@@ -1,60 +1,62 @@
...
 #if $action_options.input_format_conditional.input_format != "user_generated_tables":
     cp $studies_table_path $studies_table_out;
     cp $samples_table_path $samples_table_out;
     cp $experiments_table_path $experiments_table_out;
     cp $runs_table_path $runs_table_out;
-    #if $action_options.input_format_conditional.dry_run == "true":
-      #set $dry_run_option = "True"
-    #end if
+#else:
+    ln -s '$action_options.input_format_conditional.experiments_users_table' $experiments_table_path
```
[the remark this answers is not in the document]
ieguinoa: added

tools/ena_upload/ena_upload.xml
wm75: You're probably right about the `$pair.forward.element_identifier` vs `$pair.forward.name` distinction, but from a practical point of view, the user normally has the `$pair.name` under their control and not the `$pair.forward.name`.
The `$pair.name` will propagate through WFs that somehow start with a list of identifiers so by ensuring appropriate input, the user will automatically get good `$pair.name`s.
OTOH, a user *can* change the `forward` and `reverse` names, but only extremly tediously - by going to the pencil icon of each individual dataset.
I see the point about the assumption made about the `_1`, `_2` suffix in my suggestion, but this is standard ENA naming and we could just mention in the help that samples should be following that nomenclature in the table/xls.

ieguinoa: yes, you are right about the _1 and _2 , and also would definitely be better to have some standardization with this.  I got some comments from people that the _1 and _2 is a standard for ENA but at least for the test servers I was able to push paired data with other names, so not sure what is the strict rule and couldn't find any official documentation on this. Do you have a link for this? or any more details... e.g is _R1 and _R2 allowed too?

wm75: I don't have any documentation link for you, but even if things like `_R1`, `_R2` are allowed it doesn't mean we have to support that in the Galaxy wrapper. All that's needed imo is a help section explaining that no matter what your files were called before you uploaded them to Galaxy, once they are in Galaxy they will be called <paired_collection_name>_1 and <paired_collection_name>_2 and this is how they should be listed in the tabular/xls form.

wm75: I just played this through with a concrete list:paired collection and it just makes more sense. The point is `$pair.forward.name` and `$pair.reverse.name` will initially be the names of the datasets you created the PE collection from. Assuming you haven't set new names for these datasets during and after uploading, these names will correspond to the file names that the user may know from outside Galaxy.
Now, however, imagine that user runs a single processing step inside Galaxy on that PE collection, the names will change depending on the tool. For example after running fastp on the collection, `$pair.forward.name` may be something like `fastp on data 4 and data 3`.

ieguinoa: ok, that changes things then... I didn't know tools could change the .name value. That's actually the reason why I asked if this feature would fit the way of handling paired lists in the workflows. Anyway, I changed it to fit your proposal and also added the extension to the files because the ena-upload-cli running downstream requires the extension to be in place.

wm75:
Cool, I was just going to suggest what you did in that last commit 👍

ieguinoa:
> https://github.com/galaxyproject/tools-iuc/blob/c88b2e6ab83a055fc40295b14f7369b9f5c5a514/tools/ena_upload/samples_macros.xml#L71
> 
> 
> https://github.com/galaxyproject/tools-iuc/blob/c88b2e6ab83a055fc40295b14f7369b9f5c5a514/tools/ena_upload/samples_macros.xml#L72
> 
> The help seems to be wrong for this param: according to the validator and from inspecting existing ENA metadata the date formats seems to use `-` as the separator, not `/`.
> In addition, the validator has a typo in `not collected`!
> 
> Beyond that, it may be helpful to use _suggested options_ (https://docs.galaxyproject.org/en/latest/dev/schema.html#id31) on the text param here.

changed the collection_date parameter help, it needs to be separated with '-' and not '/'.
But didnt understand the idea of using options (was that what you meant?) as the date could be any date in the past.

wm75:
> But didnt understand the idea of using options (was that what you meant?) as the date could be any date in the past.

I meant for the `not collected`, `not provided`, `restricted access` options.

ieguinoa:
> > But didnt understand the idea of using options (was that what you meant?) as the date could be any date in the past.
> 
> I meant for the `not collected`, `not provided`, `restricted access` options.

oh ok, done

tools/ena_upload/ena_upload.xml
```diff
@@ -91,15 +93,37 @@ mkdir ./submission_files;
...
-        #set $safename_reads_file = re.sub('[^\w\-_\.]', '_', $file.element_identifier)
-        ln -s '$file' $safename_reads_file &&
-        $files_to_upload.append(str($safename_reads_file))
-    #end for
+    #if $action_options.input_format_conditional.run_input_format_conditional.run_input_format == 'paired_list':
+        #for $pair in $action_options.input_format_conditional.run_input_format_conditional.paired_end_collection:
+            #set $safename_reads_file = re.sub('[^\w\-_\.]', '_', $pair.name)
+            #if $pair.forward.is_of_type('fastq.gz', 'fastqsanger.gz'):
+                #set $safename_fwd_reads_file = $safename_reads_file + '_1.fastq.gz'
+            #else:
```
wm75, line 101: The tool accepts also `fastq.bz2`/`fastqsanger.bz2`.
Not sure if ENA allows that, actually, and if what suffix they ask for then, but looks like that case needs handling here or be excluded from the input types.

wm75: Apart from this, this looks all good to me.

bgruening:
There are now 3 additional columns available in the newest version (https://github.com/bioconda/bioconda-recipes/pull/28293):

* geographic_location_region
* geographic_location_latitude
* geographic_location_longitude

See here https://github.com/usegalaxy-eu/ena-upload-cli#supported-columns-for-viral-sample-submissions

Your call, if this should go in here or the next version.

ieguinoa:
I would put it in a separate version. I can start working right away on it and shouldn't be complicated but don't want to break anything for users that may have the metadata templates in this format. 
Also, some of the changes in this PR are part of what the reviewers commented on the paper we sent that uses this tool so I would prefer to have it pushed now.

wm75 (approved):
Thanks @ieguinoa !

PR #3624, opened 2021-04-23
Add nextclade tool

pvanheus:
FOR CONTRIBUTOR:


pvanheus:
This is a wrapper for the nextclade CLI tool (`nextclade_js` in bioconda).

It provides input for all the input files accepted by nextclade. These have been tested as there is very little documentation on how they work.

Three output types are supported: tabular, JSON and tree. The tabular and JSON formats provide the same data in the same format and the tree output provides an Auspice v.2 tree format output (in JSON). The CSV and "clades only" outputs from the underlying tool are not supported as they are very similar to the existing output options.

tools/nextclade/nextclade.xml
```diff
@@ -0,0 +1,131 @@
...
+        #end if
+    ]]></command>
+    <inputs>
+        <param argument="--input-fasta" name="input_fasta" type="data" format="fasta" label="SARS-CoV-2 consensus sequences (FASTA)" />
+        <param name="outputs" type="select" label="Output options" help="Output reports and optionally tree" multiple="true">
+            <option value="report_json">JSON format report</option>
+            <option value="report_tsv" selected="true">Tabular format report</option>
+            <option value="output_tree">Auspice v2 tree file (JSON format)</option>
+        </param>
+        <conditional name="adv">
```
gallardoalba, line 48: It would be nice to include some additional tests for the advanced parameters.

pvanheus: I'll ask the upstream tool authors for some examples. Right now there's not a lot of documentation on using these parameters.

bgruening:
Those tests would be really nice, but @pvanheus if upstream does not have examples I think the complexity is low enough that we could merge it.

tools/nextclade/nextclade.xml
pvanheus: I received some advice from the upstream authors and added a test for using stricter QC.

bgruening (approved):

PR #3632, opened 2021-04-26
Fix bad input name in jbrowse-fromdir

hexylena:
FOR CONTRIBUTOR:

@bgruening noticed that it was referencing the wrong path, leading to an empty variable and reading from /data/ (yikes!). I guess it shows how often this tool is used so maybe we just delete it?


hexylena:
```
cp: can't stat '/data': No such file or directory
```

well good that the test catches it. Bad that we have no way to create a dataset with an extra_files_path I guess. I tried something I saw in velvetg's wrapper but this fails with the same error. I'm not sure how to proceed, suggestions are welcome. Should we just remove the test? @abretaud any opinions?

image

tools/jbrowse/jbrowse-fromdir.xml
```diff
@@ -21,7 +21,7 @@ export JBROWSE_SOURCE_DIR=\$(dirname \$(command -v prepare-refseqs.pl))/../opt/j
 cp -R \${JBROWSE_SOURCE_DIR}/* $output.files_path/ &&
 
 ## Copy in data directory
-cp -R $input.dataset.extra_files_path/data/ $output.files_path/ &&
+cp -R $input.extra_files_path/data/ $output.files_path/ &&
```
bernt-matthias, line 24: Wan you add quotes? Here and everywhere else.

abretaud:
Yep, testing composite data seems tricky https://github.com/galaxyproject/galaxy/pull/9885
I'm not using this tool usually, do you have some usage stats on usegalaxy.*? I guess we can merge the fix anyway

tools/jbrowse/jbrowse-fromdir.xml
```diff
@@ -13,16 +13,16 @@
...
 export JBROWSE_SOURCE_DIR=\$(dirname \$(command -v prepare-refseqs.pl))/../opt/jbrowse  &&
 
 ## Copy in jbrowse
-cp -R \${JBROWSE_SOURCE_DIR}/* $output.files_path/ &&
+cp -R \${JBROWSE_SOURCE_DIR}/* '${output.files_path}/' &&
 
 ## Copy in data directory
-cp -R $input.dataset.extra_files_path/data/ $output.files_path/ &&
-cp $output.files_path/index.html $output
+cp -R '{$input.extra_files_path}/data/' '${output.files_path}/' &&
```
bgruening, line 24:
```suggestion
cp -R '${input.extra_files_path}/data/' '${output.files_path}/' &&
```

hexylena:
> I'm not using this tool usually, do you have some usage stats on usegalaxy.*?

tool | count
--- | ---
jbrowse               |  4650
jbrowse_to_standalone |    86

I don't know. I used to use it to take data from apollo and make a read-only jbrowse I could share, but, we should invest energy in the apollo setup instead probably.

PR #3650, opened 2021-05-04
Update BWA-MEM tool

gallardoalba:
FOR CONTRIBUTOR:


tools/bwa/bwa-mem.xml
```diff
@@ -247,11 +255,23 @@ bwa mem
...
+            <option value="name">Sort by read names  (i.e., the QNAME field) </option>
+            <option value="unsort">Not sorted = sorted as input (probably not reproducible with different cores)</option>
+        </param>
     </inputs>
 
     <outputs>
-        <data format="bam" name="bam_output" label="${tool.name} on ${on_string} (mapped reads in BAM format)">
+        <data format="bam" name="bam_indexed" label="${tool.name} on ${on_string} (mapped reads in BAM format)">
             <expand macro="dbKeyActionsBwaMem" />
+            <filter>output_sort == "coordinate"</filter>
```
bgruening, line 268: Why do you not add the `change_format` to the first output and drop the second?

gallardoalba: Because of this line  <expand macro="dbKeyActionsBwaMem" />. Is it possible to include it inside`change format`?

tools/bwa/bwa-mem.xml
```diff
@@ -247,11 +255,23 @@ bwa mem
                 </conditional>
             </when>
         </conditional>
+        <param name="output_sort" type="select" label="BAM sorting mode">
+            <option value="coordinate" selected="True">Sort by chromosomal coordinates</option>
+            <option value="name">Sort by read names  (i.e., the QNAME field) </option>
+            <option value="unsort">Not sorted = sorted as input (probably not reproducible with different cores)</option>
```
bgruening, line 261: with different number of threads

tools/bwa/bwa-mem.xml
bgruening: Mh, no, but wouldn't you want this also in the other output cases?

tools/bwa/bwa-mem.xml
```diff
@@ -106,7 +106,15 @@ bwa mem
     '${fastq_input.fastq_input1}'
 #end if
 
-| samtools sort -@\${GALAXY_SLOTS:-2} -T "\${TMPDIR:-.}" -O bam -o '$bam_output'
+#if str( $output_sort ) == "coordinate":
+        | samtools sort -@\${GALAXY_SLOTS:-2} -T "\${TMPDIR:-.}" -O bam -o '$bam_indexed'
+#elif str( $output_sort ) == "name":
+        | samtools sort -n -@\${GALAXY_SLOTS:-2} -T "\${TMPDIR:-.}" -O bam -o '$bam_not_indexed'
+#else
+        | samtools view -@ \${GALAXY_SLOTS:-1} -bS - -o '$bam_not_indexed'
```
bgruening, line 114:
```suggestion
        | samtools view -@ \${GALAXY_SLOTS:-2} -bS - -o '$bam_not_indexed'
```

tools/bwa/bwa-mem.xml
gallardoalba: Not, because I think that it is related to indexing, isn't it?

bgruening: I think the macros assigning a dbkey according the the reference genome against they are mapping ... those reference genomes are the referecen indices.

gallardoalba: Then I'll modify it

tools/bwa/bwa-mem.xml
```diff
@@ -247,11 +255,20 @@ bwa mem
...
+            <option value="name">Sort by read names  (i.e., the QNAME field) </option>
+            <option value="unsort">Not sorted = sorted as input (probably not reproducible with different number of threads)</option>
+        </param>
     </inputs>
 
     <outputs>
         <data format="bam" name="bam_output" label="${tool.name} on ${on_string} (mapped reads in BAM format)">
             <expand macro="dbKeyActionsBwaMem" />
+            <change_format>
+                <when input="output_sort" value="name" format="qname_input_sorted.bam" />
```
wm75, line 269: I think I have to admit that I didn't know about this datatype before. What's the difference between this and simply `qname_sorted.bam`?

tools/bwa/bwa-mem.xml
```diff
@@ -247,11 +255,20 @@ bwa mem
                 </conditional>
             </when>
         </conditional>
+        <param name="output_sort" type="select" label="BAM sorting mode">
+            <option value="coordinate" selected="True">Sort by chromosomal coordinates</option>
+            <option value="name">Sort by read names  (i.e., the QNAME field) </option>
+            <option value="unsort">Not sorted = sorted as input (probably not reproducible with different number of threads)</option>
```
wm75, line 261: How desperately is this option needed? Is it really worth compromising on reproducibility?
If the threads are really an issue, could you apply a fixed value for them in the command generated for this option?

It would also be more readable if the value was `unsorted` instead of `unsort`.

wm75:
Don't forget to bump the wrapper version!

wm75:
What has changed in all those test bam files?

tools/bwa/bwa-mem.xml
bgruening: One is really sorted according the qname and the other one is sorted like the input sequeces - so not-sorted.

tools/bwa/bwa-mem.xml
bgruening: In HiC its just the savest to keep the order of reads untouched. I don't know why qname-sorted should not work as well in theory, but what we are trying here is to reimplement the `--reorder` option from bowtie2 http://bowtie-bio.sourceforge.net/manual.shtml

wm75: It would seem that you need to restrict bwa to using a single thread then in this strict mode?

wm75: name-sorted would do for most users with most inputs, but if the input is not even qname_sorted you'd have to use the strictly unsorted (and very slow) option.

tools/bwa/bwa-mem.xml
wm75: Wouldn't it be easier to just reuse the input's format in any case?

wm75: Ah, forget this silly comment!

wm75: So it seems the format should be "qname_sorted.bam" with value="name" (because you know the file has been qname-sorted, so it's not qname- or input-sorted but qname-sorted for sure).
When value is "unsort", format should be "qname_input_sorted" (because you know the output order is that of the input) unless you want to go with the somehow undefined multithreaded behavior.

tools/bwa/bwa-mem.xml
gallardoalba: > It would seem that you need to restrict bwa to using a single thread then in this strict mode?

what do you think @bgruening?

gallardoalba:
> What has changed in all those test bam files?

The change must be due to some random factor, since the `line_diff=2`parameter was previously included.

tools/bwa/bwa-mem.xml
```diff
@@ -247,11 +260,20 @@ bwa mem
...
+            <option value="unsorted">Not sorted (sorted as input)</option>
+        </param>
     </inputs>
 
     <outputs>
         <data format="bam" name="bam_output" label="${tool.name} on ${on_string} (mapped reads in BAM format)">
             <expand macro="dbKeyActionsBwaMem" />
+            <change_format>
+                <when input="output_sort" value="name" format="qname_sorted.bam" />
+                <when input="output_sort" value="unsorted" format="unsorted.bam" />
```
bgruening, line 275: those formats do not exists, isn't it? datatypes needs to be defined here: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/sample/datatypes_conf.xml.sample

So one needs to be `qname_input_sorted.bam` and the other one `qname_sorted.bam`

bgruening: e.g. `description="A binary file compressed in the BGZF format with a '.bam' file extension and sorted based on the aligner output."` I think this is also what bowtie2 is using. Can you please check that.

tools/bwa/bwa-mem.xml
```diff
@@ -14,7 +14,12 @@
 ## Begin BWA-MEM command line
 
 bwa mem
--t "\${GALAXY_SLOTS:-1}"
+
+#if str( $output_sort == "unsorted")
```
wm75, line 18: This will not do what you think it does ;)
`str($output_sort) == 'unsorted'` is what you are looking for. The above would *always* run with 1 thread I guess.

tools/bwa/bwa-mem.xml
gallardoalba: Yes, it is.

tools/bwa/bwa-mem.xml
gallardoalba: Yes, thanks!

tools/bwa/bwa-mem.xml
bgruening: I don't know ... got for 1 thread and see how many poeple complain?

Delphine-L: name sorted is safer i think, when using several datasets, it allows to correct if the forward and reverse datasets have been concatenated in different orders

bgruening: So we go for one thread for the input-sorted option to be save and add a warning into the tool-help section?

gallardoalba: Yes.

gallardoalba: I included the warning @bgruening; it is ready to merge I think.

tools/bwa/bwa-mem.xml
```diff
@@ -247,11 +260,20 @@ bwa mem
                 </conditional>
             </when>
         </conditional>
+        <param name="output_sort" type="select" label="BAM sorting mode" help="The 'Not sorted' option requires using a single thread to perform the conversion from SAM to BAM format, so the runtime is extended.">
```
wm75, line 263:
```suggestion
        <param name="output_sort" type="select" label="BAM sorting mode" help="The 'Not sorted' option can extend the run time of the tool significantly (cause it requires running on only a single thread).">
```

wm75 (approved):

bgruening:
Thanks @gallardoalba!

PR #3636, opened 2021-04-27
show badges for master branch

bernt-matthias:
and use simpler URLs

FOR CONTRIBUTOR:
* [x] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [ ] - License permits unrestricted use (educational + commercial)
* [ ] - This PR adds a new tool or tool collection
* [ ] - This PR updates an existing tool or tool collection
* [x] - This PR does something else (explain below)

PR #3669, opened 2021-05-13
add a job to PR workflow

bernt-matthias:
that indicates success if no tools were changed

@mvdbeek maybe something like this helps?

FOR CONTRIBUTOR:

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -337,14 +337,23 @@ jobs:
         shed-key: ${{ secrets.TS_API_KEY }}
 
   determine-success:
-    name: check workflow success
+    name: Check workflow success
     runs-on: ubuntu-latest
     if: always()
-    needs: [setup, combine_outputs]
+    needs: [setup, lint, flake8, lintr, combine_outputs]
     strategy:
       matrix:
         python-version: ['3.7']
     steps:
-    - name: Indicate failure
+    - name: Indicate tool test failure
       if: needs.setup.outputs.tool-list != '' && needs.combine.result != 'success'
+      run: exit 1
+    - name: Indicate tool lint failure
+      if: needs.lint.result != 'success'
+      run: exit 1
+    - name: Indicate python script lint failure
+      if: needs.flake8.result != 'success'
+      run: exit 1
+    - name: Indicate R script lint failure
+      if: needs.lintr.result != 'success'
       run: exit 1
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -346,5 +346,5 @@ jobs:
         python-version: ['3.7']
     steps:
     - name: Indicate failure
-      if: needs.setup.outputs.tool-list != '' and needs.combine.result != 'success'
+      if: needs.setup.outputs.tool-list != '' && needs.combine.result != 'success'
       run: exit 1
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -84,7 +84,6 @@ jobs:
   lint:
     name: Lint tool-list
     needs: setup
-    if: needs.setup.outputs.repository-list != ''
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -340,7 +340,7 @@ jobs:
     name: check workflow success
     runs-on: ubuntu-latest
     if: always()
-    needs: [setup, combine]
+    needs: [setup, combine_outputs]
     strategy:
       matrix:
         python-version: ['3.7']
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -116,7 +116,7 @@ jobs:
   flake8:
     name: Lint Python scripts
     needs: setup
-    if: needs.setup.outputs.repository-list != ''
+    if: ${{ needs.setup.outputs.repository-list != '' }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


mvdbeek (approved):
Yeah, I think that would help, although I also like being able to just override the checks.
We probably need to merge this to see it in effect.

mvdbeek:
Ah, I think this needs to be a single status check that reflects no tools and successful checks:
Screenshot 2021-05-13 at 16 17 31
I don't think we can specify an `or` here

bernt-matthias:
Maybe something like:

```
success-if-notools:
    if: always()
    needs: [setup, combine]
```

and then implement the logic in a step.. not sure how.

bernt-matthias:
Checked here https://github.com/bernt-matthias/tools-iuc/pull/37 (https://github.com/bernt-matthias/tools-iuc/runs/2577704110) .. with the badges PR

mvdbeek:
Nice!

mvdbeek (approved):

nsoranzo:
Wouldn't adding:
```
    if: needs.setup.outputs.repository-list != ''
```
to the `combine` step be enough?

BTW, I believe `if: needs.setup.outputs.repository-list != ''` should instead be removed from the `lint` step, otherwise it won't check for missing `.shed.yml` file for a new tool.

bernt-matthias:
> Wouldn't adding:
> 
> ```
>     if: needs.setup.outputs.repository-list != ''
> ```
> 
> to the `combine` step be enough?

What if the combine step fails?

> BTW, I believe if: needs.setup.outputs.repository-list != '' should instead be removed from the lint step, otherwise it won't check for missing .shed.yml file for a new tool.

Either this or `needs.setup.outputs.repository-list != '' and needs.setup.outputs.tool-list != ''`. Not sure. We can try to remove it.

nsoranzo:
> > Wouldn't adding:
> > ```
> >     if: needs.setup.outputs.repository-list != ''
> > ```
> >   
> > to the `combine` step be enough?
> 
> What if the combine step fails?

Aren't you trying to workaround a failure when no tools were modified? In which case the combine step should be skipped with the above condition.

> > BTW, I believe if: needs.setup.outputs.repository-list != '' should instead be removed from the lint step, otherwise it won't check for missing .shed.yml file for a new tool.
> 
> Either this or `needs.setup.outputs.repository-list != '' and needs.setup.outputs.tool-list != ''`. Not sure. We can try to remove it.

We weren't skipping the `lint` step before the move to `planemo-ci-action` , so I suppose it should still work.

bernt-matthias:
> Aren't you trying to workaround a failure when no tools were modified? In which case the combine step should be skipped with the above condition.

Not sure if a skipped combine step would count as successful with the settings outlined by @mvdbeek https://github.com/galaxyproject/tools-iuc/pull/3669#issuecomment-840593344

.github/workflows/pr.yaml
```diff
@@ -335,3 +334,25 @@ jobs:
...
         shed-key: ${{ secrets.TS_API_KEY }}
+
+  determine-success:
+    name: Check workflow success
+    runs-on: ubuntu-latest
+    if: always()
+    needs: [setup, lint, flake8, lintr, combine_outputs]
+    strategy:
+      matrix:
+        python-version: ['3.7']
```
nsoranzo, line 345:
```suggestion
```

nsoranzo:
> > Aren't you trying to workaround a failure when no tools were modified? In which case the combine step should be skipped with the above condition.
> 
> Not sure if a skipped combine step would count as successful with the settings outlined by @mvdbeek [#3669 (comment)](https://github.com/galaxyproject/tools-iuc/pull/3669#issuecomment-840593344)

It does, I've just played around with those settings and requiring the "Deploy" step (which is always skipped on PRs) still allows merging.

I've now changed the settings to require "Lint tool-list (3.7)", "Lint Python scripts (3.7)", "Lint R scripts (ubuntu-20.04, 4.0.1)" and "Combine chunked test results (3.7)" (which can still force the merge as admins).

mvdbeek:
I'm a bit confused by this.

> It does, I've just played around with those settings and requiring the "Deploy" step (which is always skipped on PRs) still allows merging.

That was the starting point, and it definitely did not allow us merging a PR that skipped a required test without changing the settings allowing admins to merge. Which we should avoid, not all committers are or should be repo admins.

So you're saying that this is now not necessary anymore ?

nsoranzo:
> I'm a bit confused by this.
> 
> > It does, I've just played around with those settings and requiring the "Deploy" step (which is always skipped on PRs) still allows merging.
> 
> That was the starting point, and it definitely did not allow us merging a PR that skipped a required test without changing the settings allowing admins to merge. Which we should avoid, not all committers are or should be repo admins.
> 
> So you're saying that this is now not necessary any more ?

The starting point was not the "Deploy" step, it was the "Combine chunked test results (3.7)" , xref. https://github.com/galaxyproject/tools-iuc/pull/3669#issuecomment-840593344

nsoranzo:
OK, "interestingly" the job name changes by losing the matrix part if the job is skipped, see e.g. the difference between "Lint tool-list (3.7)" (executed) and "Lint Python scripts" (skipped) in the box below.
This discrepancy leaves the status check waiting for non-existing jobs in both following situations:
- if no tools were changed (as in this PR) and the status check configuration uses job names *with* the matrix part
- if tools were changed in a PR and the status check configuration uses job names *without* the matrix part

So I had to restore the `determine-success` job, with a couple of fixes. I also found such a solution proposed in https://github.community/t/status-check-for-a-matrix-jobs/127354

nsoranzo (approved):

bernt-matthias:
Looks good to me. Thanks for the feedback @nsoranzo and @mvdbeek

bernt-matthias:
Seems to work: https://github.com/galaxyproject/tools-iuc/pull/3636

One could think about running `lint` only `if: ${{ needs.setup.outputs.repository-list != '' && needs.setup.outputs.tool-list != '' }}`

PR #3676, opened 2021-05-17
expose filename_mapping of reads in ont_fast5_api wrapper [WIP]

drosofff:
FOR CONTRIBUTOR:

- A file containing the filename mapping of reads is generated by two of the wrapped tools. This is useful, for instance when you want to subset reads, because you can deduced the desired subset from the filename read mapping. This PR expose the file name mapping as an addition output dataset.

- Fixes an issue with an input macro (compression) which was not properly modified after call in the tool wrappers.
- Fixes the desired format for the list of reads to be subsetted, which can now be either tabular or txt.
- Added the appropriate tests.

### Remaining issue:
the filename mapping of reads is outputed even with tools that do not normally do it (compress). This results in an empty dataset (green through). This is fixable but would need not to centralise output code in a shared macro.

### DIFF
--- a/tools/ont_fast5_api/compress_fast5.xml
+++ b/tools/ont_fast5_api/compress_fast5.xml
@@ -25,17 +25,16 @@ compress_fast5
     ]]></command>
     <inputs>
         <expand macro="input"/>
-        <expand macro="compression">
-            <option value="vbz" selected="true">VBZ</option>
-        </expand>
+        <expand macro="compression"/>
     </inputs>
     <outputs>
         <expand macro="output"/>
     </outputs>
     <tests>
         <!-- #1 default -->
-        <test expect_num_outputs="1">
+        <test expect_num_outputs="2">
             <param name="input_path" value="multi.fast5.tar"/>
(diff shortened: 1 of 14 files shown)


bgruening:
@mmiladi can you take a look?

@drosofff are you planning to work on the remaining issue?

drosofff:
@bgruening Yes I am

drosofff:
@mmiladi @bgruening I have a question about formats:

Overall the compression token in all commands leads to a tar of none/gz/bz2 fast5 files. Thus formally your initial proposition to produce only fast5.tar is not completely satisfying since the state of the fast5 files in the tar archive is not exposed.

In the contrary, I took the option to say that a compression leads to fast5.tar.gz/bz2 which is not exactly true as it should be fast5.gz/bz2.tar. However, the datatypes fast5.gz.tar and fast5.bz2.tar do not exist.

What would you do to disentangle this ?

drosofff:
> @drosofff are you planning to work on the remaining issue?

Done !

PR #3700, opened 2021-05-27
Add bellerophon

astrovsky01:
FOR CONTRIBUTOR:


davebx (approved):

tools/bellerophon/bellerophon.xml
```diff
@@ -0,0 +1,40 @@
...
+    <command detect_errors="exit_code"><![CDATA[
+        bellerophon
+        --forward '${forward}'
+        --reverse '${reverse}'
+        --quality $quality
+        --output 'merged_out.bam'
+        ]]>
+    </command>
+    <inputs>
+        <param argument="--forward" type="data" format="sam" label="First set of reads" help="This is usually the forward reads in your experiment." />
```
bgruening, line 19: you might want to look at https://github.com/galaxyproject/tools-iuc/pull/3649/files#diff-5adabb44b4619b33f8bde84a20008f43bfbc8afd7156daf614d4047ec76a6430R29

Dave put a lot of effort into supporting bam files, so we should add this I think?

astrovsky01: Fully a typo

tools/bellerophon/bellerophon.xml
```diff
@@ -5,23 +5,26 @@
...
+        --reverse reverse.bam
         --quality $quality
         --output 'merged_out.bam'
+        && samtools sort --no-PG -O BAM -o '$outfile' -@ \${GALAXY_SLOTS:-1} merged_out.bam
         ]]>
     </command>
     <inputs>
-        <param argument="--forward" type="data" format="bam,sam,cram" label="First set of reads" help="This is usually the forward reads in your experiment." />
-        <param argument="--reverse" type="data" format="bam,sam,cram" label="Second set of reads" help="This is usually the reverse reads in your experiment." />
+        <param argument="--forward" type="data" format="qname_sorted.bam" label="First set of reads" help="This is usually the forward reads in your experiment." />
```
bgruening, line 22: SAM should stay in addition.

davebx: Presumably to skip the overhead of converting a sam file to bam? Because I would argue in favor of this change, so that we can at least be certain the inputs are always going to be sorted correctly.

bgruening: SAM files are a way to keep them "unsorted" that's why there are a few mappers that only output SAM. So I would say SAM as input means unsortd. So save.

davebx: That's a fair point, I would now also recommend leaving sam as an accepted input format.

tools/bellerophon/bellerophon.xml
```diff
@@ -0,0 +1,44 @@
...
+        && samtools sort --no-PG -O BAM -o '$outfile' -@ \${GALAXY_SLOTS:-1} merged_out.bam
+        ]]>
+    </command>
+    <inputs>
+        <param argument="--forward" type="data" format="qname_sorted.bam" label="First set of reads" help="This is usually the forward reads in your experiment." />
+        <param argument="--reverse" type="data" format="qname_sorted.bam" label="Second set of reads" help="This is usually the reverse reads in your experiment." />
+        <param argument="--quality" type="integer" value="20" label="Minimum mapping quality"/>
+    </inputs>
+    <outputs>
+        <data name="outfile" label="${tool.name} on ${on_string}" format="bam"/>
```
bernt-matthias, line 28: Label could be removed since it's the default

bernt-matthias (approved):

tools/bellerophon/bellerophon.xml
astrovsky01: 👍

bernt-matthias:
The error was for the version that I got from @astrovsky01 by mail. Works for the latest version.

tools/bellerophon/bellerophon.xml
```diff
@@ -0,0 +1,44 @@
...
+        <param argument="--forward" type="data" format="qname_sorted.bam,qname_sorted.bsam" label="First set of reads" help="This is usually the forward reads in your experiment." />
+        <param argument="--reverse" type="data" format="qname_sorted.bam,qname_sorted.bsam" label="Second set of reads" help="This is usually the reverse reads in your experiment." />
+        <param argument="--quality" type="integer" value="20" label="Minimum mapping quality"/>
+    </inputs>
+    <outputs>
+        <data name="outfile" format="bam"/>
+    </outputs>
+    <tests>
+        <test expect_num_outputs="1">
+            <param name="forward" value="forward.bam" />
```
bernt-matthias, line 32: Set `ftype`. Otherwise converter will run.

tools/bellerophon/bellerophon.xml
```diff
@@ -0,0 +1,44 @@
+<tool id="bellerophon" name="Filter and merge" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="20.01">
+    <description>chimeric reads with bellerophon.py</description>
```
bernt-matthias, line 2: Remove `.py`?

tools/bellerophon/bellerophon.xml
bernt-matthias: In this case I would suggest to make the tool wrapper do the conversion to sam (if bellerophon does not support it natively). Argument is the same as for the fastq/fastq.gz efforts (https://github.com/galaxyproject/tools-iuc/issues/2312)

tools/bellerophon/bellerophon.xml
```diff
@@ -0,0 +1,44 @@
...
+    <macros>
+        <token name="@TOOL_VERSION@">1.0</token>
+        <token name="@VERSION_SUFFIX@">0</token>
+    </macros>
+    <requirements>
+        <requirement type="package" version="@TOOL_VERSION@">bellerophon</requirement>
+        <requirement type="package" version="1.12">samtools</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        ln -s '${forward}' forward_input &&
```
bgruening, line 12: we have no test for SAM files, are we sure pysam is smart enough to recognize SAM and BAM?

tools/bellerophon/bellerophon.xml
astrovsky01: I wanted the auto-converter for bam to run. It will sort if not sorted. The data is sorted, but I wanted that to run.

tools/bellerophon/bellerophon.xml
astrovsky01: Specific because the tool was originally a perl script that was converted to python by @davebx

tools/bellerophon/bellerophon.xml
astrovsky01: Added

tools/bellerophon/bellerophon.xml
davebx: I wasn't sure about it myself, but there is also a mothur tool involving bellerophon.

tools/bellerophon/bellerophon.xml
davebx: > are we sure pysam is smart enough to recognize SAM and BAM?

Yeah, that was one of the first things I checked when I was writing this, we're good.

davebx:
Looks good to me. @bgruening?

bgruening (approved):

bgruening:
Yes, very cool!

bernt-matthias:
What I do not understand here is why one can run this tool with `unsorted.bam` as input (without a converter running). This seems to be due to allowing `sam`. Might be a Galaxy bug?

Also supporting `qname_sorted.bam` (sorted in some way) and `sam` (unsorted) seems unintuitive to me. Why not just supporting `sam` and `bam`?

bernt-matthias:
> What I do not understand here is why one can run this tool with unsorted.bam as input (without a converter running).

Have to be more precise... I'm still working on the old version that I got per mail. This only accepts `sam` and the test does not set the `ftype` for the input. If I'm running `planemo serve` and upload the test files they are autodetected as `unsorted.bam` .. so I'm assuming that this also happens in `planemo test`. But I do not see a converter running in the test.

bgruening:
> Also supporting qname_sorted.bam (sorted in some way) and sam (unsorted) seems unintuitive to me. Why not just supporting sam and bam?

This tool deals with HiC data. You have PE data but they are treaded as single end during mapping etc. Important for many downstream tools is that the position of the read-pair is the same. This is because tool iterate over both tools and assume a pair on position one in both fles, and the second pair in position two etc ...
SAM files does this by default with the main mapping tools that we have. They process reads as they are in the FASTQ and FASTQ has the pairs in both files on the same positions. `qname_sorted.bam` is the equivalent to that in BAM. Compressed bu sorted in a way that you can easily map over it with `zip(file1, file2)`.

bernt-matthias:
Thanks for the explanation @bgruening. Sounds reasonable. 

For the other point see https://github.com/galaxyproject/galaxy/pull/12073

PR #3732, opened 2021-06-09
Change Seurat help text

astrovsky01:
We no longer allow RDS output, so I am removing that to avoid confusion

FOR CONTRIBUTOR:

### DIFF
--- a/tools/seurat/seurat.xml
+++ b/tools/seurat/seurat.xml
@@ -164,7 +164,6 @@ types of single cell data. See the `Seurat Guided Clustering tutorial`_ for more
 
 Optionally you can choose to output
 
-    * Seurat RDS object (can use within R)
     * Rscript
 
 .. _Seurat: https://www.nature.com/articles/nbt.4096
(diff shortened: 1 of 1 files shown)


tools/seurat/seurat.xml
```diff
@@ -164,7 +164,6 @@ types of single cell data. See the `Seurat Guided Clustering tutorial`_ for more
 
 Optionally you can choose to output
 
-    * Seurat RDS object (can use within R)
     * Rscript
```
nsoranzo, line 167: We don't produce the Rscript output any more as well, right?

astrovsky01: We do, it's an option to print alongside the graphs instead of separate

nsoranzo: You mean the "Show code alongside outputs?" option? Then I'd rephrase here, e.g. "R code to generate each output".

astrovsky01: Better?

nsoranzo (approved):

PR #3701, opened 2021-05-28
Update minimap2 v.2.20

gallardoalba:
FOR CONTRIBUTOR:
* [x] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [x] - License permits unrestricted use (educational + commercial)
* [ ] - This PR adds a new tool or tool collection
* [x] - This PR updates an existing tool or tool collection
* [ ] - This PR does something else (explain below)

This PR updates minimap2 to version 2.20.

PR #3736, opened 2021-06-11
Add header with samtools view for alignments produced via minimap2

### LINKED ISSUE #3734: minimap2 multi-part (missing SAM header error)
Hi all!
I am using minimap2 (Galaxy Version 2.17+galaxy3) for my pacbio ccs reads on a big Onion genome (~16GB). It runs into the error that is mentioned here: https://github.com/lh3/minimap2/issues/171 e.g. missing SAM header. I could solve this in commandline by adding samtools view -T reference.fasta, but I figured this is no option in your galaxy tool. 
I would like to make this available for my non-bioinformatician colleagues in Galaxy. 
Let me know if you have any suggestions :)
Cheers!
Annabel Dekker

mvdbeek:
Fixes #3734. The
problem is that minimap2 splits indexes above a certain basecount
(4GB by default, configurable with the -I parameter). If that
happens the SQ header lines won't be written, so we add them here.

Not sure about the performance impact of the additional samtools view
step ... it's probably minimal since it's sam in, sam out ?

Otherwise we could do something like
`(samtools view -H new_header.sam; minimap <options>) | samtool
sort` but getting a SAM header out of a fasta file would require
a custom script, it seems.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/minimap2/macros.xml
+++ b/tools/minimap2/macros.xml
@@ -1,6 +1,6 @@
 <macros>
     <token name="@TOOL_VERSION@">2.20</token>
-    <token name="@GALAXY_TOOL_VERSION@">galaxy0</token>
+    <token name="@GALAXY_TOOL_VERSION@">galaxy1</token>
     <xml name="edam_ontology">
         <edam_topics>                                                                                  
             <edam_topic>topic_0102</edam_topic>
--- a/tools/minimap2/macros.xml
+++ b/tools/minimap2/macros.xml
@@ -24,17 +24,15 @@
     </xml>
     <xml name="citations">
         <citations>
-            <citations>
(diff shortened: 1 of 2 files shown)


wm75:
Seems the tests are failing as a fallout of #3701 - https://github.com/galaxyproject/tools-iuc/pull/3701#discussion_r650072975

wm75:
Seems planemo linting learnt to catch this only now.

mvdbeek:
> Seems planemo linting learnt to catch this only now.

planemo-ci-action didn't use pipefail when it should have, we fixed that now but we should run this on all of the IUC to make sure we're still good.

wm75 (approved):
Nice! I cannot imagine that piping through samtools view would have any measurable influence on performance.

Issue #3488, opened 2021-02-17
Add citations to MAKER.


tnabtaf:
MAKER has several publications.  Add at least one to the tool. This is a suggested Paper-Cut issue

PR #3742, opened 2021-06-18
Maker: clarify unmask option

abretaud:
FOR CONTRIBUTOR:

As suggested in GGA gitter channel, rephrasing the unmask option: by running on "unmasked sequence" the Maker authors probably meant "on sequences that have not been masked" = "ignoring masking information" = "both masked and unmasked if masking was performed"
What do you think @AnnaSyme?

Also added 2 citations for #3488


AnnaSyme:
thanks for looking into this @abretaud - I'm still a bit confused! So do you think that "Run ab-initio prediction programs on unmasked sequence" means to firstly unmask the sequence (if it has been masked), then run predictions?

abretaud:
I think when enabling this option maker will just ignore the masking information = consider the genome to be completely unmasked
The phrasing os the option is very misleading, but that's what I understand from the author explanations

AnnaSyme:
Ah right, that makes sense too as it is off by default. Thank you!

bgruening (approved):

bgruening (approved):

PR #3753, opened 2021-06-23
Update mothur fetcher

rhpvorderman:
FOR CONTRIBUTOR:

This updates the data manager that fetches mothur reference datasets.

- Updated the tool to python3 with `2to3`.  Move to python 3.8. According to the python documentation moving files has had some optimizations in 3.8.
- ~Added a container requirement to circumvent any issues that arise from biocontainers base-container shennannigans (like: https://github.com/galaxyproject/galaxy/issues/12184)~
- Add the latest mothur references from the website
- Fix a bug where `os.rename` was used to move a file. `os.rename` only works for files on the same physical disk, and this is not necessarily the case on cluster filesystems. Use `shutil.move` instead, which always works.
- Fix an issue where the entire file would be read into memory during the download. This is now done in chunks.

A test has been added for one of the new reference. The SILVA reference was too big, but it has been tested on our testgalaxy instance and confirmed to be working.

### DIFF
--- a/data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
+++ b/data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
@@ -2,7 +2,7 @@
 <tool id="data_manager_fetch_mothur_reference_data" name="Fetch Mothur toolsuite reference data" version="0.1.5" tool_type="manage_data" profile="19.05">
     <description>Fetch and install reference data for Mothur</description>
     <requirements>
-        <requirement type="package" version="2.7">python</requirement>
+        <requirement type="package" version="3.7">python</requirement>
     </requirements>
     <command><![CDATA[
         python '$__tool_directory__/fetch_mothur_reference_data.py'
--- a/data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
+++ b/data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<tool id="data_manager_fetch_mothur_reference_data" name="Fetch Mothur toolsuite reference data" version="0.2.0" tool_type="manage_data" profile="19.05">
+<tool id="data_manager_fetch_mothur_reference_data" name="Fetch Mothur toolsuite reference data" version="0.2.1" tool_type="manage_data" profile="19.05">
     <description>Fetch and install reference data for Mothur</description>
     <requirements>
         <requirement type="package" version="3.8">python</requirement>
--- a/data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
+++ b/data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
@@ -2,7 +2,7 @@
 <tool id="data_manager_fetch_mothur_reference_data" name="Fetch Mothur toolsuite reference data" version="0.1.5" tool_type="manage_data" profile="19.05">
     <description>Fetch and install reference data for Mothur</description>
     <requirements>
-        <requirement type="package" version="3.7">python</requirement>
+        <requirement type="package" version="3.8">python</requirement>
     </requirements>
     <command><![CDATA[
         python '$__tool_directory__/fetch_mothur_reference_data.py'
--- a/data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
+++ b/data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
@@ -35,12 +35,14 @@
                     <option value="lookup_titanium">GS FLX Titanium lookup files</option>
                     <option value="lookup_gsflx">GSFLX lookup files</option>
                     <option value="lookup_gs20">GS20 lookup files</option>
+                    <option value="RDP_v18">RDP reference files (training set version 18)</option>
                     <option value="RDP_v16">RDP reference files (training set version 16)</option>
                     <option value="RDP_v14">RDP reference files (training set version 14)</option>
                     <option value="RDP_v10">RDP reference files (training set version 10)</option>
                     <option value="RDP_v9">RDP reference files (training set version 9)</option>
                     <option value="RDP_v7">RDP reference files (training set version 7)</option>
                     <option value="RDP_v6">RDP reference files (training set version 6)</option>
+                    <option value="silva_release_138.1">SILVA reference files (release 138.1)</option>
                     <option value="silva_release_128">SILVA reference files (release 128)</option>
                     <option value="silva_release_123">SILVA reference files (release 123)</option>
                     <option value="silva_release_119">SILVA reference files (release 119)</option>
--- a/data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
+++ b/data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
@@ -3,6 +3,10 @@
     <description>Fetch and install reference data for Mothur</description>
     <requirements>
         <requirement type="package" version="3.8">python</requirement>
+        <!--Since we are downloading data it is more important to have an
+         up-to-date container with the latest CA certificates than a bit-by-bit
+         reproducible image.-->
+        <container type="docker">python:3.8-slim</container>
     </requirements>
     <command><![CDATA[
         python '$__tool_directory__/fetch_mothur_reference_data.py'
--- a/data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
+++ b/data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
@@ -3,10 +3,6 @@
     <description>Fetch and install reference data for Mothur</description>
     <requirements>
         <requirement type="package" version="3.8">python</requirement>
-        <!--Since we are downloading data it is more important to have an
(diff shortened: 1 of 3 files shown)


data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
```diff
@@ -1,8 +1,12 @@
...
-<tool id="data_manager_fetch_mothur_reference_data" name="Fetch Mothur toolsuite reference data" version="0.1.5" tool_type="manage_data" profile="19.05">
+<tool id="data_manager_fetch_mothur_reference_data" name="Fetch Mothur toolsuite reference data" version="0.2.1" tool_type="manage_data" profile="19.05">
     <description>Fetch and install reference data for Mothur</description>
     <requirements>
-        <requirement type="package" version="2.7">python</requirement>
+        <requirement type="package" version="3.8">python</requirement>
+        <!--Since we are downloading data it is more important to have an
+         up-to-date container with the latest CA certificates than a bit-by-bit
+         reproducible image.-->
+        <container type="docker">python:3.8-slim</container>
```
mvdbeek, line 9:
```suggestion
```
That is not necessary and not recommended.

data_managers/data_manager_mothur_toolsuite/data_manager/fetch_mothur_reference_data.py
```diff
@@ -230,14 +246,14 @@ def download_file(url, target=None, wd=None):
...
+    print(("Downloading %s" % url))
     if not target:
         target = os.path.basename(url)
     if wd:
         target = os.path.join(wd, target)
-    print("Saving to %s" % target)
+    print(("Saving to %s" % target))
     with open(target, 'wb') as fh:
-        fh.write(urllib2.urlopen(url).read())
+        fh.write(urllib.request.urlopen(url).read())
```
mvdbeek, line 256: That's been a problem before, but `.read()` will read the entire file into memory. Any chance you can read the file in chunks ? It's also good practice to set a timeout with the timeout argument to urlopen.

mvdbeek (commented):
Thanks for working on this!

data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
rhpvorderman: Using the biocontainer automated builded image gives problems. Unless these images are updated again. I rather ensure this tool keeps working rather than that the tool fail because a problem created elsewhere. 
This is a fetching tool. Presumably the toolchain does not have influence on the bit-by-bit reproducibility of downloading a link.

Do you have a better suggestion for avoiding issues like https://github.com/galaxyproject/galaxy/issues/12184?

data_managers/data_manager_mothur_toolsuite/data_manager/fetch_mothur_reference_data.py
rhpvorderman: I will take a look at this. It should not be that hard. Just reading in `io.DEFAULT_BUFFER_SIZE` chunks will do.

rhpvorderman: Should be fixed now. I tested the download_file function on a file larger than io.DEFAULT_BUFFER_SIZE and it had the correct checksum.

data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
rhpvorderman: For clarity, I am grateful you were able to fix said issue. But then it turned out the tool itself had some errors, resulting in this work and me porting it to python3. I ran into exactly the same issue for the 3.8 container, again.

Biocontainers is unfortunately completely unreliable when you rely on more than just the tool you use (it works great in 99% of the cases where you just use the tool). This is not the first time an issue like this crops up. Since in this case we rely on downloading I rather use something separate from biocontainers,

mvdbeek: Can you provide evidence of this? This is common infrastructure, we are not going to change this. The problem you encountered is with singularity. All test are run weekly I docker and work correctly.

data_managers/data_manager_mothur_toolsuite/data_manager/fetch_mothur_reference_data.py
```diff
@@ -230,14 +247,19 @@ def download_file(url, target=None, wd=None):
     Returns the name that the file is saved with.
 
     """
-    print("Downloading %s" % url)
+    print(("Downloading %s" % url))
```
mvdbeek, line 250: Can you please remove the extraneous brackets here?

rhpvorderman: I am now in the proces of moving everything to f-strings. That looks much better anyway.

data_managers/data_manager_mothur_toolsuite/data_manager/data_manager_fetch_mothur_reference_data.xml
rhpvorderman: Like I stated it works great when you only call the executable you are interested in. For example `samtools` or `bwa`. 

But things differ when you rely on OS operations. In this case the singularity image does not work because of an issue with DNS. We have had other problems as well. All biocontainers rely on busybox for their gnu tools (cat, ls), but busybox was never intended for use with large bioinformatics files. Instead it is intended for use in embedded systems. Therefore it *is compiled 32-bit* to keep the binary small. That means it largest unsigned integers are 4GB. `wc -c` does not give the expected result on big bionformatic files.

Due to its reliance on non-standard binaries (busybox) on a non-standard OS (alpine), scripts in the containers will sometimes behave differently unexpectedly. Also relying on the OS layer might sometimes lead to unexpected errors (as in this particular issue where a common file is missing). 

So in any use case where you do more than execute a binary on a mounted file, a biocontainer might fail in unexpected ways. Since this is a downloader tool, this does not fit a biocontainers ideal use case.

> This is common infrastructure, we are not going to change this without serious reason to do so.

I am not happy with this either. Frankly I am baffled that the biocontainers team did all this extra effort to make a less reliable base container in an effort to save 25 MB. Why not use debian instead? Sure it uses 25 MB extra, but you never have problems because every production container under the sun uses Debian (postgres, nginx, python etc.). 25 MB is literally nothing compared to the amounts of data we are dealing with on a daily basis in bioinformatics. In any case that is an issue I should raise at biocontainers.

I consider "breaking a tool on singularity" a serious reason. But I also rather rely on common infrastructure. I am affraid there is no ideal solution here :confused: . So the best way forward would be to delete the container stuff in the xml and update the 3.8 python container upstream?

mvdbeek: > But things differ when you rely on OS operations. In this case the singularity image does not work because of an issue with DNS. We have had other problems as well. All biocontainers rely on busybox for their gnu tools (cat, ls), but busybox was never intended for use with large bioinformatics files. Instead it is intended for use in embedded systems. Therefore it _is compiled 32-bit_ to keep the binary small. That means it largest unsigned integers are 4GB. `wc -c` does not give the expected result on big bionformatic files.

Can you mention this on the bioconda github repo ? That is indeed a serious problem.

> Frankly I am baffled that the biocontainers team did all this extra effort to make a less reliable base container in an effort to save 25 MB.

I agree with you, certainly from the docker and kubernetes perspective it doesn't seem like a smart choice, since you're going to re-use the base layer anyway, so whatever the extra size, that is a one-time cost.

> I consider "breaking a tool on singularity" a serious reason.

You can work around this by copying a host /etc/resolv.conf, but from my perspective it's not the container that is breaking singularity, it's singularity breaking the container. If you claim compatibility with docker you should make sure even these corner cases work IMO. Granted I don't know if docker works because they're doing non-standard / non-spec compliant things, in which case the story is more complicated.

> So the best way forward would be to delete the container stuff in the xml and update the 3.8 python container upstream

Yeah, I prefer that. It also means we can simply update the upstream container again if we encounter another issue. The resolver takes into account the container build numbers (the `-N` part), so there's also a trace of that for reproducibility.

mvdbeek: > Can you mention this on the bioconda github repo ? That is indeed a serious problem.

Actually is that still the case for quay.io/bioconda/base-glibc-busybox-bash ?

rhpvorderman: > Actually is that still the case for quay.io/bioconda/base-glibc-busybox-bash ?

@mvdbeek I just checked it. No, it is not a problem anymore. I collleague of mine reported this issue earlier and apparently it has been fixed :tada:. 

Still the point is not that this issues were there, but it is that these sort of issues keep popping up. There are a multitude of issues out there where the base image breaks something and the extended base is needed. It is causing a lot of headaches unnecessarily in my opinion. The whole point of automated infrastructure is that you don't want to hassle with these things manually. 

> I agree with you, certainly from the docker and kubernetes perspective it doesn't seem like a smart choice, since you're going to re-use the base layer anyway, so whatever the extra size, that is a one-time cost.

Indeed, the issue is only from a singularity perspective. And coming from the singularity side, I'd rather deal with 25 MB than with non-functional containers due to the biocontainers base image being a corner case (and thus not being part of the singularity tests).
Anyway I will raise this issue in the appropriate channels.

> Yeah, I prefer that. It also means we can simply update the upstream container again if we encounter another issue. The resolver takes into account the container build numbers (the -N part), so there's also a trace of that for reproducibility.

Thanks. Will do!

rhpvorderman:
@mvdbeek Thanks for the review! I have addressed the changes.

mvdbeek (approved):

rhpvorderman:
Thanks, it is always a pleasure to work with you.

PR #3754, opened 2021-06-24
Cutadapt: remove duplicated cut option

mblue9:
FOR CONTRIBUTOR:

An eagle-eyed user noticed that the cut option is present twice. It was in the tool form listed as the short form -u (as -U is the option for Read 2) and it was recently added again as --cut. This PR removes the --cut option and adds `(--cut)` to the help for the -u option to clarify it is the --cut option.

### DIFF
--- a/tools/cutadapt/macros.xml
+++ b/tools/cutadapt/macros.xml
@@ -1,6 +1,6 @@
 <macros>
     <token name="@TOOL_VERSION@">3.4</token>
-    <token name="@GALAXY_TOOL_VERSION@">galaxy</token>
+    <token name="@GALAXY_TOOL_VERSION@">galaxy0</token>
     <xml name="edam_ontology">
         <edam_topics>                                                                                  
             <edam_topic>topic_0632</edam_topic>
--- a/tools/cutadapt/macros.xml
+++ b/tools/cutadapt/macros.xml
@@ -265,7 +265,7 @@
                     </conditional>
                 </repeat>
 
-                <param name="cut" argument="-u" type="integer" value="0" optional="True" label="Cut bases from reads before adapter trimming" help="Remove bases from each read (first read only if paired). If positive, remove bases from the beginning. If negative, remove bases from the end. This is applied *before* adapter trimming. (--cut)" />
+                <param name="cut" argument="-u/--cut" type="integer" value="0" optional="True" label="Cut bases from reads before adapter trimming" help="Remove bases from each read (first read only if paired). If positive, remove bases from the beginning. If negative, remove bases from the end. This is applied *before* adapter trimming. (--cut)" />
 
             </section>
 
--- a/tools/cutadapt/macros.xml
+++ b/tools/cutadapt/macros.xml
@@ -1,6 +1,6 @@
 <macros>
     <token name="@TOOL_VERSION@">3.4</token>
-    <token name="@GALAXY_TOOL_VERSION@">galaxy0</token>
+    <token name="@GALAXY_TOOL_VERSION@">galaxy</token>
     <xml name="edam_ontology">
         <edam_topics>                                                                                  
             <edam_topic>topic_0632</edam_topic>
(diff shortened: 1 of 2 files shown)


mvdbeek (approved):
Thanks, that looks good! Can you update the GALAXY_TOOL_VERSION macro ?

mblue9:
Ok great, I wasn't sure whether to update the wrapper version as it's a more of a tool form/bug fix but done now. Thanks!

tools/cutadapt/macros.xml
```diff
@@ -265,7 +265,7 @@
                     </conditional>
                 </repeat>
 
-                <param name="cut" argument="-u" type="integer" value="0" optional="True" label="Cut bases from reads before adapter trimming" help="Remove bases from each read (first read only if paired). If positive, remove bases from the beginning. If negative, remove bases from the end. This is applied *before* adapter trimming." />
+                <param name="cut" argument="-u" type="integer" value="0" optional="True" label="Cut bases from reads before adapter trimming" help="Remove bases from each read (first read only if paired). If positive, remove bases from the beginning. If negative, remove bases from the end. This is applied *before* adapter trimming. (--cut)" />
```
mvdbeek, line 268: Sorry, upon closer inspection, are these actually the same parameters ? It seems this is `-u`, as indicated by `argument="-u"`, while the parameter you removed is the `--cut` parameter.

mblue9: It's confusing --cut is the long name, -u is the short name, see here https://cutadapt.readthedocs.io/en/stable/guide.html#trimming-paired-end-reads

bernt-matthias (changes requested):
The change seems good to me, but currently the `cut` parameter is not set on the command line. Also it would be nice to use the section in the test definitions.

mblue9:
>The change seems good to me, but currently the cut parameter is not set on the command line. 

Thanks for having a look. Isn't it here in the command line as `-u` ? https://github.com/galaxyproject/tools-iuc/blob/2b4ce00ae6554d456850182331d2ad3a073b4188/tools/cutadapt/macros.xml#L66

>Also it would be nice to use the section in the test definitions.

Happy to do that. I think it'll mean adding the r1 or r2 section into all 27 tests so just to check before I do that, is this what you mean?  https://github.com/galaxyproject/tools-iuc/pull/3754/commits/c842f2cdb7ab2f8402bb1cedb759b21ed1bffdc3

bernt-matthias (approved):
> Happy to do that. 

Yep, that's what I meant.

bernt-matthias:
@mvdbeek is this one good to go? looks good to me.

tools/cutadapt/macros.xml
mvdbeek:
```suggestion
                <param name="cut" argument="-u" type="integer" value="0" optional="True" label="Cut bases from reads before adapter trimming" help="Remove bases from each read (first read only if paired). If positive, remove bases from the beginning. If negative, remove bases from the end. This is applied *before* adapter trimming (which is controller by --cut)." />
```
is that how it should be ? If you use `argument="-u"` the `-u` will be appended to the end of the help text, so that might be confusing.

mvdbeek: Or should this be
```suggestion
                <param name="cut" argument="--cut" type="integer" value="0" optional="True" label="Cut bases from reads before adapter trimming" help="Remove bases from each read (first read only if paired). If positive, remove bases from the beginning. If negative, remove bases from the end. This is applied *before* adapter trimming. (--cut)" />
```

mvdbeek: Or 
```suggestion
                <param name="cut" argument="-u/--cut" type="integer" value="0" optional="True" label="Cut bases from reads before adapter trimming" help="Remove bases from each read (first read only if paired). If positive, remove bases from the beginning. If negative, remove bases from the end. This is applied *before* adapter trimming. (--cut)" />
```
? 😆

mblue9: 😄 I like the last one the best as it has -u which hopefully makes it obvious that -U is the corresponding option for read 2. But just noticed it looks like this so need to remove one `--cut`
>Remove bases from each read (first read only if paired). If positive, remove bases from the beginning. If negative, remove bases from the end. This is applied *before* adapter trimming. (--cut) (-u/--cut)

bgruening (approved):

PR #3764, opened 2021-06-29
Update RGCCA to 3.0.0, debug Galaxy ui optional files

ecamenen:
* Added an update to the R-package tool (major changes).  WIP for Conda update (https://github.com/bioconda/bioconda-recipes/pull/29272). When it will be accepted, I will make another commit for the Galaxy automatic tests.
* For the current version of Galaxy, the tests pass ... but the user interface does not work with the default options! There are bugs with the optional output files that need to be catched (next commit; done in local).


ecamenen:
Hello @bgruening! Thanks a lot for the bioconda PR!
One question: I was checking the conda with the command ``conda create -n conda-rgcca rgccacmd``, but I have the feeling that it downloads the previous version, not the latest one: bioconda/noarch::rgccacmd-3.0.0-r40h9ee0642_1. 
The last one starts with r41 and not 40 ... is there a parsing priority problem? 
noarch/rgccacmd-3.0.0-r41h9ee0642_2.tar.bz2 (https://anaconda.org/bioconda/rgccacmd/files). Maybe I should have changed the conda version to 3.0.1 to solve this problem like I hesitated?
I can try to commit to see what happens but I m afraid it ll not work.
Thanks in advance for your help :)

bgruening:
> Maybe I should have changed the conda version to 3.0.1 to solve this problem like I hesitated?

Yes, that would have been better :) 
We can force the R version here to be 4.1

ecamenen:
(Tests debugged; I will do another commit when the conda is ready)

ecamenen:
(Tests passed.)

bgruening (approved):

bgruening:
Thanks @ecamenen! Sorry for the slow turnaround.

PR #3749, opened 2021-06-21
Update data_manager_primer_scheme_bedfiles to support ARTICv4 primers

pvanheus:
FOR CONTRIBUTOR:
* [x] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [x] - License permits unrestricted use (educational + commercial)
* [ ] - This PR adds a new tool or tool collection
* [x] - This PR updates an existing tool or tool collection
* [ ] - This PR does something else (explain below)

PR #3767, opened 2021-06-29
Update data_manager_primer_scheme_bedfiles to support ARTICv4, VarSkip and Midnight primers (continuation)

pvanheus:
FOR CONTRIBUTOR:


pvanheus:
Apologies, the previous version of this was #3749 which was closed due to a git mis-configuration.

data_managers/data_manager_primer_scheme_bedfiles/tool_data_table_conf.xml.sample
```diff
@@ -1,7 +1,7 @@
 <tables>
     <!-- Locations of primer scheme BED files -->
     <table name="primer_scheme_bedfiles" comment_char="#" allow_duplicate_entries="False">
-        <columns>value, description, path</columns>
+        <columns>value, description, path, amplicon_info_path</columns>
```
wm75, line 4: @mvdbeek @nsoranzo DM table changes always confuse me :(
Is this an allowed thing to do (adding a column to an existing table)? Or would it require a *new* table?

nsoranzo: If I remember correctly, this would require a new table and loc file, otherwise people will end up with a loc file with some lines having 3 columns and others having 4.

pvanheus: Ok, I'm going to back this change out of this PR and think about where to put it in future.

wm75: @pvanheus you don't have to. Just give the table a new name and all is fine :)

pvanheus: As discussed in chat, I've decided to back out the generation of an amplicon info file, so this PR is now *only* adding support for ARTICv4. I'll make a separate PR to add the amplicon info file as needed on the ivar side.

bwlang:
@pvanheus : looks fine to me.  Is it worth adding
https://raw.githubusercontent.com/nebiolabs/VarSkip/main/schemes/NEB_VarSkip/V1a/NEB_VarSkip.scheme.bed
while you're at it?

pvanheus:
> @pvanheus : looks fine to me. Is it worth adding https://raw.githubusercontent.com/nebiolabs/VarSkip/main/schemes/NEB_VarSkip/V1a/NEB_VarSkip.scheme.bed while you're at it?

The latest iteration adds support for the VarSkip primers.

pvanheus:
The latest commit adds support for Midnight primers and also adds a warning: this tool is *not* idempotent because having duplicated keys in the data table is an error and there is no way (that I know of) for the tool to discover the existing values in the data table. The (admin) user needs to then only choose to download primer schemes that they don't already have copies of - doing anything else will cause an error.

data_managers/data_manager_primer_scheme_bedfiles/data_manager/install_primer_scheme_bedfiles.py
```diff
@@ -18,33 +18,46 @@
...
+                    score = float(fields[4])
+                except ValueError:
+                    # Alright, this is an ARTIC-style bed,
+                    # which is actually against the specs, but required by the
+                    # ARTIC pipeline.
+                    pass
+                else:
+                    # This is a regular bed with numbers in the score column.
+                    # We need to "fix" it for the ARTIC pipeline.
+                    fields[4] = '{0}'.format(score)
```
wm75, line 49: You'd want to prefix the value with an underscore like in the original code?

data_managers/data_manager_primer_scheme_bedfiles/data_manager/install_primer_scheme_bedfiles.py
```diff
@@ -86,7 +103,7 @@ def __call__(self, parser, namespace, values, option_string=None):
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(
-        description="Fetch ARTIC SARS-CoV-2 primer files for Galaxy/IRIDA use"
+        description="Fetch ARTIC and VarSkip SARS-CoV-2 primer files for Galaxy/IRIDA use"
```
wm75, line 106:
```suggestion
        description="Fetch ARTIC, VarSkip and Midnight SARS-CoV-2 primer files for Galaxy/IRIDA use"
```

data_managers/data_manager_primer_scheme_bedfiles/data_manager/install_primer_scheme_bedfiles.py
```diff
@@ -18,33 +18,46 @@
...
-                pass
+            fields = line.strip().split("\t")
+            if "Midnight" in scheme_name:
+                # Midnight primers are distributed in a tabular file, not a BED file
+                if line.startswith("Primer Name"):
+                    continue
+                if len(fields) != 8:
+                    exit("Unexpected format in Midnight primer file: {}".format(line.rstrip()))
+                (primer_name, _, pool, _, _, _, start, end) = fields
+                strand = '+' if primer_name.endswith('LEFT') else '-'
```
wm75, line 32:
```suggestion
                strand = '+' if primer_name.endswith('LEFT') else '-'
                if strand == '-':
                    start, end = end, start
```

data_managers/data_manager_primer_scheme_bedfiles/data_manager/install_primer_scheme_bedfiles.py
pvanheus: I looked at the original file (from the ARTIC folks) and there was no underscore there.

pvanheus: is the _ required in some software?

pvanheus:
The latest version of the PR uses some Cheetah code (explained using comments) to look up existing installed primer schemes and ensure that they are not installed more than once.

data_managers/data_manager_primer_scheme_bedfiles/data_manager/install_primer_scheme_bedfiles.py
pvanheus: Thanks for catching this.

data_managers/data_manager_primer_scheme_bedfiles/data_manager/install_primer_scheme_bedfiles.py
wm75: see https://github.com/galaxyproject/tools-iuc/commit/b93b0d4371450ca0ffd0da6ea5beef8cddbc84e8#diff-fff6f234da80f505e1fb773112d66ce9bb428a3add41304188eb0254d467d7a5
don't know if the cli version of the minion pipeline has been fixed since but even then that would require an update to the Galaxy wrapper.

pvanheus: The relevant software here seems to be [ARTIC minion](https://github.com/artic-network/fieldbioinformatics/blob/master/artic/vcftagprimersites.py#L76 ), which sets the data type of column 5 to a string. This gets [used](https://github.com/artic-network/fieldbioinformatics/blob/master/artic/align_trim.py#L172) to set read groups according to the primer pool. On the other hand [ivar](https://github.com/andersen-lab/ivar/blob/master/src/primer_bed.cpp#L125) interprets this as a BED score but is permissive about it not being a number. The score read in by this code is not actually used. Given all of the above, I'm re-instating the code that insists that it should be a string because this is how it used to be and the different tools seem to be able to handle that setup.

data_managers/data_manager_primer_scheme_bedfiles/data_manager/install_primer_scheme_bedfiles.py
```diff
@@ -62,19 +79,23 @@ def fetch_artic_primers(output_directory, primers):
...
         data.append(dict(value=name, path=bed_output_filename, description=description))
     return data
 
 
 def install_primer_file(
-    output_directory, input_filename, primer_name, primer_description
+    output_directory, input_filename, scheme_name, primer_description
 ):
-    name = re.sub(r"\W", "", str(primer_name).replace(" ", "_"))
+    name = re.sub(r"\W-", "", str(scheme_name).replace(" ", "_"))
```
wm75, line 95: I assume you meant:
```suggestion
    name = re.sub(r"[\W-]", "", str(scheme_name).replace(" ", "_"))
```
?

data_managers/data_manager_primer_scheme_bedfiles/data_manager/install_primer_scheme_bedfiles.xml
```diff
@@ -1,13 +1,33 @@
...
-            --artic_primers '${input.primers}'
+        #if $input.input_type == "network"
+            ## this code looks up the existing table and uses it to build a list of known primers which
+            ## is then used to filter the $input.primers variable. some notes:
+            ##
+            ## $__app__.tool_data_tables is a dictionary where the keys are data table names and the values 
+            ## are TabularToolDataTable objects (from lib/galaxy/tools/data/__init__.py)
+            ##
+            ## the get_fields() method on the TabularToolDataTable returns a list of lists, with one list
+            ## per line of the tool data table, so row[0] is the first field (i.e. the value column)
```
wm75, line 17: Thanks for that detailed explanation. Learnt something new!

data_managers/data_manager_primer_scheme_bedfiles/data_manager/install_primer_scheme_bedfiles.py
pvanheus: Actually neither seem to work (because this is a negative test, I need to exclude - from \W. `(?!-)\W"` works and I have switched to that.

pvanheus:
The test failure is due to a download error from Zenodo which in turn might have been triggered by me doing two commits in close succession and thus triggering the Zenodo anti-bot filter. I have added some explanatory text about how to deal with this in real-world circumstances.

data_managers/data_manager_primer_scheme_bedfiles/data_manager/install_primer_scheme_bedfiles.py
```diff
@@ -62,19 +79,26 @@ def fetch_artic_primers(output_directory, primers):
...
 
 def install_primer_file(
-    output_directory, input_filename, primer_name, primer_description
+    output_directory, input_filename, scheme_name, primer_description
 ):
-    name = re.sub(r"\W", "", str(primer_name).replace(" ", "_"))
+    # (?!-)\W is all non-word charactes except -,
+    # see https://stackoverflow.com/questions/17327765/exclude-characters-from-a-character-class
+    # i.e. remove all non word characters except -
+    name = re.sub(r"(?!-)\W", "", str(scheme_name).replace(" ", "_"))
```
wm75, line 98: Ah, didn't get your intention right, sorry (and thought `-` was a word character).
I guess I would have chosen the (to me) more self-explaining `'[^\w-]'` (not a word character, nor a minus), but fine with your code plus explanation.

pvanheus: Actually yours is easier to read, so I've switched to that.

data_managers/data_manager_primer_scheme_bedfiles/data_manager/install_primer_scheme_bedfiles.xml
```diff
@@ -64,10 +94,20 @@
         Amplicon sequencing for viral pathogens using the ARTIC_ pipeline or `PrimalSeq and iVar`_ relies on
         identifying primer locations in a reference sequence using BED format files. This 
         data manager populates a Galaxy tool data table, either from files provided via
-        a history or via the ARTIC_ network Github repository.
+        a history or via the ARTIC_, the New England Biolabs VarSkip_ and the Massey 
+        University Midnight_ web pages.
+
+        Note that the Midnight primer description file is downloaded from Zenodo, which has strict download
+        limits to restrict bot-like behaviour. If you get an error downloading the Midnight primers, either:
```
wm75, line 101: So I checked zenodo's documentation and it has https://developers.zenodo.org/#rate-limiting.
60 requests per minute, 2000 requests per hour seems more than sufficient for the purpose of the data manager.

In my own testing I noted temporal unavailability of zenodo today, which also may have caused your earlier test failures, but that hardly justifies such a warning, which makes it sound as if zenodo was somehow less suited for data deposition than github.

Maybe say that the user, on failures in general, can try to visit the corresponding web pages in a browser to see if that site has issues?

wm75 (approved):
Very nice enhancements. Good to go after rewording of the help section on failed downloads.
Thanks a lot!

data_managers/data_manager_primer_scheme_bedfiles/data_manager/install_primer_scheme_bedfiles.xml
pvanheus: I have simplified the help text.

PR #3810, opened 2021-07-15
Updates ena

ieguinoa:
- Refactored the parsing script to take into account dynamically loaded optional columns of metadata (using a dict).
- Applied this to the case of samples with an extended set of optional metadata columns in the case of viral submissions
- Upgrade to cli v0.3.1 which can handle these columns.
- Added a test for this case.
- Fixed the parsing of date fields which are normally stored in float values by Excel.


bgruening (approved):

bgruening:
Thanks @ieguinoa

PR #3830, opened 2021-07-20
tb_variant_filter: fix invalid xml

martenson:
this should have never passed linting?

CI error reported at https://github.com/galaxyproject/tools-iuc/issues/3831


pvanheus (approved):

pvanheus (approved):

tools/tb_variant_filter/tb_variant_filter.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="tb_variant_filter" name="TB Variant Filter" version="@TOOL_VERSION@+galaxy0" profile="21.01">
+<tool id="tb_variant_filter" name="TB Variant Filter" version="@TOOL_VERSION@+galaxy1" profile="21.01">
```
nsoranzo, line 1: I don't think we need the version bump? This would keep a non-functional tool installed.

martenson: The toolshed never generated an installable revision for any of these so it makes sense to differentiate between non-working code version and a working version I think.

nsoranzo (approved):

PR #3839, opened 2021-07-20
Updating tools/vardict tool version(s)

planemo-autoupdate:
Hello! This is an automated update of the following tool: **tools/vardict**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

Please see https://github.com/planemo-autoupdate/autoupdate for more information.

### DIFF
--- a/tools/vardict/macros.xml
+++ b/tools/vardict/macros.xml
@@ -1,6 +1,6 @@
 <macros>
     <token name="@VERSION_SUFFIX@">0</token>
-    <token name="@TOOL_VERSION@">1.8.1</token>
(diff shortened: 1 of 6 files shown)


bernt-matthias:
Would be better to change the test assumptions, eg allowing lines_diff or use assertions instead.

bgruening (approved):
Agreed, but this could be done later, maybe?

bernt-matthias (approved):
Seems odd that the VCF version is decreased and the tool version is increased in the test data.

bgruening:
Indeed, maybe they reverted the VCF version.

PR #3843, opened 2021-07-21
[proposal] update CODEOWNERS

simonbray:
The idea is to have a more clearly defined list of maintainers for each tool. This would help with review of the autoupdate PRs (or just PRs in general). See the discussion on https://github.com/galaxy-iuc/standards/issues/65.

This list was just what I was able to extract from the git logs, I imagine it would need a lot of correction.


bernt-matthias:
Would be great to update this dynamically? Maybe also maintaining this in a metadata file per tool is an option (shed.yml?).

simonbray:
> Maybe also maintaining this in a metadata file per tool is an option (shed.yml?).

That was my first thought as well (https://github.com/galaxy-iuc/standards/issues/65), and I do still like that idea, but it was suggested this might be a simpler approach.

bgruening:
@simonbray why is this really needed for your use case? You can find the contributors also as part of the bot-run and tag users.

FWIW I think adding it to the shed.yml is the more explicit solution and other smaller repos already have problems catching up with all our CI infrastructure changes and requirements.

simonbray:
> why is this really needed for your use case?

Ultimately it is the same information, whether it is stored in the shed.yml file or in this CODEOWNERS file. I don't have a strong opinion about which is better.

An advantage of this way would be that maintainers get pinged by all PRs to their tools, not just the autoupdate ones - at least, I assume so, if it works the same as on the GTN.

nsoranzo:
I'm not sure how easy it would be to merge this PR, given that a lot of people would need to approve it. So I went ahead with a smaller PR just to add myself: https://github.com/galaxyproject/tools-iuc/pull/3844
You may notice that there I've changed the paths to start with `/` (so they are considered absolute w.r.t. the root of the repository) and removed the `*` (so that also changes in subdirectories are picked up).

Happy to merge/rebase this PR on top of mine if/when that one is merged.

simonbray:
In the end I just went ahead with the original idea to put maintainers into the `.shed.yml` (e.g. https://github.com/bgruening/galaxytools/pull/1166/files).

Of course, that doesn't stop anyone from adding themselves to the CODEOWNERS file as well, but I will close this PR.

PR #3580, opened 2021-04-07
Update Mykrobe conda version from 0.7.0 to 0.9.0 

dankein:
FOR CONTRIBUTOR:
* [x] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [x] - License permits unrestricted use (educational + commercial)
* [ ] - This PR adds a new tool or tool collection
* [x] - This PR updates an existing tool or tool collection
* [ ] - This PR does something else (explain below)

Updating Mykrobe from Conda version 0.7.0 to 0.9.0

Issue #2412, opened 2019-05-13
Mykrobe Predict custom probe incorrect


dankein:
When I select a custom panel for M. tuberculosis in the Mykrobe predict tool, the tool does not use the custom panel. It runs the "201901" (tb-hunt-probe-set-jan-03-2019.fasta.gz) file instead.

Selections:
**Specify Species for AMR:** Mycobacterium tuberculosis
**Select panel for TB only** custom
**Custom Probe** uploaded file
**Custom Variant to Resistance** uploaded file

This feature works with the same "custom probe" and "custom variant to resistance" files on the command line.

PR #3849, opened 2021-07-23
Update Mykrobe to v0.10.0

emarinier:
FOR CONTRIBUTOR:

Primarily, this pull request updates Mykrobe to version 0.10.0 and adds Shigella sonnei to the Galaxy wrapper. However, as Mykrobe v0.10.0 removes `mykrobe genotype`, I have removed the corresponding wrapper.

Additionally, testing Mykrobe on the command line suggests that using custom panels with `mykrobe predict` does not work and, after confirming with a regular user of the program that this is the case, the custom panel options have been removed from the wrapper.

Finally, when specifying the TB panel "walker-2015", `mykrobe predict` will crash (outside of Galaxy), so this particular panel has been removed from the wrapper as well.


ericenns:
Looking at other tools that have been recently updated, it likely makes sense to add a `TOOL_VERSION` macro, as this would allow auto updating of the wrapper when a new version of mykrobe is released by the bot.

See https://github.com/planemo-autoupdate/tools-iuc/blob/master/tools/seqwish/seqwish.xml for an example of `TOOL_VERSION` macro.

emarinier:
> 
> 
> Looking at other tools that have been recently updated, it likely makes sense to add a `TOOL_VERSION` macro, as this would allow auto updating of the wrapper when a new version of mykrobe is released by the bot.
> 
> See https://github.com/planemo-autoupdate/tools-iuc/blob/master/tools/seqwish/seqwish.xml for an example of `TOOL_VERSION` macro.

Added.

bernt-matthias:
Can you check if this solves/superseeds:

- https://github.com/galaxyproject/tools-iuc/issues/2412
- https://github.com/galaxyproject/tools-iuc/pull/3580
- https://github.com/galaxyproject/tools-iuc/issues/2312

?

emarinier:
* [Mykrobe Predict custom probe incorrect #2412](https://github.com/galaxyproject/tools-iuc/issues/2412): This pull request addresses this issue by removing the functionality from the wrapper, as the real issue appears to be either a bug in Mykrobe or inaccurate user documentation (at least as far as I understand the problem).
* [Update Mykrobe conda version from 0.7.0 to 0.9.0  #3580](https://github.com/galaxyproject/tools-iuc/pull/3580): This pull request supersedes this issue.
* [make sure most used tools support compressed input if possible #2312](https://github.com/galaxyproject/tools-iuc/issues/2312): Unchanged. I have not made any changes to inputs.

bernt-matthias:
> Mykrobe Predict custom probe incorrect #2412: This pull request addresses this issue by removing the functionality from the wrapper, as the real issue appears to be either a bug in Mykrobe or inaccurate user documentation (at least as far as I understand the problem).

Then it might be good to open an upstream issue .

> make sure most used tools support compressed input if possible #2312: Unchanged. I have not made any changes to inputs.

Would be cool if you could add this here.

tools/mykrobe/macro.xml
```diff
@@ -97,14 +71,14 @@
         <option value="collection">Collection Paired-end Data</option>
       </param>
       <when value="single">
-        <param name="fastq_input1" type="data" format="fastqsanger, fastq,fasta,bam" label="Single end read file(s)"/>
+        <param name="seq1" type="data" format="fastqsanger,fastqsanger.gz,fastq,fastq.gz,fasta,fasta.gz,bam" label="Single end read file(s)"/>
```
bernt-matthias, line 74: `fastq` already allows for `fastqsanger` (and analogous for the gzipped types). I guess removing `fastq` and `fastq.gz` is the best option, except if the tool can deal with other quality encoding schemes.

tools/mykrobe/mykrobe_predict.xml
```diff
@@ -103,6 +97,18 @@
         </assert_contents>
       </output>
     </test>
+    <test>
+      <param name="type" value="single"/>
+      <param name="fastq_input1" value="reads.fastq"/>
```
bernt-matthias, line 102: please add the `ftype` attribute.

bernt-matthias (commented):
Thanks for adding the zipped input support. Can you adapt one of the tests to show that zipped input works?

tools/mykrobe/mykrobe_predict.xml
emarinier: Added.

emarinier:
I've added a copy of an existing test using the same data, but gzipped instead.

tools/mykrobe/macro.xml
emarinier: Sorry, I'm only somewhat familiar with both Galaxy and Mykrobe, so I'm not sure how to proceed.

As far as I'm aware, Mykrobe accepts [fasta, fastq, and bam files](https://github.com/Mykrobe-tools/mykrobe/wiki/AMR-Prediction#full-usage) and apparently can accept gzipped files, although this seems to be mentioned only [through examples](https://github.com/Mykrobe-tools/mykrobe/wiki/AMR-Prediction#examples-of-basic-usage) and not explicitly in their [usage documentation](https://github.com/Mykrobe-tools/mykrobe/wiki/AMR-Prediction#full-usage). However, command-line testing suggests it will accept fastq.gz fasta.gz files.

I decided to leave the original "fastqsanger" specifications in the wrapper (and add the corresponding gzipped variant), because I'm assuming the original author knew they were needed, but I don't know if this is the case. I tried searching the Mykrobe Git repository for "sanger" or "fastqsanger" and I don't find anything. I'm also uncertain how Galaxy treats this file types behind the scenes.

How should I proceed?

```
format="fastq,fastq.gz,fasta,fasta.gz,bam"
```

or

```
format="fastqsanger,fastqsanger.gz,fasta,fasta.gz,bam"
```

Thanks for the help! :)

bgruening (approved):
Thanks a lot!

PR #3876, opened 2021-08-11
add potential meryl-fix

bgruening:
FOR CONTRIBUTOR:


wm75 (approved):
worth a try it seems :)

I guess the same trick could be applied to trio mode?

wm75:
Note: It's not clear to me how "trio-mode" is supposed to work. I can see how it would handle multiple child input seqs, but I don't know how, conceptually, there can be more than 1 mother and 1 father sample, and the code also doesn't seem to handle these two inputs correctly (it creates mat|pat{i}.meryl dirs but then tries to use just a single mat|pat.meryl dir.

I don't know much about meryl though, so maybe this all makes perfect sense?

Anyway, not directly related to this PR:

PR #3881, opened 2021-08-16
updated wrapper so busco auto-detect lineage works

GraceAHall:
FOR CONTRIBUTOR:
* [ YES ] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [ YES ] - License permits unrestricted use (educational + commercial)
* [ NO ] - This PR adds a new tool or tool collection
* [ YES ] - This PR updates an existing tool or tool collection
* [ NO ] - This PR does something else (explain below)

This pull fixes an issue with the busco 5.2.2 wrapper. 

When running busco, either a specific lineage should be given with --lineage_dataset, or the lineage can be auto-detected by providing --auto-lineage.  It is not allowed to pass both at the same time.   The current wrapper passes them both at the same time, disabling any use of the --auto-lineage feature as --lineage_dataset has priority.  

This pull provides a sample fix.   The command section of the tool xml should now have proper behaviour, and a 'None' option (default) has been added to the lineages. The 'None' lineage option isn't needed, but I believe will help people feel more at ease when using the tool as it is more intuitive.

### DIFF
--- a/tools/busco/busco.xml
+++ b/tools/busco/busco.xml
@@ -29,10 +29,10 @@ busco
 --evalue ${adv.evalue}
 --limit ${adv.limit}
 
-#if $adv.auto_lineage:
-    $adv.auto_lineage
-#else:
-    --lineage_dataset '${lineage_dataset}'
+#if $lineage.lineage_mode == "auto_detect":
+    $lineage.auto_lineage
+#else if $lineage.lineage_mode == "select_lineage":
+    --lineage_dataset '${lineage.lineage_dataset}'
 #end if
 #if $busco_mode.mode == 'geno' and $busco_mode.use_augustus.use_augustus_selector == 'yes':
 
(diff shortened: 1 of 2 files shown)


bgruening (commented):
That looks great, thanks @GraceAHall. Can you please bump the galaxy tool version?

Thanks.

GraceAHall:
Thanks Björn! I'm still relatively new to the proper github process, so I hope I have done things the way you intended :)

tools/busco/macros.xml
```diff
@@ -109,6 +109,7 @@
     </xml>
 
     <xml name="lineages">
+        <option value="none" selected="true">None</option>
```
abretaud, line 112: I find the `None` option a bit misleading, would it be possible instead to rework both `auto_lineage` and `lineage_dataset` with a conditional, to display the list of lineages only when the user chooses not to use the auto_lineage mode? (I can help if you need!)

abretaud (commented):
Thanks for fixing it! 2 little remarks

tools/busco/macros.xml
GraceAHall: Yep! Can do! Ill do it first thing in the morning

tools/busco/macros.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <macros>
-    <token name="@TOOL_VERSION@">5.2.2</token>
+    <token name="@TOOL_VERSION@">5.2.3</token>
```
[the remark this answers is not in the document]
GraceAHall: Yep that was a bit silly of me. Busco itself isn't changing!

tools/busco/macros.xml
GraceAHall: Here is how it looks on my dev machine now. 

If this is what you had in mind, I'll make another commit for these changes! 
The tool wrapper version change has been added too, and busco has been returned to v5.2.2 haha. 

busco_conditional2

abretaud: Yes that looks great :+1: ! I'd just remove `optional="true"` on the 2 params as it's not clear what it will do if nothing is selected

tools/busco/busco.xml
```diff
@@ -96,19 +96,28 @@ busco
...
             <when value="prot" />
         </conditional>
-
-        <param argument="--lineage_dataset" type="select" label="Lineage">
-            <expand macro="lineages" />
-        </param>
-
+        <conditional name="lineage">
+            <param name="lineage_mode" type="select" label="Auto-detect or select lineage?" help="">
+                <option value="nothing_selected">Nothing selected</option>
```
abretaud, line 101: I was thinking about removing this one too, unless it is needed somehow?

GraceAHall: Not needed! Just removed it :)

abretaud (approved):

tools/busco/busco.xml
```diff
@@ -164,12 +179,14 @@ busco
         </test>
         <test expect_num_outputs="4">
             <param name="input" value="transcriptome.fa" />
-            <param name="lineage_dataset" value="arthropoda_odb10" />
+            <conditional name="lineage">
```
abretaud, line 182: Looks like this test is failing because busco doesn't find any results. I guess you could split this test in 2:
- keep the test in transcriptome mode with arthropoda_odb10 as is was before
- add a test for auto lineage alone. Not sure if there's an easy way to make busco find something in the data in auto lineage mode, otherwise you can just check that it "fails as expected"

PR #3884, opened 2021-08-17
Update nextclade to support 1.x series

pvanheus:
FOR CONTRIBUTOR:


pvanheus:
This updates the nextclade tool and switches it from using the 0.X series (javascript) tool to using the 1.X series (C++) version of nextclade. Since the new version of nextclade does not come with the database built in, this tool downloads the database from github. 

In the future nextclade will have its down download functionality, and the data will be removed from Github, at which time this tool will need and update to follow upstream changes.

pvanheus:
After some discussion from @wm75 I did some upstream work to make a bioconda package that contains the "current" version of the database associated with the nextclade release (see https://github.com/bioconda/bioconda-recipes/pull/30138) and by default this tool now uses that as the nextclade DB while also giving the user the option to download the *latest* database at runtime.

pvanheus:
Following more discussion with @bgruening and @wm75 the "download the latest data" functionality was removed as (1) it leads to non-reproducible analyses and (2) if users really want this they can download and use the latest data themselves by providing the files listed under "advanced options".

tools/nextclade/link_db.sh
```diff
@@ -0,0 +1,20 @@
...
+if [[ ! -d $DB_PATH ]] ; then
+    echo "DB_PATH $DB_PATH is not a directory" >&2
+    exit 1
+fi
+
+mkdir db
+for filename in qc.json reference.fasta tree.json genemap.gff primers.csv ; do
+    ln -s $DB_PATH/$filename db/$filename
+done
+
```
wm75, line 20: Guess I would reduce all of the above to a:

```
export DB_PATH="\$(dirname "\$(dirname "\$(which nextclade)")")/share/nextclade" &&
ln -s "\$DB_PATH" db &&
```

in the command section of the wrapper.

wm75 (approved):
Maybe consider my suggested simplification, but LGTM otherwise.

Thanks for always keeping track of upstream developments!

wm75:
Thanks @pvanheus !

PR #3885, opened 2021-08-17
Deepbgc

EngyNasr:
Screenshot from 2021-08-17 20-37-06
@bgruening 

So now I do have multiple this to ask about:

In this PR there are three wrappers for DeepBGC tool included; prepare, train and pipeline, and the test-data folder which is available on the tool's [github page](https://github.com/Merck/deepbgc#train-deepbgc-on-your-own-data)

Questions/Notes:
1- For Prepare, it works, and I get the two outputs (A GenBank file and a TSV file)
2- For Train, it works too, however, I should output a PKL (pickle file), which I cannot present on usegalaxy since I don't have its format. (I need help to either know a replacement format or to add this format)
3- For Pipeline,
    - I cannot run it, since I also cannot test it using the command line and the test data provided by the tool. Using the tool command line and guessed test data file, I get the error attached
    - I don't know yet what does this wrapper output, and I couldn't figure it out from the tool's explanation or the wrapper python file

General Note: 
I need to find more test data


bgruening:
For what is pickle used? Is it nessary for the third tool?

EngyNasr:
> Uploading Screenshot_20210818-115859~2.png…
> 
> For what is pickle used? Is it nessary for the third tool?

Two of the third tool (Pipeline) arguments; classifier and detector, both need trained model pickle files, however, the tool also suggest to use the model names available after running "deepbgc download" (Possible pre defined detectors and classifiers downloaded by the tool to choose from), which I made available through running the command and using "Select". But then we will not use the train wrapper at all, that's why I guessed that adding pkl will help in passing the train wrapper output to the pipeline wrapper. But I am still reading to understand the tools. In case of the "train" output is very needed then, I can do as you explained to me before, add both wrappers together and pass the inputs internally

bgruening:
Let's put those tools on hold for the moment pickles are not good security wise. I will tell you more next week. Please proceed with another tool/topic.

hechth:
@hechth

PR #3886, opened 2021-08-18
Made Phyml use mpirun for multi cpu - Bioconda package for phyml will…

Slugger70:
… install and configure mpi so this will work.

FOR CONTRIBUTOR:

This change allows phyml to use multiple cpus during its bootstrapping. The Bioconda package for phyml includes all the necessary mpi binaries etc and phyml has been compiled with mpi in mind.

### DIFF
--- a/tools/phyml/phyml.xml
+++ b/tools/phyml/phyml.xml
@@ -17,7 +17,7 @@
           ln -sf '${$userInputTree}' '${$userInputTree.name}' && 
           #end if
           
-          mpirun -mca orte_rsh_agent true -np "\${GALAXY_SLOTS:-4}"
+          mpirun -mca orte_rsh_agent true --allow-run-as-root -np "\${GALAXY_SLOTS:-4}"
           phyml-mpi --input '${input.name}'
                 '${phylip_format}'
                 --datatype '${seq.type_of_seq}'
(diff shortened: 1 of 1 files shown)


Slugger70:
The tests all pass on a local machine where the conda environment is used and the tool works perfectly. According to information found while searching for the error: `The value of the MCA parameter "plm_rsh_agent" was set to a path that could not be found:` it seems that openMPI has a runtime requirement for ssh (which it doesn't use.. :( ).. And so it may need open_ssh installed into the docker container.  According to the Bioconda recipe for Phyml, when running the tool in docker you need to add some workarounds on the openMPI command line. 
```
# mpirun flags needed to work around docker image:
    #   - runs as root (--allow-run-as-root)
    #   - doesn't have rsh or ssh (-mca orte_rsh_agent /bin/true)
```
so the command line would have to look like: `mpirun -mca orte_rsh_agent true --allow-run-as-root phyml-mpi` but only when running in Docker. :(

So the question is: What should we do here? Phyml is clearly a tool that benefits massively from parallelisation (and frankly is unusable on a production Galaxy server without it) and so it should be enabled...

Slugger70:
Argh, it still fails here despite passing on my production server and my dev instance. It seems that the use of Docker is the problem and I'm out of ideas...

PR #3907, opened 2021-08-27
fix pangolearn DM tag parsing

bernt-matthias:
there are tags like: 2021-07-07-2 which made the DM fail

FOR CONTRIBUTOR:

### DIFF
--- a/data_managers/data_manager_pangolearn/data_manager/pangolearn_dm.py
+++ b/data_managers/data_manager_pangolearn/data_manager/pangolearn_dm.py
@@ -15,6 +15,8 @@ import requests
 def extract_date(tag_str):
     parts = tag_str.split("_")
     assert len(parts) < 3, "expected maximum of two parts, got " + str(parts)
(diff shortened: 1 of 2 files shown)


pvanheus:
We need to consider the future of this DM - pangolin switched from using `pangoLEARN` to using `pangoLEARN`, `constellations` and `pango-designation`. The latest `pangolin` tool supports downloading these to a datadir, but this DM is based on the model of just updating `pangoLEARN`. It would be relatively straight forward to write a DM to update all three sources of data, but it should have a new table format (with versions for each database used) and a new tool name.

bernt-matthias:
Thanks @pvanheus . Question is if we can merge this to have one tool less that fails weekly CI?

pvanheus:
> Thanks @pvanheus . Question is if we can merge this to have one tool less that fails weekly CI?

go ahead! we can fix the larger question of whether to keep this tool in its current form downstream

PR #3913, opened 2021-08-28
fix pangolin tests

bernt-matthias:
output seems to have changed

not sure if the new assertions will be useful in the future. opinions?

FOR CONTRIBUTOR:

### DIFF
--- a/tools/pangolin/pangolin.xml
+++ b/tools/pangolin/pangolin.xml
@@ -88,7 +88,7 @@
             </conditional>
             <output name="output1" ftype="tabular">
                 <assert_contents>
-                    <has_text_matching expression="B.1.1\t0.0" />
+                    <has_text_matching expression="B\.1\.1\t0\.0" />
                     <has_text text="passed_qc" />
                     <has_n_lines n="1" />
                 </assert_contents>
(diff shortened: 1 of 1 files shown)


tools/pangolin/pangolin.xml
```diff
@@ -103,7 +103,7 @@
             </conditional>
             <output name="output1" ftype="tabular">
                 <assert_contents>
-                    <has_text_matching expression="B.1.1\t0.0\t*PUSHER" />
+                    <has_text_matching expression="B\.1\.1.*\t0\.0\t*PUSHER" />
```
bgruening, line 106: How could the test pass initially? I don't understand those failures :(

bgruening (approved):
Thanks for fixing it!

tools/pangolin/pangolin.xml
bernt-matthias: I think pangolin uses an online database in the tests. This seems to change over time.

bgruening: ah, the missing backslashes are not the problem but the missing `.`. Ok got it.

bernt-matthias: Maybe we should add `.*` it in all tests?

bgruening: @pvanheus do you have an idea.

It would make the test less stringent I think. But if its needed let's do it.

PR #3916, opened 2021-08-31
Update nextclade and nextalign to version 1.3.0

pvanheus:
FOR CONTRIBUTOR:


pvanheus:
This updates nextalign to use the latest version of the binary (version 1.3.0) and updates tests to correspond with the output format produced by this version (its a small change). It also updates nextclade to use the next `nextclade dataset get` command and support all the organisms currently listed in their database.

tools/nextclade/macros.xml
```diff
@@ -1,5 +1,5 @@
 <macros>
-    <token name="@TOOL_VERSION@">0.1.6</token>
+    <token name="@TOOL_VERSION@">1.3.0</token>
```
bernt-matthias, line 2: That's an odd requirement macro, isn't it?

bernt-matthias (approved):
Wondering if production Galaxy servers put viral reference sequences in all_fasta, I always considered it for large genomes...

tools/nextclade/nextalign.xml
```diff
@@ -1,9 +1,9 @@
-<tool id="nextalign" name="NextAlign" version="@TOOL_VERSION@+galaxy0" profile="20.01">
+<tool id="nextalign" name="NextAlign" version="@PACKAGE_VERSION@+galaxy0" profile="20.01">
     <macros>
         <import>macros.xml</import>
     </macros>
     <expand macro="requirements">
```
bernt-matthias, line 5: You still expand the macro here. 

The best thing would be to define in macros.xml:

```
<xml name="requirements">
        <requirements>
           <requirement type="package" version="@PACKAGE_VERSION@">nextclade</requirement>
            <yield/>
        </requirements>
    </xml>
```

Then you can use

-  in nextalign.xml : `<expand macro="requirements"/>`
- and in nextclade.xml: ```
<expand macro="requirements">
    <requirement type="package">coreutils</requirement>
</expand>
```

in case you need coreutils for `tail` (I guess so)...

Actually `TOOL_VERSION` is preferrred: https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html

pvanheus: I've switched back to using `@TOOL_VERSION@` - the situation here is a bit complicated. The two tools - nextalign and nextclade - are distinct tools but currently are being released with the same version number each time a new release comes out. This might change in the future, in which case the current system would have to change, but right now they `<requirement>` is different for each but the version number is the same.

I've added back `coreutils` - I had forgotten that I need that for `tail`.

bernt-matthias: OK. Got it. Ready to merge?

pvanheus: Yes! Merge merge merge! :)

bgruening (approved):

PR #3919, opened 2021-09-03
Add DeepVariant tool

gallardoalba:
FOR CONTRIBUTOR:


tools/deepvariant/deepvariant.xml
```diff
@@ -0,0 +1,278 @@
...
+        --call_variants_extra_args="use_openvino=true" ## Setting this will use OpenVINO on Intel CPUs, which empirically reduces call_variants runtime by 15%-25%.
+        --num_shards=\${GALAXY_SLOTS:-2}
+        && gunzip './output.vcf.gz'
+        #if $output_gvcf
+            && gunzip './output.g.vcf.gz'
+        #end if
+    ]]>    </command>
+    <inputs>
+        <conditional name="reference_genome">
+            <param name="source" type="select" label="Source for the reference genome" help="Built-in references were created using default options.">
```
bgruening, line 43: Should we add here that this tool only works reliably on human genomes?

gallardoalba: In theory is possible to use it in other organisms, but in that case perhaphs is better to restrict it to Humans? 

This is what the documentation says:
"The models included with DeepVariant are only trained on human data. For other organisms, see the blog post on non-human variant-calling for some possible pitfalls and how to handle them."

bgruening: I think if we add this sentence somewhere this is fine.

gallardoalba: It was included before in the help section, but I didn't realize about it.

bgruening (approved):

bgruening:
Awesome, thanks @gallardoalba!

mvdbeek:
@bgruening I think we should have had a discussion first on whether we will merge container-only tools and under which conditions we will do that, and whether there need to be improvements in Galaxy and the TS interface on how admins are alerted to this fact. There is https://anaconda.org/bioconda/deepvariant, was there a reason this couldn't be used ?

gallardoalba:
> @bgruening I think we should have had a discussion first on whether we will merge container-only tools and under which conditions we will do that, and whether there need to be improvements in Galaxy and the TS interface on how admins are alerted to this fact. There is https://anaconda.org/bioconda/deepvariant, was there a reason this couldn't be used ?

I just used it because it is the developer's recommendation https://github.com/google/deepvariant#how-to-run-deepvariant

bgruening:
> There is https://anaconda.org/bioconda/deepvariant, was there a reason this couldn't be used?

This package has multiple problems and no tests: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/deepvariant/meta.yaml
The real show-stopper is that it does not compile under older glibc and thus does not fulfill the usual requirements.

PR #3934, opened 2021-09-09
Update FSD: Regions and FSD: Before/After to use BAM index generated by Galaxy

### LINKED ISSUE #3931: FSD: Regions and FSD: Before/After should probably use the Galaxy-generated BAM index rather than generating one
These tools call `pysam.index(bam_file)` which indexes next to the BAM, which in Galaxy is not going to be a writable path. These tools use pysam 0.15, but later versions have an option to specify the path when creating an index. However, these tools should probably just make use of the BAM index that is automatically generated for all BAM datasets in Galaxy, I haven't dug super deep but it looks like the input in question should have the index?

natefoo:
Also, if `--bamIndex` is not specified, the index will be built in the current working directory rather than adjacent to `bamFile`.

Fixes #3931.

FOR CONTRIBUTOR:


nsoranzo:
If it was me implementing this, I would not add the `--bamIndex` option but
- symlink the bam input and its index to the working directory in the Cheetah
- in the Python scripts check for the `.bai` file and create it if it doesn't exist.

But it works this way as well, of course.

natefoo:
@nsoranzo done. I guess I can revert to the older pysam and drop the explicit index filename handling as well.

tools/fsd/fsd_beforevsafter.xml
```diff
@@ -1,14 +1,18 @@
...
     <expand macro="requirements">
         <requirement type="package" version="1.71">biopython</requirement>
-        <requirement type="package" version="0.15">pysam</requirement>
+        <requirement type="package" version="0.16.0.1">pysam</requirement>
     </expand>
-    <command>
+    <command><![CDATA[
+#if $bamFile:
+ln -s '$bamFile' 'reads.bam' &&
+ln -s '$bamFile.metadata.bam_index' 'reads.bam.bai' &&
```
nsoranzo, line 14:
```suggestion
    ln -s '$bamFile' 'reads.bam' &&
    ln -s '$bamFile.metadata.bam_index' 'reads.bam.bai' &&
```

tools/fsd/fsd_beforevsafter.xml
```diff
@@ -21,7 +25,7 @@ python '$__tool_directory__/fsd_beforevsafter.py'
 #end if
```
nsoranzo, line 25: Can you also replace

```
--bamFile '$bamFile'
```

with
```
    --bamFile reads.bam
```

just above this line please?

tools/fsd/fsd_regions.xml
```diff
@@ -17,7 +19,7 @@ python '$__tool_directory__/fsd_regions.py'
 #end if
```
nsoranzo, line 19: `--bamFile reads.bam` above

nsoranzo:
> @nsoranzo done. I guess I can revert to the older pysam and drop the explicit index filename handling as well.

Thanks @natefoo ! I suppose the explicit index filename handling can be useful if the scripts are to be used outside of Galaxy, not sure if that's of interest.

natefoo:
I went ahead and reverted to pysam 0.15 and dropped the index filename syntax so that we don't generate new containers for this when it's unnecessary. If this looks good I'll squash it.

tools/fsd/fsd_beforevsafter.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<tool id="fsd_beforevsafter" name="FSD Before/After:" version="1.0.2" profile="19.01">
+<tool id="fsd_beforevsafter" name="FSD Before/After:" version="1.0.3" profile="19.01">
```
nsoranzo, line 2: Given that you've restored the pysam requirement version, you may also restore the tool versions since the rest are bug fixes not changing input or output. Up to you!

nsoranzo (approved):

nsoranzo (approved):

PR #3938, opened 2021-09-10
Update SnpSift Annotate to avoid writing to input directory

natefoo:
Otherwise it tries to create `${dbSnp}.sidx`. There may be other tools in the suite that would also encounter this - any ideas @wm75?

FOR CONTRIBUTOR:

### DIFF
--- a/tool_collections/snpsift/snpsift/snpSift_annotate.xml
+++ b/tool_collections/snpsift/snpsift/snpSift_annotate.xml
@@ -10,12 +10,13 @@
     <expand macro="stdio" />
     <expand macro="version_command" />
     <command><![CDATA[
+ln -s '$dbSnp' 'dbSnp.vcf' &&
 SnpSift -Xmx8G annotate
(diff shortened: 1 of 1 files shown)


bernt-matthias:
Would be nice to extend planemo such that it tests if any extra files are created, or?

bgruening (approved):

bgruening:
> Would be nice to extend planemo such that it tests if any extra files are created, or?

That should be possible already. We are using Docker. I'm actually surprised that the input folder is mounted `rw` and not `r`. We could ship a job_conf file here as part of the testing. I think I had a setup on my tools repo that even used ht-condor for testing tools.

wm75:
> There may be other tools in the suite that would also encounter this

From a quick check of the source code it seems that *annotate* is the only tool doing this :crossed_fingers: 
This should really be caught during tests. Doing this with docker is nice, but if planemo would catch such cases directly, it would help during tool development directly.

PR #3973, opened 2021-09-20
Updating tools/raven tool version(s)

planemo-autoupdate:

Please see https://github.com/planemo-autoupdate/autoupdate for more information.

### DIFF
--- a/tools/raven/macros.xml
+++ b/tools/raven/macros.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <macros>
-    <token name="@TOOL_VERSION@">1.5.3</token>
+    <token name="@TOOL_VERSION@">1.6.0</token>
     <token name="@PROFILE@">20.05</token>
     <xml name="requirements">
         <requirements>


bernt-matthias:
There seem to be new options https://github.com/lbcb-sci/raven/releases/tag/1.6.0

bgruening:
https://github.com/lbcb-sci/raven/compare/1.5.3...1.6.1

bgruening:
The point is that it will not save any space, isn't it? Once its in github/TS is taking the space, reducing the space later, will just increase the overall space of the repo. So maybe we can merge those failures for older tools and try to be stricter with new tools - if possible?

bernt-matthias:
That's a good point. It will only save space if you have a shallow clone of the repo as in the CI workflows.

simonbray:
> The point is that it will not save any space, isn't it? Once its in github/TS is taking the space, reducing the space later, will just increase the overall space of the repo. So maybe we can merge those failures for older tools and try to be stricter with new tools - if possible?

+1, the file size check could be changed to only take changed files into account

bernt-matthias:
@bgruening and @simonbray I tried to include this idea here https://github.com/galaxyproject/tools-iuc/pull/4077

I will merge this one.

PR #3977, opened 2021-09-21
Presto update

aerijman:
FOR CONTRIBUTOR:


bgruening:
@aerijman please update the tool version. https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html?highlight=version#tool-versions

tools/presto/presto_pairseq.xml
```diff
@@ -34,12 +33,19 @@
             <expand macro="text-regex-validator"/>
         </param>
         <expand macro="presto-coord-param"/>
+        <param argument="failed_fastq" type="boolean" value="False" truevalue="--failed" falsevalue="" label="Failed reads fastq output"/>
```
bernt-matthias, line 36: Argument should be ` -- failed`, or?

bernt-matthias: Hmm you extend a test for this functionality?

bernt-matthias (changes requested):

tools/presto/presto_pairseq.xml
aerijman: `--failed` outputs reads that couldn't be synced between the reads fwd and rev. We used those to generate histograms with all the data, not only reads that could be paired.  Without this argument we only get fastqs with reads that could be synced

bernt-matthias: Argument is `argument="failed_fastq` which will lead to adding (failed_fastq) to the parameter's help. This is confusing for the user since the cli parameter seems to be `--failed`.

bernt-matthias:
Please also add a recent profile to the tools.

tools/presto/presto_pairseq.xml
aerijman: My apologies if there are more succinct ways of doing this.

aerijman:
profile reflects newest galaxy release. Let me know if this is ok or should I change it! Danke!

tools/presto/presto_pairseq.xml
```diff
@@ -33,7 +33,7 @@
             <expand macro="text-regex-validator"/>
         </param>
         <expand macro="presto-coord-param"/>
-        <param argument="failed_fastq" type="boolean" value="False" truevalue="--failed" falsevalue="" label="Failed reads fastq output"/>
+        <param argument="failed_fastq" type="boolean" value="False" truevalue="--failed" falsevalue="" label="Failed reads fastq output" help="--failed"/>
```
bernt-matthias, line 36:
```suggestion
        <param argument="--failed" type="boolean" value="False" truevalue="--failed" falsevalue="" label="Failed reads fastq output" />
```

otherwise help will become `"--failed (failed_fastq)"` and you also need to adapt the command block. 

The idea of `argument` is to add the command line argument to the end of the help text. Then users know which command line argument is controlled with the tool parameter. Furthermore the cheetah variable for the command is auto generated (leading dashed are removed and internal dashes replaced by underscore).

bernt-matthias (commented):
Profile 21.05 is to new. Then users with not so recent installations can not use the tool. Make it 20.05 at most.

tools/presto/presto_pairseq.xml
aerijman: Thank you for the explanation!!!

aerijman:
@bernt-matthias ready to merge?

bernt-matthias:
@aerijman there are still a few of my suggestions open. Please accept them or respond why they should not be accepted...

aerijman (commented):
I don't have the power to approve the changes. I DO accept them

bernt-matthias:
There is a button `accept changes` below the suggestion. But I will do this for you :)

bernt-matthias:
@aerijman .. my mistake .. actually I forgot to submit my last review :(

tools/presto/presto_macros.xml
```diff
@@ -10,8 +10,9 @@
             <xref type="bio.tools">presto</xref>
         </xrefs>
     </xml>
-    <token name="@TOOL_VERSION@">0.6.2</token>
-    
+    <token name="@TOOL_VERSION@">0.6.2.1</token>
```
bernt-matthias, line 13:
```suggestion
    <token name="@TOOL_VERSION@">0.6.2</token>
```

bernt-matthias (approved):

PR #3986, opened 2021-09-24
avoid bam2msa to create BAM index in inputdir

pavanvidem:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/bioext/bam2msa.xml
+++ b/tools/bioext/bam2msa.xml
@@ -13,7 +13,7 @@
             #if $region_start and $region_end:
                 -r $region_start:$region_end
             #end if
-            '$input' '$output'
+            'input_bam' '$output'
         ]]></command>
     <inputs>
         <param name="input" type="data" format="bam,sam" label="Input BAM file"/>
--- a/tools/bioext/bam2msa.xml
+++ b/tools/bioext/bam2msa.xml
@@ -7,6 +7,8 @@
     </macros>
     <expand macro="requirements"/>
     <command detect_errors="exit_code"><![CDATA[
+        ## avoid bam2msa to create .bai in inputdir
+        ln -s '$input' input_bam &&
         bam2msa
             #if $region_start and $region_end:
                 -r $region_start:$region_end
--- a/tools/bioext/bam2msa.xml
+++ b/tools/bioext/bam2msa.xml
@@ -9,6 +9,7 @@
     <command detect_errors="exit_code"><![CDATA[
         ## avoid bam2msa to create .bai in inputdir
         ln -s '$input' input_bam &&
+        ln -s '$input.metadata.bam_index' input_bam.bai &&
         bam2msa
             #if $region_start and $region_end:
                 -r $region_start:$region_end


bernt-matthias:
Hey @pavanvidem .. how do you identify tools writing to the input dir?

pavanvidem:
First, we extracted the dataset names from input dir that do not end with `.dat` using something like `find . -not -name "*.dat" -type f | grep -v 'dataset_.*_files'`. Then we used  [`gxadmin query q`](https://galaxyproject.github.io/gxadmin/#/README.query?id=query-q) to query the `job` table using each dataset name and extracted the tools out of it. Finally, we manually check/run each tool and fix it.

bernt-matthias:
@pavanvidem I started some discussion here: https://github.com/galaxyproject/planemo/issues/1189

bernt-matthias:
Still the manual approach will be needed (since I guess we won't have perfect test coverage)...

tools/bioext/bam2msa.xml
```diff
@@ -7,11 +7,13 @@
     </macros>
     <expand macro="requirements"/>
     <command detect_errors="exit_code"><![CDATA[
+        ## avoid bam2msa to create .bai in inputdir
+        ln -s '$input' input_bam &&
```
bernt-matthias, line 11: Galaxy already stores the index files (only for bam) and you can access it with:`$input.metadata.bam_index`. So `ln -s '$input.metadata.bam_index' input_bam.bai`  may avoid that `bam2msa` needs to recreate it?

But I guess then we need a version bump.

bgruening (approved):

bernt-matthias:
> First, we extracted the dataset names from input dir that do not end with `.dat` using something like `find . -not -name "*.dat" -type f | grep -v 'dataset_.*_files'`. Then we used [`gxadmin query q`](https://galaxyproject.github.io/gxadmin/#/README.query?id=query-q) to query the `job` table using each dataset name and extracted the tools out of it. Finally, we manually check/run each tool and fix it.

Hey @pavanvidem just thought a bit more about this. The problem is that old versions of the tool will still write to Galaxy's file dir. How about configuring your galaxy to run jobs as a separate user that does not have write permissions for Galaxy's file dir: 

https://github.com/galaxyproject/galaxy/blob/40ddc72f485ae233f3a4aed63847ccb041003320/lib/galaxy/config/sample/galaxy.yml.sample#L2121

Still the problem needs to be fixed: I' currently running IUC's weekly CI with an extended version of planemo (https://github.com/galaxyproject/planemo/pull/1190) : https://github.com/bernt-matthias/tools-iuc/actions/runs/1277704614

pavanvidem:
> Hey @pavanvidem just thought a bit more about this. The problem is that old versions of the tool will still write to Galaxy's file dir. How about configuring your galaxy to run jobs as a separate user that does not have write permissions for Galaxy's file dir:

This is a good idea. I should have mentioned that I queried the EU Galaxy database, not my local instance. This might have covered most of the old tool versions and the tools outside IUC.

bernt-matthias:
@pavanvidem can you check if a bump of the tool version to 0.20.4 fixes this?

PR #3988, opened 2021-09-27
Make bealign parallelism configurable

wm75:
FOR CONTRIBUTOR:


wm75:
@bernt-matthias not a huge expert on pipefail :) Am I doing things right here?

@davebx @astrovsky01 does the new background test data look reasonable to you? Previously that test failed silently.

@mvdbeek The proposed multiprocessing handling looks about right?

tools/bioext/bealign.xml
```diff
@@ -18,7 +19,10 @@
     ## and replace any unwanted characters in the sequence names with underscores.
     ## This should not affect the actual alignment, since any non-standard character
     ## in the sequences is already ignored, but the possibility remains.
+    set -o pipefail;
     cat '$input' @SANITIZE@ reads.fa &&
+    ## bealign's parallel job executor uses the following env variable
+    export NCPU=\${GALAXY_SLOTS:-2} &&
```
mvdbeek, line 25: I would just prefix bealign with the variable and not export it. `NCPU=\${GALAXY_SLOTS:-2} bealign ...`. I guess in practice it doesn't make a difference, but NCPU is often but not always set by the DRM.

tools/bioext/bealign.xml
```diff
@@ -18,7 +19,10 @@
     ## and replace any unwanted characters in the sequence names with underscores.
     ## This should not affect the actual alignment, since any non-standard character
     ## in the sequences is already ignored, but the possibility remains.
+    set -o pipefail;
```
bernt-matthias, line 22: Since there seems to be no pipe I guess this is not needed.

wm75: The pipe is hidden inisde the SANITIZE macro.

bernt-matthias:
Looks good to me. Do we want/need to bump the version?

bgruening:
suffix is bumped, imho. I think @mvdbeek comments need to be addressed and then we can go.

bgruening (approved):

bernt-matthias (approved):

bernt-matthias:
@wm75 PR is fine as it is. Simplifications are welcome in a separate PR, anytime.

Issue #3682, opened 2021-05-19
Request for tool wrap for BBtools


xinhuang66:
We are trying to deploy our pipeline on Galaxy so more people can use it, and we would need to use the BBtools suite. If it can be wrapped into Galaxy, that would be greatly appreciated!

pvanheus:
@xinhuang66 is there any particular functionality you want to BBtools? This could help focus work on a wrapper.

xinhuang66:
Hi, thank you for getting back to me. We would love to use bbduk.sh, bbmap.sh and callvariants.sh in the BBtools suite.

Thank you!

Xin Huang, PhD | Senior Bioinformatician
Discovery Research | Gene Therapy Program<https://gtp.med.upenn.edu/>
Perelman School of Medicine, University of Pennsylvania

Check us out on LinkedIn<https://www.linkedin.com/company/gene-therapy-program-perelman-school-of-medicine-university-of-pennsylvania?trk=ppro_cprof>

***CONFIDENTIALITY NOTICE***
This e-mail message and any documents accompanying this e-mail transmission contain confidential and/or proprietary information from the University of Pennsylvania and shall not be used, disclosed, or reproduced, in whole or in part, without the prior written consent of the University of Pennsylvania.   Title to this document and all information contained herein remains at all times in The University of Pennsylvania.   If you have received this transmission in error, please reply to the sender advising of the error and delete the message and any accompanying documents from your system immediately.

gregvonkuster:
I'll work on these - please let me know if work has already started on them.

xinhuang66:
Thank you very much, Greg!

All the best,

Xin Huang, PhD | Senior Bioinformatician
Discovery Research | Gene Therapy Program<https://gtp.med.upenn.edu/>
Perelman School of Medicine, University of Pennsylvania

Check us out on LinkedIn<https://www.linkedin.com/company/gene-therapy-program-perelman-school-of-medicine-university-of-pennsylvania?trk=ppro_cprof>

***CONFIDENTIALITY NOTICE***
This e-mail message and any documents accompanying this e-mail transmission contain confidential and/or proprietary information from the University of Pennsylvania and shall not be used, disclosed, or reproduced, in whole or in part, without the prior written consent of the University of Pennsylvania.   Title to this document and all information contained herein remains at all times in The University of Pennsylvania.   If you have received this transmission in error, please reply to the sender advising of the error and delete the message and any accompanying documents from your system immediately.

[...]

PR #3993, opened 2021-09-28
Add BBMap tool

gregvonkuster:
FOR CONTRIBUTOR:


bernt-matthias:
Fixes https://github.com/galaxyproject/tools-iuc/issues/3682

tools/bbtools/bbmap.xml
```diff
@@ -0,0 +1,155 @@
...
+                <param name="read1" type="data" format="fastqsanger.gz,fastqsanger" label="Read1 fastq file"/>
+                <param name="read2" type="data" format="fastqsanger.gz,fastqsanger" label="Read2 fastq file"/>
+            </when>
+            <when value="paired">
+                <param name="reads_collection" type="data_collection" format="fastqsanger,fastqsanger.gz" collection_type="paired" label="Collection of fastqsanger paired read files"/>
+            </when>
+        </conditional>
+        <conditional name="ref_source_cond">
+            <param name="ref_source" type="select" label="Select reference genome source; a cached reference or one from the history">
+                <option value="cached" selected="True">Use a cached reference</option>
```
bernt-matthias, line 84: Can you add a test for a cached reference?

tools/bbtools/bbmap.xml
```diff
@@ -0,0 +1,155 @@
...
+                <option value="history">Use a reference from the history</option>
+            </param>
+            <when value="cached">
+                <param name="reference" type="select" label="Using reference genome">
+                    <options from_data_table="fasta_indexes">
+                        <validator type="no_options" message="A built-in reference genome is not available"/>
```
bernt-matthias, line 90: Should there be a filter taking only those entries with the correct dbkey?

tools/bbtools/LICENSE
```diff
@@ -0,0 +1,14 @@
+BBTools Copyright (c) 2014, The Regents of the University of California, through Lawrence Berkeley National Laboratory (subject to receipt of any required approvals from the U.S. Dept. of Energy).  All rights reserved.
```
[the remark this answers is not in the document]
gregvonkuster: It states the following, so I was just being careful.  Since it refers to source code, the conda recipe is likely sufficient.  I'll remove it here.
```
Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
```

tools/bbtools/bbmap.xml
```diff
@@ -0,0 +1,155 @@
...
+    ln -s '${read2}' '${read2_file}' &&
+#end if
+
+bbmap.sh t=\${GALAXY_SLOTS:-4} ref='${ref}'
+#if str($input_type_cond.input_type) == 'single':
+    in='${read1_file}'
+#else:
+    in1='${read1_file}' in2='${read2_file}'
+#end if
+2>'bbmap_stderr.txt'
```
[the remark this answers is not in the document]
gregvonkuster: The script execute Java, which writes a lot of stuff to stderr, and all of it is displayed in the history item.  I'm happy to remove this redirect if that is more ideal.

tools/bbtools/bbmap.xml
```diff
@@ -0,0 +1,155 @@
...
+#if str($output_sort) == 'coordinate':
```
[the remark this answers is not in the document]
gregvonkuster: I figured I'd leverage another iuc supported mapper, so I leverage bwahaha here https://github.com/galaxyproject/tools-iuc/blob/master/tools/bwa/bwa-mem.xml#L266-L270.  The default for BBMap is no sorting.  Is that preferable?

bs.sh is doing the sorting on chromosome name - https://github.com/BioInfoTools/BBMap/blob/master/docs/guides/BBMapGuide.txt#L153.

tools/bbtools/bbmap.xml
gregvonkuster: The first 2 tests use a cached reference - it is the default setting.  The 3rd test uses a reference from the history.

tools/bbtools/bbmap.xml
gregvonkuster: This is a mapper, so downstream tools would use a filter.  In this case, the user is choosing the reference against which to map the inputs, so no filter here unless I'm missing something.

tools/bbtools/LICENSE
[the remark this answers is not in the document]
bernt-matthias: Should be fine to remove, since this is just a wrapper and we are not distributing bbmap source code.

tools/bbtools/bbmap.xml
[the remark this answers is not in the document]
bernt-matthias: I remember that IUC guidelines include not to redirect. The history item should only contain the first few lines (but I thought its stdout). You may redirect to an (optional) log file if you prefer...

tools/bbtools/bbmap.xml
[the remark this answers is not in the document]
bernt-matthias: We can leave it, but then I would prefer samtools, since this can use multiple cores.

tools/bbtools/bbmap.xml
bernt-matthias: I see.

tools/bbtools/bbmap.xml
bernt-matthias: The idea is: 

- If you set the dbkey for the input reads then the select should only show the matching entries
- If you don't set the dbkey all entries are shown (due to a 'bug' in Galaxy this did not work until recently .. )

This helps the user to not choose the wrong entry...  or?

tools/bbtools/LICENSE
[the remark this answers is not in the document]
gregvonkuster: it's removed.

tools/bbtools/bbmap.xml
gregvonkuster: Hmm...but the input reads don't have a dbkey - they are the raw fastqsanger samples.  The BBMap tool in this PR is actually aligning the reads against a selected reference genome, associating the selected dbkey with the bam output.  This tool is performing the same task as other iuc mapping tools (e.g., bwa, bowtie, etc).   

I did add a sort filter so the references are sorted.  But I don't understand the dbkey filter you're asking for.  Can you point to an example in another mapping tool?

tools/bbtools/bbmap.xml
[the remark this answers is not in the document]
gregvonkuster: I switched to using samtools.

tools/bbtools/bbmap.xml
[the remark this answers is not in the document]
gregvonkuster: Ok, I removed the redirect.

tools/bbtools/bbmap.xml
bernt-matthias: > Hmm...but the input reads don't have a dbkey - they are the raw fastqsanger samples. The BBMap tool in this PR is actually aligning the reads against a selected reference genome, associating the selected dbkey with the bam output. This tool is performing the same task as other iuc mapping tools (e.g., bwa, bowtie, etc).

Also sounds reasonable. Can you extend the tests such that they check if the dbkey is really set. Looking at your loc file the dbkey should be 89 for the `cached` case, or? Since this is not a valid dbkey I would also expect at least some warning in the logs.

An example how it might be done is here .. https://github.com/galaxyproject/galaxy/blob/dev/test/functional/tools/dbkey_filter_input.xml .. but not for a wrapper.

gregvonkuster: 89 is a valid dbkey.  It is the UCSC dbkey for the AF2122 reference.  It is in the builds.txt file.
```
$ grep 89 ./shared/ucsc/builds.txt
28921	Acinetobacter baumannii (28921)
18965	Sulfurovum sp. NBC37-1 (18965)
18963	Nitratiruptor sp. SB155-2 (18963)
18981	Finegoldia magna ATCC 29328 (18981)
89	Mycobacterium bovis AF2122/97 (89)
12716	Citrobacter koseri ATCC BAA-895 (12716)
caldSacc_DSM8903	Caldicellulosiruptor saccharolyticus DSM 8903 (caldSacc_DSM8903)
17237	Rickettsia bellii OSU 85-389 (17237)
189	Streptomyces avermitilis MA-4680 (189)
16259	Escherichia coli UTI89 (16259)
18789	Shewanella sediminis HAW-EB3 (18789)
10689	Kineococcus radiotolerans SRS30216 (10689)
13389	Shewanella baltica OS195 (13389)
12720	Enterobacter sakazakii ATCC BAA-894 (12720)
thioDeni_ATCC33889	Sulfurimonas denitrificans DSM 1251 (thioDeni_ATCC33889)
19489	Staphylococcus aureus subsp. aureus USA300_TCH1516 (19489)
```
I think the tests cover this tool quite well, so can this PR be merged?  I've got the next tool ready to commit, but have to wait til this one is merged since suites like this share files (e.g., macros. etc).

tools/bbtools/bbmap.xml
```diff
@@ -0,0 +1,137 @@
...
+        <data format="bam" name="output" label="${tool.name} on ${on_string} (mapped reads in BAM format)">
+            <expand macro="dbKeyActionsBBMap"/>
+        </data>
+    </outputs>
+    <tests>
+        <!-- Single file, cached reference, output coordinate sorted -->
+        <test expect_num_outputs="1">
+            <param name="input_type" value="single"/>
+            <param name="read1" value="13-1941-6_S4_L001_R1_600000.fastq.gz"/>
+            <output name="output" file="output1.bam" ftype="bam" lines_diff="4"/>
```
bernt-matthias, line 98: Something like this in all tests would be nice:

```suggestion
<output name="output" file="output1.bam" ftype="bam" lines_diff="4">
    <metadata name="dbkey" value="89" />
</output>
```

tools/bbtools/bbmap.xml
```diff
@@ -0,0 +1,137 @@
...
+        </test>
+        <!-- Paired reads in separate datasets, cached reference, output name sorted -->
+        <test expect_num_outputs="1">
+            <param name="input_type" value="pair"/>
+            <param name="read1" value="13-1941-6_S4_L001_R1_600000.fastq.gz"/>
+            <param name="read2" value="13-1941-6_S4_L001_R2_600000.fastq.gz"/>
+            <param name="output_sort" value="name"/>
+            <output name="output" file="output2.bam" ftype="bam" lines_diff="4"/>
```
bernt-matthias, line 106: This should then be `qname_sorted_bam`.. The output Dietmar needs to be set according to the value of `output_sort`.

tools/bbtools/bbmap.xml
gregvonkuster: Added the metadata tag.

tools/bbtools/bbmap.xml
gregvonkuster: Sorted bam outputs have been fixed.

bernt-matthias (approved):

bernt-matthias:
Good to go @gregvonkuster ?

gregvonkuster:
> Good to go @gregvonkuster ?

Yes, thanks!

PR #3997, opened 2021-09-30
Jbrowse2 tool

abretaud:
FOR CONTRIBUTOR:

This has been standing too long on my hard drive: here's a wrapper for jbrowse2 :tada: :fireworks: !
Implemented the same way as Jbrowse (though there might be other cool things to do now that it's more embeddable). I mostly copied the jbrowse1 wrapper, and tried to adapt.
It mostly works, but it needs some clean up, there are some options that are still there but don't work yet (or should be removed), some new datatypes to add support to, ....

An interesting thing in jbrowse2 is that many display options can be easily changed from the jbrowse ui (while in jbrowse1 it required tweaking json), which means I'm not sure if we need to (re)implement them all in the form...

Also to test with `planemo server` I had to hack a little as planemo doesn't support byte-range headers. I can probably share my notes if there's interest


hexylena:
Love how much can be re-used for this, that's fantastic

tools/jbrowse2/jbrowse2.xml
```diff
@@ -350,14 +359,15 @@ cp $output.files_path/index.html $output;
 
     <!-- TODO genetic code not yet supported in jbrowse2 https://github.com/GMOD/jbrowse-components/issues/1765 -->
 
+    <!-- TODO show linear view by default? -->
```
hexylena: +1

hexylena:
Interesting, if you use `config.json` in the root, next to index.html, then it'll by default open on to a working jbrowse in the first assembly. Too easy.

image

abretaud:
Ok, it seems like this one is *finally* ready to be reviewed/merged :tada: :tada: :tada: 

There's still a linting error `Unknown datatype [hic] used in param element`, which should be fixed by https://github.com/galaxyproject/galaxy/pull/20874 --> it's not clear to me why the new datatype is not taken into account :sos: 

I think the features from [the tool that got merged in bgruening earlier](https://github.com/bgruening/galaxytools/tree/master/tools/jbrowse2) are present in this one, but overall looks much cleaner (e.g. cleaned up the useless code from jbrowse1 wrapper), and there are some additional features (ref name aliases, cytobands, customizable feature panel, metadata). Happy to make changes if anything is missing.

The tool supports linear genome view and synteny view too. And updating existing jbrowse instance works too, at least in most simple cases. I also added the possibility to reference data from remote urls: I know it's controversial, but I think it can be useful when reusing this tool in largest ebp-like projects, and the warning are quite visible about that.

I've left a short todo list in the README, but nothing blocking for a first merge I think.

Any reviews welcome!

ping @hexylena (if you're still interested :) ) @bgruening @cmdcolin

abretaud:
@Delphine-L would you have a VGP workflow and datasets to check if this tool is able to replace the bgruening one?

tools/jbrowse2/jbrowse2.xml
```diff
@@ -0,0 +1,1075 @@
...
+    </edam_operations>
+    <xrefs>
+        <xref type="bio.tools">jbrowse_2</xref>
+    </xrefs>
+    <expand macro="requirements"/>
+    <command><![CDATA[
+mkdir -p '$output.files_path/data';
+
+#if $action.action_select == "update":
+    cp -R $action.update_jbrowse.extra_files_path/data $output.files_path/;
```
bgruening: quote

tools/jbrowse2/jbrowse2.xml
```diff
@@ -0,0 +1,1075 @@
...
+                                <param name="annotation_source" value="remote" />
+                                <param name="uri" value="https://s3.amazonaws.com/jbrowse.org/genomes/hg19/mappability/wgEncodeDukeMapabilityRegionsExcludable.bed.gz" />
+                                <param name="label" value="wgEncodeDukeMapabilityRegionsExcludable" />
+                            </conditional>
+                        </conditional>
+                    </repeat>
+                </repeat>
+            </repeat>
+            <output name="output">
+                <extra_files name="galaxy.xml" type="file" value="out/02_remote/galaxy.xml" compare="sim_size" />
```
bgruening, line 706: can we avoid here sim_size? do asserts work here?

tools/jbrowse2/jbrowse2.xml
```diff
@@ -0,0 +1,1075 @@
...
+                    <repeat name="data_tracks">
+                        <conditional name="data_format">
+                            <param name="data_format_select" value="gc" />
+                        </conditional>
+                    </repeat>
+                </repeat>
+            </repeat>
+            <output name="output">
+                <extra_files name="galaxy.xml" type="file" value="out/01_all_tracks/galaxy.xml" compare="sim_size" />
+                <extra_files name="config.json" type="file" value="out/01_all_tracks/config.json" compare="sim_size" />
```
bgruening, line 597: asserts?

tools/jbrowse2/macros.xml
```diff
@@ -0,0 +1,355 @@
...
+        <requirements>
+            <requirement type="package" version="@TOOL_VERSION@">jbrowse2</requirement>
+            <requirement type="package" version="1.22.1">samtools</requirement>
+            <requirement type="package" version="1.11">tabix</requirement>
+            <requirement type="package" version="3.13.7">python</requirement>
+            <requirement type="package" version="2.32.5">requests</requirement>
+            <yield/>
+        </requirements>
+    </xml>
+    <token name="@WRAPPER_VERSION@">galaxy0</token>
```
bgruening: maybe just the 0 here? https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tool-versions

name it version_suffix

tools/jbrowse2/macros.xml
```diff
@@ -0,0 +1,355 @@
...
+
+    <xml name="general_options">
+        <section name="jbgen" title="General JBrowse Options [Advanced]" expanded="false">
```
bgruening: I personally don't like the [] - we don't use that anywhere else.

tools/jbrowse2/macros.xml
```diff
@@ -0,0 +1,355 @@
...
+                <when value="LinearHicDisplay"/>
+            </conditional>
+        </section>
+    </xml>
+
+    <xml name="track_styling_gc">
+        <section name="jbstyle" title="Styling options [Advanced]" expanded="false">
+            <conditional name="track_style">
+                <param name="display" type="select" label="Display style" help="How the track will be displayed by default">
+                    <option value="LinearGCContentDisplay" selected="true">GC content</option>
```
bgruening: only one option? Is that conditional needed then?

tools/jbrowse2/macros.xml
```diff
@@ -0,0 +1,355 @@
...
+                <when value="LinearMafDisplay"/>
+            </conditional>
+        </section>
+    </xml>
+
+    <xml name="track_styling_hic">
+        <section name="jbstyle" title="Styling options [Advanced]" expanded="false">
+            <conditional name="track_style">
+                <param name="display" type="select" label="Display style" help="How the track will be displayed by default">
+                    <option value="LinearHicDisplay" selected="true">HiC</option>
```
bgruening: only one option?

tools/jbrowse2/macros.xml
```diff
@@ -0,0 +1,355 @@
...
+                </when>
+            </conditional>
+        </section>
+    </xml>
+
+    <xml name="track_styling_maf">
+        <section name="jbstyle" title="Styling options [Advanced]" expanded="false">
+            <conditional name="track_style">
+                <param name="display" type="select" label="Display style" help="How the track will be displayed by default">
+                    <option value="LinearMafDisplay" selected="true">MAF</option>
```
bgruening: only one option

tools/jbrowse2/macros.xml
```diff
@@ -0,0 +1,355 @@
...
+
+        <param label="Features description"
+                type="text"
+                name="descriptions_name"
+                value="jexl:get(feature,'note') || get(feature,'description')"
+                help="See https://jbrowse.org/jb2/docs/config_guide/#configuration-callbacks for syntax">
+            <expand macro="jexl_sanitize" />
+        </param>
+
+        <param label="Max height" name="max_height" type="integer" value="600" help="Maximum height that the track is permitted to reach in pixels."/>
```
bgruening: add min (max) maybe?

tools/jbrowse2/macros.xml
```diff
@@ -0,0 +1,355 @@
...
+            </param>
+
+            <param label="Font size" name="font_size" type="integer" value="10" />
+        </section>
+    </xml>
+
+    <xml name="track_metadata">
+        <section name="metadata" title="Track metadata [Advanced]" expanded="false">
+            <param name="galaxy_metadata" label="Show Galaxy track metadata" type="boolean" checked="true" truevalue="true" falsevalue="false" help="Make Galaxy metadata available in 'About track' window and faceted track selector" />
+            <param name="metadata_bonus" type="data" format="tsv" label="Additional track metadata" optional="true" help="Display additional track metadata from a TSV file" />
```
bgruening: tabular or at least in addition

tools/jbrowse2/macros.xml
```diff
@@ -0,0 +1,355 @@
...
+            </param>
+            <param name="quaternary_color" type="color" label="Quaternary color" value="#ffb11d">
+                <sanitizer>
+                    <valid initial="string.ascii_letters,string.digits">
+                        <add value="#" />
+                    </valid>
+                </sanitizer>
+            </param>
+
+            <param label="Font size" name="font_size" type="integer" value="10" />
```
bgruening: min/max

tools/jbrowse2/macros.xml
```diff
@@ -0,0 +1,355 @@
...
+            <param name="tertiary_color" type="color" label="Tertiary color" value="#135560">
+                <sanitizer>
+                    <valid initial="string.ascii_letters,string.digits">
+                        <add value="#" />
+                    </valid>
+                </sanitizer>
+            </param>
+            <param name="quaternary_color" type="color" label="Quaternary color" value="#ffb11d">
+                <sanitizer>
+                    <valid initial="string.ascii_letters,string.digits">
```
bgruening: do we really need a sanitizer for type=color?

tools/jbrowse2/macros.xml
hexylena: no, but it's also a bigger and more complex tool than 99%. Would () be better? We want to indicate that most folks probably don't need to look in there, unlike the selects that are more common options. Or prefixing with "Advanced options: " or something like that?

tools/jbrowse2/jbrowse2.xml
```diff
@@ -0,0 +1,1075 @@
...
+                                <add value="="/>
+                                <add value="'"/>
+                                <add value='"'/>
+                                </valid>
+                            </sanitizer>
+                        </param>
+                        <expand macro="track_visibility" />
+                    </when>
+
+                    <when value="gc">
```
hexylena, line 482: is this missing an input?

tools/jbrowse2/macros.xml
hexylena: GC data could be visualised with other track types I guess, but I feel less certain anyone would want to use anything other than the builtin gc here.

tools/jbrowse2/macros.xml
hexylena: I've only found the one MAF plugin @abretaud maybe you know of others? otherwise yeah, I can see @bgruening's argument :)

abretaud: Yep there are many other plugins in https://jbrowse.org/jb2/plugin_store/, I have added to the TODO list (in readme) to see if we can add more in the tool (but later :smile: )
Also you can easily add custom plugins https://jbrowse.org/jb2/docs/config_guides/customizing_feature_colors/

tools/jbrowse2/jbrowse2.xml
abretaud: Nope, gc track is auto generated by jbrowse from the sequence. I added it as a specific track type though to let users decide in which category they want to place it + possibly add more styling options at some point

abretaud:
Thanks for the comments!

For the sim_size/asserts: I was planning to check with a full diff, but there are many paths and md5 hashes in the output which made it unusable. However these paths and hashes have a constant length which means the sim_size should be pretty reliable to check if you get the expected output. Not so fond of using asserts as it would be hard to check that the json structure is correct while avoiding portions where there are paths/hashes. Anyway we can change that later if it proved to be inconvenient

For "[Advanced]" I proposed something, let me know how you feel about it (+1 to make it clear that it's advanced)

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/18189601014

abretaud:
Fixed now, thanks thanks thanks, quite happy to have it merged now!

PR #4001, opened 2021-10-01
Updating tools/pangolin from version 3.1.11 to 3.1.14

planemo-autoupdate:

I have updated tools/pangolin from version 3.1.11 to 3.1.14.


bernt-matthias (approved):
Changes in pangolin seem to cover only Bugfixes.

PR #4005, opened 2021-10-04
Let PR workflow check file sizes

bernt-matthias:
Large test data should be avoided. So lets automatize this in the PR workflow. .. Sometimes it can even create problems: https://github.com/galaxyproject/galaxy/pull/12604

At the moment we have 410 files > 500k and 225 files > 1M

We can argue if we want to fail `determine-success` if large files are found.

FOR CONTRIBUTOR:

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -244,7 +244,18 @@ jobs:
       with:
         name: 'File size report'
         path: file_size_report.txt
-
+    - name: PR Comment
+      uses: actions/github-script@v2
+      with:
+        github-token: ${{ secrets.PAT }}
+        script: |
+          github.issues.createComment({
+            issue_number: ${{ github.event.number }},
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            body: ':warning: Some files in this PR are large. Try to reduce the size, e.g. by using content assertions.'
+          })
+  
   # Planemo test the changed repositories, each chunk creates an artifact
   # containing HTML and JSON reports for the executed tests
   test:
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -244,6 +244,14 @@ jobs:
       with:
         name: 'File size report'
         path: file_size_report.txt
+    - name: Comment PR
+      if: ${{ failure() }}
+      uses: peter-evans/create-or-update-comment@v1
+      with:
+        token: ${{ secrets.PAT }}
+        repository: ${{ github.event.client_payload.github.payload.repository.full_name }}
+        issue-number: ${{ github.event.pull_request.number }}
+        body: ':warning: Some files in this PR are large. Try to reduce the size, e.g. by using content assertions.'
 
   # Planemo test the changed repositories, each chunk creates an artifact
   # containing HTML and JSON reports for the executed tests
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -219,7 +219,7 @@ jobs:
   file_sizes:
     name: Check file sizes
     needs: setup
-    if: ${{ github.ref != 'refs/heads/master' && github.ref != 'refs/heads/main' && needs.setup.outputs.repository-list != '' }}
+    if: ${{ github.event_name == 'pull_request' && needs.setup.outputs.repository-list != '' }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -4,6 +4,7 @@ env:
   GALAXY_FORK: galaxyproject
   GALAXY_BRANCH: release_21.09
   MAX_CHUNKS: 4
+  MAX_FILE_SIZE: 500k
 concurrency:
   # group runs by PR, but keep runs on master separate
   # because we do not want to cancel toolshed uploads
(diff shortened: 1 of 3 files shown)


nsoranzo:
> At the moment we have 410 files > 500k and 225 files > 1M

I get a bit different numbers:

```shell
$ find . -size +500k | grep -v '^./.git/' | wc -l
324
$ find . -size +1M | grep -v '^./.git/' | wc -l
153
```

These are the worst offenders:

```shell
$ find . -size +10M | grep -v '^./.git/' | sort | xargs ls -1sh
24M ./tools/feelnc/test-data/genome_chr38.fa
62M ./tools/gatk4/test-data/chr20.fa
15M ./tools/kallisto/test-data/cached_locally/sacCer2_chrX.kallistei
20M ./tools/khmer/test-data/test-abund-read-2.oxlicg
11M ./tools/maxbin2/test-data/3/out.reassem/out.reads.noclass
13M ./tools/maxbin2/test-data/interleavedPE_unmapped_Sample3_small.fasta
15M ./tools/nanoplot/test-data/alignment.bam
47M ./tools/pangolin/test-data/2021-04-21/data/decision_tree_rules.txt
46M ./tools/pangolin/test-data/2021-04-21/data/lineages.metadata.csv
14M ./tools/pygenometracks/test-data/Li_et_al_2015.h5
22M ./tools/ucsc_blat/test-data/amaVit1_Gallus/amaVit1.fa
```

nsoranzo:
Can you add a (temporary) commit adding a large test file to see it at work?

bernt-matthias:
> Can you add a (temporary) commit adding a large test file to see it at work?

Excellent idea. I would expect to get reported at least the following files (which seemed to be the smallest (larger than 500k) and the largest in the repo):

- ./tools/artic/test-data/SRR11410539_seqtk_sample_500_1.fastq
- ./tools/gatk4/test-data/chr20.fa

bernt-matthias:
Looks good:

```
Files larger than 500k found
tools/artic/test-data/SRR11410539_seqtk_sample_500_1.fastq
tools/gatk4/test-data/chr20.fa
Error: Process completed with exit code 1.
```

If you like I could add the size for each file.

Also the artifact is produced.

.github/workflows/pr.yaml
```diff
@@ -215,6 +216,35 @@ jobs:
         name: 'R linting output'
         path: rlint_report.txt
 
+  file_sizes:
+    name: Check file sizes
+    needs: setup
+    if: ${{ needs.setup.outputs.repository-list != '' }}
```
nsoranzo, line 222:
```suggestion
    if: ${{ github.ref != 'refs/heads/master' && github.ref != 'refs/heads/main' && needs.setup.outputs.repository-list != '' }}
```

nsoranzo:
Nice!
Two additional points for discussion:
- This would fail builds also for PRs where the offending large file is pre-existing and not changed.
- Instead of failing the build, the CI job could add a comment to the PR?

bernt-matthias:
> Nice! Two additional points for discussion:
> 
>     * This would fail builds also for PRs where the offending large file is pre-existing and not changed.
> 
>     * Instead of failing the build, the CI job could add a comment to the PR?

Indeed. I also could imagine that there are tools where it is hard/impossible to find smaller test data. Then we should probably still accept it. 

Is there a way to still be able to merge if this job fails? Just remove it from `Check workflow success`?

Comment is also an idea.

.github/workflows/pr.yaml
```diff
@@ -215,6 +216,41 @@ jobs:
         name: 'R linting output'
         path: rlint_report.txt
 
+  file_sizes:
+    name: Check file sizes
+    needs: setup
+    if: ${{ github.ref != 'refs/heads/master' && github.ref != 'refs/heads/main' && needs.setup.outputs.repository-list != '' }}
```
nsoranzo, line 222:
```suggestion
    if: ${{ github.event_name == 'pull_request' && needs.setup.outputs.repository-list != '' }}
```

.github/workflows/pr.yaml
```diff
@@ -215,6 +216,41 @@ jobs:
...
+          exit 1
+        fi
+    - uses: actions/upload-artifact@v2
+      if: ${{ failure() }}
+      with:
+        name: 'File size report'
+        path: file_size_report.txt
+    - name: Comment PR
+      if: ${{ failure() }}
+      uses: thollander/actions-comment-pull-request@v1
```
nsoranzo, line 249: We are already using `peter-evans/create-or-update-comment@v1` in `.github/workflows/ci.yaml` , maybe we could stick with that?

.github/workflows/pr.yaml
```diff
@@ -354,7 +390,7 @@ jobs:
 
   determine-success:
     name: Check workflow success
-    needs: [setup, lint, flake8, lintr, combine_outputs]
+    needs: [setup, lint, flake8, lintr, file_sizes, combine_outputs]
```
nsoranzo, line 393: Should we drop it from here?

nsoranzo:
It may very well be that the comment won't work until this get merged.

bernt-matthias:
> @nsoranzo any idea why the workflow isn't running anymore?

Also this one does not run: https://github.com/galaxyproject/tools-iuc/pull/4010

So maybe its a problem at github?

nsoranzo:
It seems workflows are running now.

bernt-matthias:
From https://github.com/peter-evans/create-or-update-comment: `Note: In public repositories this action does not work in pull_request workflows when triggered by forks.`

PR #4006, opened 2021-10-04
Add the bbtools callvariants tool

gregvonkuster:
FOR CONTRIBUTOR:


tools/bbtools/callvariants.xml
```diff
@@ -0,0 +1,110 @@
...
+    <inputs>
+        <param name="input" type="data" format="bam" label="Bam file" help="Bam files produced by BBMap are recommended"/>
+        <expand macro="reference_source_cond"/>
+        <param name="ploidy" type="integer" value="1" min="0" label="Ploidy" help="See help section below"/>
+        <param name="output_format" type="select" label="Select output format">
+            <option value="vcf" selected="true">vcf</option>
+            <option value="gff">gff</option>
+            <option value="txt">txt</option>
+        </param>
+        <param name="output_variant_score_hist" type="boolean" truevalue="yes" falsevalue="no" checked="false" label="Output variant score histogram?"/>
```
bernt-matthias, line 47: I sometimes prefer a select instead of multiple bool parameter. Because I think it is more compact.. But this is only my preference.

tools/bbtools/callvariants.xml
```diff
@@ -0,0 +1,110 @@
...
+#elif str($output_format) == 'gff':
+    outgff='out.gff' && mv 'out.gff' '$output'
+#else:
+    out='output.txt' && mv 'output.txt' '$output'
+#end if
+]]></command>
+    <inputs>
+        <param name="input" type="data" format="bam" label="Bam file" help="Bam files produced by BBMap are recommended"/>
+        <expand macro="reference_source_cond"/>
+        <param name="ploidy" type="integer" value="1" min="0" label="Ploidy" help="See help section below"/>
```
bernt-matthias, line 41: What is a value of 0 doing here?

gregvonkuster: Setting the lower bounds.  Is this not correct?

tools/bbtools/callvariants.xml
gregvonkuster: Yeah, I've been told the opposite by other reviewers.  The iuc should define a standard rule.

tools/bbtools/callvariants.xml
bernt-matthias: I meant: what does it mean if a user selects a value of 0?

tools/bbtools/callvariants.xml
```diff
@@ -0,0 +1,110 @@
...
+            <option value="txt">txt</option>
+        </param>
+        <param name="output_variant_score_hist" type="boolean" truevalue="yes" falsevalue="no" checked="false" label="Output variant score histogram?"/>
+        <param name="output_zygosity_hist" type="boolean" truevalue="yes" falsevalue="no" checked="false" label="Output zygosity histogram?"/>
+        <param name="output_quality_hist" type="boolean" truevalue="yes" falsevalue="no" checked="false" label="Output quality histogram?"/>
+    </inputs>
+    <outputs>
+        <data name="output" format="vcf">
+            <change_format>
+                <when input="output_format" value="gff" format="gff"/>
```
bernt-matthias, line 54: This should be `gff3`. There is no `gff` datatype.

bernt-matthias: > This should be `gff3`. There is no `gff` datatype.

Sorry, I was wrong about this.

tools/bbtools/callvariants.xml
gregvonkuster: Are you implying that the lower bounds should be 1?

bernt-matthias: > Are you implying that the lower bounds should be 1?

I have no idea :) .. was just wondering.

gregvonkuster: I'm not a biologist , so I have no idea if some cell of some organism can ever be in a state that has a ploidy value of 0.  The bbtools callvariants.sh script does not set a lower bounds, so I cam remove it here if you want.

bernt-matthias: In the sources I found `if(ploidy<1){System.err.println("WARNING: ploidy not set; assuming ploidy=1."); ploidy=1;}`. Can you check if this is triggered if setting the value to 0? If so I would suggest `min="1"`.

gregvonkuster: Setting ploidy to 0 throws this exception, so I changed the min to 1.
```
Processing variants.
Exception in thread "main" java.lang.AssertionError
	at var2.VarMap.processVariantsMT_inner(VarMap.java:254)
	at var2.VarMap.processVariantsMT(VarMap.java:247)
	at var2.CallVariants.processVariants(CallVariants.java:592)
	at var2.CallVariants.process(CallVariants.java:472)
	at var2.CallVariants.main(CallVariants.java:62)
```

bernt-matthias (approved):

PR #4034, opened 2021-10-11
gfa_to_fa: make test data gfa1

bernt-matthias:
it was not according to https://github.com/GFA-spec/GFA-spec/blob/master/GFA1.md
(there are no 'a' lines in gfa1/2). makes no difference in the tool, but
I'm afraid that the wrong test data perpetuates through this / other
repos

FOR CONTRIBUTOR:


bernt-matthias:
> > I'm afraid that the wrong test data perpetuates through this / other repos
> 
> What do you mean here?

I mean't that we could also leave the test data as is and force the dataype to gfa via `ftype`. But this seemed wrong to me in particular because others might use the test file by copy and pasting it.

PR #4041, opened 2021-10-11
deprecate kobas

bernt-matthias:
there are tabs, but its not a proper tab separated file

I would suggest to deprecate the tool, since the download URL of the kobas sqlite db (for the `kobas_db` parameter) is gone. The underlying tool uses python 2.7 and the github repo (https://github.com/xmao/kobas) has not been updated for > 10years.

FOR CONTRIBUTOR:


bernt-matthias:
> Yes, we should probably deprecate the tools, but I don't think we should update the XML before. So either do a last release and then deprecate or just deprecate and not change anything.

Agreed

bgruening (approved):

bgruening:
And deprecated ... :)

Thanks!

PR #4045, opened 2021-10-11
hamronize: use tabular instead of tsv

bernt-matthias:
in summary the test input tsv files are sniffed as tabular
since the input uses mixed datatypes we can not use ftype

I guess using tabular everywhere is the best solution

FOR CONTRIBUTOR:


bernt-matthias:
> I never understood the difference between tsv and tabular :(

Good point, me neither...

PR #4053, opened 2021-10-13
fix #4050

### LINKED ISSUE #4050: qiime error
We got this error from a user.

```
Error in validate_mapping_file.py: no such option: -d

If you need help with QIIME, see:
http://help.qiime.org
```

bgruening:
this might fix #4050


wm75:
@bernt-matthias
https://github.com/galaxyproject/tools-iuc/blob/5685169cb3d003d3e10765e757a8e453aec003ee/tools/qiime/qiime_core/validate_mapping_file.xml#L51-L54
- should planemo maybe catch such cases of pseudo-tests of boolean params?
- should correct use of boolean params in tests be part of IUC recomendations?

bernt-matthias:
>     * should planemo maybe catch such cases of pseudo-tests of boolean params?

Maybe, setting the true/falsevalue as well as true/false works. I guess the former will only fail if `truevalue="false" falsevalue="true"`.

Currently the linting of tests is indepent of the inputs section, i.e. the test linter does not know anything about the inputs .. would be quite nice to change this.

>     * should correct use of boolean params in tests be part of IUC recomendations?

I guess if we would lint for it it comes automatically. But adding this somewhere to the docs would be also really nice.

bernt-matthias:
Lets see if tests run ...

PR #4062, opened 2021-10-13
seqtk mutfa: in_snp is not necesarrily tab separated

bernt-matthias:
so add txt, alternatively refomat test data

FOR CONTRIBUTOR:


bernt-matthias:
> I think using a tabular test file is indeed the better solution here. The tool help is rather explicit: `the SNP inputs consist of 4 columns: chromosome, 1-based-pos, any, base-changed-to.`
> 
> So offering general txt datasets as input seems very liberal.
> 
> Is it enough maybe to just set ftype="tabular" in the test?

I guess than its better to "fix" the test data itself and let the sniffer do its job :)

PR #4082, opened 2021-10-18
Updating tools/abyss from version 2.3.1 to 2.3.2

ghost:
Hello! This is an automated update of the following tool: **tools/abyss**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

I have updated tools/abyss from version 2.3.1 to 2.3.2.

**Project home page:** http://www.bcgsc.ca/platform/bioinfo/software/abyss


bernt-matthias:
Some parameters may need updates https://github.com/bcgsc/abyss/releases/tag/2.3.2

PR #4125, opened 2021-10-29
Trinity: scratch folder has to be unique to each Trinity instance 

gmauro:
Trinity wrapper can use a scratch directory defined by the variable $TRINITY_SCRATCH_DIR but, in the current implementation, could happen that multiple Trinity instance running at the same time can overwrite results in that common path.
I am using mktemp to enforce a directory creation with a unique name  to be used as scratch dir per each instance.

Trinity has also the bad attitude to create millions of files in the same directory (we have had a dir with 9M), so I added a cleaning task using rsync for ensuring better performances.

Added coreutils and rsync to the dependencies.

FOR CONTRIBUTOR:
* [X ] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [X ] - This PR updates an existing tool or tool collection

### DIFF
--- a/tools/trinity/macros.xml
+++ b/tools/trinity/macros.xml
@@ -3,8 +3,6 @@
     <xml name="requirements">
         <requirements>
             <requirement type="package" version="@WRAPPER_VERSION@">trinity</requirement>
-            <requirement type="package" version="8.32">coreutils</requirement>
-            <requirement type="package" version="3.2.3">rsync</requirement>
             <yield/>
         </requirements>
     </xml>
--- a/tools/trinity/macros.xml
+++ b/tools/trinity/macros.xml
@@ -3,6 +3,8 @@
     <xml name="requirements">
         <requirements>
             <requirement type="package" version="@WRAPPER_VERSION@">trinity</requirement>
+            <requirement type="package" version="8.32">coreutils</requirement>
+            <requirement type="package" version="3.2.3">rsync</requirement>
             <yield/>
         </requirements>
     </xml>
--- a/tools/trinity/trinity.xml
+++ b/tools/trinity/trinity.xml
@@ -1,4 +1,4 @@
-<tool id="trinity" name="Trinity" version="@WRAPPER_VERSION@+galaxy1">
+<tool id="trinity" name="Trinity" version="@WRAPPER_VERSION@+galaxy2">
     <description>de novo assembly of RNA-Seq data</description>
     <expand macro="bio_tools"/>
     <macros>
@@ -14,7 +14,8 @@
 
         if [ ! -z "\$TRINITY_SCRATCH_DIR" ] ; then
             workdir=`pwd` ;
-            cd "\$TRINITY_SCRATCH_DIR" ;
+            scratchfolder=`mktemp -d -p \$TRINITY_SCRATCH_DIR` ;
+            cd "\$scratchfolder" ;
         fi ;
 
         #if $additional_params.guided.is_guided == "yes":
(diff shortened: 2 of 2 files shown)


tools/trinity/macros.xml
```diff
@@ -3,6 +3,8 @@
     <xml name="requirements">
         <requirements>
             <requirement type="package" version="@WRAPPER_VERSION@">trinity</requirement>
+            <requirement type="package" version="8.32">coreutils</requirement>
+            <requirement type="package" version="3.2.3">rsync</requirement>
```
wm75, line 7: While this combination of requirements would be fine, some xml wrappers in the folder are adding additional ones that create conflicts conda can't resolve.
I think you can simply move the two new requirements into the trinitiy.xml requirements section specifically to make conda happy.
See for example *align_and_estimate_abundance.xml* on how to add wrapper-specific requirements on top of the macros-defined general ones.

tools/trinity/trinity.xml
```diff
@@ -149,6 +150,10 @@
         if [ ! -z "\$TRINITY_SCRATCH_DIR" ] ; then
             mkdir -p "\$workdir/trinity_out_dir";
             cp -p trinity_out_dir/Trinity* "\$workdir/trinity_out_dir";
+            mkdir -p "\$TRINITY_SCRATCH_DIR/empty_dir";
+            cd "\$TRINITY_SCRATCH_DIR";
+            rsync -a --delete "\$TRINITY_SCRATCH_DIR/empty_dir/ \$scratchfolder/";
```
wm75, line 155: I assume "\$TRINITY_SCRATCH_DIR/empty_dir/ \$scratchfolder/" should be broken into two arguments?

tools/trinity/trinity.xml
```diff
@@ -149,6 +150,10 @@
...
+            rmdir "\$TRINITY_SCRATCH_DIR/empty_dir \$scratchfolder/";
```
wm75, line 156: same here?

tools/trinity/macros.xml
gmauro: Oh, sure.

tools/trinity/trinity.xml
```diff
@@ -149,6 +153,10 @@
         if [ ! -z "\$TRINITY_SCRATCH_DIR" ] ; then
             mkdir -p "\$workdir/trinity_out_dir";
             cp -p trinity_out_dir/Trinity* "\$workdir/trinity_out_dir";
+            mkdir -p "\$TRINITY_SCRATCH_DIR/empty_dir";
```
wm75, line 156: Shouldn't the empty dir also be job-specific?
Can rsync sync an outer with an inner dir maybe?

wm75: Otherwise you could create it simply in the job working dir

gmauro: They are at the same level but, can happen that a job removes it while another is syncing.

gmauro: job working dir can be a completely different file system. I will arrange differently

wm75: `$ cat test_del_from_inner.sh`

```
mkdir -p test_folder/empty
touch test_folder/test_file
ls -lha test_folder
rsync -av --delete test_folder/empty/ test_folder/
ls -lha test_folder
rmdir test_folder
```

`$ sh test_del_from_inner.sh`

```
total 12K
drwxrwxr-x. 3 wolma wolma 4,0K  4. Nov 20:21 .
drwxr-xr-x. 5 wolma wolma 4,0K  4. Nov 20:21 ..
drwxrwxr-x. 2 wolma wolma 4,0K  4. Nov 20:21 empty
-rw-rw-r--. 1 wolma wolma    0  4. Nov 20:21 test_file
sending incremental file list
deleting empty/
deleting test_file

sent 55 bytes  received 35 bytes  180.00 bytes/sec
total size is 0  speedup is 0.00
total 8,0K
drwxrwxr-x. 2 wolma wolma 4,0K  4. Nov 20:21 .
drwxr-xr-x. 5 wolma wolma 4,0K  4. Nov 20:21 ..
```

Deleting an outer dir by syncing it with an empty inner dir seems to work, which would simplify your code a bit.
Up to you though whether you want to use that behavior.

wm75:
I guess you should add a comment explaining why removing the scratch_dir via rsync is necessary.

wm75 (approved):
Cool! Thanks a lot @gmauro for tackling this.

Feel free to consider my comment https://github.com/galaxyproject/tools-iuc/pull/4125#discussion_r743140254 or stay with the current, less surprising empty folder location as you like.

gmauro:
I feel a bit weird using an inner empty dir but, it is a personal taste, indeed.
Thanks for your time @wm75

PR #4128, opened 2021-10-30
Integrating MaAsLin2 into Galaxy

EngyNasr:
Adding [**MaAsLin2**](http://huttenhower.sph.harvard.edu/maaslin) tool: which is comprehensive R package for efficiently determining multivariable association between phenotypes, environments, exposures, covariates and microbial meta’omic features. MaAsLin2 relies on general linear models to accommodate most modern epidemiological study designs, including cross-sectional and longitudinal, and offers a variety of data exploration, normalization, and transformation methods.

This wrapper includes: 1 tool xml file, 1 macros, 1 yml file and the test data folder. 

I need guidance in the boolean arguments, which is not a big problem but better if improved, the false value is set to FALSE however when the button is set to no (in the local use galaxy) the false value is not given to the command.  Any advice ?
I handled that with filtering the output later on in the code.

Thanks a lot,
Engy :)


EngyNasr:
Doing: Reducing the test data size and updating tests accordingly

tools/maaslin2/maaslin2.xml
```diff
@@ -0,0 +1,373 @@
...
+                <option value="CPLM">CPLM</option>
+                <option value="NEGBIN">NEGBIN</option>
+                <option value="ZINB">ZINB</option>
+            </param>
+            <param argument="--correction" type="text" value="BH" optional="true" label="Correction" help="The correction method for computing  the q-value"/>
+            <param argument="--standardize" type="boolean" truevalue="TRUE" falsevalue="FALSE" checked="true" label="Apply z-score so continuous metadata are on  the same scale"/>
+            <param argument="--cores" type="integer" value="1" optional="true" label="Cores" help="The number of R processes to run in parallel"/>        
+        </section>          
+        <section name="output" title="Set Plotting Output" expanded="true">
+            <param argument="--plot_heatmap" type="boolean" truevalue="TRUE" falsevalue="FALSE" checked="true" label="Generate a heatmap for the significant  associations"/>
```
bgruening, line 95: How does the commandline need to look like? I assume simply `--plot_heatmap` and not `--plot_heatmap TRUE`?

If so try this.
```suggestion
            <param argument="--plot_heatmap" type="boolean" truevalue="--plot_heatmap" falsevalue="" checked="true" label="Generate a heatmap for the significant  associations"/>
```

tools/maaslin2/maaslin2.xml
```diff
@@ -0,0 +1,373 @@
...
+    --fixed_effects '$fixed_effects'
+#end if
+#if $additional_options.correction
+    --correction '$additional_options.correction'
+#end if
+#if $additional_options.standardize
+    --standardize '$additional_options.standardize'
+#end if
+#if $output.plot_heatmap
+    --plot_heatmap '$output.plot_heatmap'
```
bgruening, line 45: Replace the entire if with "$output.plot_heatmap" if you can use my suggestion below.

tools/maaslin2/maaslin2.xml
```diff
@@ -0,0 +1,373 @@
...
+        <section name="output" title="Set Plotting Output" expanded="true">
+            <param argument="--plot_heatmap" type="boolean" truevalue="TRUE" falsevalue="FALSE" checked="true" label="Generate a heatmap for the significant  associations"/>
+            <param argument="--heatmap_first_n" type="integer" value="50" optional="true" label="Heatmap plot first N" help="In heatmap, plot top N features with significant associations"/>
+            <param argument="--plot_scatter" type="boolean" truevalue="TRUE" falsevalue="FALSE" checked="true" label="Generate scatter plots for the significant associations"/>    
+        </section>   
+   </inputs>
+   <outputs>    
+        <data name="all_results" format="tsv" from_work_dir="outputFolder/all_results.tsv" label="All results ordered by increasing q-value"/>
+        <data name="significant_results" format="tsv" from_work_dir="outputFolder/significant_results.tsv" label="Q-values smaller than or equal to the threshold"/>
+        <data name="residuals" format="rdata" from_work_dir="outputFolder/residuals.rds" label="Data frame with residuals for each feature"/>
```
bgruening, line 103: is this output needed? Should it be optional?

tools/maaslin2/maaslin2.xml
```diff
@@ -0,0 +1,373 @@
...
+   <outputs>    
+        <data name="all_results" format="tsv" from_work_dir="outputFolder/all_results.tsv" label="All results ordered by increasing q-value"/>
+        <data name="significant_results" format="tsv" from_work_dir="outputFolder/significant_results.tsv" label="Q-values smaller than or equal to the threshold"/>
+        <data name="residuals" format="rdata" from_work_dir="outputFolder/residuals.rds" label="Data frame with residuals for each feature"/>       
+        <data format="pdf" name="headmap" from_work_dir="outputFolder/figures/heatmap.pdf" label="Heatmap of the significant associations" >
+            <filter>(output['plot_heatmap'] is True)</filter>
+        </data>        
+        <collection name="figures_pdfs" type="list" label="Plots" >
+            <discover_datasets pattern="(?P&lt;designation&gt;.+)" directory="outputFolder/figures/" format="pdf"/>
+            <filter>(output['plot_scatter'] is True)</filter>
```
bgruening, line 109: the brackets are not needed imho

tools/maaslin2/maaslin2.xml
```diff
@@ -0,0 +1,373 @@
...
+    --cores '$additional_options.cores'
+#end if
+    'input_data.tsv'
+    'input_metadata.tsv'
+    'outputFolder'
+&&
+cd outputFolder &&  mkdir -p figures/ && cp *.pdf figures
+    ]]></command>
+    <inputs>
+        <param name="input_data" type="data" format="tsv" label="Data (or features) file"/>
```
bgruening, line 63: please use `tabular`, `tsv` should probably be never used

tools/maaslin2/maaslin2.xml
```diff
@@ -0,0 +1,373 @@
...
+        <param name="input_metadata" type="data" format="tsv" label="Metadata file"/>
+        <param argument="--fixed_effects" type="text" value="diagnosis,dysbiosisnonIBD,dysbiosisUC,dysbiosisCD,antibiotics,age" multiple="true" optional="true" label="Interactions: Fixed effects" help="The fixed effects for the model,  comma-delimited for multiple effects"/>
```
bgruening, line 65: from where does the user get this list? Do we need to add a validator here?

bgruening (commented):
Thanks @EngyNasr

tools/maaslin2/maaslin2.xml
EngyNasr: sorrryy

EngyNasr: I will change it directly

tools/maaslin2/maaslin2.xml
EngyNasr: supposedly, the default value of this argument is ALL so I checked their documentation to know what is this ALL and wrote all options they support such that user by default will have all the list and have the option to remove from it

tools/maaslin2/maaslin2.xml
EngyNasr: it is outputted in all ways, there is no tool's argument that makes it optional, but I can do it

tools/maaslin2/maaslin2.xml
bgruening: In that case you can use `truevalue="--plot_heatmap TRUE"`

tools/maaslin2/maaslin2.xml
bgruening: Yes, this is usually the case. But the Galaxy tool dev can decide if its a good idea to expose all those files by default to the user or not.

tools/maaslin2/maaslin2.xml
bgruening: if this is a fixed set of values and no free text, why not offering a select field here?

tools/maaslin2/maaslin2.xml
EngyNasr: I will do this but that is actually not the only problem, the main problem is when I turn it off, the argument is not given as False, else its just not written in the command, and that for the tool means true and not false

EngyNasr: that solved the problem, thanks a lot :))

EngyNasr:
Can I create a PR now for [step 3](https://galaxyproject.eu/posts/2020/08/22/three-steps-to-galaxify-your-tool/) 🎉 ?

bgruening:
> Can I create a PR now for [step 3](https://galaxyproject.eu/posts/2020/08/22/three-steps-to-galaxify-your-tool/) tada ?

only when this one is merged :) otherwise your step 3 will fail :)

tools/maaslin2/macros.xml
```diff
@@ -0,0 +1,25 @@
...
+            <edam_topic>topic_3305</edam_topic>
+        </edam_topics>           
+    </xml>
+    <xml name="requirements">
+        <requirements>
+            <requirement type="package" version="@TOOL_VERSION@">maaslin2</requirement>
+            <yield/>
+        </requirements>
+    </xml>
+    <token name="@HELP_HEADER@">
```
bgruening, line 19: isn't that redundant with what you have in the other help section?

EngyNasr: I will remove it

bgruening (approved):

bgruening:
Thanks @EngyNasr! Now its time for step 3.

PR #4149, opened 2021-11-08
Updating tools/ruvseq from version 1.26.0 to 1.44.0

ghost:
Hello! This is an automated update of the following tool: **tools/ruvseq**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

I have updated tools/ruvseq from version 1.26.0 to 1.28.0.

**Project home page:** https://www.bioconductor.org/packages/release/bioc/html/DESeq2.html


bernt-matthias:
new parameter `--range`

PR #4157, opened 2021-11-10
funannotate clean: add cpu option

abretaud:
FOR CONTRIBUTOR:

Just adding the --cpus option (it's not in the --help, but it works)

### DIFF
--- a/tools/funannotate/funannotate_clean.xml
+++ b/tools/funannotate/funannotate_clean.xml
@@ -15,7 +15,7 @@ funannotate clean
 --cov ${cov}
 --minlen ${minlen}
 ${exhaustive}
---cpus \${GALAXY_SLOTS:-2}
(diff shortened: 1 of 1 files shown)


pvanheus:
You have a test file that is too large (> 500K), but besides that, why is the default for the CPU option set to 2?

hexylena (approved):

abretaud:
> You have a test file that is too large (> 500K), but besides that, why is the default for the CPU option set to 2?

Yep for file size, the file is already in the repo and I can't really reduce it
For the default of 2, hum... why not? I see different values for other tools, I don't think we have a standard for this

hexylena:
Defaulting to 1 would retain the older behaviour a bit better maybe, in case there's any parallelization differences from the previous version?

PR #4164, opened 2021-11-12
Fix samtools wrappers to work with fasta.gz refs

wm75:
Galaxy's fasta.gz datatype is currently used for gzipped and bgzipped
fasta files alike, but samtools faidx can only handle the latter
directly.
Until Galaxy starts supporting a dedicated fasta.bgzip datatype, this
change works around the existing datatype's ambiguity with an
"easier to ask for forgiveness than permission" approach letting
samtools faidx try to index the file, then, if that fails, try again with the
uncompressed (likely regular gzipped) file.

FOR CONTRIBUTOR:


wm75:
@bernt-matthias if you like this solution, feel free to borrow it for https://github.com/galaxyproject/tools-iuc/pull/3549 :)

wm75:
The disadvantage of this workaround is that the standard user that passes in a gzip-compressed fasta.gz file will not be alerted to the fact that they are doing something rather inefficient, but at least things will just work for now.
The dedicated fasta.bgzip datatype is the cleaner long-term solution though, imo.

tool_collections/samtools/macros.xml
```diff
@@ -50,23 +50,56 @@
...
+                        samtools faidx \$reffa;
+                    }
+                } &&
+            #end if
+            reffai=\$reffa.fai &&
         #elif $addref_cond.addref_select == "cached":
-            #set reffa=str($addref_cond.ref.fields.path)
-            #set reffai=str($addref_cond.ref.fields.path)+".fai"
+            ## in case of cached the absolute path is used which allows to read 
+            ## a cram file  without specifying the reference
```
wm75, line 98: I didn't really understand this comment so I decided to keep it, but move it to the relevant section of the code.

wm75:
Very weird linting problem:

https://github.com/galaxyproject/tools-iuc/runs/4191216022?check_suite_focus=true#step:5:346

why does it think there is an optional="false" combined with checkboxes, when that is not present in the xmls?

bernt-matthias:
> Very weird linting problem:
> 
> https://github.com/galaxyproject/tools-iuc/runs/4191216022?check_suite_focus=true#step:5:346
> 
> why does it think there is an optional="false" combined with checkboxes, when that is not present in the xmls?

Indeed. Its a minor bug in the linter: https://github.com/galaxyproject/galaxy/pull/12920

bgruening:
And lets retrigger the tests.

bernt-matthias (approved):

bernt-matthias:
https://github.com/galaxyproject/galaxy/issues/12995

PR #4167, opened 2021-11-15
Updating tools/shasta from version 0.6.0 to 0.14.0

ghost:
Hello! This is an automated update of the following tool: **tools/shasta**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

I have updated tools/shasta from version 0.6.0 to 0.8.0.

**Project home page:** https://github.com/chanzuckerberg/shasta/releases

### DIFF
--- a/tools/shasta/macros.xml
+++ b/tools/shasta/macros.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <macros>
-    <token name="@TOOL_VERSION@">0.6.0</token>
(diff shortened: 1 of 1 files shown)


bernt-matthias:
Fixes https://github.com/galaxyproject/tools-iuc/issues/4454

mthang:
it seems like the merging of the code is blocked. Any idea?

bernt-matthias:
@mthang this is because tests are failing.

Issue #4090, opened 2021-10-20
Fix mandatory selects with checkboxes display


bernt-matthias:
Linter complains: `display="checkboxes"` is incompatible with `optional="false"`, remove the `display` attribute. 

The problem is that select parameters are by default `optional="false"`. If `display="checkboxes"` is chosen the user can unselect all options which contradicts that the parameter is mandatory.

I think that most of the cases are easy to solve, 

- If the code using the corresponding parameter checks for `None` then the parameter was probably intended to be `optional="true"` and we can leave checkboxes. Otherwise remove checkboxes.
- For other cases I think its necessary to check the underlying application if selecting no options is fine / not.

Here is the list of tools that currently suffer from this (please put your name / link to the PR next to the tool if you are working on this): 

- [ ] tools/kobas/kobas_identify.xml
  - db
- [ ] tools/breseq/breseq.xml
  - formats
- [ ] tools/pear/pear.xml
  - outputs

PR #4172, opened 2021-11-16
fix select that must not be checkboxes

bernt-matthias:
fixes https://github.com/galaxyproject/tools-iuc/issues/4090

FOR CONTRIBUTOR:

### DIFF
--- a/tools/pear/pear.xml
+++ b/tools/pear/pear.xml
@@ -3,6 +3,22 @@
     <macros>
         <token name="@TOOL_VERSION@">0.9.6</token>
         <token name="@VERSION_SUFFIX@">2</token>
+        <xml name="format_action">
+            <actions>
+                <conditional name="library.type">
+                    <when value="paired">
+                        <action type="format">
+                            <option type="from_param" name="library.forward" param_attribute="ext" />
+                        </action>
+                    </when>
+                    <when value="paired_collection">
+                        <action type="format">
+                            <option type="from_param" name="library.input_collection" param_attribute="forward.ext" />
+                        </action>
+                    </when>
+                </conditional>
+            </actions>
+        </xml>
     </macros>
     <xrefs>
         <xref type="bio.tools">pear</xref>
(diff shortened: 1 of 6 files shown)


bgruening:
We need bumps here, do we?

tools/pear/pear.xml
```diff
@@ -50,6 +54,18 @@
...
         $nbase
+        ## set output datatypes using galaxy.json
+        #if str( $library.type ) == "paired":
+            #set ext = $library.forward.ext
+        #else
+            #set ext = $library.input_collection.ext
+        #end if
+        #import json
+        #set gxy_json = {}
+        #for o in $outputs:
```
bgruening, line 65: I'm missing the context here, Is that really necessary? Have we ever done this in IUC before?

bernt-matthias: Do you mean using `galaxy.json`? Pretty sure, we did. But let me undo the changes and see what fails .. if anything.. I can't recall why I did these changes.

bernt-matthias: Maybe because due to the new profile the `format="input"` does not work anymore ... from tool xml docs:

_Disable format="input" and require explicit metadata targets (metadata_source, format_source)._ 

using `format_source` seems tricky due to the conditional.

bgruening: I don't have a better idea, but this looks like a major hack. Do we really encourage people to hack the galaxy.json on the fly?

bgruening:
This now needs a rebase.

tools/pear/pear.xml
bernt-matthias: The alternative would be to use `format_source` which would imply to have either

- separate outputs depending on the input type (pair/paired collection).
- or maybe use https://docs.galaxyproject.org/en/master/dev/schema.html#tool-outputs-data-actions-action ... which I never understood :(

bernt-matthias: **But** maybe the complete `paired_collection` branch makes no sense at all .. there is no loop over the collection in the code.. I doubt that this works .. test is missing .. could try to add one to verify ..

bernt-matthias:
@bgruening  pear now uses output actions instead of `galaxy.json`. The linter problem should be fixed here https://github.com/galaxyproject/galaxy/pull/13583 (but I guess we need to ignore it since it will need a while until this change finds its way to `planemo lint`)

bernt-matthias:
> I'm fine ignoring it, but the test also fails left?

Ups. My bad. Fixed it and also reverted changes to kobas which is now deprecated.

bgruening (approved):
Thanks @bernt-matthias.

Maybe your galaxy fixes will make the unix cat wrapper now pass :-)

bgruening:
I had a button for merging... But you probably need to push manually.

bernt-matthias:
Hmm.. Of course this isn't deployed when linting fails..

PR #4177, opened 2021-11-17
Add the vSNP sample names tool

gregvonkuster:
FOR CONTRIBUTOR:


gregvonkuster:
Tests all pass, just file size linting is failing.

bernt-matthias:
Isn't this what https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/iuc/collection_element_identifiers/collection_element_identifiers/0.0.2 does?

gregvonkuster:
> Isn't this what https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/iuc/collection_element_identifiers/collection_element_identifiers/0.0.2 does?

Not quite.  That tool only allows for an input collection - list or list:paired.
```
    <inputs>
        <param name="input_collection" type="data_collection" collection_type="list,list:paired" label="Dataset collection"/>
    </inputs>
```
I need to handle 3 input types, single file, paired reads in separate files, or a list:paired collection.

bernt-matthias:
Wouldn't it be an option to put the single dataset in a list. Would have the advantage that the identifier is kept when the dataset is processed.

gregvonkuster:
> Wouldn't it be an option to put the single dataset in a list. Would have the advantage that the identifier is kept when the dataset is processed.

Do you mean change this tool wrapper to not accept a single dataset on input, but a list?  Or do you mean use the "Build List from one or more datasets" tool, or some other tool, to add the dataset to a list as part of the workflow?

gregvonkuster:
@bgruening and @bernt-matthias what are your thoughts on adding this tool to this repo?  No other tool (or combination of tools as far as I've seen) does what this one does.

gregvonkuster:
Should I close this PR?

bgruening:
@gregvonkuster I do not know. I would need to play with those tools and the use case. My current thinking goes in the same direction of @bernt-matthias ... if we can use collections, and with this element-identifiers, I think the workflow would be more stable, less affected by bad input names etc ... in addition you could use the above mentioned tool that we already have.

PR #4192, opened 2021-11-22
Circos: Include housekeeping configuration

gallardoalba:
FOR CONTRIBUTOR:

This PR includes the housekeeping configuration in the tableviewer.xml file. The reason for this modification is to increase the max_ideogram value up to 210 (previously 200), in order to adapt it to a user request.


hexylena:
Thanks for contributing this! But I think we should do it a bit differently. The existing circos.xml has a `limits` section, I think it might make more sense to copy that, than change the (sensible) limits for everyone. I think I really prefer the idea of keeping the defaults the developer chose, and then letting users can on a case-by-case basis, so we're not stuck bumping them every time a user asks, and so we don't risk (eventually) having expanded them enough we're not catching the silly mistakes users expect.

Maybe we can move that from circos.xml to the macros, and then re-use in tableviewer?

gallardoalba:
> Thanks for contributing this! But I think we should do it a bit differently. The existing circos.xml has a `limits` section, I think it might make more sense to copy that, than change the (sensible) limits for everyone. I think I really prefer the idea of keeping the defaults the developer chose, and then letting users can on a case-by-case basis, so we're not stuck bumping them every time a user asks, and so we don't risk (eventually) having expanded them enough we're not catching the silly mistakes users expect.
> 
> Maybe we can move that from circos.xml to the macros, and then re-use in tableviewer?

Sure; I wasn't sure how to include it.

hexylena:
I'll update the PR!

hexylena:
I've added a quick test case too, let's see if it works!

hexylena:
It's passing despite the limit, unexpected.

image

probably because it's not passed along the limit I set.

image

gallardoalba:
Do we need to include some assertion_content for those files?

hexylena:
> Do we need to include some assertion_content for those files?

Just a stupid mistake on my end, no, no assertion_content since the test should fail.

It looks like it's passed now though! Fantastic

hexylena (approved):

PR #4228, opened 2021-11-25
Add the MEGAN daa2info, read-extractor, daa-meganizer tools

gregvonkuster:
FOR CONTRIBUTOR:


tools/megan/daa2info.xml
```diff
@@ -0,0 +1,149 @@
...
+#end if
+#if $extractSummaryFile:
+    --extractSummaryFile '$output_summary'
+#end if
+--out '${output}'
+]]></command>
+    <inputs>
+        <param name="input" argument="--in" type="data" format="daa" label="Input DIAMOND file"/>
+        <conditional name="input_is_meganized_cond">
+            <param name="input_is_meganized" type="boolean" truevalue="true" falsevalue="false" checked="false" label="Input is meganized?" help="Implies the input file was produced by the MEGAN: DAA Meganizer tool"/>
```
bgruening, line 43: please no boolean here

tools/megan/daa2info.xml
```diff
@@ -0,0 +1,149 @@
...
+            <when value="false"/>
+            <when value="true">
+                <conditional name="list_class2count_cond">
+                    <param name="list_class2count" type="boolean" truevalue="true" falsevalue="false" checked="false" label="List class to count for named classification(s)?"/>
```
bgruening, line 47: no boolean, according to IUC guildelines

tools/megan/read_extractor.xml
```diff
@@ -0,0 +1,115 @@
+<tool id="megan_read_extractor" name="MEGAN: Extract reads" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
+    <description>by classification</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="bio_tools"/>
+    <expand macro="requirements"/>
+    <command detect_errors="exit_code"><![CDATA[
+#if $input.ext.endswith('.gz'):
+    #set input_ext = $input.ext.upper() + '.gz'
```
bgruening, line 10: you create here .DAA.GZ.gz?

tools/megan/read_extractor.xml
```diff
@@ -0,0 +1,115 @@
...
+--gzipOutputFiles 'false'
+## Output must be a directory.
+--output ./output.txt
+&& mv ./output.txt '${output}'
+]]></command>
+    <inputs>
+        <param name="input" argument="--input" type="data" format="daa,rma6" label="Input file"/>
+        <param argument="--frameShiftCorrect" type="boolean" truevalue="true" falsevalue="" checked="false" label="Extract frame-shift corrected reads?" help="Set to Yes only if blast mode of input is BlastX"/>
+        <conditional name="specify_classification_cond">
+            <param name="specify_classification" type="boolean" truevalue="true" falsevalue="false" checked="true" label="Specify classification to use?"/>
```
bgruening, line 45: please no boolean

tools/megan/daa2info.xml
gregvonkuster: Ah, I'll change this, but I don't understand the reasoning in the following guideline (of which I wasn't aware until now - sorry I missed it):
```
Options that are conditionally hidden (using the <conditional> element) should use a select param type and not a boolean. The user may not expect a boolean checkbox to change the content of a form.
```
Booleans in tool forms are not checkboxes, but Yes / No switches:

image

If flipping the switch from one option to another, form refreshes occur exactly like flipping the options in a select list:
image

gregvonkuster: Although I'm currently in Safari - not sure if booleans look different in different browsers...

nsoranzo: The guidelines were written when boolean were still shown as checkboxes, but the reasoning still applies in my opinion (that a switch making new options appear can be confusing / go unnoticed).

gregvonkuster: Thanks for the context - I'll get this fixed asap.

tools/megan/daa2info.xml
```diff
@@ -0,0 +1,149 @@
...
+                    <when value="false"/>
+                    <when value="true">
+                        <param argument="--class2count" type="select" label="Select class to count for named classification(s)">
+                            <expand macro="classification_options"/>
+                        </param>
+                        <param argument="--sum" type="boolean" truevalue="--sum" falsevalue="" checked="false" label="Use summarized rather than assigned counts when listing class to count?"/>
+                    </when>
+                </conditional>
+                <conditional name="list_read2class_cond">
```
bernt-matthias, line 56: Instead of this conditional + select you may simply use a select with one additional option, e.g. `"No classification"`

Maybe also for the other cases here .. not sure.

Problem with conditionals is always, that you can not change them in workflows (in all cases .. not sure)

gregvonkuster: This will work with some conditionals, but I don't see it working for this one due to this param which should not be displayed when the value is no.
```
<param argument="--sum" type="boolean" truevalue="--sum" falsevalue="" checked="false" label="Use summarized rather than assigned counts when listing class to count?"/>
```

tools/megan/read_extractor.xml
gregvonkuster: Fixed - thanks!

tools/megan/daa2info.xml
gregvonkuster: Fixed - thanks!

tools/megan/daa2info.xml
gregvonkuster: Fixed - thanks!

tools/megan/read_extractor.xml
gregvonkuster: Fixed - thanks!

tools/megan/daa_meganizer.xml
```diff
@@ -0,0 +1,88 @@
+<tool id="daa_meganizer" name="MEGAN: meganize a DIAMOND file" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
```
bgruening, line 1: prefix this tool-id with megan again?

gregvonkuster: Ah!  Fixed - thanks for catching.

bgruening (approved):

PR #4236, opened 2021-11-26
raceid: parameter fix

bernt-matthias:
- clustering: rfseed is commented in the command file

also rename version tokens

FOR CONTRIBUTOR:
* [x] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [x] - License permits unrestricted use (educational + commercial)
* [ ] - This PR adds a new tool or tool collection
* [x] - This PR updates an existing tool or tool collection
* [ ] - This PR does something else (explain below)

PR #4253, opened 2021-11-27
Raceid unused param

mtekman:
FOR CONTRIBUTOR:

Fork of: https://github.com/galaxyproject/tools-iuc/pull/4236

I was able to reduce the test data for `raceid_filtnormconf.xml`, and `raceid_clustering.xml`, but not the test-data for the other XML files

The problem is that the tree generation seems to use the raw data, and it's not clear which features it uses. I tried many tricks including randomly subsetting and testing, and reducing the overall decimal places in each internal sparse matrix in the hopes that it would compress better.... but no, I was unable to get it under 2MB.

@bgruening  is it okay to allow these two files:

* out_cluster_default.rdat (6.5 MB)
* out_traject_default.ltree.rdat (7.5 MB)

I have no idea how to reduce them


bgruening:
Yes, old tools can keep those files. They are not increasing the bloat of the repo, since they are already in. But now the tests fail.

bernt-matthias:
Jep, as @bgruening wrote: counterintuitively making the test data smaller increases the repo size.

But in case that the test failures are of interest the PR might still be useful.

mtekman:
Ah I see, that makes sense about the file reduction in the git history being in vain haha

In that case all tests appear to pass on my system with the edits I've made w.r.t the file size changes, but likely all tests were passing on @bernt-matthias's original PR too.

Feel free to close this then, in favour of the previous PR :+1:

PR #4254, opened 2021-11-28
added wtdbg2 wrapper

GraceAHall:
FOR CONTRIBUTOR:


bgruening (commented):
Should we update this tool https://github.com/bgruening/galaxytools/blob/master/tools/wtdbg/wtdbg.xml instead or do you think its good to have both tools seperate?

GraceAHall:
Hey Björn! Hope you're well 

Since its the same tool, I think we should update your wrapper. What do you think? 
I actually didn't see your wrapper when looking on the toolshed, but I guess that's because the package is wtdbg rather than wtdbg2. 

From what I can see, there are a lot of big differences between version 2.0 (your wrapper) and 2.5 (this one). 
Lots of options have been changed, added or removed, sequencing technology presets have been added, and the fastq/fasta file inputs now need to be in the order of fastqs then fastas if using multiple files. Since tool versions are independent, maybe this could just be the tool version for 2.5? 

That said, I really like some of the extra details you have added in your wrapper, so I would definitely add those to this version if ok.

bgruening:
Sure this is super ok with me. The main advantage I'm after is what we call "tool-linage" if we "upgrade" one tool, Galaxy displays just one tool but you can switch between versions of that one tool - browse the tool linage :)

If we create a new tool, Galaxy will treat them as two independent tools - which might confuse users.

GraceAHall:
Hey Björn! 

Lets go with that then. Should I make a PR with the updates in https://github.com/bgruening/galaxytools/blob/master/tools/wtdbg/wtdbg.xml?  

What do you want to do in regards to the tool version and wrapper version? 
Version = 2.5 and wrapper = 0 I assume.

bgruening:
> Lets go with that then. Should I make a PR with the updates in https://github.com/bgruening/galaxytools/blob/master/tools/wtdbg/wtdbg.xml?

Yes that's probably the easiest option.

> Version = 2.5 and wrapper = 0 I assume.

Sounds good :)

PR #4255, opened 2021-11-29
added necat wrapper

GraceAHall:
FOR CONTRIBUTOR:


tools/necat/macros.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0"?>
+
+<macros>
+    <token name="@TOOL_VERSION@">0.0.1_update20200803</token>
+    <token name="@VERSION_SUFFIX@">0</token>
+    <xml name="job_configfile_macro">
```
bgruening, line 6: macro suffix could be removed I think

tools/necat/necat.xml
```diff
@@ -0,0 +1,180 @@
...
+                <option value="yes">Perform assembly on corrected reads</option>
+            </param>
+            <when value="no" />
+            <when value="yes">
+                <param name="assembly_coverage" type="integer" value="30" label="Assembly coverage" help="Number of reads to use in genome assembly in terms of genome coverage" />
+                <param name="polish_contigs" type="boolean" checked="true" truevalue="true" falsevalue="false" label="Polish contigs" help="Polish contigs as final step after briding" />
+            </when>
+        </conditional>
+        <section name="advanced" title="Advanced" expanded="false" help="Warning: only change these if you really know what you are doing">
+            <param name="overlap_fast_options" type="text" value="-n 500 -z 20 -b 2000 -e 0.5 -j 0 -u 1 -a 1000" label="OVLP_FAST_OPTIONS" help="Overlap fast options">
```
bgruening, line 60: Maybe it would be better to explicitly put those 7 parameters explicitly (as seperate params) in a macro, or are there more parameters that a user can add?

bgruening (commented):
Thanks @GraceAHall. Small review inline.

tools/necat/necat.xml
```diff
@@ -0,0 +1,180 @@
...
+            #if $ext == "fastqsanger"
+                #set $ext = "fastq"
+            #end if
+            #set filename = "reads_" + str($i) + "." + $ext
+            #return $filename
+        #end def
+
+        ## push each input file and everything in input collections into read_list.txt
+        #set i = 1
+        #for input in $input_fastqs
```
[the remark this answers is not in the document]
GraceAHall: Love this!

tools/necat/necat.xml
```diff
@@ -0,0 +1,180 @@
...
+            #set filename = $make_filename($i, $input)
+            cp '$input' $filename  
+            && echo $filename >> read_list.txt &&
+            #set i = $i + 1
+        #end for
+
+        ## necat commands
+        necat correct '${job_configfile}' 
+        ## && gunzip project/1-consensus/cns_final.fasta.gz
```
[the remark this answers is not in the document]
GraceAHall: ah haha yes meant to

tools/necat/necat.xml
```diff
@@ -0,0 +1,180 @@
...
+        #if $assembly.should_assemble == "yes":
+            && necat assemble '${job_configfile}'
+            && necat bridge '${job_configfile}' 
+        #end if
+    ]]></command>
+    <configfiles>   
+        <expand macro="job_configfile_macro" />
+    </configfiles>
+    <inputs>
+        <param name="input_fastqs" type="data" format="fastq,fasta" multiple="true" label="input reads" help="Input read files (fastq or fasta). To select more than one file or collection from your history, use the 'ctrl' key" />
```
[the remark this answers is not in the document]
GraceAHall: Yes seems you are right! I just checked on command line. Will update.

tools/necat/necat.xml
GraceAHall: I believe there are more that a user can add, and these are the defaults. The documentation does not explain tool options well, so I am really not sure. I thought it was better to keep it broad, knowing that only advanced users would set these params.

tools/necat/macros.xml
GraceAHall: the @VERSION_SUFFIX@ token?

bgruening: This is really nitpicking I would just name this "job_configfile"

tools/necat/necat.xml
bgruening: I guess we can leave it like that. Let's see if others have an opinion @galaxyproject/iuc 
In general, we try to avoid those things and try to be as explicit as possible with the options. Also for security risks. Since this is using the config file and not exposing the text on the CMD I guess this is safe.

tools/necat/.shed.yml
```diff
@@ -0,0 +1,13 @@
+categories: 
+- Assembly
+description: Error correction and de-novo assembly for ONT Nanopore reads
+homepage_url: https://github.com/xiaochuanle/NECAT
+long_description: |
+    NECAT performs error correction to remove complex errors in nanopore reads. It can also optionally de novo assembly. 
+
+    See https://doi.org/10.1038/s41467-020-20236-7 for more information
```
gallardoalba, line 8: Not sure if this is necessary.

tools/necat/necat.xml
hexylena: > Since this is using the config file and not exposing the text on the CMD I guess this is safe.

but doesn't mean that the tool doesn't turn around and use this value directly on the command line in a subprocess or so. Which I checked, it does (ugh). 
- var used https://github.com/xiaochuanle/NECAT/blob/0c5fc856aedc855c96efb78d8340372104cb5cfe/src/pipeline/necat.pl#L165
- and stuffed in a subprocess https://github.com/xiaochuanle/NECAT/blob/0c5fc856aedc855c96efb78d8340372104cb5cfe/src/pipeline/necat.pl#L197

Not sure what the escape opportunities are like there. But given that `taint` mode isn't enabled I'm dubious. They use `>` in other pipelines (which would be caught by sanitiser here) so it can read and write to files. With `.` but not `/` you _probably_ can't get an arbitrary file read.

I think I'd be more comfortable with explicit options too. If it helps @GraceAHall I compiled the tool and since these are getting passed to `oc2pmov` I think this should be the full list of options available :)

```
$ Linux-amd64/bin/oc2pmov
USAGE:
Linux-amd64/bin/oc2pmov [options] wrk-dir volume-id output

OPTIONS AND DESCRIPTIONS:
-k <Integer>	kmer size
-z <Integer>	scan window size
-q <Integer>	kmer occurs > q times will be ignored
-b <Integer>	block size
-n <Integer>	number of candidates
-a <Integer>	min align length
-d <Real>	ddf score cutoff
-e <Real>	sequencing error
-m <Integer>	number of output
-t <Integer>	number of cpu threads
-j <0 or 1>	job: 0 = find candidates only, 1 = perform alignemnt
-u <0 or 1>	output binary results: 0 = no, 1 = yes
-i <0 or 1>	use header as sequence id: 0 = no, 1 = yes

DEFAULT OPTIONS:
-k 15 -z 10 -q 500 -b 2000 -s 3 -n 500 -a 500 -d 0.250000 -e 0.500000 -m 500 -t 1 -j 1 -u 0 -i 1
```

Also definitely thought this was a `netcat` tool when I first saw @bgruening's ping :joy: But it's even better instead!

tools/necat/necat.xml
```diff
@@ -0,0 +1,180 @@
...
+            <param name="overlap_sensitive_options" type="text" value="-n 500 -z 10 -e 0.5 -j 0 -u 1 -a 1000" label="OVLP_SENSITIVE_OPTIONS" help="Overlap sensitive options">
+                <sanitizer>
+                    <valid initial="string.letters,string.digits">
+                        <add value="-" />
+                        <add value="." />
+                        <add value=" " />
+                    </valid>
+                </sanitizer>
+            </param>
+            <param name="consensus_fast_options" type="text" value="-a 2000 -x 4 -y 12 -l 1000 -e 0.5 -p 0.8 -u 0" label="CNS_FAST_OPTIONS" help="Consensus fast options">
```
hexylena, line 78:
```
11:51:47|(master) [hxr@cosima:/tmp/tmp.mvixDjib8b/NECAT]$ ./Linux-amd64/bin/oc2cns
USAGE:
./Linux-amd64/bin/oc2cns [options] wrk_dir candidates cns_out raw_out

If Multiple Nodes Are Used:
./Linux-amd64/bin/oc2cns [options] wrk_dir candidates cns_out raw_out -mn node_id num_nodes

OPTIONS AND DESCRIPTIONS:
-a <Integer>	align length cutoff
-x <Integer>	minimal coverage
-y <Integer>	maximal coverage
-l <Integer>	minimal length of corrected reads.
-f <0 or 1>	full consensus or not: 1 = yes, 0 = no
-e <Real>	sequencing error
-p <Real>	minimal mapping ratio
-t <Integer>	number of cpu threads
-r <0 or 1>	rescue long indels or not: 1 = yes, 0 = no
-u <0 or 1>	use dynamic or fixed ident cutoff: 1 = fixed, 0 = dynamic
-s <0 or 1>	use small memoty
DEFAULT OPTIONS:
-a 400 -x 4 -y 12 -l 500 -f 0 -e 0.500000 -p 0.800000 -t 1 -r 0 -u 0 -s 0

File: tools/necat/necat.xml [hunk 8]
```diff
@@ -0,0 +1,180 @@
...
+            <param name="consensus_sensitive_options" type="text" value="-a 2000 -x 4 -y 12 -l 1000 -e 0.5 -p 0.8 -u 0" label="CNS_SENSITIVE_OPTIONS" help="Consensus sensitive options">
+                <sanitizer>
+                    <valid initial="string.letters,string.digits">
+                        <add value="-" />
+                        <add value="." />
+                        <add value=" " />
+                    </valid>
+                </sanitizer>
+            </param>
+            <param name="trim_overlap_options" type="text" value="-n 100 -z 10 -b 2000 -e 0.5 -j 1 -u 1 -a 400" label="TRIM_OVLP_OPTIONS" help="Trimming overlap options">
```
hexylena, line 96: i'm struggling to find this one + assembly_overlap_options, it's difficult to even trace where they go.

tools/necat/necat.xml
```diff
@@ -0,0 +1,180 @@
...
+
+**Output**
+
+- Corrected reads (fasta)
+- Genome assembly (fasta) (Optional)
+
+    ]]></help>
+    <citations>
+        <citation type="doi">https://doi.org/10.1038/s41467-020-20236-7</citation>
+    </citations>
```
gallardoalba, line 179:
```suggestion
    <expand macro="citations" />
```

tools/necat/macros.xml
```diff
@@ -0,0 +1,44 @@
...
+            CLEANUP=1
+            USE_GRID=false 
+            GRID_NODE=0 
+            GRID_OPTIONS=
+            SMALL_MEMORY=0 
+            FSA_OL_FILTER_OPTIONS= 
+            FSA_ASSEMBLE_OPTIONS= 
+            FSA_CTG_BRIDGE_OPTIONS= 
+            ]]>
+        </configfile>
```
gallardoalba, line 42:
```suggestion
        </configfile>
        <xml name="citations">
                <citations>
                    <citation type="doi">10.1038/s41467-020-20236-7</citation>
                </citations>
        </xml>
```

tools/necat/necat.xml
gallardoalba, line 44:
```suggestion
        <param name="input_fastqs" type="data" format="fastq,fasta" multiple="true" label="Input reads" help="Input read files (FASTQ or FASTA). To select more than one file or collection from your history, use the 'ctrl' key" />

```

tools/necat/necat.xml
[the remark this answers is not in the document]
GraceAHall: Couldn't quite get this working. both $i and $input seemed to be blank.

GraceAHall:
Thank you @bgruening @hexylena @gallardoalba for the input! I have added all your suggestions. 

Thanks to @hexylena I was able to convert those text params into int/float/bool params. They each have a section now. 

I also added 2 more tests, one for multiple input files of different types, and one to test some of the new advanced options. 

What do you think?

gallardoalba:
It seems that you need to resize some files; or [include assertions](https://github.com/gallardoalba/galaxytools/blob/1c67d07f66629a08437eb3d7bc9408c33f7de478/tools/flye/flye.xml#L135).

gallardoalba:
According to the test reports, the estimated genome sizes are 0, when are supposed to be 1300; could you check it? I think that it is not a problem with the wrapper, but not sure.

Screenshot from 2021-12-06 11-26-59

gallardoalba:
Also, I think that it would be nice to explain a little bit the different sections included in the Advanced options sections in the tool help section (I mean, e.g. what is the differences between "Overlap sensitive options" and "Overlap fast options"?

bgruening:
@GraceAHall do you need help here?

GraceAHall:
Hi all, 

Have expanded the advanced settings and include descriptions in the <help> section. 
Necat is very poorly documented. The option descriptions in the <help> section were only accessible by downloading the binaries and running them. 

I actually do not know what a few of these mean.  The only description available is via the above, and these descriptions are very lacking. I have left them as they appear in the binaries.

Necat also is still in development and can be quite buggy. It has a particular error that occurs often - `[FATAL] No overlaps were loaded`. This issue has been raised multiple times in the github, but authors have not replied. 

This is as good as I can do for this tool.

tools/necat/necat.xml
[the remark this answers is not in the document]
bgruening: See for example here: https://github.com/galaxyproject/tools-iuc/blob/e22821f51ae326b3696a9456cbc44f9982512a52/tools/qiime/qiime_core/upgma_cluster.xml#L12 and there are many other examples in the repo. Please feel free to push the stuff that don't work and we can help you.

bernt-matthias:
How to continue here? The tool wrapper seems good. 

> Necat also is still in development and can be quite buggy. It has a particular error that occurs often - [FATAL] No overlaps were loaded. This issue has been raised multiple times in the github, but authors have not replied.

Question to me is if the underlying tool is valuable. Seems that the upstream repo shows no activity in the last two years. Also the 2nd link (http://www.tgsbioinformatics.com/necat/) given in the original publication is dead.

Can we contact the author and ask what the plans are for the future? Are there alternative tools?

gallardoalba:
> How to continue here? The tool wrapper seems good.
> 
> > Necat also is still in development and can be quite buggy. It has a particular error that occurs often - [FATAL] No overlaps were loaded. This issue has been raised multiple times in the github, but authors have not replied.
> 
> Question to me is if the underlying tool is valuable. Seems that the upstream repo shows no activity in the last two years. Also the 2nd link (http://www.tgsbioinformatics.com/necat/) given in the original publication is dead.
> 
> Can we contact the author and ask what the plans are for the future? Are there alternative tools?

Definitely it seems to be interesting since it is currently used by Genoscope for assembling plant genomes.

gallardoalba (approved):

PR #4262, opened 2021-11-29
Updating tools/last from version 1205 to
1453

planemo-autoupdate:
Hello! This is an automated update of the following tool: **tools/last**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

I have updated tools/last from version 1205 to 1256.

**Project home page:** http://last.cbrc.jp


bernt-matthias:
Update upstream URL to https://gitlab.com/mcfrith/last

Free seem to be 2 relevant changes:

- 1219 Add fasta-nr
- 1243 Add last-split -r option

Issue #4270, opened 2021-12-03
Error run-in g latest version of freebayes


gregvonkuster:
I'm running Galaxy 21.09, and executing this version of freebayes: 
```
toolshed.g2.bx.psu.edu/repos/devteam/freebayes/freebayes/1.3.5+galaxy0
```
produces this error:
```
tool error
An error occurred with this dataset:
cannot find 'standard_filters'
```
This may be related to this recent change - https://github.com/galaxyproject/tools-iuc/pull/4250

Is this a bug, or have I missed something during the tool install?

bernt-matthias:
@gregvonkuster can you have a look: https://github.com/galaxyproject/tools-iuc/pull/4271

PR #4271, opened 2021-12-03
fix standard-filters argument in cheetah

bernt-matthias:
fixes https://github.com/galaxyproject/tools-iuc/issues/4270

FOR CONTRIBUTOR:

### DIFF
--- a/tools/freebayes/freebayes.xml
+++ b/tools/freebayes/freebayes.xml
@@ -1,4 +1,4 @@
-<tool id="freebayes" name="FreeBayes" version="@TOOL_VERSION@+galaxy0">
+<tool id="freebayes" name="FreeBayes" version="@TOOL_VERSION@+galaxy1">
     <description>bayesian genetic variant detector</description>
     <xrefs>
         <xref type="bio.tools">freebayes</xref>
@@ -181,7 +181,7 @@
 
         ##INPUT FILTERS
             #if str( $options_type.input_filters.input_filters_selector ) == "set":
-                $standard_filters
+                $options_type.input_filters.standard_filters
                 ${options_type.input_filters.use_duplicate_reads}
                 --min-mapping-quality ${options_type.input_filters.min_mapping_quality}
                 --min-base-quality ${options_type.input_filters.min_base_quality}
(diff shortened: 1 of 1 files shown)


gregvonkuster:
Awesome - thanks for the quick fix!

gregvonkuster:
This looks like it will now work, but is this the same tool that is in the TS here https://toolshed.g2.bx.psu.edu/view/devteam/freebayes/8a7a42541080?  

When installed from the TS, the tool only has this:
```
$ grep standard_filters *
freebayes.xml:                $standard_filters
```
But in the tools-iuc repo it has this:
```
$ grep standard_filters *
freebayes.xml:                        <param name="standard_filters" argument="--standard-filters" type="boolean" truevalue="-0" falsevalue="" checked="false"
```

bernt-matthias:
Have you updated you master branch? It should look like this: https://github.com/galaxyproject/tools-iuc/blob/b63133897598f2881de7bad1f18fcd69d7d7e9df/tools/freebayes/freebayes.xml#L485

tools/freebayes/freebayes.xml
```diff
@@ -181,7 +181,7 @@
 
         ##INPUT FILTERS
             #if str( $options_type.input_filters.input_filters_selector ) == "set":
-                $standard_filters
+                $options_type.input_filters.standard_filters
```
nsoranzo, line 184:
```suggestion
                ${options_type.input_filters.standard_filters}
```

bernt-matthias: Thanks for the suggestion @nsoranzo .. can you tell me what the curly braces are good for?

gregvonkuster:
> Have you updated you master branch? It should look like this:
> 
> https://github.com/galaxyproject/tools-iuc/blob/b63133897598f2881de7bad1f18fcd69d7d7e9df/tools/freebayes/freebayes.xml#L485

Yes, that is in the tools-iuc repo, but the TS version must be older (or something?) than what is in the tools-iuc repo.

tools/freebayes/freebayes.xml
nsoranzo: See https://cheetahtemplate.org/users_guide/language.html#placeholder-syntax-rules

They are optional and really only needed if you have something like `${var}string`, but they may improve readability. Same as in Bash, basically.
In this case my suggestion was mainly for consistency with the rest of the Cheetah code.

nsoranzo (approved):

PR #4272, opened 2021-12-03
Change vSNP processors

gregvonkuster:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/vsnp/vsnp_build_tables.py
+++ b/tools/vsnp/vsnp_build_tables.py
@@ -295,7 +295,7 @@ def preprocess_tables(task_queue, annotation_dict, timeout):
 
 
 def set_num_cpus(num_files, processes):
-    num_cpus = int(multiprocessing.cpu_count())
+    num_cpus = len(os.sched_getaffinity(0))
     if num_files < num_cpus and num_files < processes:
         return num_files
     if num_cpus < processes:
--- a/tools/vsnp/vsnp_get_snps.py
+++ b/tools/vsnp/vsnp_get_snps.py
@@ -22,7 +22,7 @@ def get_time_stamp():
 
 
 def set_num_cpus(num_files, processes):
-    num_cpus = int(multiprocessing.cpu_count())
+    num_cpus = len(os.sched_getaffinity(0))
     if num_files < num_cpus and num_files < processes:
         return num_files
     if num_cpus < processes:


gregvonkuster:
This fix is per instruction from @natefoo

gregvonkuster:
The linting error is due to the large files sizes required for testing.

bernt-matthias (changes requested):
Please bump the tool version.

gregvonkuster:
> Please bump the tool version.

No inputs and outputs change, so why the version bump?  Also, I'm the only user that has ever installed them.
```
Times cloned / installed: 2
```
If the version bump is required, we'll need to change the use of the macro to separate the version to each tool since only 2 our of the 5 vSNP tools are affected by this.

bernt-matthias:
>  I'm the only user that has ever installed them.

In this case it might be OK. 

Note, in my experience Galaxy loads only one tool if there are multiple installed with the same version and unfortunately this seems to be the older one. 

I mentioned this somewhere here for data managers https://github.com/galaxyproject/galaxy/pull/12255 .. and also noted this behavior just yesterday on my instance. Guess I should open an issue for this.

bernt-matthias (approved):
Up to you :)

bernt-matthias:
https://github.com/galaxyproject/galaxy/issues/13030

gregvonkuster:
> > I'm the only user that has ever installed them.
> 
> In this case it might be OK.
> 
> Note, in my experience Galaxy loads only one tool if there are multiple installed with the same version and unfortunately this seems to be the older one.
> 
> I mentioned this somewhere here for data managers [galaxyproject/galaxy#12255](https://github.com/galaxyproject/galaxy/pull/12255) .. and also noted this behavior just yesterday on my instance. Guess I should open an issue for this.

Hmm...this is a bit concerning.  If we merge this PR without bumping the version, the the Galaxy tool install / uninstall UI will show the tool as installed, but the Install button will be displayed instead of the Uninstall button since the currently installed revision has an update available in the Tool Shed.  

I've always clicked the Install button to get the update.  And after the update is installed, the Uninstall button is displayed for the installed version.

Are you saying that the update is not actually installed in this scenario, and that the original version of the tool (without the update) is used?

bernt-matthias:
The admin UI is not the problem. All revisions are shown (independent of if they are bumped).

The problem is the user interface. If multiple tools with the same id and version are installed then Galaxy will show the tool with the oldest revision to the user (in my experience .. still need a way to verify this).

gregvonkuster:
Can this be merged?

gregvonkuster:
Thanks @bernt-matthias  :)

PR #4275, opened 2021-12-06
Updating tools/raven from version 1.6.1 to 1.7.0

planemo-autoupdate:

I have updated tools/raven from version 1.6.1 to 1.7.0.

**Project home page:** https://github.com/lbcb-sci/raven/releases


simonbray:
There seem to be no new features (https://github.com/lbcb-sci/raven/compare/1.6.1...1.7.0) and only the file size check is failing.

PR #4284, opened 2021-12-10
Add fastq,fastq.gz datatypes

yvanlebras:
Add fastq,fastq.gz datatypes as tools like illuminapairend output these formats but obigrep, who is a tool who can be used on output dataset from illuminapairend was not allowing fastq/fastq.gz input files. OR maybe better to update obigrep to also use this @INPUT_FORMATS@ ?

FOR CONTRIBUTOR:

### DIFF
--- a/tools/obitools/illuminapairedend.xml
+++ b/tools/obitools/illuminapairedend.xml
@@ -29,6 +29,7 @@
             --sanger
         #end if
         --without-progress-bar
+        --fastq-output
         --score-min='$score'
         -r fastq3p.fastq 
         fastq5p.fastq
@@ -44,7 +45,7 @@
         <param name="score" type="float" value="40.0" label="minimum score for keeping aligment"/>
     </inputs>
     <outputs>
-        <data name="output" format_source="inputfastq3p" label="${tool.name} on ${on_string}: assembly results" />
+        <data name="output" format="fastqsanger" label="${tool.name} on ${on_string}: assembly results" />
     </outputs>
 
     <tests>
--- a/tools/obitools/illuminapairedend.xml
+++ b/tools/obitools/illuminapairedend.xml
@@ -22,7 +22,7 @@
             ##input file is in fastq nucleic format produced by solexa sequencer
             --solexa
         #else if $inputfastq3p.ext.startswith("fastqillumina")
-            ##input file is in fastq nucleic format produced by solexa sequencer
+            ##input file is in fastq nucleic format produced by illumina sequencer
             --illumina
         #else
             ## input file is in sanger fastq nucleic format (standard fastq)
--- a/tools/obitools/illuminapairedend.xml
+++ b/tools/obitools/illuminapairedend.xml
@@ -40,8 +40,8 @@
         ]]>
     </command>
     <inputs>
-        <param name="inputfastq3p" type="data" format="fastq,fastq.gz" label="Read from file" help="file of 3p (1:) Illumina pair-end reads to assemble in sanger fastq nucleic format (standard fastq)" />
-        <param name="inputfastq5p" type="data" format="fastq,fastq.gz" label="Read from file" help="file of 5p (2:) Illumina pair-end reads to assemble in sanger fastq nucleic format (standard fastq)" />
+        <param name="inputfastq3p" type="data" format="@INPUT_FORMATS@" label="Read from file" help="file of 3p (1:) Illumina pair-end reads to assemble in sanger fastq nucleic format (standard fastq)" />
+        <param name="inputfastq5p" type="data" format="@INPUT_FORMATS@" label="Read from file" help="file of 5p (2:) Illumina pair-end reads to assemble in sanger fastq nucleic format (standard fastq)" />
         <param name="score" type="float" value="40.0" label="minimum score for keeping aligment"/>
     </inputs>
     <outputs>
--- a/tools/obitools/macros.xml
+++ b/tools/obitools/macros.xml
@@ -19,7 +19,7 @@
         </stdio>
     </xml>
 
-    <token name="@INPUT_FORMATS@">fastqsanger,fastqsanger.gz,fastqsolexa,fastqsolexa.gz,fasta,fasta.gz</token>
+    <token name="@INPUT_FORMATS@">fastq,fastq.gz,fastqsanger,fastqsanger.gz,fastqsolexa,fastqsolexa.gz,fasta,fasta.gz</token>
     <token name="@GUNZIP_INPUT@"><![CDATA[
         #if $input.ext.endswith(".gz")
             gunzip -c '$input' > input &&


tools/obitools/macros.xml
```diff
@@ -19,7 +19,7 @@
         </stdio>
     </xml>
 
-    <token name="@INPUT_FORMATS@">fastqsanger,fastqsanger.gz,fastqsolexa,fastqsolexa.gz,fasta,fasta.gz</token>
+    <token name="@INPUT_FORMATS@">fastq,fastq.gz,fastqsanger,fastqsanger.gz,fastqsolexa,fastqsolexa.gz,fasta,fasta.gz</token>
```
bernt-matthias, line 22: Already includes fastqsanger.gz.

yvanlebras: Seems to me fastq.gz and fastqsanger.gz are 2 different datatypes isn't it?

yvanlebras: As an alternative maybe it is best to update obigrep tool it can use @INPUT_FORMAT@ instead of fastq,fastq.gz manually described?

bgruening:
@yvanlebras which tool is that `illuminapairend`?

yvanlebras:
https://github.com/yvanlebras/tools-iuc/blob/patch-3/tools/obitools/illuminapairedend.xml

bgruening:
Are you saying that if you put fastqsanger into the illuminapariedend tool you get `fastq` out?

According to https://github.com/yvanlebras/tools-iuc/blob/patch-3/tools/obitools/illuminapairedend.xml#L47 you should get that out what you put in.

yvanlebras:
If we give fastqsanger files as input, output is fastqsanger so it's ok. So yes we can "just" mention as a best practice it is good to use fastqsanger as input. The ""problem"" is that we can use some formats as input of illuminapairend who are not available for obigrep, naturally next tool on a classical workflow. For example taking a fastq format as input of illuminapairend gives a fastq output than is not recognized by obigrep

bgruening:
But this behaviour is kind of independent. fastqsanger is more strict than fastq. And if obgrep can only handle fastqsanger, it should only accept that. We have this in a few mapping tools as well. In this case we tell people to annotate their inputs with `fastqsanger` explicitly, or in a workflow, change the datatype to `fastqsanger` after the illuminapaired step.

yvanlebras:
Ok.. Trying to look at this more into details, there is something not clear for me...

On illuminapairedend tool wml file we have this : 

```
  illuminapairedend
        ##--index-file=
        #if $inputfastq3p.ext.startswith("fastqsolexa")
            ##input file is in fastq nucleic format produced by solexa sequencer
            --solexa
        #else if $inputfastq3p.ext.startswith("fastqillumina")
            ##input file is in fastq nucleic format produced by solexa sequencer
            --illumina
        #else
            ## input file is in sanger fastq nucleic format (standard fastq)
            --sanger
```
and on obigrep this :
```
        #if $input.ext.startswith("fastq")
            #set ext = "fastqsanger"
```
So if I am not wrong, obigrep only take fastqsanger format for fastq input files, so maybe we need to modify the input format as for now we can put no fastqsanger file (like fastq illumina 1.5+) as input and if I am not wrong, such files are considered as fastqsanger ones...

And for illuminapairedend, if I am not wrong again (sorry not sure to understand well the code here), if files are named .fastqsolexa or fastqillumina, ok, but if .fastq for example, these files are assimilated to fast sanger ones even if encoding in illumina 1.5+ for example... 

Am I wrong ?

yvanlebras:
..And If the fastq format is not sanger one, Illumina 1.5+ for example, is there a way on Galaxy to convert it to sanger / illumina 1.8+ format ? I am thinking in my poor tired brain these is such functionality.. but don't remember where ..

yvanlebras:
... just if needed, here the history where I am testing obitools tools (to create a tutorial for Gallantries) => https://usegalaxy.eu/u/ylebras/h/tuto-obitools . We notably see that there is this mention `file of 3p (1:) Illumina pair-end reads to assemble in sanger fastq nucleic format (standard fastq)` so in fact, this tool can take non sanger fastq file as input but output sanger ones...

yvanlebras:
So now it appears to me it's almost ok (I hope not making too much bad things with my big fingers like sausages ;) ). The ~last important think seems to me be to change this :
```
        #if $inputfastq3p.ext.startswith("fastqsolexa")
            ##input file is in fastq nucleic format produced by solexa sequencer
            --solexa
        #else if $inputfastq3p.ext.startswith("fastqillumina")
            ##input file is in fastq nucleic format produced by illumina sequencer
            --illumina
        #else
            ## input file is in sanger fastq nucleic format (standard fastq)
            --sanger
```
maybe by this:
```
        #if $inputfastq3p.ext.startswith("fastqsanger")
            ##input file is in fastq nucleic format produced by illumina 1.8+ or sanger sequencer
            --sanger
        #else
            ## input file is in solexa fastq nucleic format (illlumina < 1.8+ or others)
            --solexa
```

But maybe too naive ?

tools/obitools/macros.xml
bernt-matthias: Problem is that obigrep (and all other tools excep `illuminapairedend`) really only accepts `fastqsanger` and `fastqsolexa`. Only `illuminapairedend` takes also `fastqillumina`. 

The `fastq` data type contains all `fastq` derivatives, i.e. `fastqsanger`, `fastqillumina`, and `fastqsolexa`. So this would be wrong - for all tools except `illuminapairedend`.

tools/obitools/illuminapairedend.xml
```diff
@@ -39,12 +40,12 @@
         ]]>
     </command>
     <inputs>
-        <param name="inputfastq3p" type="data" format="fastq,fastq.gz" label="Read from file" help="file of 3p (1:) Illumina pair-end reads to assemble in sanger fastq nucleic format (standard fastq)" />
-        <param name="inputfastq5p" type="data" format="fastq,fastq.gz" label="Read from file" help="file of 5p (2:) Illumina pair-end reads to assemble in sanger fastq nucleic format (standard fastq)" />
+        <param name="inputfastq3p" type="data" format="@INPUT_FORMATS@" label="Read from file" help="file of 3p (1:) Illumina pair-end reads to assemble in sanger fastq nucleic format (standard fastq)" />
```
bernt-matthias, line 43: This change would also accept `fasta` and `fasta.gz` which is not an input that is accepted by this tool.

I would leave this as `fastq,fastq.gz` which includes all three fastq datatypes (zipped and unzipped).

tools/obitools/illuminapairedend.xml
```diff
@@ -39,12 +40,12 @@
...
     <inputs>
-        <param name="inputfastq3p" type="data" format="fastq,fastq.gz" label="Read from file" help="file of 3p (1:) Illumina pair-end reads to assemble in sanger fastq nucleic format (standard fastq)" />
-        <param name="inputfastq5p" type="data" format="fastq,fastq.gz" label="Read from file" help="file of 5p (2:) Illumina pair-end reads to assemble in sanger fastq nucleic format (standard fastq)" />
+        <param name="inputfastq3p" type="data" format="@INPUT_FORMATS@" label="Read from file" help="file of 3p (1:) Illumina pair-end reads to assemble in sanger fastq nucleic format (standard fastq)" />
+        <param name="inputfastq5p" type="data" format="@INPUT_FORMATS@" label="Read from file" help="file of 5p (2:) Illumina pair-end reads to assemble in sanger fastq nucleic format (standard fastq)" />
         <param name="score" type="float" value="40.0" label="minimum score for keeping aligment"/>
     </inputs>
     <outputs>
-        <data name="output" format_source="inputfastq3p" label="${tool.name} on ${on_string}: assembly results" />
+        <data name="output" format="fastqsanger" label="${tool.name} on ${on_string}: assembly results" />
```
bernt-matthias, line 48: This seems (nearly) correct to me .. Apparently all obitools output `fastqsanger`, but the Galaxy tool still gzips the output if the input is gzipped. This should be reflected by the tool. 

Easiest is with a `galaxy.json` file. But `<change_format>` might also work.

bernt-matthias: I'm pretty sure that we can reuse the `@GENERATE_GALAXY_JSON@` token here if we also support the `--fasta-output` argument in this tool .. which seems reasonable.

PR #4288, opened 2021-12-13
Updating tools/seqwish from version 0.7.2 to 0.7.3

planemo-autoupdate:
Hello! This is an automated update of the following tool: **tools/seqwish**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

I have updated tools/seqwish from version 0.7.2 to 0.7.3.

**Project home page:** https://github.com/ekg/seqwish/releases


bernt-matthias (changes requested):
Supports now `--version`.. So we should add `version_command`

bernt-matthias (approved):

PR #4291, opened 2021-12-14
CI: Check file sizes only for changed files

bernt-matthias:
as pointed out here: https://github.com/galaxyproject/tools-iuc/pull/3973#issuecomment-944247753

reducing the size of test data does not decrease the repo size
if the files are already in. so we should only check for the sizes
of changed files and be strict with them

took this from https://github.com/galaxyproject/tools-iuc/pull/4077 .. which seems more controversial 

FOR CONTRIBUTOR:

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -228,8 +228,6 @@ jobs:
     - uses: actions/checkout@v2
       with:
         fetch-depth: 0
-    - name: Save repositories to file
-      run: echo '${{ needs.setup.outputs.repository-list }}' > repository_list.txt
     - name: Check file sizes
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yaml
```diff
@@ -226,14 +227,12 @@ jobs:
     steps:
     - uses: actions/checkout@v2
       with:
-        fetch-depth: 1
+        fetch-depth: 0
     - name: Save repositories to file
       run: echo '${{ needs.setup.outputs.repository-list }}' > repository_list.txt
```
nsoranzo, line 232:
```suggestion
```

nsoranzo (approved):

PR #4292, opened 2021-12-15
only check added files

bernt-matthias:
otherwise find fails for deleted files

not sure if we should add more. here is the man

```
       --diff-filter=[(A|C|D|M|R|T|U|X|B)...[*]]
           Select only files that are Added (A), Copied (C), Deleted (D),
           Modified (M), Renamed (R), have their type (i.e. regular file,
           symlink, submodule, ...) changed (T), are Unmerged (U), are Unknown
           (X), or have had their pairing Broken (B). Any combination of the
           filter characters (including none) can be used. When *
           (All-or-none) is added to the combination, all paths are selected
           if there is any file that matches other criteria in the comparison;
           if there is no file that matches other criteria, nothing is
           selected.

           Also, these upper-case letters can be downcased to exclude. E.g.
           --diff-filter=ad excludes added and deleted paths.

           Note that not all diffs can feature all types. For instance, diffs
           from the index to the working tree can never have Added entries
           (because the set of paths included in the diff is limited by what
           is in the index). Similarly, copied and renamed entries cannot
           appear if detection for those types is disabled.
```

FOR CONTRIBUTOR:

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -231,7 +231,8 @@ jobs:
     - name: Check file sizes
       run: |
         touch file_size_report.txt
-        GITDIFF=$(git diff --diff-filter=d --name-only ${{ needs.setup.outputs.commit-range }})
+        GITDIFF=$(git diff --diff-filter=d --name-only ${{ needs.setup.outputs.commit-range }} | tr '\n' ' ')
+        echo GITDIFF "$GITDIFF"
         if [ -n "$GITDIFF" ]; then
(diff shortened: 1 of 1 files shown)


bernt-matthias:
> Save `git diff` output to a file instead of an environment variable?

Check

> A good test would be to add a space in `tools/trinotate/trinotate.log2` filename.

Check

nsoranzo:
LGTM! If you remove the test changes, we can merge.

nsoranzo (approved):

PR #4293, opened 2021-12-15
Update pangolin to use the latest scorpio (necessary for BA.3 lineage)

pvanheus:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/pangolin/pangolin.xml
+++ b/tools/pangolin/pangolin.xml
@@ -46,7 +46,7 @@
         <param name="include_header" type="boolean" label="Include header line in output file" 
             truevalue="true" falsevalue="false" />
         <conditional name="db">
-            <param type="select" name="source" label="pangoLEARN source" help="Where to find the pangoLEARN database">
+            <param type="select" name="source" label="pangoLEARN source" help="Where to find the pangoLEARN database. While 'Download latest from web' is recommended, if errors occur see the warning in the main help text below.">
                 <option value="download">Download latest from web</option>
                 <option value="builtin">Use database from Galaxy server</option>
                 <option value="default">Use default database built in to pangolin (not recommended)</option>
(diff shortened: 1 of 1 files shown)


pvanheus:
The latest [scorpio](https://github.com/cov-lineages/scorpio) is updated to support a new format of [constellations](https://github.com/cov-lineages/constellations). This is needed to support reporting on BA.3 lineage.

Since the pangolin version itself hasn't changed, this updates the wrapper by bumping its version and specifying scorpio 0.3.16 as a requirement.

tools/pangolin/pangolin.xml
```diff
@@ -1,10 +1,11 @@
-<tool id="pangolin" name="Pangolin" version="@TOOL_VERSION@+galaxy0" profile="20.01">
+<tool id="pangolin" name="Pangolin" version="@TOOL_VERSION@+galaxy1" profile="20.01">
     <description>Phylogenetic Assignment of Outbreak Lineages</description>
     <macros>
         <token name="@TOOL_VERSION@">3.1.17</token>
     </macros>
     <requirements>
         <requirement type="package" version="@TOOL_VERSION@">pangolin</requirement>
+        <requirement type="package" version="0.3.16">scorpio</requirement>
```
mvdbeek, line 8: That should not be necessary, we updated the main pangolin dependency yesterday, so you should be getting the latest scorpio version. @wm75 also fixed this upstream as well.

mvdbeek: I see, that was a very short-lived release then

pvanheus: Scorpio was updated today, so if anyone updated their pangolin yesterday it would have picked up the older scorpio and it will now break when it downloads the new constellations.

pvanheus: > I see, that was a very short-lived release then

welcome to the world of SARS-CoV-2 ;)

wm75: Kind of crazy, indeed. Thanks for keeping things working @pvanheus !

wm75:
Can you maybe add a warning message somewhere that the functionality of the "Latest from web" option can only be relied on for the current latest version of the wrapper, but may break without warning for older versions?

Unfortunately, we don't have any way to backport changes like this one.

pvanheus:
> Can you maybe add a warning message somewhere that the functionality of the "Latest from web" option can only be relied on for the current latest version of the wrapper, but may break without warning for older versions?
> 
> Unfortunately, we don't have any way to backport changes like this one.

Done. I've added a note in the help text for the DB source option and a longer note in the main help text.

bgruening (approved):

PR #4298, opened 2021-12-20
Exclude deleted files from file size check

simonbray:
Deleting a file leads to an error with the file size check, as the file can no longer be found: https://github.com/galaxyproject/tools-iuc/runs/4580754087

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -230,7 +230,7 @@ jobs:
         fetch-depth: 0
     - name: Check file sizes
       run: |
-        find $(git diff --name-only ${{ needs.setup.outputs.commit-range }}) -type f -size +${{ env.MAX_FILE_SIZE }} > file_size_report.txt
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yaml
```diff
@@ -230,7 +230,7 @@ jobs:
         fetch-depth: 0
     - name: Check file sizes
       run: |
-        find $(git diff --name-only ${{ needs.setup.outputs.commit-range }}) -type f -size +${{ env.MAX_FILE_SIZE }} > file_size_report.txt
+        find $(git diff --name-only --diff-filter=d ${{ needs.setup.outputs.commit-range }}) -type f -size +${{ env.MAX_FILE_SIZE }} > file_size_report.txt
```
nsoranzo, line 233: If the only changes are file deletions, these will return all files in the repository that are larger than `env.MAX_FILE_SIZE`.
I think this needs to be split in 2 commands, with a check that the output of `git diff` is non-empty.

simonbray: Thank you @nsoranzo - I'll close this in favour of #4292.

PR #4306, opened 2022-01-03
Updating tools/bwameth from version 0.2.2 to 0.2.3

planemo-autoupdate:

I have updated tools/bwameth from version 0.2.2 to 0.2.3.

### DIFF
--- a/tools/bwameth/bwameth.xml
+++ b/tools/bwameth/bwameth.xml
@@ -1,7 +1,7 @@
-<tool id="bwameth" name="bwameth" version="@TOOL_VERSION@+galaxy1" profile="20.05">
+<tool id="bwameth" name="bwameth" version="@TOOL_VERSION@+galaxy0" profile="20.05">
     <description>Fast and accurate aligner of BS-Seq reads.</description>
     <macros>
-        <token name="@TOOL_VERSION@">0.2.2</token>
+        <token name="@TOOL_VERSION@">0.2.3</token>
     </macros>
     <requirements>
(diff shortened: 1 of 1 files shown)


bgruening:
The new version added support for BWA2 indices, nothing we need here strictly.

bgruening (approved):

bernt-matthias:
So we could use `bwa_mem2_indexes` here: https://github.com/galaxyproject/tools-iuc/blob/5275667f1a59335340e07f7633a56821fb73f407/tools/bwameth/bwameth.xml#L94

in addition / as alternative?

bgruening:
I would decide for one and since BWA is more common and we all have indices for that one already I would keep it like it is for the time being. Or is it really worth the pain of creating new indices and maintaining them?

bernt-matthias:
> since BWA is more common and we all have indices for that one already I would keep it like it is for the time being

Just to be sure: its not the bwa data data table, but the bwameth data table.

Currently there are 3 data data tables:

- bwa_mem_indexes
- bwa_mem2_indexes
- bwameth_indexes

(where the second has apparently no data manager). So I thought that it might be a possibility to drop one? Or create a unifying data table that also stores the creation method in an extra column.

But we can merge for now .. and maybe open an issue.

bgruening:
> But we can merge for now .. and maybe open an issue.

I like that.

bernt-matthias:
Opened an issue https://github.com/galaxyproject/tools-iuc/issues/4319

PR #4309, opened 2022-01-03
Updating tools/minia from version 3.2.4 to 3.2.6

planemo-autoupdate:
Hello! This is an automated update of the following tool: **tools/minia**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

I have updated tools/minia from version 3.2.4 to 3.2.6.

**Project home page:** https://gatb.inria.fr/software/minia

### DIFF
--- a/tools/minia/minia.xml
+++ b/tools/minia/minia.xml
@@ -4,7 +4,7 @@
         <xref type="bio.tools">minia</xref>
     </xrefs>
     <macros>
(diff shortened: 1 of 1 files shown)


bernt-matthias:
Seems to be a bugfix only release according to the release notes

bernt-matthias (approved):

PR #4312, opened 2022-01-03
Updating tools/stringtie from version 2.1.7 to
2.2.1

planemo-autoupdate:
Hello! This is an automated update of the following tool: **tools/stringtie**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

I have updated tools/stringtie from version 2.1.7 to 2.2.0.

**Project home page:** http://ccb.jhu.edu/software/stringtie


simonbray:
```
v2.2.0 release hilights:

    --mix option allows StringTie to take both short and long read alignments; when this option is used, the 2nd BAM (or CRAM) file in the command line must be a long reads alignment file (the 1st being the short-reads alignment file); -L option should not be used in this case.
    improved transcriptome assembly on mixed data with annotation
    added support for CRAM input files, as StringTie is now built using HTSlib
```

gallardoalba:
> ```
> v2.2.0 release hilights:
> 
>     --mix option allows StringTie to take both short and long read alignments; when this option is used, the 2nd BAM (or CRAM) file in the command line must be a long reads alignment file (the 1st being the short-reads alignment file); -L option should not be used in this case.
>     improved transcriptome assembly on mixed data with annotation
>     added support for CRAM input files, as StringTie is now built using HTSlib
> ```

It seems that the --mix option was included in a previous PR.

gallardoalba (approved):

PR #4314, opened 2022-01-03
add new tool rpextractsink v5.11.1

kenza12:
Add new tool **rpextractsink** : Generate the RetroPath2.0 sink file from an SBML input.

Homepage: https://github.com/brsynth/rptools/tree/master/rptools/rpextractsink

LICENSE : https://github.com/brsynth/rptools/blob/master/LICENSE

FOR CONTRIBUTOR:


tools/rpextractsink/.shed.yml
```diff
@@ -0,0 +1,11 @@
...
+- Synthetic Biology
+description: Generate the RetroPath2.0 sink file from an SBML input
+long_description: 'Sink refers to the collection of chemical species used by the restrosynthesis
+  algorithm of RetroPath2.0 to finish metabolic route exploration. This tool uses
+  an SBML file of the desired chassis organism, parses all the molecules within a
+  specified compartment (example: cytoplasm, Golgi apparatus, nucleus, etc) and uses
+  its MIRIAM annotation to find their InChI structures. Source code may be found at
+  the following location: https://github.com/brsynth/rptools/tree/master/rptools/rpextractsink'
+name: rpextractsink
+owner: tduigou
```
bgruening, line 11: this needs to be iuc

tools/rpextractsink/rpextractsink.xml
```diff
@@ -0,0 +1,214 @@
+<tool id="rpExtractSink" name="Sink from SBML" version="5.11.1">
```
bgruening, line 1: Can we have all lower case for the ID?
Please add `profile="20.09"` to the tool as well.

tools/rpextractsink/rpextractsink.xml
```diff
@@ -0,0 +1,214 @@
...
+        #end if
+    ]]></command>
+    <inputs>
+        <conditional name="input_strain_type">
+            <param name="strain_type" type="select" label="Input type">
+                <option value="BiGG" selected="true">BiGG database</option>
+                <option value="user">Custom file</option>
+            </param>
+            <when value="user">
+                <param name="input" type="data" format="sbml" optional="false" label="Strain" />
```
bgruening, line 31: @kenza12 is SMBL working for you?

I had this open https://github.com/galaxyproject/galaxy/pull/10144 but not sure if needed anymore.

tools/rpextractsink/rpextractsink.xml
```diff
@@ -0,0 +1,214 @@
...
+            </when>
+            <when value="BiGG">
+                <param name="input" type="select" label="Strain">
+                    <option value="iCN718">(iCN718) Acinetobacter baumannii AYE</option>
```
bgruening, line 35: I would put the brackets at the end and the description at the beginning.

tools/rpextractsink/rpextractsink.xml
```diff
@@ -0,0 +1,214 @@
...
+            </when>
+        </conditional>
+        <!-- <param name="input" type="data" format="sbml" optional="false" label="SBML model" /> -->
+        <section name="adv" title="Advanced Options" expanded="false">
+            <param name="compartment_id" type="text" value="c" label="SBML compartment ID" />
+            <param name="remove_dead_end" type="boolean" checked="true" label="Remove dead-end metabolites using FVA evaluation?" />
+        </section>
+    </inputs>
+    <outputs>
+        <data name="sink" format="csv" label="${tool.name} - ${input_strain_type.input}" />
```
bgruening, line 92: Can you convert the output to TSV? Galaxy has a lot of tools that can deal with TSV but less with CSV.

tools/rpextractsink/rpextractsink.xml
```diff
@@ -0,0 +1,214 @@
...
+  author={Le Nov{\`e}re, Nicolas and Finney, Andrew and Hucka, Michael and Bhalla, Upinder S and Campagne, Fabien and Collado-Vides, Julio and Crampin, Edmund J and Halstead, Matt and Klipp, Edda and Mendes, Pedro and others},
+  journal={Nature biotechnology},
+  volume={23},
+  number={12},
+  pages={1509--1515},
+  year={2005},
+  publisher={Nature Publishing Group}
+}
+        </citation>
+        <citation type="bibtex">
```
bgruening, line 179: can you use DOI here instead of bibtex?

tools/rpextractsink/rpextractsink.xml
```diff
@@ -0,0 +1,214 @@
...
+    <description>Generate the RetroPath2.0 sink file from an SBML input</description>
+    <requirements>
+        <requirement type="package" version="5.11.1">rptools</requirement>
+    </requirements>
+    <stdio>
+		<regex match="Could not retreive any species in the compartment" level="fatal" />
+	</stdio>
+    <command detect_errors="exit_code"><![CDATA[
+        #if str($input_strain_type.strain_type) == "BiGG":
+            curl http://bigg.ucsd.edu/static/models/$(input_strain_type.input).xml.gz --output $(input_strain_type.input).xml.gz;
```
bgruening, line 11: please use && instead of ; and single-quote all paths

bgruening (commented):
@kenza12 thanks a lot for your contribution. I have added some comments inline.

tools/rpextractsink/rpextractsink.xml
kenza12: Hi @bgruening , thanks for your return.
Do you mind if we can specify `profile="19.09"` instead of `profile="20.09"` in order to install this tool in our galaxy instance too (https://galaxy-synbiocad.org/) which doesn't use the 20.09 version yet ?

tools/rpextractsink/rpextractsink.xml
kenza12: Yes it's working well. We didn't encounter any problems while uploading sbml files. Thank you.

tools/rpextractsink/rpextractsink.xml
kenza12: We have several tools that support CSV files either as input or as output. It would take a lot of work to make the changes. So it would be easier for us to keep the files in CSV format but from now on we will work with the TSV files for the next tools in order to be in line with Galaxy, if that's okay with you?

tools/rpextractsink/rpextractsink.xml
bgruening: Sure, if the tests pass we can do that. But you should really update :)

tools/rpextractsink/rpextractsink.xml
bgruening: Cool.

tools/rpextractsink/rpextractsink.xml
bgruening: Can we add this as on option for the time being. So that people that already want to use tsv can do that?

kenza12:
Hi @bgruening ,
I just made all the necessary changes including the TSV option.
Thank you for your work.

tools/rpextractsink/rpextractsink.xml
kenza12: Hi @bgruening ,
Before submitting the new changes in this pull request, We'd like to know what is the purpose of modifying our tools when it is already possible on Galaxy with `Edit Attributes` , `Convert CSV to tabular`. Thank you.

bernt-matthias (commented):
Can you put all tools requiring rptools in a single folder?

kenza12:
> Can you put all tools requiring rptools in a single folder?

Hi @bernt-matthias ,
We prefer not to use a suite of rptools and to be able to install each tool individually because they are from different categories (retrosynthesis, pathway analysis ..). We also plan to separate these tools in the future.
Thank you.

bgruening:
> 

I think there is a misunderstanding here. A suite of tools just means you can share here in this repo, much code, between many tools. The tools will be decomposed by planemo and end up in different repositories. We actually recommend having only one tool per repository. However, if tool development can profit from having them in one folder sharing macros etc, we use "suites" for that. @kenza12 having seen your GTN PR. What is needed to get the GTN tools merged here?

kenza12:
> > 
> 
> I think there is a misunderstanding here. A suite of tools just means you can share here in this repo, much code, between many tools. The tools will be decomposed by planemo and end up in different repositories. We actually recommend having only one tool per repository. However, if tool development can profit from having them in one folder sharing macros etc, we use "suites" for that. @kenza12 having seen your GTN PR. What is needed to get the GTN tools merged here?

Hi @bgruening ,

Ok I see, thank you. I thought it was like _bedtools_ suite where you can't install each tool from the suite individually. So, to do that, I will create a rptools directory and move rpfba and rpextractsink on it. I will also modify the ID tools. Example: from rpfba to rptools_rpfba. Do I need to do something more ?

About rpExtractSink, I got an error :

`PermissionError: [Errno 13] Permission denied: '/usr/local/lib/python3.9/site-packages/rr_cache/cache'`

I'm not able to reproduce this error locally yet. I have troubles with `--biocontainers` option. But tests passed locally without --biocontainers option. I need time to figure out how to make this command line working in my local Galaxy, maybe I’ll install a new local galaxy instance.

About the PR in Galaxy Training, yes we're planning to move all the tools available in the toolshed in _synthetic biology_ to IUC repository and making tutorials for all the available workflows. I'll do that as soon as possible.

bernt-matthias:
The error is because the tool writes to the installation directory which is impossible in containers and should also not be done in conda environments.

kenza12:
> The error is because the tool writes to the installation directory which is impossible in containers and should also not be done in conda environments.

Hi, unfortunately , I will close this PR because of this unsolved technical error. The developer cannot change that because It's time consuming. I did the same for the other tool: https://github.com/galaxyproject/tools-iuc/pull/4456

Thank you.

bgruening:
@kenza12 what we can try is to copy the binary (the tool) into the CWD and run the tool from the CWD. Not sure if this will work, but worth a try?

kenza12:
> @kenza12 what we can try is to copy the binary (the tool) into the CWD and run the tool from the CWD. Not sure if this will work, but worth a try?

Yes sure, I'll give it a try ! Thank you @bgruening !

kenza12:
Hi, **rpextractsink** was updated to enable a new argument where I can specify the localization of the rr_cache output. Before, it was writing at the package directory and it was not allowed.
I have one question about the output directory. In the wrapper, I specified `"\${TMPDIR:-.}"` but this is specific to a user. I'd like to save this output in a directory where all users can have access to it. Otherwise, This directory will be generated each time and it's time consuming. I don't know if you have such a "common" directory in Galaxy ?
Note that this output won't be written again if he already exists.
Thanks.

bgruening:
> I have one question about the output directory. In the wrapper, I specified "\${TMPDIR:-.}" but this is specific to a user. I'd like to save this output in a directory where all users can have access to it. Otherwise, This directory will be generated each time and it's time consuming. I don't know if you have such a "common" directory in Galaxy ?

This is where Galaxy data tables come into play. Same concepts as with genome indices or reference databases, as BLAST. They are shared with *.loc files. The process is just different, there is one tool that creates those shared data and then tools that can use them.

kenza12:
> 

Thank you for your answer.
rpExtractSink uses rr_cache to generate a cache data which is stored on disk and loaded in memory each time the tool is used in this format : `json.gz` . I don't see how I can define the loc file in this particular situation. Do I need to specify the location of this `json.gz` file in the loc file ? Where can I store it ? I can see that genome indices are stored in this directory : `/galaxy/tool-data/shared/` . Do you think I can store the shared file in this location ?

kenza12:
Hi,
Thank you for your answer.
I just started learning about DM but I think I can't create it in this case because this data (_cid_strc.json.gz_) is generated by rpextractsink tool (_--cache-dir option_) and used by this same tool to generate the final output. I also don't need a loc file, I just need to specify an output folder (accessible to all users). I was thinking about something like that:

`mkdir -p '/galaxy/tool-data/shared/rpextractsink' && python -m rptools.rpextractsink '/galaxy/database/files/002/dataset_2036.dat' '/galaxy/database/files/002/dataset_2037.dat' --compartment_id 'c' --remove_dead_end --cache-dir '/galaxy/tool-data/shared/rpextractsink'
`

This is working on my local galaxy but I don't know if It's possible to work this way and if it's accessible to all users ?

bgruening:
@kenza12 just for my understanding how big is the likelihood of a cache hit? I mean is it likely that the users have the same input and the same DB is generated? If so we can pre-generate those DBs with a data manager. If this is rather unlikely we just generate the DB every time, I don't think this is a big deal, isn't it?

kenza12:
> @kenza12 just for my understanding how big is the likelihood of a cache hit? I mean is it likely that the users have the same input and the same DB is generated? If so we can pre-generate those DBs with a data manager. If this is rather unlikely we just generate the DB every time, I don't think this is a big deal, isn't it?

@bgruening Yes, indeed the same DB is generated and I will create a DM to pre-generate these data.

PR #4323, opened 2022-01-11
bedtools GetFastaBed: add -nameOnly flag

FleurGaBru:
The "-nameOnly" flag differs from the "-name" flag by adding only the content of the 4th column in the bed file to the output fasta header. The "-name" flag adds both the name and the coordinates. Some tools require very specific requirements for the fasta header and there -nameOnly appears to be very convenient.

FOR CONTRIBUTOR:


tools/bedtools/getfastaBed.xml
```diff
@@ -61,13 +64,15 @@ $split
             <param name="fasta" value="nucBed1.fasta" ftype="fasta" />
             <param name="tab" value="False" />
             <param name="split" value="False" />
+            <param name="nameOnly" value="False" />
```
bernt-matthias, line 67: Would be cool to have one test with nameOnly true

FleurGaBru: Agree!

FleurGaBru: Can I change it without sending a new pull request? I am new here :)

bernt-matthias: > Can I change it without sending a new pull request? I am new here :)

Simply push to the branch in your fork. New commits will be added automatically to the PR.

tools/bedtools/getfastaBed.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="bedtools_getfastabed" name="bedtools GetFastaBed" version="@TOOL_VERSION@" profile="@PROFILE@">
+<tool id="bedtools_getfastabed" name="bedtools GetFastaBed" version="@TOOL_VERSION@+galaxy1" profile="@PROFILE@">
```
bernt-matthias, line 1:
```suggest
<tool id="bedtools_getfastabed" name="bedtools GetFasta" version="@TOOL_VERSION@+galaxy1" profile="@PROFILE@">
```

I would suggest to have the name identical to the actual (sub)command. As long as we do not change the id of the tool this should not have a negative side effect.

bernt-matthias:
The changed test implies that the tests file getfastaBed_result2.tabular needs to be updated. The diff is

```diff
--- local_file
+++ history_data
@@ -1 +1 @@
-chr1:10-100	TTCTTACCTATTAGTGGTTGAACATCGTGATATGTATGTTGACGGCCATAAGGCTGCTTCTTGTTGTCGATAGAACTTCATGTGCCTGTA
+	TTCTTACCTATTAGTGGTTGAACATCGTGATATGTATGTTGACGGCCATAAGGCTGCTTCTTGTTGTCGATAGAACTTCATGTGCCTGTA
```

Even better would be to update the input fasta such that it has a header format that bedtools understand.

tools/bedtools/getfastaBed.xml
FleurGaBru: I absolutely agree!

bgruening (approved):

bgruening:
Thanks @FleurGaBru!

Issue #4322, opened 2022-01-10
Range of allowed values for max-depth parameter of bcftools mpileup


profgiuseppe:
Greetings, one user asked why the max-depth parameter of the bcftools mpileup tool has a [maximum allowed value of 1024](https://github.com/galaxyproject/tools-iuc/blob/master/tools/bcftools/bcftools_mpileup.xml#L226), given that the corresponding command line argument has no such boundary., according to the tool [man page](https://samtools.github.io/bcftools/bcftools.html)
Is there a galaxy-related reason for the choice of boundary? I would rather avoid having a custom xml for one tool the local installation of galaxy.
Thanks!

nsoranzo:
@profgiuseppe I don't think there is a Galaxy-specific reason, a PR removing the `max` value from the parameter looks the best way forward.

bernt-matthias:
Solved by https://github.com/galaxyproject/tools-iuc/pull/4326

PR #4326, opened 2022-01-14
Removing max limit to bcftools max-depth parameter

profgiuseppe:
FOR CONTRIBUTOR:

Removing an upper limit to a parameter of bcftools mpilup, as per #4322

### DIFF
--- a/tools/bcftools/bcftools_annotate.xml
+++ b/tools/bcftools/bcftools_annotate.xml
@@ -1,11 +1,11 @@
 <?xml version='1.0' encoding='utf-8'?>
 <tool name="bcftools @EXECUTABLE@" id="bcftools_@EXECUTABLE@" version="@TOOL_VERSION@">
     <description>Annotate and edit VCF/BCF files</description>
-    <expand macro="bio_tools" />
     <macros>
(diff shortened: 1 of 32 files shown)


nsoranzo (approved):

wm75:
@nsoranzo do we want to bump the wrapper version?

nsoranzo:
> @nsoranzo do we want to bump the wrapper version?

Doesn't seem needed to me.

nsoranzo:
> The result is going to be different, it's a problem for the workflow state, this needs a version bump.

@mvdbeek Can you elaborate a bit please? I'd like to understand this well for future reviews.

mvdbeek:
There's a problem that as a user you could define a workflow with the fixed tool version and set the value beyond the previous max. As a user you could reasonably expect that you can take the workflow to any server with the right tool version, except if the tool version is the non-fixed one the workflow won't run. If there's a version bump that's a clean and clear separation.

nsoranzo (approved):

nsoranzo:
Thanks for taking the time to explain! I've bumped the tool version.

PR #4329, opened 2022-01-14
bbtools, ensure read pair identifiers differ

gregvonkuster:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/bbtools/bbduk.xml
+++ b/tools/bbtools/bbduk.xml
@@ -10,18 +10,16 @@
 
 #if str($input_type_cond.input_type) in ['single', 'pair']:
     #set read1 = $input_type_cond.read1
-    #set read1_identifier = re.sub('[^\s\w\-]', '_', str($read1.element_identifier))
     ## bbduk uses the file extension to determine the input format.
-    #set ext = $read1_identifier + '.fastq'
+    #set ext = '.fastq'
     #if $read1.ext.endswith('.gz'):
         #set ext = $ext + '.gz'
     #end if
-    #set read1_file = $read1_identifier + $ext
+    #set read1_file = 'forward'  + $ext
     ln -s '${read1}' '${read1_file}' &&
     #if str($input_type_cond.input_type) == 'pair':
         #set read2 = $input_type_cond.read2
-        #set read2_identifier = re.sub('[^\s\w\-]', '_', str($read2.element_identifier))
-        #set read2_file = $read2_identifier + $ext
+        #set read2_file = 'reverse' + $ext
         ln -s '${read2}' '${read2_file}' &&
     #end if
 #else:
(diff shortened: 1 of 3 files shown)


gregvonkuster:
The error on this the the MAX_FILE_SIZE issue:
```
Run exit 1
  exit 1
  shell: /usr/bin/bash -e {0}
  env:
    GALAXY_FORK: galaxyproject
    GALAXY_BRANCH: release_21.09
    MAX_CHUNKS: 4
    MAX_FILE_SIZE: 500k
Error: Process completed with exit code 1.
```

tools/bbtools/bbduk.xml
```diff
@@ -10,18 +10,16 @@
...
     #set read1 = $input_type_cond.read1
-    #set read1_identifier = re.sub('[^\s\w\-]', '_', str($read1.element_identifier))
     ## bbduk uses the file extension to determine the input format.
-    #set ext = $read1_identifier + '.fastq'
+    #set ext = '.fastq'
     #if $read1.ext.endswith('.gz'):
         #set ext = $ext + '.gz'
     #end if
-    #set read1_file = $read1_identifier + $ext
+    #set read1_file = 'forward'  + $ext
```
bgruening, line 18: That looks in general fine, but no the identifier is also not unique, left? So how does this help you in contrast to the previous version?

tools/bbtools/macros.xml
```diff
@@ -1,6 +1,6 @@
 <macros>
     <token name="@WRAPPER_VERSION@">1.0.0</token>
-    <token name="@VERSION_SUFFIX@">1.0.0</token>
+    <token name="@VERSION_SUFFIX@">1.0.1</token>
```
bgruening, line 3: whenever you bump the wrapper version, this suffix is supposed to be just 0,1,2,3,4... and so on. Like a build number.

tools/bbtools/bbduk.xml
gregvonkuster: A user first ran BBDuk on samples to decontaminate them and then attempted to map them with BBMap.  I think this is unusual (generally mapping occurs first in an analysis pipeline, although I'm not familiar with BBTools pipelines)  but in this case the output of BBDuk was the input of BBMap, and the element identifier was the history label.  Since the outputs of BBDuk were not unique (due to the missing forward / reverse label items that I've added in this PR) the symlink attempt resulted in the following error and (to make things cleaner) I thought it would be better to not attempt to symlink hid label strings.
```
error
An error occurred with this dataset:
ln: BBTools_ BBduk on data 37 and data 36 _Unmatched_BBTools_ BBduk on data 37 and data 36 _Unmatched_.fastq: File exists
```

tools/bbtools/macros.xml
gregvonkuster: Fixed, thanks!  I'm not sure how this incorrect version string got in there!

gregvonkuster:
@bgruening - the bbduk label fixes here are necessary.  If you'd like me to switch the symlink names back to using the string identifiers, let me know.  Also, should the version strings now be pulled out of macros since not all tools are affected?  Either way is fine with me.  I'd like to get this merged soon if possible.  Thanks!

bgruening (approved):

bgruening:
Thanks @gregvonkuster

PR #4338, opened 2022-01-21
Add hap.py

JasperO98:
FOR CONTRIBUTOR:

Could you have a look at it @hexylena ?


JasperO98:
Check fails at "Combine chunked test results" with `Unsuccessful tests found, inspect the 'All tool test results' artifact for details.`
What does it mean?

bgruening:
@JasperO98 if you click on the failed test you can navigate to this page, under Summary.

image

Below you find the "All tool test results", download this artifact and look at the included HTML. It will give you the planemo summary with all errors included.

JasperO98:
Thanks, I overlooked that! Apparently it did not generate the output results.tsv, however the problem is that --update_test_data should be added to the command according to the html file. However, shouldn't it also work without this option?
image

hexylena:
> Apparently it did not generate the output results.tsv, however the problem is that --update_test_data should be added to the command according to the html file. However, shouldn't it also work without this option?

It's a bit confusingly worded, what it means is the test data is missing from the repo, so it has nothing to compare against.

tools/hap.py/hap.py.xml
```diff
@@ -0,0 +1,52 @@
...
+		<requirement type="package" version="@TOOL_VERSION@">hap.py</requirement>
+	</requirements>
+	
+    <command detect_errors="exit_code"><![CDATA[
+		export HGREF='$reference' && 
+		samtools faidx '$reference' &&
+        som.py
+		'$truth'
+		'$query'
+		-r '$reference'
```
hexylena, line 19:
```suggestion
		-r '$ref_genome'
```
based on the suggestion above

tools/hap.py/hap.py.xml
```diff
@@ -0,0 +1,52 @@
...
+	</command>
+	
+	<inputs>
+		<param name="truth" type="data" format="vcf" label="ground truth variant calls"/>
+		<param name="query" type="data" format="vcf" label="query variant calls"/>
+		<param name="reference" argument="-r" type="data" format="fa,fasta,fna" label="reference genome"/>
+	</inputs>
+	
+    <outputs>
+		<data name="results" label="${tool.name} on ${on_string}" format="tabular"/>
```
hexylena, line 31:
```suggestion
		<data name="results" label="${tool.name} on ${on_string}" format="tabular" from_work_dir="results.tsv" />
		<data name="stats" label="${tool.name} on ${on_string}: stats" format="csv" from_work_dir="happy.stats.csv" />
		<data name="metrics" label="${tool.name} on ${on_string}: metrics" format="json" from_work_dir="happy.metrics.json"/>
```

tools/hap.py/hap.py.xml
```diff
@@ -0,0 +1,52 @@
...
+	</requirements>
+	
+    <command detect_errors="exit_code"><![CDATA[
+		export HGREF='$reference' && 
+		samtools faidx '$reference' &&
+        som.py
+		'$truth'
+		'$query'
+		-r '$reference'
+		-o 'happy' > results.tsv
```
hexylena, line 20:
```suggestion
		-o 'happy' | sed 's/\s\+/\t/g' | tail -n+2> results.tsv
```

It seems hap.py's results are manually aligned with spaces, rather than proper tabs. So, we need this to convert those. It also has a blank first line which is odd. And so a tail to remove that.

hexylena: This is the other reason for the error, here you write `results.tsv`, so even if the results file was included, Galaxy hasn't figured out where the results file is (when checking the output stanza)

hexylena (commented):
Overall great start @JasperO98! I've added the test data and made some code suggestions. 

Do you reckon you'll need any more of the CLI options for [som.py](https://github.com/Illumina/hap.py/blob/master/doc/sompy.md#all-options)? We don't *need* to add them all from the start, we can always update the tool later and if people find they need them, depends on your use case I guess.

tools/hap.py/hap.py.xml
```diff
@@ -0,0 +1,52 @@
...
+        ]]>
+	</command>
+	
+	<inputs>
+		<param name="truth" type="data" format="vcf" label="ground truth variant calls"/>
+		<param name="query" type="data" format="vcf" label="query variant calls"/>
```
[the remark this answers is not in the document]
JasperO98: Resolved in #9c9f87a0c842f52c57656f4c9b662e35e785299c

JasperO98:
Test is now working.

hexylena (approved):
> Ive approved all the commits and fixed the ordering of the input parameters. I don't think I'll need the other parameters in the near future. I'll add them needed. Regarding the hap.py script I did not implement it since the conda version is bugged only som.py works which is also the one that I need.
> Thanks for all the help and edits!

I had the same experience with conda versions, would not run locally due to missing libraries I couldn't solve. The tests seem to be running in a container, those should be identical in theory, but maybe not for some reason.

I've:

- restored the additional outputs (in case someone cares about them)
- and added tests for them (we can use `compare="sim_size"` to make sure the files are about the same size, since they're different due to file paths included.)
- replaced tabs with spaces since that's IUC style preference

Let's see if it passes and if it does, I think good to go

JasperO98:
Ah the "sim_size" is a good idea, but I think it still failed...

hexylena:
Fantastic @JasperO98! Thanks for wrapping this. It'll be uploaded soon to the toolshed and we'll get it installed in emc.

PR #4340, opened 2022-01-22
Add parameter to VarDict to chunk up chromosome.

JasperO98:
To prevent OutOfMemoryError the genome needs to be split into smaller chunks, as suggested here:
- https://github.com/AstraZeneca-NGS/VarDictJava/issues/197
- https://github.com/AstraZeneca-NGS/VarDictJava/wiki/Error-java.lang.OutOfMemoryError:-GC-overhead-limit-exceeded

In the current state vardict will crash when the regions are to large but it will keep on running endlessly in galaxy.
With this additional parameter this can be prevented.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/vardict/vardict.xml
+++ b/tools/vardict/vardict.xml
@@ -32,19 +32,7 @@
         #else
             ln -s ./ref.fa.fai ./chromosomes.fa.fai &&
         #end if
-        awk 'BEGIN{chunk=$chunksize}{
-			i=1;
-			while(i <= $2) {
-				printf $1"\t"i"\t";
-				if(i+chunk< $2){
-					print i+chunk
-				}
-				else {
-					print $2
-				};
-				i=i+chunk
-			}
-		}' ./chromosomes.fa.fai > ./regions.bed &&
+	python $__tool_directory__/split.py ./chromosomes.fa.fai '$chunksize' > ./regions.bed &&
 
         vardict-java
         #if $select_mode.mode == "paired"
--- a/tools/vardict/vardict.xml
+++ b/tools/vardict/vardict.xml
@@ -33,7 +33,7 @@
         #else
             ln -s ./ref.fa.fai ./chromosomes.fa.fai &&
         #end if
-	python $__tool_directory__/split.py ./chromosomes.fa.fai '$chunksize' > ./regions.bed &&
+        python $__tool_directory__/split.py ./chromosomes.fa.fai "\${VARDICT_CHUNKSIZE:-1000000}" > ./regions.bed &&
 
         vardict-java
         #if $select_mode.mode == "paired"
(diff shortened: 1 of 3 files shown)


hexylena:
To review the concept: I think having this run in a constant amount of memory is an improvement than an unbounded version based on the size of the file.

I'm not sure if we should expose the chunk size, or make it an environment variable with a default value, which would let administrators control it.

I've posted a similar message to the IUC matrix channel and we'll see if there are opinions there as well.

tools/vardict/vardict.xml
```diff
@@ -32,7 +32,19 @@
         #else
             ln -s ./ref.fa.fai ./chromosomes.fa.fai &&
         #end if
-        awk 'BEGIN {FS=OFS="\t"} {print $1, "1", $2}' ./chromosomes.fa.fai > ./regions.bed &&
+        awk 'BEGIN{chunk=$chunksize}{
```
hexylena, line 35: this currently gets stuck in an infinite loop for me, I'm not sure why. We can also write this in perl or python, maybe, and it's a bit easier.

I think in any case, however, it should output multiple regions files, rather than a single file.

tools/vardict/vardict.xml
```diff
@@ -32,7 +32,19 @@
...
+					print i+chunk
+				}
+				else {
+					print $2
+				};
+				i=i+chunk
+			}
+		}' ./chromosomes.fa.fai > ./regions.bed &&
 
         vardict-java
```
hexylena, line 49: and then here we'd need something like `for region in regions*.bed`, and then some way to combine all of the results.

Either we could have

JasperO98:
Here is another relevant link for this particular option: https://github.com/AstraZeneca-NGS/VarDict/issues/2
For targeted sequencing or WES we can use bed files provided by the manufacturer.
For WGS an overlap of 150bp is recommended with 5kb segments, the segments could be larger but then it requires more memory.

tools/vardict/vardict.xml
JasperO98: I can do it in python, but why should we use multiple files rather than a single file? One file should work fine right?

JasperO98:
The test failed, since:
 `/tmp/tmpc_fg9_go/job_working_directory/000/4/tool_script.sh: line 22: python: command not found`
Which corresponds to this line:
 https://github.com/galaxyproject/tools-iuc/blob/bb0d8aaafd3c7756954c7d2b253654686e92b91e/tools/vardict/vardict.xml#L35

I looked at other .xml files and those access python in the same way, what is going wrong?

hexylena:
You need to add python to the dependency list.

tools/vardict/vardict.xml
hexylena: Ok, re-read the issue, they really just chunk it and it's fine.

hexylena:
Ok, rewrote it to admin-only controllable variable, per IUC discussion. This looks good otherwise, if tests pass!

hexylena (approved):

JasperO98:
Can we merge this PR now, since the bug should be fixed in the current state?

hexylena:
Bumped the minor version, but, yes, ok from my side. @bgruening fyi potential behaviour change to decrease memory usage :)

PR #4343, opened 2022-01-24
Fix happy repo name

hexylena:
FOR CONTRIBUTOR:

This wasn't caught in linting - only at deployment. I think it's relatively easy to test for but I guess it needs to be added in planemo somewhere.


nsoranzo:
For clarity, this is to fix `Repository names must contain only lower-case letters, numbers and underscore.` in the repository deploy: https://github.com/galaxyproject/tools-iuc/runs/4921056782?check_suite_focus=true

hexylena:
Indeed, thanks @nsoranzo I should've included context.

nsoranzo:
Should we rename also the directory to be `happy` ?

hexylena:
As IUC likes. I don't see a big difference either way as long as `remote_repository_url` is correct which is what folks will look at

hexylena:
@nsoranzo renamed.

nsoranzo (approved):

nsoranzo:
@hexylena Upload failed again with a different error: https://github.com/galaxyproject/tools-iuc/runs/4935722720?check_suite_focus=true

Metadata may have been defined for some items in revision '5095e751304c'.  Correct the following problems if necessary and reset metadata.`hap.py.xml` - This file requires an entry in the `tool_data_table_conf.xml` file.  Upload a file named `tool_data_table_conf.xml.sample` to the repository that includes the required entry to correct this error.

hexylena:
urgh ok. I'm really surprised this isn't getting caught in linting or tests, it's too bad

hexylena:
opened https://github.com/galaxyproject/tools-iuc/issues/4349 as a result of this, I found reference to that in the reviewer guide but since it wasn't mentioned in an planemo linting errors, didn't even think about this potential case. TIL there's a 40 bullet point checklist I should've followed

PR #4346, opened 2022-01-25
Update nextclade to 1.10.1

pvanheus:
FOR CONTRIBUTOR:


pvanheus:
Nextclade CLI 1.10.0 has been [released](https://github.com/nextstrain/nextclade/releases). Besides a version bump this introduces a new JSON configuration file (which is supported in the "advanced options")

bgruening:
Failing the tests as it seems.

pvanheus:
> Failing the tests as it seems.

Yes, that was a problem with nextalign. Passing now.

pvanheus:
The latest change in this PR is switching how the Nextclade reference data is handled: instead of specifying each file individually, the dataset directory is specified and then individual files are specified if the user provides values for them. This change was made after the Nextclade team clarified that the semantics of the tool allow to "override" like this.

wm75:
Thanks @pvanheus! Looks like a real simplification of things :)

Couple of remarks:
- 1.10.1 is available by now and while it shouldn't include any changes relevant for the wrapper you might just want to update to it to show that everything's up to date
- why the new more complicated checks like `#if $adv.input_root_seq and str($adv.input_root_seq).strip()`? When would the first condition be True, but not the second?
- I just noticed on re-reviewing the code that your
  ```
            #if str($adv.advanced_options) == "yes"
                $adv.include_reference
            #end if
  ```
  doesn't work since `include_reference` doesn't specify true/falsevalues.
  So this leads to literal `true` / `false` being included on the command line and the tool always failing when `output_fasta` is selected in `outputs`.

pvanheus:
> Thanks @pvanheus! Looks like a real simplification of things :)
> 
> Couple of remarks:
> 
>     * 1.10.1 is available by now and while it shouldn't include any changes relevant for the wrapper you might just want to update to it to show that everything's up to date
> 
Good point. The main change on their side was w.r.t. error messages.
>     * why the new more complicated checks like `#if $adv.input_root_seq and str($adv.input_root_seq).strip()`? When would the first condition be True, but not the second?
> 
I'm reverting to a simpler form. I was confusing myself here.
>     * I just noticed on re-reviewing the code that your
>       ```
>                 #if str($adv.advanced_options) == "yes"
>                     $adv.include_reference
>                 #end if
>       ```
>       doesn't work since `include_reference` doesn't specify true/falsevalues.
>       So this leads to literal `true` / `false` being included on the command line and the tool always failing when `output_fasta` is selected in `outputs`.
Well spotted and now fixed, with a test added.

wm75 (approved):
Perfect. This looks good to go to me!

PR #131, opened 2015-05-08
[WIP] Emboss 6.6.0

erasche:
Just opening for comments along the way. There's an existing acd2xml parser of some sort in debian, but this was an excuse to learn lex/yacc ;)

I've got ACD files converted to XML (it's ugly but whatever). This can then be reparsed into CTD or Galaxy XML by some other scripts (since ACD->GalaxyXML would be nice but my c++/flex/yacc skills aren't up to that challenge)

Issue #65, opened 2015-02-27
EMBOSS tools needs an update


bgruening:
We need a new version of all EMBOSS tools. We should write a converter from [EMBOSS ACD](http://emboss.sourceforge.net/developers/acd/syntax.html) files to Galaxy xml files.

Please get in touch with us if you are interested in this project.

erasche:
I had one somewhere at one point. Crap. It's probably gone now. Willing to help though :)

erasche:
Nope, gone now. I remember the main problem with my automatic conversion was that there were a bunch of "unlisted" arguments like "seqNformat" (or something like that)

bgruening:
Binaries can be build with: https://github.com/natefoo/docker-build/pull/2

erasche:
+1 for prebuilt binaries!

erasche:
Depends on https://github.com/galaxyproject/docker-build/issues/10

nsoranzo:
:+1:

innovate-invent:
Emboss 6 piece-meal update:

- ALIGNMENT
  - [ ] extractalign
- ALIGNMENT CONSENSUS
  - [ ] cons
  - [ ] consambig
  - [ ] megamerger
  - [ ] merger
- ALIGNMENT DIFFERENCES
  - [ ] diffseq
- ALIGNMENT DOT PLOTS
  - [ ] dotmatcher
  - [ ] dotpath
  - [ ] dottup
  - [ ] polydot
- ALIGNMENT GLOBAL
  - [ ] est2genome
  - [ ] needle
  - [ ] needleall
  - [ ] stretcher
- ALIGNMENT LOCAL
  - [ ] matcher
  - [ ] seqmatchall
  - [ ] supermatcher
  - [ ] water
  - [ ] wordfinder
  - [ ] wordmatch
- ALIGNMENT MULTIPLE
  - [ ] edialign
  - [ ] emma
  - [ ] infoalign
  - [ ] plotcon
  - [ ] prettyplot
  - [ ] showalign
  - [ ] tranalign
- DATA RESOURCES
  - [ ] dbtell
  - [ ] servertell
  - [ ] showdb
  - [ ] showserver
- DATA RETRIEVAL
  - [ ] codcopy
  - [ ] seqret
- DATA RETRIEVAL CHEMISTRY DATA
  - [ ] infobase
  - [ ] inforesidue
- DATA RETRIEVAL FEATURE DATA
  - [ ] featcopy
[...]

PR #4347, opened 2022-01-25
Add EMBOSS 6: fuzznuc

innovate-invent:
It looks like people keep starting to implement automated EMBOSS 6 wrapper solutions but then abandoning them. I propose a more manual peace-meal approach based on interest.

I have created a new folder for emboss 6 given that the current emboss folder has its version number baked in. For future versions of emboss we can then reuse this folder. I have also removed special characters from the tool ID to improve compatibility with downstream scripts. The perl dependency was removed. I think it would be best to avoid extra dependencies if possible, removing the need for a special mulled container.

New parameters for fuzznuc have been included.

~~Some work should be done to remove the emboss_format_corrector.py but this appears like it might be non-trivial.~~
emboss_format_corrector.py was outdated and removed

FOR CONTRIBUTOR:

Related #131


innovate-invent:
The test failure says to check the artifacts but I don't see any...

wm75:
> The test failure says to check the artifacts but I don't see any...

They are a bit hard to discover nowadays: click on "Summary" in the navigation bar on the left of the Action page. The Artifacts are at the very bottom of that section.

tools/emboss/emboss_format_corrector.py
```diff
@@ -0,0 +1,47 @@
+# EMBOSS format corrector
+import operator
```
bernt-matthias, line 2: Always thought code files are evil :)?

Can this also be done using

- output actions or
- `galaxy.json`, see tools/obitools/macros.xml, or?

innovate-invent: I was procrastinating on my work and decided to dig into this. It looks like that format corrector was just serving outdated functionality and could be easily removed.

tools/emboss/emboss_fuzznuc.xml
```diff
@@ -0,0 +1,133 @@
+<tool id="emboss_fuzznuc37" name="fuzznuc" version="6.6.0">
```
bernt-matthias, line 1: Can you use to version token here? And maybe rename to `TOOL_VERSION`?

bernt-matthias: In there a reason for `37` in the id?

tools/emboss/emboss_fuzznuc.xml
```diff
@@ -0,0 +1,133 @@
...
+      <param name="complement" value="no"/>
+      <param name="out_format1" value="excel"/>
+      <param name="seq_qualifiers|supper1" value="true"/>
+      <param name="seq_qualifiers|sbegin1" value="100"/>
+      <param name="out_qualifiers|rmaxseq2" value="5"/>
+      <output name="out_file1" file="emboss_fuzznuc_out2.tabular"/>
+    </test>
+  </tests>
+  <help>
+.. class:: warningmark
```
bernt-matthias, line 122: A bit more help would be appreciated. Maybe just what the tool does?

tools/emboss/macros.xml
```diff
@@ -0,0 +1,69 @@
...
+                <add value="["/>
+                <add value="]"/>
+                <add value="."/>
+                <add value=","/>
+            </valid>
+        </sanitizer>
+        <validator type="empty_field" />
+        <validator type="regex" message="Pattern must not end with backslash.">.*[^\\]$</validator>
+    </xml>
+    <xml name="report_formats" token_output="out_format1">
```
bernt-matthias, line 49:
```suggestion
    <xml name="report_formats" tokens="output">
````

would render the xml macro token a required token.

File: tools/emboss/macros.xml [hunk 5]
```diff
@@ -0,0 +1,69 @@
...
+        </xrefs>
+    </xml>
+    <xml name="citations">
+        <citations>
+            <citation type="doi">10.1016/S0168-9525(00)02024-2</citation>
+            <citation type="doi">10.1101/gr.5578007</citation>
+            <yield />
+        </citations>
+    </xml>
+    <xml name="regex_sanitizer">
```
bernt-matthias, line 26: This seems unused.

bernt-matthias (commented):
Maybe we can add a todo list to this old issue https://github.com/galaxyproject/tools-iuc/issues/65 .. keeping track of updating all the emboss tools?

tools/emboss/emboss_fuzznuc.xml
innovate-invent: sure, I have no idea why the 37. It was copied from emboss_5.

tools/emboss/macros.xml
innovate-invent: Macros were copied forward from emboss_5 if anyone wants to reuse them for other tools. I can remove them if you think it isn't worth it.

tools/emboss/emboss_fuzznuc.xml
bernt-matthias: Now I see. I did non get in the first place that you created a new folder. 

With the change of the id we are destroying tool lineare (I guess) .. but keeping the space and upper case letters also seems to be no good option? 

I guess we need someone here who knows better about tool lineage...

innovate-invent: This will be published to a new tool repo and will not have the same lineage as emboss_5.
The only way to have it in the same lineage would be to have emboss 6 in the emboss_5 repo.

tools/emboss/.shed.yml
```diff
@@ -0,0 +1,13 @@
...
+- Sequence Analysis
+- Fasta Manipulation
+description: EMBOSS tools
+homepage_url: http://emboss.open-bio.org/
+long_description: |
+    The European Molecular Biology Open Software Suite (EMBOSS) is a high quality, well documented package of open
+    source software tools for molecular biology. It includes over 200 applications for molecular sequence analysis
+    and other common tasks in bioinformatics.
+name: emboss
+owner: devteam
```
nsoranzo, line 11:
```suggestion
owner: iuc
```

tools/emboss/.shed.yml
```diff
@@ -0,0 +1,13 @@
+categories:
+- Sequence Analysis
+- Fasta Manipulation
+description: EMBOSS tools
+homepage_url: http://emboss.open-bio.org/
```
nsoranzo, line 5:
```suggestion
homepage_url: http://emboss.sourceforge.net/
```

tools/emboss/emboss_fuzznuc.xml
nsoranzo: > The only way to have it in the same lineage would be to have emboss 6 in the emboss_5 repo.

True, as far I as I understand, tool lineage is maintained only if also the ToolShed owner and repository name are the same (i.e. the whole `guid` minus the version).

> This will be published to a new tool repo and will not have the same lineage as emboss_5.

As you've already noticed, 3 years ago I updated one tool in `emboss_5` to v6.6.0 to keep the lineage. I see 2 alternative options here:
1. Keep updating (or adding) tools in the `emboss_5` directory.
   Pros: preserve the lineage of the existing tools
   Cons: the `emboss_5` repository becomes increasingly a misnomer; all-in-one repository
2. Create a new `emboss_suite` suite with a separate repository for each new/updated tool and sane tool IDS (like in this PR, but you would need to modified the `.shed.yml` like in https://github.com/galaxyproject/tools-iuc/blob/master/tools/bcftools/.shed.yml )
   Pros: one tool per repository; sane tool IDs; no "5" in the repo names
   Cons: tool lineages broken, i.e. no clean upgrade path for workflows using these tools

innovate-invent: Is there value in a suite if it is all the same dependency?

I am not sure that there would be a clean upgrade path for workflows sticking with the emboss_5 repo. There are going to be a lot of incompatible changes in the major version bump of the tools.

nsoranzo: > Is there value in a suite if it is all the same dependency?

I'll leave this one for @mvdbeek to reply :)

bgruening:
What is the status here @nsoranzo and @bernt-matthias? This will be deployed as IUC. I guess the idea is to deprecate the `devteam` tools?

bernt-matthias:
Main/Last question is tool lineage. 

Wondering if we can just mix versions in the old dir. Given that emboss has seen no updates in a long time (http://emboss.sourceforge.net/developers/changelog.html) we can hope that the tools will converge to the latest at one point. 

Has someone ever tried to bump emboss?

bgruening:
The old tools are all in one repo aren't they?

bernt-matthias:
> The old tools are all in one repo aren't they?

yes

> Has someone ever tried to bump emboss?

remember that there are barely any tests...

PR #4348, opened 2022-01-26
Add compression tool

astrovsky01:
FOR CONTRIBUTOR:


bernt-matthias:
I would prefer this https://github.com/galaxyproject/galaxy/pull/13223

mvdbeek:
It serves a different purpose, this tool can compress anything, https://github.com/galaxyproject/galaxy/pull/13223 is a datatype converter that can convert to datatypes that can be compressed

tools/uncompressed_to_gz/uncompressed_to_gz.xml
```diff
@@ -0,0 +1,29 @@
...
+        <requirement type="package" version="1.11">gzip</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        gzip -cf ${input} > output_file.gz
+    ]]></command>
+    <inputs>
+        <param name="input" type="data" format="data" label="Select data to compress"/>
+    </inputs>
+    <outputs>
+        <data name="output_file" format="gz" from_work_dir="output_file.gz" label="GZIP compression on ${on_string}"/>
```
bgruening, line 12: gz is not an official Galaxy filetype. https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/sample/datatypes_conf.xml.sample

Maybe it should be?

astrovsky01: That was actually what prompted the tool rather than the converter. I can either switch it over to 'data' or add a datatype. It's so generic I don't see it being passed to many different tools after compression, mostly exporting.

tools/uncompressed_to_gz/uncompressed_to_gz.xml
```diff
@@ -0,0 +1,29 @@
...
+            <param name="input" value="1.csv"/>
+            <output name="output_file" file="1.csv.gz" compare="sim_size"/>
+        </test>
+        <test expect_num_outputs="1">
+            <param name="input" value="1.csv"/>
+            <output name="output_file" file="1.csv" decompress="true"/>
+        </test>
+    </tests>
+    <help><![CDATA[
+        Compress files before exporting. If compressing a collection, all elements within that collection will be compressed, but the collection itself will not be.
```
bgruening, line 26: the tool has nothing to do with exporting ... so I would not write this into the help.

tools/uncompressed_to_gz/uncompressed_to_gz.xml
```diff
@@ -0,0 +1,29 @@
+<tool id="uncompressed_to_gz" name="Compress file(s) using gzip" version="0.1.0">
```
bgruening, line 1: maybe we make the ID and the name more generic, we might want to add more compressions algorithms later?

bgruening:
@astrovsky01 here are now two tools included, is that correct?

astrovsky01:
It showed it as moved, rather than duplicated before, and it's not letting me delete the redundant files locally -- it should just be compress_file.xml

bgruening:
+1 from my side, if we add the gz filetype to Galaxy.

bgruening (approved):

PR #4358, opened 2022-01-29
add bioconductor IDs (dois) as xref

bgruening:
This PR adds Bioconductor DOI as a cross-reference to some main bioconductor tools.

Every BioC package has a DOI, that we can use, those are DOIs but I added them as `xref`. I did that since I assume people want to cite the manuscript DOI and we want to render the BioC xref maybe with a fancy logo. I also only annotated the *main* dependency of a tool. It can happen that a tool has multiple dependencies. In that case I only added the main BioC package as xref.

@csoneson does this look ok for you?


tools/deseq2/deseq2_macros.xml
```diff
@@ -49,7 +49,8 @@
     </xml>
     <xml name="xrefs">
         <xrefs>
-          <xref type='bio.tools'>DESeq2</xref>
+            <xref type='bio.tools'>DESeq2</xref>
+            <xref type='bioconductor'>10.18129/B9.bioc.deseq2</xref>
```
nsoranzo, line 53:
```suggestion
            <xref type="bio.tools">DESeq2</xref>
            <xref type="bioconductor">10.18129/B9.bioc.deseq2</xref>
```

csoneson:
This looks good to me!

bgruening:
We need to remove the `10.18129/B9.bioc.` from this and then it is ready to go.

bgruening:
Ok, if someone wants to pull the trigger we have a few tools freshly annotated and linked them out to BioC with the new Galaxy release.

hrhotz:
is this something for our discussion (Galaxy/Bioc) on Wednesday (Sep.28th)

bgruening:
@galaxyproject/iuc do you think we can merge this (without uploading)?

bernt-matthias:
> without uploading

Why not? 

Maybe its worth to fix lint and test issues?

bgruening:
> > without uploading
> 
> Why not?
> 
> Maybe its worth to fix lint and test issues?

That seems to be a lot of work, that I have no time in the next two month. If we merge it it will get populated with the next update I had hoped :-(

bernt-matthias:
Fine with me .. given that its likely that the bioconductor packages will have updates (otherwise the updates would never become effective).

bgruening:
@bernt-matthias can you pull the trigger, please?

bernt-matthias:
Sure. Since tests / lint will fail the deploy job won't start at all. 

Wondering if this should also trigger an error message here?

PR #4364, opened 2022-01-31
Updating tools/proteinortho from version 6.0.32 to
6.1.2

planemo-autoupdate:
Hello! This is an automated update of the following tool: **tools/proteinortho**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

I have updated tools/proteinortho from version 6.0.32 to 6.0.33.

**Project home page:** https://gitlab.com/paulklemm_PHD/proteinortho

PR #4375, opened 2022-02-03
Updating tools/proteinortho from version 6.0.32 to 6.0.33 - added "--subparaMakeBlast"

mira-miracoli:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/proteinortho/proteinortho.xml
+++ b/tools/proteinortho/proteinortho.xml
@@ -108,6 +108,7 @@
             <param argument="--identity" type="integer" value="25" min="0" max="100" label="Minimal percent identity of best blast hits in %"/>
             <param argument="--selfblast" type="boolean" checked="false" truevalue="--selfblast" falsevalue="" label="Apply selfblast, detects paralogs without orthologs "/>
             <param argument="--singles" type="boolean" checked="false" truevalue="--singles" falsevalue="" label="Report singleton genes without any hit "/>
+            <param argument="--subparaMakeBlast" type="text" optional="true" label="additional parameters for the database generation (-p=blast -> makeblastdb,diamond -> diamond makedb,...) example -subpara='--ignore-warnings' for a common diamond error" />
             <param argument="--isoform" type="select" label="Use isoform information" help="The reciprocal best hit graph is build using isoform information (isoforms are treated equivalent). For ncbi : simply add the additional files to the input (file names need to match). For uniprot : the isoforms need to contain the word isoform and the corresponding identifier. For trinity simply use the trinity output format.">
                 <option value="no" selected="true">Don't use isoform information</option>
                 <option value="ncbi">ncbi style (..._additional.fasta)</option>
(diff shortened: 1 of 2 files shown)


tools/proteinortho/proteinortho.xml
```diff
@@ -108,6 +108,7 @@
             <param argument="--identity" type="integer" value="25" min="0" max="100" label="Minimal percent identity of best blast hits in %"/>
             <param argument="--selfblast" type="boolean" checked="false" truevalue="--selfblast" falsevalue="" label="Apply selfblast, detects paralogs without orthologs "/>
             <param argument="--singles" type="boolean" checked="false" truevalue="--singles" falsevalue="" label="Report singleton genes without any hit "/>
+            <param argument="--subparaMakeBlast" type="text" optional="true" label="additional parameters for the database generation (-p=blast -> makeblastdb,diamond -> diamond makedb,...) example -subpara='--ignore-warnings' for a common diamond error" />
```
simonbray, line 111: @mira-miracoli thanks a lot for adding the new parameter! A couple of comments:
* you should add the `--subparaMakeBlast` to the `<command>` section above, so the command which is executed will be modified when the new parameter is specified.
* I think we should not allow users to enter parameters directly via a text `<param>` - we should choose a selection (such as `--ignore-warnings`) and expose each via a separate `<param>` via the Galaxy interface. I am not sure which are useful to include, we probably need some feedback from @bernt-matthias here.

simonbray (commented):
Did you try pushing to the original PR? Something like
```
git remote add pau git@github.com:planemo-autoupdate/tools-iuc.git
git push pau tools/stringtie
```
should work?

I think it would be cleaner to stick to a single PR if possible for these updates.

mira-miracoli:
@simonbray ah I'm sorry, I'll do it the way you suggested :)

The parameter I added could be a security risk (allows uncontrolled input) - would be better to merge it without and delete my commit, I guess.

simonbray:
> @simonbray ah I'm sorry, I'll do it the way you suggested :)

No problem, maybe something for next time, seeing as you already opened this PR :)

> The parameter I added could be a security risk (allows uncontrolled input) - would be better to merge it without and delete my commit, I guess.

Yes, exactly. Or we individually parameterize all the options that could be passed in, which is a lot more work. Unfortunately I never used this tool so I don't know if that is useful or worthwhile.

tools/proteinortho/proteinortho.xml
bernt-matthias: As long as the blast and diamond requirement is not pinned (can someone check if it is in bioconda?) this might cause a lot of trouble. 

`--ignore-warnings` is certainly of interest because otherwise nucleotide sequences might not work.

What else is available?

simonbray: Blast and diamond are both pinned in the requirements.

> What else is available?

From my cursory read of that gitlab issue it seems any of the `diamond makedb` parameters can be passed in this way.

bgruening:
How do we proceed here?

pmjklemm:
most up-to-date version can be found here: #4364

PR #4377, opened 2022-02-03
Update fastp to version 0.23.2

pvanheus:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/fastp/macros.xml
+++ b/tools/fastp/macros.xml
@@ -1,5 +1,5 @@
 <macros>
-    <token name="@WRAPPER_VERSION@">0.20.1</token>
+    <token name="@WRAPPER_VERSION@">0.23.2</token>
 
     <xml name="adapter_trimming_options">
         <section name="adapter_trimming_options" title="Adapter Trimming Options" expanded="False">


bgruening (approved):
No new options?

pvanheus:
> No new options?

Do you mean, does the fastp command offer new options in this version? I don't know. What I *do* know is that the current version of fastp available in the toolshed (and on usegalaxy.eu) is broken because it load plotly from a http URL (http://opengene.org/plotly-1.2.0.min.js), and this loading gets blocked by modern web browsers (causing the fastp graphs to not be rendered). This impacts all Galaxy servers that are served over https. The latest version of fastp (0.23.2) doesn't have this problem (it loads plotly from a https URL).

bgruening:
That is a good reason to merge it :)

wm75:
Ah, just a bit late - there is in fact a new deduplication option in this version, but it's ok we don't need this urgently.
The new version is also supposed to be faster for compressed fastq input so that's relevant and the update comes just in time to include it in our next SARS-CoV-2 variation workflow version :)
Thanks @pvanheus !

PR #4384, opened 2022-02-04
Updating tools/read_it_and_keep from version 0.1.0 to
0.2.2

planemo-autoupdate:
Hello! This is an automated update of the following tool: **tools/read_it_and_keep**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

I have updated tools/read_it_and_keep from version 0.1.0 to 0.2.1.

**Project home page:** https://github.com/GenomePathogenAnalysisService/read-it-and-keep/releases


simonbray (approved):
only change is `Improved MacOS compiler compatibility`

simonbray:
> only change is `Improved MacOS compiler compatibility`

Sorry, I was too fast; there's also `Adds new command line option --enumerate_names` from version 0.2.0

bgruening:
ping @mira-miracoli :)

tools/read_it_and_keep/read-it-and-keep.xml
```diff
@@ -124,6 +125,7 @@
                 <param name="ref_fasta" value="reference.fasta" ftype="fasta" />
             </conditional>
             <param name="sequencing_tech" value="illumina" />
+            <param name="enumerate_names" value="True" />
```
simonbray, line 128:
```suggestion
            <param name="enumerate_names" value="--enumerate_names" />
```
Is the test passing at the moment? I think you need to specify it like this.

tools/read_it_and_keep/read-it-and-keep.xml
```diff
@@ -79,6 +79,7 @@
         <section name="adv" title="Advanced options">
             <param argument="--min_map_length" type="integer" min="0" value="50" label="Shortest match required to keep a read (in bp)" />
             <param argument="--min_map_length_pc" type="float" min="0.0" max="100.0" value="50.0" label="Minimum length of match required to keep a read (as percentage of read length" />
+            <param name="enumerate_names" type="boolean" truevalue="--enumerate_names" falsevalue="" label="rename the reads 1,2,3,... (for paired reads, will also add /1 or /2 on the end of names))" />
```
simonbray, line 82:
```suggestion
            <param name="enumerate_names" type="boolean" truevalue="--enumerate_names" falsevalue="" label="Rename the reads 1,2,3,... (for paired reads, will also add /1 or /2 to the end of names)" />
```

wm75:
@mira-miracoli you also need to use the new param somewhere in the command block

wm75:
and while you're at it, you may also want to implement a `<version_command>`- `readItAndKeep --version` should be all that's needed.

tools/read_it_and_keep/read-it-and-keep.xml
```diff
@@ -124,6 +126,7 @@
                 <param name="ref_fasta" value="reference.fasta" ftype="fasta" />
             </conditional>
             <param name="sequencing_tech" value="illumina" />
+            <param name="enumerate_names" value="--enumerate_names" />
```
wm75, line 129: Here, `value="true"` is correct :)
You want to test that the param, when checked, evaluates to its truevalue.
Your current test will also work but it short-circuits the logic so is less conclusive.

wm75: and, of course, you'll have to adjust the test data since the option will change all read names in the output. It's actually a good thing though to have the test fail once to demonstrate it can catch something, then fix it.

mira-miracoli: ok first version was correct @simonbray 
I'll fix that, too

tools/read_it_and_keep/read-it-and-keep.xml
```diff
@@ -106,8 +109,16 @@
                 <param name="ref_fasta" value="reference.fasta" ftype="fasta" />
             </conditional>
             <param name="sequencing_tech" value="illumina" />
-            <output name="output_reads1" value="output_test1.fastq.gz" />
-            <output name="output_reads2" value="output_test2.fastq.gz" />
+            <output name="output_reads1" file="output_test1.fastq.gz" >
```
bernt-matthias, line 112: Tests fail because: 

```
Collection element forward of collection output_collection:  different than expected, difference (using diff):
( /tmp/tmpesn8_wjsoutput_test1.fastq.gz v. /tmp/tmpstccecuhoutput_test1.fastq.gz )
Binary data detected, not displaying diff
```

Maybe we can just remove the test files and rely on the size comparison?

tools/read_it_and_keep/read-it-and-keep.xml
```diff
@@ -139,7 +158,12 @@
                 <param name="ref_fasta" value="reference.fasta" ftype="fasta" />
             </conditional>
             <param name="sequencing_tech" value="ont" />
-            <output name="output_reads1" value="output_test3.fastq.gz" />
+            <param name="enumerate_names" value="True" />
+            <output name="output_reads1" file="output_test3.fastq.gz">
```
bernt-matthias, line 162:
```suggestion
            <output name="output_reads1" file="output_test3.fastq.gz">
```

then it should work

bgruening: You can accept this suggestion via the github interface @mira-miracoli

bgruening (approved):

wm75:
@mira-miracoli This is unrelated to your update efforts, but can you check whether this tool can actually handle *uncompressed* fastq input (from a very quick look at the source, it seems it can't, but knows how to handle `fasta.gz` instead).
If that's the case, the `@FASTQ_FORMATS@` token would have to be adjusted (and mayb be renamed to just `@INPUT_FORMATS@`.

Thanks for all your work!

mira-miracoli:
> @mira-miracoli This is unrelated to your update efforts, but can you check whether this tool can actually handle *uncompressed* fastq input (from a very quick look at the source, it seems it can't, but knows how to handle `fasta.gz` instead).
> If that's the case, the `@FASTQ_FORMATS@` token would have to be adjusted (and mayb be renamed to just `@INPUT_FORMATS@`.
> 
> Thanks for all your work!

Sure, I can do! Would next week be fast enough? Thank you for your reviews :)

wm75:
> Sure, I can do! Would next week be fast enough? Thank you for your reviews :)

Of course, take your time :)

mira-miracoli:
@wm75 I had a look at the source code and the relevant libraries and I think it should be able to handle fastq as well.
I found no specific file format check for the input file in the read-it-and-keep source.
According to the [source](https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/zlib.h#L1328) of zlib.h it doesn't matter whether the file is compressed or not. I found similar information about the [kseq.h](http://attractivechaos.github.io/klib/#Kseq%3A%20stream%20buffer%20and%20FASTA%2FQ%20parser) library - kseq should be able to parse both fasta and fastq, compressed or uncompressed.
I can adjust the format tokens now, if you want.

wm75:
@mira-miracoli Thanks for checking, and for explaining why that works!
Yes, then just adding fasta and fasta.gz as supported input formats would be awesome.
In addition, it would be good to stick to the input format on output. The current version behaves like the command line tool, i.e. produces gzipped output unconditionally, and you'd have to add a decompression step at the end if the input was uncompressed, plus make the format in the output section input-dependent, too.

tools/read_it_and_keep/read-it-and-keep.xml
```diff
@@ -19,17 +20,23 @@
         #if $reads.read_type == "paired"
             ln -s '$reads.read1' read1.fastq &&
             ln -s '$reads.read2' read2.fastq &&
+            ln -s '$reads.read1.ext' read1.ext &&
+            ln -s '$reads.read2.ext' read2.ext &&
```
wm75, line 24: With `.ext` you're only getting the typical extension of the input dataset's type, not the full file name of the file.
You could do something like
`ln -s '$reads.read1' read1.${reads.read1.ext}`
where `$reads.read1` would evaluate to some Galaxy-managed .dat file and `${reads.read1.ext}` would be `fasta` for example. However, I don't think this tool looks at the file suffix at all so you can also just symlink the read1 dataset to `read1` and the read2 dataset to `read2` in all cases.

tools/read_it_and_keep/read-it-and-keep.xml
```diff
@@ -79,19 +105,42 @@
         <section name="adv" title="Advanced options">
             <param argument="--min_map_length" type="integer" min="0" value="50" label="Shortest match required to keep a read (in bp)" />
             <param argument="--min_map_length_pc" type="float" min="0.0" max="100.0" value="50.0" label="Minimum length of match required to keep a read (as percentage of read length" />
+            <param name="enumerate_names" type="boolean" truevalue="--enumerate_names" falsevalue="" label="Rename the reads 1,2,3,... (for paired reads, will also add /1 or /2 to the end of names)" />
         </section>
     </inputs>
     <outputs>
         <data name="output_reads1" format="fastqsanger.gz" label="Filtered reads ${on_string} - reads1" from_work_dir="output.reads_1.fastq.gz">
```
wm75, line 112:
```suggestion
        <data name="output_reads1" format_source="reads.read1" label="Filtered reads ${on_string} - reads1" from_work_dir="output.reads_1.fastq*">
```

This will discover gzipped or uncompressed files and set the format according to the input.
Do the same for the other output cases.

mira-miracoli: Thank you for your help! :)

mira-miracoli:
> @mira-miracoli I fixed an issue with `format_source` not handling qualified/nested parameters. I also found that the output suffix of the command line tool reflects the detected input format (i.e. it's .fasta(.gz) for fasta, and .fastq(.gz) for fastq) and included logic to handle this.
> 
> What's missing are additional tests for fasta and uncompressed fastq inputs. Can you add those? Then I think we're done here.

Thank you! Yes I can add those tests.

wm75:
Thanks @mira-miracoli !
Final cleanup:
- I think all `output.test1.*` test files are identical to their `output.test2.*` counterparts (cause the only difference between those tests is whether the input comes as two individual datasets or as a paired collection) -> please remove the duplicates
- All `output.*.fasta.gz` are actually plain text fasta files. That's because you're declaring `ftype="fasta"` for the `fasta.gz` inputs in those tests. When you do that, planemo uploads the .gz files to Galaxy, which decompresses them *before* feeding them to the tool. In other words you're not really testing for handling of fasta.gz in the wrapper (only if Galaxy knows how to decompress the format).
- A better way would be the other way round: use the fasta.gz input files in all tests related to Fasta format. Declare `ftype="fasta.gz"` for tests of fasta.gz handling, but `ftype="fasta" for testing fasta handling. This way you can get rid of the uncompressed input files. I assume the same trick should work for fastqsanger tests.

wm75:
Thank you @mira-miracoli for bearing with me and my requests for additional changes.
This started out as a simple version bump, but the result now is a significantly enhanced wrapper so thanks for all your work on this!

wm75 (approved):

tools/read_it_and_keep/read-it-and-keep.xml
```diff
@@ -1,12 +1,13 @@
 <tool id="read_it_and_keep" name="Read It and Keep" version="@TOOL_VERSION@+galaxy0" profile="20.09">
     <macros>
-        <token name="@FASTQ_FORMATS@">fastq,fastq.gz,fastqsanger,fastqsanger.gz</token>
-        <token name="@TOOL_VERSION@">0.1.0</token>
+        <token name="@INPUT_FORMATS@">fasta,fastq,fastq.gz,fasta.gz,fastqsanger,fastqsanger.gz</token>
```
bernt-matthias, line 3: `fastq` should imply fastqsanger and also the other fastq formats. I think this is unwanted / redundant?

tools/read_it_and_keep/read-it-and-keep.xml
```diff
@@ -79,22 +97,176 @@
...
+            </conditional>
+            <param name="sequencing_tech" value="ont" />
+            <param name="enumerate_names" value="True" />
+            <output name="output_reads1" file="output.test3.reads_1.fastq" ftype="fastqsanger"/>
+        </test>
+        <!--Testing uncompressed fasta input-->
+        <test expect_num_outputs="2">
+            <conditional name="reads">
+                <param name="read_type" value="paired" />
+                <param name="read1" value="test1.fasta.gz" ftype="fasta" />
```
bernt-matthias, line 169: `ftype` should be `fasta.gz`

tools/read_it_and_keep/read-it-and-keep.xml
```diff
@@ -79,22 +97,176 @@
...
+            <param name="sequencing_tech" value="illumina" />
+            <output_collection name="output_collection" type="paired">
+                <element name="forward" file="output.test1.reads_1.fastq" ftype="fastqsanger"/>
+                <element name="reverse" file="output.test1.reads_2.fastq" ftype="fastqsanger"/>
+            </output_collection>
+        </test>
+        <test expect_num_outputs="1">
+            <conditional name="reads">
+                <param name="read_type" value="single" />
+                <param name="read1" value="test3.fastq.gz" ftype="fastqsanger"/>
```
bernt-matthias, line 155: `ftype` should be `fastqsanger.gz`

tools/read_it_and_keep/read-it-and-keep.xml
```diff
@@ -79,22 +97,176 @@
...
+            <param name="sequencing_tech" value="illumina" />
+            <output name="output_reads1" file="output.test1.reads_1.fasta" ftype="fasta"/>
+            <output name="output_reads2" file="output.test1.reads_2.fasta" ftype="fasta"/>
+        </test>
+        <test expect_num_outputs="3">
+            <conditional name="reads">
+                <param name="read_type" value="paired_collection" />
+                <param name="paired_reads">
+                    <collection type="paired">
+                        <element name="forward" value="test1.fasta.gz" ftype="fasta" />
```
bernt-matthias, line 185: please check all the `ftype` attributes of test in and outputs

tools/read_it_and_keep/read-it-and-keep.xml
```diff
@@ -106,16 +278,16 @@
                 <param name="ref_fasta" value="reference.fasta" ftype="fasta" />
             </conditional>
             <param name="sequencing_tech" value="illumina" />
-            <output name="output_reads1" value="output_test1.fastq.gz" />
-            <output name="output_reads2" value="output_test2.fastq.gz" />
+            <output name="output_reads1" decompress="true" file="output.test1.reads_1.fastq" ftype="fastqsanger.gz"/>
```
bernt-matthias, line 281: `ftype`?

bernt-matthias: `decompress`?

bernt-matthias (changes requested):

tools/read_it_and_keep/read-it-and-keep.xml
wm75: see my explanation above

tools/read_it_and_keep/read-it-and-keep.xml
wm75: Here the wrapper is expected to produce "fastqsanger.gz", but the result gets compared to a decompressed fastq used in other tests.

tools/read_it_and_keep/read-it-and-keep.xml
bernt-matthias: Not sure if this really works. 

I think the automatic extraction only works if the tool accepts only the uncompressed type and a compressed file is given in the test (then an additional converter job should run). But here the tool accepts compressed and uncompressed. 

I'm not 100% sure. So I would suggest to add `file ...` calls to the tool command to verify what file types the command gets.

If I'm wrong we should document this somewhere (if it isn't already) .. clearly this behavior would be handy.

wm75: Well, the tests themselves kind of prove that's working. The wrapper respects the input format on output and the test is checking the output format -> the tool must have seen fasta.

bernt-matthias: Note that Galaxy currently basically does not check `ftype` at all in tests: https://github.com/galaxyproject/galaxy/pull/12073 .. but it's difficult to fix.

wm75: So lets just ask the tests: I've pushed a commit that violates the test assumptions. Lets see if that makes the tests fail.

bernt-matthias: Seems really to work :) I checked with `file ...`. Also your argument with `format_source` and checking the output `ftype` is convincing. 

Do you know if this is documented somewhere? This is really cool, because it can save loads of test data size. 

Last thing I noticed is that the version of the python requirement should be specified

wm75: > Do you know if this is documented somewhere? This is really cool, because it can save loads of test data size.

Don't know if it's documented. I thought it's a logical extension of Galaxy's normal upload behavoir (you can do the same thing from the UI Upload data dialog). Never thought of using this in tests before until @mira-miracoli did so by accident in one of her commits :)

wm75: Ok, and the tests are just failing now as expected with the expected format complaints.

bernt-matthias (approved):

bernt-matthias:
https://github.com/galaxyproject/galaxy/pull/13568

PR #4389, opened 2022-02-08
InterProScan: add doc for licensed components

abretaud:
FOR CONTRIBUTOR:

Just adding some doc on how to install signalp etc for interproscan
ping @bgruening if you want to test this on .eu

### DIFF
--- a/tools/interproscan/README.md
+++ b/tools/interproscan/README.md
@@ -10,3 +10,71 @@ The standard install of InterProScan does not include the following components, 
 - SMART data (THRESHOLD file)
 
 As a Galaxy instance admin, you can install those manually if needed, as long as you (and the Galaxy users) respect the respective licenses.
+
+## Installing licensed components manually
+
+All the following steps assume that:
+
+- You agree with each software license
+- InterProScan was installed using Conda in the `IPRSCAN_DIR` directory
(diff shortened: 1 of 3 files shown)


tools/interproscan/README.md
```diff
@@ -10,3 +10,71 @@ The standard install of InterProScan does not include the following components,
 - SMART data (THRESHOLD file)
 
 As a Galaxy instance admin, you can install those manually if needed, as long as you (and the Galaxy users) respect the respective licenses.
+
+## Installing licensed components manually
```
bernt-matthias, line 14: Don't know if I really like this. 

- Some of the linked software seems to be GPL / LGPL .. might be distributed via bioconda?
- Additional "modules" (?) are nontransparent to the user
- License agreement needs to be done by each user for each module?
  - Can single modules be disabled?

I always advocate that scientific software must be free: non-free software can not be called reproducible (since not everyone can use it) and therefore can not be used in reproducible research .. which any research should be.

abretaud: Sure, I also hate non-free software :D
These instructions I have added are for the last bits of interproscan that are non-free and can't be installed from bioconda. It's up to the server admin to install them or not, and up to the user to use them or not (they can select each one individually https://github.com/galaxyproject/tools-iuc/blob/master/tools/interproscan/interproscan.xml#L81)
I think that's the best we could do with this restricted stuff..

bgruening (approved):

Issue #4299, opened 2021-12-20
Clair3 variant caller


bgruening:
That seems to be an interesting ONT variant caller: https://github.com/nanoporetech/Clair3

mira-miracoli:
I would like to try it :)

pvanheus:
I made a very small start on this:

```xml
<tool id="clair3" name="Clair3" version="@TOOL_VERSION@+galaxy0" profile="21.09">
    <macros>
        <token name="@TOOL_VERSION@">0.1.10</token>
    </macros>
    <requirements>
        <requirement type="package" version="@TOOL_VERSION@">clair3</requirement>
    </requirements>
    <command detect_errors="exit_code"><![CDATA[
        #if $ref_source.source == "history"
            ln -s '$ref_source.ref_fasta' reference.fasta &&
        #elif $ref_source.source == "builtin"
            ln -s '$ref_source.ref_fasta_builtin.path' reference.fasta &&
        #end if
        BIN_PATH=\$(dirname \$(which run_clair3.sh))
        #import os
        #if os.getenv('BIN_DIR') is not None
            #set $model_prefix = os.environ.get('BIN_PATH') + '/models/'
        #end if
        
        run_clair3.sh
            --bam_fn='$bam_input'
            --ref_fn=reference.fasta
            --model_path='$model_prefix/$model'
            --plaform='ont'
            --output='clair3_output'
            --threads=\${GALAXY_SLOTS:-1}

    ]]></command>
    <inputs>
        <conditional name="ref_source">
            <param type="select" label="Reference genome source" name="source">
                <option value="history" selected="true">History</option>
                <option value="builtin">Built-in</option>
            </param>
            <when value="history">
                <param type="data" format="fasta" name="ref_fasta" label="Reference genome" />
            </when>
            <when value="builtin">
                <param type="select" name="ref_fasta_builtin" label="Reference genome">
                    <options from_data_table="all_fasta" />
                </param>
            </when>
        </conditional>        
    </inputs>
    <outputs>
    </outputs>
    <help><![CDATA[
        TODO: Fill in help.
    ]]></help>
```

gallardoalba:
Closed in https://github.com/galaxyproject/tools-iuc/pull/4392

PR #4392, opened 2022-02-08
New tool: Clair3

mira-miracoli:
FOR CONTRIBUTOR:
* [x] - License permits unrestricted use (educational + commercial) ?


tools/clair3/clair3.xml
```diff
@@ -0,0 +1,179 @@
...
+            ln -s '${ref_source.ref_fasta_builtin.fields.path}' reference.fasta &&
+        #end if
+        samtools faidx 'reference.fasta' &&
+        BIN_PATH=\$(dirname \$(which run_clair3.sh))
+        #import os
+        #if os.getenv('BIN_DIR') is not None
+            #set $model_prefix = os.environ.get('BIN_PATH') + '/models/'
+        #end if
+
+        ln -s '${$bam_input}' 'file.bam' &&
```
pvanheus, line 21: You don't need to index the BAM - Galaxy is kind enough to do that for you and you can get the path to the index from `bam_input.metadata.bam_index` if you want it, see this [example](https://github.com/galaxyproject/tools-iuc/blob/master/tools/jvarkit/jvarkit_wgscoverageplotter.xml#L27).

tools/clair3/clair3.xml
```diff
@@ -0,0 +1,179 @@
...
+        samtools index 'file.bam' &&
+
+        run_clair3.sh
+            --bam_fn='file.bam'
+            --ref_fn='reference.fasta'
+            --model_path='$model.fields.path'
```
pvanheus, line 27: the Claire3 conda package comes with some models built in (they're in the `models/` directory that is a subdirectory of the `bin/` directory where the `run_clair3.sh` tool is located). Since many people will probably want to use the built-in models, perhaps offer them as an option so that people don't need to populate a data table to use this tool? I realise that that creates another kind of problem because you need to keep the tool up-to-date with a list of models based on whatever is "shipped" with the Clair3 tool.

pvanheus:
Thanks for tackling this! Can I suggest including the `--haploid_precise` and `--haploid_sensitive` and `--no_phasing_for_fa` options? These are useful for variant calling on SARS-CoV-2, which is what I'd like to use Clair3 for.

tools/clair3/clair3.xml
```diff
@@ -0,0 +1,179 @@
...
+                <param type="data" format="fasta" name="ref_fasta" label="Reference genome" />
+            </when>
+            <when value="builtin">
+                <param type="select" name="ref_fasta_builtin" label="Reference genome">
+                    <options from_data_table="all_fasta" />
+                </param>
+            </when>
+        </conditional>
+        <param argument="--model_path" type="select" name="model" label="Select model" >
+            <options from_data_table="model">
```
pvanheus, line 74: Perhaps use a more descriptive name for the data table - there might be multiple things in Galaxy uses models so something like `clair3_models` might be more suitable?

tools/clair3/clair3.xml
mira-miracoli: Thank you for your review. I tried with your [start](https://github.com/galaxyproject/tools-iuc/issues/4299), but the os.getenv('BIN_PATH') seems to to stay 'None'. I would be very greatful for any suggestions, since I found no helpful results for os.getenv or os.environ.get in the tools folder.

tools/clair3/clair3.xml
```diff
@@ -0,0 +1,179 @@
...
+            <param type="boolean" name="include_all_ctgs" truevalue="--include_all_ctgs" falsevalue="" label="Call variants on all contigs, otherwise call in chr{1..22,X,Y} and {1..22,X,Y}, default: disable." />
+            <param type="boolean" name="gvcf" truevalue="--gvcf" falsevalue="" label="Enable GVCF output, default: disable" />
+            <param type="boolean" name="enable_phasing" truevalue="--enable_phasing" label="Output phased variants using whatshap, default: disable" />
+            <param type="float" name="snp_min_af" value="0.08" min="0.001" max="0.08" label="Minimum SNP AF required for a candidate variant. Lowering the value might increase a bit of sensitivity in trade of speed and accuracy, default: ont:0.08,hifi:0.08,ilmn:0.08." />
+            <param type="float" name="indel_min_af" value="0.15" min="0.001" max="0.15" label="Minimum INDEL AF required for a candidate variant. Lowering the value might increase a bit of sensitivity in trade of speed and accuracy, default: ont:0.15,hifi:0.08,ilmn:0.08." />
```
wm75, line 125: Why such strange min/max bounds on these params. AF can range from 0 to 1, and normally it should be left to the user to configure anything in that range.

tools/clair3/clair3.xml
```diff
@@ -0,0 +1,179 @@
...
+                    <param type="boolean" name="print_ref_calls" truevalue="--print_ref_calls" falsevalue="" label="Show reference calls (0/0) in vcf file, default: disable." />
+                </when>
+            </conditional>
+            <param type="integer" name="qual" value="0" min="0" max="255" label="If set, variants with >qual will be marked PASS, or LowQual otherwise." />
+            <param type="integer" name="chunk_size" value="5" min="1" max="255" label="The size of each chuck for parallel processing, default: 5Mbp." optional="true" />
+            <param type="boolean" name="include_all_ctgs" truevalue="--include_all_ctgs" falsevalue="" label="Call variants on all contigs, otherwise call in chr{1..22,X,Y} and {1..22,X,Y}, default: disable." />
+            <param type="boolean" name="gvcf" truevalue="--gvcf" falsevalue="" label="Enable GVCF output, default: disable" />
+            <param type="boolean" name="enable_phasing" truevalue="--enable_phasing" label="Output phased variants using whatshap, default: disable" />
```
wm75, line 123: falsevalue missing here

tools/clair3/clair3.xml
```diff
@@ -0,0 +1,179 @@
...
+                    <param type="data" name="bed_fn" format="bed" label="Call variants only in the provided bed regions." optional="true" />
+                </when>
+                <when value="vcf">
+                    <param type="data" name="vcf_fn" format="vcf" label="Candidate sites VCF file input, variants will only be called at the sites in the VCF file if provided." optional="true" />
+                    <param type="boolean" name="print_ref_calls" truevalue="--print_ref_calls" falsevalue="" label="Show reference calls (0/0) in vcf file, default: disable." />
+                </when>
+            </conditional>
+            <param type="integer" name="qual" value="0" min="0" max="255" label="If set, variants with >qual will be marked PASS, or LowQual otherwise." />
+            <param type="integer" name="chunk_size" value="5" min="1" max="255" label="The size of each chuck for parallel processing, default: 5Mbp." optional="true" />
+            <param type="boolean" name="include_all_ctgs" truevalue="--include_all_ctgs" falsevalue="" label="Call variants on all contigs, otherwise call in chr{1..22,X,Y} and {1..22,X,Y}, default: disable." />
```
wm75, line 121: My preference would be to have this param `checked="true"`, i.e. to invert the default. As restricting to canonical human chromosomes really only makes sense if you're analyzing human data.

tools/clair3/clair3.xml
```diff
@@ -0,0 +1,179 @@
...
+                </param>
+                <when value="bed">
+                    <param type="data" name="bed_fn" format="bed" label="Call variants only in the provided bed regions." optional="true" />
+                </when>
+                <when value="vcf">
+                    <param type="data" name="vcf_fn" format="vcf" label="Candidate sites VCF file input, variants will only be called at the sites in the VCF file if provided." optional="true" />
+                    <param type="boolean" name="print_ref_calls" truevalue="--print_ref_calls" falsevalue="" label="Show reference calls (0/0) in vcf file, default: disable." />
+                </when>
+            </conditional>
+            <param type="integer" name="qual" value="0" min="0" max="255" label="If set, variants with >qual will be marked PASS, or LowQual otherwise." />
```
wm75, line 119: Variant QUAL scores are not capped at 255 so this is a case where you shouldn't specify a max value.

tools/clair3/clair3.xml
```diff
@@ -0,0 +1,179 @@
...
+        </param>
+        <param type="data" name="bam_input" format="bam" label="BAM file input. The input file must be samtools indexed." />
+        <section name="adv" title="Advanced Parameters" expanded="false">
+            <conditional name="bed_or_vcf">
+                <param type="select" name="bed_or_vcf_selector" label="Choose if you want to provide an additional vcf or bed-file.">
+                    <option value="bed" selected="true">Provide bed-file</option>
+                    <option value="vcf">Provide vcf-file</option>
+                </param>
+                <when value="bed">
+                    <param type="data" name="bed_fn" format="bed" label="Call variants only in the provided bed regions." optional="true" />
```
wm75, line 112: does `bed_fn` not support `print_ref_calls`?

tools/clair3/clair3.xml
```diff
@@ -0,0 +1,179 @@
...
+                    <options from_data_table="all_fasta" />
+                </param>
+            </when>
+        </conditional>
+        <param argument="--model_path" type="select" name="model" label="Select model" >
+            <options from_data_table="model">
+                <validator type="no_options" message="no modles available" />
+            </options>
+        </param>
+        <param type="data" name="bam_input" format="bam" label="BAM file input. The input file must be samtools indexed." />
```
wm75, line 104: Remove mention of samtools indexing as this is not relevant in Galaxy.

tools/clair3/clair3.xml
```diff
@@ -0,0 +1,179 @@
...
+            <param type="boolean" name="gvcf" truevalue="--gvcf" falsevalue="" label="Enable GVCF output, default: disable" />
+            <param type="boolean" name="enable_phasing" truevalue="--enable_phasing" label="Output phased variants using whatshap, default: disable" />
+            <param type="float" name="snp_min_af" value="0.08" min="0.001" max="0.08" label="Minimum SNP AF required for a candidate variant. Lowering the value might increase a bit of sensitivity in trade of speed and accuracy, default: ont:0.08,hifi:0.08,ilmn:0.08." />
+            <param type="float" name="indel_min_af" value="0.15" min="0.001" max="0.15" label="Minimum INDEL AF required for a candidate variant. Lowering the value might increase a bit of sensitivity in trade of speed and accuracy, default: ont:0.15,hifi:0.08,ilmn:0.08." />
+        </section>
+    </inputs>
+    <outputs>
+        <data name="merge_output" format="vcf_bgzip" />
+        <data name="pileup" format="vcf_bgzip" />
+        <data name="full_alignment" format="vcf_bgzip" />
```
wm75, line 134: Not sure what outputs Clair3 produces (can you explain?), but if the names mean anything, then they cannot all be of format "vcf_bgzip"

tools/clair3/clair3.xml
mira-miracoli: You're right, I thought it caused an error, but it doesn't. Fixed it.

tools/clair3/clair3.xml
mira-miracoli: Didn't know that - the description said "Lowering the value might increase a bit of sensitivity in trade of speed and accuracy". So I assumed only a small decrease of the value would do it.

tools/clair3/clair3.xml
mira-miracoli: I'm not sure about the output files. I don't understand what they mean. Would need help on this.

wm75: The github page also says this:
`--remove_intermediate_dir Remove intermediate directory, including intermediate phased BAM, pileup and full-alignment results. default: disable.`
implying that there is at least one other intermediate output (a "phased" BAM file). Not exactly sure what that would look like, but it can't hurt to offer it as an optional output, too.

mira-miracoli:
I'm on vacation until Monday 28th Feb, you can finish the tool without me if you want.

tools/clair3/clair3.xml
mira-miracoli: Thanks for the explanation :)

tools/clair3/clair3.xml
wm75: I'm still not entirely sure how/if `--bed_fn` and `--vcf_fn` interact and `--print_ref_calls` interact.
It could be that  `--print_ref_calls` only works in combination with `--vcf_fn` and is silently ignored without a VCF file.
Currently you're using it outside the whole conditional and if that is accepted and doesn't cause *every* position to be reported, then it might just be ignored. Please check the different possibilities and either move `print_ref_calls` into both branches inside the conditional or back into the "vcf" branch only.

Different issue: the conditional as such is a bit suboptimal. The user could also choose not to provide *any* additional file and that's currently covered with optional="true" in both branches, but that feels strange.
Would be cleaner to have the outer param say: "Restrict variant calling to" and show the options:
- Whole reference (unrestricted)
- Regions defined in BED dataset
- Sites defined in VCF dataset

The extra files would only be asked for in branches 2 and 3, and would not have to be optional then.
I would then move that whole conditional out of advanced, show it right underneath the `bam_input` selection, and move `include_all_ctgs` into the "Whole reference" branch.

tools/clair3/clair3.xml
```diff
@@ -0,0 +1,291 @@
...
+            <param type="boolean" name="haploid_precise" truevalue="--haploid_precise" falsevalue="" label="EXPERIMENTAL: Enable haploid calling mode. Only 1/1 is considered as a variant, default: disable." />
+            <param type="boolean" name="haploid_sensitive" truevalue="--haploid_sensitive" falsevalue="" label="EXPERIMENTAL: Enable haploid calling mode. 0/1 and 1/1 are considered as a variant, default: disable." />
+            <param type="boolean" name="no_phasing_for_fa" truevalue="--no_phasing_for_fa" falsevalue="" label="EXPERIMENTAL: Call variants without whatshap phasing in full alignment calling, default: disable." />
+            <param type="boolean" name="print_ref_calls" truevalue="--print_ref_calls" falsevalue="" label="Show reference calls (0/0) in vcf file, default: disable." />
+            <param type="boolean" name="no_phasing_for_fa" truevalue="--no_phasing_for_fa" falsevalue="" label="EXPERIMENTAL: Call variants without whatshap phasing in full alignment calling, default: disable." />
```
wm75, line 142: Duplicated option (see above). Group with `--enable_phasing`.

tools/clair3/clair3.xml
```diff
@@ -0,0 +1,291 @@
...
+                <when value="bed">
+                    <param type="data" name="bed_fn" format="bed" label="Call variants only in the provided bed regions." optional="true" />
+                </when>
+                <when value="vcf">
+                    <param type="data" name="vcf_fn" format="vcf" label="Candidate sites VCF file input, variants will only be called at the sites in the VCF file if provided." optional="true" />
+                </when>
+            </conditional>
+            <param type="integer" name="qual" value="0" min="0" label="If set, variants with >qual will be marked PASS, or LowQual otherwise." />
+            <param type="integer" name="chunk_size" value="5000000" min="1" label="The size of each chuck for parallel processing, default: 5Mbp." optional="true" />
+            <param type="boolean" name="gvcf" truevalue="--gvcf" falsevalue="" label="Enable GVCF output, default: disable" />
```
wm75, line 134: Move this into `output_options`.

tools/clair3/clair3.xml
```diff
@@ -0,0 +1,291 @@
...
+                </param>
+                <when value="bed">
+                    <param type="data" name="bed_fn" format="bed" label="Call variants only in the provided bed regions." optional="true" />
+                </when>
+                <when value="vcf">
+                    <param type="data" name="vcf_fn" format="vcf" label="Candidate sites VCF file input, variants will only be called at the sites in the VCF file if provided." optional="true" />
+                </when>
+            </conditional>
+            <param type="integer" name="qual" value="0" min="0" label="If set, variants with >qual will be marked PASS, or LowQual otherwise." />
+            <param type="integer" name="chunk_size" value="5000000" min="1" label="The size of each chuck for parallel processing, default: 5Mbp." optional="true" />
```
wm75, line 133: Move to the end because it's a very technical param that few users would ever want to think about.

tools/clair3/clair3.xml
```diff
@@ -0,0 +1,291 @@
...
+            <param type="boolean" name="gvcf" truevalue="--gvcf" falsevalue="" label="Enable GVCF output, default: disable" />
+            <param type="boolean" name="enable_phasing" truevalue="--enable_phasing" falsevalue="" label="Output phased variants using whatshap, default: disable" />
+            <param type="float" name="snp_min_af" value="0.08" min="0" max="1" label="Minimum SNP AF required for a candidate variant. Lowering the value might increase a bit of sensitivity in trade of speed and accuracy, default: ont:0.08,hifi:0.08,ilmn:0.08." />
+            <param type="float" name="indel_min_af" value="0.15" min="0" max="1" label="Minimum INDEL AF required for a candidate variant. Lowering the value might increase a bit of sensitivity in trade of speed and accuracy, default: ont:0.15,hifi:0.08,ilmn:0.08." />
```
wm75, line 137: move up to underneath `qual`

tools/clair3/clair3.xml
```diff
@@ -0,0 +1,291 @@
...
+            <param type="boolean" name="haploid_precise" truevalue="--haploid_precise" falsevalue="" label="EXPERIMENTAL: Enable haploid calling mode. Only 1/1 is considered as a variant, default: disable." />
+            <param type="boolean" name="haploid_sensitive" truevalue="--haploid_sensitive" falsevalue="" label="EXPERIMENTAL: Enable haploid calling mode. 0/1 and 1/1 are considered as a variant, default: disable." />
```
wm75, line 139: You cannot reasonably combine these two. This is more elegantly combined into a select labeled "Call with the following ploidy model" and with three options:
- diploid (with selected="true" and an empty string option value)
- haploid (sensitive)
- haploid (precise)

Move the current labels of the boolean params into the select's help.

.gitignore
```diff
@@ -72,3 +72,6 @@ tool_test_output.json
 
 # VScode configuration
 .vscode
+tools/clair3/test-data/run_clair3.log
+tools/clair3/test-data/test1.bam.bai
+tools/clair3/test-data/test1.fasta.fai
```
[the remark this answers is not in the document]
mira-miracoli: I'm sorry I forgot to do that.

tools/clair3/clair3.xml
mira-miracoli: Sorry should I move it to `output_options` or to the end of `adv` and do you mean only `chunk_size`?

tools/clair3/clair3.xml
```diff
@@ -0,0 +1,309 @@
...
+        #elif $ref_source.source == "builtin"
+            ln -s '${ref_source.ref_fasta_builtin.fields.path}' reference.fasta &&
+        #end if
+        samtools faidx reference.fasta &&
+
+        #if $model_source.source == "datatable"
+            #set model_name = $model_source.model.fields.path
+        #else
+            #set model_name = $model_source.select_built_in
+        #end if
```
wm75, line 21: Just noticed that this won't work with a real datatable: the DM would *not* install the model into the conda environment, but in a Galaxy-internal folder.
In other words, in the "datatable" case you should trust the .fields.path thing to be a complete path to the model dir, while in the built_in case it's only a simple folder name.
=> you need to invoke your \$(dirname \$(which run_clair3.sh))/models magic only in the built_in branch.

For the test datatable record (and only there), you'd have to experiment with prefixing the model name with \$(dirname \$(which run_clair3.sh))/models/ verbatim.

wm75: and regarding the datatable, I think it would also be prudent to add a platform column there, then bypass your platform autodetection logic for models from the datatable cause who knows what gets installed there under what folder names.

tools/clair3/clair3.xml
```diff
@@ -0,0 +1,310 @@
...
+            #if $model_source.source == "datatable"
+                --platform='${model_source.model.fields.platform}'
+                --model_path=${model_source.model.fields.path}
+            #else
+                #if $model_source.select_built_in == "hifi" or $model_source.select_built_in == "ilmn":
+                    --platform='${$model_source.select_built_in}'
+                #else
+                    --platform='ont'
+                #end if
+                --model_path=\$(dirname \$(which run_clair3.sh))/models/$model_source.select_built_in
```
gallardoalba, line 42:
```suggestion
                --model_path='\$(dirname \$(which run_clair3.sh))/models/$model_source.select_built_in'
```

tools/clair3/.shed.yml
```diff
@@ -0,0 +1,15 @@
+
```
gallardoalba, line 1:
```suggestion
```

tools/clair3/clair3.xml
```diff
@@ -0,0 +1,310 @@
+<tool id="clair3" name="Clair3" version="@TOOL_VERSION@+galaxy0" profile="21.09">
+    <macros>
+        <token name="@TOOL_VERSION@">0.1.11</token>
```
gallardoalba, line 3:
```suggestion
<tool id="clair3" name="Clair3" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="21.09">
    <macros>
        <token name="@TOOL_VERSION@">0.1.11</token>
        <token name="@VERSION_SUFFIX@">0</token>
```

gallardoalba (approved):

tools/clair3/clair3.xml
```diff
@@ -0,0 +1,311 @@
...
+            #if $model_source.source == "datatable"
+                --platform='${model_source.model.fields.platform}'
+                --model_path=${model_source.model.fields.path}
+            #else
+                #if $model_source.select_built_in == "hifi" or $model_source.select_built_in == "ilmn":
+                    --platform='${$model_source.select_built_in}'
+                #else
+                    --platform='ont'
+                #end if
+                --model_path='\$(dirname \$(which run_clair3.sh))/models/$model_source.select_built_in'
```
gallardoalba, line 43:
```suggestion
                --model_path=\$(dirname \$(which run_clair3.sh))/models/$model_source.select_built_in
```

tools/clair3/clair3.xml
```diff
@@ -0,0 +1,291 @@
...
+            #set model_name = $model_source.select_built_in
+        #end if
+
+        ln -s '${$bam_input}' 'file.bam' &&
+        ln -s '${$bam_input.metadata.bam_index}' 'file.bai' &&
+
+        run_clair3.sh
+            --bam_fn='file.bam'
+            --ref_fn='reference.fasta'
+            --model_path=\$(dirname \$(which run_clair3.sh))/models/$model_name
```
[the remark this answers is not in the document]
gallardoalba: I included here the quotes, but it causes the tests to fail, so not sure how to quote it.

mira-miracoli:
@wm75 could you review/approve it so we can merge, please? :)

bgruening:
We will trust @gallardoalba here and get it in. Wolfgang is on vacation this week.

Thanks @mira-miracoli

PR #4401, opened 2022-02-11
Add the binning_refiner tool

gregvonkuster:
FOR CONTRIBUTOR:


bgruening (commented):
There are two binning refiner tools in your PR one in the test data folder I believe

gregvonkuster:
> There are two binning refiner tools in your PR one in the test data folder I believe

Thanks for helping @bgruening - so appreciated!

tools/binning_refiner/binning_refiner.xml
```diff
@@ -0,0 +1,175 @@
...
+#end for
+
+Binning_refiner
+-i input_bin_dir
+-p 'refined'
+&& mv 'refined_Binning_refiner_outputs/refined_contigs.txt' '$output_refined_contigs'
+&& mv 'refined_Binning_refiner_outputs/refined_sources_and_length.txt' '$output_sources_and_length'
+    ]]></command>
+    <inputs>
+        <repeat name="binned_fasta_files" title="Binned fasta files" min="1">
```
bgruening, line 38: Why does this needs to be a repeat and a collection?

`type="data" multiple="true"` would also work, does it? If you want to feed in multiple collections, the logic should be part of the workflow, not the tool. E.g. a user could use the merge collection tool to convert X collections into one before feeding into this tool

gregvonkuster: > Why does this needs to be a repeat and a collection?
> 
> `type="data" multiple="true"` would also work, does it? If you want to feed in multiple collections, the logic should be part of the workflow, not the tool. E.g. a user could use the merge collection tool to convert X collections into one before feeding into this tool

I've changed the Galaxy wrapper to perform in this way.  Notice that the test now has binned outputs from the MetaBAT tool and the CONCOCT tool merged into the same single set of inputs.  A possible concern resulting from this change is the difference in the test outputs when the same inputs (but merged into 1 collection) are used.

bgruening (approved):
I'm only asking questions Greg :)
If you think the other approach is better for this tool, go for it. I have no clue about the tool. Just try to keep workflow logic out of the tool.

gregvonkuster:
> I'm only asking questions Greg :)
> If you think the other approach is better for this tool, go for it. I have no clue about the tool. Just try to keep workflow logic out of the tool.

In this case, I think the wrapper should model the underlying tool.  I'm not an expert on binning, but I think the various binning tools bin contigs in various ways.  So I think this tool was written to expect each input subdirectory would contain a set of binned fasta files that were produced by a single tool.

I find it strange that fewer refined bins are output when a 2nd input subdirectory is added (see test 2 of the original version), but I expect this must be correct.

I know it is more optimal to not allow for input collections with tools when it comes to Galaxy's ability to split jobs across cores, but doing that is not possible with every tool, and I think this tool should be allowed to have input collections since that is what is produced by the upstream binning tools.  

We could eliminate the repeat blocks, but then we'd be restricted to a specific number of input collections.

Would you like me to revert back to the original?

bgruening:
What about a repeat block, but multiple=true data inputs?

gregvonkuster:
> What about a repeat block, but multiple=true data inputs?

If you're thinking about job splitting, I think the input binned fasta files may need to be all handled together to get properly filtered - this line in the refiner https://github.com/songweizhi/Binning_refiner/blob/master/bin/Binning_refiner#L158 may filter inappropriately when the inputs are split up (although I'm not absolutely sure).

Is there another reason you want a multi-select instead of a collection on input?

gregvonkuster:
I've reverted the tool for now because I think merging the inputs from various binning tools produced incorrect results.  I'll make more changes if you want though - just let me know. :)

bernt-matthias (commented):
Looks good to me, but I would prefer to have fewer input test datasets if possible.

gregvonkuster:
> Looks good to me, but I would prefer to have fewer input test datasets if possible.

It's tricky to find the correct inputs that will produce at least 1 refined output.  The 2nd test is the reason there are this many inputs - outputs from multiple binning tools seem to get filtered with smaller inputs (I've used the smallest inputs I could find).

bernt-matthias (approved):
Then its great as it is. Lets merge?

gregvonkuster:
> Then its great as it is. Lets merge?

Sure, thanks!  :)

PR #4402, opened 2022-02-14
Updating tools/bwameth from version 0.2.3 to 0.2.4

planemo-autoupdate:

I have updated tools/bwameth from version 0.2.3 to 0.2.4.

### DIFF
--- a/tools/bwameth/bwameth.xml
+++ b/tools/bwameth/bwameth.xml
@@ -1,7 +1,7 @@
 <tool id="bwameth" name="bwameth" version="@TOOL_VERSION@+galaxy0" profile="20.05">
     <description>Fast and accurate aligner of BS-Seq reads.</description>
     <macros>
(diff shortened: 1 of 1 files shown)


bgruening:
Is there a way to skip the release? This one only contains fixes for bwa2 that we do not support.

bernt-matthias:
I would suggest that we simply close PRs which we want to skip. Or will the autoupdater open a new PR?

simonbray:
It will reopen them, as for BiocondaBot, which I followed quite closely for the implementation.

Bioconda are dealing with this situation using the `do-not-close` tag: https://github.com/bioconda/bioconda-recipes/pulls?q=is%3Aopen+is%3Apr+label%3Ado-not-close

Maybe that could be an option here as well?

simonbray:
Hm, maybe

planemo-autoupdate:
let's test

PR #4419, opened 2022-02-22
add labels fw and rv when use the same filenames

PlushZ:


bgruening (approved):
I think this is fine. I guess I would prefer `rv_` and `fw_` but that is not a strong feeling.

@gallardoalba any preferences?

bernt-matthias:
Added a commit examplifying what should be done in my opinion: 

- output a paired collection with the corrected reads (can this actually be multiple .. would like to see a test with multiple inputs) and one for the uncorrected
- also the fixes for
  -  input formats: having fastq and fastqsanger is redundant and fastq should be removed
  - add missing operation mode to all tools (or remove from tests)

bernt-matthias:
> somehow the collection misses the reverse file 

got it. the tool symlinks the inputs using element identifiers (for paired in separate datasets) and therefore also the outputs will have unique identifiers .. i.e. forward and reverse reads end up in separate collections,

solution: do not use element identifiers for data inputs with `multiple="true"`  .. 

funnily the tool symlinks to indexes for the list:paired collection where identifiers actually would make sense

bernt-matthias:
Added a few more comments to the code that need fixing. 

Biggest question so far is how to treat the `library_number` parameter(s). If I get it right use can select if the `additional_reads` come from a different library. 

But its unclear to me why there is another unused `library_number` parameter in `input_files_paired` that is apparently unused. 

Maybe we can just use a repeat for this, with one fastq input macro for each library. Then we could also have more than 2 libraries. If users have data from the same library in different collections they need to restructure them with collection operations. We could still allow to consider each input element (e.g. if its in a collection) as coming from a different library with an additional switch.

PR #4420, opened 2022-02-23
Srst2

EngyNasr:
FOR CONTRIBUTOR:
* [x] - This PR adds a new tool ([SRST2](https://github.com/katholt/srst2#srst2))
* [x] Next step: is to reduce the size of test data

Thank youuu :)


tools/srst2/macros.xml
```diff
@@ -0,0 +1,21 @@
...
+        <edam_topics>
+            <edam_topic>topic_0114</edam_topic>
+            <edam_topic>topic_1311</edam_topic>
+            <edam_topic>topic_0783</edam_topic>
+            <edam_topic>topic_3697</edam_topic>
+        </edam_topics>
+    </xml>
+    <xml name="requirements">
+        <requirements>
+            <requirement type="package" version="0.1.18">samtools</requirement>
```
bgruening, line 16: Do you need this very very old version of samtools?

tools/srst2/srst2.xml
```diff
@@ -0,0 +1,423 @@
...
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="edam_ontology"/>
+    <expand macro="requirements"/>
+    <version_command>srst2 --version</version_command>
+    <command detect_errors="exit_code"><![CDATA[
+#if $input.selector == "single"
+#set ext=$input.single_input.datatype.file_ext
+    ln -s $input.single_input ./input_read1.$ext &&
```
bgruening, line 12: please single-quote all data paramerters

tools/srst2/srst2.xml
```diff
@@ -0,0 +1,423 @@
...
+#else if $input.selector == "paired"
+#set ext_1=$input.paired_input1.datatype.file_ext
+#set ext_2=$input.paired_input2.datatype.file_ext
+    ln -s $input.paired_input1 ./input_read1.$ext_1 &&
+    ln -s $input.paired_input2 ./input_read2.$ext_2 &&
+#end if
+#for $i, $s in enumerate($prev_output)
+    #if $s
+        input_count=$i
```
bgruening, line 21: what is input count? its not used anymore, is this option tested? I guess it should crash.

tools/srst2/srst2.xml
```diff
@@ -0,0 +1,423 @@
...
+    <inputs>
+        <conditional name="input">
+            <param name="selector" type="select" optional="true" label="Reads files type for anaylsis">
+                <option value="single">Single-end</option>
+                <option value="paired">Paired-end</option>
+                <option value="only_compiling_previous_results">Only Compiling Previous SRST2 Results</option>
+            </param>
+            <when value="single">
+                <param name="single_input" type="data" format="fasta,fastq,fastqsanger" label="Single end read file(s) for analysing (may be gzipped)"/>
+                <param argument="--read_type" type="select" label="Read file type">
```
bgruening, line 123: This is used twice, should this be a macro?

tools/srst2/srst2.xml
```diff
@@ -0,0 +1,423 @@
...
+                    <option value="q" selected="true">fastq</option>
+                    <option value="qseq">solexa</option>
```
bgruening, line 125: uha, SOLEXA is super old, this option is probably not needed

tools/srst2/srst2.xml
```diff
@@ -0,0 +1,423 @@
...
+                    <option value="f">fasta</option>
+                </param>     
+            </when>
+            <when value="paired">
+                <param name="paired_input1" type="data" format="fasta,fastq,fastqsanger" label="Paired end read files for analysing (may be gzipped)"/>
```
bgruening, line 130: if it can be gzipped you can add the datatpyes fasta.gz, fastq.gz, and fasqsanger.gz

tools/srst2/srst2.xml
```diff
@@ -0,0 +1,423 @@
...
+                <param argument="--gene_db" type="data" optional="true" multiple="true" format="fasta" label="Gene database(s)" help="Fasta file/s for gene databases"/>
+                <param argument="--no_gene_details" type="boolean" truevalue="" falsevalue="--no_gene_details" checked="false" label="Do you want reporting of gene typing?"/>
+                <param argument="--gene_max_mismatch" type="integer" value="10" label="Maximum number of mismatches per read for gene detection and allele calling"/>
+                <param argument="--min_coverage" type="integer" value="90" label="Minimum %coverage cutoff for gene reporting"/>
+                <param argument="--max_divergence" type="integer" value="10" label="Maximum %divergence cutoff for gene reporting"/>            
+            </when>
+            <when value="no">
+            </when>
+        </conditional>
+        <param argument="--prob_err" type="float" value="0.01" label="Probability of sequencing error"/>
```
bgruening, line 186: min max

tools/srst2/srst2.xml
```diff
@@ -0,0 +1,423 @@
...
+                    <add value="\" />
+                    <add value="-" />
+                </valid>
+            </sanitizer>
+            <validator type="regex">[A-Za-z0-9 =-_/+]+</validator>
+        </param>        
+        <param argument="--max_unaligned_overlap" type="integer" value="10" label="Read discarded from alignment" help="if either of its ends has unaligned overlap with the reference that is longer than this value"/>
+        <param argument="--mapq" type="integer" value="1" label="Samtools -q parameter (Minimum mapping quality)"/>
+        <param argument="--baseq" type="integer" value="20" label="Samtools -Q parameter (Minimum base quality)"/>
+        <param argument="--samtools_args" type="text" optional="true" label="Other arguments to pass to samtools mpileup" help="E.g. \-A">
```
bgruening, line 201: is this parameter really needed? @wm75 safe enough?

tools/srst2/srst2.xml
```diff
@@ -0,0 +1,423 @@
...
+            <sanitizer invalid_char="">
+                <valid initial="string.letters,string.digits">
+                    <add value="\" />
+                    <add value="-" />
+                </valid>
+            </sanitizer>
+            <validator type="regex">[A-Za-z0-9 =-_/+]+</validator>
+        </param>
+        <param argument="--log" type="boolean" truevalue="--log" falsevalue="" checked="false" label="Do you want to save the log?"/>
```
bgruening, line 210: this could be a select field with multiple=true

bgruening (commented):
Small review included. Your test data is also too big it seems.

tools/srst2/macros.xml
EngyNasr: unfortunately yes: https://github.com/katholt/srst2#current-release---v020---july-28-2016

tools/srst2/srst2.xml
EngyNasr: They (prev_output and gene_db) are multiple input, that I wanted to rename since they appear in the outputs' names, as well as when they are given without this for loop, it is done with a ',' separating them, which the tool doesnot accept.

By doing this for loop it is tested and it doesnot crash, the gene_db example is tested in the test in the wrapper.

However, if it not used anymore then I will think of something else :)

tools/srst2/srst2.xml
EngyNasr: yeaah, I thought it is automatically included in the initial datatype :D, I will update it right away

tools/srst2/srst2.xml
EngyNasr: Not really its optional, but since the tool uses bowtie2 and samstool it allows the user to add more arguments but it requires the users to already know the arguments of samstool and bowtie2, which might not be always the case.

tools/srst2/srst2.xml
EngyNasr: The samstool & bowtie2 arguments you mean ?

tools/srst2/srst2.xml
EngyNasr: > Small review included. Your test data is also too big it seems.

Enormous :D :D I was able to reduce all except, the Paired reads files, still trying

I have one more question left: if I want to add an if condition, where I copy the output files into folders. Such that I check if there is no file with this name, it by pass the if condition and dont give an error. I tried it by different ways but I failed, can you please help me in this as well <3 ?

tools/srst2/macros.xml
bernt-matthias: Should be in the biopython requirements, or?

tools/srst2/srst2.xml
bgruening: Why is this not working?

```
#for $i, $s in enumerate($prev_output)
      ln -s $s './$i-prev_output.txt' &&
#end for
```

tools/srst2/srst2.xml
bgruening: No, all those files. You can convert all those output-file-booleans into one single multiple="true" select option. That makes your tool form simpler imho

tools/srst2/srst2.xml
bgruening: Those "pass any parameters to tool X" options are not secure. I guess we should not offer them in the tool form. Or if they are important model them as real parameters. E.g. real floats and booleans.

wm75: Whether this is safe or not depends also heavily on what the tool does internally with the string (from the docs it seems at least as if you'd pass them as *one* argument to srst2 meaning you could quote the whole thing in the command section.

wm75: The docs also list:
```
--use_existing_bowtie2_sam
                        Use existing SAM file generated by Bowtie2 if
                        available, otherwise they will be generated
  --use_existing_pileup
                        Use existing pileups if available, otherwise they will
                        be generated
```

so it might be best to split at least bowtie2 out of this wrapper and force the user to run it separately (why do we have workflows after all?).
This way, we can offer our existing super-well tested bowtie2 wrapper, which is much more user-friendly than anything you can incorporate here.

wm75: The existing pileup option might not be so easy unfortunately since it would have to be generated with a compatible samtools version and our existing samtools pileup wrapper is not in particular good shape either :(

bgruening: This tool also seems to need samtools >1.0

tools/srst2/srst2.xml
EngyNasr: That is a wrong thing sorrryyy, just got it now :D :D

bgruening:
@EngyNasr this is still WIP isn't it?

EngyNasr:
@bgruening yes please I still need to check its importance before we merge it :)

EngyNasr:
Dears, 

I appreciate a review from tool-iuc  :)

Thanks a lot,
Engy

tools/srst2/srst2.xml
```diff
@@ -83,28 +82,26 @@ srst2
         #end if
     #end for  
 #end if    
-    $log
-    $save_scores
-    $report_new_consensus
-    $report_all_consensus
-    $keep_interim_alignment
-#if $report_new_consensus and $use_gene_db.selector == "yes" and $use_mlst_db.selector == "yes"
+    --log
```
bgruening, line 85: I don't see how this can work? You are writing here all options, no matter what a user selects.

tools/srst2/macros.xml
```diff
@@ -4,12 +4,6 @@
     <token name="@VERSION_SUFFIX@">0</token>
     <token name="@PROFILE@">20.01</token>
     <xml name="edam_ontology">
```
bgruening, line 6: why not removing the xml here entirely?

tools/srst2/srst2.xml
EngyNasr: Yeah I let the tool produce everything and only show the files the user choose to show, can change that as well no problem

tools/srst2/macros.xml
EngyNasr: i did so but failed linting, maybe I can search the ontology for a related edam topic

bgruening: you need to remove the macro here and the macro wherever you use it, isn't it?
Removing the EDAM onotoogies should be fine if you have a bio.tools ID

EngyNasr: thank youuu :) yeah I missed removing it from the tool's wrapper file :D now it linting successfully

bgruening (approved):

bgruening:
Thanks!

PR #4427, opened 2022-02-28
Updating tools/cutadapt from version 3.5 to 3.7

planemo-autoupdate:
Hello! This is an automated update of the following tool: **tools/cutadapt**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

I have updated tools/cutadapt from version 3.5 to 3.7.

**Project home page:** https://cutadapt.readthedocs.org/en/stable


bernt-matthias:
There seems to be only a new placeholder for the `--rename` option which I documented now: https://github.com/marcelm/cutadapt/blob/main/CHANGES.rst

bgruening (approved):

PR #4430, opened 2022-02-28
Updating tools/tracy from version 0.6.1 to
0.7.5

planemo-autoupdate:
Hello! This is an automated update of the following tool: **tools/tracy**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

I have updated tools/tracy from version 0.6.1 to 0.7.1.


simonbray:
> https://github.com/gear-genomics/tracy/compare/v0.6.1...v0.7.1

No changelog unfortunately, but the only major change seems to be a new `consensus` subcommand.

PR #4432, opened 2022-02-28
Add the dada2_phyloseq_plot_richness tool

gregvonkuster:
FOR CONTRIBUTOR:


bernt-matthias:
I would just rename the directory to `phyloseq`.

tools/dada2_phyloseq/.shed.yml
```diff
@@ -0,0 +1,18 @@
+name: dada2_phyloseq
+owner: iuc
+description: Various phyloseq plots of DADA2 outputs.
```
bernt-matthias, line 3: Can we keep this generic independent of `dada2`?

tools/dada2_phyloseq/dada2_phyloseq_plot_richness.R
```diff
@@ -0,0 +1,44 @@
...
+# Construct a tax_table object.  The rownames of
+# tax_tab must match the OTU names (taxa_names)
+# of the otu_table defined below.
+tax_tab <- tax_table(tax_table_matrix);
+
+# Construct an otu_table object.
+otu_tab <- otu_table(seq_table_numeric_matrix, taxa_are_rows = TRUE);
+
+# Construct a phyloseq object.
+phyloseq_obj <- phyloseq(otu_tab, tax_tab);
```
bernt-matthias, line 38: What do you think of creating a tool for the import to a R phyloseq import which would output `rds` that can be used as input for the plotting tool?

tools/dada2_phyloseq/.shed.yml
bernt-matthias: Here, in tool names, etc?

bernt-matthias (commented):
Really good start.

tools/dada2_phyloseq/.shed.yml
gregvonkuster: The intent of these tools very specific to dada2 - they accept only the dada2 sequence table and taxonomy table as inputs and produce plots based on them.  I don't think we should consider these as generic phyloseq plotting tools.

bernt-matthias: > The intent of these tools very specific to dada2

At the moment :) 

To me it seems easy to generalize by having one dada2 specific import tool (other import tools can be added later) + generic plotting tool.

tools/dada2_phyloseq/dada2_phyloseq_plot_richness.R
gregvonkuster: We could do that, but the weakness in that approach would be that these plotting tools would accept any ```rds``` dataset, and would not be restricted to the specific tool that provides the phyloseq object.   

I think these plotting tools are simple enough that the few lines of code that instantiate the phyloseq object may not require a separate tool.  But these few lines of code would be duplicated in the various plotting tools, so I'll create the separate tool if you feel it would be better.

tools/dada2_phyloseq/.shed.yml
gregvonkuster: So would you place the dada2 specific tool that outputs the rds dataset in the dada2 directory, and the plotting tools in the renamed phyloseq directory?

tools/dada2_phyloseq/dada2_phyloseq_plot_richness.R
bernt-matthias: > the weakness in that approach would be that these plotting tools would accept any rds

true. we could create a datatype for that (would still be possible for 22.01 I guess).

> so I'll create the separate tool if you feel it would be better.

That would be better in my opinion

gregvonkuster: What should the new datatype be?  Would rds_dada2 work?

nsoranzo: > true. we could create a datatype for that (would still be possible for 22.01 I guess).

It is always possible to add new datatypes to a frozen branch or the latest tagged release branch (i.e. 22.01 or 21.09 currently).

tools/phyloseq/.shed.yml
```diff
@@ -0,0 +1,18 @@
+name: phyloseq
+owner: iuc
+description: A suite of tools that produce various phyloseq plots.
```
bernt-matthias, line 3:
```suggestion
description: Handling and analysis of high-throughput microbiome census data
```

tools/phyloseq/.shed.yml
```diff
@@ -0,0 +1,18 @@
...
+long_description: |
+  A suite of tools that produce various phyloseq plots.
```
bernt-matthias, line 5:
```suggestion
    phyloseq provides a set of classes and tools to facilitate the import, storage, analysis, and graphical display of microbiome census data.
```

tools/phyloseq/.shed.yml
```diff
@@ -0,0 +1,18 @@
...
+categories:
+- Metagenomics
+remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/master/tools/phyloseq
+homepage_url: https://www.bioconductor.org/packages/release/bioc/html/phyloseq.html
+type: unrestricted
+auto_tool_repositories:
+  name_template: "{{ tool_id }}"
+  description_template: "Phyloseq plot {{ tool_name }}."
```
bernt-matthias, line 13:
```suggestion
    description_template: "Phyloseq {{ tool_name }}."
```

tools/phyloseq/.shed.yml
```diff
@@ -0,0 +1,18 @@
...
+suite:
+  name: "suite_phyloseq"
+  description: "A suite of tools that produce various phyloseq plots."
```
bernt-matthias, line 16: Same as aboove

bernt-matthias (commented):
Looks good to me. Do you plan to support the other parameters of the plotting function (eg for grouping and coloring)? Or leave it for later?

In order to implement this one may also use an auxillary tabular dataset that helps the user to select metadata variables / values like here https://github.com/galaxyproject/tools-iuc/pull/4294

An alternative might be to implement this in the metadata of the rds datatype. 

I guess the rds datatype (I could add it .. we just should discuss what is needed). I think it should have boolean metadata variables that allow to investigate what info is stored in the object, and maybe the phyloseq version used to generate the object.

As far as I see the following info ca be stored in phyloseq objects

- otu_table a single object of class otu_table.
- sam_data a single object of class sample_data.
- tax_table a single object of class taxonomyTable.
- phy_tree a single object of the phylo-class, from the ape package. 
- refseq a biological sequence set object of a class that inherits from the XStringSet-class, from
the Biostrings package.

gregvonkuster:
> Do you plan to support the other parameters of the plotting function (eg for grouping and coloring)? Or leave it for later?

I'm not clear on what you mean by grouping.  With regard to color, I'm not sure how to go about supporting it in the tool form.  The docs for the current version here http://master.bioconductor.org/packages/release/bioc/manuals/phyloseq/man/phyloseq.pdf are basically the same as the docs for the older version (which is not pdf) here https://www.rdocumentation.org/packages/phyloseq/versions/1.16.2/topics/plot_richness, and setting the color seems to be dependent on the input data, so we cannot use a static select list.  Do you know of another way to allow the user to specify the color value?   The result of sample_variables(physeq_obj) is NULL with the current tool, and the only static string I could get to work for the values of both ```x``` and ```color``` is "samples".

Also, the following detail is included in the pdf doc:
```
NOTE: Because this plotting function incorporates the output from estimate_richness, the variable names of that output
should not be used as x or color (even if it works, the resulting plot might be kind of strange, and not the intended
behavior of this function). The following are the names you will want to avoid using in x or color:
c("Observed","Chao1","ACE","Shannon","Simpson","InvSimpson","Fisher").
```

gregvonkuster:
> I guess the rds datatype (I could add it .. we just should discuss what is needed).

This would be great, but can this be added as a later enhancement to the current version of the tools?  I need to get an initial working version of these tools made available as soon as possible.

tools/phyloseq/.shed.yml
gregvonkuster: Fixed - thanks!

tools/phyloseq/.shed.yml
gregvonkuster: Fixed - thanks!

tools/phyloseq/.shed.yml
gregvonkuster: Fixed - thanks!

tools/phyloseq/.shed.yml
gregvonkuster: Fixed - thanks!

bernt-matthias:
> > Do you plan to support the other parameters of the plotting function (eg for grouping and coloring)? Or leave it for later?
> 
> I'm not clear on what you mean by grouping. With regard to color, I'm not sure how to go about supporting it in the tool form. The docs for the current version here http://master.bioconductor.org/packages/release/bioc/manuals/phyloseq/man/phyloseq.pdf are basically the same as the docs for the older version (which is not pdf) here https://www.rdocumentation.org/packages/phyloseq/versions/1.16.2/topics/plot_richness, and setting the color seems to be dependent on the input data, so we cannot use a static select list. Do you know of another way to allow the user to specify the color value? The result of sample_variables(physeq_obj) is NULL with the current tool, and the only static string I could get to work for the values of both `x` and `color` is "samples".
> 
> Also, the following detail is included in the pdf doc:
> 
> ```
> NOTE: Because this plotting function incorporates the output from estimate_richness, the variable names of that output
> should not be used as x or color (even if it works, the resulting plot might be kind of strange, and not the intended
> behavior of this function). The following are the names you will want to avoid using in x or color:
> c("Observed","Chao1","ACE","Shannon","Simpson","InvSimpson","Fisher").
> ```

In the ampvis wrappers I created an additional tabular output in the input wrapper (`phyloseq_from_dada2` here). In the other tools this file is then used to create dynamic options (with `from_dataset`). Might also be possible directly with Metadata... which would be more convenient for the user.
.

bernt-matthias:
@gregvonkuster is 21.09 fine for you: https://github.com/galaxyproject/galaxy/pull/13462 ?

gregvonkuster:
> @gregvonkuster is 21.09 fine for you: [galaxyproject/galaxy#13462](https://github.com/galaxyproject/galaxy/pull/13462) ?

Yes, this works - thanks!

bernt-matthias:
Then you only need to adapt the datatypes here and we are good to go once the Galaxy PR is accepted. Fine for you?

gregvonkuster:
> Then you only need to adapt the datatypes here and we are good to go once the Galaxy PR is accepted. Fine for you?

Yes, I'll do that - thanks!

bernt-matthias:
Tests are failing because `Output output: Test output file (output.phyloseq) is missing. If you are using planemo, try adding --update_test_data to generate it.
`

gregvonkuster:
Looks like the new ```phyloseq``` datatype is not yet available:
```
Input staging problem: parameter 'file_type': an invalid option ('phyloseq') was selected
```

bernt-matthias (approved):
Good to go?

gregvonkuster:
> Good to go?

Yes!  :)

gregvonkuster:
Thanks to everyone for all of the help on this!

bernt-matthias:
Just realized that there are already a few phyloseq tools in the toolshed .. from here https://github.com/QFAB-Bioinformatics/metaDEGalaxy  .. maybe we can ask there to join forced and move those tools to IUC?

gregvonkuster:
> Just realized that there are already a few phyloseq tools in the toolshed .. from here https://github.com/QFAB-Bioinformatics/metaDEGalaxy .. maybe we can ask there to join forced and move those tools to IUC?

I took a look at those before I started on these, and I didn't see anything we could leverage from them and it doesn't look like they are being maintained.

Issue #3606, opened 2021-04-19
bcftools isec error: Expected the -p option


gallardoalba:
**Problem**: the tool generates the following standard error: Expected the -p option. However, setting that option doesn't solve the error.

- Tool: [bcftools isec](https://toolshed.g2.bx.psu.edu/repository?repository_id=71621a8a85b93435)
- Test history: https://usegalaxy.org/u/gallardoalba/h/bcftools-isec-error
- GHelp post: https://help.galaxyproject.org/t/issues-to-use-bcftools-isec-to-compare-two-different-vcf-files/5734

PR #4437, opened 2022-03-03
Bcftools: Allow vcf.gz as input files

FredericBGA:
FOR CONTRIBUTOR:


bgruening:
@FredericBGA can you please bump the version number as well?

bgruening:
https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html?highlight=version#tool-versions

We recommend to use +galaxyX for those cases. :) Thanks a lot @FredericBGA for all your help!

FredericBGA:
Thank you @bgruening 

so there is now a new token `<token name="@VERSION_SUFFIX@">1</token>` in the macros.

I've replaced: `version="@TOOL_VERSION@"` by `version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@"`

and the former `version="@TOOL_VERSION@+galaxy1"` by `version="@TOOL_VERSION@+galaxy2"`

I hope it's ok.

bgruening (approved):
Thanks.

bernt-matthias (commented):
Test would be nice.

FredericBGA:
> Test would be nice.

Yes, of course. In this particular case, how can I deal with this? Can I only add one test for one of the tool (let say bcftool call or stat as I was working on it for one of my user) using a vcf.gz file? I can compress one of the file already available in test-data and add a test accordingly.
If it's ok, I will do this asap.
Thank you @bernt-matthias

bernt-matthias:
That sounds like a good plan. Thanks @FredericBGA

tools/bcftools/bcftools_stats.xml
```diff
@@ -198,6 +198,16 @@ ${section.verbose}
             </output>
             <output name="output_pdf" file="summary.pdf" compare="sim_size" delta="25000" />
         </test>
+        <test>
+            <param name="input_file" ftype="vcf" value="mpileup.vcf.gz" />
```
bernt-matthias, line 202: I think that `ftype` should not be `vcf`, but I am unsure what would be correct. `vcf_bgzip`?

bernt-matthias:
@FredericBGA do you want to include a fix for https://github.com/galaxyproject/tools-iuc/issues/4438 here?

bernt-matthias:
There is also another open issue https://github.com/galaxyproject/tools-iuc/issues/3606.. If you have any idea how this could be fixed..

FredericBGA:
> @FredericBGA do you want to include a fix for #4438 here?

I will create a PR for this, my test is ok here.

FredericBGA:
> There is also another open issue #3606.. If you have any idea how this could be fixed..

This tool was not already installed in our Galaxy. I'm not an expert in bcftools, I'm working with my colleague (she's absent this week). The test of bcftools isec works. I think that the `-p` option should not be a boolean as it needs a value (`-p, --prefix <dir>            if given, subset each of the input files accordingly, see also -w`).

Issue #4438, opened 2022-03-04
validator of --set-id  for bcftools annotate 


FredericBGA:
Hi,

I need help on this one.

https://github.com/galaxyproject/tools-iuc/blob/d1004c06207be773c278e12745aada276b63172e/tools/bcftools/bcftools_annotate.xml#L140

The current regexp does not allow to use `%CHROM\_%POS` it fails in `lib/galaxy/tools/parameters/validation.py`

I've tried to set it to `^([+]?(%[A-Z]+)(\\_%[A-Z]+)*)?$` but it does not match either... I don't really understand right now.

The match occurs with three \ or I need to add the raw string notation `r` (https://docs.python.org/3/howto/regex.html#the-backslash-plague) before the regexp string.

With the help of https://regex101.com/ I've tried:

```
import re
reg = '^([+]?(%[A-Z]+)(\\_%[A-Z]+)*)?$'
set_id = "%CHROM\_%POS"
print(re.match(reg, set_id), 'set_id', set_id)
None set_id %CHROM\_%POS
```

```
import re
reg = '^([+]?(%[A-Z]+)(\\\_%[A-Z]+)*)?$'
set_id = "%CHROM\_%POS"
print(re.match(reg, set_id), 'set_id', set_id)
<_sre.SRE_Match object; span=(0, 12), match='%CHROM\\_%POS'> set_id %CHROM\_%POS
```

```
import re
reg = r'^([+]?(%[A-Z]+)(\\_%[A-Z]+)*)?$'
set_id = "%CHROM\_%POS"
print(re.match(reg, set_id), 'set_id', set_id)
<_sre.SRE_Match object; span=(0, 12), match='%CHROM\\_%POS'> set_id %CHROM\_%POS
```

So? Do I need to use `\\\` (and then create a PR and a test for bcftools annotate)
Or use only `\\`  and raw string notation (and then create a PR and a test for bcftools annotate and a PR for  `lib/galaxy/tools/parameters/validation.py`?

Thank you.

bernt-matthias:
Adding a test (or extending a present test) seems like a great idea, since the parameter is untested at the moment.

I guess 3 backslashes are correct.

FredericBGA:
Solved in https://github.com/galaxyproject/tools-iuc/pull/4442

PR #4442, opened 2022-03-07
Change the validator for set_id, add test, bump version

FredericBGA:
FOR CONTRIBUTOR:

Regex modified to allow backslash (`\_` between tags for command line) and underscores in tags (for `%FIRST_ALT`) for the `--set-id` option. The example given on the interface works (used in the test: `%CHROM\_%POS\_%REF\_%FIRST_ALT`)
I've modified the second test (add an assert_contents: `<has_text text="1_3062915_GTTT_G" />`

### DIFF
--- a/tools/bcftools/bcftools_annotate.xml
+++ b/tools/bcftools/bcftools_annotate.xml
@@ -118,8 +118,8 @@ $sec_annofile.annofile.header_lines.__str__.strip()#slurp
                         Note that in case of tab-delimited file, the coordinates POS, FROM and TO are one-based and inclusive.
                         ]]></help>
                     </param>
-                    <param name="header_file" type="data" format="txt" label="Header Lines File" optional="True" help="lines which should be appended to the VCF header" />
-                    <param name="header_lines" type="text" area="True" label="Header Lines" optional="True" help="lines which should be appended to the VCF header" >
+                    <param name="header_file" type="data" format="txt" label="Header Lines File" optional="true" help="lines which should be appended to the VCF header" />
+                    <param name="header_lines" type="text" area="true" label="Header Lines" optional="true" help="lines which should be appended to the VCF header" >
                         <help><![CDATA[
                         ##INFO=<ID=NUMERIC_TAG,Number=1,Type=Integer,Description="Example header line">
                         ##INFO=<ID=STRING_TAG,Number=1,Type=String,Description="Yet another header line">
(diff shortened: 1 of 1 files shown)


nsoranzo:
@FredericBGA I've added a couple of commits, hope you are OK with the changes.

FredericBGA:
> @FredericBGA I've added a couple of commits, hope you are OK with the changes.

yes, it's nice to lint the file.
By the way are you sure about the `unnecessary backslash in regex`?


I've tested with:
```
reg = '^([+]?(%[A-Z]+)(\\_%[A-Z_]+)*)?$'
set_id = "%CHROM\_%POS\_%REF\_%FIRST_ALT"
print(re.match(reg, set_id), 'set_id', set_id)
# None set_id %CHROM\_%POS\_%REF\_%FIRST_ALT
```
and
```
reg = '^([+]?(%[A-Z]+)(\\\_%[A-Z_]+)*)?$'
set_id = "%CHROM\_%POS\_%REF\_%FIRST_ALT"
print(re.match(reg, set_id), 'set_id', set_id)
# <_sre.SRE_Match object; span=(0, 30), match='%CHROM\\_%POS\\_%REF\\_%FIRST_ALT'> set_id %CHROM\_%POS\_%REF\_%FIRST_ALT
```
It does not match with only two backslash. I've opened an issue here: https://github.com/galaxyproject/tools-iuc/issues/4438 to discuss about that. I'm not an expert in regex! I didn't know about raw string notation before I had to dig in this issue. Any help and explanations are welcomed.

nsoranzo:
The tests pass though. And also in the UI it works for me with 2 slashes, no?

nsoranzo (approved):

FredericBGA:
indeed ^^ it works with only two. But I don't yet really understand how. And sorry for the mistake, I've tested with a python script without trying then with two in the XML.

I've looked at `https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/parameters/validation.py` and I've tried to reproduce (with a python shell from the .venv of Galaxy) what we have in the docstring of `RegexValidator`. And here again I need three. Nevermind, it's perfect as it works but if somebody can explain me what I miss, I will be happy.
```
>>> from galaxy.util import XML
>>> from galaxy.tools.parameters.basic import ToolParameter
>>> p = ToolParameter.build(None, XML('''<param name="blah" type="text" value="%CHROM\_%POS\_%REF\_%FIRST_ALT"><validator type="regex" message="Nope" >^([+]?(%[A-Z]+)(\\\_%[A-Z_]+)*)?$</validator></param>'''))
>>> t = p.validate("%CHROM\_%POS\_%REF\_%FIRST_ALT")
>>>
>>>
>>> p = ToolParameter.build(None, XML('''<param name="blah" type="text" value="%CHROM\_%POS\_%REF\_%FIRST_ALT"><validator type="regex" message="Nope" >^([+]?(%[A-Z]+)(\\_%[A-Z_]+)*)?$</validator></param>'''))
>>> t = p.validate("%CHROM\_%POS\_%REF\_%FIRST_ALT")
  File "<stdin>", line 1, in <module>
  File "/softs/bioinfo/galaxy-dev/lib/galaxy/tools/parameters/basic.py", line 295, in validate
    return super(TextToolParameter, self).validate(value, trans)
  File "/softs/bioinfo/galaxy-dev/lib/galaxy/tools/parameters/basic.py", line 233, in validate
    validator.validate(value, trans)
  File "/softs/bioinfo/galaxy-dev/lib/galaxy/tools/parameters/validation.py", line 91, in validate
    raise ValueError(self.message)
ValueError: Nope
```

nsoranzo:
@FredericBGA The issue with your Python code is that if you want to write a literal `\` inside a Python string you need to escape it as `\\`. The regular expression in the tool XML doesn't need that though because it's read from the file by Galaxy.

PR #4453, opened 2022-03-14
Updating tools/ncbi_datasets from version 13.4.0 to
14.1.0

planemo-autoupdate:
Hello! This is an automated update of the following tool: **tools/ncbi_datasets**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

I have updated tools/ncbi_datasets from version 13.4.0 to 13.5.1.

**Project home page:** https://www.ncbi.nlm.nih.gov/datasets


bernt-matthias:
According to https://github.com/ncbi/datasets/issues/106 gbff download for the virus data package was disabled. I guess we do not reflect this in the tool form. Maybe we should document it?

PR #4455, opened 2022-03-15
GetOrganelle tool 

o-william-white:
Hello, 

I have tried to create an .xml wrapper for GetOrganelle https://github.com/Kinggerm/GetOrganelle. I think this is a useful tool and it would be great if it was widely accessible on Galaxy.

At the moment I have included the main input parameters and a selection of more advanced parameters that are commonly used. There are many other parameters and I was not sure how important these would be to the average user. However, I am more than happy to keep adding to this as I am keen to learn how to develop galaxy tools. In terms of output, I only kept the main assembly output files and not included the output from the seed and spades output. 

There are a couple of issues I am already aware of. For example, the --reduce-reads-for-coverage is coded as an integer in the xml, but you can also set this as inf on the command line to denote no limit. Also, if you select the anonymous organelle type, you must specifiy -s and --genes inputs. I will work on this next. 

I created this pull request to let you know I am working on this and perhaps get some initial feedback. 

Best wishes
Ollie

 

FOR CONTRIBUTOR:


tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,110 @@
+<tool id="get_organelle_from_reads" name="Get organelle from reads" version="0.1.0" python_template_version="3.5">
```
bernt-matthias, line 1: The tool version should be the same as the version of the used conda package. It would be best to introduce a macro token for this, see https://planemo.readthedocs.io/en/latest/writing_advanced.html#macros-reusable-elements

See for instance here https://github.com/galaxyproject/tools-iuc/blob/40a09cbfd6052f7b0295946621db1bdf58228b09/tools/abricate/abricate.xml#L1 .. the @TOOL_VERSION@ macro is in this example defined in an extra file https://github.com/galaxyproject/tools-iuc/blob/40a09cbfd6052f7b0295946621db1bdf58228b09/tools/abricate/macros.xml#L13

But you may also define it in the same file .. guess this would be better here.

bernt-matthias: `python_template_version` is not needed. But add a `profile` attribute. I would suggest `21.01`

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,110 @@
...
+    <requirements>
+        <requirement type="package" version="1.7.5">GetOrganelle</requirement>
```
bernt-matthias, line 3:
```suggestion
        <requirement type="package" version="1.7.5">getorganelle</requirement>
```

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,110 @@
...
+	    #if $settings.genes
+                --genes $settings.genes
+            #end if
+	#end if
+    ]]></command>
+    <inputs>
+	<conditional name="fastq_input">
+            <param name="fastq_input_selector" type="select" label="Single-end or paired reads">
+                <option value="single" selected="true">Single-end</option>
+                <option value="paired">Paired</option>
```
bernt-matthias, line 37: Would be cool to also have paired collection input: https://planemo.readthedocs.io/en/latest/writing_advanced.html#processing-pairs

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,110 @@
...
+            </param>
+            <when value="single">
+                <param type="data" name="fastq_input1" format="fastq" label="Single-end fastq"/>
```
bernt-matthias, line 40: Do you know if the tool supports `fastq.gz`? Then this might be added to the list of formats. 

Are you aware of the fastq encoding supported by getorganelle? Is it sanger / illumina encoding? Galaxy has special datatypes for this.

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,110 @@
...
+            </when>
+        </conditional>
+        <param type="select" name="organelle_type" label="Organelle type">
+            <option value="embplant_pt" selected="true">Plant chloroplast</option>
+            <option value="embplant_mt">Plant mitochondrial</option>
+            <option value="embplant_nr">Plant ribosomal</option>
+            <option value="fungus_mt">Fungus mitochondrial</option>
+            <option value="fungus_nr">Fungus nuclear</option>
+            <option value="animal_mt">Animal mitochondrial</option>
+	    <option value="other_mt">Other mitochondrial</option>
```
bernt-matthias, line 54: Please fix indentation.

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,110 @@
...
+        </param>
+	<conditional name="settings">
+	    <param name="advanced" type="select" label="Specify advanced parameters">
+                <option value="simple" selected="true">No, use program defaults.</option>
+                <option value="advanced">Yes, see full parameter list.</option>
+	    </param>
+            <when value="simple">
+            </when>
+	    <when value="advanced">
+		<param type="data" name="s" format="fasta" optional="true" label="Seed sequence(s)" help="Fasta file to use as initial seed. A seed sequence in GetOrganelle is only used for identifying initial organelle reads. The assembly process is purely de novo."/>
```
bernt-matthias, line 65: Also here indentation is off.

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,110 @@
...
+        #if str( $fastq_input.fastq_input_selector ) == "single":
+	    -u '${fastq_input.fastq_input1}'
+        #end if
+        #if str($fastq_input.fastq_input_selector) == "paired":
+	    -1 '${fastq_input.fastq_input1}' -2 '${fastq_input.fastq_input2}'
+        #end if
+	-o results_directory -F '$organelle_type'
+	#if $settings.advanced == "advanced"
+	    #if $settings.s
+                -s $settings.s
```
bernt-matthias, line 16: this should be single quoted as well.

tools/getorganelle/get_organelle_from_reads.xml
bernt-matthias: Please add the `argument` attribute where possible. Also note that the name attribute might then not be necessary anymore -- since name is generated from argument by removing leading dashes and replacing internal dashes by underscore.

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,110 @@
...
+	<conditional name="settings">
+	    <param name="advanced" type="select" label="Specify advanced parameters">
+                <option value="simple" selected="true">No, use program defaults.</option>
+                <option value="advanced">Yes, see full parameter list.</option>
+	    </param>
+            <when value="simple">
+            </when>
+	    <when value="advanced">
+		<param type="data" name="s" format="fasta" optional="true" label="Seed sequence(s)" help="Fasta file to use as initial seed. A seed sequence in GetOrganelle is only used for identifying initial organelle reads. The assembly process is purely de novo."/>
+		<param type="integer" name="max_reads" optional="true" label="Maximum number of reads" help="Maximum number of reads to be used. An input larger than 536870911 will be treated as infinity (INF). Default: 1.5E7 (plant chloroplast, plant ribosomal, fungus mitochondrial, fungus nuclear) 7.5E7 (Plant mitochondrial) 3E8 (Animal mitochondrial)"/>
```
bernt-matthias, line 66: Since you have `optional="true"` you may also use the case that no value is given as `inf`

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,110 @@
...
+		<param type="integer" name="w" optional="true" label="Spades kmer settings" help="Word size for pre-grouping and extending process. This script would try to guess (auto-estimate) a proper word size using an empirical function based on average read length, reads quality, target genome coverage, and other variables that might influence the extending process. Default: estimated automatically"/>
+		<param type="text" name="k" value="21,55,85,115" label="Spades kmer settings" help="Use the same format as in SPAdes. Illegal kmer values would be automatically discarded by GetOrganelle. Default: 21,55,85,115"/>
+		<param type="data" name="genes" format="fasta" optional="true" label="Gene sequence(s)" help="Fasta file containing protein coding genes and ribosomal RNAs extracted from a reference genome that you want to assemble."/>
+	    </when>
+	</conditional>
+    </inputs>
+    <outputs>
+	<collection name="path_sequence_list" type="list" label='${tool.name} on ${on_string}: Path sequences'>
+            <discover_datasets pattern="(?P&lt;designation&gt;.+)\.path_sequence.fasta" format="Fasta" directory="results_directory" />
```
bernt-matthias, line 76: lower case `fasta` in format

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,110 @@
...
+        </collection>
+        <data format="txt" name="selected_graph" from_work_dir="results_directory/*.selected_graph.gfa" label='${tool.name} on ${on_string}: Selected graph'/>
```
bernt-matthias, line 78: Galaxy supports gfa1 and gfa2 data types.

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,110 @@
...
+	<data format="txt" name="logfile" from_work_dir="results_directory/get_org.log.txt" label='${tool.name} on ${on_string}: Logfile'/>
+	<data format="Fastg" name="assembly_graph" from_work_dir="results_directory/extended_K*.assembly_graph.fastg" label='${tool.name} on ${on_string}: Raw assembly graph'/>
```
bernt-matthias, line 80: lower case `fastg`

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,110 @@
...
+		<param type="integer" name="w" optional="true" label="Spades kmer settings" help="Word size for pre-grouping and extending process. This script would try to guess (auto-estimate) a proper word size using an empirical function based on average read length, reads quality, target genome coverage, and other variables that might influence the extending process. Default: estimated automatically"/>
+		<param type="text" name="k" value="21,55,85,115" label="Spades kmer settings" help="Use the same format as in SPAdes. Illegal kmer values would be automatically discarded by GetOrganelle. Default: 21,55,85,115"/>
+		<param type="data" name="genes" format="fasta" optional="true" label="Gene sequence(s)" help="Fasta file containing protein coding genes and ribosomal RNAs extracted from a reference genome that you want to assemble."/>
+	    </when>
+	</conditional>
+    </inputs>
+    <outputs>
```
bernt-matthias, line 74: With filters (https://docs.galaxyproject.org/en/master/dev/schema.html#tool-outputs-data-filter) you may make some of the outputs optional. In the bast case in combination with an additional select.

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,110 @@
...
+	        <has_text text="Thank you!" />
+	    </assert_stdout>
+        </test>
+    </tests>
+    <help><![CDATA[
+	    GetOrganelle v1.7.5.3 get_organelle_from_reads.py assembles organelle genomes from genome skimming data. 
+	    See https://github.com/Kinggerm/GetOrganelle for more information.
+	    ]]></help>
+    <citations>
+        <citation type="bibtex">
```
bernt-matthias, line 100: Use a doi style citation. With [10.1186/s13059-020-02154-5](https://doi.org/10.1186/s13059-020-02154-5)

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,110 @@
...
+        <data format="txt" name="selected_graph" from_work_dir="results_directory/*.selected_graph.gfa" label='${tool.name} on ${on_string}: Selected graph'/>
+	<data format="txt" name="logfile" from_work_dir="results_directory/get_org.log.txt" label='${tool.name} on ${on_string}: Logfile'/>
+	<data format="Fastg" name="assembly_graph" from_work_dir="results_directory/extended_K*.assembly_graph.fastg" label='${tool.name} on ${on_string}: Raw assembly graph'/>
+	<data format="Fastg" name="simplfied_graph" from_work_dir="results_directory/extended_K*.assembly_graph.fastg.extend-*.fastg" label='${tool.name} on ${on_string}: Simplfied graph'/>
+	<data format="tsv" name="assembly_csv" from_work_dir="results_directory/extended_K*.assembly_graph.fastg.extend-*.csv" label='${tool.name} on ${on_string}: Contig labels'/>
+    </outputs>
+    <tests>
+	<test>
+	    <param name="fastq_input_selector" value="paired"/>	    
+            <param name="fastq_input1" value="Arabidopsis_simulated.1.fq.gz"/>
```
bernt-matthias, line 87: Please add `ftype`. 

So the tool is supporting fastq.gz :) ..

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,110 @@
...
+	    <param name="fastq_input2" value="Arabidopsis_simulated.2.fq.gz"/>
+	    <param name="organelle_type" value="embplant_pt" />
+	    <assert_stdout>
```
bernt-matthias, line 90: Tests for the outputs would be great.

bernt-matthias (commented):
Really good start

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: Hi @bernt-matthias, apologies I was not certaian what part needed to be single quoted here? Was it on a specific line or an entire section?

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: Yes the tool supports ```fastq/gz```. The documentation says illumina pair-end/single-end data (fastq or fastq.gz). I will change the datatype for all inputs to ```format="fastqsanger,fastqsanger.gz" ```

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: Good idea, adding this now

tools/getorganelle/get_organelle_from_reads.xml
bernt-matthias: Most likely the tool relies on the file extension to check if its compressed or not. But in Galaxy all files all files have a `.dat` extension.

The solution is in most cases a symlink: `ln -s '$galaxy_input' forward.$ext`.

Here `$ext` is a variable that you need to determine first by checking `$galaxy_input.ext`  which holds the Galaxy datatype name of the input. Maybe this is a good example https://github.com/galaxyproject/tools-iuc/blob/5fcc308928ab2417b7e40227b27a3955f227649d/tools/colibread/macros.xml#L21

o-william-white: Hi @o-william-white many thanks for explaining, I have been able to make most changes, I will try to resolve this and upload my changes for review tomorrow.

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: Hi @bernt-matthias I am not sure why, but when I add ```profile="@PROFILE@"```, my test fails. For some reason it cannot find the reference datasets use for identifyng seed sequences. See error message below. I will not include the profile option for now. 


```
############################################################################
ERROR: default embplant_pt,embplant_mt database not added yet!

Install it by: get_organelle_config.py -a embplant_pt,embplant_mt
or
Install all types by: get_organelle_config.py -a all
```

bernt-matthias: If you like you can push the failing version. Then I can have a look

o-william-white: Yes I will push this afternoon, just wanted to address one more of your earlier comments

o-william-white: Hi @bernt-matthias I have updated the getorganelle tool here https://github.com/o-william-white/tools-iuc/tree/master/tools/getorganelle I think I was able to include most comments you suggsted. Thank you again for the feedback! Do I need to create another PR?

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: Added

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: I checked my indentation and I couldn't see where I was going wrong. Could you highlight one example where I went wrong and I will fix the rest?

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: Added quotes where necessary. I wasn't sure if curly brackets were needed. For example ```'${settings.s}'``` I have included but miught be unnecessary

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: Done

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: Changed to gfa1

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: Done

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: Done

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: Done

tools/getorganelle/get_organelle_from_reads.xml
bernt-matthias: The last two options here have a different indentation than the others. I guess you mix tabs and spaces. Please use only spaces.

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: I tried to add another test for the output collection such that it must hae a count of two. At least ths is what I hoped to achieve! ```<output_collection name="path_sequence_list" type="list" count="2" />```

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: Ah I undersand now thanks. I use vim and I think it sometimes adds tabs. I will change now

tools/getorganelle/get_organelle_from_reads.xml
bgruening: In that case I guess we should output an empty file?

o-william-white: That's what I thought might happen. I think novoplasty works in a similar way, it produces all possible output files, even if some are empty.

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: The indentations should all spaces now

bernt-matthias:
Seems that the tool also fails without the profile with: 

```
############################################################################
ERROR: default embplant_pt,embplant_mt database not added yet!

Install it by: get_organelle_config.py -a embplant_pt,embplant_mt
or
Install all types by: get_organelle_config.py -a all
```

Probably on you local system you have the data bases in your home dir.. which is why it works for you.

I previously reported this here: https://github.com/Kinggerm/GetOrganelle/issues/64 

So the seed DBs need to be downloaded before running the tool. The preferable way would be to store them in a data table (https://galaxyproject.org/admin/tools/data-tables/). For now the data table could be managed manually, 

So the Galaxy admin would need to initialize the db with:  `get_organelle_config.py --use-version 0.0.1 --config-dir GALAXY_TOOL_DATA_DIR/get_organelle/0.0.1/ -a all` and create an entry in the data table specifying the version and path.
Alternatively one could also think about a data manager tool .. but this may come later (if needed).

Problem is that the seed databases are quite large .. I will do some experiments.

There is also small test data that we might use https://github.com/Kinggerm/GetOrganelle/issues/64#issuecomment-763482665

Last thing that I noticed is that you created the PR from your `master` branch .. on the long run branches would be better :)

tools/getorganelle/get_organelle_from_reads.xml
bernt-matthias: > From what I could understand, this works where an argument can be used to determine the output?

Yes, often tool developers add an additional select that allows users to choose which (additional) outputs they are interested in. So one can "hide" output that are not always of interest.

o-william-white:
Hi @bernt-matthias

I understand now, yes I had installed the get organelle seed databses locally. Are the seed databases large enough to be problematic for usage on galaxy? 

Regarding the test data, I am already using the smallest test data availble from get organelle, specifically example 1 from here https://github.com/Kinggerm/GetOrganelle/wiki/Examples

Apologies I didn't realise I shouldn't create a pull request from the master branch. If I understand correctly, should I create a new branch and create a new PR from there? 

I will also make the changes suggested above this week, specifically: 
 - add output filters
 - add ```inf``` option for max reads

Thanks for your pateince with this

Cheers
Ollie

bernt-matthias:
> I understand now, yes I had installed the get organelle seed databses locally. Are the seed databases large enough to be problematic for usage on galaxy?

They are not problematic for Galaxy, but too large for this repo :( We have a limit of 1MB per file and the seed databases are much larger. 

Not sure if we can downsample one of the given ones. Such that still results are produced? I also asked at the upstream issue.

Maybe we can just proceed and add a data table and use the original (large) seed database for now just to get the tool and tests finished for now and try to reduce test data later...

> Regarding the test data, I am already using the smallest test data availble from get organelle, specifically example 1 from here https://github.com/Kinggerm/GetOrganelle/wiki/Examples

Then its fine.

> Apologies I didn't realise I shouldn't create a pull request from the master branch. If I understand correctly, should I create a new branch and create a new PR from there?

Its fine for now. But after the PR is merged you will need to fix/reset your branches (don't do it before since all changes will be lost):

```
git fetch --all
git reset --hard upstream/master
git push -f origin master
```

(assuming that your remotes are called `origin` and `master`)

> I will also make the changes suggested above this week, specifically:
> 
>     * add output filters
> 
>     * add `inf` option for max reads

Wonderful.

o-william-white:
Hi @bernt-matthias 

Apologies for my slow reply on this. 

I have added output filters for the advanced settings and the inf option for --reduce-reads-for-coverage.

However, I have tested the .xml and when I set an advanced option in galaxy (for example changing the word size or spades kmer sizes), it does not change the command, instead using only the simple input paramters. 

Can you see where I might be going wrong https://github.com/o-william-white/tools-iuc/blob/master/tools/getorganelle/get_organelle_from_reads.xml 

I have been looking for the source of the error this afternoon but can't find it. I expect I have missed something simple! 

Best wishes
Ollie

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,185 @@
...
+            #if '${settings.w}':
+                -w '${settings.w}'
+            #end if
+            #if '${settings.k}':
+                -k '${settings.k}'
+            #end if
+            #if '${settings.genes}':
+                --genes '${settings.genes}'
+            #end if
+        #end if
```
bernt-matthias, line 84: Hi @o-william-white I think the single quotes in the if statements are wrong. 

Can you try this

```suggestion
       #if $settings.advanced == "advanced":
            #if $settings.s
                -s '$settings.s'
            #end if
            #if $settings.max_reads:
                --max-reads $settings.max_reads
            #end if
            #if $settings.reduce_reads_for_coverage:
                #if $settings.reduce_reads_for_coverage == 0:
                    --reduce-reads-for-coverage inf
                #else:
                    --reduce-reads-for-coverage $settings.reduce_reads_for_coverage
                #end if
            #end if
            #if $settings.w':
                -w $settings.w
            #end if
            #if $settings.k:
                -k $settings.k
            #end if
            #if $settings.genes:
                --genes '$settings.genes'
            #end if
```

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,185 @@
...
+        </data>
+        <data format="sam" name="seed_sam" from_work_dir="results_directory/seed/*.sam" label='${tool.name} on ${on_string}: Seed sam'>
+            <filter>settings['advanced'] == 'advanced' and "seed_sam" in settings['out']</filter>
+        </data>
+        <data format="txt" name="spades_log" from_work_dir="results_directory/extended_spades/spades.log" label='${tool.name} on ${on_string}: Spades log'>
+            <filter>settings['advanced'] == 'advanced' and "spades_log" in settings['out']</filter>
+        </data>
+    </outputs>
+    <tests>
+        <test>
```
bernt-matthias, line 167: Can you add the `expect_num_outputs` attribute to `test` tags?

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: This has fixed it thank you!

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: I added ```expect_num_outputs``` to the ```test``` tag. 

The advanced paramters seem to be working as expected now.

bgruening:
I think there is now version 1.7.6.1 available :)

Thanks @o-william-white for working on it.

bernt-matthias:
> I think there is now version 1.7.6.1 available :)
> 
> Thanks @o-william-white for working on it.

Indeed. And it seems that the authors provided a minimal test database which is just a bit larger than 1mb.

bernt-matthias:
> Is the this the test data discussed here https://github.com/Kinggerm/GetOrganelle/wiki/Example-1? I used this data in my test.

See discussion here https://github.com/Kinggerm/GetOrganelle/issues/64 .. inparticular at the end

> Is there anything else needed for this tool? @bernt-matthias mentioned a data manager for reference datasets, I can have a look at this next, perhaps you could recommend an example and I will try to replicate for get organelle.

Most data managers in https://github.com/galaxyproject/tools-iuc/tree/master/data_managers/ will be a good example .. maybe https://github.com/galaxyproject/tools-iuc/tree/master/data_managers/data_manager_mitos .. just because it also considers mitogenomes

o-william-white:
Hi @bernt-matthias 

Thanks again for the additional details. If I understand correctly, the discussion at the end of https://github.com/Kinggerm/GetOrganelle/issues/64 refers to the seed and label database used to generate the assembly. It does indeed look a lot smaller in size to the orginal seed and label databases. 

The orginal seed/label databse is 1.3M and the smaller databse is 292K. 

Would it be better to use the smaller seed/label database in data manager? 

However, the data used in the test from here https://github.com/Kinggerm/GetOrganelle/wiki/Example-1 is still the same. 
```
7.8M    test-data/Arabidopsis_simulated.1.fq.gz
8.1M    test-data/Arabidopsis_simulated.2.fq.gz
```

bernt-matthias:
The data manager should only download the real seed/label databases.

The smaller/minimal seed/label databases should be used for the tests only. The idea would be to create a data table containing an entry for these databases.

You need to create 

- `tool_data_table_conf.xml.sample` files
- `tool_data_table_conf.xml.test`
- and mock data table (referenced in the `.test` file .. as here [.loc](https://github.com/galaxyproject/tools-iuc/blob/master/tools/mitos/test-data/mitos.loc)
- and use them in a test like this https://github.com/galaxyproject/tools-iuc/blob/8b58f3f03d6430689d228029bb2eb46c16cfff23/tools/mitos/mitos2.xml#L169

See for instance here: https://github.com/galaxyproject/tools-iuc/tree/master/tools/mitos

Kinggerm:
> Hi @bernt-matthias
> 
> Thanks again for the additional details. If I understand correctly, the discussion at the end of [Kinggerm/GetOrganelle#64](https://github.com/Kinggerm/GetOrganelle/issues/64) refers to the seed and label database used to generate the assembly. It does indeed look a lot smaller in size to the orginal seed and label databases.
> 
> The orginal seed/label databse is 1.3M and the smaller databse is 292K.
> 
> Would it be better to use the smaller seed/label database in data manager?
> 
> However, the data used in the test from here https://github.com/Kinggerm/GetOrganelle/wiki/Example-1 is still the same.
> 
> ```
> 7.8M    test-data/Arabidopsis_simulated.1.fq.gz
> 8.1M    test-data/Arabidopsis_simulated.2.fq.gz
> ```

Actually, if you want the test data to be as small as possible, the top 20k reads (in the file size of 1Mb * 2) of the  `test-data/Arabidopsis_simulated` set are enough to generate a successful complete genome assembly. The only difference from usual is that the tool will tune the parameters to an extreme for shallow coverage and generate some warnings indicating that the coverage is too shallow.
Let me know if you need that smallest dataset, I can upload it to a public somewhere or you may simply get it through `gunzip -c Arabidopsis_simulated.1.fq.gz|head -n 40000 > Arabidopsis_simulated.10k.1.fq`

o-william-white:
Hi @bernt-matthias and @Kinggerm,

Thanks both for the comments. I have used the small input data set and seed/label databases as suggested. I made the aditional scripts required and my planemo test passed succesfully.
 
I only kept the embplant_pt and embplant_mt databses as these seem to be the minimimum requirement. 

The test data directory is the following size.

```
2.2M    test-data/
```

Best wishes
Ollie

bernt-matthias:
Great. I will have a look. One thing I noticed: You should also add a sample `.loc` file like here https://github.com/galaxyproject/tools-iuc/blob/master/tools/mitos/mitos.loc.sample

Another question is if the reference datasets are/will be versioned (and specific tool versions require reference data version), then we would need a version column. Do you have plans in this direction @Kinggerm ?

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,193 @@
...
+            #set $in1_name = re.sub('[^\w\-\s]', '_', str($fastq_input.paired_input.name)) + $ext
+            #set $in2_name = re.sub('[^\w\-\s]', '_', str("%s_%s" % ($fastq_input.paired_input.name, "R2"))) + $ext
+            ln -s '$in1' '$in1_name' &&
+            ln -s '$in2' '$in2_name' &&
+        #end if    
+
+        ## GetOrganelle cmd
+
+        get_organelle_from_reads.py
+        #if str( $fastq_input.fastq_input_selector ) == "single":
```
bernt-matthias, line 51: Can you remove spaces around parentheses?

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,193 @@
...
+            <option value="animal_mt">Animal mitochondrial</option>
+            <option value="other_mt">Other mitochondrial</option>
+            <option value="anonym">Anonymous</option>
+        </param>
+        <param argument="--config-dir" label="Reference data" type="select" help="Contact the administrator of this Galaxy instance if you miss reference data">
+            <options from_data_table="getorganelle">
+                <filter type="static_value" value="getorganelle" column="2"/>
+            </options>
+            <validator message="No reference annotation is available for getorganelle" type="no_options" />
+        </param>
```
bernt-matthias, line 120: Actually that could be a single input, if the loc file would contain an entry for each reference set. 

Or we keep it as is and assume that we can select versions, which would imply that the reference databases always contain references contained in the select list. If the select list changes then we would have a problem. Maybe a minor argument in favor of my first idea.

What do you think?

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,193 @@
...
+            </when>
+            <when value="advanced">
+                <param name="out" type="select" label="Select output files" multiple="true" display="checkboxes">
+                    <option value="path_sequences">Path sequences</option>
+                    <option value="selected_graph">Selected graph</option>
+                    <option value="raw_assembly_graph">Raw assembly graph</option>
+                    <option value="simplified_graph">Simplified graph</option>
+                    <option value="contig_labels">Contig labels</option>
+                    <option value="seed_fastq">Seed fastq</option>
+                    <option value="seed_sam">Seed sam</option>
```
bernt-matthias, line 136: Maybe capitalize SAM?

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,193 @@
...
+        </data>
+        <data format="sam" name="seed_sam" from_work_dir="results_directory/seed/*.sam" label='${tool.name} on ${on_string}: Seed sam'>
+            <filter>settings['advanced'] == 'advanced' and "seed_sam" in settings['out']</filter>
+        </data>
+        <data format="txt" name="spades_log" from_work_dir="results_directory/extended_spades/spades.log" label='${tool.name} on ${on_string}: Spades log'>
+            <filter>settings['advanced'] == 'advanced' and "spades_log" in settings['out']</filter>
+        </data>
+    </outputs>
+    <tests>
+        <test>
```
bernt-matthias, line 174: Can you add the `expect_num_outputs` attribute?

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,193 @@
...
+            <param name="fastq_input2" value="Arabidopsis_simulated.10k.2.fq.gz" ftype="fastqsanger.gz"/>
+            <param name="organelle_type" value="embplant_pt" />
+            <assert_stdout>
+                <has_text text="Thank you!" />
+            </assert_stdout>
+            <output_collection name="path_sequence_list" type="list" count="2" />
+        </test>
+    </tests>
+    <help><![CDATA[
+        GetOrganelle v1.7.5.3 get_organelle_from_reads.py assembles organelle genomes from genome skimming data.
```
bernt-matthias, line 187: I would remove the tool verison from the help.

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,193 @@
...
+        </conditional>
+        <param type="select" argument="-F" label="Organelle type">
+            <option value="embplant_pt" selected="true">Plant chloroplast</option>
+            <option value="embplant_mt">Plant mitochondrial</option>
+            <option value="embplant_nr">Plant ribosomal</option>
+            <option value="fungus_mt">Fungus mitochondrial</option>
+            <option value="fungus_nr">Fungus nuclear</option>
+            <option value="animal_mt">Animal mitochondrial</option>
+            <option value="other_mt">Other mitochondrial</option>
+            <option value="anonym">Anonymous</option>
```
bernt-matthias, line 113: What is Anonymous?

Kinggerm: Maybe not a good name. It refers to any loci other than the above, e.g., people may be interested in doing an animal nr assembly from WGS, but I do not have a default database prepared. In the anonym mode, people have to provide their own seed/label databases and have no fine-tuned parameter set.

Kinggerm:
> Great. I will have a look. One thing I noticed: You should also add a sample `.loc` file like here https://github.com/galaxyproject/tools-iuc/blob/master/tools/mitos/mitos.loc.sample
> 
> Another question is if the reference datasets are/will be versioned (and specific tool versions require reference data version), then we would need a version column. Do you have plans in this direction @Kinggerm ?

Yes, the GetOrganelleDB was versioned already. 
But currently, we only have two main versions (`0.0.0` and `0.0.1`), where
`GetOrganelle <1.7.3` only works with `GetOrganelleDB 0.0.0` while `GetOrganelle 1.7.3+` work with both `GetOrganelleDB 
 0.0.0` and `GetOrganelleDB 0.0.1`. I have plans for updating GetOrganelleDB but I cannot do it very soon. So `0.0.1` will lasts long.

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: Apologies, thought I had included this already, added back

tools/getorganelle/get_organelle_from_reads.xml
bernt-matthias: So there needs to be a parameter to specify database fasta (?) file?

bernt-matthias:
> Yes, the GetOrganelleDB was versioned already. But currently, we only have two main versions (`0.0.0` and `0.0.1`), where `GetOrganelle <1.7.3` only works with `GetOrganelleDB 0.0.0` while `GetOrganelle 1.7.3+` work with both `GetOrganelleDB 0.0.0` and `GetOrganelleDB 0.0.1`. I have plans for updating GetOrganelleDB but I cannot do it very soon. So `0.0.1` will last long.

Thanks for the clarification.

tools/getorganelle/get_organelle_from_reads.xml
Kinggerm: > So there needs to be a parameter to specify the database fasta (?) file?

Yeah, for `-F anonym`, specifying the database files is required through `-s` and `--genes`. An error will prompt with guidelines if users did not specify the database files but attempted to use `-F anonym`.

o-william-white: Hi both, I have added paramters```-s``` and ```--genes``` under advanced options, however, I haven't yet added a rule that if ```-F anonym``` is used, the user must specify these advanced options. To be honest I wasn't sure how to do this yet.

o-william-white: I will have a look at this again later today

bgruening:
> > Great. I will have a look. One thing I noticed: You should also add a sample `.loc` file like here https://github.com/galaxyproject/tools-iuc/blob/master/tools/mitos/mitos.loc.sample
> > Another question is if the reference datasets are/will be versioned (and specific tool versions require reference data version), then we would need a version column. Do you have plans in this direction @Kinggerm ?
> 
> Yes, the GetOrganelleDB was versioned already. But currently, we only have two main versions (`0.0.0` and `0.0.1`), where `GetOrganelle <1.7.3` only works with `GetOrganelleDB 0.0.0` while `GetOrganelle 1.7.3+` work with both `GetOrganelleDB 0.0.0` and `GetOrganelleDB 0.0.1`. I have plans for updating GetOrganelleDB but I cannot do it very soon. So `0.0.1` will last long.

I think this is mostly ready. But as you pointed out that the version might change, I think it does not hurt to add a version column in the loc file. What do you think?

o-william-white:
Hello @bernt-matthias and @bgruening,

Thanks both for the feedback and apologies for the delay in getting back to you. I have made the following changes: 

- added a ```getorganelle.loc.sample``` file
- added a column for the get organelle database version to the ```test-data/getorganelle.loc``` file

I was not sure if ```tool_data_table_conf.xml.sample``` and ```tool_data_table_conf.xml.test``` needed to be adjusted to reflect this. 

If you have any further comments or suggestions, please do let me know.

Best wishes
Ollie

o-william-white:
image

I tried to test the tool in my local galaxy but it cannot find the reference data. I am assuming this is something that needs to be set up with the galaxy tool data manager?

bernt-matthias:
> image
> 
> I tried to test the tool in my local galaxy but it cannot find the reference data. I am assuming this is something that needs to be set up with the galaxy tool data manager?

Actually it should be possible to edit the .loc file and entries should show up. During startup the log should show how many entries are loaded from each data table.

You can also see the content of the data tables in the admin interface.

tools/getorganelle/getorganelle.loc.sample
```diff
@@ -0,0 +1 @@
+#ID	name	getorganelle-version	location	getorganelle-db-version
```
bernt-matthias, line 1: Note that columns must be tab separated  .. not sure if they are here.

tools/getorganelle/macros.xml
```diff
@@ -0,0 +1,3 @@
+<macros>
+    <token name="@TOOL_VERSION@">1.7.6.1</token>
```
bernt-matthias, line 2: Note that you can include the macros also in the main xml file. Having a separate file is really useful if tere a multiple tools using the same macros. But it does not hurt :)

tools/getorganelle/tool_data_table_conf.xml.sample
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<tables>
+    <!-- Location of seed and label data for getorganelle -->
+    <table name="getorganelle" comment_char="#">
+        <columns>value, name, type, path</columns>
```
bernt-matthias, line 5: What is the type column good for? Do we need a version column?

o-william-white: I copied the mitos version, so we can remove the type column if you think it cleans things up?

I added a version column to ```test-data/getorganelle.loc``` but I was not sure if ```tool_data_table_conf.xml.sample``` and ```tool_data_table_conf.xml.test``` needed to be adjusted to reflect this.

tools/getorganelle/macros.xml
bernt-matthias: Up to you.

tools/getorganelle/tool_data_table_conf.xml.sample
bernt-matthias: The important thing is the column definition in the `tool_data_table_conf.xml*` files. This is what Galaxy uses. 

The lines in the `.loc.sample` file  are just comments to the Galaxy admin (which should match the definition in the `tool_data_table_conf.xml` files.

tools/getorganelle/macros.xml
o-william-white: I will leave as it is for for now, but good to know that it's not always necessary

tools/getorganelle/tool_data_table_conf.xml.sample
o-william-white: Thanks for explaining this, much appreciated

o-william-white:
Hello again, 

I upated the main xml to include the macro as suggested. 

I tried to remove the column for type from ```test-data/getorganelle.loc```, ```tool_data_table_conf.xml.sample``` and ```tool_data_table_conf.xml.test``` but when I did, I got the error below with planemo test

```
parameter 'config_dir': requires a value, but no legal values defined
```

Can you suggest how to edit the files to remove the type column, just in case I am missing something obvious?

tools/getorganelle/tool_data_table_conf.xml.sample
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<tables>
+    <!-- Location of seed and label data for getorganelle -->
+    <table name="getorganelle" comment_char="#">
+        <columns>value, name, type, path, db</columns>
```
bernt-matthias, line 5:
```suggestion
        <columns>value, name, path, db-version</columns>
```

Not sure if the tool version is needed .. I removed it. Add it if you think its needed

tools/getorganelle/tool_data_table_conf.xml.test
```diff
@@ -0,0 +1,7 @@
+<tables>
+    <!-- Location of seed and label data for getorganelle test -->
+    <table name="getorganelle" comment_char="#">
+        <columns>value, name, type, path, db</columns>
```
bernt-matthias, line 4:
```suggestion
        <columns>value, name, path, db-version</columns>
```

tools/getorganelle/test-data/getorganelle.loc
```diff
@@ -0,0 +1 @@
+getorganelle_refdata	getorganelle seed and label data	getorganelle	${__HERE__}/0.0.1.minima/	0.0.1
```
bernt-matthias, line 1:
```suggestion
getorganelle_refdata	getorganelle seed and label data	${__HERE__}/0.0.1.minima/	0.0.1
```

tools/getorganelle/getorganelle.loc.sample
```diff
@@ -0,0 +1 @@
+#ID name  getorganelle-version  location  getorganelle-db-version
```
bernt-matthias, line 1:
```suggestion
#ID	name	path	version
```

I think this file needs to go into a subdirectory tool-data/.

o-william-white:
Hi @bernt-matthias, 

Thanks for the suggestions, I made the changes and moved ```getorganelle.loc.sample``` to a directory called ```tools-data```. But when I test the tool using planemo I get the same error below

```galaxy.tool_util.verify.interactor.RunToolException: parameter 'config_dir': requires a value, but no legal values defined```

I assume it is an issue with my data table and locating the reference data.

I refer to the data tables in the command section:
```--config-dir '$config_dir.fields.path'```

and in the inputs section:
```
<param argument="--config-dir" label="Reference data" type="select" help="Contact the administrator of this Galaxy instance if you miss reference data">
            <options from_data_table="getorganelle">
                <filter type="static_value" value="getorganelle" column="2"/>
            </options>
            <validator message="No reference annotation is available for getorganelle" type="no_options" />
        </param>
```
I tried to follow the mitos2 example to get the tool working with data tables but the options may not be relavant to get organelle. Do you have any suggestions on how to resolve the error? 

Best wishes
Ollie

**Comment by bernt-matthias** [comment]
Remove the filter `<filter type="static_value" value="getorganelle" column="2"/>` .. its removing all options that do not have the text `getorganelle` in the 3rd column.

File: tools/getorganelle/get_organelle_from_reads.xml [hunk 31]
```diff
@@ -0,0 +1,190 @@
+<tool id="get_organelle_from_reads" name="Get organelle from reads" version="@TOOL_VERSION@">
+    <macros>
+        <token name="@TOOL_VERSION@">1.7.6.1</token>
+    </macros>    
+    <requirements>
```
bernt-matthias, line 5:
```suggestion
    <xrefs>
        <xref type="bio.tools">getorganelle</xref>
    </xrefs>
    <requirements>
```

o-william-white:
> Remove the filter <filter type="static_value" value="getorganelle" column="2"/> .. its removing all options that do not have the text getorganelle in the 3rd column.

Thanks @bernt-matthias this fixed the issue, I accepted all suggested changes. 

> Actually it should be possible to edit the .loc file and entries should show up. During startup the log should show how many entries are loaded from each data table.

When I load the tool in my local galaxy I still get the message that no reference data can be found. I am not sure how to edit the loc file to do this. I tried adding a file called ```tool-data/getorganelle.loc``` which looked like this:
```
getorganelle_refdata    getorganelle seed and label data        /home/owhite/.GetOrganelle/     0.0.1
```

Where ```/home/owhite/.GetOrganelle/```` is the path to my locally installed get organelle database. Can you suggest where I might be going wrong? 

In general, are there any other edits that need to be made to the getorganelle tool before I start to focus on the data manager?

Cheers
Ollie

bernt-matthias:
Did you restart after editing the loc file? 

Can you execute `find tool-data -name "getorganelle.loc"` .. I guess the loc file is just at a different location. If I understand it correctly each tool version has its own loc file .. and the contents are merged during startup of Galaxy.

o-william-white:
Yes I restart galaxy each time I edit the files. 

When I run ```find tool-data -name "getorganelle.loc"``` from the same directory as the getorganelle xml, I get the following output ```tool-data/getorganelle.loc```

Note that for my local galaxy, I have the xml files in a separate folder within my tools directory. Below is the section block from my ```config/tool_conf.xml``` file with custom tools that I have worked on:
```
<section name="MyTools" id="mTools">
    <tool file="myTools/at_content.xml" />
    <tool file="myTools/split_fasta.xml"/>
    <tool file="myTools/get_organelle_from_reads.xml" />
    <tool file="myTools/count_kmers.xml" />
  </section>
```

tools/getorganelle/get_organelle_from_reads.xml
bernt-matthias: I guess a conditional should work here.  Just put this very select into the conditional. The only when tag needed will be for anonymous (move the two parameters here .. I assume that they are only useful for this case?). The other when tags need to be listed, but can be empty, i.e. `</when value="...">`

bernt-matthias (commented):
Apart from the comment regarding `anonymous` this seems ready from my side.

o-william-white:
Hi @bernt-matthias 

Thanks for the update.

> The only when tag needed will be for anonymous (move the two parameters here .. I assume that they are only useful for this case?).

The ```-s``` and ```--genes``` options that are required when ```-F anonym``` are also useful when this parameter is not selected. So I thought it might be could to keep these variables separate.

Is it possible/sensible to have the same parameters being set a two different place in an xml?

As it stands, if someone were to set  ```-F anonym``` without the ```-s``` and ```--genes``` options, there is a good error meesage to catch this.

```
get_organelle_from_reads.py -1 Arabidopsis_simulated.1.fq.gz -2 Arabidopsis_simulated.2.fq.gz -o test -F ano                                nym

############################################################################
ERROR: "-s" and "--genes" must be specified when "-F anonym"!


```


The easiest thing to do would be to ignore the anonym setting. 

Best wishes
Ollie

bernt-matthias:
So the two parameters are mandatory in the anonymous case and optional otherwise?

o-william-white:
> So the two parameters are mandatory in the anonymous case and optional otherwise?

Yes exactly

bernt-matthias:
Then a conditional might be the best solution https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-inputs-conditional

o-william-white:
Hi @bernt-matthias 

> Then a conditional might be the best solution https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-inputs-conditional

Thanks again for the feedback, I added a conditional, so when a user selects ```anonym```, the user must specify the ```-s``` and ```--genes``` parameters. 

I tested the tool on my local galaxy and thought there might be one potential issue, if the user selects the ```anonym``` option and advanced parameters, the user could potentiially set the specify the ```-s``` and ```--genes``` parameters twice. 

Do you think this is okay? Or is it necessary to have another conditional? 

Best wishes
Ollie

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,212 @@
...
+            <when value="fungus_nr">
+            </when>
+            <when value="animal_mt">
+            </when>
+            <when value="other_mt">
+            </when>
+            <when value="anonym">
+                <param type="data" argument="-s" format="fasta" optional="false" label="Seed sequence(s)" help="Fasta file to use as initial seed. A seed sequence in GetOrganelle is only used for identifying initial organelle reads."/>
+                <param type="data" argument="--genes" format="fasta" optional="false" label="Gene sequence(s)" help="Fasta file containing protein coding genes and ribosomal RNAs extracted from a reference genome that you want to assemble."/>
+            </when>
```
bernt-matthias, line 134: So, if the user selects `anonym` then `-s` and `--genes` are mandatory and otherwise optional? 

Then you can have a macro:

```
<xml name="seed_and_genes" tokens="optional">
    <param type="data" argument="-s" format="fasta" optional="@OPTIONAL@" label="Seed sequence(s)" help="Fasta file to use as initial seed. A seed sequence in GetOrganelle is only used for identifying initial organelle reads."/>
     <param type="data" argument="--genes" format="fasta" optional="@OPTIONAL@" label="Gene sequence(s)" help="Fasta file containing protein coding genes and ribosomal RNAs extracted from a reference genome that you want to assemble."/>
</xml>
```

and use it here as follows:

```suggestion
            <when value="embplant_pt">
                <expand macro="seed_and_genes" optional="true"/>
            </when>
            <when value="embplant_mt">
                <expand macro="seed_and_genes" optional="true"/>
            </when>
            <when value="embplant_nr">
                <expand macro="seed_and_genes" optional="true"/>
            </when>
            <when value="fungus_mt">
                <expand macro="seed_and_genes" optional="true"/>
            </when>
            <when value="fungus_nr">
                <expand macro="seed_and_genes" optional="true"/>
            </when>
            <when value="animal_mt">
                <expand macro="seed_and_genes" optional="true"/>
            </when>
            <when value="other_mt">
                <expand macro="seed_and_genes" optional="true"/>
            </when>
            <when value="anonym">
                <expand macro="seed_and_genes" optional="false"/>
            </when>
```

Hope I did not made a typo here .. code is from the top of my head. Docs on Galaxy xml macros is here https://planemo.readthedocs.io/en/latest/writing_advanced.html#macros-reusable-elements

o-william-white: > So, if the user selects anonym then -s and --genes are mandatory and otherwise optional?

Exactly, and thank you for the suggested code above, I think it looks much better this way. 

Are ther any other issues that I might need to address, or can I move onto the data manager? 

I would also like to write an .xml wrapper for one of the supporting scripts to get annotated regions from an genbank file in the correct format for get organelle https://github.com/Kinggerm/PersonalUtilities/blob/master/get_annotated_regions_from_gb.py But this can be a separate PR. 

Best wishes
Ollie

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,226 @@
...
+        #if str($fastq_input.fastq_input_selector) == "paired_collection":
+            -1 '${$in1_name1}' -2 '${$in2_name}'
+        #end if
+        -o results_directory
+        #if str($organelle_type_input.F) == "anonym":
+            -F '$organelle_type_input.F'
+            -s '$organelle_type_input.s'
+            --genes '$organelle_type_input.genes'
+        #else:
+            -F '$organelle_type_input.F'
```
bernt-matthias, line 64: In this branch you need to specify `-s` and `--genes` if they are given by the user: 

```suggestion
            -F '$organelle_type_input.F'
            #if $organelle_type_input.s
                -s '$organelle_type_input.s'
            #end if
            #if $organelle_type_input.genes
                -s '$organelle_type_input.genes'
            #end if
```

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,226 @@
...
+        </data>
+        <data format="sam" name="seed_sam" from_work_dir="results_directory/seed/*.sam" label='${tool.name} on ${on_string}: Seed sam'>
+            <filter>settings['advanced'] == 'advanced' and "seed_sam" in settings['out']</filter>
+        </data>
+        <data format="txt" name="spades_log" from_work_dir="results_directory/extended_spades/spades.log" label='${tool.name} on ${on_string}: Spades log'>
+            <filter>settings['advanced'] == 'advanced' and "spades_log" in settings['out']</filter>
+        </data>
+    </outputs>
+    <tests>
+        <test expect_num_outputs="3">
```
bernt-matthias, line 207: A few more tests would be great if you have the time to add them: 

- cover a few more (ideally all parameters)
- cover all outputs

If not, then we may go ahead.

tools/getorganelle/get_organelle_from_reads.xml
bernt-matthias: > I would also like to write an .xml wrapper for one of the supporting scripts to get annotated regions from an genbank file

Sounds good. I would prefer a separate PR.

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: I have added more tests to cover all of the output files generated. This was useful as it highlighted some errors that otherwise would have gone unnoticed.

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: I have added this change, however, now the ```-s``` and ```--genes``` appear by default, rather than appearing only when ```anonym``` is selected. Is this expected? 

I suppose we could have  ```-s``` and ```--genes``` as simple optional paramters, then remove these from the advanced list?

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: I have added more parameters to the .xml file. Unfortunately, I do not have the time to add and test all of the parameters, there are quite a lot, so I have tried to select those which are most likely to be used based on the FAQ https://github.com/Kinggerm/GetOrganelle/wiki/FAQ 

I would quite like to move onto the data manager and https://github.com/Kinggerm/PersonalUtilities/blob/master/get_annotated_regions_from_gb.py as I am hoping to use this for a training course

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -0,0 +1,297 @@
...
+            -F '$organelle_type_input.F'
+            -s '$organelle_type_input.s'
+            --genes '$organelle_type_input.genes'
+        #else:
+            -F '$organelle_type_input.F'
+            #if $organelle_type_input.s
+                -s '$organelle_type_input.s'
+            #end if
+            #if $organelle_type_input.genes
+                -s '$organelle_type_input.genes'
```
bernt-matthias, line 69: This should be `--genes` instead of `-s`

tools/getorganelle/get_organelle_from_reads.xml
bernt-matthias: Yep remove them from the advanced list.

`-s` and `--genes` should now only appear if 

- anonym is selected or 
- another option is selected and files are given

If you like you can verify this using `<assert_command>`

tools/getorganelle/get_organelle_from_reads.xml
bernt-matthias: Thanks for the effort. Lets move on with the DM.

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: Well spotted, thanks!

o-william-white:
I removed ```--genes``` and ```-s``` from the advanced list. They are now optional parameters in the simple list, unless ```anonym``` mode is selected.

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: Thank you for all the feedback, it's been quite the learning curve...

bernt-matthias (approved):

bernt-matthias:
Thanks for the contribution @o-william-white .. since I worked a lot on mitochondria in the past (and still do if my time allows) I'm particularly happy to see organelle related tools added to the Galaxy ecosystem.

o-william-white:
Thanks @bernt-matthias for all the help with this, pleased to have contributed! 

I will likely be in touch again to finish the the data manager and some additional tools that will help with getorganelle.

For the next pull rquest I will remember to create a branch. I think I worked on the master branch for this pull request by mistake. Now this has been mergered, is it best to keep a copy of this repository, or does one normally delete it? I am still new to github! 

Best wishes
Ollie

bernt-matthias:
Can you post the output of `git remote -v`?

o-william-white:
Hi @bernt-matthias 

The output looks like this? 
```
(base) [owhite@gruffalo tools-iuc]$ git remote -v
origin  https://github.com/o-william-white/tools-iuc.git (fetch)
origin  https://github.com/o-william-white/tools-iuc.git (push)
```

I ran this from within the repository downloaded from github. 

With the exception of ```git clone```, I haven't really used the command line options for git, I tend to edit the scripts and upload the files manually to the webpage... I am still learning how to use git properly! 

Best wishes
Ollie

bernt-matthias:
OK. Then you can add an upstream remote as follows: 

```
git remote add upstream https://github.com/galaxyproject/tools-iuc.git
```

then fetch the latest updates

```
git fetch --all
```

and reset your master branch to the upstream 

```
git reset --hard upstream/master
```

In the future you can synchronize the master branch of upstream, local, and your fork (origin) by

```
git checkout master
git pull upstream master
git push origin master
```

To create a new branch, commits and push to your origin (for the next PR)

```
git checkout -b BRANCH_NAME
... add commits
git push --set-upstream origin 
```

The last command will also tell you the URL for opening the PR

Hope I did not forget anything / added typos :)

PR #4486, opened 2022-04-06
EGA Download Client updates

shiltemann:
This PR updates the tool in the following way:
- output the list of files in a dataset as a proper table
- support downloading multiple files at once by providing a table with IDs
- optionally output the log file

The idea is that users can request the table of available files, filter this, and pass it back to the tool to download the files of interest.

FOR CONTRIBUTOR:


tools/pyega3/pyega3.xml
```diff
@@ -18,9 +18,13 @@
 #elif $action.action_type == "list_dataset_files"
     pyega3 -cf '$credentials'
       files '$action.dataset_id'
+    &&
+    echo -e 'File ID\tStatus\tBytes\tCheck sum\tFile name' > '$dataset_file_list' &&
+    cat pyega3_output.log | grep EGAF | cut -d "]" -f 2 | sed -e 's/^ \(.*\)/\1/g' | sed  's/\s\+/\t/g' >> '$dataset_file_list'
```
bernt-matthias, line 23: You can simplify this a bit by using `grep EGAF pyega3_output.log` instead of `cat pyega3_output.log | grep EGAF`.

What is `sed -e 's/^ \(.*\)/\1/g'` doing?

tools/pyega3/pyega3.xml
```diff
@@ -18,9 +18,13 @@
...
 #elif $action.action_type == "download_file"
-    pyega3 -cf '$credentials'
+    pyega3 -c 30 -cf '$credentials'
```
bernt-matthias, line 25: Is using 30 connections a good idea? Maybe this should be admin controllable by using an environment variable?

tools/pyega3/pyega3.xml
```diff
@@ -30,7 +34,28 @@
           --end $action.range.end
         #end if
       #end if
-      --saveto '$downloaded_file'
+    && mv ${action.file_id} downloads && rm -f downloads/*.md5
```
bernt-matthias, line 37: Interesting, there are md5 files. Should those be used to verify the download?

tools/pyega3/pyega3.xml
```diff
@@ -70,18 +96,35 @@
                 <param argument="--end" type="integer" optional="true" min="0" label="End Position" help="0-based, exclusive. Only used if a reference sequence name was specified"/>
                 </section>
             </when>
+            <when value="download_files">
+                <param name="id_table" type="data" format="tabular" label="Table with IDs to download" help="A tabular file where one column contains the set of file IDs. This will output a collection. Please select files that are all the same format (e.g. all BAM or all VCF)."/>
+                <param name="id_column" type="integer" value="1" min="1" label="Column containing the file IDs" help="File Identifiers starting with 'EGAF'. For example: EGAF00001753735" />
```
bernt-matthias, line 101: You can make this a `data_column` parameter.

tools/pyega3/pyega3.xml
```diff
@@ -70,18 +96,35 @@
...
+                <section name="range" title="Request a specific Genomic range? (will be applied to ALL requested files)" expanded="false">
+                <param argument="--reference-name" type="text" optional="true" label="Reference Sequence Name" help="For example 'chr1', '1', or 'chrX'. If unspecified, all data is returned." />
```
bernt-matthias, line 103: This should be indented.

tools/pyega3/pyega3.xml
```diff
@@ -1,6 +1,6 @@
 <tool id="pyega3" name="EGA Download Client" version="@VERSION@+galaxy0" profile="19.09" >
     <macros>
-        <token name="@VERSION@">3.4.0</token>
+        <token name="@VERSION@">4.0.0</token>
```
bernt-matthias, line 3: Can you rename the token to `TOOL_VERSION`? Then it will be covered by the auto updater.

tools/pyega3/pyega3.xml
```diff
@@ -1,6 +1,6 @@
 <tool id="pyega3" name="EGA Download Client" version="@VERSION@+galaxy0" profile="19.09" >
```
bernt-matthias, line 1: Guess we can increase the profile a bit.

tools/pyega3/pyega3.xml
shiltemann: it was the recommended value in the README of pyEGA, but I agree it would be best to make it admin controllable, will do :+1:

tools/pyega3/pyega3.xml
shiltemann: the verification is already done by the program, we don't really need these files on the Galaxy side (and if the checksum fails, the download is restarted)

tools/pyega3/pyega3.xml
shiltemann: oh, neat :)

tools/pyega3/pyega3.xml
shiltemann: what value would you recommend here?

shiltemann:
thanks for the fast review @bernt-matthias! I will update and I also still wanted to add an additional test case or two, so marking this WIP for now

tools/pyega3/pyega3.xml
bernt-matthias: Then something like `\${PYEGA_CONNECTIONS:-30}` ...

tools/pyega3/pyega3.xml
shiltemann: yeah this entire line can be simplified heaps, will do

shiltemann:
@bernt-matthias any insight in how to fix this error I keep getting in test 5? (locally it passes so hard to debug)

```
galaxy.tool_util.verify.interactor.JobOutputsError: Output collection 'downloaded_file_collection': identifier 'ENCFF284YOU.bam' found out of order, expected order of {0: 'ENCFF000VWO.bam', 1: 'ENCFF284YOU.bam'} for the tool generated collection elements ['ENCFF284YOU.bam', 'ENCFF000VWO.bam']
```

bernt-matthias:
I'm guessing that the collection elements come from a nested folder, i.e. they are in subdirectories of the `downloads` folder. If so, then this is a due to a bug in Galaxy that I fixed recently https://github.com/galaxyproject/galaxy/pull/13688 .. but it did not find its way into 22.01 https://github.com/galaxyproject/galaxy/pull/13697

bernt-matthias:
Anything left todo from your side @shiltemann ? Otherwise we can merge.

shiltemann:
nope, good to go from my side! Thanks a lot @bernt-matthias !!

bernt-matthias (approved):

PR #4488, opened 2022-04-07
Eliminate auto selection of Excel filter so tool will work with Pulsar

gregvonkuster:
FOR CONTRIBUTOR:


bgruening (approved):

bgruening:
@gregvonkuster we should not use __app__ in the tools. The proper way would be to enhance the excel metadata in Galaxy ro populate the fields and then use that from the tool. This would also improve UX as it happens in the front end.

gregvonkuster:
Merci! :)

PR #4494, opened 2022-04-11
Updating tools/pangolin from version 3.1.20 to 4.0.5

planemo-autoupdate:

I have updated tools/pangolin from version 3.1.20 to 4.0.4.

### DIFF
--- a/tools/pangolin/pangolin.xml
+++ b/tools/pangolin/pangolin.xml
@@ -11,7 +11,7 @@
     <version_command><![CDATA[pangolin --version]]></version_command>
     <command detect_errors="exit_code"><![CDATA[
         #if str($db.source) == "download"
-            # Pangolin version 4 tries to update from an existing directory
+            ## Pangolin version 4 tries to update from an existing directory
             mkdir datadir &&
             pangolin --update-data --datadir datadir &&
         #else if str($db.source) == "builtin"
--- a/tools/pangolin/pangolin.xml
+++ b/tools/pangolin/pangolin.xml
@@ -50,8 +50,7 @@
                 <option value="pangolearn">pangoLEARN</option>
             </param>
             <when value="usher">
-                <param argument="--use-assignment-cache" type="boolean" truevalue="--add-assignment-cache --use-assignment-cache" falsevalue=""
-                label="Use latest UShER assignment cache"
+                <param argument="--use-assignment-cache" type="boolean" truevalue="--add-assignment-cache --use-assignment-cache" falsevalue="" label="Use latest UShER assignment cache"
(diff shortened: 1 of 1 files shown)


wm75:
That's an important update. I guess @pvanheus or me will take care of it. Whoever has some time first :)

wm75:
Here's my preliminary work on this. Makes the tool run again at least and updates the documentation.
Tests and the optional header line still need to be fixed (maybe more).

wm75:
The test in usher mode is passing now. pangolearn mode (all the failing tests) has some issue, but debugging this from the error alone is prevented by https://github.com/cov-lineages/pangolin/issues/419.
Also cannot inspect local files cause pangolin in pangolearn mode gets killed by my OS OOM protection.

wm75:
After updating the misleading tempdir message has gone away, but no other error message gets revealed :(
After rebooting my own machine, it survives the pangolin run, but all tests are passing then.
Mysterious.

bgruening:
> Job in error state.. tool_id: pangolin, exit_code: 1, stderr: [36mWarning: Ignoring specified datadir /tmp/tmp3pcvfcra/job_working_directory/000/2/working/datadir - could not find __init__.py file to check versions 

That looks like the error.

wm75:
No, that you're getting with the --update-data option (because the datadir gets checked before its populated) and if the data was already up to date (because in that case --update-data is not doing anything, but I checked locally and in both cases things work afterwards (also you're getting the same warnings in usher mode).

wm75:
I got some helpful hints over in https://github.com/cov-lineages/pangolin/issues/433 and https://github.com/cov-lineages/pangolin/issues/395 has more information.
So it seems we cannot test the pangolearn branch right now.
Another consideration might be this: if the pangolearn mode uses that much more memory than usher mode, Galaxy admins would have to give more memory to the tool just because someone might run it not in the default mode. Do we want this?
The alternative would be to split the tool into two.

bgruening:
> Another consideration might be this: if the pangolearn mode uses that much more memory than usher mode, Galaxy admins would have to give more memory to the tool just because someone might run it not in the default mode. Do we want this?

We can control memory requirements based on parameters. Its not super easy, but possible.

bgruening (approved):
Style-wise the outputs are locking strange, but overall this looks good.

I think we need to play with the tool and estimate the memory consumption in this one mode.

wm75:
@bgruening strange because of the conditional actions you mean?
Have never used those myself before but apparently they work just as described in the tool xml syntax docs.

I agree that we should maybe just gather some experience with the new tool. If the pangolearn mode causes problems we can still decide to split it out into a separate tool.

wm75:
If you are talking about the strange control chars on stdout/stderr - those come from https://github.com/cov-lineages/pangolin/blob/f2f30b07059e1c959ad943ce3e4aa4704a65079b/pangolin/utils/custom_logger.py.
If you have any idea how we could control that from outside, that would be great.

wm75:
It's just my old-fashioned preference for being able to read things at an 80 chars width without much line wrapping :)

https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#coding-style is somewhat ambiguous here, but not sold to anything anyway.

PR #4496, opened 2022-04-11
Updating tools/raven from version 1.8.0 to 1.8.1

planemo-autoupdate:

I have updated tools/raven from version 1.8.0 to 1.8.1.

**Project home page:** https://github.com/lbcb-sci/raven/releases

### DIFF
--- a/tools/raven/macros.xml
+++ b/tools/raven/macros.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <macros>
-    <token name="@TOOL_VERSION@">1.8.0</token>
+    <token name="@TOOL_VERSION@">1.8.1</token>
     <token name="@PROFILE@">20.05</token>
     <xml name="requirements">
         <requirements>


bernt-matthias:
New release seems only to concern installation which is not relevant for us. So I think we can skip this release.

abretaud:
Should we close this, or the bot will open a new one if we close this one?

simonbray:
> Should we close this, or the bot will open a new one if we close this one?

It will open a new one if you close it. If you leave it open, the bot will push a new commit for the next release.

abretaud:
Ok, maybe we can set a label for this kind of situation? And have it deleted by the bot on next release?
By the way, I think adding a label to all autoupdate PRs would be great too, PRs proposed by human are a bit "drowned" in the automatic ones

simonbray:
> Ok, maybe we can set a label for this kind of situation? And have it deleted by the bot on next release?

That would be great, that's how Bioconda does it as well.

> By the way, I think adding a label to all autoupdate PRs would be great too, PRs proposed by human are a bit "drowned" in the automatic ones

Yes, I suggested this a while ago, but the bot account doesn't have permissions to add labels.

bgruening:
> Yes, I suggested this a while ago, but the bot account doesn't have permissions to add labels.

This we can fix. We just need to add the bot to this repo I think. @martenson can you do this?

PR #4499, opened 2022-04-11
Fix malt_run code for pulsar

gregvonkuster:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/malt/macros.xml
+++ b/tools/malt/macros.xml
@@ -1,6 +1,6 @@
 <macros>
     <token name="@TOOL_VERSION@">0.5.3</token>
-    <token name="@VERSION_SUFFIX@">0</token>
+    <token name="@VERSION_SUFFIX@">1</token>
     <token name="@PROFILE@">20.09</token>
     <xml name="requirements">
         <requirements>
--- a/tools/malt/malt_run.xml
+++ b/tools/malt/malt_run.xml
@@ -7,9 +7,6 @@
     <command detect_errors="exit_code"><![CDATA[
 #import re
 
-## This will point to a directory.
-#set ref = str($reference.fields.path)
-
 #set input_identifier = re.sub('[^\s\w\-]', '_', str($input.element_identifier))
 ## malt-run uses the file extension to determine the input format.
 #if $input.is_of_type('fasta', 'fasta.gz'):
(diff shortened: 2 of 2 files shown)


bernt-matthias:
> The word 'path' has to be in the parameter template, so I eliminated the unnecessary variable $ref.

I like the simplification. But I also think that the two versions should be equivalent, or am I missing something?

What exactly is the problem that you see on pulsar?

Also "funny": If I execute the tests with conda requirement resolution the tests fail, because there is 1byte size difference. Maybe we can allow for this by using `delta`.

bgruening:
@gregvonkuster you fixed that in Galaxy do you?

gregvonkuster:
> @gregvonkuster you fixed that in Galaxy do you?

No, that fix was for tools that use chromosome lengths (e.g., bedtools), which use the path_len setting.  It was a similar fix for Pulsar though.

bgruening:
But we can close this one isn't it?

gregvonkuster:
> But we can close this one isn't it?

This PR should be merged to make sure the malt_run tool works with the Pulsar job runner.  Thanks!  :)

mvdbeek:
> The word 'path' has to be in the parameter template, so I eliminated the unnecessary variable $ref.

Pulsar receives the templated command, so that comes as a surprise to me. It sounds like you know where that happens though, can you let us know where that is ? 

> This PR should be merged to make sure the malt_run tool works with the Pulsar job runner. Thanks! :)

I strongly disagree, we have to figure out the bugs or else we're playing eternal catch up and that makes it a trial-and-error game to write tools that work on pulsar.

gregvonkuster:
> > The word 'path' has to be in the parameter template, so I eliminated the unnecessary variable $ref.
> 
> Pulsar receives the templated command, so that comes as a surprise to me. It sounds like you know where that happens though, can you let us know where that is ?
> 

I'm just making sure that the word 'path' is in the template to ensure this code works https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/wrappers.py#L54

> > This PR should be merged to make sure the malt_run tool works with the Pulsar job runner. Thanks! :)
> 
> I strongly disagree, we have to figure out the bugs or else we're playing eternal catch up and that makes it a trial-and-error game to write tools that work on pulsar.

Perhaps I've misunderstood at which point in the code the templated command is translated.  I thought that this component of the command line ```--index '$ref'``` would not properly be translated, but if ```'$ref'``` is changed to ```'$reference.fields.path'``` before this code https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/wrappers.py#L54 is reached, I'll close this PR.

PR #4501, opened 2022-04-13
Fix bug in the just updated heatmap2

lldelisle:
I am really sorry. I made a mistake in the R script...
I corrected it here and add more tests.


lldelisle:
@bgruening I am so sorry...
- Is it correct to change tool version or should I keep the same as it is a bug fix?
- Would you mind to update usegalaxy.eu again when it is merged?

PR #4522, opened 2022-04-25
Updating tools/pangolin from version 4.0.5 to
4.1.1

planemo-autoupdate:

I have updated tools/pangolin from version 4.0.5 to 4.0.6.


wm75:
According to the release notes, v4.0.6 would enable use of `fasta.gz` as additional input format. The scorpio update to 0.3.17 is not essential at this point.
The failing test is because of `<has_text_matching expression="B\.1\.1.*\t\d\.\d\t*PUSHER" />`, which expects a float with one decimal, but the output has `0.3333333333333333`. It's a bit weird that the test passed on the previous update to 4.0.5 because the output of the installed tool looks identical to 4.0.6.

wm75:
addressed by #4647
we should still consider supporting fasta.gz as input format if possible though.

wm75:
fasta.gz support added in #4666

PR #4526, opened 2022-04-25
Add the adapter_removal tool

gregvonkuster:
FOR CONTRIBUTOR:


gregvonkuster:
@bgruening or @bernt-matthias  Can this PR be "fast-tracked"?  This tool differs from other adapter trimmers like cutadapt in that it works very well with ancient dna.  As soon as this tool is available for install, I'll be doing significant testing with it, and I'll maintain it.  Thanks so much! :)

tools/adapter_removal/adapter_removal.xml
```diff
@@ -0,0 +1,438 @@
...
+#end if
+#if str($output_options.output_discarded) == 'yes':
+    --discarded '$output_discarded'
+#end if
+    ]]></command>
+    <inputs>
+        <conditional name="input_type_cond">
+            <param name="input_type" type="select" label="Choose the category of the reads files to be analyzed">
+                <option value="single" selected="true">Single dataset containing reads</option>
+                <option value="pair">Dataset pair of forward and reverse reads</option>
```
bernt-matthias, line 114: Would appreciate paired collection input

tools/adapter_removal/adapter_removal.xml
```diff
@@ -0,0 +1,438 @@
...
+            </when>
+            <when value="interleaved">
+                <param name="read1" type="data" format="fasta,fasta.gz,fastqsanger.gz,fastqsanger" label="Interleaved read file"/>
+                <expand macro="interleaved_output_param"/>
+                <expand macro="combined_output_param"/>
+            </when>
+        </conditional>
+        <!-- FASTQ Options -->
+        <section name="fastq_options" title="FASTQ Options" expanded="false">
+            <param argument="--qualitybase" type="select" label="Choose the Phred quality scores encoding used in the input reads">
```
bernt-matthias, line 135: Is it possible to set this parameter automatically based on the datatype of the input?

tools/adapter_removal/adapter_removal.xml
```diff
@@ -0,0 +1,438 @@
...
+                <expand macro="sanitizer"/>
+            </param>
+            <param argument="--adapter_list" type="data" format="tabular" optional="true" label="Tabular file containing adapter sequences" help="The first two columns of each line in the file must correspond to values passed to --adapter1 and --adapter2"/>
+            <param argument="--minadapteroverlap" type="integer" value="0" min="0" label="Trim reads only if the overlap between read and the adapter is at least this number of bases long" help="Ignored unless input files are single-end reads"/>
+            <param argument="--mm" type="float" value="3.0" label="The fraction of mismatches allowed in the aligned region" help="The value is used directly if less than 1, the rate s set to 1 / value if the value is greater than 1"/>
+            <param argument="--shift" type="integer" value="2" min="0" max="93" label="Slip the alignment by this number of bases in the 5' end to allow for missing bases in the 5' end of the read"/>
+            <conditional name="trim5p_cond">
```
bernt-matthias, line 159: Are these conditionals needed?

tools/adapter_removal/adapter_removal.xml
```diff
@@ -0,0 +1,438 @@
...
+            <param argument="--minalignmentlength" type="integer" value="11" min="0" label="Minimum overlap between mate 1 and mate 2 before the reads are collapsed into one when collapsing paired-end reads, or when attempting to identify complete template sequences in single-end mode"/>
+            <param argument="--collapse_deterministic" type="boolean" truevalue="--collapse-deterministic" falsevalue="" checked="false" label="Collapse deterministically?" help="Setting this option also sets --collapse"/>
+            <param argument="--collapse_conservatively" type="boolean" truevalue="--collapse-conservatively" falsevalue="" checked="false" label="Collapse conservatively?" help="Setting this option also sets --collapse"/>
+        </section>
+        <!-- Output Options -->
+        <section name="output_options" title="Output Options" expanded="false">
+            <param name="output_singleton" type="select" label="Output file containing paired reads for which the mate has been discarded?">
```
bernt-matthias, line 208: A single select with 4 options might be better instead of the for selects. Otherwise make them boolean.

tools/adapter_removal/adapter_removal.xml
gregvonkuster: Thanks so much for reviewing this on shirt notice @bernt-matthias !  I considered allowing a paired collection on input, but I'm still not clear on when this is desired and when it is not due to the affect input collections have on Galaxy job splitting.  I thought this was a case where we'd want job splitting across the inputs, which would not occur with an input collection.  Can you clarify this for me?  I'll certainly add the collection option on input.

tools/adapter_removal/adapter_removal.xml
gregvonkuster: Hmm...possibly, but I'm not sure.  We'd have to know which Illumina version the inputs are, which isn't defined by the fastqsanger datatype.  Is there another way to determine this?

tools/adapter_removal/adapter_removal.xml
gregvonkuster: I think so, because we probably don't want the associated parameters displayed.
```
                <when value="yes">
                    <param name="trim5p_mate1" type="integer" value="0" min="0" label="Value for trimming the 5' of mate1 reads"/>
                    <param name="trim5p_mate2" type="integer" value="0" min="0" label="Value for trimming the 5' of mate1 reads"/>
                </when>
```
But I could be overlooking a better approach here.

tools/adapter_removal/adapter_removal.xml
gregvonkuster: I know that the associated filters don't work with booleans, and I'm not sure if they would work with a multi-select.
```
        <data name="output_singleton_truncated" format="fastqsanger" label="${tool.name} on ${on_string} (singleton truncated)">
            <filter>output_options['output_singleton'] == 'yes'</filter>
        </data>
        <data name="output_collapsed" format="fastqsanger" label="${tool.name} on ${on_string} (collapsed)">
            <filter>output_options['output_collapsed'] == 'yes'</filter>
        </data>
        <data name="output_collapsed_truncated" format="fastqsanger" label="${tool.name} on ${on_string} (collapsed truncated)">
            <filter>output_options['output_collapsed_truncated'] == 'yes'</filter>
        </data>
        <data name="output_discarded" format="fastqsanger" label="${tool.name} on ${on_string} (discarded)">
            <filter>output_options['output_discarded'] == 'yes'</filter>
        </data>
```

gregvonkuster: Ah, I see that multi-selects do work with filters - I'll change this to be a multi-select!

tools/adapter_removal/adapter_removal.xml
bernt-matthias: You need to distinguish 

1. `paired` and
2.  `list:paired`. 

If you input a list of pairs to 1. then a separate job is started for each pair (which is what you want here). In the second case all pairs would be processed jointly in a single job.

gregvonkuster: Yes, I understand the difference between ```paired``` and ```list:paired```.  But I don't think we'd want a large collection run as a single job for this tool, would we?  That's why I left out the ```list:paired``` option - it would create a single very long running job, but the more optimal solution would be to split the job across the inputs.

bernt-matthias: Yes, `list:paired` is not what we want here. 

But `paired` is exactly what we want. Job splitting will happen automatically if you give a list of pairs to such an input.

tools/adapter_removal/adapter_removal.xml
gregvonkuster: This is now a single multi-select param.  Thanks!

tools/adapter_removal/adapter_removal.xml
```diff
@@ -0,0 +1,438 @@
...
+            </when>
+        </conditional>
+        <!-- FASTQ Options -->
+        <section name="fastq_options" title="FASTQ Options" expanded="false">
+            <param argument="--qualitybase" type="select" label="Choose the Phred quality scores encoding used in the input reads">
+                <option value="33" selected="true">Phred+33 (Illumina 1.8+)</option>
+                <option value="64">Phred+64 (Illumina 1.3+ and 1.5+)</option>
+                <option value="solexa">Solexa encoded scores</option>
+            </param>
+            <param argument="--qualitybase_output" type="select" label="Choose the base of the quality score for output reads">
```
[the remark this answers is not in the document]
gregvonkuster: What is the "best practice" here?  For inputs, I've typically just used fastqsanger, but should tools support all fastq variants?  And what about outputs - should tools support all fastq variants on output as well?

For this particular tool, the default value for --qualitybase-output is the value given for --qualitybase, so handling that is fairly simple.

But using your proposed ```format_source``` poses problems if the inputs are compressed.  I don't think that the ```change_format``` feature works with ```format_source```.  Should I just restrict output to fastqsanger, no matter the fastq variant on input?

tools/adapter_removal/adapter_removal.xml
bernt-matthias: An unexpanded section might be an option.

gregvonkuster: These 2 conditionals are already contained within the unexpanded fastq_trimming_options section.  We need to have these 4 trimming parameter settings under user control so that trimming is not automatic, and these conditionals enable this.  I'm not seeing a cleaner approach for enabling this user control, although I could certainly be missing something here.  I'm happy to change this if you have a better approach.  Im wondering why these conditionals are undesirable though.

tools/adapter_removal/adapter_removal.xml
```diff
@@ -0,0 +1,452 @@
...
+            <filter>input_type_cond['input_type'] in ['pair', 'paired'] and input_type_cond['interleaved_output'] == 'no'</filter>
+        </data>
+        <data name="output_reverse_truncated" format="fastqsanger" label="${tool.name} on ${on_string} (reverse truncated)">
+            <filter>input_type_cond['input_type'] in ['pair', 'paired'] and input_type_cond['interleaved_output'] == 'no'</filter>
+        </data>
+        <data name="output_interleaved_truncated" format="fastqsanger" label="${tool.name} on ${on_string} (interleaved truncated)">
+            <filter>input_type_cond['input_type'] in ['pair', 'paired'] and input_type_cond['interleaved_output'] == 'yes'</filter>
+        </data>
+        <data name="output_singleton_truncated" format="fastqsanger" label="${tool.name} on ${on_string} (singleton truncated)">
+            <filter>'output_singleton' in output_select</filter>
```
bernt-matthias, line 230: You will need: 
```suggestion
            <filter>output_select and 'output_singleton' in output_select</filter>
```

Since `output_select` is optional and therefore the variable can be `None`

Maybe also add a test where no option is selected.

gregvonkuster: I've made this change, but I don't think it is necessary.  The output_select parameter is not optional, so can never be the value None - by default, the value is the string "none".  So current tests cover the scenario where no option is selected and the default value of "none" is used.

gregvonkuster:
@bernt-matthias or @bgruening can this be merged?  Thanks!  :)

bgruening (approved):
Good from my side. @bernt-matthias ?

tools/adapter_removal/adapter_removal.xml
bernt-matthias: Sorry, my mistake. If you like you can revert https://github.com/galaxyproject/tools-iuc/pull/4526/commits/9217dbfc37f123ac79f09fd61c4d6b040e8e1c1a and then we merge.

bernt-matthias (approved):

tools/adapter_removal/adapter_removal.xml
gregvonkuster: Reverted,  thanks!

gregvonkuster:
@bernt-matthias Sorry to keep bugging you on this, I think it's ready.  Thanks!

PR #4539, opened 2022-05-03
FastQC: detects compression directly instead of depending on user input

bwlang:
Users sometimes make mistakes when specifying the type of input files (fastqsanger instead of fastqsanger.gz, etc.)
Fastqc is finicky about this - requiring that gzipped files have a .gz extension.

This switches out the user-specified file format detection for a more reliable direct detection using the file command.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/fastqc/rgFastQC.xml
+++ b/tools/fastqc/rgFastQC.xml
@@ -25,10 +25,10 @@
         #else
             #set format = 'fastq'
         #end if
-        detected_suffix=`file --mime-type "${input_file}"
+        detected_suffix=\$(file --mime-type "${input_file}"
                          | grep -e "bzip2" -e "gzip"
                          | sed "s/.*application\///"
-                         | sed "s/.*bzip2/.bz2/" | sed "s/.*gzip/.gz/"` &&
+                         | sed "s/.*bzip2/.bz2/" | sed "s/.*gzip/.gz/") &&
         ln -s '${input_file}' '${input_name}'\$detected_suffix &&
         mkdir -p '${html_file.files_path}' &&
         fastqc
--- a/tools/fastqc/rgFastQC.xml
+++ b/tools/fastqc/rgFastQC.xml
@@ -1,4 +1,4 @@
-<tool id="fastqc" name="FastQC" version="0.73+galaxy0">
+<tool id="fastqc" name="FastQC" version="0.73+galaxy1">
     <description>Read Quality reports</description>
     <xrefs>
         <xref type="bio.tools">fastqc</xref>
@@ -18,14 +18,6 @@
         #import re
         #set input_name = re.sub('[^\w\-\s]', '_', str($input_file.element_identifier))
 
-        #if $input_file.ext.endswith('.gz'):
-            #set input_file_sl = $input_name + '.gz'
-        #elif $input_file.ext.endswith('.bz2'):
-            #set input_file_sl = $input_name + '.bz2'
-        #else
-            #set input_file_sl = $input_name
-        #end if
-
         #if 'bam' in $input_file.ext:
             #set format = 'bam'
         #elif 'sam' in $input_file.ext:
@@ -33,8 +25,11 @@
         #else
             #set format = 'fastq'
         #end if
-
-        ln -s '${input_file}' '${input_file_sl}' &&
+        detected_suffix=`file --mime-type "${input_file}"
+                         | grep -e "bzip2" -e "gzip"
+                         | sed "s/.*application\///"
+                         | sed "s/.*bzip2/.bz2/" | sed "s/.*gzip/.gz/"` &&
+        ln -s '${input_file}' '${input_name}'\$detected_suffix &&
         mkdir -p '${html_file.files_path}' &&
         fastqc
             --outdir '${html_file.files_path}'
@@ -58,7 +53,7 @@
             $nogroup
             --kmers $kmers
             -f '${format}'
-            '${input_file_sl}'
+            '${input_name}'\$detected_suffix
 
         && cp '${html_file.files_path}'/*/fastqc_data.txt output.txt
         && cp '${html_file.files_path}'/*\.html output.html


tools/fastqc/rgFastQC.xml
```diff
@@ -18,23 +18,18 @@
...
         #else
             #set format = 'fastq'
         #end if
-
-        ln -s '${input_file}' '${input_file_sl}' &&
+        detected_suffix=`file --mime-type "${input_file}"
+                         | grep -e "bzip2" -e "gzip"
+                         | sed "s/.*application\///"
+                         | sed "s/.*bzip2/.bz2/" | sed "s/.*gzip/.gz/"` &&
+        ln -s '${input_file}' '${input_name}'\$detected_suffix &&
```
bernt-matthias, line 32:
```suggestion
        detected_suffix=\$(file --mime-type "$input_file"
                         | grep -e "bzip2" -e "gzip"
                         | sed "s@.*application/@@"
                         | sed "s/.*bzip2/.bz2/" | sed "s/.*gzip/.gz/") &&
```

bernt-matthias (commented):
One might also think about just raising a helpful error message in case of a mismatch between `detected_suffix` and the datatype of the input, because the dataset with the wrong type will likely also cause problems in downstream analyses.

bwlang:
Most processing tools (seqtk, fastp, cutadapt, flexbar, etc.) and aligners handle gzip transparently in my experience...  I don't see trouble with other tools in my environment.

bernt-matthias:
Could the `apply rule` tool be used to change datatypes? Otherwise I would suggest to open an issue at the Galaxy repo that a tool/method is needed for this task.

mvdbeek:
> Most processing tools (seqtk, fastp, cutadapt, flexbar, etc.) and aligners handle gzip transparently in my experience... I don't see trouble with other tools in my environment.
cutadapt definitely fails if you provide them a fastq disguised as fastq.gz and reverse. Most tools just look at the extension and don't do inference (which I think is the most robust and predictable thing to do).

fastp actually illustrates perfectly why we should absolutely not do this:
```
gzip -c ~/src/galaxy/test-data/1.fastqsanger > 1.fastq
 cp 1.fastq 1.fastq.gz
```
So now I have 2 compressed fastq files, one with the right extension, one with the wrong extension.
```
 ~/src/planemo   planemo_gravity ● ? ⍟55  fastp -i 1.fastq.gz                                                                                                                                    ✔  10199  12:56:41  .venv  (fastp)
Detecting adapter sequence for read1...
No adapter detected for read1

Read1 before filtering:
total reads: 2
total bases: 68
Q20 bases: 42(61.7647%)
Q30 bases: 16(23.5294%)

Read1 after filtering:
total reads: 2
total bases: 68
Q20 bases: 42(61.7647%)
Q30 bases: 16(23.5294%)

Filtering result:
reads passed filter: 2
reads failed due to low quality: 0
reads failed due to too many N: 0
reads failed due to too short: 0
reads with adapter trimmed: 0
bases trimmed due to adapters: 0

Duplication rate (may be overestimated since this is SE data): 0%

JSON report: fastp.json
HTML report: fastp.html

fastp -i 1.fastq.gz
fastp v0.23.2, time used: 1 seconds
```

```
 ~/src/planemo   planemo_gravity ● ? ⍟55  fastp -i 1.fastq                                                                                                                                       ✔  10200  12:58:49  .venv  (fastp)
Detecting adapter sequence for read1...
No adapter detected for read1

Read1 before filtering:
total reads: 0
total bases: 0
Q20 bases: 0(nan%)
Q30 bases: 0(nan%)

Read1 after filtering:
total reads: 0
total bases: 0
Q20 bases: 0(nan%)
Q30 bases: 0(nan%)

Filtering result:
reads passed filter: 0
reads failed due to low quality: 0
reads failed due to too many N: 0
reads failed due to too short: 0
reads with adapter trimmed: 0
bases trimmed due to adapters: 0

Duplication rate (may be overestimated since this is SE data): 0%

JSON report: fastp.json
HTML report: fastp.html

fastp -i 1.fastq
fastp v0.23.2, time used: 1 seconds
```
The result is just wrong for 1.fastq, it doesn't even quite with a non-zero exit code. That can have really bad consequences on scientific results. I'm -1 on this.

Assunta is working on adding bulk changing of datatypes in Galaxy, this should hopefully appear in the next Galaxy release. Until then you could use https://galaxyworks.io/blog/change-datatype-on-many-datasets if you have a collection in your history. If that needs to go back into a library you can import the collection to your library.

bwlang:
That seems like a nasty fastp bug (opened that here https://github.com/OpenGene/fastp/issues/410)... I think it'e because they chose to use file extension detection instead of inspecting the first few bytes.  I think that's evidence that inspection is more reliable.  

Sounds like you want the metadata to be correct... maybe that's the best path here- moving this responsibility to galaxy to reliably sanity check the input type for these common cases.  Unfortunately we still have to do some encoding in the filenames to accommodate this tool. So which is worse - sniffing the file  or trusting the possibly faulty metadata. Code seem similarly complex to me (but maybe i'm biased ;P) .

bwlang:
looking forward to the bulk operations on data sets - seems a nice addition!

bernt-matthias:
Seems that we can close this and also no issue at  the main repo is needed, or?

Maybe we could need a sanity check for zipped VS unzipped datatypes which might be easy to mix up.

bwlang:
I think fa.gz/fa metadata mixup is a fairly common issue for users, and it's trivial to check for gzip. I'm not sure where that would fit in the main galaxy repo. I think @mvdbeek  doesn't like this approach - but maybe he's convinced by the fastp example that direct detection is superior to extension inspection (which is a form of user format specification)? What's that thumbs up voting for - "direct detection is better" or "fix metadata in galaxy code"?    If we fix it upstream, this is not needed - but I don't see harm in merging this even if the metadata is eventually corrected upstream.

bernt-matthias:
Just asking, this is about users choosing the wrong data type and not relying on the datatype autodetection during upload, or? 

Just checked: the autodetection work independent of the extension for `fastq` and `fastq.gz`.

bwlang:
@bernt-matthias 
I'm not sure how they got assigned the wrong type, maybe as you suggest - user specifying fastqsanger instead of fastqsanger.gz

mvdbeek:
> I'm not sure where that would fit in the main galaxy repo. I think @mvdbeek doesn't like this approach

We should deal with this on the framework level in Galaxy, I don't think we want (or can, in the case of non-iuc/devteam/bgruening/... tools) add this or similars hacks to all tools that have auto-compressed datatypes, which I think we'd have to do if we merge this PR. The fastp example shows how dangerous it is to work around wrong datatypes (in a pipeline fastq->fastqc->fastp all may look fine and green, so you'd have to be very careful in interpreting your analysis results / lack of analysis results), and I would assume there are many more bioinformatics tools out there that fail in similar ways. It's better to fail the fastqc step. As a user at least you can infer that something was wrong with the direct input dataset. This is maybe not such a huge concern for fastqc, but we should be consistent. If some downstream tool fails because the upstream tool worked around the wrong datatype but the downstream tool didn't you'd need to be pretty savvy to figure out what happened somewhere in the middle. I hope that made it a little clearer ?

Additional reasons why we shouldn't do this:
- I think it's important that we try to limit the stuff in the command line to what is necessary to run the tool. The command line should remain as readable as possible
- If we discover something wrong with the approach we'll have to copy-paste the fix to every tool (we can make it a macro, but we'd still have to release a new version ....)
- The old versions of fixed tools are still going to be around for a long time

So what can we do ?
I guess there is prevention, warning and fixing.

I have a couple of questions before we can pick how we want to prevent this situation:

> I think fa.gz/fa metadata mixup is a fairly common issue for users,

It'd be great to know how this happens. If you upload a fastq.gz file and you don't set a datatype you get the proper fastq.gz datatype. If you select fastq then the auto-decompress setting kicks in and unpacks the archive. The one thing that is wrong is if you select fastq.gz but you uploaded fastq ... is that a common issue ? Is there something we can do to make it clear that the upload process is not converting datasets ? Do we need to discuss whether that would actually be an option (Not really loving the idea, but if it could help we could work out the pro's and con's) ?

Or is this an issue where users change the auto-sniffed datatype, thinking that changing the datatype is the same as converting a dataset ? In that case, do we need to improve the help text ? Prevent doing that if we're certain we're switching from compressed to uncompressed or vice versa and that that would be "wrong"?

One idea we've discussed is that we could do a quick check of the first few files (could be client-side for local files, would have to be server side for ftp/urls/uris) and then return an error in the user interface if you're going to do something that won't work / isn't correct.

In terms of warning:
- We could display a warning saying "this dataset doesn't actually look like what you selected. Click here to fix this / learn more". I could have sworn we already do that, but maybe that was only working with the upload1 tool and not the __DATA__FETCH__ tool.
- If a job fails we can do some sort of "common problems" diagnosis. We already have something like this that we can extend

In terms of fixing wrong datatypes:
- Bulk changing dataypes is hopefully going to be part of 22.05

bwlang:
I think this is better than what we have now and I don't propose that we make all tools compression-agnostic, but I don't want to suck up more energy on a not very important topic ;).

PR #4516, opened 2022-04-20
recentrifuge new tool

pimarin:
I made a new pull request due to some errors in the previous closed PR [4506](https://github.com/galaxyproject/tools-iuc/pull/4506).
I also integrated some remarks :
- [X] [discussion_r851772801](https://github.com/galaxyproject/tools-iuc/pull/4506#discussion_r851772801)
- [X] [discussion_r851773718](https://github.com/galaxyproject/tools-iuc/pull/4506#discussion_r851773718)
- [X] [discussion_r851773800](https://github.com/galaxyproject/tools-iuc/pull/4506#discussion_r851773800)
- [X] [discussion_r851773910](https://github.com/galaxyproject/tools-iuc/pull/4506#discussion_r851773910)
- [X] [discussion_r851774335](https://github.com/galaxyproject/tools-iuc/pull/4506#discussion_r851774335)

For other comments : 
- [discussion_r851774516](https://github.com/galaxyproject/tools-iuc/pull/4506#discussion_r851774516)
    - I added when available most of the options made by the developers.
- [discussion_r851774580](https://github.com/galaxyproject/tools-iuc/pull/4506#discussion_r851774580)
    - It seem to be in development and this option isn't clear for me.
- [discussion_r851774825](https://github.com/galaxyproject/tools-iuc/pull/4506#discussion_r851774825)
    - To be the most open to previous taxonomic tools it could be better to have a more general option.

FOR CONTRIBUTOR:
* [X] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [X] - License permits unrestricted use (educational + commercial)
* [X] - This PR adds a new tool or tool collection
* [ ] - This PR updates an existing tool or tool collection
* [ ] - This PR does something else (explain below)

PR #4540, opened 2022-05-04
PR for recentrifuge tool

pimarin:
Follow the last PR [4516](https://github.com/galaxyproject/tools-iuc/pull/4516) accientally closed by forcing a bad push
FOR CONTRIBUTOR:


pimarin:
I'm not sure about the good use of ncbi_taxonomy into the code, I copy the code made for the tool `taxonomy_filter_refseq`

pimarin:
Other suggestions ?a

pimarin (commented):
Recentrifuge tools finished

tools/recentrifuge/test-data/.~lock.result_rcf.data.csv#
```diff
@@ -0,0 +1 @@
+,pierre,pmarin-hp,10.03.2022 15:37,file:///home/pierre/.config/libreoffice/4;
```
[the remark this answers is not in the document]
gallardoalba: Great, thank you so much!

bgruening:
@pimarin we also tried to reduce the datasets or compress it - during that we actually wondered if its actually needed to provide the databese via this history. This seems to be brittle, filenames need to have the correct names etc ... in Blast we created an own composite datatype for this and a tool that creates such datatypes. So you think it makes sense to stip out this functionality for the time being here and work towards a tool that can create those databases?

pimarin:
@bgruening , thank you for the suggestion. I agree but I don't understand how to do ? Can you give me link to the blast tool ? I searched into the BLAST IUC tools but I don't see where is the tool to create the composite database.

bgruening:
@pimarin this PR as it contains one error that you can see here: https://github.com/galaxyproject/tools-iuc/runs/6858745039?check_suite_focus=true#step:8:1293

You can reproduce it with `planemo test --biocontainers ...`

I have a few stupid questions, sorry to be naiive. You are using the NCBI taxonomy table here, is that correct?
If so will the data manager work https://github.com/galaxyproject/tools-iuc/tree/master/data_managers/data_manager_fetch_ncbi_taxonomy ? For example this here is a taxonomy table used by kraken: https://github.com/galaxyproject/tools-iuc/tree/master/tool_collections/kraken/test-data/test_db/taxonomy

Can you use the same test data?

The Blast Datatype is here: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/datatypes/blast.py#L182
But this might be some work, maybe we just comment out the option (to supply the DB from the history) from this PR and get it in?

pimarin:
@bgruening, I used your suggestion and I successfully reduced the database size using the same as for kraken tools. And I fixed the problem of a test updating test outputs related to the new database

pimarin:
Oky I fixed the test error ... And all test pass with planemo

pimarin:
I don't understand why the test file in the action ?

tools/recentrifuge/recentrifuge.xml
```diff
@@ -0,0 +1,388 @@
...
+                </conditional>
+            </section>
+            <section name="output_option">
+                <param name="extra" value="MULTICSV"/>
+                <param name="nohtml" value="true"/>
+            </section>
+            <section name="advanced_option">
+                <param name="controls" value="0"/>
+                <param name="scoring" value="NORMA"/>
+                <param name="minscore" value="specify_minscore"/>
```
gallardoalba, line 272: There's not "minscore" parameter defined I think.

tools/recentrifuge/recentrifuge.xml
```diff
@@ -0,0 +1,388 @@
...
+                <param name="minscore_value" value="0"/>
+                <param name="avoidcross" value="true"/>
+            </section>
+            <section name="more_advanced_option">
+                <param name="minscore_select" value="specify_minscore"/>
```
gallardoalba, line 277: Also this parameter is not defined I think.

gallardoalba:
The error of the second test is related to the MULTICSV output; it triggers an error when using Galaxy inside Docker (used by default when running the tests in Github). I'm investigating the problem, but I suggest removing the MULTICSV output format (according to the[ recentrifuge source code](https://github.com/khyox/recentrifuge/blob/master/rcf#L651) the CSV output provides similar information).

pimarin:
Hi @gallardoalba, 
Thank you I will remove this option. But how do you find the problem about multicsv and docker ? Is it explicit in the output of github action ?

pimarin:
I pushed a new update but I'm wondering how I can test other output option, because `FULL` and `DYNOMICS` give an excel file...

gallardoalba:
> I pushed a new update but I'm wondering how I can test other output option, because `FULL` and `DYNOMICS` give an excel file...
Hi @pimarin, I tested both FULL and DYNOMICS, and both work fine; in order to identify the problem, I just ran Galaxy under Docker by running _planemo s --docker_. Regards

gallardoalba (approved):

gallardoalba:
Great job @pimarin !

pimarin:
Thank to your help and time spent to help me.

PR #4544, opened 2022-05-06
Modify anndata modify_loom so it doesn't write to the input

natefoo:
As it is currently, it modifies the input, then copies the input to the output. This just moves the copy up front: copy the input to the output, then modify the output.

I could update the galaxyN version, but that would affect all the tools in the suite, right?

FOR CONTRIBUTOR:

### DIFF
--- a/tools/anndata/export.xml
+++ b/tools/anndata/export.xml
@@ -110,8 +110,8 @@ adata.write_csvs('.', sep="\t", skip_data = False)
     </tests>
     <help><![CDATA[
 This tool exports an AnnData dataset to a Loom file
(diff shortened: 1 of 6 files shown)


bgruening:
> I could update the galaxyN version, but that would affect all the tools in the suite, right?

Yes, but you can for that specific tool change version="@VERSION@+@GALAXY_VERSION@"> to version="@VERSION@+galaxy1">

natefoo:
Ok, I wasn't sure if that would be annoying for the next person to do a major upgrade. I'll update it that way, thanks!

bgruening (approved):

PR #4549, opened 2022-05-09
allow user to set sample ID in stacks2_ustacks

TomHarrop:
FOR CONTRIBUTOR:

---

Hi,

Some Galaxy Australia users want to combine the output of multiple `ustacks` runs at the `cstacks` (catalog building) stage. The current wrapper sets the sample ID for the first sample to 1 and then increments by 1 for each sample. This means sample IDs collide across `ustacks` runs, which causes an error in `cstacks`.

I've added a new parameter (and a test) to `ustacks` to allow the user to set the ID for the first sample. I couldn't think of a lower-impact way of doing it.

Unrelated: one of the existing ustacks tests is failing. I think this might be because the locus ID in the ustacks output changes between runs. I had to modify the tests from diff to `assert_contents`.

Gruß!
Tom

### DIFF
--- a/tools/stacks2/stacks_ustacks.xml
+++ b/tools/stacks2/stacks_ustacks.xml
@@ -126,7 +126,11 @@ true
             <!-- 1st test checks for file content allowing differences in the 2 comment lines that contain date and version -->
             <output_collection name="tabs" type="list" count="6">
                 <element name="PopA_01.tags" ftype="tabular" file="ustacks/PopA_01.tags.tsv" lines_diff="4"/>
-                <element name="PopA_01.snps" ftype="tabular" file="ustacks/PopA_01.snps.tsv" lines_diff="4"/>
+                <element name="PopA_01.snps">
+                    <assert_contents>
+                        <has_text text="# ustacks version" />
+                    </assert_contents>
+                </element>
                 <element name="PopA_01.alleles">
                     <assert_contents>
                         <has_text text="# ustacks completed on " />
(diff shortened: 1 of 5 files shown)


TomHarrop:
It looks like one of the other tools has issues - I'm adding some more `<assert_contents>`s.

tools/stacks2/stacks_ustacks.xml
```diff
@@ -255,6 +284,23 @@ true
...
         </test>
+        <!-- test setting i -->
+        <test expect_num_outputs="2">
+            <param name="input_type|input_type_select" value="single"/>
+            <param name="input_type|fqinputs" value="demultiplexed/PopA_01.1.fq,demultiplexed/PopA_02.1.fq" ftype="fastqsanger"/>
+            <param name="add_log" value="yes"/>
+            <param name="i" value="42"/>
+            <output name="output_log" ftype="txt"><assert_contents><has_text text="ustacks is done."/></assert_contents></output>
+            <output_collection name="tabs" count="6">
+                <element name="PopA_01.tags"><assert_contents><has_text text="# ustacks completed on "/></assert_contents></element>
```
bernt-matthias, line 295: Just a suggestion. You may use a macro for `<element name="PopA_01.tags"><assert_contents><has_text text="# ustacks completed on "/></assert_contents></element>` and parametrize the value of the `name` attribute.

TomHarrop: Good idea, I have added that macro and applied it where any check was looking for "stacks completed on".

bernt-matthias:
Ready to merge @TomHarrop ?

TomHarrop:
Yes, thanks @bernt-matthias

bgruening:
Thanks @TomHarrop and @bernt-matthias

PR #4558, opened 2022-05-12
add mcl

bgruening:
adjusted from Nolan, with permission to move to IUC


tools/mcl/mcl.xml
```diff
@@ -0,0 +1,118 @@
...
+    <macros>
+        <token name="@TOOL_VERSION@">14.137</token>
+        <token name="@VERSION_SUFFIX@">0</token>
+    </macros>
+    <edam_topics>
+        <edam_topic>topic_3570</edam_topic>
+    </edam_topics>
+    <edam_operations>
+        <edam_operation>operation_3432</edam_operation>
+    </edam_operations>
```
gallardoalba, line 12:
```suggestion
    <xrefs>
        <xref type="bio.tools">mcl</xref>
    </xrefs>
```

tools/mcl/mcl.xml
```diff
@@ -0,0 +1,118 @@
...
+    </outputs>
+    <tests>
+        <test expect_num_outputs="1">
+            <conditional name="input_type_select">
+                <param name="input_type" value="--abc" />
+                <param name="input" value="mash-dist.tabular" ftype="tabular" />
+            </conditional>
+            <param name="transform" value="mul(-1),add(1),gq(0.96)" />
+            <output name="output" file="mash-dist-cluster.tabular" ftype="tabular" />
+        </test>
```
gallardoalba, line 93: It would be nice to include some additional tests for the remaining options.

tools/mcl/mcl.xml
```diff
@@ -0,0 +1,118 @@
...
+        #else if $input_type_select.input_type == ""
+            -tf '$transform'
+        #end if
+    #end if
+    -o '$output'
+
+    ]]></command>
+    <inputs>
+        <conditional name="input_type_select">
+            <param name="input_type" type="select" argument="--abc" label="Input Type">
```
gallardoalba, line 45:
```suggestion
            <param name="input_type" type="select" label="Input type">
```

gallardoalba: In that case, the argument tag is wrong I think since it depends on the selected option.

tools/mcl/mcl.xml
```diff
@@ -0,0 +1,118 @@
...
+                <option value="--abc" selected="true">Labeled</option>
+                <option value="--sif">SIF</option>
+                <option value="">Matrix</option>
```
gallardoalba, line 48: There's also an etc input format, not sure if necessary.

tools/mcl/mcl.xml
```diff
@@ -0,0 +1,118 @@
...
+            </when>
+            <when value="--sif">
+                <param type="data" format="sif" name="input" argument="input" label="Input" optional="false"/>
+            </when>
+        </conditional>
+        <param name="inflation" argument="-I" type="float" value="4.0" label="Inflation"
+            help="This value is the main handle for affecting cluster granularity. It is usually chosen somewhere in the range [1.2-5.0]. -I 5.0 will tend to result in fine-grained clusterings, and -I 1.2 will tend to result in very coarse grained clusterings. Your mileage will vary depending on the characteristics of your data. That is why it is a good idea to test the quality and coherency of your clusterings using clm dist and clm info. This will most likely reveal that certain values of -I are simply not right for your data." />
+        <param name="transform" type="text" area="true" label="Transform input matrix values"
+            help="See https://micans.org/mcl/man/mcxio.html#trans for the transformation syntax" />
+        <param argument="--discard-loops" type="boolean" checked="true" truevalue="" falsevalue="--discard-loops=n" label="Discard loops in input"
```
gallardoalba, line 64:
```suggestion
        <param argument="--discard-loops" type="boolean" checked="true" truevalue="--discard-loops=y" falsevalue="--discard-loops=n" label="Discard loops in input"
```

tools/mcl/mcl.xml
```diff
@@ -0,0 +1,118 @@
...
+                <param type="data" format="mclmat" name="input" argument="input" label="Input" optional="false"/>
+            </when>
+            <when value="--abc">
+                <param type="data" format="tabular" name="input" argument="input" label="Input" optional="false"/>
+            </when>
+            <when value="--sif">
+                <param type="data" format="sif" name="input" argument="input" label="Input" optional="false"/>
+            </when>
+        </conditional>
+        <param name="inflation" argument="-I" type="float" value="4.0" label="Inflation"
```
gallardoalba, line 60: According to the documentation, the default value should be 2.

tools/mcl/mcl.xml
```diff
@@ -0,0 +1,118 @@
...
+            help="This value is the main handle for affecting cluster granularity. It is usually chosen somewhere in the range [1.2-5.0]. -I 5.0 will tend to result in fine-grained clusterings, and -I 1.2 will tend to result in very coarse grained clusterings. Your mileage will vary depending on the characteristics of your data. That is why it is a good idea to test the quality and coherency of your clusterings using clm dist and clm info. This will most likely reveal that certain values of -I are simply not right for your data." />
+        <param name="transform" type="text" area="true" label="Transform input matrix values"
```
gallardoalba, line 62: I guess it requires to include sanitizers.

tools/mcl/mcl.xml
```diff
@@ -0,0 +1,118 @@
...
+        <param name="sum_loops" type="boolean" checked="false" truevalue="--sum-loops" falsevalue="" argument="--sum-loops"
+            label="Set loops to sum of other arcs weights"  />
+        <param name="reweight" argument="-c" optional="true" type="float" value="1" label="Reweight loops"
+            help="As the final step of loop computation (i.e. after initialization and shadowing) all loop weights are multiplied by the provided value." />
+        <param name="verbosity" type="select" multiple="true" optional="true" label="Verbosity">
+            <option value="pruning">pruning</option>
+            <option value="explain">explain</option>
+            <option value="cls">cls</option>
+        </param>
+    </inputs>
```
gallardoalba, line 75: There are a bunch of additional options which haven't been included, not sure if are important at all.

```
-o <fname>      !  write output to file <fname>

--version       i  show version
-h              i  output description of most important options
--help          i  output description of options
-z              i  show some of the default settings
--jury-charter  i  show the meaning of the jury pruning synopsis
-how-much-ram <int> i  show estimated RAM usage for graphs with <int> nodes
--show-schemes  i  show the preset -scheme options

-overlap <split|cut|keep>    what to do with overlap (default cut)
-force-connected y/n    analyze clustering, make sure it induces cocos
-check-connected y/n    analyze clustering, see whether it induces cocos
--write-limit      output the limit matrix
-show-log y/n      send log to stdout
-discard-loops y/n    remove loops in input graphs if any
--sum-loops        set loop weight to sum of node arc weights
-c <num>           increase loop-weights <num>-fold
-q log-spec        quiet level of logging
-analyze y/n       append performance/characteristics measures
-sort <mode>       order clustering by one of lex|size|revsize|none
-digits <int>      precision in interchange (intermediate matrices) output
--write-binary     write binary output

--abc              expect abc-format (label input), write label output
--sif              expect sif-format (label input), write label output
--etc              expect etc-format (label input), write label output
--expect-values    accept extended SIF or ETC format (label:weight fields)
-use-tab fname     expect native network format, write label output using dictionary
-abc-tf tf-spec    transform label values

--abc-neg-log      log-transform label value, negate sign
--abc-neg-log10    log10-transform label value, negate sign
-write-graph fname    write input matrix to file
-write-graphx fname    write transformed matrix to file

-write-expanded <fname>    file name to write expanded graph to

-annot <description>    string describing the experiment
-aa <suffix>       append <suffix> to mcl-created output file name
-odir <directory> !  use this directory for output
-az             i  show output file name mcl would construct
-ax             i  show the suffix mcl constructs from parameters
-ap <prefix>       prepend <prefix> to mcl-created output file name
--d                use automatic naming and use input directory for output

-tf tf-spec        transform matrix values
-icl fname         subcluster this clustering
-pi <num>          preprocess by applying inflation with parameter <num>
-ph <num>          as -pi, applied before shadowing
-if <num>          assume expanded input, inflate with parameter <num>
-I <num>        !  main inflation value (default 2.0)
-scheme <int>   !  use a preset resource scheme (cf --show-schemes)
-resource <int> !  allow <int> neighbours throughout computation
-sparse <num>      estimated sparse matrix-vector overhead per summand (default 10)
-te <int>       !  expansion thread number, use with multiple CPUs

--show             (small graphs only [#<20]) dump iterands to *screen*
-v {pruning|explain|clusters|all}    mode verbose
-V {pruning|explain|clusters|all}    mode silent

-p <num>           the rigid pruning threshold
-P <int>           (inverted) rigid pruning threshold (cf -z)
-R <int>           recover to maximally <int> entries if needed
-S <int>           select down to <int> entries if needed
-pct <pct>         try recovery if mass is less than <pct>

-warn-factor <int>    warn if pruning reduces entry count by <int>
-warn-pct <pct>    warn if pruning reduces mass to <pct> weight

-dump-stem <str>    use <str> to construct dump (file) names
-dump <mode>       <mode> in chr|ite|cls|dag (cf manual page)
-dump-interval <int>:<int>    only dump for iterand indices in this interval
-dump-modulo <int>    only dump if the iterand index modulo <int> vanishes

Legend:
  (i)   for informative options that exit after usage
  (!)   for important options that you should be aware of
Consult the manual page for more information

```

tools/mcl/mcl.xml
coeit: There is a bunch of programs in the [mcl suite](https://micans.org/mcl/index.html?sec_networktools) that understand the "native" matrix format. In case clustering is part of a larger workflow using multiple programs from that suite it is recommended to use the native format to avoid any confusion.

Realistically if only the `mcl` clustering tool is to be added I'm not sure how much extra gain there is by using "native" mode. However, thinking maybe more long-term, in case the whole suite could be included as galaxy tools using the native format does make a lot of sense.

bgruening: Ok, good to now. Do you have a reference how the mclmat format is defined? We will try to add this format to Galaxy.

coeit: The format is defined here:
https://micans.org/mcl/man/mcxio.html
in section "Specifying matrices".

tools/mcl/mcl.xml
```diff
@@ -0,0 +1,115 @@
+<tool id="mcl" name="MCL" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="21.05">
+    <description>The Markov Cluster Algorithm, a cluster algorithm for graphs</description>
```
bernt-matthias, line 2: Pretty long description

tools/mcl/mcl.xml
```diff
@@ -0,0 +1,115 @@
...
+    '$input_type_select.input_type'
+    -V all -te \${GALAXY_SLOTS:-1}
+    $sum_loops
+    $discard_loops
+    #if $verbosity
+        #for $v in $verbosity
+            -v $v
+        #end for
+    #end if
+    #if $reweight and $reweight != 1.0
```
bernt-matthias, line 27: I would drop the the check for equality with - 1

tools/mcl/mcl.xml
bernt-matthias: Would it be possible to drop the conditional? And replace it with a single data param?

tools/mcl/mcl.xml
```diff
@@ -0,0 +1,115 @@
...
+        <param name="transform" type="text" area="true" label="Transform input matrix values"
+            help="See https://micans.org/mcl/man/mcxio.html#trans for the transformation syntax" />
+        <param argument="--discard-loops" type="boolean" checked="true" truevalue="--discard-loops=y" falsevalue="--discard-loops=n" label="Discard loops in input"
+            help="Remove any loops that are present in the input. Bear in mind that loops will still be modified in all cases where the loop weight is not maximal among the list of edge weights for a given node." />
+        <param name="sum_loops" type="boolean" checked="false" truevalue="--sum-loops" falsevalue="" argument="--sum-loops"
```
bernt-matthias, line 63: Me name attribute can be removed

tools/mcl/mcl.xml
coeit: You mean having only a single pre-defined input format? It depends on how much of the CLI should be mapped here. I think it would be sufficient for most use-cases to only offer a pre-defined input format `--abc`, since it is very easy to get to such an input e.g from BLAST or Diamond tabular outputs. The other formats would most likely be useful in case other tools from the [mcl suite](https://micans.org/mcl/index.html?sec_networktools) would also be included that use these formats as input.

coeit:
Hi @bgruening , I finally got around to revise the tool. I have a revision ready, however am not sure what would be the best way to add my commit to this PR?

Cheers!

coeit:
> Hi @coeit; I think that you could open a PR against Bjorn's repository https://github.com/bgruening/tools-iuc/tree/mcl

Thanks @gallardoalba I did so in https://github.com/bgruening/tools-iuc/pull/556

gallardoalba (approved):

PR #4611, opened 2022-06-23
Update to merqury tool wrapper

neoformit:
Minor update to merqury: add Boolean option "Add headers to tabular outputs"

Previously, tabular outputs were produced which had no headers. This may be desirable to keep the files machine-readable, but these files are annoying to read for humans without column headers. When this option is checked, the tool will inject appropriate headers (taken from the tool docs) into the tabular files.

**NOTE**: I had to modify the test a bit to get them to pass before I even started working on this. These lines in test outputs were not tolerated by Planemo since they had no file to compare against:

```xml
<output_collection name="qv_files" type="list" count="3">
    <!-- Had to remove these to pass test due to planemo errors: "no file to compare"
    <element name="output_03.assembly_01" ftype="tabular"/>
    <element name="output_03.assembly_02" ftype="tabular"/>
    -->
    <element name="output_03" file="output_03_qv.tabular" ftype="tabular"/>
</output_collection>
```

I have updated the test `expect_num_output` and collection `count` properties to reflect these updates.


tools/merqury/merqury.xml
```diff
@@ -204,13 +212,12 @@
...
             <param name="label" value="output_01"/>
             <param name="output_selector" value="sizes"/>
-            <output_collection name="sizes_files" type="list" count="2">
-                <element name="output_01.assembly.100_20000.phased_block" ftype="tabular"/>
+            <output_collection name="sizes_files" type="list">
                 <element name="output_01.assembly.contig" file="output_01.assembly.contig.tabular" ftype="tabular"/>
             </output_collection>
         </test>
-        <!--Test 02: standard mode-->
-        <test  expect_num_outputs="11">
```
bgruening, line 213: please keep the expect_num_outputs

tools/merqury/merqury.xml
```diff
@@ -159,6 +166,7 @@
             <option value="wig">WIG: hap-mer tracks</option>
             <option value="log">Log file</option>
         </param>
+        <param name="output_add_headers" type="boolean" value="false" label="Add column header on tabular outputs" />
```
bgruening, line 169: would it be enough to add the column headers as metadata annotation (Galaxy then displays them in the UI) or do you need this as part of the file on disc?

tools/merqury/merqury.xml
```diff
@@ -292,14 +299,14 @@
             <output_collection name="stats_files" type="list" count="1">
                 <element name="output_03.completeness" file="output_03.completeness.stats.tabular" ftype="tabular"/>
             </output_collection>
-            <output_collection name="qv_files" type="list" count="3">
-                <element name="output_03.assembly_01" ftype="tabular"/>
-                <element name="output_03.assembly_02" ftype="tabular"/>
+            <output_collection name="qv_files" type="list">
```
bgruening, line 302: what is the problem here?

bgruening (commented):
ping also @gallardoalba

tools/merqury/merqury.xml
```diff
@@ -204,13 +212,12 @@
...
             <param name="output_selector" value="sizes"/>
-            <output_collection name="sizes_files" type="list" count="2">
-                <element name="output_01.assembly.100_20000.phased_block" ftype="tabular"/>
+            <output_collection name="sizes_files" type="list">
                 <element name="output_01.assembly.contig" file="output_01.assembly.contig.tabular" ftype="tabular"/>
             </output_collection>
         </test>
-        <!--Test 02: standard mode-->
-        <test  expect_num_outputs="11">
+        <!--Test 02: standard mode with tabular headers -->
```
gallardoalba, line 219: I would prefer to include additional tests instead of modifying an existing one.

tools/merqury/merqury.xml
neoformit: Sorry I forgot to mention the issues with test - I have extended the PR description

bgruening:
@neoformit wouldn't this be a better approach? https://docs.galaxyproject.org/en/latest/dev/schema.html#unconditional-actions-and-column-names

tools/merqury/merqury.xml
neoformit: Do you mean like this? (Sorry I haven't seen that before)

```xml
<outputs>
    <data format="tabular" name="table">
        <actions>
            <action name="column_names" type="metadata" default="Firstname,Lastname,Age" />
        </actions>
    </data>
</outputs>
```

tools/merqury/merqury.xml
neoformit: I've created a new test for this

tools/merqury/merqury.xml
bgruening: Yes https://docs.galaxyproject.org/en/latest/dev/schema.html#unconditional-actions-and-column-names

This way we do not need to touch the output file from the tool.

bgruening: And in general we should use this more often and annotate our tables better :)

neoformit:
> wouldn't this be a better approach? https://docs.galaxyproject.org/en/latest/dev/schema.html#unconditional-actions-and-column-names

Hah I just copied and pasted this in reply to you. Yep that should work, I'll just check with the requester before I push. Thanks.

neoformit:
@bgruening How do I add `<actions>` to these?

```xml
<collection name="stats_files" type="list" label="${tool.name} on ${on_string}: stats">
    <discover_datasets pattern="(?P&lt;name&gt;.+)\.stats" format="tabular" directory="output_files"/>
    <filter>"stats" in output_selector</filter>
</collection>
```

bgruening:
No idea :( please grep through this repo, hopefully you will find an example.

neoformit:
> No idea :( please grep through this repo, hopefully you will find an example.

@bgruening no examples of this in `tools-iuc` - I tried two obvious ways that it might work and both were rejected by `planemo lint`. I guess it's not possible on collections.

gallardoalba:
Could you try to re-generate the outputs by using Docker? The command is _planemo t --update_test_data --docker_.

neoformit:
> Could you try to re-generate the outputs by using Docker? The command is _planemo t --update_test_data --docker_.

I ran this and it reverted the changes I made to the test data headers. Now two tests fail with this:

```
Collection element output_06 of collection qv_files:  different than expected, difference (using diff):
( /tmp/tmpccfal523output_06_qv.tabular v. /tmp/tmpqamlu8v6output_06_qv.tabular )
--- local_file
+++ history_data
@@ -1,2 +1,2 @@
 assembly	unique k-mers	common k-mers	QV	error rate
-assembly	0	1037953		
+assembly	0	1037953	+inf	0
```

Why does running with docker produce a different output? Which one should I use to pass the IUC tests?

bgruening:
Docker or non docker should not produce different results. If it's does there is an underlying problem. We run tests with docker here as they are more strict. Concrete we are running with planemo test --biocontainers...

neoformit:
> Docker or non docker should not produce different results. If it's does there is an underlying problem. We run tests with docker here as they are more strict. Concrete we are running with planemo test --biocontainers...

I'm not sure how to remedy that @bgruening but the CI tests are passing now

gallardoalba:
Thanks @neoformit!

PR #4627, opened 2022-06-28
Added edam operation and topic to annotateMyIDs

LucieLamothe:
operations: Genome annotation, Data retrieval
topic: Genomics


tools/annotatemyids/annotateMyIDs.xml
```diff
@@ -1,5 +1,12 @@
 <tool id="annotatemyids" name="annotateMyIDs" version="3.12.0+galaxy1">
     <description>annotate a generic set of identifiers</description>
+    <edam_topics>
```
bgruening, line 3: I think it would be better to add the bio.tools ID: https://bio.tools/annotatemyids

gallardoalba:
I included the bio.tools ID in this PR https://github.com/galaxyproject/tools-iuc/pull/4636

PR #4628, opened 2022-06-28
add new tool retropath2 v2.3.1

kenza12:
Add new tool **retropath2 v2.3.1** : https://github.com/brsynth/RetroPath2-wrapper

FOR CONTRIBUTOR:


tools/retropath2/.shed.yml
```diff
@@ -0,0 +1,7 @@
+categories:
+- Synthetic Biology
+description: Build a reaction network from a set of source compounds to a set of sink compounds
+long_description: Perform retrosynthesis search of possible metabolic routes between a source molecule and a collection of sink molecules.  It takes for input the minimal (dmin) and maximal (dmax) diameter for the reaction rules and the maximal path length (maxSteps). The tool expects the following files: rules.csv, sink.csv and source.csv and produce a retrosynthesis network as a CSV file providing reactions in the reaction SMILES format and chemicals in both SMILES and InChI formats along with other information like the score for each reaction.
```
bgruening, line 4: this is invalid YAML since you use a `:`. I can help you fix those things, but I do not have write access to this PR.

tools/retropath2/retropath2.xml
```diff
@@ -0,0 +1,135 @@
...
+            --source_name '$source_name'
+            --rp2_version '$adv.version'
+            --max_steps '$max_steps'
+            --topx '$adv.topx'
+            --dmin '$adv.dmin'
+            --dmax '$adv.dmax'
+            --mwmax_source '$adv.mwmax_source'
+            --timeout '$adv.timeout' &&
+        if compgen -G 'out/*_scope.csv' > /dev/null; then
+            cp out/*_scope.csv '$Reaction_Network';
```
bgruening, line 32: use && instead of `;`

tools/retropath2/retropath2.xml
```diff
@@ -0,0 +1,135 @@
...
+        else
+            cp out/results.csv '$Reaction_Network';
+        fi
+    ]]></command>
+    <inputs>
+        <param name="rulesfile" type="data" format="csv,tar" label="Rules File" help="File containing reaction rules which reflect the enzymatic potential of the chassis organism."/>
```
bgruening, line 38: why tar?

tools/retropath2/retropath2.xml
```diff
@@ -0,0 +1,135 @@
...
+        <param name="sinkfile" type="data" format="csv" label="Sink File" help=" Sink file which comprises all compounds that are considered as granted in your system."/>
+        <param name="source_inchi" type="text" label="Source InChI" optional="false" help="InChI of compound to produce.">
+            <validator type="empty_field" message="You must provide the InChI string"/>
+        </param>
+        <param name="max_steps" type="integer" value="3" min="1" max="10" label="Maximal Pathway length" help="The maximal number of metabolic steps (i.e. reactions) for exploring around the target."/>
+        <param name="source_name" type="text" value="target" optional="true" label="Source name" help="Name of compound to produce.">
+            <validator type="empty_field" message="Source name is required."/>
+        </param>
+        <section name="adv" title="Advanced Options" expanded="false">
+            <param name="version" type="select" label="Workflow version" help="Retropath2.0 Workflow versions, formatted as 'r{year}{month}{day}' (the latest version 'r20220104' is used by default).">
```
bgruening, line 48: the help does not fit the selection options I think

tools/retropath2/retropath2.xml
```diff
@@ -0,0 +1,135 @@
...
+            <param name="dmin" type="integer" value="0" min="0" max="1000" label="Minimum rule diameter" help="Minimum rule diameter of the sphere including the atoms around the reacting center. The higher is the diameter, the more specific are the rules." />
+            <param name="dmax" type="integer" value="1000" min="0" max="1000" label="Maximum rule diameter" help="Maximum rule diameter of the sphere including the atoms around the reacting center. The higher is the diameter, the more specific are the rules."/>
+            <param name="mwmax_source" type="integer" value="1000" min="0" max="2000" label="Molecular weight of source (Da)" help="The molecular weight cutoff (in Da), above which initial source (ie target) and intermediate compounds will be filtered out." />
+            <param name="timeout" type="integer" value="60" min="30" max="600" label="Timeout (min)" help="Maximal time of RetroPath2.0 execution (60 minutes by default)." />
+        </section>
+    </inputs>
+    <outputs>
+        <data name="Reaction_Network" format="csv" label="${tool.name}" />
```
bgruening, line 61: you could add the column names as metadata: https://docs.galaxyproject.org/en/latest/dev/schema.html#unconditional-actions-and-column-names

tools/retropath2/.shed.yml
kenza12: Hi,
I think I fixed it with adding a single quote.
Thank you.

tools/retropath2/retropath2.xml
kenza12: I can't, it doesn't work with && in line 32. I used && in line 30.

tools/retropath2/retropath2.xml
kenza12: Because the tool can handle 2 different formats of this input, a csv format or compressed tar.

tools/retropath2/retropath2.xml
kenza12: Ok, done, thanks.

tools/retropath2/retropath2.xml
kenza12: I confirm you that the help is correct. The latest version 'r20220104' is used by default and selected in the following option:

`<option value="r20220104" selected="true">r20220104</option>`

kenza12:
Hi, The developer just added a new option to this tool `--kinstall` in order to install some dependencies in a specific folder (using kinstall option) instead of tool directory which is forbidden in this case because it's running in a container.

The problem is that I got another error: 

```
Job in error state.. tool_id: retropath2_wrapper, exit_code: 1, stderr: /usr/local/lib/python3.10/runpy.py:126: RuntimeWarning: 'retropath2_wrapper.__main__' found in sys.modules after import of package 'retropath2_wrapper', but prior to execution of 'retropath2_wrapper.__main__'; this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))
[32mChecking input data[0m
[32m   |- InChI[0m
[32m   |- Source in Sink (simple)[0m
[32mDownloading KNIME 4.5.0...[0m
  File "/usr/local/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.10/site-packages/retropath2_wrapper/__main__.py", line 207, in <module>
    _cli()
  File "/usr/local/lib/python3.10/site-packages/retropath2_wrapper/__main__.py", line 95, in _cli
    r_code, result_files = retropath2(
  File "/usr/local/lib/python3.10/site-packages/retropath2_wrapper/RetroPath2.py", line 195, in retropath2
    install_knime(
  File "/usr/local/lib/python3.10/site-packages/retropath2_wrapper/RetroPath2.py", line 441, in install_knime
    chown_r(kinstall, getuser())
  File "/usr/local/lib/python3.10/getpass.py", line 169, in getuser
    return pwd.getpwuid(os.getuid())[0]
KeyError: 'getpwuid(): uid not found: 1001'
```

Note that I don't get this error when I use `planemo test` without biocontainers option. So I think the tool can't access to the process ID because it's containerized. I'm not sure to understand this error. Maybe you can help me to fix this error ?

Thank your for your time.

bgruening:
I have built the container "manually" https://quay.io/repository/biocontainers/retropath2_wrapper and this is now used here - but it's still happening. I don't know why this software is doing the chown thing when the same user just installed the software before - I would assume this is not needed.

Is there any way to install knime as part of the conda package?

bgruening:
As a hack you can put this tool on the https://github.com/galaxyproject/tools-iuc/blob/e982ed53ff833f7792a24e5223327c286dac2bda/.tt_biocontainer_skip list. But that might be too hacky for IUC. If you like you can move your tool to https://github.com/bgruening/galaxytools where we can have slightly more hacky tools.

tools/retropath2_wrapper/wrap.xml
```diff
@@ -2,24 +2,20 @@
     <description>Build a reaction network from a set of source compounds to a set of sink compounds</description>
     <macros>
         <token name="@VERSION_SUFFIX@">0</token>
-        <token name="@TOOL_VERSION@">2.4.0</token>
+        <token name="@TOOL_VERSION@">2.6.0</token>
     </macros>
     <requirements>
-        <requirement type="package" version="@TOOL_VERSION@">retropath2_wrapper</requirement>
+        <requirement type="docker">ggricourt/retropath2_wrapper</requirement>
```
bgruening, line 8: Is there any reason why we can not use Conda packages as well? Not everyone can run containers.

tools/retropath2_wrapper/wrap.xml
```diff
@@ -0,0 +1,129 @@
...
+            --dmin '$adv.dmin'
+            --dmax '$adv.dmax'
+            --mwmax_source '$adv.mwmax_source'
+            --kinstall '/opt'
+            --kver '4.7.0'
+            --log 'debug'
+            --msc_timeout '$adv.timeout'
+            '$sinkfile'
+            '$rulesfile'
+            "\${_GALAXY_JOB_TMP_DIR:-.}/out";
```
bgruening, line 33: why not use the CWD? So simply: `./out/`

bgruening: please also use && instead of ;

tools/retropath2_wrapper/wrap.xml
```diff
@@ -0,0 +1,129 @@
...
+            code=\$?;
+            if compgen -G '\${_GALAXY_JOB_TMP_DIR:-.}/out/*_scope.csv' > /dev/null; then
+                cp \${_GALAXY_JOB_TMP_DIR:-.}/out/*_scope.csv '$Reaction_Network';
+            else
+                cp \${_GALAXY_JOB_TMP_DIR:-.}/out/results.csv '$Reaction_Network';
+            fi;
+            (exit \$code)
```
bgruening, line 40: I think when you use && you can skip the exit hack.

tools/retropath2_wrapper/wrap.xml
guillaume-gricourt: added in 99d3b9e

tools/retropath2_wrapper/wrap.xml
guillaume-gricourt: added in 99d3b9e

guillaume-gricourt:
Please don't merge the PR now, the docker requirement is not acceptable. `retropath2_wrapper` uses Knime and it's not easy to manage it only using conda.

bgruening:
@guillaume-gricourt if its not possible with conda, we probably can go with Docker only.

guillaume-gricourt:
> 

You are right but I would like to find a workaround before

bgruening:
Cool, let me know if you need help! Thanks a lot @guillaume-gricourt and @kenza12!

bernt-matthias:
> retropath2_wrapper uses Knime

@guillaume-gricourt .. would you be so kind and explain this to me? How do you use KNIME in a command line pipeline? I'm asking because we have some KNIME users at our institution .. and I would like to see them using Galaxy :)

bernt-matthias:
Interesting, did not know that you can call knime workflows from the command line :) 

So, why not put knime itself to conda (bioconda / conda-forge)? If I get it right its GPL, or? Are there any pre-compiled components? The code that you have looks quite close to a recipe for conda. 

Mirroring/archiving to zenodo seems to be a good strategy  to me.

guillaume-gricourt:
Knime has developed its own python wrapper [knime](https://github.com/KNIME/knimepy) available on pypi.
Unfortunately, it seems to be not maintained and it uses an old version of Knime. Also, it doesn't provide Knime packages.
I don't known if there is an issue about licensing. As pre-compiled components, I guess they are only those mentioned in.the meta.yaml corresponding to `retropath2_wrapper` like `cups-devel`.
Adding Knime on conda will involve extensive maintenance work, especially if it stores each version, for each platform with their packages to upload on Zenodo. Also, Knime can add extra dependencies at each update.

bernt-matthias:
> Adding Knime on conda will involve extensive maintenance work, especially if it stores each version, for each platform with their packages to upload on Zenodo.

In bioconda each version is stored, but typically only the latest receives updates (technically it is possible to update multiple versions, e.g. the latest of each major version, but seldomly done).

For each version a container is created which usually guarantees that also older versions of the software are running even if requirements change.

Uploads to zenodo (or another platform) would only be needed for Linux (and if desired MacOS) since bioconda only supports these platforms anyway. I stumble over more and more Java projects that have platform dependent releases .. I always thought that the point of Java is platform independence...

My idea would be to add an initial recipe and hope that its of interest for a but larger community which might help with maintenance and updates. 

Xref: https://github.com/bioconda/bioconda-recipes/pull/3649 There was an earlier attempt. @bgruening what do you think about this? 

An important question might be: is knime itself actually needed in order to execute a knime workflow as you do in your case...?

guillaume-gricourt:
> An important question might be: is knime itself actually needed in order to execute a knime workflow as you do in your case...?

Yes, it's mandatory. Knime and its packages.

guillaume-gricourt:
Hi, 
We have decided not to  incorporate this tool, so please close this pull request.
Thanks

bgruening:
Sorry to hear that. It seems to be complicated. Let us know if we can help anyhow next time.

PR #4629, opened 2022-06-28
Added edam topic and operation to quast

LucieLamothe:
operations: Visualisation, Sequence assembly validation
topic: sequence assembly


bernt-matthias:
Since the change is not needed I'm going to close this. Thanks for your contribution @LucieLamothe

PR #4631, opened 2022-06-28
Added edam annotations to augustus

LucieLamothe:
operations: Gene prediction, Ab-initio gene prediction, Homology-based gene prediction
topics: Gene and protein families, Gene transcripts


bernt-matthias:
Since the change is not needed I'm going to close this as well.

PR #4632, opened 2022-06-28
Added edam annotations to egsea

LucieLamothe:
topics: Gene expression, Systems biology
operations: Gene-set enrichment analysis

### DIFF
--- a/tools/egsea/egsea.xml
+++ b/tools/egsea/egsea.xml
@@ -1,5 +1,12 @@
 <tool id="egsea" name="EGSEA" version="1.20.0">
     <description> easy and efficient ensemble gene set testing</description>
+    <edam_topics>                                                                                  
+        <edam_topic>topic_2259</edam_topic>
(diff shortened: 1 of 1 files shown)


tools/egsea/egsea.xml
```diff
@@ -1,5 +1,12 @@
 <tool id="egsea" name="EGSEA" version="1.20.0">
     <description> easy and efficient ensemble gene set testing</description>
+    <edam_topics>
```
bgruening, line 3: @LucieLamothe if we have a bio.tools ID we do not need the EDAM terms, as they will be retrieved automatically from bio.tools.

LucieLamothe: Oh I didn't know that, it's great! :) Shall I add bio.tools id instead when they exists then?

nsoranzo: Yes, that's the idea. Thanks!

bernt-matthias:
Then I guess I can close this one

PR #4633, opened 2022-06-28
Add data manager for pangolin-data

pvanheus:
FOR CONTRIBUTOR:


wm75:
@pvanheus I cannot currently planemo serve the DM so presumably there's something wrong (name or ID mismatch between the different DM/data table files, or something along these lines) still. Or is that not a problem for you?

pvanheus:
> @pvanheus I cannot currently planemo serve the DM so presumably there's something wrong (name or ID mismatch between the different DM/data table files, or something along these lines) still. Or is that not a problem for you?

It works for me. When I am in the top level directory, then `planemo serve --docker --galaxy_root MYGALAXYROOT data_manager/pangolin_data_dm.xml` works for me.

wm75:
Hmm, ok, then I need to debug this on my end first. Except for --docker I'm running the same command as you, but the DM doesn't show up in the tools panel for me.

wm75:
In the meantime, I have started working on version 4.1.1 of the pangolin wrapper, which will already offer your new data tables. So before we merge this one here, I'd like to test the use of these tables on the pangolin tool wrapper side. Should be done tomorrow hopefully.

wm75:
Populating the options from a csv file is somewhat an undocumented hack maybe atm (see https://github.com/galaxyproject/galaxy/issues/14324) but in the very worst case the data manager help would have to tell users to run a csv to tabular conversion first on the downloaded file.

pvanheus:
@wm75 please have a look at the new version of the DM PR.

wm75:
I've changed the requirements a bit, but looks good to me.

Personally, what I would do is decrease the nesting of the interface a bit by presenting three top-level options per data source:
`Download latest version`, `Download specific version`, `Do not download`.
We could also do this later if you feel like you've spent enough time on this by now.

data_managers/data_manager_pangolin_data/data_manager/pangolin_data_dm.py
```diff
@@ -0,0 +1,220 @@
...
+        print("XXXX versions:", args.versions)
+        releases_wanted = set(args.versions)
+        releases = []
+        for release in downloadable_releases:
+            version = release["tag_name"].lstrip('v.')
+            if version in releases_wanted:
+                if version in compatibility:
+                    # only add the releases for which we have compatibility info
+                    release[min_version_key] = compatibility[version]
+                    releases.append(release)
```
wm75, line 193: This bit could be made a bit more robust if upon appending to releases, you'd also remove the corresponding item from releases_wanted.
This way you could i) break out of the loop early once you've consumed/found all releases_wanted, ii) fail with a proper error message if not all of the requested releases could be found.

data_managers/data_manager_pangolin_data/data_manager/macros.xml
```diff
@@ -0,0 +1,26 @@
...
+                <option value="latest" selected="true">Latest</option>
+                <option value="versions">Choose from release list</option>
+            </param>
+            <when value="latest">
+            </when>
+            <when value="versions">
+                <param name="compatibility_file" 
+                       type="data" format="csv"
+                       label="Version compatibility file"
+                       help="File describing compatibility between @SOFTWARE_NAME@ releases and databases" />
```
wm75, line 14:
```suggestion
                       help="CSV dataset describing compatibility between @SOFTWARE_NAME@ releases and databases; typically, you would want to download fresh from https://raw.githubusercontent.com/cov-lineages/pangolin/master/pangolin/data/data_compatibility.csv" />
```
Makes the link easier to discover than from the general help.

pvanheus: Thanks, added.

wm75 (commented):
Really cool, thanks!
Just tiny things left to do here.

wm75:
One more minor thing: I just noticed that pangolin-data/-assignment version 1.2.133 cannot be installed because the github release tags for them are `v1.2.133.2` for pangolin-data and `v1.2.133.1` for pangolin-assignment.
I guess we can leave that as is for now (alternatives would be to fix the upstream csv file, and ask if those two minor versions are actually compatible with each other, or to introduce something like a tag_translation dictionary into the DM) since the error message you're getting from the DM is clear enough.
Maybe we can fix the upstream file with its next update.

pvanheus:
> One more minor thing: I just noticed that pangolin-data/-assignment version 1.2.133 cannot be installed because the github release tags for them are `v1.2.133.2` for pangolin-data and `v1.2.133.1` for pangolin-assignment. I guess we can leave that as is for now since the error message you're getting from the DM is clear enough (alternatives would be to fix the upstream csv file, and ask if those two minor versions are actually compatible with each other, or to introduce something like a tag_translation dictionary into the DM). Maybe we can fix the upstream file with its next update.

My goodness! pangolin-data versions that look like PANGO lineages.

I'd hope that the version in Github is the "true version" and we can get around this by fixing the CSV file. Perhaps best to make a PR and wait for comments before taking action on the Galaxy side.

data_managers/data_manager_pangolin_data/data_manager/pangolin_data_dm.py
```diff
@@ -0,0 +1,247 @@
...
+            releases = []
+        else:
+            releases = [latest_release]
+    else:
+        compatibility = read_compatibility_info(
+            args.version_compatibility_file, package_name
+        )
+        downloadable_releases = get_model_list(existing_release_tags, package_name)
+        print("XXXX versions:", args.versions)
+        releases_wanted = set(args.versions)
```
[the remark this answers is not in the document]
wm75: Smart :-) Thanks!

PR #4634, opened 2022-06-30
Gemini query: fix region filter option

gallardoalba:
FOR CONTRIBUTOR:

This PR fixes a bug in the region filtering option. It was reported in Ghelp https://help.galaxyproject.org/t/gemini-query-empty/8245/3


wm75:
I'm not quite sure what you're trying to fix here.
If a user specifies a chomosome identifier as part of the region filter, they need to provide the correct name.
I don't think it's helpful to force a `chr` prefix unconditionally since you will now punish users that typed it.

gallardoalba:
> I'm not quite sure what you're trying to fix here. If a user specifies a chomosome identifier as part of the region filter, they need to provide the correct name. I don't think it's helpful to force a `chr` prefix unconditionally since you will now punish users that typed it.

Yes, I agree that it is strange, but if you check [the dataset used for generating the input](https://github.com/galaxyproject/tools-iuc/blob/67ea0ffc9b2e08f5b99ffea8bbe654e810e691cc/tools/gemini/test-data/util/build-data/gemini_load_input.vcf#L37), the chromosome field is 1, but in order to work, it requires that modification, otherwise it generates an empty output.

gallardoalba:
> Was there a specific bug report that prompted you to propose this change?

Yes, the problem was reported here https://help.galaxyproject.org/t/gemini-query-empty/8245

wm75:
I see. So GEMINI automatically adds a `chr` prefix when building its database from any input VCF. As a result users need to use a `chr` prefix for region filters even if their original VCF did not have it, which is confusing.
Your proposed change is not a fix though for users that already had a `chr` prefix in their VCF and are now using it also in their region filter expression because you will change their query to `chrchr`.

wm75:
Two viable options here:
- simply update the help to mention that region filters MUST use a `chr` prefix
- rewrite the logic to check whether the prefix was user-typed already and only add it when it's missing

gallardoalba:
> The problem with that second option is that you cannot do this in the expert mode where users type the full query directly, can you? So beginners would learn syntax that doesn't work later for them in expert mode or on the command line.

So do you think is better to include the advice? If so, I'll include it in the wrapper.

wm75:
Hmm, still thinking about better ways.
Maybe a validator that doesn't accept text without a leading `chr` as the chrom field?

wm75:
I guess that should work since GEMINI load appears to enforce `chr` on all chromosomes here:
https://github.com/arq5x/gemini/blob/d393bf3d76a6ff91f711525cb00b6954d2193651/gemini/gemini_load_chunk.py#L513

gallardoalba:
> I guess that should work since GEMINI load appears to enforce `chr` on all chromosomes here: https://github.com/arq5x/gemini/blob/d393bf3d76a6ff91f711525cb00b6954d2193651/gemini/gemini_load_chunk.py#L513

Great, thanks for checking!

gallardoalba:
Do you think it is fine @wm75?

PR #4644, opened 2022-07-05
instrain: add zip requirement to all

bernt-matthias:
tool currently fails CI

compare also calls unzip and should have the requirement

tool still fails, because Galaxy unzips the zip inputs: I see `ZIP file contained more than one file, only the first file was added to Galaxy.` in the Dataset Info of the input datasets (and the actual tool fails with `unzip: short read`).

FOR CONTRIBUTOR:

### DIFF
--- a/tools/instrain/instrain_profile.xml
+++ b/tools/instrain/instrain_profile.xml
@@ -4,9 +4,7 @@
         <import>macros.xml</import>
     </macros>
     <expand macro="edam_ontology"/>
-    <expand macro="requirements">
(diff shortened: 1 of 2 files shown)


bernt-matthias:
zip problem has been solved here https://github.com/galaxyproject/galaxy/pull/14216/commits/da381443e0b7e3f1999816b40930e13b83dd1b22

so the PR just unifies the compute environment which is good in my opinion.

bgruening (approved):

PR #4654, opened 2022-07-08
Gemini query:  fix region filter option

gallardoalba:
FOR CONTRIBUTOR:

This PR fixes a bug in the region filtering option. It was reported in Ghelp https://help.galaxyproject.org/t/gemini-query-empty/8245/3. It replaces https://github.com/galaxyproject/tools-iuc/pull/4634


tools/gemini/gemini_macros.xml
```diff
@@ -184,7 +184,7 @@
         <repeat name="regions" title="Region Filter" default="0" min="0"
         help="Filter variant sites by their position in the genome. If multiple Region Filters are specified, all variants that fall in ONE of the regions are reported.">
             <param name="chrom" type="text" label="Chromosome">
-                <validator type="expression" message="A chromosome identifier is required when specifying a region filter">value.strip()</validator>
+                <validator type="expression" message="A chromosome identifier is required when specifying a region filter. It should start with 'chr'">^chr\d+</validator>
```
wm75, line 187:
```suggestion
                <validator type="expression" message="A chromosome identifier is required when specifying a region filter. It should start with 'chr'">value.startswith('chr')</validator>
```
You can't assume digits right after or you'd make chrM impossible to specify and with type="expression", a Python expression is expected (otherwise you'd have to use type="regex".

wm75:
This time I've also looked at the failing tests and they are probably benign.
Gemini has tons of unpinned dependencies and so it's not that surprising that the 5 gemini load tests fail due to unexpected size differences in the resulting gemini.sqlite databases. I guess it will be easiest to just increase the delta of all these 5 tests to 17000 and be done with this.

gallardoalba:
It seems to be fine now; I tried to update the test datasets instead of increasing the delta value but without success (it generated the same datasets when using `planemo t --update_test_data --docker`)

wm75:
Yes, this looks good now. Only thing to address still is: `gemini_gene_wise.xml` uses that same macro so that wrapper version needs a bump, too.

wm75 (approved):

PR #4662, opened 2022-07-12
Improve robustness of column_remove_by_header to encodings

wm75:
The previous version of the tool would use whatever encoding was configured in
its environment to parse the input file and could fail with non-ASCII
characters as a consequence.
With the change the tool can handle nearly any file encoding and special
characters in column names. The only limitations should be:
- the column delimiter characters need to be from the ASCII range
- special characters of column names in non-UTF-8-encoded files may be
  parsed incorrectly and so it may be impossible to select such columns

FOR CONTRIBUTOR:


wm75:
@bgruening can we get this merged so that @shiltemann can start using the new version on usegalaxy.eu?

bgruening (approved):
Looks good for me.

shiltemann:
thanks a lot @wm75!

PR #4674, opened 2022-07-20
column_maker on steroids

wm75:
Enables multiple expressions to be computed sequentially on every input
data row. In addition, the resulting columns can now not only be
appended to existing columns, but also be inserted at arbitrary
positions or be used to replace any existing column.

FOR CONTRIBUTOR:


tools/column_maker/column_maker.xml
```diff
@@ -1,101 +1,123 @@
...
+        label="Avoid scientific notation in any newly computed columns"
+        help="If yes, use fully expanded decimal representation when writing new columns with floating point values. To prevent scientific notation in just specific new columns, you can use numpy's format_float_positional function in the corresponding expression." />
     </inputs>
     <outputs>
         <data name="out_file1" format_source="input" metadata_source="input"/>
     </outputs>
     <tests>
         <test>
-            <param name="cond" value="c3-c2"/>
+            <param name="cond" value="float(c3-c2)"/>
```
bgruening, line 91: How hard is it to make it work the old way? I think this is a major downside for UX if a simple `c3-c2` does not work anymore.

wm75: That test is quite deceiving. `c3` and `c2` are integer numbers and the fact that the result is a float is actually a flaw of the old version. It had to cast ints to floats before doing any computation because of python2's div behavior. That's why the round option was so important previously.
The new version does the right thing by default, but I didn't want to change the test files and forced the old behavior.

wm75: See also the changed usage examples in the help section.

wm75:
@shiltemann might be interesting for your https://github.com/galaxyproject/training-material/pull/3542 ?

pvanheus:
There's a bug in this latest version - if you specify a column that doesn't exist, you get a successful tool run but no output.

I discovered this while doing an experiment:
1. do something that appends a column
2. add a second operation that uses the result in the newly created column.

The help does not say this this is *not* possible and it seems like a thing I might want to do.

wm75:
> * do something that appends a column
> 
> * add a second operation that uses the result in the newly created column.

This should definitely be supported and I just tested that it is. Can you specify what exactly you've tried?

> if you specify a column that doesn't exist, you get a successful tool run but no output.

This is true, but is not specific to the new version. You should get a (relatively) clear warning in that case on stdout with a recommendation to verify your column choices (which I even tried to improve in the new version), but such runs have never been considered failed afaik.

tools/column_maker/column_maker.py
```diff
@@ -20,143 +21,253 @@
...
+            skipped_lines += 1
+            if not invalid_line:
+                first_invalid_line = i + 1
+                invalid_line = line
+            continue
+        fields = line.split('\t')
+        typed_fields = cast_types(*fields)
+        try:
+            for fun, col_idx, mode, col_name in ops:
+                new_val = fun(*typed_fields)
```
wm75, line 234: @pvanheus one *could* make the tool fail on non-existing column names by putting a `try: ... except NameError` around this single line.
Question to everyone: objections to such a change, which would make the tool fail in a situation tolerated by previous versions?

wm75: This could break hypothetical WFs in which an upstream tool generates output with input-dependent variable number of columns and where column_maker is used to either compute a new column or produce empty output if the expected column is not present.
The case that a user typed a wrong column by mistake is, of course, *much* more frequent but only requires the user to realize the problem and fix it.

wm75: Hmm, I could add a "permissive" option to the tool that makes it skip "NameError" lines like previous versions, but make strict mode the default?

pvanheus:
> > * do something that appends a column
> > * add a second operation that uses the result in the newly created column.
> 
> This should definitely be supported and I just tested that it is. Can you specify what exactly you've tried?

On testing again, it works. I'm not sure what I did wrong the first time.

> 
> > if you specify a column that doesn't exist, you get a successful tool run but no output.
> 
> This is true, but is not specific to the new version. You should get a (relatively) clear warning in that case on stdout with a recommendation to verify your column choices (which I even tried to improve in the new version), but such runs have never been considered failed afaik.

The message is indeed clear on stdout, but the blank result is nonetheless confusing. While this didn't provoke failure in previous versions of this tool, I'd argue that that was a mistake.

shiltemann:
@wm75 ooh very nice! that will definitely simplify some of the tutorial steps ..making this tutorial really highlighted a bunch of tool improvements we can make, this one was on my list, thanks!

There was one more for this tool actually: my dataset had columns with missing or invalid data, and this tool only outputs those rows for which it could perform the requested computation, losing a lot of rows in the process. But it would be nicer to have the option to keep all rows of the original files, and for those that it could not perform the computation for, simply write out a null value in the column (and let the user choose whether this would be "NA", NaN, empty cell, whatever other value). Not sure how much work that would be to implement, but might be nice to make that an option?

e.g. see final exercise 3 of my tutorial ( https://training.galaxyproject.org/training-material/topics/introduction/tutorials/data-manipulation-olympics/tutorial.html#exercises-putting-it-all-together ) for an example of this, I made it an exercise to work around this issue and add the original rows back, but it could all be solved in this tool if it has an option to output all the original rows, even if it couldn't perform the computation

tools/column_maker/column_maker.xml
bgruening: Ok, I see, thanks.

tools/column_maker/column_maker.py
pvanheus: Yep. I see @shiltemann has a request for a different kind of "permissive" mode. I wonder if the two are or can be connected.

wm75:
@shiltemann thanks for the feedback. I think this idea and @pvanheus suggestion that there should be at least an option for failing on non-existent columns can both be implemented. I think I'll add a new "Error handling" section to the tool.

wm75:
@shiltemann your olympics case is now covered. If you run the new tool version as described in the tutorial, it will still drop NA lines, but you can now go into the new Error handling section and select either of:
- **If an expression cannot be computed for a row**: `Keep the row unchanged` (which will emit a mix of lines with BMI added and lines without that additional column, hmm)
- **If an expression cannot be computed for a row**: `Produce an empty column value for the row` (which will produce lines with equal number of columns, but some last columns are empty)
- **If an expression cannot be computed for a row**: `Fill in a replacement value`
     **Replacement value**: `NA` (which will give NA values for BMI for lines for which it could not be computed)

wm75:
@pvanheus if an expression references a non-existent column the tool will now fail by default, but you can uncheck **Fail on references to non-existent columns** to get the old skip-all-lines behavior.

Wrapper tests will follow tomorrow.

wm75:
Puh, the tests revealed several complex edge cases, but now everything should be fine and there is a test for the olympics data with BMI calculation and a nice usage example for the error handling section.

wm75:
@shiltemann @pvanheus and everyone else interested: the exception handling in this code is really finnicky with tons of corner cases. I *think* I found and dealt with the most important ones, but please test yourselves on your own tabular data and let me know if you run into more issues.

wm75:
From my side, I'm happy with this PR. To summarize the changes it brings:

new features
==========
- can chain mutltiple expressions in a single run of the tool
- can insert new columns in arbitrary position
- can replace existing columns with newly calculated ones
- can now handle unicode outside the ASCII range independent of locale settings

changes
======
- No longer converts int columns to float internally before computing an expression (ints are ints, and floats are floats). This means pure integer arithmetics won't result in a float output column anymore.
- The list of allowed built-in functions has been updated to Python3 (unicode, unichr and cmp, which are Python2-only have been removed; ascii, format, list, map and set have been added)
- The tool will still fail unconditionally on invalid expressions (expressions that
  use words and characters outside the predefined set) and on syntax errors, but
  now also on NameErrors
- By default, the new version will fail if any field of any line cannot be cast to its expected type (e.g., a NA in an int column). Old behavior was to skip the line. Relying on expected types can be turned off for a complete run of the tool, in which case expressions have to handle type conversions explicitly.
- By default, the new version will fail on references to non-existent columns instead of skipping the line.
  This behavior can be turned off, in which case such invalid references will get treated like regular errors while computing the expression(s)
- For any errors arising during computation of an expression that haven't caused immediate failure, handling is now highly configurable (including the option to skip problematic lines like in previous versions), but the default is to fail the tool run.

fixed
====

- the list column type is handled correctly by the new version

tools/column_maker/column_maker.xml
```diff
@@ -1,101 +1,161 @@
-<tool id="Add_a_column1" name="Compute" version="1.6">
+<tool id="Add_a_column1" name="Compute" version="2.0">
     <description>an expression on every row</description>
```
nekrut, line 2: I would simplify description as "Compute on rows"

tools/column_maker/column_maker.xml
```diff
@@ -126,52 +301,136 @@ python '$__tool_directory__/column_maker.py'
 
 **What it does**
 
-This tool computes an expression for every row of a dataset and appends the result as a new column (field).
+This tool computes an expression for every row of a dataset and appends or inserts the result as a new column (field). Several expressions can be specified and will be applied sequentially to each row.
```
nekrut, line 304: This tool computes an expression on every row of a dataset and appends or inserts the result as a new column (field). Several expressions can be specified and will be applied sequentially to each row.

tools/column_maker/column_maker.xml
```diff
@@ -126,52 +301,136 @@ python '$__tool_directory__/column_maker.py'
...
+   chr1  151077881  151077918  2  200  -  True
+   chr1  151081985  151082078  3  500  +  True
+
+-----
+
+**Error handling**
+
+The tool will always fail on syntax errors in and other unrecoverable parsing
+errors with any of your expressions. For other problems, however, it offers
+quite some control over how they should be handled:
```
nekrut, line 358: control over how they should be handled:

tools/column_maker/column_maker.xml
```diff
@@ -126,52 +301,136 @@ python '$__tool_directory__/column_maker.py'
...
+
+1. The default for "Autodetect column types" is "Yes", which means the tool
+   will evaluate each column value as the type that Galaxy assumes for the
+   column. This default behavior will allow you to write simpler expressions.
+   The arithmetic expression "c4 * c5" from the first simple example,
+   for instance, works only because Galaxy realizes that c4 and c5 are integer
+   columns. Occasionally, however, this autodetection can cause issues. A common
```
nekrut, line 365: columns. Occasionally this autodetection can cause issues. A common

nekrut (approved):

tools/column_maker/column_maker.xml
wm75: Done, but I'm never sure about the implications for the tool panel.
@bgruening Is that an ok kind of change or will it force us to have yet another duplicated tool in our panel, or do you see other downsides?

bgruening (approved):

tools/column_maker/column_maker.py
```diff
@@ -308,12 +311,15 @@ def to_str(obj):
...
-                if args.fail_on_non_computable:
-                    sys.exit(
-                        'Could not compute a new column value using "%s" on '
-                        'line #%d: "%s".  Error was "%s"'
-                        % (ex, i, line, str(e))
-                    )
+                if args.skip_non_computable:
+                    # log that a line got skipped and why, then stop
+                    # computing for this line
+                    print(e)
```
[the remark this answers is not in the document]
wm75: You're right :)
I thought printing skipped lines was the original version's behavior, but it isn't, and doing this would have Galaxy store an unnecessarily huge amount of stdout info.

Thanks.

PR #4676, opened 2022-07-20
Wrapper for bcftools split-vep plugin

DM2zzion:
FOR CONTRIBUTOR:

Wrapper for the previously unwrapped bcftools plugin split-vep


tools/bcftools/bcftools_plugin_split_vep.xml
```diff
@@ -0,0 +1,113 @@
...
+
+@OUTPUT_TYPE@
+
+## Primary Input/Outputs
+@INPUT_FILE@
+## Plugin options
+-a '${a}'
+#if $c:
+    -c '${c}'
+#end if
```
wm75, line 35: Why the `#if` when -c seems to be required?

tools/bcftools/bcftools_plugin_split_vep.xml
```diff
@@ -0,0 +1,113 @@
...
+@EXCLUDE@
+@REGIONS@
+@TARGETS@
+
+@OUTPUT_TYPE@
+
+## Primary Input/Outputs
+@INPUT_FILE@
+## Plugin options
+-a '${a}'
```
wm75, line 32:
```suggestion
-a '$a'
```
remove the curly braces here and around the other parameters. They are required only in cases where cheetah wouldn't be able to figure out on its own where a variable name ends and something else begins. In simple cases like this one they only make things less readable.

wm75 (commented):
Thanks @DM2zzion, this looks really good already!
Just minor requests/questions from me.

Also, I know that you need this plugin for your work so I'm happy to get this merged even if the current PR should not offer all options that the command line tool has to offer.
In case some command line functionality is still missing (haven't checked) could you please state it here (e.g. as a TO DO in the wrapper) so that we remember to add it when we touch the wrapper next time?

tools/bcftools/bcftools_plugin_split_vep.xml
DM2zzion: At first I thought it's optional and when I changed it, I forgot that bit.

wm75 (approved):
Perfect, thanks for the changes!

Issue #4687, opened 2022-07-25
Magicblast : DB not found


eHirchaud:
## Bug description:

I Install  magic blast tool on local Galaxy instance.

I use local db like blast tool.

And I have this error message

`BLAST Database error: No alias or index file found for nucleotide database [/db/nt_blast/nt/blastdb] in search path [/srv/galaxy/jobs/003/3361/working::]`

* My blastdb.loc :

`nt      NCBI Partially non-redundant (nt)       /srv/db//nt_db/nt`

This works for ncbi_blast_plus

In magicblast.xml wrapper I see this line for db selection `'${os.path.join($db_opts.database.fields.path, "blastdb")}'`

This assumes that  index name  is allways blastdb.

gregvonkuster:
This should be fixed by this PR https://github.com/galaxyproject/tools-iuc/pull/4707

bernt-matthias:
@eHirchaud feel free to reopen if the problem persists.

PR #4707, opened 2022-08-08
magicblast: don't hard-code default db name

gregvonkuster:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/blast/macros.xml
+++ b/tools/blast/macros.xml
@@ -1,6 +1,6 @@
 <macros>
     <token name="@TOOL_VERSION@">1.6.0</token>
-    <token name="@VERSION_SUFFIX@">0</token>
+    <token name="@VERSION_SUFFIX@">1</token>
     <token name="@PROFILE@">20.09</token>
     <xml name="requirements">
         <requirements>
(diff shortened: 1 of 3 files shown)


gregvonkuster:
This should fix this issue https://github.com/galaxyproject/tools-iuc/issues/4687.

bernt-matthias (commented):
Thanks for working on this @gregvonkuster. 

I guess this will break old data table entries? We could check at run time whether

- $db_opts.database.fields.path or
-  '${os.path.join($db_opts.database.fields.path, "blastdb")}'

exist and adapt the parameter accordingly. 

But I'm also fine as is .. since I would consider it a bug fix.

gregvonkuster:
> I guess this will break old data table entries? 

I think this would only be a problem for those admins that installed blast databases only for the magicblast tool, but with no other blast tool installed.  And in this case the admin set the path value to enable the magicblast tool to function correctly, working around the bug in the tool.  I think this scenario would be a small percentage of installations, with the more usual scenario being the one that occurred that uncovered this issue.  So I think we're ok to not add the additional runtime check.

By the way, none of the older blast tools have any functional tests that cover cached blast databases (they only test for blast files from the history), so it was a bit confusing at the time as to how the path should be set, and that resulted in this bug.  Of course, hindsight makes it easy now.

> But I'm also fine as is .. since I would consider it a bug fix.

bernt-matthias (approved):

PR #4734, opened 2022-08-19
emboss implement missing parameters

bernt-matthias:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/emboss_5/emboss_sixpack.xml
+++ b/tools/emboss_5/emboss_sixpack.xml
@@ -1,4 +1,4 @@
-<tool id="EMBOSS: sixpack90" name="sixpack" version="@VERSION@+galaxy2">
+<tool id="EMBOSS: sixpack90" name="sixpack" version="@VERSION@.1+galaxy2">
   <!-- tool adds file description and timestamp to output data -->
   <description>Display a DNA sequence with 6-frame translation and ORFs</description>
   <expand macro="bio_tools" />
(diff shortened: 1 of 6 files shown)


bgruening:
This is not a tool suite so it will bump all tools. Are any other changes we want to include here? Should we bump all tool versions?

tools/emboss_5/emboss_textsearch.xml
```diff
@@ -1,12 +1,13 @@
...
   <description>Search sequence documentation. Slow, use SRS and Entrez!</description>
   <expand macro="bio_tools" />
   <macros>
     <import>macros.xml</import>
   </macros>
   <expand macro="requirements" />
   <code file="emboss_format_corrector.py" />
-  <command>textsearch -sequence '$input1' -outfile '$out_file1' -pattern '$pattern' -casesensitive -heading $heading -usa $usa -accession $accession -name $search_name -description $description -html
+  <command>textsearch 
+    -sequence '$input1' -outfile '$out_file1' -pattern '$pattern' -casesensitive $casesensitive -heading $heading -usa $usa -accession $accession -name $search_name -description $description -html
```
nsoranzo, line 10: Maybe split this with one option per line?

bernt-matthias: I agree that this would be nicer, but then we should do it consistently in all emboss tools. Don't think that I have the time for this.

bernt-matthias:
> This is not a tool suite so it will bump all tools. Are any other changes we want to include here? Should we bump all tool versions?

I have no more changes that I want to include. Don't know about bumping. Can you tell me?

tools/emboss_5/emboss_sixpack.xml
bgruening: why that?

bgruening:
Should we add profile versions? We update all tools no matter what ... so we could bump all tools and include a profile in all tools.

PR #4762, opened 2022-09-01
Add DRAM tools

gregvonkuster:
FOR CONTRIBUTOR:


tools/dram/dram_distill.xml
```diff
@@ -0,0 +1,109 @@
...
+        <data name="output_product_tsv" format="tabular" label="${tool.name} on ${on_string}: product_tsv"/>
+    </outputs>
+    <tests>
+        <!--
+        These tests succeed, but the outputs are empty because the DRAM
+        database is not available in the test environment. This warning
+        is displayed in the log...
+        warnings.warn('Database does not exist at path %s' % self.description_loc)
+        ...and this exception is raised:
+        ValueError: Invalid file path or buffer object type: <class 'NoneType'
```
bgruening, line 52: Should the tool fail in this case?
Please note that you can test for a tool error. You can define that a test should fail and then check the stderr

tools/dram/dram_merge_annotations.xml
```diff
@@ -0,0 +1,195 @@
...
+#if $rrnas_collection:
+    && test -f 'output_dir/rrnas.tsv' && mv 'output_dir/rrnas.tsv' '$output_rrnas' || echo 'No rrnas.tsv output produced'
+#end if
+#if $trnas_collection:
+    && test -f 'output_dir/trnas.tsv' && mv 'output_dir/trnas.tsv' '$output_trnas' || echo 'No trnas.tsv output produced'
+#end if
+&& test -f 'output_dir/annotations.tsv' && mv 'output_dir/annotations.tsv' '$output_annotations' || echo 'No annotations.tsv output produced'
+    ]]></command>
+    <inputs>
+        <param name="annotations_collection" type="data_collection" format="tabular" collection_type="list" label="Collection of annotation files"/>
```
bgruening, line 54: why is it not possible here to just take 2 single files

Is the assumption that dram_annotate is used on a collection and then produces collections? Before this tool is used.

Alternatively you could use multiple=true here

tools/dram/dram_neighborhoods.xml
```diff
@@ -0,0 +1,133 @@
+<tool id="dram_neighborhoods" name="DRAM find neighborhoods" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
+    <description>around genes of interest</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements"/>
+    <command detect_errors="exit_code"><![CDATA[
+#import gzip
```
bgruening, line 8: where is this needed?

tools/dram/dram_neighborhoods.xml
```diff
@@ -0,0 +1,133 @@
...
+            <param name="distance_genes" value="10"/>
+            <output name="output_neighborhood_annotations">
+                <assert_contents>
+                    <has_size value="2046495" delta="1000"/>
+                </assert_contents>
+                <assert_contents>
+                    <has_text text="neighborhood_number"/>
+                </assert_contents>
+            </output>
+       </test>
```
bgruening, line 79: indentation

tools/dram/dram_strainer.xml
```diff
@@ -0,0 +1,92 @@
+<tool id="dram_strainer" name="DRAM strain annotations" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
+    <description>down to genes of interest</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements"/>
+    <command detect_errors="exit_code"><![CDATA[
+#import gzip
```
bgruening, line 8: is this needed?

tools/dram/dram_distill.xml
gregvonkuster: Failing the tool would be tricky with the current DRAM version.  DRAM returns a 0 exit code when the database is missing, so we'd have to check for the database in the tool command line and fail if it's missing.

But DRAM accesses a config file stored on disk to locate the database (this is why these DRAM tools don't include data tables).  So to locate the database, we'd have to parse the output of this command which prints the config file.
```
$DRAM-setup.py print_config
```
The config file includes entries like this, so we'd have to parse the line and get the database location and then check to see if it exists.
```
KOfam db: /home/galaxy/tool-data/DRAM/1.3.5/kofam_profiles.hmm
```
To me this seems a bit brittle, but I'll do this if you feel it is best - let me know.

Also, I'll submit a ticket to the DRAM developers to return a 1 exit code if the database is missing - this is a better solution, but will  not be available until a later version.

tools/dram/dram_neighborhoods.xml
gregvonkuster: Ah, cruft from development.  Removed - thanks!

tools/dram/dram_neighborhoods.xml
gregvonkuster: Fixed - thanks!

tools/dram/dram_strainer.xml
gregvonkuster: Removed - thanks!

tools/dram/dram_merge_annotations.xml
bgruening: ping

tools/dram/dram_distill.xml
bgruening: @gregvonkuster where is this written to?

> ValueError: Invalid file path or buffer object type: <class 'NoneType'

You can with `stdio` filter stdout and raise an error level on your own if needed. (checking with the DRAM devs is a very good idea as well.)

Where is this config-file located? Can it be in the HOME or in the JWD directory? How do you think admins should configure the tool? If *.loc file is not possible we need a README I think.

tools/dram/dram_merge_annotations.xml
gregvonkuster: > why is it not possible here to just take 2 single files
> Alternatively you could use multiple=true here

The DRAM tools are very restrictive with inputs / parameters.  This tool requires a directory for the input and a directory for the output - files do not work in either case.   I even tried implementing the wrapper without using the * wildcard on the input directory and the underlying tool failed.
```
usage: DRAM.py merge_annotations [-h] [-i INPUT_DIRS] [-o OUTPUT_DIR]

options:
  -h, --help            show this help message and exit
  -i INPUT_DIRS, --input_dirs INPUT_DIRS
                        Path with wildcards pointing to DRAM annotation output directories (default: None)
  -o OUTPUT_DIR, --output_dir OUTPUT_DIR
                        Path to output merged annotations files (default: None)
```

tools/dram/dram_distill.xml
gregvonkuster: (checking with the DRAM devs is a very good idea as well.)

I've opened this issue https://github.com/WrightonLabCSU/DRAM/issues/209

gregvonkuster: > You can with `stdio` filter stdout and raise an error level on your own if needed. (checking with the DRAM devs is a very good idea as well.)

I've added this to dram_annotate and dram_distill, so tools and tests now fail when the database is missing.

The dram_neighborhoods and dram_strainer tools use the database only if the optional --categories param is used., but this param is broken in version 3.5.1 - https://github.com/WrightonLabCSU/DRAM/issues/194.

So I believe all tools are now covered for this.

PR #4766, opened 2022-09-05
fixing an issue with the tool by updating output files copying condit…

EngyNasr:
fixing an issue with the tool by updating output files copying conditions, where user get no error when no databases are provided, based on that no output means no error rather no output

Thanks a lot,
Engy

### DIFF
--- a/tools/srst2/srst2.xml
+++ b/tools/srst2/srst2.xml
@@ -98,6 +98,7 @@ srst2
 #if 'keep_interim_alignment' in str($output_files_selector)
     --keep_interim_alignment
 #end if
+## |true here is added in order not to search for a file that is not produced at all, such that if the user provided no gene/MLST databases or there are no outputs found, the tool will run successfully and only notify the user that no outputs are found
 #if 'report_new_consensus' in str($output_files_selector) and $use_gene_db.selector == "yes" and $use_mlst_db.selector == "yes"
 && mkdir -p allelesOutput/ && cp *.output__input.*.pileup allelesOutput | true
 #end if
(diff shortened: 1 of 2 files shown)


wm75:
I think this `| true` magic justifies an explanatory comment in the source code or future contributors will be very puzzled what this is about.

wm75:
In addition, you need to bump the wrapper version.

EngyNasr:
> In addition, you need to bump the wrapper version.

Can you please explain more for both comments please :)

gallardoalba:
> > In addition, you need to bump the wrapper version.
> 
> Can you please explain more for both comments please :)

You need to increase the version suffix https://github.com/galaxyproject/tools-iuc/blob/38b55875aabf987c81372688d8c7462afefd263b/tools/srst2/macros.xml#L4

tools/srst2/srst2.xml
```diff
@@ -97,17 +99,17 @@ srst2
     --keep_interim_alignment
 #end if
 #if 'report_new_consensus' in str($output_files_selector) and $use_gene_db.selector == "yes" and $use_mlst_db.selector == "yes"
-&& mkdir -p allelesOutput/ && cp *.output__input.*.pileup allelesOutput
+&& mkdir -p allelesOutput/ && cp *.output__input.*.pileup allelesOutput | true
 #end if
 #if $use_gene_db.selector == "yes" and  $use_gene_db.no_gene_details
-&& mkdir -p geneTypingOutput/ && cp output__genes__*__results.txt geneTypingOutput && cp output__fullgenes__*__results.txt geneTypingOutput
+&& mkdir -p geneTypingOutput/ && cp output__genes__*__results.txt geneTypingOutput | true && cp output__fullgenes__*__results.txt geneTypingOutput | true
```
gallardoalba, line 105: Also as Wolfgag suggested, it would be nice to explain why did you include the `| true` expressions.

gallardoalba (approved):

PR #4770, opened 2022-09-07
Kraken: Alot of things, first updating the tools version based on the updated…

EngyNasr:
1.  updating the tools version based on the updated conda package from 1.0.1 to 1.2.0, 
2.  based on that updated some arguments along with their test
3.  fixing the otu argument and adding a test for that. 
4. updating some of the output format based on the real tool's outputs formats

Only one thing missing the metadata test which cannot be tested unless the update in the conda package is merged after that the tool is totally successful with more arguments to use for following Phyloseq, Phinch and other metagenomics analysis tools

Thanks a lot,
Engy


tools/kraken_biom/kraken_biom.xml
```diff
@@ -78,24 +90,39 @@ kraken-biom
...
+                <has_text text="Biological Observation"/>
+            </assert_contents>
+        </output>
+    </test>
+    <test expect_num_outputs="2">
+        <param name="kraken_reports" value="MG_AIR20200707.tabular,MG_AIR20200708.tabular,MG_AIR20200709.tabular"/>
+        <param name="otu_fp" value="true"/>
         <param name="fmt" value="json"/>
-        <output name="biomOutput" ftype="json">
+        <output name="biomOutput" ftype="biom1">
```
bgruening, line 115: can we add here an biom2 test as well ... I guess biom2 is what most people will use?

Should we maybe put biom2 as default?

tools/kraken_biom/kraken_biom.xml
```diff
@@ -49,27 +56,32 @@ kraken-biom
...
             <option value="hdf5" selected="true">Biom2 (HDF5)</option>
         </param>
     </inputs>
     <outputs>
-        <data name="biomOutput" format="biom2" from_work_dir="./table.biom" label="Kraken biom output file">
+        <data name="biomOutput" format="tabular" from_work_dir="./table.biom" label="Kraken-biom output file">
             <change_format>
-                <when input="fmt" value="json" format="json" />
-            </change_format>
+                <when input="fmt" value="json" format="biom1" />
```
bgruening, line 72: json is biom1?

What is the default that is called tabular?

bernt-matthias:
If possible have also a look at the open PRs  and issues concerning kraken.

bgruening:
> If possible have also a look at the open PRs and issues concerning kraken.

That's unfortunately a completely different tool that only consumes kraken. I need to fix the kraken tool probably.

tools/kraken_biom/kraken_biom.xml
EngyNasr: Phinch only take Biom1

bernt-matthias: @EngyNasr test would still be welcome

@bgruening the default is `biom2`, or? See the `--fmt` parameter.

tools/kraken_biom/kraken_biom.xml
bernt-matthias: > What is the default that is called tabular?

Is there an example in the `test-data` directory?

tools/kraken_biom/kraken_biom.xml
EngyNasr: yeah default is HDF5 which is Biom2 and it is set in the parameter --fmt, but I just played with formats in the output section to suit the tool's real output

EngyNasr: I will add the test right now :)

tools/kraken_biom/kraken_biom.xml
EngyNasr: I created the test-data myself and they are all tested, but the tool itself didnot come with any test data

EngyNasr:
Just added the last test, really the tool is now perfectly working I tried it on my project datasets using the project metadata files, it worked and Phinch was finally able to visualize the output with its metadata. 

Kindly merge updates, i need the tool urgently 

Thanks a lot <3 
Engy

bgruening (approved):

PR #4777, opened 2022-09-12
Allow value -1 for --twopass1readsN

JasperO98:
The minimum value for --twopass1readsN is currently 1, however it should allow a -1 value. I don't know if it should allow a 0, but -1 is actually the default value, see the manual.

https://physiology.med.cornell.edu/faculty/skrabanek/lab/angsd/lecture_notes/STARmanual.pdf

FOR CONTRIBUTOR:

### DIFF
--- a/tools/rgrnastar/rg_rnaStar.xml
+++ b/tools/rgrnastar/rg_rnaStar.xml
@@ -313,7 +313,7 @@
             </when>
             <when value="Basic --twopass1readsN">
                 <param name="sj_precalculated" type="hidden" value="" />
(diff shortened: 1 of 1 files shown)


bernt-matthias (commented):
Is this parameter actually used in the command block? Could not find it.

I guess the idea should be to use it in the command block only if a value is set. Then we have the same behavior, ie the default of -1 is implicitly used if no value is given in the tool form.

JasperO98:
You are right! Sorry for the sloppy mistake.

Issue #4170, opened 2021-11-16
Tool request: riboWaltz


bgruening:
https://github.com/LabTranslationalArchitectomics/riboWaltz for Ribo-Seq.

pavanvidem:
@bgruening this can be closed

bgruening:
Thanks Pavan!

PR #4796, opened 2022-09-21
Add new tool riboWaltz

pavanvidem:
FOR CONTRIBUTOR:


pavanvidem:
Solves https://github.com/galaxyproject/tools-iuc/issues/4170

tools/ribowaltz/.shed.yml
```diff
@@ -0,0 +1,11 @@
+categories:
```
bgruening, line 1: since those are 2 tools, this needs to be suite defintion

pavanvidem: Both of them are based on functions from the same R package. It should fit to both.

bgruening: Yes, but we only want to have one tool per toolshed repo

bgruening:
Its green :)

bgruening:
> galaxy.tool_util.verify.interactor.JobOutputsError: Output collection 'out_plots': identifier 'read_ends_heatmap' found out of order, expected order of {3: 'metaplots', 2: 'psites_per_region', 1: 'read_ends_heatmap', 0: 'read_lengths', 4: 'trinucleotide_periodicity'} for the tool generated collection elements ['metaplots', 'psites_per_region', 'read_ends_heatmap', 'read_lengths', 'trinucleotide_periodicity']

That is the recent error

pavanvidem:
that's strange!. From the error report it seems the collection elements are sorted lexicographically but the order in the test is not. I name sorted the elements in the test. I don't know  if that makes sense.

bgruening (approved):

pavanvidem:
gracias!!

PR #4812, opened 2022-09-29
updating medaka version from 1.4.4 to the latest conda 1.7.1

EngyNasr:
updating medaka version from 1.4.4 to the latest conda 1.7.1 and adding a wrapper for SNP medaka tool

FOR CONTRIBUTOR:
* [yes] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [yes] - License permits unrestricted use (educational + commercial)
* [yes] - This PR adds a new tool or tool collection: Medaka SNP
* [yes] - This PR updates an existing tool or tool collection


tools/medaka/macros.xml
```diff
@@ -1,5 +1,5 @@
 <macros>
-    <token name="@TOOL_VERSION@">1.4.4</token>
+    <token name="@TOOL_VERSION@">1.7.1</token>
     <token name="@VERSION_SUFFIX@">1</token>
```
gallardoalba, line 3:
```suggestion
    <token name="@VERSION_SUFFIX@">0</token>
```

EngyNasr:
Shall we merge the PR

gallardoalba:
> Shall we merge the PR

I think Wolfgang needs to review it, since he knows this tool much better than me.

EngyNasr:
In the last commit I just added the new basecallers models added by the tool's new version

wm75 (commented):
Nice, thanks @EngyNasr!

The previous `medaka_variant` pipeline is not completely dead though but lives on in the form of  `medaka_haploid_variant` reflecting mostly its now restricted recommended usecase.
Would you mind adjusting the existing medaka_variant.xml wrapper accordingly instead of removing it altogether?

wm75:
@neoformit brought up the question of how to transition from `medaka_variant` to `medaka_haploid_variant` exactly in the wg-tools matrix room.
I guess I'd change only the name of the existing tool (i.e. leave the tool_id as is) in the existing wrapper. This way users will see only *medaka haploid variant pipeline* in the tool panel of instances that have the new version, but will be able to go back to previous ones via the versions menu.

tools/medaka/snp.xml
```diff
@@ -0,0 +1,168 @@
...
+        </test>
+    </tests>
+    <help><![CDATA[
+.. class:: infomark
+
+**What it does**
+
+@WID@
+
+The module *SNP* decodes probabilities.
```
wm75, line 140:
```suggestion
This module decodes probabilities to SNPs but NOT indels.

For a more general solution see the medaka *variant* tool.
```

neoformit:
It's worth noting that the `haploid_variant_pipeline` is significantly different, i.e. takes a different set of models and parameters. I've updated those in [my fork](https://github.com/neoformit/tools-iuc/blob/medaka-upgrade/tools/medaka/medaka_variant.xml) though it's not passing tests yet.

EngyNasr:
in the medaka_haploid_variant, the -i is different, the wrapper needs some changes to work

EngyNasr:
> in the medaka_haploid_variant, the -i is different, the wrapper needs some changes to work

I will suggest to remove this tool wrapper from the suit until we get a replacement for the fastx, but I updated the version again from 1.7.1 to 1.7.2

wm75:
@EngyNasr I think *fastx* simply means fastq *or* fasta. The pipeline that gets executed with the *medaka_haploid_variant* command is this one here: https://github.com/nanoporetech/medaka/blob/master/scripts/medaka_haploid_variant

I'm not sure what kind of error you got when trying to run it, but looking at the code I see that:
https://github.com/nanoporetech/medaka/blob/481a2839f019ad0dce032b9a655b5e309968ca18/scripts/medaka_haploid_variant#L123
requires a mini_align script. This refers to https://github.com/nanoporetech/pomoxis/blob/master/scripts/mini_align, which should be compiled into the conda env. If the error is about this command not being found you will have to dig through the conda env to see if it's there.

wm75:
That's an option for me, yes.
Would that be ok with you @neoformit? The old medaka_variant wrapper will stick around on servers that have it installed already and because @EngyNasr removed it from this version will stay at 1.4.4.

The new medaka_haploid_variant script is, in fact, an extremely thin wrapper around minimap, medaka consensus, variant and tools annotate, and it should be relatively simple to replace it with a WF using existing wrappers of minimap2 and this PR here.
With a true Galaxy WF you'd gain flexibility and efficient parallelization. It's only that command line users of the script will not find it in the tools panel, but would have to be pointed to the workflow.

neoformit:
Sounds good to me. If you this PR I can most likely make another one to add `medaka_haploid_variant` as I'm half way there already.

wm75 (approved):
Ok, then lets merge this. Thanks @EngyNasr!
@neoformit looking forward to your PR and happy to help you with medaka_haploid_variant.

PR #4830, opened 2022-10-04
Make IUC workflows reusable

bernt-matthias:
This PR adds:

1. Reusable Workflows which can be used in all IUC style repos
    - The hope/aim is that users of these workflows will need to update/sync their workflows less often. In particular the `galaxy-branch` will be updated automatically per default once its updated at IUC. Still users can fix the branch if they like. Also parameters for linting and testing are by default as they are at IUC, but can be overwritten.
    - Bugs in Galaxy (`tool-util`,...)  and `planemo` will impact more repos by default (if repos do not fix the for and branch). I would argue that this is good because it might help to detect bugs early.
    - [x] **TODO** it would be handy if we could do the switch from `master` to `main`, since called worflows apparently can't refer to the default branch. otherwise we would need to change the calling workflows again.
2. Adds `report-level` and `fail-level`
   - supersedes https://github.com/galaxyproject/galaxy-tool-repository-template/pull/8
3. Add `planemo-version` parameter to slash command which should allow to test planemo releases
    - usage, e.g. `planemo-version="planemo==0.74"`
3. Update all called actions
    - checkout v2 -> v3
    - cache v2 -> v3
    - upload-artifact v2 -> v3
    - download-artifact v2 -> v3
    - setup-python v1 -> v4
    - peter-evans/create-or-update-comment v1 -> v2
    - peter-evans/slash-command-dispatch v2 -> v3
4. Make used flake8 plugins configurable  fixes https://github.com/galaxyproject/tools-iuc/issues/4988
5. Execute slash command only if `PAT` is defined (only relevant for template repo .. once synced .. there the problem is that the slash workflow always fails until `PAT` is defined .. which is a bit annoying since it runs on every comment)

The PR can also be seen in action here: https://github.com/galaxyproject/galaxy-tool-repository-template/pull/11

TODOs:

- [ ] test to use from devteam / template repo (https://github.com/galaxyproject/galaxy-tool-repository-template/pull/11)
- [ ] decide if the called workflows should live at the IUC repo / template repo

FOR CONTRIBUTOR:


nsoranzo:
@bernt-matthias Can you take a look at the conflicts please?

bernt-matthias:
Will do. Seems that two PR were merged during my vacation: 

- https://github.com/galaxyproject/tools-iuc/pull/4892 the `set-output` things need to be incorporated here .. one moment
- https://github.com/galaxyproject/tools-iuc/pull/4850 I think I also updated all versions here .. I will recheck

bernt-matthias:
Reminder: 

- [ ] remove test bumps of tools before merging

Note that the WF does not run because of `error parsing called workflow "galaxyproject/tools-iuc/.github/workflows/wf_lint.yaml@master": failed to fetch workflow: workflow was not found.` .. If you like I could change it temporarily to the branch in my for to see it in action?

bernt-matthias:
Integrate https://github.com/peterjc/galaxy_blast/pull/151

mvdbeek (approved):
Awesome, let's do this!

bernt-matthias:
As discussed on gitter I will try to add the called workflows to the planemo-ci-action repo. Putting this to draft until it is done.

PR #4834, opened 2022-10-06
add qiime qzv visualisation extractor

bernt-matthias:
the qiime2 display application (view.qiime2.org) does not work for non-public Galaxy instances. This is a workaround.

qiime2 viz artifacts are just zip files which also contain the html + data + files (which is shown at the view website). So this tool is equivalent

If I find the time I will explore the possibility to do the same with a generic / viz plugin .. at least for so long (and maybe for workflows) the tool will be useful.

Thanks to @EngyNasr for the discussion and help.

FOR CONTRIBUTOR:


tools/qiime_extract_viz/qiime_extract_viz.xml
```diff
@@ -0,0 +1,78 @@
...
+        for nonjs in \$(cat '${viz.extra_files_path}'/*.html | grep '<script src="' | sed 's/^\s*<script src="//; s/".*//' | grep -v "\.js\$"); do
+            js=\$(echo "\$nonjs" | sed 's/\.[^.]\+$//').js &&
+            mv '${viz.extra_files_path}'"/\$nonjs" '${viz.extra_files_path}'"/\$js" &&
+            sed -i -e "s@\"\${nonjs}\"@\"\${js}\"@" '${viz.extra_files_path}'/*.html;
+        done &&
+
+        mv '${viz.extra_files_path}'/index.html '$viz'
+    ]]></command>
+    <inputs>
+        <param name="infile" type="data" format="qzv" label="QIIME visualization artifact" />
```
bernt-matthias, line 34: Qza?

bernt-matthias: Seem to contain only data. Metadata should be accessible also without the tool. So qzv should be sufficient.

tools/qiime_extract_viz/.shed.yml
```diff
@@ -0,0 +1,10 @@
+name: qiime_extract_viz
+owner: iuc
+description: Extract vizualization from QIIME artifacts
+long_description: |
+  TODO
```
pvanheus, line 5: This needs to be completed before merge.

tools/qiime_extract_viz/qiime_extract_viz.xml
```diff
@@ -0,0 +1,78 @@
...
+            </output>
+        </test>
+    </tests>
+    <help><![CDATA[
+
+.. class:: infomark
+
+**What it does**
+
+The tool extracts visualisation from QIIME artivarct datasets (qza, qzv).
```
pvanheus, line 65: Do you mean "QIIME artifacts"?

bernt-matthias: Indeed. My bad 😅

tools/qiime_extract_viz/.shed.yml
bernt-matthias: Thanks for checking carefully. I would have forgotten about this.

bernt-matthias:
Put this on draft .. want to finish the discussion with the qiime2 developers first https://github.com/qiime2/galaxy-tools/issues/38#issuecomment-1272332723

shiltemann:
@bernt-matthias I just wrote this exact same tool, then found this PR. What are your thoughts on this currently? Should we try to get this in? 

The "display at qiime2view" links are still broken (https://github.com/galaxyproject/galaxy/issues/17658) in Galaxy, pasting shared dataset URLs in qiime2view manually is also broken due to CORS settings, so currently no good way to view qzv files in/from Galaxy. Would be nice if the qiime wrappers output html outputs such as these by default, but until then this at least provides a way for users now to view qiime qzv outputs?

bernt-matthias:
> I just wrote this exact same tool, then found this PR. What are your thoughts on this currently? Should we try to get this in?

Sorry for that. We should have merged this a long time ago. Lets go ahead. 

Any comments from your side?

shiltemann (approved):

shiltemann:
@bernt-matthias ok great, just tested your tool locally and looks good to me :+1:

PR #4838, opened 2022-10-10
cutadapt name collusion

tuncK:
Added a check against name collusion between two input files. This becomes a problem if the tool is part of a workflow that provides a list of list of read pairs. The current version does not generate an error or warning, but rather generates inaccurate results by overwriting input1 with input2.

FOR CONTRIBUTOR:
* [Y] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [Y ] - License permits unrestricted use (educational + commercial)
* [] - This PR adds a new tool or tool collection
* [Y] - This PR updates an existing tool or tool collection
* [] - This PR does something else (explain below)


wm75 (commented):
A good catch!

I'm wondering though if there's a good reason to deal with element_identifiers at all here and why not to call the symlinks reads1.* and reads2.*?

wm75:
@mblue9 seems you've added the use of element identifiers in https://github.com/galaxyproject/tools-iuc/commit/1fdba6d4837aa529684df4eef443e70866df433b - do you remember why that is important?

wm75:
Aha, worked my way back through #1828 and the linked https://github.com/galaxyproject/tools-iuc/pull/1821#issuecomment-381175523.
So MultiQC takes the last word it finds in Cutadapt's "Command line parameters" line in the Report output and uses that as a sample name. So there *is* a reason for the element_identifier juggling even if not in the tool itself.
If we're still doing things in such an exotic way in 2022, then I have no objections against the fix here.

wm75:
@tuncK in any case you will have to bump the @VERSION_SUFFIX@ in the macros file to create a new tool wrapper version.

bgruening (approved):

bgruening:
thanks

PR #4847, opened 2022-10-17
Disable sanitizer for map_param_value expression tool

mvdbeek:
Sanitization is not necessary for expression tools, and all upstream/downstream tools pass through the regular parameter sanitization.

FOR CONTRIBUTOR:


tools/map_param_value/map_param_value.xml
```diff
@@ -1,10 +1,16 @@
-<tool name="Map parameter value" id="map_param_value" version="0.1.0" tool_type="expression" profile="22.01">
+<tool name="Map parameter value" id="map_param_value" version="0.1.1" tool_type="expression" profile="22.01">
     <macros>
         <xml name="when_element" tokens="type_selection">
-            <param name="input_param" type="@TYPE_SELECTION@" optional="true" label="Map this parameter value to a different value"/>
+            <param name="input_param" type="@TYPE_SELECTION@" optional="true" label="Map this parameter value to a different value">
```
lldelisle, line 4:
```suggestion
            <param name="input_param" type="@TYPE_SELECTION@" optional="true" label="Value to map to a different one">
```

lldelisle: I don't know why but when I used the tool I found the label confusing.

mvdbeek: How about 
```suggestion
            <param name="input_param" type="@TYPE_SELECTION@" optional="true" label="Value to map">
```
?

lldelisle: Perfect

lldelisle:
I approve

gallardoalba (approved):

wm75:
The javascript expression gets evaluated client-side only?

mvdbeek:
No, in a javascript sandbox, and inputs to that scope are just plain dicts with primitive types.

wm75:
Also, I may have missed the discussion about this, but why are `$input_param` and `mappings.from` optional params?

mvdbeek:
There are 2 reasons, one is that you could include None/null as input (a.k.a optional input parameter), and otherwise we need to set a default value (that's a limitation of Galaxy tool parameters), which doesn't really make sense.

wm75:
So in case I leave both unset to map the optional param not provided by the user to some other value:

1. am I not just replicating an optional param *with* a default (i.e. I should have done that at the WF input param level?
2. what gets actually compared in
    https://github.com/galaxyproject/tools-iuc/blob/eedbc5b3f712eaea1ac35ac9a771cbe5431d09bd/tools/map_param_value/map_param_value.xml#L39? Can there ever be a match between the null object and it's string repr, and, if it can I could also map it explicitly using that string as the `mappings.from` value?

mvdbeek:
> 1. am I not just replicating an optional param _with_ a default (i.e. I should have done that at the WF input param level?

I think you can do a lot of things here much more complex than that. For starters you could have parse_param_value step before this one, so you don't know what the input is going to be, or what the default is going to be.

> 2\. Can there ever be a match between the null object and it's string repr

Ah, yes, that won't work, should be easy to fix though. This needs one more test.

> 2\. could also map it explicitly using that string as the `mappings.from` value

Are you asking if you can map to null? That's the idea, but let me add a test

mvdbeek:
Alright, I've made the two from/to parameters required. Unfortunately Galaxy's tool state for text parameters doesn't discriminate between empty string and None/null. So instead of optional parameters we can go with `""` as the default value. For the input parameter I still think it makes sense to have it be optional, so you can connect it to an optional parameter.

wm75:
> Justed started playing with it and I don't think the current comparison:
> 
> https://github.com/galaxyproject/tools-iuc/blob/eedbc5b3f712eaea1ac35ac9a771cbe5431d09bd/tools/map_param_value/map_param_value.xml#L39
> 
> is working for anything but text inputs.

Sorry, it is actually working for ints (and likely floats), but not for booleans it seems.

wm75:
There is also an issue with the "Use unmodified input ..." option since this can lead to violation of the output type if you're mapping from one type to another.

mvdbeek:
Could you include what values you've checked ?

mvdbeek:
> There is also an issue with the "Use unmodified input ..." option since this can lead to violation of the output type if you're mapping from one type to another.

Can you explain what violation this is ? 'bla' for instance gets turned into true, but that's fine ?

mvdbeek:
> The default `value=''` also doesn't seem to work for ints and floats.

How so ? I mean it's silly that a required parameter needs a default value, but that's Galaxy for you.

tools/map_param_value/map_param_value.xml
```diff
@@ -144,6 +174,19 @@ return { output: coerceToOutput(source) };
             </conditional>
             <param name="output_param_text" value_json="&quot;D&quot;" />
```
nsoranzo, line 175:
```suggestion
            <output name="output_param_text" value_json="&quot;D&quot;" />
```

tools/map_param_value/map_param_value.xml
```diff
@@ -144,6 +174,19 @@ return { output: coerceToOutput(source) };
...
+                <repeat name="mappings">
+                    <param name="from" value="A" />
+                    <param name="to" value="B" />
+                </repeat>
+            </conditional>
+            <conditional name="unmapped">
+                <param name="on_unmapped" value="default"/>
+                <param name="default_value" value="default"/>
+            </conditional>
+            <param name="output_param_text" value_json="&quot;default&quot;" />
```
nsoranzo, line 188:
```suggestion
            <output name="output_param_text" value_json="&quot;default&quot;" />
```

and similarly below for more output_param_text, output_param_float and output_param_boolean

tools/map_param_value/map_param_value.xml
```diff
@@ -118,6 +135,19 @@ return { output: coerceToOutput(source) };
...
+                <conditional name="input_param_type">
+                    <param name="input_param" value="A" />
+                    <repeat name="mappings">
+                        <param name="from" value="A" />
+                    </repeat>
+                </conditional>
+                <output name="output_param_text" value_json="null" />
+            </test>
+        -->
         <test expect_num_outputs="1" expect_failure="true">
```
nsoranzo, line 151:
```suggestion
        <test expect_failure="true">
```

mvdbeek: What's the reason for that ?

nsoranzo: I don't see much of a point in checking the number of outputs when the job is expected to fail, but no problem if you prefer to keep it.

wm75:
> > There is also an issue with the "Use unmodified input ..." option since this can lead to violation of the output type if you're mapping from one type to another.
> 
> Can you explain what violation this is ? 'bla' for instance gets turned into true, but that's fine ?

If, e.g., your input datatype is text and you'rr mapping to boolean, but the input text doesn't match any of your mappings, then with this option 'bla' for example will be returned unchanged, i.e., as text.

wm75:
> > The default `value=''` also doesn't seem to work for ints and floats.
> 
> How so ? I mean it's silly that a required parameter needs a default value, but that's Galaxy for you.

Yeah, but Galaxy won't let you leave the param blank for int and float types, but will insist on a compatible value. => You can't map the null value case for an optional int or float param.

wm75:
> > Justed started playing with it and I don't think the current comparison:
> > https://github.com/galaxyproject/tools-iuc/blob/eedbc5b3f712eaea1ac35ac9a771cbe5431d09bd/tools/map_param_value/map_param_value.xml#L39
> > 
> > is working for anything but text inputs.
> 
> Sorry, it is actually working for ints (and likely floats), but not for booleans it seems.

I've checked with a boolean input param and mapping that to text, but no matter the value of the input param or how I toggle the mappings.from switch, I never get a match.

mvdbeek:
> Yeah, but Galaxy won't let you leave the param blank for int and float types, but will insist on a compatible value. => You can't map the null value case for an optional int or float param.

You can with a default, everything else is unfortunately just not possible in Galaxy.

mvdbeek:
> I've checked with a boolean input param and mapping that to text, but no matter the value of the input param or how I toggle the mappings.from switch, I never get a match.

I've added more tests for this in https://github.com/galaxyproject/tools-iuc/pull/4847/commits/70aaafd4eab720efac9f5848bcc44b759d8300ce

mvdbeek:
> If, e.g., your input datatype is text and you'rr mapping to boolean, but the input text doesn't match any of your mappings, then with this option 'bla' for example will be returned unchanged, i.e., as text.

I've added a test for that in https://github.com/galaxyproject/tools-iuc/pull/4847/commits/32d9b7616e749c919be17fc790afce65efdedf9b

mvdbeek:
I'm happy to keep working on bugs too, but can we get this in so we can make progress on the workflows ? All I wanted to do here is disable the sanitizer :)

wm75:
Fine with me :-) I start losing the overview in this PR already.

mvdbeek:
Haha, yeah, I think I could make the point for expressions being a lot easier to write (and understand IMO) just on the complexities you have to deal with when any input and any output type are possible ... .

mvdbeek:
And thanks for the thorough review @wm75 and @nsoranzo, https://github.com/galaxyproject/tools-iuc/pull/4847/commits/08f8e89022290cdbec19db5341d418656cd62b19 was an issue (that I didn't catch because I didn't use the right test output tag)

PR #4848, opened 2022-10-19
Add feature to Arriba to let it control the STAR mapping

JasperO98:
FOR CONTRIBUTOR:

With the current STAR implementation in Galaxy the `--alignSJstitchMismatchNmax` and `--chimoutType` parameters are missing, which are recommended parameters for Arriba. All recommended parameters for Arriba are available, see [documentation](https://arriba.readthedocs.io/en/latest/workflow/). To solve this I've added a STAR-Fusion like option to Arriba. With this feature the user can provide reads to Arriba and it maps it automatically with the optimal parameters. This problem could also be resolved by adding the parameter not included yet in STAR. This could be added later by someone if they want. To me STAR is too big of a program to change, so I would rather not touch it.


JasperO98:
Not sure what is going wrong here? What does Lint R scripts do?
Any help would be appreciated.

nsoranzo:
> Not sure what is going wrong here, because the test succeed on my local machine? What does Lint R scripts do? Any help would be appreciated.

This was caused by a change in the GitHub action we use to lint R scripts, it's now fixed by https://github.com/galaxyproject/tools-iuc/pull/4850 , sorry for the inconvenience. Can you rebase or merge the master branch?

JasperO98:
@nsoranzo could you merge this pull request?

shiltemann:
So, so after discussing with @JasperO98, I have a question to IUC:

Downside of letting Arriba run STAR, is that servers will now have to reserve large nodes for this tool, which will cause it to have to wait longer, even if the user doesn't choose the option to run STAR as part of this tool.

As I understand it, the two main advantages here are that it is a bit faster this way, and that Arriba can choose its own settings for STAR to optimize its own performance (and these STAR parameters are not currently exposed in the STAR wrapper). So we can:

- Merge this, it is no different than say STAR-fusion I guess
- Update the STAR wrapper to expose the missing parameters and advise users in Arriba's help text to run star with these parameters

Thoughts?

nsoranzo:
>     * Update the STAR wrapper to expose the missing parameters and advise users in Arriba's help text to run star with these parameters

I think I prefer this option, since arriba doesn't actually run star as part of its script. And having the extra options available in the standalone STAR wrapper is a good thing any way.

bgruening:
> Downside of letting Arriba run STAR, is that servers will now have to reserve large nodes for this tool, which will cause it to have to wait longer, even if the user doesn't choose the option to run STAR as part of this tool.

With some advanced job configuration, this can be eliminated so that we only allocate memory when Star is run.
I agree with @nsoranzo that exposing this option in Star is a good way and enhancing the help - if in addition we merge this PR here I do not mind.

shiltemann:
oh, didn't realise jobs could be configured based in individual parameter settings, that takes away that concern then :+1: 

@JasperO98 would you be up for making the changes to the STAR wrapper as well?

JasperO98:
Sure I can do it. Thanks for the feedback!

JasperO98:
Parameters have been added in #4889.

bgruening:
@JasperO98 @shiltemann do we still want to get this in, or is the other solution ok?

bernt-matthias:
Closing this here. @JasperO98 @shiltemann (and anyone else) feel free to reopen if needed. Thanks for the contributions @JasperO98

PR #4852, opened 2022-10-20
Re-add the PIMA pv data manager

gregvonkuster:
FOR CONTRIBUTOR:

### DIFF
--- a/data_managers/data_manager_pima_pv_downloader/.shed.yml
+++ b/data_managers/data_manager_pima_pv_downloader/.shed.yml
@@ -1,0 +1,13 @@
+name: data_manager_pima_pv_downloader
+owner: iuc
+description: Data manager to download the PIMA plasmids_and_vectors.fasta files for use by the PIMA pipeline tools.
+homepage_url: https://github.com/appliedbinf/pima_md
+long_description: |
+  PIMA (Plasmid, Integrations, Mutations, and Antibiotic) resistance annotation pipeline is a high-throughput
+  sequence analysis pipeline. PIMA is an end-to-end solution encompassing all the steps required to transform
+  raw ONT output into high-quality annotated assemblies. PIMA supports providing optional Illumina paired end
+  reads for polishing and error correcting assemblies.
+remote_repository_url: "https://github.com/galaxyproject/tools-iuc/tree/master/data_managers/data_manager_pima_pv_downloader"
+type: unrestricted
+categories:
+- Data Managers
--- a/data_managers/data_manager_pima_pv_downloader/data_manager/pima_pv_downloader.py
+++ b/data_managers/data_manager_pima_pv_downloader/data_manager/pima_pv_downloader.py
@@ -1,0 +1,60 @@
+import argparse
+import datetime
+import json
+import os
+import sys
+from urllib.request import Request, urlopen
+
+
+def url_download(url, workdir):
+    file_path = os.path.abspath(os.path.join(workdir, os.path.basename(url)))
+    src = None
+    dst = None
+    try:
+        req = Request(url)
+        src = urlopen(req)
+        with open(file_path, 'wb') as dst:
+            while True:
+                chunk = src.read(2**10)
+                if chunk:
+                    dst.write(chunk)
+                else:
+                    break
+    except Exception as e:
+        sys.exit(str(e))
+    finally:
+        if src:
+            src.close()
+    return file_path
+
+
+def download(url, out_file):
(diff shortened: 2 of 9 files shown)


gregvonkuster:
@bgruening can this be merged?  Thanks!

bgruening:
What does this DM doing?
Would it not be better to have a very generic - give me random fasta files - DM?

gregvonkuster:
> What does this DM doing?
> Would it not be better to have a very generic - give me random fasta files - DM?

It's downloading a fasta file, but a fairly specific one that consists of plasmids and vectors that are used in the [PIMA: Plasmid, Integrations, Mutations, and Antibiotic resistance annotation pipeline] (https://github.com/galaxyproject/tools-iuc/pull/4852#pima-plasmid-integrations-mutations-and-antibiotic-resistance-annotation-pipeline).  

This particular fasta file is updated regularly, so it is dated when installed into Galaxy, and my intent with they PR is to make it available to appropriate tools via the new "pima_pv" data table.  I will soon be opening PRs for tools that are currently contained within this PIMA pipeline that will make use of this fasta file via this data table

I'm not sure where the idea to make these kinds of data manager tools more generic originates, but it is an interesting idea.  It would obviously eliminate some tools and code, but it would place the responsibility of choosing the correct fasta file on the user.  Is this something that the IUC has been pursuing?

bgruening:
> I'm not sure where the idea to make these kinds of data manager tools more generic originates, but it is an interesting idea. It would obviously eliminate some tools and code, but it would place the responsibility of choosing the correct fasta file on the user. Is this something that the IUC has been pursuing?

We can add a special column to the data table that can be identified by your tools, isn't it.

gregvonkuster:
> > I'm not sure where the idea to make these kinds of data manager tools more generic originates, but it is an interesting idea. It would obviously eliminate some tools and code, but it would place the responsibility of choosing the correct fasta file on the user. Is this something that the IUC has been pursuing?
> 
> We can add a special column to the data table that can be identified by your tools, isn't it. Means in your PIMA tools you filter on that column and your users will only see those files.

Until now, column selections have been on things like dbkey, so I'm not clear on how this generic table approach would work.  What do you want the data table name to be and what do you want the column names to be?  What would the values of this new special column be?

bgruening:
You can filter on any column with various filters: https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-inputs-param-options-filter

E.g. `static_value` as already used here could be useful: https://github.com/galaxyproject/tools-iuc/blob/d7751920c666de56e21cf48ffe19191b2d19d534/tools/gemini/gemini_macros.xml#L45

A name for the column could be "group" or "class"? @wm75 you have a similar DM for annotation files do you think this is a good idea?

gregvonkuster:
> > I'm not sure where the idea to make these kinds of data manager tools more generic originates, but it is an interesting idea. It would obviously eliminate some tools and code, but it would place the responsibility of choosing the correct fasta file on the user. Is this something that the IUC has been pursuing?
> 
> We can add a special column to the data table that can be identified by your tools, isn't it. Means in your PIMA tools you filter on that column and your users will only see those files.

In some cases filters have non-optimal side-effects (e.g., I've experienced the select list in the workflow tool form changed to a text box when a filter is used).

gregvonkuster:
> You can filter on any column with various filters: https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-inputs-param-options-filter
> 
> E.g. `static_value` as already used here could be useful:
> 
> https://github.com/galaxyproject/tools-iuc/blob/d7751920c666de56e21cf48ffe19191b2d19d534/tools/gemini/gemini_macros.xml#L45
> 
> A name for the column could be "group" or "class"? @wm75 you have a similar DM for annotation files do you think this is a good idea?

I can see some pros for doing this, but I also see some cons, and I'm not sure the pros outweigh the cons.  

A con with keeping these similar tools separate as we've done until now is more similar tools.  But these simple data manager tools require little to no ongoing maintenance.  At most, some update may be required if the Python http request code is somehow changed.  

But making a more complex tool that will accommodate many data table downloads could potentially incur more maintenance overhead, especially if any of the supported data sources change their process for downloading the data.

I don't have strong opinions either way on this, so other input is welcome.  However, my plate is so full that I can't see myself finding time in the short term to enhance this tool as discussed.  I need this tool as soon as possible, but I can wait a bit longer if someone else wants to take on this enhancement.

bgruening (approved):

gregvonkuster:
Closing this PR since we decided a more generic FASTA data manager downloader is needed.

PR #4867, opened 2022-10-24
Updating tools/pangolin from version 4.1.2 to 4.1.3

planemo-autoupdate:

I have updated tools/pangolin from version 4.1.2 to 4.1.3.


wm75:
This new version of the tool only fixes an issue with pip-installs of pangolin. Conda-installs are not affected and there's no functional change so we don't need an update in Galaxy.

PR #4910, opened 2022-11-08
Fix python linting errors for table_compute

bernt-matthias:
- tests 20 and 21 were suddenly failing in CI 
  - failure is legit, but I don't know which galaxy change triggered it
- not sure what triggers the failing test 41 (it appears that the test is supposed to fail, but it is not recognized)

FOR CONTRIBUTOR:

### DIFF
--- a/tools/table_compute/scripts/table_compute.py
+++ b/tools/table_compute/scripts/table_compute.py
@@ -57,7 +57,7 @@ class Utils:
                         for col in data.columns]
         data.index = [row.strip() if type(row) is str else row
                       for row in data.index]
-        return(data)
+        return data
 
     @staticmethod
     def rangemaker(tab):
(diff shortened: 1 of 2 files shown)


wm75:
Tests 20 and 21 are also *supposed* to fail since `[]:` are indeed not allowed in expressions.
So in fact all tests (and only those) of this tool that expect failure are behaving differently from before. My guess is that it's because of the combination `expect_num_outputs="1" expect_failure="true"`, which doesn't make much sense but probably hasn't always been checked for?

bernt-matthias:
Thanks for pointing this out @wm75 .. overlooked this. 

Let me bisect this.

mvdbeek:
Alright, it's green again! Thanks for the report and fix @bernt-matthias!

bernt-matthias:
Cool. Then lets just merge the python linter fixes?

mvdbeek (approved):

PR #4914, opened 2022-11-10
Bandage upgrade

neoformit:
Upgrade bandage tool to use the new `bandage_nc` package. This [fork of Bandage](https://github.com/asl/BandageNG) is better maintained with performance upgrades.

~~N.B. failing file size limits, but these files are already in IUC.~~

N.B. Test output (PNG/SVG/JPG) sizes vary a lot - apparently the tools is no longer deterministic.


tools/bandage/bandage_image.xml
```diff
@@ -69,7 +69,7 @@ Bandage
             <param name="output_format" value="png"/>
             <output name="outfile" ftype="png">
               <assert_contents>
-                  <has_size value="2752" delta="1000"/>
+                  <has_size value="2752" delta="4000"/>
```
bgruening, line 72: the filesize is only 2700 und the delta is 4000 are we sure those files are generated or correct?

neoformit: No, I was tired last night and this probably doesn't make sense :joy:

bgruening: I'm getting tired now :) haha

Thanks for working on all of that Cam!

neoformit: All good, just wanna make sure this gets through :)

bernt-matthias (dismissed):
Good upgrade, but I would prefer to rename the tool as well.

neoformit:
> Good upgrade, but I would prefer to rename the tool as well.

@bernt-matthias  Like this?

```xml
<tool id="bandage_image" name="BandageNG Image" version="@TOOL_VERSION@+galaxy4" profile="21.05">
```

bernt-matthias:
More like 

```xml
<tool id="bandage_ng_image" name="BandageNG Image" version="@TOOL_VERSION@+galaxy4" profile="21.05">
```

Making a complete new suite.

bgruening:
@bernt-matthias I think it's fine as it is. The old tool is not maintained and the new is a friendly fork, for the user the functionality does not change and keeping it will make workflow updates and tutorials easier to migrate.

neoformit:
Ok so new name, same ID? Or same name too? @bgruening @bernt-matthias

bgruening:
I would keep the ID as it is.

neoformit:
Agreed, but currently the tools have been changed to `name="BandageNG Image"`. So it can be merged like this or I can roll the name back too. I don't think a name change has implications for workflows etc?

bgruening:
The name has no implications. I will wait for @bernt-matthias to merge maybe he has strong feelings about the ID.

bernt-matthias:
You convinced me. We can keep the ID and name if you like.

neoformit:
@bgruening @bernt-matthias what's the action here? Do you want to merge with `name="BandageNG *"` or revert to `name="Bandage *"`?

neoformit:
Also - Gareth has pointed out that the input data for Bandage should probably be `gfa` datatype, not `tabular`. Any thoughts on that? Or why it was done this way originally?

bgruening:
I'm fine with keeping the name as it is now. Gfa looks more accurate isn't it. Or is anything else supported?

bernt-matthias:
> Also - Gareth has pointed out that the input data for Bandage should probably be gfa datatype, not tabular. Any thoughts on that? Or why it was done this way originally?

Galaxy seems to support fastg and gfa (as gfa1 or gfa2). Not sure what the others are. Probably we need to keep `txt` to allow for the others. Still listing `fastg,gfa[12],txt` would increase usability a bit (since the user sees what is allowed in the form field .. in addition to the help text).

Also lets revert name and id of the tool to the original values.

bgruening (approved):

PR #4919, opened 2022-11-12
Improve vapor wrapper

wm75:
FOR CONTRIBUTOR:


tools/vapor/vapor.xml
```diff
@@ -12,6 +12,9 @@
         <requirement type="package" version="@TOOL_VERSION@">vapor</requirement>
     </requirements>
     <command detect_errors="exit_code"><![CDATA[
+    #if int($fasta_file.metadata.data_lines) != 2 * int($fasta_file.metadata.sequences)
+        echo "This tool can only parse FASTA input that has each actual sequence on a single line." 1>&2; exit 1
```
bgruening, line 16: This should be possible as a data validator isn't it?

wm75: There's no way to do the comparison in a validator because expressions are only available for text params.

wm75:
The failing lint I've expected from local testing and deserves an issue.

tools/vapor/vapor.xml
bernt-matthias: True. Might be an option to convert any given fasta file into a single-line-fasta, e.g. using awk, sed, or python?

wm75: Yes, I've been considering that briefly but then wasn't feeling like implementing it.
Now on second thought that might actually be a good idea, in particular, because one could do it *only* in this if block.

wm75:
Remembered an idea I had forgotten to implement here. Done now :)

bgruening (approved):

kciy (approved):
Useful changes that ease debugging and allow the user to see more output in case of errors. Thank you!

PR #4926, opened 2022-11-14
Update ncbi datasets

bernt-matthias:
Recover https://github.com/galaxyproject/tools-iuc/pull/4886 (which was deleted for some reason .. I only marked it as draft while working on it).

FOR CONTRIBUTOR:

### DIFF
--- a/tools/ncbi_datasets/.shed.yml
+++ b/tools/ncbi_datasets/.shed.yml
@@ -5,6 +5,6 @@ long_description: |
   NCBI Datasets is a new resource that lets you easily gather data from across NCBI databases. Find and download gene, transcript, protein and genome sequences, annotation and metadata.
 name: ncbi_datasets
 owner: iuc
-homepage_url: https://www.ncbi.nlm.nih.gov/datasets/
+homepage_url: https://github.com/ncbi/datasets
 remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/master/tools/ncbi_datasets
 type: unrestricted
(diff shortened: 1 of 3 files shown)


bernt-matthias:
I guess this removes the "virus-functionality". Since I removed the `excludes_virus_protein` macro (and all the include/exclude arguments which are now covered by `--include`

Downloading virus data seems now to work with `datasets download virus [genome|protein]`

mvdbeek:
This is awesome, thanks @bernt-matthias!

bernt-matthias:
Two questions @mvdbeek: 

- is the removal of the virus functionality fine (it moved to a separate subcommand)? Or should I consider to readd it as separate tool? 
- To download the data of multiple `taxon` I created an auxiliary workflow using an expression tool. This will then create a parallel download job for each genome. Might this be sanctioned by NCBI (aborted downloads)? I see two alternatives
  - is it possible to limit the number of parallel jobs in galaxy thereby I could serialize the processing of the downloads
  - I could change the tool such that it uses a bash for loop for downloading multiple `taxon` (eg given in a file)

mvdbeek:
> * is the removal of the virus functionality fine (it moved to a separate subcommand)? Or should I consider to readd it as separate tool?

That's fine I think. It used to be a separate tool (I had implemented everything, but it was decided to go one tool at a time), and we only kept the macro. I guess at this point the options have changed; starting from scratch is probably easier than picking this back up.

I think the batch/parallel question is something we can't really answer, let's keep it as it is for now ?

mvdbeek (approved):

bernt-matthias:
> let's keep it as it is for now ?

Fine with me. 

Unsure about the test failures. All tests pass locally for me.

mvdbeek:
> All tests pass locally for me.

Same, could be the container again ... although then I'd expect everything to fail.

PR #4939, opened 2022-11-23
Add ggupset

hexylena:
FOR CONTRIBUTOR:

image

not 100% sure what to label the axes since this is in our case specifically for DEGs, but can be equally useful for other cases. We could expose those to configuration I guess but maybe it's not necessary?


bgruening:
@hexylena that looks super useful. Thanks. I think the label are good for a first version - but offering them does not hurt.

hexylena:
Ok, I will add the parameters then. Thanks for the feedback!

tools/ggupset/ggupset.xml
```diff
@@ -0,0 +1,46 @@
...
+		<data name="upset_degree" format="png" from_work_dir="upset-degree.png" label="${tool.name} on ${on_string}: Upset by degree" />
+		<data name="commonalities" format="tabular" from_work_dir="upset.tsv" label="${tool.name} on ${on_string}: Identified Overlapping Groups" />
+	</outputs>
+	<tests>
+		<test>
+			<param name="input" value="HEP3B,HUH,JHH6,JHH7,SNU"/>
+			<param name="n_int" value="40"/>
+			<param name="width" value="1000"/>
+			<param name="height" value="800"/>
+			<output name="upset_degree" file="upset-degree.png" ftype="png" compare="sim_size" delta="10000"/>
```
bernt-matthias, line 38: Use a size assertion. Then we do not need to save the files.

hexylena:
sure can do. This would be something good to lint on.

hexylena:
I will pay someone money (it's worth like 200€ to me, for the lifetime cumulative annoyance), if they can just extract the test error into a github bot comment here.

Just so I don't have to
1. go to failed action
2. go to summary because the action doesn't include it
3. download a zip file
4. which I unpack for a single html
5. which I can then use to view the report.

bernt-matthias (approved):

bgruening:
Thanks Helena!

hexylena:
Danke Björn, Matthias!

PR #4942, opened 2022-11-24
PyEGA3 Downloaded vcf.gz files are recognized as vcf format

JasperO98:
FOR CONTRIBUTOR:

Currently, when downloading multiple files from the EGA i.e., as a collection `vcf.gz` files are recognized as `vcf` instead of `vcf_bgzip`. Before everything would be interpreted as a `data` format, which was fixed in #4841. However, now it's the same problem but different wrongly recognized format. This causes problems when using the VCFs in tools downstream. In #4841 I simply added the `.vcf_bgzip` extension to `vcf.gz` files. I thought they are still not recognized properly because the `vcf.gz` extension is still in the filename. However, this is not the case because when I test the download with `planemo serve` it still recognizes the `vcf.gz` files as `vcf`? How can I let Galaxy recognize them properly any tips? When downloading a single `vcf.gz` file it does recognize it as `vcf_bgzip`. So I'm afraid something is wrong with `pattern="__designation_and_ext__"` in this output line:
```
<collection name="downloaded_file_collection" type="list" label="${tool.name} on ${on_string}: Downloaded datasets">
    <filter> action['action_type'] == 'download_files' </filter>
    <discover_datasets pattern="__designation_and_ext__" recurse="true" directory="downloads" />
</collection>
```


JasperO98:
As you can see test 8 fails, which downloads a collection of `vcf.gz` files, but it detects it as `vcf`. Are there any solutions?

hexylena (approved):

PR #4947, opened 2022-11-25
Update bakta to 1.6.0 and database options

pimarin:
- Update Bakta tool to the last tool version
- Add a filter option to filter available bakta database


pimarin:
In the last update maybe, I will add the amrfinderplus database option when the DM will be updated

pimarin:
?

pimarin:
I finished to implement test to the last version and also I pre-added the choise of database from bakta-db data manager and amrfinderplus data manager

bgruening (approved):

pimarin:
thank you to your patience !

oschwengers:
Thank you so much @pimarin for all the effort you've put into this, also thanks @bgruening for all your support!

May I kindly ask you to fast-track this PR to [v1.6.1](https://github.com/oschwengers/bakta/releases/tag/v1.6.1) which is a patch release fixing some minor but annoying bugs? Then we can cut corners skipping an additional PR for 1.6.0->1.6.1

tools/bakta/macro.xml
```diff
@@ -1,5 +1,7 @@
+
 <macros>
-    <token name="@TOOL_VERSION@">1.5.0</token>
+    <token name="@TOOL_VERSION@">1.6.0</token>
```
bernt-matthias, line 3:
```suggestion
    <token name="@TOOL_VERSION@">1.6.1</token>
```

tools/bakta/bakta.xml
```diff
@@ -92,6 +94,9 @@
         #if "skip_cds" in $workflow.skip_analysis
             --skip-cds
         #end if
+        #if "skip_pseudo" in $workflow.skip_analysis
```
bernt-matthias, line 97: Just for the fun of it:

```
#echo " ".join(["--"+s.replace("_", "-") for s in $workflow.skip_analysis])
```

might replace all this. 

Alternatively, if you rename the values of the select such that they are equal to the command line options then this would simplify to 

```
#echo " ".join($workflow.skip_analysis)
```

bernt-matthias (dismissed):

bernt-matthias:
> May I kindly ask you to fast-track this PR to [v1.6.1](https://github.com/oschwengers/bakta/releases/tag/v1.6.1)

Lets see if this just works out.

@pimarin let us know if this is finished from your side

tools/bakta/bakta.xml
```diff
@@ -49,9 +53,7 @@
             gram type, prodigal/protein file
         ======================================*#
         $annotation.complete
-        #if $annotation.prodigal
```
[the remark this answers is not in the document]
oschwengers: No. It was deactivated due to a bug that, however, was fixed by 1.6.1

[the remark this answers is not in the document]
bernt-matthias: OK. Then lets keep it.

bernt-matthias (changes requested):
@pimarin 

could you restore the argument? 

the tests fail because output contains versions and dates .. they will be successful if a few lines of difference are allowed: `lines_diff`. could you adapt this?

bgruening:
Great work everyone. This work is very well received by the community. On EU we already have more than 1000 tool runs.

pimarin:
I will change all thank's to your comments !

tools/bakta/bakta.xml
pimarin: this is so good !
I updated it !

thank you !

pimarin:
I take comments and update for database new options (amrfinderplus and bakta DM)
I also change code with your suggestion @bernt-matthias

bgruening (approved):
Good from my side! Awesome work @pimarin!

tools/bakta/test-data/TEST_3/TEST_3.log
```diff
@@ -42,13 +43,14 @@ annotation summary:
 	oriCs/oriVs: 0
 	oriTs: 0
 
-export annotation results to: /tmp/tmpmnqj1xog/job_working_directory/000/6/working
```
bernt-matthias, line 45: One more minor thing. There is quite a lot of test data that is only changed because of differences in: 

- version
- dates
- paths

I would prefer if this would be reverted and in the test differences would be allowed by `lines_diff`.

bernt-matthias: The advantage is that it (hopefully) simplifies future updates.

Also, differences due to date or paths will just fail the weekly CI test running next sunday.

pimarin:
Thank you to you work also @bernt-matthias for all my PR !

pimarin:
I update following your comments, 
- I changed number of output but keeping the toolbox, just by default their is a minimum of 4 files

bernt-matthias (commented):
Seems that you restored the changes to the test data (probably you run `planemo test` with `--update-test-data` and just commit everything. 

Can you undo this?

bernt-matthias (approved):
I removed the unneeded test file changes again. 

Also I removed the svg test data .. its always better to use a size assertion instead of a sim_size comparison.

Also removed unneeded lines_diff from outputs that show no difference.

bernt-matthias:
Thanks a lot @pimarin.

PR #4965, opened 2022-11-30
Update bamtools and fix Split BAM by Reference

wm75:
FOR CONTRIBUTOR:

The changes to the split by reference tool include:
- The tool is now guaranteed to produce a collection element per sequence mentioned in the input BAM header to make it more useful in workflows. Previously, sequences without reads mapped to them did not result in a collection element. With the change, the input BAM header will be used as the content for an "empty" element.
- The tool will now respect the order of sequences in the input BAM also when the user selects a subset of references to split on, i.e., this fixes a bug in the previous version.
- The tests got updated and should now be more stringent.
- The tool now declares profile="20.09", which made it possible to simplify the generated command line by using a galaxy.json file for output discovery and to test for collection element sort order.


tool_collections/bamtools/bamtools_split/bamtools_split_ref.xml
```diff
@@ -1,32 +1,49 @@
...
+          for i in \$REFS_FROM_BAM_METADATA;
+          do FN=`printf "split_bam.REF_%s.bam" "\$i"`;
+            if [ ! -f \$FN ]; then cp header.bam "\$FN"; fi;
+          done)
+        && cp '$c1' galaxy.json
+        && cat galaxy.json
+    ]]></command>
+    <environment_variables>
+        <environment_variable name="REFS_FROM_BAM_METADATA">#import re
+#set $ref_list = [$re.sub('^.*__sq__(.+)__sq__.*$','\\1',n) if n.find('__sq__') >= 0 else n for n in str($input_bam.metadata.reference_names).split(',')]
```
wm75, line 26: I would really love to get rid off this weirdness here and in the configfile section, which I copied over from the previous version.
I fail to see when it would ever be useful to do something special in case of `__sq__` in a reference name and likely this is just some historic cruft.
I would like to get another opinion on this first though.

bgruening: I do not know, but I trust your judgment. Remove it.

wm75: Ahh, turns out that this dance is done because a bam dataset's `metadata.reference_names` is a Galaxy-sanitized string representation of the list of names like `__ob____sq__chr1__sq__, __sq__chr2__sq____cb__`.
The re.sub is trying to re-extract the original ref names from that.

Not sure this is intended, @mvdbeek?

bgruening (approved):

wm75:
@mvdbeek do you still want to comment on the strange formatting of `metadata.reference_names`?
Otherwise this seems mergeable.

PR #4966, opened 2022-11-30
Small fix for shape option in ggplot2

npinter:
Just a small fix for ggplot2. When selecting "Plot multiple groups of data on one plot" at "Plotting multiple groups" in the advanced options the shape selection was ignored.

### DIFF
--- a/tools/ggplot2/ggplot_point.xml
+++ b/tools/ggplot2/ggplot_point.xml
@@ -52,8 +52,8 @@ names(input)[$yplot] <- "ycol"
         gg_point = geom_point(size=1, alpha=1, gg_factor)
         gg_line = geom_line(size=1, alpha=1, gg_factor)
     #else
-        gg_point = geom_point(size=$adv.type_conditional.points.size, alpha=$adv.type_conditional.points.alpha, gg_factor)
-        gg_line = geom_line(size=$adv.type_conditional.points.size, alpha=$adv.type_conditional.points.alpha, gg_factor)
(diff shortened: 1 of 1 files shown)


bgruening:
Can you please bump?

tools/ggplot2/ggplot_point.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="ggplot2_point" name="Scatterplot with ggplot2" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="20.01">
+<tool id="ggplot2_point" name="Scatterplot with ggplot2" version="@TOOL_VERSION@+galaxy2" profile="20.01">
```
bernt-matthias, line 1: Would keep the VERSION_SUFFIX. Guess the autoupdate bot may get confused. Then we will need to bump all tools.
Maybe merge with https://github.com/galaxyproject/tools-iuc/pull/4907?

bgruening: Yes, I wanted to avoid bumping all tools. Merging with the other PR is a good idea.

bernt-matthias: > Yes, I wanted to avoid bumping all tools.

OK. Do you know how/if the autoupdate bot will handle this (some tools have version suffix other galaxyX)?

bernt-matthias:
Was this https://github.com/galaxyproject/tools-iuc/pull/4973 the same?

npinter:
no, but I'll close the PR and resubmit it

PR #4967, opened 2022-12-01
Updating tools/metaphlan from version 3.0.14 to 4.0.3

tuncK:
FOR CONTRIBUTOR:
* [ *] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [ *] - License permits unrestricted use (educational + commercial)
* [ *] - This PR updates an existing tool or tool collection


tools/metaphlan/metaphlan.xml
```diff
@@ -123,9 +123,15 @@ metaphlan
...
     --bt2_ps '$inputs.in.mapping.bt2_ps'
     --min_mapq_val $inputs.in.mapping.min_mapq_val
+    #if $ext == "sam"
+        --nreads \$(cat '$file_path' | grep -v '^@' | wc -l)
+    #end if
 #else
     '$inputs.in.in'
     --input_type '$inputs.in.selector'
+    #if $inputs.in.selector == "sam"
+        --nreads \$(cat '$inputs.in.in' | grep -v '^@' | wc -l)
```
bgruening, line 133: nread is ignore when -t is not spcified, is this correct?

You can probably use `grep -c -v` to count without `wc`

bgruening (commented):
I see that the old test size was also over he IUC limit. Do you think its possible to reduce the bowtie index below 1MB, or do we need an exception here?

tools/metaphlan/macros.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <macros>
-    <token name="@TOOL_VERSION@">3.0.14</token>
+    <token name="@TOOL_VERSION@">4.0.3</token>
     <token name="@VERSION_SUFFIX@">0</token>
     <token name="@PROFILE@">20.01</token>
```
bgruening, line 5: you can increase the profile to 21.01 if you like

tuncK:
It does not seem to be possible to reduce the bt2l sizes, it is still about the same size even if I include only 1 sequence, so as if an exception is needed (or is there a way to make the test function invoke a custom preprocessing step somehow?)

bgruening:
Not yet, but we will work on that. I'm happy to merge and push it manually if you answer/fix the other two comments. 

Great work @tuncK!

tools/metaphlan/metaphlan.xml
tuncK: nreads is to be used only if the input type is *.sam, ignored otherwise.

Good suggestion, will switch to grep -cv.

bgruening:
I will push it over the line so you can use it next week, thanks a lot!

PR #4970, opened 2022-12-01
Update samtools_coverage.xml

o-william-white:
Corrected arguments for --rf and --ff

FOR CONTRIBUTOR:

### DIFF
--- a/tool_collections/samtools/samtools_coverage/samtools_coverage.xml
+++ b/tool_collections/samtools/samtools_coverage/samtools_coverage.xml
@@ -22,11 +22,11 @@
 
             #set $filter = $additional_options.required_flags
             @FLAGS@
-            -rf $flags
+            --rf $flags
 
             #set $filter = $additional_options.skipped_flags
             @FLAGS@
-            -ff $flags
+            --ff $flags
 
             #if $condition_histogram.histogram_select == "yes"
                 -m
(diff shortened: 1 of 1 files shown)


o-william-white:
Hello, 

I was running samtools coverage today and I hit an error that I think is related to the input arguments being specified as -ff and -rf rather than --ff and --rf. 

Think this is a simple fix

Best wishes
OIlie

o-william-white:
image

bernt-matthias:
Maybe use the long options:

```
--rf, --incl-flags STR|INT
Required flags: skip reads with mask bits unset [null]

--ff, --excl-flags STR|INT
Filter flags: skip reads with mask bits set
```

Would also look more readable.

o-william-white:
> Maybe use the long options

Done

o-william-white:
Is there anything else I need to do for this? Or is it good to go? @bernt-matthias

bernt-matthias:
The original problem was that `samtools coverage` does not have a non zero exit code for the wrong argument. Therefore Galaxy did not realize the error (consequently the test data files are all empty).

@o-william-white I guess the following should be done to finish this:

- update the test data to make tests pass (use eg planemo's `--update_test_data` flag)
- remove the `from_workdir` from the output since the tool writes directly to `$output`
- could you file an upstream issue that samtools should always return a non-zero exit code for missing or wrong options?

PR #5008, opened 2022-12-16
Ivar annotation as input added

tuncK:
FOR CONTRIBUTOR:


tools/ivar/ivar_consensus.xml
```diff
@@ -1,10 +1,10 @@
-<tool id="ivar_consensus" name="ivar consensus" version="@VERSION@+galaxy0">
+<tool id="ivar_consensus" name="ivar_consensus" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
```
bgruening, line 1: I would not change the name. The name is exposed to the user - and what does a underscore tells the user?

tools/ivar/ivar_removereads.xml
```diff
@@ -1,12 +1,10 @@
-<tool id="ivar_removereads" name="ivar removereads" version="@VERSION@+galaxy3" profile="21.01">
+<tool id="ivar_removereads" name="ivar_removereads" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
     <description>Remove reads from trimmed BAM file</description>
     <macros>
         <import>macros.xml</import>
     </macros>
-    <expand macro="requirements">
-        <requirement type="package" version="3.8.1">python</requirement>
```
bgruening, line 7: python is only a dependency for this tool, not the others isn't it?

Moving it to the macro will make it a dependency fo all tools.

tools/ivar/ivar_trim.xml
```diff
@@ -1,13 +1,10 @@
-<tool id="ivar_trim" name="ivar trim" version="@VERSION@+galaxy4" profile="21.01">
+<tool id="ivar_trim" name="ivar_trim" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
     <description>Trim reads in aligned BAM</description>
     <macros>
         <import>macros.xml</import>
     </macros>
-    <expand macro="requirements">
-        <requirement type="package" version="3.8.1">python</requirement>
-        <requirement type="package" version="1.15">samtools</requirement>
```
bgruening, line 8: Is samtools not needed anymore?

tuncK: The tool depends on samtools 1.18 (?), so I dont see why we need to add in an earlier version explicitly.

tools/ivar/ivar_consensus.xml
tuncK: Will fix, copy-paste accident.

tools/ivar/ivar_removereads.xml
tuncK: 3 of 5 tools need python, I thought it would be easier to maintain.

tools/ivar/ivar_trim.xml
bgruening: So the conda packages as a samtools dependency? Then we can remove it here.

tools/ivar/ivar_removereads.xml
bgruening: its ok, we can do that, less containers I guess.

tools/ivar/macros.xml
```diff
@@ -1,9 +1,13 @@
+<?xml version="1.0"?>
 <macros>
-  <token name="@VERSION@">1.3.1</token>
+  <token name="@TOOL_VERSION@">1.3.1</token>
+  <token name="@VERSION_SUFFIX@">3</token>
```
wm75, line 4: This does not make sense. There should not be one version suffix for all wrappers since they do not always get updated all at once. In fact, this change would *downgrade* ivar_trim.

tools/ivar/ivar_removereads.xml
wm75: It's ok, but also no big gain, and Python is only a requirement for those subcommands where @pvanheus and me added significant extra logic to the command line tool, i.e., it's not a very native requirement.

tools/ivar/ivar_removereads.xml
```diff
@@ -31,10 +29,10 @@
     <inputs>
         <param name="input_bam" argument="-i" type="data" format="bam"
         label="Bam input"
-        help="BAM dataset, preprocessed with ivar trim, to remove reads from" />
+        help="BAM dataset, preprocessed with ivar trim, to remove reads from"/>
```
wm75, line 32: I would prefer not having all these cosmetic changes. It takes a lot of time inspecting such a non-minimal diff and understand what the *functional* change is in the PR.

tools/ivar/ivar_variants.xml
```diff
@@ -1,19 +1,21 @@
...
+<tool id="ivar_variants" name="ivar variants" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
     <description>Call variants from aligned BAM file</description>
     <macros>
         <import>macros.xml</import>
     </macros>
-    <expand macro="requirements">
-        <requirement type="package" version="3">python</requirement>
-    </expand>
-    <expand macro="version_command" />
+    <expand macro="requirements"/>
```
wm75, line 6: Ha, self-defeating argument above ;)
This wrapper relying on ivar's shipped samtools for something non-trivial. Haven't written this wrapper though, but @pvanheus did. Imo, this use here should require a pinned samtools version, but I'm open for other views.

tools/ivar/ivar_variants.xml
```diff
@@ -28,26 +30,27 @@
         <param name="ref" type="data" format="fasta" label="Reference"/>
         <param name="min_qual" argument="-q" type="integer" min="1" value="20" label="Minimum quality score threshold to count base"/>
         <param name="min_freq" argument="-t" type="float" min="0" max="1" value="0.03" label="Minimum frequency threshold"/>
+        <param name="gtf" argument="-g" optional="true" type="data" format="gff" label="Coordinates of ORFs, in GFF3 format (optional)"/>
```
wm75, line 33: Have you tested using this option when asking for VCF output. Not sure the wrapper will be able to handle this adequately.

tuncK: Yes, tests 4 &5 check this (or I think they do so)

tools/ivar/test-data/zika/Z52_a.vcf
```diff
@@ -22,7 +22,7 @@ PRV	2922	.	C	T	.	PASS	DP=275;REF_DP=264;REF_RV=0;REF_QUAL=36;ALT_DP=11;ALT_RV=0;
 PRV	3148	.	Y	T	.	PASS	DP=1707;REF_DP=0;REF_RV=0;REF_QUAL=0;ALT_DP=1324;ALT_RV=264;ALT_QUAL=36;AF=0.77563
 PRV	3148	.	Y	C	.	PASS	DP=1707;REF_DP=0;REF_RV=0;REF_QUAL=0;ALT_DP=381;ALT_RV=75;ALT_QUAL=36;AF=0.223199
 PRV	3295	.	A	G	.	PASS	DP=1040;REF_DP=1002;REF_RV=1002;REF_QUAL=35;ALT_DP=38;ALT_RV=38;ALT_QUAL=33;AF=0.0365385
-PRV	5680	.	C	T	.	PASS	DP=35;REF_DP=27;REF_RV=5;REF_QUAL=44;ALT_DP=8;ALT_RV=2;ALT_QUAL=46;AF=0.228571
+PRV	5680	.	C	T	.	PASS	DP=35;REF_DP=27;REF_RV=10;REF_QUAL=44;ALT_DP=8;ALT_RV=3;ALT_QUAL=46;AF=0.228571
```
wm75, line 25: This might actually be an illustration of what I'm talking about above since the changed count could come from a slightly different pileup produced by the latest version of samtools.

wm75 (changes requested):
Sorry for the many comments, but ivar being a core tool suite for some of our most important workflows needs to be handled extra-carefully.

tools/ivar/ivar_variants.xml
wm75: Aha, so test 4 shows that it's working with tabular output and test 5 would actually demonstrate that annotation is non-functional for VCF (cause you didn't have to change that test file).

It doesn't make much sense then to reveal this param for VCF output, but it should be a suboption for tabular only.
VCF files are better annotated after the fact with e.g. SnpEff anyway.

tuncK: Good idea, done.

tools/ivar/macros.xml
tuncK: Did not think of it, fixed.

tools/ivar/test-data/zika/Z52_a.vcf
tuncK: I am not sure if this can explain the change in output. The deployed version uses samtools without specifying its version (https://github.com/galaxyproject/tools-iuc/blob/8487d2c73793be0afa5b34388b122e686ac8a094/tools/ivar/ivar_variants.xml#L7). Only trimming does. Regardless, I added a samtools v1.16 to the entire container to avoid such risks.

tools/ivar/ivar_variants.xml
```diff
@@ -35,19 +37,22 @@
                 <option value="tabular_and_vcf">Both Tabular and VCF</option>
             </param>
             <when value="vcf">
-                <param argument="--pass_only" type="boolean" truevalue="--pass_only" falsevalue="" label="In VCF only output variants that PASS all filters" />
+                <param argument="--pass_only" type="boolean" truevalue="--pass_only" falsevalue="" label="In VCF only output variants that PASS all filters"/>
             </when>
             <when value="tabular_and_vcf">
-                <param argument="--pass_only" type="boolean" truevalue="--pass_only" falsevalue="" label="In VCF only output variants that PASS all filters" />
+                <param argument="--pass_only" type="boolean" truevalue="--pass_only" falsevalue="" label="In VCF only output variants that PASS all filters"/>
+                <param name="gtf" argument="-g" optional="true" type="data" format="gff" label="Coordinates of ORFs, in GFF3 format (optional)"/>
```
wm75, line 44: It would be good to have some help here to inform the user what these coordinates will be used for.

wm75 (changes requested):

wm75:
While you're at it: could you polish the output handling of ivar variant a bit.
The current way of copying the .tsv file to its final location and of leeting the python script write the VCF to the final destination results in rather long command lines.
A modern, better way would be to use `from_work_dir` in the output section and pick up variants.tsv and variants.vcf files by their names.

tools/ivar/macros.xml
```diff
@@ -1,18 +1,22 @@
+<?xml version="1.0"?>
 <macros>
-  <token name="@VERSION@">1.3.1</token>
+  <token name="@TOOL_VERSION@">1.3.1</token>
+  <token name="@PROFILE@">22.05</token>
```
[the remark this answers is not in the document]
tuncK: Can you name a suitable profile? 20.01? 

I am aware that newer profile causes additional problems, so I tested from 19.?? up to 22.05. I thought it was backwards-compatible, as other people had asked me to bump it in the past.

tools/ivar/ivar_variants.xml
```diff
@@ -1,26 +1,27 @@
...
         samtools mpileup -A -d 0 --reference ref.fa -B -Q 0 sorted.bam | ivar variants 
         -p variants
         -q $min_qual
-        -t $min_freq 
+        -t $min_freq
         #if str($output_format.choice) == "tabular" or str($output_format.choice) == "tabular_and_vcf"
-            && cp variants.tsv '$output_variants_tabular'
+            #if str($gtf) != 'None':
+                -r ref.fa
+                -g $gtf
```
wm75, line 18: This needs quoting.

tools/ivar/macros.xml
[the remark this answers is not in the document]
wm75: `ivar_trim`, in the previous version, declared 21.01 as the profile version so seems reasonable to just require that everywhere.

New profile versions usually bring improvements with them so we're trying to update the requirements from time to time, but they are a requirement on the Galaxy server side, and asking for 22.05 for a tool that doesn't benefit from it just seems rather aggressive.

tools/ivar/ivar_variants.xml
```diff
@@ -35,50 +36,69 @@
...
-            <when value="tabular" />
         </conditional>    
     </inputs>
     <outputs>
-        <data name="output_variants_tabular" format="tabular" label="${tool.name} tabular output on ${on_string}">
+        <data name="output_variants_tabular" from_work_dir="./variants.tsv" format="tabular" label="${tool.name} tabular output on ${on_string}">
             <filter>output_format['choice'] == 'tabular' or output_format['choice'] == 'tabular_and_vcf'</filter>
             <actions>
-                <action name="column_names" type="metadata" default="REGION,POS,REF,ALT,REF_DP,REF_RV,REF_QUAL,ALT_DP,ALT_RV,ALT_QUAL,ALT_FREQ,TOTAL_DP,PVAL,PASS,GFF_FEATURE,REF_CODON,REF_AA,ALT_CODON,ALT_AA" />
+                <action name="column_names" type="metadata" default="REGION,POS,REF,ALT,REF_DP,REF_RV,REF_QUAL,ALT_DP,ALT_RV,ALT_QUAL,ALT_FREQ,TOTAL_DP,PVAL,PASS,GFF_FEATURE,REF_CODON,REF_AA,ALT_CODON,ALT_AA"/>
```
wm75, line 54: when the $gtf option is in use, there are additional "column_names", aren't they?

tools/ivar/ivar_variants.xml
```diff
@@ -35,50 +36,69 @@
...
             </conditional>
             <output name="output_variants_tabular" file="zika/Z52_a.tsv" ftype="tabular" lines_diff="10"/>
             <output name="output_variants_vcf" file="zika/Z52_a_pass.vcf" ftype="vcf"/>
         </test>
-
+        <test expect_num_outputs="1">
+            <param name="input_bam" value="zika/Z52_a.masked.sorted.bam"/>
+            <param name="ref" value="zika/db/PRV.fa"/>
+            <param name="gtf" value="zika/db/PRV.gff"/>
+            <output name="output_variants_tabular" file="zika/Z52_a_annotated.tsv" ftype="tabular" lines_diff="10"/>
```
wm75, line 89: why the lines_diff in this and the following test's output check?

tools/ivar/ivar_variants.xml
tuncK: Done

tools/ivar/ivar_variants.xml
tuncK: No. They are already included in the current version: GFF_FEATURE,REF_CODON,REF_AA,ALT_CODON,ALT_AA

If you dont provide a gtf, these are filled with "NA". We could perhaps remove these 5 columns from non-gtf to save storage, but then some user-specific codes that rely on the number of columns might break.

tools/ivar/ivar_variants.xml
tuncK: I copied it from the existing code and dont know why exactly it was needed :) I eliminated.

tuncK:
All tests work here, but 1 test times out on Github for some reason, maybe a rerun is all it needs? ("Timed out after 900.25 seconds waiting on tool test run.")

bgruening:
restarted the tests :)

tuncK:
Restarting the test indeed did the trick, thanks Björn.

@wm75, all good now?

wm75 (approved):

wm75:
Thanks @tuncK for a good number of small improvements that make this tool suite better :heart:

PR #5033, opened 2023-01-05
Add tool hifiasm meta

TomHarrop:
FOR CONTRIBUTOR:


TomHarrop:
Hi folks! The CI test may fail because this tool uses a lot of RAM. It used 17 GB for this tiny dataset on my computer. Can it still be added to tools-iuc?

bgruening (commented):
Does the tool crash with a specific error code or error message? Some out of memory error or so? If so the you can define the test Inna way that you expect this error and grep for the error message or the error code.

Thanks @TomHarrop

TomHarrop:
Thanks @bgruening   

The test only fails on the GitHub PR workflow for this repo. It passes locally with planemo test. Setting `expect fail` would instead cause the test to fail everywhere else. 

The error running on GitHub is `Killed 9`. I assume this was because it tried to use too much RAM but I don’t know how much is available on the GitHub runners. I benchmarked the exact same command with the same data locally without Galaxy and Planemo, and it used 17 GB of RAM.

bgruening:
@TomHarrop have you seen this "For small genomes, use -f0 to disable the initial bloom filter which takes 16GB memory at the beginning." From https://github.com/chhylp123/hifiasm - would that work for a small test?

If not what about we test the command line and check that this looks valid? In addition, we add a real test and test this locally.

TomHarrop:
@bgruening thanks, I had *not* seen that! It's a hifiasm option but it works for hifiasm_meta too.

I've made the change but GH wouldn't let me re-open this PR, so I've created a new one here: https://github.com/galaxyproject/tools-iuc/pull/5059

PR #5037, opened 2023-01-06
Add UCSC faToVcf tool

### LINKED ISSUE #5026: Requested tool UCSC faToVcf
A tool to extract VCF from a multisequence FASTA alignment.

I start writing the wrapper.

See [bioconda package](https://bioconda.github.io/recipes/ucsc-fatovcf/README.html)

kciy:
FOR CONTRIBUTOR:

I didn't find any citation-

closes #5026


tools/ucsc_tools/fatovcf/fatovcf.xml
```diff
@@ -0,0 +1,150 @@
+<tool id="fatovcf" name="faToVcf" version="@TOOL_VERSION@+galaxy0" profile="21.05" license="MIT">
+    <description>
+        Convert a FASTA alignment file to Variant Call Format (VCF) single-nucleotide diffs
+    </description>
+    <macros>
+        <token name="@TOOL_VERSION@">426</token>
+    </macros>
+    <xrefs>
+        <xref type="bio.tools">ucsc-fatovcf</xref>
```
nsoranzo, line 9:
```suggestion
        <xref type="bio.tools">UCSC_Genome_Browser_Utilities</xref>
```

tools/ucsc_tools/fatovcf/fatovcf.xml
```diff
@@ -0,0 +1,150 @@
...
+            <param argument="-vcfChrom" type="text" optional="true" label="Use this sequence for the CHROM column in the VCF" help="Default: name of the reference sequence." />
+        </section>
+    </inputs>
+    <outputs>
+        <data name="out" format="vcf" from_work_dir="out.vcf" />
+    </outputs>
+    <tests>
+        <test expect_num_outputs="1"> <!-- default params -->
+            <param name="in_fasta" value="input.fa" />
+            <output name="out" ftype="vcf" file="out1.vcf" compare="sim_size" delta="50" />
```
nsoranzo, line 103:
```suggestion
            <output name="out" ftype="vcf" file="out1.vcf" lines_diff="4" />
```

I haven't tested, but unless the output has some randomness, the only differences should be on 2 header lines (which makes 2 additions and 2 deletions, i.e. 4 lines diff).

tools/ucsc_tools/fatovcf/fatovcf.xml
```diff
@@ -0,0 +1,150 @@
...
+        <param argument="-minAf" type="float" min="0.0" max="1.0" value="0.0" label="Minimum allele frequency" help="Ignore alternate alleles observed in less than F of non-N bases. Default: 0.0" />
+
+        <section name="output" title="Output VCF options" expanded="true">
+            <param argument="-includeRef" type="boolean" truevalue="-includeRef" falsevalue="" label="Include the reference in the genotype columns" help="Default: omitted as redundant (false)" />
+            <param argument="-noGenotypes" type="boolean" truevalue="-noGenotypes" falsevalue="" label="Output 8-column VCF" help="VCF without the sample genotype columns. Default: false" />
+            <param argument="-vcfChrom" type="text" optional="true" label="Use this sequence for the CHROM column in the VCF" help="Default: name of the reference sequence." />
+        </section>
+    </inputs>
+    <outputs>
+        <data name="out" format="vcf" from_work_dir="out.vcf" />
```
nsoranzo, line 98:
```suggestion
        <data name="out" format="vcf" />
```

tools/ucsc_tools/fatovcf/fatovcf.xml
```diff
@@ -0,0 +1,150 @@
...
+    #end if
+    #if $output.vcfChrom
+        ln -sfn '$output.vcfChrom' vcfChromFile.vcf &&
+    #end if
+    #if $in_fasta
+        ln -sfn '$in_fasta' input.fa &&
+    #end if
+    faToVcf
+        input.fa
+        out.vcf
```
nsoranzo, line 27:
```suggestion
        '$out'
```

tools/ucsc_tools/fatovcf/fatovcf.xml
```diff
@@ -0,0 +1,150 @@
...
+            <param name="minAc" value="1" />
+            <param name="minAf" value="0.1" />
+
+            <section name="output">
+                <param name="includeRef" value="true" />
+                <param name="noGenotypes" value="true" />
+                <param name="vcfChrom" value="sample1" />
+            </section>
+
+            <output name="out" ftype="vcf" file="out2.vcf" compare="sim_size" delta="50" />
```
nsoranzo, line 138:
```suggestion
            <output name="out" ftype="vcf" file="out2.vcf" lines_diff="4" />
```

tools/ucsc_tools/fatovcf/fatovcf.xml
```diff
@@ -0,0 +1,150 @@
...
+        #if $startOffset
+            -startOffset=$startOffset
+        #end if
+        #if $ambiguous == "-ambiguousToN"
+            -ambiguousToN
+        #elif $ambiguous == "-resolveAmbiguous"
+            -resolveAmbiguous
+        #end if
+        #if $excl_seq.excludeFile
+            -excludeFile='excludeFile.txt'
```
nsoranzo, line 40:
```suggestion
            -excludeFile='$excl_seq.excludeFile'
```

tools/ucsc_tools/fatovcf/fatovcf.xml
```diff
@@ -0,0 +1,150 @@
...
+        #end if
+    ]]></command>
+    <inputs>
+        <param name="in_fasta" format="fasta" type="data" label="FASTA Alignment" help="Must contain a series of sequences with different names and the same length. Both N and - are treated as missing information." />
+        
+        <conditional name="ref_seq">
+            <param name="ref" type="boolean" checked="true" truevalue="" falsevalue="false" label="Determine reference sequence" help="Use the first sequence in the FASTA file as the reference sequence." />
+            <when value="" />
+            <when value="false">
+                <param name="refSeqName" type="text" label="Name of sequence that should be used as reference sequence:" help="Must be present in the FASTA file. (-ref)" />
```
bgruening, line 64:
```suggestion
                <param argument="-ref" type="text" label="Name of sequence that should be used as reference sequence:" help="Must be present in the FASTA file." />
```

Maybe? using `argument` is in general recommended if possible.

tools/ucsc_tools/fatovcf/fatovcf.xml
```diff
@@ -0,0 +1,150 @@
...
+        ln -sfn '$excl_seq.excludeFile' excludeFile.txt &&
+    #end if
+    #if $output.vcfChrom
+        ln -sfn '$output.vcfChrom' vcfChromFile.vcf &&
+    #end if
+    #if $in_fasta
+        ln -sfn '$in_fasta' input.fa &&
+    #end if
+    faToVcf
+        input.fa
```
nsoranzo, line 26:
```suggestion
        '$in_fasta'
```

tools/ucsc_tools/fatovcf/fatovcf.xml
bgruening, line 138: sim_size is not very accurate, can you use asserts instead, and look at some content of the VCF file? Or maybe `contains`?

tools/ucsc_tools/fatovcf/fatovcf.xml
```diff
@@ -0,0 +1,150 @@
...
+    <command detect_errors="exit_code"><![CDATA[
+    #if $excl_seq.excludeFile
+        ln -sfn '$excl_seq.excludeFile' excludeFile.txt &&
+    #end if
+    #if $output.vcfChrom
+        ln -sfn '$output.vcfChrom' vcfChromFile.vcf &&
+    #end if
+    #if $in_fasta
+        ln -sfn '$in_fasta' input.fa &&
+    #end if
```
nsoranzo, line 24:
```suggestion
```

tools/ucsc_tools/fatovcf/fatovcf.xml
```diff
@@ -0,0 +1,150 @@
...
+        #end if
+        -windowSize=$mask_sites.windowSize
+        -minAmbigInWindow=$mask_sites.minAmbigInWindow
+        $includeNoAltN
+        -minAc=$minAc
+        -minAf=$minAf
+        $output.includeRef
+        $output.noGenotypes
+        #if $output.vcfChrom
+            -vcfChrom='vcfChromFile.vcf'
```
nsoranzo, line 54:
```suggestion
            -vcfChrom='$output.vcfChrom'
```

tools/ucsc_tools/fatovcf/fatovcf.xml
kciy: Hm, why should I remove these? Since at least the input.fasta file name is displayed in the VCF header and I remove the symlink, the ugly /tmp/ path is displayed instead. So I would rather keep it as it was?

tools/ucsc_tools/fatovcf/fatovcf.xml
kciy: same question here; why not use the short `out.vcf` file name as it appears in the ##source comment in the vcf header?

nsoranzo: This one would save a file move from the tool working directory to the Galaxy dataset storage (what `from_work_dir` does), but in the great scheme of things is probably not that important, so you can keep it as it is, if you prefer.

tools/ucsc_tools/fatovcf/fatovcf.xml
nsoranzo: I think I preferred to keep the output files since they are small and don't seem to change across re-runs?

bernt-matthias (commented):
Wondering if we can turn the ucsc tools into a suite? Seems advantageous in terms of consistency. Maybe also use the same version for all of them....?

wm75:
> Wondering if we can turn the ucsc tools into a suite? Seems advantageous in terms of consistency. Maybe also use the same version for all of them....?

@bernt-matthias not sure this would offer more advantages, then it would cause issues.

Some points to consider, imo:
1. The ucsc tools are a *very* big suite of tools with very diverse function so it seems rather unlikely to me that anyone would want to install all of them.
2. We already have wrappers for some of them in different places, e.g., https://github.com/galaxyproject/galaxy/blob/dev/tools/filters/wig_to_bigwig.xml, so restructuring things would be some effort
3. Most of these tools get updated much less frequently than the suite sees releases (compare https://github.com/ucscGenomeBrowser/kent/tree/v442_base/src/hg/utils) so we would bump wrapper versions very often when there's no underlying change to a tool (this is a similar problem as for @bgruening 's text_processing tools and I do not think we want to recreate such a situation?)
4. this kind of suite level versioning is not what bioconda does. They seem to be using (at least for fatovcf) the suite version at which the tool in question received an update.

tools/ucsc_tools/fatovcf/fatovcf.xml
nsoranzo: Yep, `lines_diff` is what I've suggested in another comment.

tools/ucsc_tools/fatovcf/fatovcf.xml
```diff
@@ -0,0 +1,158 @@
...
+                <option value="" selected="true">Use the first sequence as reference</option>
+                <option value="customRef">Use a different sequence as reference</option>
+            </param>
+            <when value="customRef">
+                <param argument="-ref" type="text" label="Name of sequence that should be used as reference sequence:" help="Must be present in the FASTA file." />
+            </when>
+            <when value="" />
+        </conditional>
+
+        <param argument="-startOffset" type="integer" min="0" value="0" optional="true" label="Start offset" help="Add N bases to each position, e.g. for trimmed alignments. Default: 0" />
```
wm75, line 65: This could be moved to the `output` section, I guess? and it doesn't have to be an *optional* param - just treat a value of 0 as not used.

tools/ucsc_tools/fatovcf/fatovcf.xml
```diff
@@ -0,0 +1,158 @@
...
+            <param argument="-windowSize" type="integer" min="0" value="0" label="Window size to mask bases" help="Mask any base for which there are at least -minAmbigWindow bases in a window of +-N bases around the base. Masking approach adapted from https://github.com/roblanf/sarscov2phylo/blob/master/scripts/mask_seq.py Use -windowSize=7 for same results. Default: 0" />
+            <param argument="-minAmbigInWindow" type="integer" min="0" value="2" label="Minimum of ambiguous characters within the window given above" help="When -windowSize is provided, mask any base for which there are at least this many N, ambiguous or gap characters within the window. Default: 2" />
+        </section>
+        
+        <param argument="-includeNoAltN" type="boolean" truevalue="-includeNoAltN" falsevalue="" label="Include base positions" help="Include base positions with no alternate alleles observed, but at least one N (missing base/no-call). Default: false" />
```
wm75, line 84: The label here could be better, e.g. "Include positions without defined ALT allele" ?

tools/ucsc_tools/fatovcf/fatovcf.xml
```diff
@@ -0,0 +1,158 @@
...
+            -ref=$ref_seq.ref
+        #end if
+        #if $startOffset
+            -startOffset=$startOffset
+        #end if
+        #if $ambiguous == "-ambiguousToN"
+            -ambiguousToN
+        #elif $ambiguous == "-resolveAmbiguous"
+            -resolveAmbiguous
+        #end if
```
wm75, line 32: Can't you just use `$ambiguous` instead of this whole block?

tools/ucsc_tools/fatovcf/fatovcf.xml
```diff
@@ -0,0 +1,158 @@
...
+        </section>
+
+        <section name="mask_sites" title="Mask sites" expanded="true">
+            <param argument="-maskSites" format="vcf" type="data" optional="true" label="Mask sites at given positions (VCF file)" help="Exclude variants in positions recommended for masking in file. Typically https://github.com/W-L/ProblematicSites_SARS-CoV2/raw/master/problematic_sites_sarsCov2.vcf" />
+            <param argument="-windowSize" type="integer" min="0" value="0" label="Window size to mask bases" help="Mask any base for which there are at least -minAmbigWindow bases in a window of +-N bases around the base. Masking approach adapted from https://github.com/roblanf/sarscov2phylo/blob/master/scripts/mask_seq.py Use -windowSize=7 for same results. Default: 0" />
+            <param argument="-minAmbigInWindow" type="integer" min="0" value="2" label="Minimum of ambiguous characters within the window given above" help="When -windowSize is provided, mask any base for which there are at least this many N, ambiguous or gap characters within the window. Default: 2" />
```
wm75, line 81: If I understand things correctly, then `min` should be `1`?

tools/ucsc_tools/fatovcf/fatovcf.xml
```diff
@@ -0,0 +1,158 @@
...
+        #end if
+        #if $excl_seq.excludeFile
+            -excludeFile='$excl_seq.excludeFile'
+        #end if
+        -maxDiff=$excl_seq.maxDiff
+        #if $mask_sites.maskSites
+            -maskSites='$mask_sites.maskSites'
+        #end if
+        -windowSize=$mask_sites.windowSize
+        -minAmbigInWindow=$mask_sites.minAmbigInWindow
```
wm75, line 41: Unless these are required options, you can make the typical command line shorter by checking if the window size is > 0 before printing those.

wm75 (commented):
Very nice, thanks @kciy !
Just a few suggestions.

tools/ucsc_tools/fatovcf/fatovcf.xml
kciy: True, it could be 1 aswell. The example in the original code was with 2

nsoranzo (approved):

tools/ucsc_tools/fatovcf/fatovcf.xml
```diff
@@ -0,0 +1,145 @@
...
+        </section>
+
+        <section name="mask_sites" title="Mask sites" expanded="true">
+            <param argument="-maskSites" format="vcf" type="data" optional="true" label="Mask sites at given positions (VCF file)" help="Exclude variants in positions recommended for masking in file. Typically https://github.com/W-L/ProblematicSites_SARS-CoV2/raw/master/problematic_sites_sarsCov2.vcf" />
+            <param argument="-windowSize" type="integer" min="0" value="0" label="Window size to mask bases" help="Mask any base for which there are at least -minAmbigWindow bases in a window of +-N bases around the base. Masking approach adapted from https://github.com/roblanf/sarscov2phylo/blob/master/scripts/mask_seq.py Use -windowSize=7 for same results. Default: 0" />
+            <param argument="-minAmbigInWindow" type="integer" min="0" value="1" label="Minimum of ambiguous characters within the window given above" help="When -windowSize is provided, mask any base for which there are at least this many N, ambiguous or gap characters within the window. Default: 1" />
```
wm75, line 79: Not quite what I meant. I think `min="1" value="2"` would be good. `min="0"` does not really have a meaning and it's generally good to stick to the command line tool defaults where possible (keep surprises for experienced command line users to a minimum).

tools/ucsc_tools/fatovcf/fatovcf.xml
```diff
@@ -0,0 +1,145 @@
...
+        #end if
+        -maxDiff=$excl_seq.maxDiff
+        #if $mask_sites.maskSites
+            -maskSites='$mask_sites.maskSites'
+        #end if
+        #if $mask_sites.windowSize > 0
+            -windowSize=$mask_sites.windowSize
+        #end if
+        #if $mask_sites.minAmbigInWindow > 1
+            -minAmbigInWindow=$mask_sites.minAmbigInWindow
```
wm75, line 37: I would only do the windowSize check. If that's > 0, it activates application of whatever minAmbiginWindow setting. (This is the behavior that the help describes.)

wm75 (commented):
Almost there, thanks :)

tools/ucsc_tools/fatovcf/fatovcf.xml
kciy: Makes sense, OK

wm75 (approved):
Perfect, thanks!

PR #5039, opened 2023-01-07
Update Virhunter tool: new tabular output format

bdartigues:
FOR CONTRIBUTOR:


bernt-matthias (commented):
Looks good. If it's always the same columns you could add the column names metadata.

bdartigues:
I don't think we can do this given that we used a fake model to pass galaxy github filesize test. The output files produced using test models have different number of column than the files produced by the real models than Bjoern dropped in Galaxy Europe.

bernt-matthias (approved):

bgruening:
Thanks!

PR #5048, opened 2023-01-11
Humann 3.0 - pinned metaphlan version

tuncK:
humann v3.0 otherwise fails after the release of metaphlan v4.0 as this still satisfies the conda dependency requirements (>=3.1) but is actually incompatible.

FOR CONTRIBUTOR:


tools/humann/macros.xml
```diff
@@ -16,6 +16,7 @@
     <xml name="requirements">
         <requirements>
             <requirement type="package" version="@TOOL_VERSION@">humann</requirement>
+            <requirement type="package" version="3.1">metaphlan</requirement>
```
bernt-matthias, line 19: Where / How is metaphan used in the humann tools? 

Latest metaphlan in IUC is 4.x, or? Gets a bit confusing, or?

bgruening: If I understand the PR description correctly `humann` depends on `methaphlan`. The conda package has a dependency defined but this is now broken because of insufficient pins.

We could and maybe should fix the conda package instead.

bgruening: In case you are interested @tuncK ... you need to restore the older version of the conda recipe and put it into a separate folder (inside the humann folder, something like /bioconda-recipes/recipes/humann/3.1/meta.yaml).

And then this old recipe needs to change the pin to metaphlan and the build number needs to be bump.

tuncK:
The bioconda PR is merged and the tool works. The other fixed typos being trivial, this PR is not needed anymore.

PR #5056, opened 2023-01-14
humann v3.0 --> v3.6

tuncK:
Updates humann to v. 3.6, but humann v. 3.0 still needs the fix in #5048 . Would it suffice if this is merged after #5048  so that both tools & containers are generated?

FOR CONTRIBUTOR:


bernt-matthias:
> that both tools & containers are generated?

I do not understand this comment

bgruening:
I think @tuncK means that we need to merge the other PR first, wait until the container is generated - so the old version is fixed and then merge that one with new updates and get a new container.

That is correct. 
@tuncK fyi, assuming just the container is broken for whatever reason we can also fix this container without touching the Galaxy tool

tuncK:
I meant, merging this will update humann to 3.6 which uses metaphlan 4. But I dont know what we need to do so that humann 3.0 + galaxy 2 is also generated. Does it do the trick if you merge that other PR first and then this?

tuncK:
@bgruening The current container of humann 3.0 is OK per se, the problem will only start if anyone attempts to re-compile the container for some reason. Because then it pulls the latest version of the dependencies, which means metaphlan 4.0.3, and it breaks.

bgruening:
@tuncK see here: https://github.com/galaxyproject/tools-iuc/pull/5048#discussion_r1070245995

The best practice would be to fix the older conda recipe. I can help with that or do it myself, just let me know.

tools/humann/macros.xml
```diff
@@ -32,7 +32,7 @@ Read more about the tool: http://huttenhower.sph.harvard.edu/humann
     </token>
     <xml name="citations">
         <citations>
-            <citation type="doi">10.1371/journal.pcbi.1002358</citation>
```
bebatut, line 35: We can maybe leave both citations here

bebatut (approved):
Thanks a lot @tuncK for this PR!!

bgruening:
Will push manually.

PR #4285, opened 2021-12-11
Add limits also to star solo

bgruening:
FOR CONTRIBUTOR:
* [ ] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [ ] - License permits unrestricted use (educational + commercial)
* [ ] - This PR adds a new tool or tool collection
* [ ] - This PR updates an existing tool or tool collection
* [ ] - This PR does something else (explain below)

Issue #1777, opened 2018-03-16
RNA STAR allows job submission when required inputs are missing, results in obscure cluster error


jennaj:
The option "Count number of reads per gene" can be set to "Yes" with the option "Gene model (gff3,gtf) file for splice junctions" left without a selected ref anno dataset OR with the option "Reference genome with or without an annotation" set an index without built-in annotation.

"Yes" for the count function is internally dependent on the inclusion of reference annotation from one of these two sources.

When this usage problem occurs, it results in a job failure that is difficult for end users to troubleshoot.

It would be better if the input dependency was trapped and the user was warned at job submission that there are missing required inputs. Tools do this in different ways - but most will highlight the missing input with help text to guide in proper usage.

If we do that, will avoid the job from even being able to be submitted before the inputs are correct and in sync.

Does this PR and others related to the upgraded tool capture this usage issue? Should/could it be added before the MTS is updated? Or possibly at the next update if too late now?
 https://github.com/galaxyproject/tools-iuc/pull/1726

ping @bgruening

Issue #1742, opened 2018-02-23
Enhancement: Add "--outWigType" && "--outWigStrand" options to RNA STAR too wrapper


jennaj:
MTS: https://toolshed.g2.bx.psu.edu/view/iuc/rgrnastar/99b17b74a8cd

> Name | Description | Version | Minimum Galaxy Version
> -- | -- | -- | --
> RNA STAR | Gapped-read mapper for RNA-seq data | 2.5.2b-1 | 17.01

**Options:** https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4631051/ then use find with "--outwig" to locate options in the context of usage examples

I don't think there is a simple way to get the same results from other tools that generate bigWig files from BAMs in quite this same way without filtering the BAM first, but I didn't test that (command-line results vs Galaxy alternative-tool outputs). If someone knows of one, please post back the method as a workaround.

PR #5060, opened 2023-01-16
Update STAR version + new options for STARsolo

lldelisle:
FOR CONTRIBUTOR:


lldelisle:
@mtekman , @wm75, @JasperO98 , @pavanvidem, @ieguinoa 
If you are interested in reviewing...
Thanks

tools/rgrnastar/rg_rnaStarSolo.xml
```diff
@@ -376,6 +405,10 @@
             <expand macro="dbKeyActions" />
         </data>
         <data format="txt" name="output_stats" label="${tool.name} on ${on_string}: Barcode/Feature Statistic Summaries"/>
+        <data name="reads_per_gene" format="tabular" label="${tool.name} on ${on_string}: combined reads per gene" from_work_dir="ReadsPerGene.out.tab">
```
mtekman, line 408: Does this output file need to be uploaded too?

lldelisle: What do you mean by uploaded? For me, this is an output I like to check with multiQC when I do single-cell.

lldelisle:
Just to be clear:
- I did not updated the bam for the STAR tests (the sizes were matching and passing tests).
- When using `outSAMattributes` `CB` or `UB`, the way the bam is sorted is changed but I did not make it appear to the user.

tools/rgrnastar/rg_rnaStarSolo.xml
mtekman: Oh it's an output directory not a file -- my mistake

mtekman (approved):
Looks good on my side

tools/rgrnastar/rg_rnaStarSolo.xml
```diff
@@ -317,6 +328,24 @@
...
+                <option value="NM">NM (edit distance of the aligned read to the reference)</option>
+                <option value="MD">MD (string for mismatching positions)</option>
+                <option value="jM">jM (intron motifs for all junctions)</option>
+                <option value="jI">jI (1-based start and end of introns for all junctions)</option>
+                <option value="CR">CR Cellular barcode sequence bases (uncorrected)</option>
+                <option value="CY">CY Phred quality of the cellular barcode sequence in the CR tag</option>
+                <option value="GX">GX Gene ID</option>
+                <option value="GN">GN Gene name</option>
+                <option value="CB">CB Cell identifier (corrected)</option>
+                <option value="UB">UB UMI (corrected)</option>
```
wm75, line 346: Could you turn these into a macro shared between the wrappers?

pavanvidem:
@lldelisle I can look at it in the evening.

I have the following features/params/UI rearrangements that I have in my TODO for STARSolo. 

- Smart-seq mapping sometimes fails with 
```
EXITING because of fatal error: buffer size for SJ output is too small
Solution: increase input parameter --limitOutSJcollapsed
```
this can be fixed with adding `--limitOutSJcollapsed` param
- ​Allow compatibility with CEL-seq. For this, we just have to make `--soloCBwhitelist` optional and adapt the cheetah
- When filtering is turned on, we should also additionally output the non-filtered data. At the moment, it only outputs the filtered matrix. This does not allow users to try other filtering options for example with DropletUtils.
- ​Organize the UI better. Some options are not compatible with all protocols. For example, I am not sure whether `--soloUMIfiltering` is used for Smart-seq.

Do you think is it reasonable to add them to this PR?

lldelisle:
@pavanvidem sure.
Do you want to write it or do you want me to write it?

lldelisle:
I just realize that the Chemistry parameter propose "Cell Ranger v2" while Cell Ranger is the name of the software, for me it should be "Chromium chemistry v2"...

pavanvidem:
> @pavanvidem sure. Do you want to write it or do you want me to write it?

I will work on it. If you wish please contribute :)

pavanvidem:
> I just realize that the Chemistry parameter propose "Cell Ranger v2" while Cell Ranger is the name of the software, for me it should be "Chromium chemistry v2"...

Good cath, it is correct.

lldelisle:
@pavanvidem , here is what I have in mind, tell me if this would work for you:
- For `--limitOutSJcollapsed`, I can just take the block limits from STAR to STARsolo
- CEL-seq, I see 2 possibilities:
  - We add a new solo_type "CEL-seq" which would use: 
  ```
  --soloType CB_UMI_Simple 
  --soloCBstart 7
  --soloCBlen 6  
  --soloUMIstart 1  
  --soloUMIlen 6  
  --soloBarcodeMate 2 
  --soloCBwhitelist None
  --clip5pNbases 0 36
  ```
  - Or we simply set `soloCBwhitelist` optional in `CB_UMI_Simple`, this is more flexible.
- For the Counts/UMI outputs, I think most of the case people don´t really want to see the counts, I propose that we add a parameter to make them appear.
- For `--soloUMIfiltering` you are right this is only compatible with `–soloUMIdedup 1MM CR`. I can make a conditional.

pavanvidem:
- True, there are already Limits in STAR. Maybe not all limits are useful, but it doesn't hurt to add them too. Then we can move Limits block to macros.
- I like the 2nd option. With the first option, we're not flexible in choosing barcode and UMI lengths.
-  Then, we can add a check box to make the unfiltered results to appear.
- Perfect

FYI, I just started working on the above options sequentially.

lldelisle:
Perfect, tell me if you need help.

bernt-matthias:
Would it be possible to include 

- https://github.com/galaxyproject/tools-iuc/pull/5101 (which I just opened :( )
- https://github.com/galaxyproject/tools-iuc/pull/4285

Maybe also check if those two old issues are still relevant:

- https://github.com/galaxyproject/tools-iuc/issues/1742
- https://github.com/galaxyproject/tools-iuc/issues/1777

lldelisle:
Good catch @bernt-matthias , thanks. Can you target your new PR to my branch?
@pavanvidem where are your with your implementation?

pavanvidem:
> Good catch @bernt-matthias , thanks. Can you target your new PR to my branch? @pavanvidem where are your with your implementation?

 last week I created a patch on your fork: https://github.com/lldelisle/tools-iuc/pull/2

bernt-matthias:
I just pushed https://github.com/galaxyproject/tools-iuc/pull/5060/commits/033cd4ac2aaacae14b6a141f974b33c6747e5cbf .. hape that is fine

tools/rgrnastar/rg_rnaStar.xml
```diff
@@ -1,4 +1,8 @@
-<tool id="rna_star" name="RNA STAR" version="@VERSION@+galaxy1" profile="20.01" license="MIT">
+<<<<<<< HEAD
```
lldelisle, line 1: @bernt-matthias we need to solve this conflict. Is there a reason to increase the profile?

tools/rgrnastar/rg_rnaStarSolo.xml
```diff
@@ -317,6 +335,16 @@
                 <when value="no_filter" />
             </conditional>
             <param argument="--soloOutFormatFeaturesGeneField3" type="text" value="Gene Expression" label="Field 3 in the Genes output." help="Input '-' to remove the 3rd column from the output." />
+            <param argument="--outSAMattributes" type="select" display="checkboxes" multiple="true" optional="true"
+            label="Read alignment tags to include in the BAM output">
+                <expand macro="common_SAM_attributes"/>                <option value="CR">CR Cellular barcode sequence bases (uncorrected)</option>
```
lldelisle, line 340: This is my mistake

tools/rgrnastar/rg_rnaStar.xml
```diff
@@ -1,4 +1,8 @@
...
+<tool id="rna_star" name="RNA STAR" version="@VERSION@+galaxy0" profile="20.01" license="MIT">
+=======
+<tool id="rna_star" name="RNA STAR" version="@VERSION@+galaxy2" profile="21.01" license="MIT">
+>>>>>>> 3ce631f61 (STAR: allow fasta.gz for reference)
```
bernt-matthias, line 5:
```suggestion
<tool id="rna_star" name="RNA STAR" version="@VERSION@+galaxy2" profile="@PROFILE@" license="MIT">
```

tools/rgrnastar/rg_rnaStarSolo.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="rna_starsolo" name="RNA STARSolo" version="@VERSION@" profile="20.01" license="MIT">
+<tool id="rna_starsolo" name="RNA STARSolo" version="@VERSION@+galaxy1" profile="21.01" license="MIT">
```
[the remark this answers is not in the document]
bernt-matthias:
```suggestion
<tool id="rna_starsolo" name="RNA STARSolo" version="@VERSION@+galaxy1" profile="@PROFILE@" license="MIT">

File: tools/rgrnastar/macros.xml [hunk 7]
```diff
@@ -5,7 +5,7 @@
     the index versions in sync, but you should manually adjust the +galaxy
     version number. -->
     <!-- STAR version to be used -->
-    <token name="@VERSION@">2.7.8a</token>
+    <token name="@VERSION@">2.7.10b</token>
```
bernt-matthias, line 8:
```suggestion
    <token name="@VERSION@">2.7.10b</token>
    <token name="@PROFILE@">21.01</token>
```

bernt-matthias (commented):
I always like to keep the profile version up to date .. without to much pressure .. usually -1 year.

tools/rgrnastar/macros.xml
```diff
@@ -161,8 +167,13 @@
         @FASTQ_GZ_OPTION@
     #end if
     ]]></token>
+    <token name="@LIMITS@" ><![CDATA[
+        --limitOutSJoneRead $algo.params.limits.limitOutSJoneRead
+        --limitOutSJcollapsed $algo.params.limits.limitOutSJcollapsed
+        --limitSjdbInsertNsj $algo.params.limits.limitSjdbInsertNsj
+    ]]></token>
```
lldelisle, line 174: @pavanvidem 
As discussed here: https://github.com/galaxyproject/tools-iuc/pull/4285/files#r767802436 this will not work like this as in one case it is in a conditional and in another one it is not...
@bernt-matthias would you know how to do?

lldelisle:
Regarding https://github.com/galaxyproject/tools-iuc/pull/4285/commits/aa42e4ac4719830f391d32b1ebe6f1acf69880fa
@wm75 have you tested: `\$((\${GALAXY_MEMORY_MB:-31} * 1000000))`?

tools/rgrnastar/macros.xml
bernt-matthias: See https://pythonhosted.org/Cheetah/users_guide/tipsAndTricks.html?highlight=getvar

Something like `$getVar("cond.param_in_cond", $param_not_in_cond)` should work. Or use the `varExists` in an `if` statement.

(Sadly, more use of these functions just adds arguments against this ... https://github.com/galaxyproject/galaxy/pull/12724 .. but this is fine.)

tools/rgrnastar/rg_rnaStarSolo.xml
```diff
@@ -269,6 +286,7 @@
                     <option value="Exact" >Exact</option>
                     <option value="NoDedup">Do not deduplicate UMIs</option>
                 </param>
+                <expand macro="limits" />
```
lldelisle, line 289: In #4285 , @bgruening proposed to set the limits for all not only for Smart-seq, would it be ok for you @pavanvidem ?

wm75:
> Regarding [aa42e4a](https://github.com/galaxyproject/tools-iuc/commit/aa42e4ac4719830f391d32b1ebe6f1acf69880fa) @wm75 have you tested: `\$((\${GALAXY_MEMORY_MB:-31} * 1000000))`?

I didn't personally. Not sure if anyone else tried (@bgruening @mtekman ?)

tools/rgrnastar/rg_rnaStarSolo.xml
pavanvidem: I used it under Smart-Seq because I saw the need for those params only with Smart-Seq. It should be fine to use those limits globally.

bernt-matthias:
At least `\$((\${GALAXY_MEMORY_MB:-31} * 1000000))` seems to be correct `bash` code. One might ask if the multiplication should be `1000^2`  or `1024^2`?

tools/rgrnastar/rg_rnaStarSolo.xml
lldelisle: Perfect, I'll move them and check how to use the getVar

tools/rgrnastar/rg_rnaStar.xml
```diff
@@ -403,7 +393,8 @@ with Cufflinks if your sequences come from an unstranded library preparation.">
             label="Would you like all alignments with the best score labeled
             primary?"/> -->
             <param name="outSAMprimaryFlag" type="hidden" value="OneBestScore" />
-            <param argument="--outSAMmapqUnique" type="integer" value="60" min="30" max="255"
+            <!-- default value in STAR is 255 while here it is 60... -->
+            <param argument="--outSAMmapqUnique" type="integer" value="60" min="0" max="255"
```
lldelisle, line 397: @wm75  apparently it is you in 2019... Should we change?

lldelisle:
I solved #1777 but in a PR to this branch as it is a big change (https://github.com/lldelisle/tools-iuc/pull/3).
If you have time to review and tell me if you want to integrate that change.

lldelisle:
> At least `\$((\${GALAXY_MEMORY_MB:-31} * 1000000))` seems to be correct `bash` code. One might ask if the multiplication should be `1000^2` or `1024^2`?

Good point. The thing is that the default value is 31000000000, so we would need:  `\$((\${GALAXY_MEMORY_MB:-29.5} * 1048576))` but I think it is also fine.

lldelisle:
Hi all,
This PR is quite big but I feel like we reach the end.
This PR:
- update STAR version to 2.7.10b and samtools to 1.16.1
- accept fasta.gz for genome file solves #5101 
- set limit of RAM to generate the index using `$GALAXY_MEMORY_MB` which was in #4285
- allow to set the limitsSJ in STARsolo solves #4285 
- add column names to gene count
- allow to set outWig options solves #1742
- allow more options for --alignEndsType
- put soloCBwhitelist optional in CB_UMI_Simple mode
- change CR2 to Cv2 and CR3 to Cv3 in `chemistry` to match Chromium version instead of Cell Ranger which is the name of the software.
- allow to get the gene counts globally on STARsolo
- more options for soloFeatures
- more options for soloUMIfiltering
- more SAM attributes in STARsolo

In progress:
- I solved #1777 in https://github.com/lldelisle/tools-iuc/pull/3 but I would prefer review before I merge it here.

In discussion:
- default value of outSAMmapqUnique is 255 in STAR and 60 in Galaxy: @wm75  should we put it back to 255?
- should we add EditDist_2 in soloCBmatchWLtype
- should we allow the user to get both the non-filtered data and filtered @pavanvidem 
- should we only unable soloUMIfiltering with soloUMIdedup 1MM CR @pavanvidem

lldelisle:
> Understatement of the month (at least) :)
I prefer to understate than overstate but let say this is a huge PR.
> I always prefer to use upstream defaults (or document that and why a different value is chosen)
Me too

lldelisle:
@pavanvidem perfect, the checkbox is on going.

lldelisle:
Ready for final review and merge...

pavanvidem (approved):
@lldelisle Fantastic! I have a final comment on cell filtering. Does it make sense to allow droplet-based filtering for Smart-Seq? Technically it is possible. But do people use this option? Maybe I'm trying to optimize too much. I'm already very happy with the current state of the wrapper and it can be merged. Thanks a lot for your efforts!!

lldelisle:
Maybe you are right, it does not make a lot of sense but then this suppose to put the conditional block `filter` inside the conditional block `sc` instead of the advanced setting. I don't mind doing it but I feel like it is ok to keep all options even if the user selected 'Smart-seq'. Tell me what you prefer.

pavanvidem:
As I said, it is all good and It does not break anything. Ready to merge from my side.

lldelisle:
@bgruening I think @pavanvidem is not a 'member' and his approval does not allow me to merge the PR. Could you approve it or you prefer a second review?

tools/rgrnastar/macros.xml
```diff
@@ -81,11 +82,16 @@
     <token name="@TEMPINDEX@"><![CDATA[
     ## Create temporary index for custom reference
     #if str($refGenomeSource.geneSource) == 'history':
+        #if $refGenomeSource.genomeFastaFiles.ext == "fasta"
+            ln -s '$refGenomeSource.genomeFastaFiles' refgenome.fa &&
+        #else
+            gunzip -c '$refGenomeSource.genomeFastaFiles' > refgenome.fa &&
```
wm75, line 88: require `gzip` then?

tools/rgrnastar/rg_rnaStar.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="rna_star" name="RNA STAR" version="@VERSION@+galaxy1" profile="20.01" license="MIT">
+<tool id="rna_star" name="RNA STAR" version="@VERSION@+galaxy2" profile="@PROFILE@" license="MIT">
```
wm75, line 1:
```suggestion
<tool id="rna_star" name="RNA STAR" version="@VERSION@+galaxy0" profile="@PROFILE@" license="MIT">
```
new tool version means resetting the wrapper version

tools/rgrnastar/rg_rnaStarSolo.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="rna_starsolo" name="RNA STARSolo" version="@VERSION@" profile="20.01" license="MIT">
+<tool id="rna_starsolo" name="RNA STARSolo" version="@VERSION@+galaxy1" profile="@PROFILE@" license="MIT">
```
wm75, line 1:
```suggestion
<tool id="rna_starsolo" name="RNA STARSolo" version="@VERSION@+galaxy0" profile="@PROFILE@" license="MIT">
```

tools/rgrnastar/macros.xml
```diff
@@ -5,7 +5,8 @@
     the index versions in sync, but you should manually adjust the +galaxy
     version number. -->
```
wm75, line 6: https://github.com/galaxyproject/tools-iuc/blob/02d1d482bda9804c69d2d03c890151bc491e5c73/data_managers/data_manager_star_index_builder/data_manager/rna_star_index_builder.xml#L1

is what this comment refers to.
This PR should, because of the linked macros file, trigger deployment of a new version of the DM, too, so you need to bump the DM version to `version="@IDX_VERSION@+galaxy1"`

wm75 (commented):
Just a bit of (important) housekeeping.
For everything else I trust you and other reviewers :)

tools/rgrnastar/macros.xml
lldelisle: good point. thank you

tools/rgrnastar/rg_rnaStar.xml
```diff
@@ -386,8 +386,7 @@ with Cufflinks if your sequences come from an unstranded library preparation.">
             label="Would you like all alignments with the best score labeled
             primary?"/> -->
             <param name="outSAMprimaryFlag" type="hidden" value="OneBestScore" />
-            <!-- default value in STAR is 255 while here it is 60... -->
-            <param argument="--outSAMmapqUnique" type="integer" value="60" min="0" max="255"
+            <param argument="--outSAMmapqUnique" type="integer" value="255" min="0" max="255"
```
wm75, line 389: for this one I had actually a good reason iirc, let me check

tools/rgrnastar/rg_rnaStar.xml
lldelisle: Yes I put it into a macro token

tools/rgrnastar/rg_rnaStarSolo.xml
lldelisle: same

tools/rgrnastar/rg_rnaStar.xml
wm75: remembering again: mapq 255 is a problematic value because according to SAM/BAM specs it means "undefined". Older mappers like tophat used this value as their max, and STAR retains that behavior for compatibility with Cufflinks, which we do not recommend anymore.
In contrast, using 255 as the max mapq causes problem with modern downstream tools like mutect2: https://sites.duke.edu/workblog/2021/08/18/star-rnaseq-gatk-mutect2/ and 60 has become an inofficial replacement for 255.

lldelisle: OK I put it into comment

lldelisle: Would you agree with:
```
            <!-- MAPQ 255 is the default in STAR (coming from tophat behaviour and compatibility for Cufflinks) but it is a problematic value
            - according to SAM/BAM specs it means "undefined".
            - Using 255 as the max mapq causes problem with modern downstream tools like mutect2: https://sites.duke.edu/workblog/2021/08/18/star-rnaseq-gatk-mutect2/ and 60 has become an inofficial replacement for 255. -->
            <param argument="--outSAMmapqUnique" type="integer" value="60" min="0" max="255"
            label="MAPQ value for unique mappers"
            help="STAR bases the mapping quality scores of alignment records in its BAM output on the number of alternative mappings for the read. If a read maps to multiple locations on the reference genome, the following MAPQ scoring scheme is
used: >=5 mappings => MAPQ=0; 3-4 mappings => MAPQ=1; 2 mappings => MAPQ=3. This setting lets you control the MAPQ used for reads mapped to a single location. Set to 255 for compatibility with Cufflink (default in STAR) but keep to 60 for modern downstream tools like mutect2." />
```

wm75: Yes, much better than my insufficient original remark :)

lldelisle:
Let see if I did not break everything...

wm75 (approved):
Thanks for the fixes, looks great to me now!

tools/rgrnastar/macros.xml
bernt-matthias: I guess bumping the `IDX_VERSION_SUFFIX` does not hurt, but I do not understand why we should do it?

lldelisle: Maybe to indicate that we use the STAR version @TOOL_VERSION@ instead of the STAR version @IDX_VERSION@ to build the index...

wm75: @bernt-matthias @lldelisle  Just to explain things: the reason for symlinking the macros file is to keep the IDX_VERSION in one place only so that when you update the tool wrapper to a STAR version that requires a newer index format, you'd automatically deploy also a DM that can create these indexes.

The "downside" is that any changes to the tool wrapper macros file will silently affect the DM. So in this case the next version of the DM will use the 2.7.10b version of star for building indexes. These *should* be identical to ones built with older versions, but it's good to bump the DM wrapper version to be able to trace things back.

lldelisle:
Youhou! Who wants to click or find something to change?

tools/rgrnastar/macros.xml
bernt-matthias: Seems legit.

Could a more expressive filter help here. For instance we could just store the star version that was used to create an entry in the datatable ... and then just filter datatable entries for a min (or max) required star version.

lldelisle: I don't think it is easy to add a new column in a table, this suppose to change the table (which happened when we add a new column for the 'genomeVersion')...

wm75: Yes, that would require a new table. I'm also not so sure whether that would improve the situation much.
min and max version checks in tool wrappers also need quite some discipline to maintain, and the max check in particular doesn't work backwards, i.e., at the time of writing a tool wrapper version the max value is typically unknown still so there's always at least one wrapper version that will display all newer index versions.
What would be comparably easy to do is to remove the symlink and have the DM use its own macro, which then needs to be maintained separately, but would maybe come with fewer surprises.

Anyway, I don't think this should hold back this PR any longer. If we want to decouple the DM from the tool wrapper, we should do it in its own PR where the decision will be more discoverable than as part of a giant PR like this one.

lldelisle: I agree.

wm75:
Merging is ok with you @bernt-matthias ?

bernt-matthias:
Sure. Thanks @lldelisle

lldelisle:
:tada: this was a HUGE PR.

lldelisle:
A great collaboration. I love you galaxy people.

lldelisle:
This worth probably a blog post...

pavanvidem:
> Let see if I did not break everything...

The most important thing is that this was a huge progress. Thanks! I'm using this tool very frequently for 10x and Smart-seq data. I also have some iCell (needs the optional barcode whitelist param), cel-seq and large Smart-seq (needs limit params) data in the queue. I can already tell in the next weeks if we broke something ;-)

bgruening:
> This worth probably a blog post...

YEAH! :100:

bernt-matthias:
Have we fixed all those issues https://github.com/galaxyproject/tools-iuc/pull/5060#issuecomment-1405100636.. If so we can close them.

bernt-matthias:
Hmm. Can't find the workflow run for the merge. Can someone double check?

bernt-matthias:
At least the updates did not arrive at the TS yet.

bernt-matthias:
Arg, the commit message contained `[no ci]` ...

bgruening:
I take care. Thanks @bernt-matthias

lldelisle:
Sorry I just found a bug see #5144

lldelisle:
In fact 2 bugs... 😔 I PR a fix in #5145

Issue #5061, opened 2023-01-17
Test dev prior to release freeze


mvdbeek:

mvdbeek:
I suspect the first chunk to fail, the weekly tests seem to be failing on this:

```
galaxy.objectstore CRITICAL 2023-01-17 18:45:59,219 [pN:main.1,p:19633,tN:LocalRunner.work_thread-0] Error copying /tmp/tmpqorr81b5/job_working_directory/000/60/working/dataset_60_files/mpa_vJan21_CHOCOPhlAnSGB_202103/mpa_vJan21_CHOCOPhlAnSGB_202103_SGB.fna to /tmp/tmpqorr81b5/files/000/dataset_60_files/mpa_vJan21_CHOCOPhlAnSGB_202103/mpa_vJan21_CHOCOPhlAnSGB_202103_SGB.fna: [Errno 28] No space left on device
galaxy.job_execution.output_collect DEBUG 2023-01-17 18:45:59,221 [pN:main.1,p:19633,tN:LocalRunner.work_thread-0] Error in collect_associated_files: [Errno 28] No space left on device
galaxy.jobs DEBUG 2023-01-17 18:45:59,222 [pN:main.1,p:19633,tN:LocalRunner.work_thread-0] (60) setting dataset 60 state to ERROR

Shut down
supervisord has terminated
There were problems with 1 test(s) - out of 1 test(s) executed. See tool_test_output.html for detailed breakdown.
data_manager_metaphlan_download (Test #1): failed
```
Interestingly after cleaning up the container tests continue and pass, but the runner crashes on shutdown.
I think I'll just disable data_manager_metaphlan_download tests for now.

gxydevbot:
Summary:

      3 error
     90 failure
   4279 success

[Find all tool test results here][1]

[1]: https://github.com/galaxyproject/tools-iuc/actions/runs/3957256233

gxydevbot:
Summary:

     86 failure
   4286 success

[Find all tool test results here][1]

[1]: https://github.com/galaxyproject/tools-iuc/actions/runs/3959234246

mvdbeek:
Nice, so I think the difference comes from a new assertion in the tool data table branch:

```
Traceback (most recent call last):
  File "/tmp/tmpmk87ahga/galaxy-dev/lib/galaxy/jobs/__init__.py", line 1909, in finish
    self.app, inp_data, out_data, param_dict, job=job, final_job_state=final_job_state
[...]

PR #5062, opened 2023-01-17
Exclude metaphlan datamanager from tests

mvdbeek:
Probably causes weekly test failures #5061 

FOR CONTRIBUTOR:

### DIFF
--- a/.tt_skip
+++ b/.tt_skip
@@ -1,3 +1,4 @@
+data_manager_metaphlan_database_downloader/
 data_managers/data_manager_rsync_g2/
 tools/coverage_report/
 tools/ebi_tools/
--- a/.tt_skip
+++ b/.tt_skip
@@ -1,4 +1,4 @@
-data_manager_metaphlan_database_downloader/
+data_managers/data_manager_metaphlan_database_downloader/
 data_managers/data_manager_rsync_g2/
 tools/coverage_report/
 tools/ebi_tools/


bwlang:
I can tell what you did... but not why... 
looks reasonable to me

mvdbeek:
.tt_skip contains paths to tools for which we skip testing. I think the metaphlan database is so big it takes down the github test runner, which means we don't get the summary for the weekly testing results. By skipping it I hope we can get that failing chunk to work again.

tuncK:
It is probably not only because of downloading 2.6GB, but also because of some indexing etc. steps that follow it. I dont know what causes a problem & what does not, but humann DM might also cause similar issues.

.tt_skip
```diff
@@ -1,3 +1,4 @@
+data_manager_metaphlan_database_downloader/
```
bgruening, line 1: Should this not start with "data_managers/"

mvdbeek, line 1:
```suggestion
data_managers/data_manager_metaphlan_database_downloader/
```

mvdbeek: Thanks!

bgruening (approved):
I'm not a big fan of that, we should probably have a separate skip list for the weekly tests.
But to get things moving let's get this in.

bgruening:
Thanks @mvdbeek

mvdbeek:
That was the problem, it's finishing properly now : https://github.com/galaxyproject/tools-iuc/actions/runs/3957256233/jobs/6777633519

PR #5065, opened 2023-01-18
Add tool chopin2 v1.0.7

cumbof:
Add new tool [chopin2](https://github.com/cumbof/chopin2) v1.0.7

FOR CONTRIBUTOR:


tools/chopin2/chopin2.xml
```diff
@@ -0,0 +1,152 @@
...
+
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="creator"/>
+    <expand macro="requirements"/>
+
+    <command detect_errors="exit_code">
+<![CDATA[
+    ln -s ${dataset};
```
bgruening, line 13: this needs to be single-quotes. Why is this needed in the first place since you are not renaming the file?

And please use && instead of `,`

tools/chopin2/chopin2.xml
```diff
@@ -0,0 +1,152 @@
...
+                
+                <param name="accuracy_uncertainty_perc" type="float" value="5.0" min="0.0" max="100.0" 
+                       label="Accuracy uncertainty percentage" 
+                       help="Consider non optimal solutions if model accuracy is greater than the best accuracy minus this percentage." />
+            </when>
+        </conditional>
+    </inputs>
+
+    <outputs>
+        <data format="tabular" name="summary" label="${tool.name} on ${on_string}: Summary" from_work_dir="summary.txt" />
```
bgruening, line 92: you might want to add column names to the tabular metadata

tools/chopin2/macros.xml
```diff
@@ -0,0 +1,27 @@
...
+        <creator>
+            <person givenName="Fabio" familyName="Cumbo" 
+                    url="https://cumbof.github.io/" 
+                    email="fabio.cumbo@gmail.com" />
+        </creator>
+    </xml>
+
+    <xml name="requirements">
+        <requirements>
+            <requirement type="package" version="@PYTHON_VERSION@">python</requirement>
```
bgruening, line 17: Python can be removed as it is part of the conda dependencies

tools/chopin2/chopin2.xml
```diff
@@ -0,0 +1,152 @@
...
+
+    --dump
+    --cleanup
+    --nproc "\${GALAXY_SLOTS:-4}"
+    --verbose
+]]>
+    </command>
+
+    <inputs>
+        <param name="dataset" type="data" format="csv"
```
bgruening, line 39: Can this tool support tabular data?

tools/chopin2/chopin2.xml
cumbof: Thanks for your reviews @bgruening! The reason why I need to make a symbolic link here is because the software generates some files (including the output) in the same folder of the input file. Thus, this is required to force `chopin2` to build its output in the work dir

tools/chopin2/chopin2.xml
cumbof: Well, the summary file contains some header lines before actually reporting a table. Here is an example:

```
# Dataset: iris.csv
# Dimensionality: 10000
# Number of levels: 100
# Max retraining iterations: 10
# Accuracy threshold (stop condition): 60.0
# Accuracy uncertainty (percentage): 5.0
# Run ID	Group Size	Retraining	Accuracy
8f0e142ff27db7f8d2cc66cfcc05e27c	4	1	78.69843527738266
```

Any suggestion about how to add the column names to the tabular metadata? Would it work in this case?

tools/chopin2/macros.xml
cumbof: Right, thanks for pointing this out

tools/chopin2/chopin2.xml
cumbof: It accepts `csv` files in input by default, but yes, tabular data are also supported by specifying `--fieldsep "\t"`

tools/chopin2/chopin2.xml
bgruening: Galaxy supports metadata for header lines, I guess you can try to set those in the same way you set tabular columns. I haven't seen this in the wild but would be cool to use it.

tools/chopin2/chopin2.xml
```diff
@@ -10,11 +10,17 @@
...
-    ln -s ${dataset};
+    ln -s '${dataset}' &&
 
     chopin2
 
     --dataset `basename $dataset`
+    #if str($datatype) == 'csv':
+        --fieldsep ','
+    #else:
+        --fieldsep $'\t'
```
bgruening, line 21: I guess you need to `\$` because `$` is cheetah.

tools/chopin2/chopin2.xml
```diff
@@ -36,9 +42,14 @@
     </command>
 
     <inputs>
-        <param name="dataset" type="data" format="csv" 
+        <param name="dataset" type="data" format="csv,tabular" 
                label="Select a dataset" 
-               help="Input dataset with features on columns and observations on rows. Last column must contain classes." />
+               help="Input dataset with features on columns and observations on rows. The first column must contain the observation IDs, while the last column must contain classes. The header line is also required." />
+        
+        <param name="datatype" type="select" label="Input dataset type">
```
bgruening, line 49: You do not need that. See above.

tools/chopin2/chopin2.xml
```diff
@@ -10,11 +10,17 @@
 
     <command detect_errors="exit_code">
 <![CDATA[
-    ln -s ${dataset};
+    ln -s '${dataset}' &&
 
     chopin2
 
     --dataset `basename $dataset`
+    #if str($datatype) == 'csv':
```
bgruening, line 18: you can use `#if $dataset.ext == 'csv':`

tools/chopin2/chopin2.xml
cumbof: Right..

bgruening:
summary.txt is missing

blankenberg (approved):

tools/chopin2/chopin2.xml
```diff
@@ -0,0 +1,164 @@
...
+                <action name="comment_lines" type="metadata" default="7" />
+            </actions>
+        </data>
+        <data format="txt" name="selection" label="${tool.name} on ${on_string}: Selection" from_work_dir="selection.txt">
+            <filter>feature_selection["enable_fs"]</filter>
+        </data>
+    </outputs>
+
+    <tests>
+        <test>
```
bgruening, line 111: Can you maybe utilize the conditional in a test and also test the tabular input?

bgruening (approved):
It would be great to add an additional test, but otherwise looks great.

tools/chopin2/chopin2.xml
cumbof: I'm not sure how to do that to be honest. Should I simply add a second test which is actually exactly like the first one except for the input dataset, or is there something I could add to the test that is already defined in the xml file?

bgruening: I would add a second test, same input but tabular and then use `<conditional name="feature_selection"><param name="feature_selection" value="true">` ... and so on.

cumbof: Oh ok, I see what you mean. Working on that

bgruening:
Cool, thanks!

bernt-matthias:
Tool tests failed when this was merged (error message was `ln: /usr/local/share/gdb/auto-load//usr/local/lib/libarrow.so.1000.1.0-gdb.py: File exists`). 


So this has not been deployed to the TS

bernt-matthias:
I will try to restart, but I'm not very optimistic.

bernt-matthias:
Seems to be some problem in container building.

mvdbeek:
> Seems to be some problem in container building.

The container is building fine, but somewhere in the dependency chain gdb autoloading had been enabled ... probably not intentionally ?
```
ln: /usr/local/share/gdb/auto-load//usr/local/lib/libarrow.so.1000.1.0-gdb.py: File exists
```

bernt-matthias:
I'm quite sure that it is a problem with mulled container building due to the changed conda package format (see also https://github.com/galaxyproject/galaxy/pull/15446)

The following is in the logs of the test job:

```
galaxy.tool_util.deps.container_resolvers.mulled DEBUG 2023-01-30 12:13:16,846 [pN:main.1,p:2446,tN:LocalRunner.work_thread-3] Image name for tool chopin2: chopin2:1.0.7

galaxy.tool_util.deps.mulled.mulled_build WARNING 2023-01-30 12:14:51,914 [pN:main.1,p:2446,tN:LocalRunner.work_thread-3] Could not load metadata.yaml for 'chopin2', version '1.0.7'
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/tarfile.py", line 1674, in bz2open
    t = cls.taropen(name, mode, fileobj, **kwargs)
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/tarfile.py", line 1623, in taropen
    return cls(name, mode, fileobj, **kwargs)
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/tarfile.py", line 1486, in __init__
    self.firstmember = self.next()
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/tarfile.py", line 2289, in next
    tarinfo = self.tarinfo.fromtarfile(self)
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/tarfile.py", line 1094, in fromtarfile
    buf = tarfile.fileobj.read(BLOCKSIZE)
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/bz2.py", line 178, in read
    return self._buffer.read(size)
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/_compression.py", line 68, in readinto
    data = self.read(len(byte_view))
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/_compression.py", line 103, in read
    data = self._decompressor.decompress(rawblock, size)
OSError: Invalid data stream

During handling of the above exception, another exception occurred:

  File "/tmp/tmpxftgwlct/galaxy-dev/lib/galaxy/tool_util/deps/mulled/mulled_build.py", line 161, in base_image_for_targets
    tarball = get_file_from_recipe_url(hit["url"])
  File "/tmp/tmpxftgwlct/galaxy-dev/lib/galaxy/tool_util/deps/mulled/util.py", line 355, in get_file_from_recipe_url
    return tarfile.open(mode="r:bz2", fileobj=BytesIO(r.content))
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/tarfile.py", line 1593, in open
    return func(name, filemode, fileobj, **kwargs)
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/tarfile.py", line 1678, in bz2open
    raise ReadError("not a bzip2 file")
tarfile.ReadError: not a bzip2 file
```

mvdbeek:
I guess we're saying the same thing ? The container builds fine, but we can't inspect the metadata to determine what the base environment should be. Note that that was also the case when the tests succeeded https://github.com/galaxyproject/tools-iuc/actions/runs/4038989913/jobs/6943401508#step:8:569.

bernt-matthias:
Ahh OK. I see. Then some default base image is used. 

And the problem in the merge job is because of the gdb problem ..

bernt-matthias:
Which has been merged a few days before https://github.com/conda-forge/arrow-cpp-feedstock/issues/935

Guess this is the culprit https://github.com/conda-forge/arrow-cpp-feedstock/blob/18fc7053eea2efafad381d5fbc4f661f165291f0/recipe/activate.sh#L23

@cumbof do you see a chance to fix this?

nsoranzo:
https://github.com/conda-forge/arrow-cpp-feedstock/issues/946 also seems relevant.

bernt-matthias:
Seems that the upstream fix worked .. the tool has been deployed now.

Would a success message from the gxydevbot be helpful? Currently we only message in case of error...

nsoranzo:
> Seems that the upstream fix worked .. the tool has been deployed now.

:tada: 

> Would a success message from the gxydevbot be helpful? Currently we only message in case of error...

I think it's fine as it is, less notifications less noise!

cumbof:
> Seems that the upstream fix worked .. the tool has been deployed now.
> 
> Would a success message from the gxydevbot be helpful? Currently we only message in case of error...

That's awesome! Thanks @bernt-matthias @mvdbeek @nsoranzo

PR #5084, opened 2023-01-22
fix mirnature tests

bernt-matthias:
stdout contains dataset IDs which are different in weekly CI tests

FOR CONTRIBUTOR:

### DIFF
--- a/tools/mirnature/macros.xml
+++ b/tools/mirnature/macros.xml
@@ -1,6 +1,6 @@
 <macros>
     <token name="@TOOL_VERSION@">1.1</token>
-    <token name="@VERSION_SUFFIX@">0</token>
+    <token name="@VERSION_SUFFIX@">1</token>
     <xml name="requirements">
         <requirements>
         <requirement type="package" version="@TOOL_VERSION@">mirnature</requirement>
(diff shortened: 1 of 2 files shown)


bgruening:
We don't need a new release for those fixes, isn't it? Can we merge and stop the CI?

bernt-matthias:
> We don't need a new release for those fixes, isn't it?

Never understood when its desirable / not. If I think about people testing tools installed on their instances ...?

>  Can we merge and stop the CI?

This is still possible (and won't be detected by the new notification system for failed/skipped deployment).

bgruening:
This fix is for the weekly tests correct, it has no user-facing changes just increases the tool count on Galaxy server including potential workflow updates and all the tool chain that we have built. 

Does "CI skip" still works? Can we maybe merge those changes with that?

bernt-matthias:
> This fix is for the weekly tests correct,

Yep, thats my motivation

>  it has no user-facing changes just increases the tool count on Galaxy server including potential workflow updates and all the tool chain that we have built.

+1

> Does "CI skip" still works? Can we maybe merge those changes with that?

What do you mean?

bgruening:
https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs

bgruening (approved):

bernt-matthias:
> If you add "[ci skip]" in the merge message, it should skip the action imho.

Seems to work.

bgruening:
Cool!

PR #5088, opened 2023-01-23
Export pick_value expression tool from Galaxy codebase

mvdbeek:
I think this will see a good deal of use in combination with conditional workflow steps, and this will allow us to make changes more easily and in a versioned way.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/pick_value/.shed.yml
+++ b/tools/pick_value/.shed.yml
@@ -1,0 +1,10 @@
+categories:
+- Text Manipulation
+description: Compose a text parameter value using text, integer and float values
+homepage_url: 
+long_description: |
+  This tool lets user pick an incoming parameter from multiple possible input sources in a workflow.
+name: pick_value
+owner: iuc
+remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/main/tools/pick_value
+type: unrestricted
--- a/tools/pick_value/pick_value.xml
+++ b/tools/pick_value/pick_value.xml
@@ -1,0 +1,373 @@
+<tool name="Pick parameter value" id="pick_value" version="0.1.0" profile="21.01" tool_type="expression">
+    <macros>
+        <xml name="booleans">
+            <repeat name="pick_from" title="Pick from">
+                <param name="value" type="boolean" optional="true" label="Value" />
+            </repeat>
+        </xml>
+        <xml name="integers">
+            <repeat name="pick_from" title="Pick from">
+                <param name="value" type="integer" label="Value" optional="true" />
+            </repeat>
+        </xml>
+        <xml name="floats">
+            <repeat name="pick_from" title="Pick from">
+                <param name="value" type="float" label="Value" optional="true" />
+            </repeat>
+        </xml>
+        <xml name="texts">
+            <repeat name="pick_from" title="Pick from">
+                <param name="value" type="text" optional="true" label="Value" />
+            </repeat>
+        </xml>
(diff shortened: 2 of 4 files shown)


jmchilton:
-0 - if it is a valuable tool we want to keep working at multiple versions and we want to be available in all Galaxies - this is a reason to keep all versions in Galaxy not move it to the tool shed IMO.

mvdbeek:
> in all Galaxies - this is a reason to keep all versions in Galaxy not move it to the tool shed IMO.

admins never pick out updates to tool_conf.xml, and we'll probably be able to install expression tools with an update 3 years from now on 21.01 -- no need to tie updates to a Galaxy update.

bernt-matthias:
+1 from me: for linting and testing this tool .. and also for the possibility of proper versioning.

Is a bump of the tool version needed? `0.1.0` is the current in Galaxy https://github.com/galaxyproject/galaxy/blob/dev/tools/expression_tools/pick_value.xml

mvdbeek:
There are no changes to the tool, I think it's good to start without modification. I do want to tighten the `==` comparisons later, but want to have the unmodified version on the shed before that.

bernt-matthias:
Thanks for the explanation. Might be a good strategy also for other tools that might follow https://github.com/galaxyproject/galaxy/issues/15253 .. as long as linting / tests are successful (if not we might need to deploy manually) ..

bernt-matthias:
For admins it would be cool to have something to install tools moved from the core: https://github.com/galaxyproject/galaxy/issues/15253#issuecomment-1368819936

jmchilton:
If we put the tool in a directory and loaded the directory in the tool conf - the admins wouldn't need to update their tool conf and we could retain the out of the box experience for this tool.

bernt-matthias:
I'm pretty sure that the IUC is willing to accept the tool, but the main question is maybe if the Galaxy committers are willing to either 

- remove the tool from the core (and move them to IUC) or 
- if there is an alternative as @jmchilton that also allows for testing + versioning

So maybe we need the corresponding PR at the Galaxy code base first and see what the Galaxy committers say?

mvdbeek:
> If we put the tool in a directory and loaded the directory in the tool conf - the admins wouldn't need to update their tool conf and we could retain the out of the box experience for this tool.

but you'd have to update Galaxy to get a new version. I think we could think of this as a "polyfill", but for tools ?

bwlang:
I would like to use this tool... even before we have completed a galaxy upgrade... maybe that's +1 for IUC??

bernt-matthias:
Let's do this?

mvdbeek:
Still in favor

bernt-matthias (approved):
The tool source seems to be in the galaxy sources https://github.com/galaxyproject/galaxy/blob/dev/tools/expression_tools/pick_value.xml but not in tool conf https://github.com/galaxyproject/galaxy/blob/fc6dafaec83b45f0ed5f2201f9b2458465d0145b/lib/galaxy/config/sample/tool_conf.xml.sample#L47

I guess we should remove it from the Galaxy code base.

How about moving `parse_values_from_file` as well?

mvdbeek:
No, needs to stay.

bernt-matthias:
> No, needs to stay.

Both, `parse_values_from_file` and `pick_value`?

mvdbeek:
Yes, we can't break existing, working setups when upgrading Galaxy, and those tools are heavily used in tests.

Issue #5089, opened 2023-01-24
Scatterplot with ggplot2 multiple user defined points broken


ElectronicBlueberry:
Attempting to set a `relative size of points` under `Advanced options` in `Scatterplot with ggplot2`, together with the option `Plot multiple groups of data on one plot` (as seen in [Galaxy 101 for everyone](https://training.galaxyproject.org/training-material/topics/introduction/tutorials/galaxy-intro-101-everyone/tutorial.html)) causes the job(s) to fail. Each option works on it's own.

Tested in galaxy 23.0 on usegalaxy.org
Version of `Scatterplot with ggplot2`: (Galaxy Version 3.4.0+galaxy0)
Tool ID: toolshed.g2.bx.psu.edu/repos/iuc/ggplot2_point/ggplot2_point/3.4.0+galaxy0

Tool Standard error:
```
Warning message:
In Sys.setlocale("LC_MESSAGES", "en_US.UTF-8") :
  OS reports request to set locale to "en_US.UTF-8" cannot be honored
Warning message:
package 'ggplot2' was built under R version 4.1.3 
Error: unexpected symbol in:
"        gg_point = geom_point(size=2.0, alpha=1.0, gg_factor, shape=as.numeric('0')
        gg_line"
```

bernt-matthias:
Fixed in https://github.com/galaxyproject/tools-iuc/pull/5090?

PR #4982, opened 2022-12-05
Small fix for shape option in ggplot2

npinter:
Just a small fix for ggplot2. When selecting "Plot multiple groups of data on one plot" at "Plotting multiple groups" in the advanced options the shape selection was ignored.

(resubmitted #4966)

PR #5090, opened 2023-01-24
Add missing parentheses to ggplot2_point

wm75:
FOR CONTRIBUTOR:


wm75:
Should fix #5089. The bug got introduced with #4982, which got cherry-picked into #4907.

I've bumped the single wrapper's version only here, which now makes the @VERSION_SUFFIX@ token somewhat misleading. Suggestions for improving this situation are welcome.

bernt-matthias:
Could we have a test covering this case?

wm75:
Yeah, would be nice to have. I suspect that in the affected branch `colour='$adv.type_conditional.points.pointcolor'` should also be used just like in the other two.

I'm absolutely no expert for this tool though so if somebody else wants to keep pushing improvements to this branch, you're very welcome.

bgruening (approved):

PR #5092, opened 2023-01-24
fix nextclade tests

bernt-matthias:
No idea how this could end up wrong in the repo, any idea @pvanheus ?

FOR CONTRIBUTOR:

### DIFF
--- a/tools/nextclade/nextclade.xml
+++ b/tools/nextclade/nextclade.xml
@@ -164,7 +164,7 @@
             <param name="organism" value="sars-cov-2"/>
             <output name="report_tsv">
                 <assert_contents>
-                    <has_n_columns n="67"/>
+                    <has_n_columns n="71"/>
                     <has_text text="20A"/>
                 </assert_contents>
             </output>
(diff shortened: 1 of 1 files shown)


wm75:
This seems very weird. Just trying to figure out what those 4 extra columns are because they should also be set at the metadata level then.

wm75:
{'partiallyAliased', 'clade_nextstrain', 'clade_legacy', 'clade_who'}
new columns 4 -7. very mysterious.

wm75:
Hmm, turns out the exact number of columns, and their names, are not determined by the version of nextclade alone, but custom columns may be added through the virus-specific dataset used (which the Galaxy wrapper, by default and in those tests, downloads fresh via the `nextclade dataset get` command.
What the failing tests revealed now is that the current version of the SARS-CoV-2 dataset that gets downloaded defines those additional columns, which weren't defined in the dataset version that was the latest when the wrapper was last updated.

related: https://github.com/nextstrain/nextclade_data/blob/master/CHANGELOG.md#2022-10-27

bernt-matthias:
When analyzing this I noted that the header of the TSV file that is removed by the tool (`head -n+2`) is also inconsistent. If this would be fixed upstream then it might be possible to set the metadata correctly with `galaxy.json`?

wm75:
What a coincidence :-)
https://mstdn.science/@nextstrain/109744600249264364

wm75:
So likely things will become simpler once we're using version 2.10.

bernt-matthias:
Good to know. 

Do we know the columns in advance, or do we need to have a dynamic select?

wm75:
What we could do is to restrict a given version of the wrapper to the columns that are available at that wrapper version's release time. That way, the set of columns is guaranteed to be stable for a specific wrapper version.
Then when either a new nextclade version gets released or new remote datasets offer very interesting new columns, we release a new wrapper version with updated column definitions.

bernt-matthias:
Question by @pvanheus from gitter: 

> Is there any way to set that metadata dynamically, instead of in the tool description XML?

My answer would be that it may be done with `galaxy.json`

pvanheus:
> Question by @pvanheus from gitter:
> 
> > Is there any way to set that metadata dynamically, instead of in the tool description XML?
> 
> My answer would be that it may be done with `galaxy.json`

Thanks, that's worth investigating. The columns are specified in a JSON file in the database... a helper script could parse these out and populate a galaxy.json. How does one use galaxy.json to set column metadata?

pvanheus:
> When analyzing this I noted that the header of the TSV file that is removed by the tool (`head -n+2`) is also inconsistent. If this would be fixed upstream then it might be possible to set the metadata correctly with `galaxy.json`?

Can you explain a bit more about this? nextclade (the CLI tool) puts a header on the file. if you want "bare" output that can be concatenated together to form a single tabular output, this header needs to be removed (thus the head).

pvanheus:
> > Can you explain a bit more about this? nextclade (the CLI tool) puts a header on the file. if you want "bare" output that can be concatenated together to form a single tabular output, this header needs to be removed (thus the head).
> 
> Sure. I had a look at the file before the Galaxy tool removes the header .. and I noted that the number of columns in the header seemed not to match the number of columns in the data. Hope I do remember correctly.

The number of columns in the metadata does not correctly match that in the header - but the nextclade-created header should match the data (there was a bug some time back but that was resolved).

PR #5094, opened 2023-01-25
Add fasta regex finder

bernt-matthias:
Tool was requested.. moved from https://github.com/bernt-matthias/mb-galaxy-tools/tree/master/tools/fasta_regex_finder 

FOR CONTRIBUTOR:
* [x] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [x] - License permits unrestricted use (educational + commercial)
* [ ] - This PR adds a new tool or tool collection
* [x] - This PR updates an existing tool or tool collection
* [ ] - This PR does something else (explain below)

PR #5096, opened 2023-01-25
rename bioinf cafe suite

bernt-matthias:
because TS does not allow dash
followup for https://github.com/galaxyproject/tools-iuc/pull/5094

FOR CONTRIBUTOR:

### DIFF
--- a/tools/bioinformatics_cafe/.shed.yml
+++ b/tools/bioinformatics_cafe/.shed.yml
@@ -1,4 +1,4 @@
-name: bioinformatics-cafe
+name: bioinformatics_cafe
 owner: mbernt
 description: Miscellanea of scripts for bioinformatics
 long_description: |
@@ -10,9 +10,9 @@ homepage_url: https://github.com/dariober/bioinformatics-cafe/
 type: unrestricted
 auto_tool_repositories:
   name_template: "{{ tool_id }}"
-  description_template: "bioinformatics-cafe {{ tool_name }}."
+  description_template: "bioinformatics_cafe {{ tool_name }}."
 suite:
-  name: "suite_bioinformatics-cafe"
+  name: "suite_bioinformatics_cafe"
   description: Miscellanea of scripts for bioinformatics
   long_description: |
     Miscellanea of scripts for bioinformatics


bgruening (approved):

nsoranzo:
I think the issue is that `suite_bioinformatics_cafe` has been created with owner `iuc` instead of `mbernt`, so we need to update this repo's `.shed.yml` to be like `tools/spades/.shed.yml` .

bernt-matthias:
Uff. If I would have known this ..

Would it be an alternative to add my user as admin to `suite_bioinformatics_cafe`? 

Or is there a way to mix manually listed repos and `auto_tool_repositories` .. ?

bgruening:
Just push as mbernt to the TS locally and give those repos admin access. This should work.

bernt-matthias:
Thanks @bgruening this worked. But I guess we will have the same problem if we add tools to the suite at some point? 

Also the suite owned by the IUC should be removed now, or?

bgruening:
> Thanks @bgruening this worked. But I guess we will have the same problem if we add tools to the suite at some point?

yes.

> Also the suite owned by the IUC should be removed now, or?

yes


sorry, those migrations are not easy.

nsoranzo:
> > Thanks @bgruening this worked. But I guess we will have the same problem if we add tools to the suite at some point?
> 
> yes.
> 
> > Also the suite owned by the IUC should be removed now, or?
> 
> yes
> 
> sorry, those migrations are not easy.

Sorry, I think these suggestions go in the wrong direction, we should keep the non-iuc owner only when adopting widely used repositories for which we don't want to break the tool lineage.
As a policy, I believe we should not depend on the original repo owner to manually add other repositories to a suite in the future. We can have a mixed suite like Spades, or I'd rather break the tool lineage when adopting it and have everything under iuc.

bgruening:
> we should keep the non-iuc owner only when adopting widely used repositories for which we don't want to break the tool lineage.

agree

> We can have a mixed suite like Spades

I would not recommend that. What we did with Spades is a hack imho. Its confusing for the consumer of a suite of tools, that tools in one repo, belong to different owners. In the end this is how we "trust" tools. Based on owners and repos.

My answer was not meant to be a general IUC recommendation, but he is an IUC member and it that is the easiest and cleanest (for the consumer of the tools) way currently imho.

I would also rather break tool linage or fix planemo to handle those cases instead of doing the Spades hack or depending on the original tool owner.

bernt-matthias:
OK. How to proceed with this case? 

Maybe open a PR switching everything to `iuc` and deprecate the suite and tool owned by me, in the TS?

bernt-matthias:
Problem with just taking iuc as owner of the tool is that this will lead to two entries in the tool menu (with the same name and description) instead of two versions of the tool. ... But this could be solved by moving the tool to a deprecated section (as I have on my instance .. not sure if there is any other deprecation mechanism that keeps the tool executable)

PR #5098, opened 2023-01-26
Update stringtie default parameter values

gallardoalba:
FOR CONTRIBUTOR:

Currently StringTie uses as default -f parameter a value of 0.15, but it implies to lose a lot of information by removing relative low abundance isoforms, even if sometimes the transcripts that are filtered out are highly expressed. The default value according the developers should be 0.01.


bgruening (approved):

bernt-matthias:
Could you check also the other parameters 

I just had a quick look in the stringtie sources and have seen that `-c` is documented to use a default of 1 and has a default of  2 in the wrapper...

gallardoalba:
> > Could you check also the other parameters
> > I just had a quick look in the stringtie sources and have seen that `-c` is documented to use a default of 1 and has a default of 2 in the wrapper...
> 
> Sure, I checked a few of them, but it seems that I missed some.



> Could you check also the other parameters
> 
> I just had a quick look in the stringtie sources and have seen that `-c` is documented to use a default of 1 and has a default of 2 in the wrapper...

Also the parameter -M was different with respect to the default parameters. I think all of them are fine now.

tools/stringtie/stringtie.xml
```diff
@@ -417,7 +417,13 @@ $adv.multi_mapping
                 <param name="input_bam_short" ftype="bam" value="short_reads.bam" />
                 <param name="input_bam_long" ftype="bam" value="long_reads.bam" />
             </conditional>            
-            <output name="output_gtf" file="stringtie_out12.gtf" ftype="gtf" compare="re_match" sort="true"/>
+            <output name="output_gtf" ftype="gtf">
+                <assert_contents>
+                    <has_text text="5.043160"/>
+                    <has_text text="StringTie version 2.2.1"/>
```
bernt-matthias, line 423: You might want to use the `@TOOL_VERSION@` token here?

bernt-matthias (approved):

tools/stringtie/stringtie.xml
gallardoalba, line 423:
```suggestion
                    <has_text text="StringTie version @TOOL_VERSION@"/>
```

PR #5122, opened 2023-02-10
Nextclade tool and data manager revision

wm75:
FOR CONTRIBUTOR:


wm75:
These first commits of the revision improve the DM, which now:
- uses nextclade 2.11 to retrieve datasets
- can download the latest version of any dataset within a specific time
  window
- can download multiple "latest" versions (using different reference sequences)
  of a dataset and keeps track of references
- uses a new datatable, nextclade2, with unique values including the
  reference ID, and a new max_nextclade_version column for future
compatibility
- has improved code documentation

They also make corresponding changes to the nextclade tool wrapper test data tables and sample tables already and add some test data for tests that I still have to write.

Next steps are to update the nextclade and nextalign tool wrappers to 2.11.0 and to address #5092 along the way.

pvanheus:
We had a little discussion on the microGalaxy meeting this morning about how to support the ever-changing columns that the nextclade tool produces and I see it is already addressed using the suggestion in #5092 about choosing a set of columns to display.

Is it perhaps possible to let advanced users select certain columns? If the DM added the list of column names to the data table, this could be used to feed the list of columns to `<options from_data_table [...]>`. Of course this means that a select list of columns could only be chosen if the user is using a DB from a data table,  but I don't see any other way to do this.

PR #5129, opened 2023-02-14
bbnorm added

tuncK:
FOR CONTRIBUTOR:


tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,297 @@
...
+    meo='$error_corr_params.meo'
+    mue='$error_corr_params.mue'
+    overlap='$error_corr_params.overlap'
+#end if
+]]></command>
+    <inputs>
+        <expand macro="input_type_cond"/>
+        <section name="hashing_params" title="Hashing parameters">
+            <param argument="k" type="integer" value="31" label="kmer length" help="Values under 32 are most efficient, but arbitrarily high values are supported."/>
+            <param argument="bits" type="integer" value="16" label="Bits per cell in bloom filter; must be 2, 4, 8, 16." help="Maximum kmer depth recorded is 2^c bits. Large values decrease accuracy for a fixed amount of memory, so use the lowest number you can that will still capture highest-depth kmers."/>
```
bernt-matthias, line 110: Can you add `min` and `max` to integer and float parameters (if applicable). 

Here you may even use a `expression` (or maybe `regex`) validator to ensure that the values are even https://docs.galaxyproject.org/en/master/dev/schema.html#generic-validators

tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,297 @@
...
+                    <element name="forward" value="bbnorm/input_R1.fastq"/>
+                    <element name="reverse" value="bbnorm/input_R2.fastq"/>
+                </collection>
+            </param>
+            <section name="norm_params">
+                <param name="target" value="4"/>
+                <param name="deterministic" value="true"/>
+                <param name="mindepth" value="0"/>
+            </section>
+            <output name="output_normalised_R1" ftype="fastq" value="bbnorm/normalised_R1.fastq"/>
```
bernt-matthias, line 230: Both tests have the same outcome .. and seem to be identical.

bernt-matthias (commented):
Few comments inline

tools/bbtools/bbnorm.xml
tuncK: Done

tools/bbtools/bbnorm.xml
tuncK: There are indeed 3 tests with identical outcomes, with 3 input scenarios. 1 might fail if the automatic  PE/SE detection fails for some reason, while 2 &3 should still pass.

I honestly did not understand why the inputs of all other BBtools needed to be explicitly managed with a "select" like this: https://github.com/galaxyproject/tools-iuc/blob/02d1d482bda9804c69d2d03c890151bc491e5c73/tools/bbtools/macros.xml#L46 so I continued that trend. Maybe I/we can revise this when next version of BBtools arrives.

wm75: Hmm, that macro seems technically correct, but can you fix the wording perhaps?
https://github.com/galaxyproject/tools-iuc/blob/02d1d482bda9804c69d2d03c890151bc491e5c73/tools/bbtools/macros.xml#L45-L63

I think the correct text for the options should be:
- Single dataset (single-end or interleaved paired-end reads)
- Paired-end data as separate datasets for forward and reverse reads
- Paired-end data as a collection of forward and reverse reads

then within the when clauses, the labels should be:
- Dataset of sequenced reads
- Forward reads dataset/Reverse reads dataset
- Paired reads collection

Specifically, this avoids talking of a "List of dataset pairs" when what's meant is a paired collection, mentions interleaved mode, makes it clearer when two separate files are expected, and doesn't talk about fastq files (when the tool is actually more permissive in terms of formats).

wm75: If you want to keep wrapper versions in sync across all tools, feel free to change this in a separate PR after this one got merged.

tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,299 @@
...
+                <element name="kmer_hist_input" ftype="tabular" file="bbnorm/kmer_hist_input.tabular"/>
+                <element name="kmer_hist_output" ftype="tabular" file="bbnorm/kmer_hist_output.tabular"/>
+            </output_collection>
+        </test>
+    </tests>
+    <help>
+**What it does**
+
+BBNorm downsamples a provided sequencing output, while paying attention to potential heteregeneities in sequencing depth obtained from the wet-lab workflow. The reads corresponding to regions with low coverage will be kept as is, whereas some of the reads contributing to an above-threshold coverage depth will be subsampled. The resulting data set is expected to be smaller in size, whereas the genome regions with low coverage levels will still be represented in the subsampled dataset. This provides a more uniform coverage depth against all genomic coordinates while the computational resources needed for subsequent steps such as assembly can be substantially reduced without losing coverage anywhere.
+
```
wm75, line 245: You may want to include the "When Not To Use BBNorm" section from https://jgi.doe.gov/data-and-tools/software-tools/bbtools/bb-tools-user-guide/bbnorm-guide/ here.

tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,299 @@
...
+#if str($input_type_cond.input_type) == 'single':
+    in='${read1_file}'
+#else:
+    in1='${read1_file}' in2='${read2_file}'
+    interleaved=f
+#end if
+
+#### Output options
+out=normalised_R1.fastq
+out2=normalised_R2.fastq
```
wm75, line 53: Does this work in the case of single-end reads? There's no test for this, but doesn't look promising.

wm75:
Also from the usage instructions:
"""
Threads:

BBNorm is fully multithreaded both when counting kmers and when doing other operations such as normalization. The counting is lock-free, using atomic counters. As a result, it will default to using all available hardware threads; this can be adjusted with the “t” flag.
"""

That default of using all available threads is not what you' want in Galaxy so better implement the t option and pass $GALAXY_SLOTS into it.

wm75:
More info is available in the general bbtools usage instructions at https://jgi.doe.gov/data-and-tools/software-tools/bbtools/bb-tools-user-guide/usage-guide/.
This addresses also *memory*, where there's a similar issue in that the tool will try to grab all available memory per default.

tools/bbtools/bbnorm.xml
tuncK: Done

tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,299 @@
...
+                    <param argument="meo" type="boolean" checked="false" label="Marks errors by reducing quality value of suspected errors; does not correct anything."/>
+                    <param argument="mue" type="boolean" checked="true" label="Mark errors only on uncorrectable reads."/>
+                    <param argument="overlap" type="boolean" checked="false" label="Correct errors by read overlap."/>
+                </when>
+            </conditional>
+        </section>
+    </inputs>
+    <outputs>
+        <data format="fastq" name="output_normalised_R1" from_work_dir="normalised_R1.fastq" label="${tool.name} on ${on_string} (normalised R1 reads)"/>
+        <data format="fastq" name="output_normalised_R2" from_work_dir="normalised_R2.fastq" label="${tool.name} on ${on_string} (normalised R2 reads)"/>
```
wm75, line 175: Only one file should be produced if the input was a single dataset. This is the only reasonable option for single-end input and, for interleaved pe reads, would preserve th input format of the data.

tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,299 @@
...
+        <data format="fastq" name="output_discarded" from_work_dir="discarded.fastq" label="${tool.name} on ${on_string} (discarded reads)"/>       
+        <collection type="list" name="output_hists" label="${tool.name} on ${on_string} (kmer histogram)">
+            <data name="kmer_hist_input" format="tabular" from_work_dir="kmer_hist_input.tabular"/>
+            <data name="kmer_hist_output" format="tabular" from_work_dir="kmer_hist_output.tabular"/>
+       </collection>
```
wm75, line 180: I think these should be *optional* outputs. Especially the discarded reads would consume a lot of user quota when most people will not really be interested in them.

tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,299 @@
...
+    #set read1_file = 'forward' + $ext
+    ln -s '${read1}' '${read1_file}' &&
+    #if str($input_type_cond.input_type) == 'pair':
+        #set read2 = $input_type_cond.read2
+        #set read2_file = 'reverse' + $ext
+        ln -s '${read2}' '${read2_file}' &&
+    #end if
+#else:
+    #set read1 = $input_type_cond.reads_collection['forward']
+    #set read1_identifier = re.sub('[^\s\w\-]', '_', str($read1.element_identifier))
```
wm75, line 28: I don't think this dance is necessary. Different from when a bam file gets produced (like by bbmap) the identifiers won't get recorded in a fastq output anyway. Same below for read2.

tuncK:
> More info is available in the general bbtools usage instructions at https://jgi.doe.gov/data-and-tools/software-tools/bbtools/bb-tools-user-guide/usage-guide/. This addresses also _memory_, where there's a similar issue in that the tool will try to grab all available memory per default.

I did not understand the memory part. Is there a Galaxy variable that I should set java's max memory allowed to?

wm75:
> > More info is available in the general bbtools usage instructions at https://jgi.doe.gov/data-and-tools/software-tools/bbtools/bb-tools-user-guide/usage-guide/. This addresses also _memory_, where there's a similar issue in that the tool will try to grab all available memory per default.
> 
> I did not understand the memory part. Is there a Galaxy variable that I should set java's max memory allowed to?

`$GALAXY_MEMORY_MB` (see https://docs.galaxyproject.org/en/latest/dev/schema.html#reserved-variables).

I'm not sure whether you can pass this to the shell script somehow. Could be that you have to go via the mentioned java option instead.
Again that is something that should alos be addressed for, for example, bbmap, but I'm not sure we should first release a rather broken bbnorm before addressing things in all tools.

wm75:
something like this could be an option:
https://github.com/galaxyproject/tools-iuc/blob/02d1d482bda9804c69d2d03c890151bc491e5c73/tools/qualimap/qualimap_macros.xml#L27-L29

if the shell script doesn't overwrite it.

tools/bbtools/bbnorm.xml
tuncK: No, so fixed.

tools/bbtools/bbnorm.xml
tuncK: Done

tools/bbtools/bbnorm.xml
tuncK: Done

tools/bbtools/bbnorm.xml
tuncK: Done. All these problems stem from trying to stick to input block from the macro.xml. I decoupled and wrote a simplified custom version.

tools/bbtools/bbnorm.xml
tuncK: Thanks a lot for your feedback. I re-wrote this input section, as if tools are not that similar anyways. However, I have 4 cases rather than 3 in case auto-interleaving function does not work for some header format for some reason, I wonder what you will think:
*single end
*single end interleaved
*PE 2 files
*PE collection.

tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,359 @@
...
+            <param argument="minq" type="integer" value="6" min="0" label="Ignore kmers containing bases with quality below this threshold"/>
+            <param argument="minprob" type="float" value="0.5" min="0" max="1" label="Ignore kmers with overall probability of correctness below this threshold"/>
+            <param argument="rdk" type="boolean" checked="true" label="Remove duplicate kmers" help="When true, a kmer's count will only be incremented once per read pair, even if that kmer occurs more than once."/>
+        </section>
+        
+        <section name="norm_params" title="Normalization parameters">
+            <param argument="fixspikes" type="boolean" checked="false" label="Do a slower, high-precision bloom filter lookup of kmers that appear to have an abnormally high depth due to collisions."/>
+            <param argument="target" type="integer" value="100" min="1" label="Target normalization depth." help="All depth parameters control kmer depth, not read depth. For kmer depth Dk, read depth Dr, read length R, and kmer size K: Dr=Dk*(R/(R-K+1))"/>
```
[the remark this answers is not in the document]
wm75: I think it would also be good to present parameters (and sections) in order of importance.
Definetily, you'd want every user to think about the depth they want to normalize to, but very few would want to be bothered with fixspikes for example.

Maybe target should even be moved out of this section and follow directly after the input choice.

tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,359 @@
...
+        #set read2 = $input_type_cond.read2
+    #end if
+#end if
+
+#if $read1.ext.endswith('.gz'):
+    #set read1_file = 'forward.fastq.gz'
+#else
+    #set read1_file = 'forward.fastq'
+#end if
+ln -s '${read1}' '${read1_file}' &&
```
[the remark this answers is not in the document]
tuncK: Thanks for the explanation that cheetah code behaves similar to a shell script. I am thinking of keeping it as is though, because consistency reduces potential future errors that one might make while revising etc. I think.

tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,359 @@
...
+#if $str($input_type_cond.input_type) in ['PE_2files', 'paired_list']:
+    #if $read2.ext.endswith('.gz'):
+        #set read2_file = 'reverse.fastq.gz'
+    #else
+        #set read2_file = 'reverse.fastq'
+    #end if
+    ln -s '${read2}' '${read2_file}' &&
+#end if
+
+bbnorm.sh t="\${GALAXY_SLOTS}" -Xmx"\${GALAXY_MEMORY_MB:-1024}m"
```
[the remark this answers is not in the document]
tuncK: Yes, it does work in the current form. It would stop if one decreases the RAM to a few kBs, for instance.

Done (without explicitly setting, java was defaulting to 1 thread now 2).

tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,359 @@
...
+        interleaved=f
+    #end if
+#end if
+
+#### Output options
+out=normalised_R1.fastq
+#if $str($input_type_cond.input_type) in ['PE_2files', 'paired_list']:
+    out2=normalised_R2.fastq
+#end if
+outt=discarded.fastq
```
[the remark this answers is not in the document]
tuncK: Done.

Out of curiosity, though: do such files that are not imported to the history get immediately deleted after the job ends?

tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,359 @@
...
+touppercase=t
+hist=kmer_hist_input.tabular
+histout=kmer_hist_output.tabular
```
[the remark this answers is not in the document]
tuncK: Done

tools/bbtools/bbnorm.xml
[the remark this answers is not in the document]
bernt-matthias: set a default for the number of slots

```
t="\${GALAXY_SLOTS:-1}"
```

I don't think that setting `-Xmx` in the tool wrapper is a good idea. In particular it seems a bad idea to set it equal to `GALAXY_MEMORY_MB` (.. one always should leave some space for -Xms ..). In my opinion the needs to be set by the galaxy admin.

tools/bbtools/bbnorm.xml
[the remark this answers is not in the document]
bernt-matthias: By default Galaxy removes the job working directory after the completion of the job. But it can be configured to keep it in case of error or always (which is helpful for debugging), see https://github.com/galaxyproject/galaxy/blob/0742d6e27702c60d1b8fe358ae03a267e3f252c3/lib/galaxy/config/sample/galaxy.yml.sample#L2412

tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,359 @@
...
+            </when>
+            <when value="PE_1file">
+                <param name="read1" type="data" format="fastqsanger.gz,fastqsanger" label="Interleaved fastq file"/>
+            </when>
+            <when value="PE_2files">
+                <param name="read1" type="data" format="fastqsanger.gz,fastqsanger" label="Forward reads fastq file"/>
+                <param name="read2" type="data" format="fastqsanger.gz,fastqsanger" label="Reverse reads fastq file"/>
+            </when>
+            <when value="paired_list">
+                <param name="reads_collection" type="data_collection" format="fastqsanger,fastqsanger.gz" collection_type="paired" label="Collection of fastqsanger paired read files"/>
```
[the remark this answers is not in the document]
tuncK: Done

tools/bbtools/bbnorm.xml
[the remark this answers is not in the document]
tuncK: It is my understanding that given that the final assembly sequence is unknown the algorithm cannot/does not keep track of reads, but rather kmers (by building a DAG?).

bbnorm.sh -h: "NOTE:  All depth parameters control kmer depth, not read depth. For kmer depth Dk, read depth Dr, read length R, and kmer size K:  Dr=Dk*(R/(R-K+1))"

tools/bbtools/bbnorm.xml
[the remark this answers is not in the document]
tuncK: @bernt-matthias are you saying that one should set the Xmx to some other galaxy variable (if nothing else, 80% of  `GALAXY_MEMORY_MB`)? Or do you disagree with @wm75's comment regarding setting a memory limit altogether and let the admin export JAVA_OPTS=...?

tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,359 @@
...
+out=normalised_R1.fastq
+#if $str($input_type_cond.input_type) in ['PE_2files', 'paired_list']:
+    out2=normalised_R2.fastq
+#end if
+outt=discarded.fastq
+touppercase=t
+hist=kmer_hist_input.tabular
+histout=kmer_hist_output.tabular
+
+#### Hashing parameters
```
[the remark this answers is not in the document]
tuncK: Done

tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,359 @@
...
+            </when>
+            <when value="paired_list">
+                <param name="reads_collection" type="data_collection" format="fastqsanger,fastqsanger.gz" collection_type="paired" label="Collection of fastqsanger paired read files"/>
+            </when>
+        </conditional>
+
+        <section name="hashing_params" title="Hashing parameters">
+            <param argument="k" type="integer" value="31" min="1" label="kmer length" help="Values under 32 are most efficient, but arbitrarily high values are supported."/>
+            <param argument="bits" type="integer" value="16" label="Bits per cell in bloom filter; must be 2, 4, 8, 16, ..." help="Maximum kmer depth recorded is 2^c bits. Large values decrease accuracy for a fixed amount of memory, so use the lowest number you can that will still capture highest-depth kmers.">
+                <validator type="expression" message="The bits per cell value must be a power of 2.">value>0 and (value &amp; (value-1) == 0)</validator>
```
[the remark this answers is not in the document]
tuncK: This quote does not seem to be true, I tried and was able to go at least up to 256 bits. I guess with realistic fastq files, the cap would come from the memory available for the hash array?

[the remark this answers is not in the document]
tuncK: Hmmm, it actually seems to be overriding down to 16 bits...

k:                	31
deterministic:    	true
toss error reads: 	false
passes:           	1
bits per cell:    	16
cells:            	356.35M
hashes:           	3
base min quality: 	6
kmer min prob:    	0.5

tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,359 @@
...
+hdp='$error_det_params.hdp'
+ldp='$error_det_params.ldp'
+tossbadreads='$error_det_params.tossbadreads'
+requirebothbad='$error_det_params.requirebothbad'
+errordetectratio='$error_det_params.errordetectratio'
+highthresh='$error_det_params.highthresh'
+lowthresh='$error_det_params.lowthresh'
+
+### Error correction parameters
+#if str($error_corr_params.ecc) == "true":
```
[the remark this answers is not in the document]
tuncK: Done

tools/bbtools/bbnorm.xml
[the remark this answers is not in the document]
tuncK: Done. I switched to a select as you suggested. It turned out that if 2 passes is selected, it is auto-reduced to 16bits. If you disable multiple-pass, you can select as many bits as you want but then the program crashes.

tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,359 @@
...
+                <validator type="expression" message="The bits per cell value must be a power of 2.">value>0 and (value &amp; (value-1) == 0)</validator>
+            </param>
+            <param argument="hashes" type="integer" value="3" min="1" label="Number of times each kmer is hashed and stored." help="Higher is slower. Higher is more accurate if there is enough memory, but less accurate if there is not enough memory."/>
+            <conditional name="prefilter">
+                <param argument="prefilter" type="select" label="Use a prefilter to eliminate low-depth kmers" help="True is slower, but generally more accurate; filters out low-depth kmers from the main hashtable. The prefilter is more memory-efficient because it uses 2-bit cells.">
+                    <option value="true" >True</option>
+                    <option value="false" selected="true">False</option>
```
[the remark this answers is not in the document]
tuncK: Done

tools/bbtools/bbnorm.xml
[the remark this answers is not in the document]
bernt-matthias: I would let the admins set `JAVA_OPTS`.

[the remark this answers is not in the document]
wm75: I've asked @bgruening how we do this on .eu.

tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,366 @@
...
+        interleaved=f
+    #end if
+#end if
+
+#### Output options
+out=normalised_R1.fastq
+#if $str($input_type_cond.input_type) in ['PE_2files', 'paired_list']:
+    out2=normalised_R2.fastq
+#end if
+#if $str($output_options.save_discarded_reads):
```
wm75, line 58: `$str` of a boolean with no `truevalue`/`falsevalue` defined will, I believe without having checked it, be evaluated as the *strings* `True`/`False`  (just like in regular python), so this conditional and the next one won't work as you expect.

```suggestion
#if $output_options.save_discarded_reads:
```
, i.e., evaluating the boolean directly, is probably what you need.

tools/bbtools/bbnorm.xml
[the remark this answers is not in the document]
wm75: That's convincing so lets go with the current help. We can still experiment with the tool later and see if it behaves accordingly.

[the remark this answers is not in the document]
wm75: > I think it would also be good to present parameters (and sections) in order of importance. Definetily, you'd want every user to think about the depth they want to normalize to, but very few would want to be bothered with fixspikes for example.
> 
> Maybe target should even be moved out of this section and follow directly after the input choice.

This ones valid still though.

tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,366 @@
...
+            </when>
+            <when value="PE_1file">
+                <param name="read1" type="data" format="fastqsanger.gz,fastqsanger" label="Interleaved paired-end data"/>
+            </when>
+            <when value="PE_2files">
+                <param name="read1" type="data" format="fastqsanger.gz,fastqsanger" label="Forward reads"/>
+                <param name="read2" type="data" format="fastqsanger.gz,fastqsanger" label="Reverse reads"/>
+            </when>
+            <when value="paired_list">
+                <param name="reads_collection" type="data_collection" format="fastqsanger,fastqsanger.gz" collection_type="paired" label="Collection of forward/reverse dataset pairs"/>
```
wm75, line 134:
```suggestion
            <param name="input_type" type="select" label="Choose the category of inputs to be analyzed">
                <option value="single_end">Single-end reads</option>
                <option value="PE_1file">Paired-end reads as a single, interleaved dataset</option>
                <option value="PE_2files" selected="true">Paired-end reads as two separate datasets</option>
                <option value="paired_list">Paired-end reads as a collection</option>
            </param>
            <when value="single_end">
                <param name="read1" type="data" format="fastqsanger.gz,fastqsanger" label="Single-end data"/>
            </when>
            <when value="PE_1file">
                <param name="read1" type="data" format="fastqsanger.gz,fastqsanger" label="Interleaved paired-end data"/>
            </when>
            <when value="PE_2files">
                <param name="read1" type="data" format="fastqsanger.gz,fastqsanger" label="Forward reads"/>
                <param name="read2" type="data" format="fastqsanger.gz,fastqsanger" label="Reverse reads"/>
            </when>
            <when value="paired_list">
                <param name="reads_collection" type="data_collection" format="fastqsanger,fastqsanger.gz" collection_type="paired" label="Collection of forward and reverse reads"/>
```

Again, in Galaxy there are three types of collections:
- `list`
- `paired`
- `list:paired`
A "collection of dataset pairs" would be the `list:paired` type, but what this tool consumes is just the `paired` type, i.e., a collection the elements of which are a single forward reads and a single reverse reads dataset.
Phrasing these things inconsistently only leads to confusion.

tools/bbtools/bbnorm.xml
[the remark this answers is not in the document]
tuncK: I have already re-arranged the order of params. The depth is now immediately after input file box.

tools/bbtools/bbnorm.xml
tuncK: Yes, thanks for the catch.

tools/bbtools/bbnorm.xml
tuncK: Hmm, I was imaging "paired" to be a vector of pairs of length 1. Am I supposed to handle list:paired with a 4th category, or if I handle "paired", galaxy automatically takes care of "list:paired" that might come as part of workflows?

wm75: Aha, now we're getting to the core of why I'm so picky about this :)
If you handle `paired` and a workflow feeds in a `list:paired` what you'll get automatically is parallel job execution for every `paired` element inside the outer list, and the result of this will be a *new* `list:paired`. This is the behavior you want here (and in most cases).

If you were handling the `list:paired` case explicitly in the wrapper, then the complete thing would get passed to *one* execution of the wrapper code and you would be confined to the one compute node that's handling this run.
This would only be desired for tools that somehow need all the data, e.g., because they are trying to merge it.
So when I'm reading "List of dataset pairs" I'm immediately worried that this wrapper will do something wrong with my collection.

wm75: > Hmm, I was imaging "paired" to be a vector of pairs of length 1.

You can have, of course, a `list:paired` with just one outer element, but `paired` is more like a struct of a forward and a reverse element.

tools/bbtools/bbnorm.xml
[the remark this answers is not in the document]
wm75: Ah, I missed that, thanks!

tools/bbtools/bbnorm.xml
```diff
@@ -0,0 +1,366 @@
...
+#### Output options
+out=normalised_R1.fastq
+#if $str($input_type_cond.input_type) in ['PE_2files', 'paired']:
+    out2=normalised_R2.fastq
+#end if
+#if $output_options.save_discarded_reads:
+    outt=discarded.fastq
+#end if
+touppercase=t
+#if $str($output_options.save_kmer_hists):
```
wm75, line 62: Don't forget this second case of "str(bool)" :)

tuncK: Done. Copy-paste has this ripple effect problem ;(

wm75:
Ok , last remaining issue now is how to handle memory allocation. grepping shows that tools-iuc has a whole zoo of apparent solutions for this, and I really don't know, which one should be preferred.

picard tools use:
```
./picard/picard_macros.xml:    _JAVA_OPTIONS=\${_JAVA_OPTIONS:-'-Xmx2048m -Xms256m'} &&
./picard/picard_macros.xml:    export _JAVA_OPTIONS &&
```
which is what @bernt-matthias seems to favor - just that it uses _JAVA_OPTIONS instead of JAVA_OPTS

another case is:
```
./snpeff-covid19/snpEff_macros.xml:  <token name="@JAVA_OPTIONS@">-Xmx\${GALAXY_MEMORY_MB:-8192}m</token>
./snpeff-covid19/snpEff.xml:        snpEff @JAVA_OPTIONS@ eff
```
which is more similar to what I suggested.

There are also other examples of tools that simply hard code the -Xmx value, but that doesn't seem to be a good idea.

The major difference between using GALAXY_MEMORY_MB and _JAVA_OPTIONS from an admin point of view would be that with the java-specific variable an admin would have to know that the tool is a java one, while GALAXY_MEMORY_MB is the regular way, but I am really not an expert on this.

tuncK:
> another case is:
> 
> ```
> ./snpeff-covid19/snpEff_macros.xml:  <token name="@JAVA_OPTIONS@">-Xmx\${GALAXY_MEMORY_MB:-8192}m</token>
> ./snpeff-covid19/snpEff.xml:        snpEff @JAVA_OPTIONS@ eff
> ```
> 
> which is more similar to what I suggested.
> 
> There are also other examples of tools that simply hard code the -Xmx value, but that doesn't seem to be a good idea.

Is not the macro above essentially doing the same thing as hard coding Xmx by copy-pasting the token after the command?

BTW, none of the other bbtools seem to impose a memory constraint while they explicitly cap the threads. Maybe there is already something in the deployment scripts and it will also properly take care of this if we follow the trend?

In either case, happy to implement either of these solutions once we decide what needs to be done, they look fairly straight-forward. @bgruening ?

wm75:
Ok, learnt more about Java and bbmap weirdness than I intended:

- the bbmap documentation is a bit misleading when it talks about the shell script grabbing all available memory if not restricted via the -Xmx option. In fact, the shell script will run very sophisticated logic to autodetect how much memory it may use at the moment it runs (this logic is in the script `calcmem.sh`) and that detected value then gets passed via the -Xmx option if the user hasn't provided it.

- with java, settings passed via _JAVA_OPTIONS will, surprisingly, take precedence over java command line options like -Xmx, which means whatever the script autodetects will be overridden by what an admin might have configured.

This behavior seems okish though it comes a disadvantage:
- as an edge-case, but considerable issue, autodetection may in some situations depend on the system's available memory at run time, and the algorithm of the tool itself is to some degree memory-dependent, so there could be situations in which memory shortage at job runtime affects the result, i.e. leads to non-reproducible output. Likewise, the admin could simply have set -Xmx too low for the data at hand via _JAVA_OPTIONS, but in this case the results, though suboptimal, would at least be reproducible.

wm75:
What I would do here is:
don't touch the -Xmx option ,but rely on autodetection or admin _JAVA_OPTIONS config, but instead turn tool memory warnings into errors.
When memory is not enough to create a reasonably sized hashtable you'll start seeing:
```
Warning:  This table is very full, which may reduce accuracy.  Ideal load is under 60% used.
For better accuracy, use the 'prefilter' flag; run on a node with more memory; quality-trim or error-correct reads; or increase the values of the minprob flag to reduce spurious kmers.  In practice you should still get good normalization results even with loads over 90%, but the histogram and statistics will be off.
```
If memory is super-critical, it'll say:
```
Warning:  This table is crazy full, which may reduce accuracy.  Ideal load is under 60% used.
For better accuracy, use the 'prefilter' flag; run on a node with more memory; quality-trim or error-correct reads; or increase the values of the minprob flag to reduce spurious kmers.  In practice you should still get good normalization results even with loads over 90%, but the histogram and statistics will be off.
```

You can detect those warnings and have the job run fail when they are present as explained [here](https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-stdio-regex).

This way, you'd ensure reproducible tool runs and prevent reproducible, but poor-quality results at the same time.
Relying on the user to inspect the job's stderr and spot the problem when the history dataset is green is not a very good option anyway, I think.

wm75:
Besides, I found two more issues while testing:
1. when the input is a paired collection, the output should be of that same structure, but currently the result is two individual datasets. Preserving the structure is crucial if you want to use the tool in WFs.
2. ideally, also the output *formats* should match the input formats. So if the input is fastqsanger.gz the output should be fastqsanger.gz, etc., currently the output is always just fastq. Not sure if bbnorm determines the output format from the file suffix as it does for inputs, but if so then this might not be that hard to do.

Finally, I found that bbnorm has a `tmpdir=` option. If the command line help is correct you can use `tmpdir="\$TMPDIR"`, then if $TMPDIR is set, it will be respected, but if not, the output directory will be used for tmp files. That happens to be also the standard practice for wrappers.

tuncK:
> You can detect those warnings and have the job run fail when they are present as explained [here](https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-stdio-regex).
> 
> This way, you'd ensure reproducible tool runs and prevent reproducible, but poor-quality results at the same time. Relying on the user to inspect the job's stderr and spot the problem when the history dataset is green is not a very good option anyway, I think.

If my understanding of the algorithm is correct, the inaccuracy comes from the probability of hash value clashes and therefore the low quality is actually a gradual decline.  If I parse stderr with regexp, it will be only triggered if a threshold is reached. The problem I see with this approach is that the results executed on different nodes/deployments will potentially have different results. I am wondering what the risk of introducing a systematic bias into, say, a control/test set because they happen to consistently land on 2 different types of nodes.

How about setting the required memory ourselves, instead? I could manually run loglog.sh beforehand, retrieve the cardinality estimate, convert it into a memory requirement estimate ("3hashes*16bits/kmer/8bits/byte*1000000000kmers/0.5load=12 GB to achieve a table under 50% full.") and hard set it. If the memory required is higher than actually available, the process would just error out.

tuncK:
I wonder what you would think of this. Basically I pre-calculate a RAM amount based on user-set hash table occupancy level. If something dictates it use less memory, it errors out.

tools/bbtools/bbnorm.xml
```diff
@@ -34,7 +34,48 @@ ln -s '${read1}' '${read1_file}' &&
...
+	        JAVA_OPT_MEMORY_MB=\${z:4:-1};
+	    elif [[ \${z:0-1} == 'g' ]]; then
+	        JAVA_OPT_MEMORY_MB=\$((\${z:4:-1}*1024));
+	    fi;
+        if [ \$required_RAM_MB -gt "\$JAVA_OPT_MEMORY_MB" ]; then
+            >&2 echo Your Java configuration limits RAM usage to "\$JAVA_OPT_MEMORY_MB"MB, whereas you need at least \${required_RAM_MB}MB to accurately perform this task. Increase \"Hashing parameters\" ">" \"Percent memory filled\" or contact your Galaxy administrator.;
+            exit 42;
+        fi;
+    fi;
+fi &&
```
wm75, line 74: Woah, impressive!
Most importantly, if you're really willing to maintain this extra code, it will have to move to an extra file that you're calling from the wrapper, or all this code would end up on the command line stored along with runs of the tool.
Second, I don't think the current version would work correctly (but I only looked at it very briefly) because
- you don't use/respect $GALAXY_MEMORY_MB anyway, so no need to fail the job (this happens to simplify the logic quite a bit :) )
- wouldn't you need to query _JAVA_OPTIONS (instead of _JAVA_OPT)?

Conceptually, to help me review this: can you explain what this extra logic adds in terms of stringency over the much simpler to implement stderr parsing of the tool, which I suggested previously?
Is it *only* that you can fail the run now on > 0.5 already, or is the tool's own warning not reliable somehow?

tools/bbtools/bbnorm.xml
```diff
@@ -121,14 +162,14 @@ lowthresh=$error_det_params.lowthresh
                 <option value="paired">Paired-end reads as a collection</option>
             </param>
             <when value="single_end">
-                <param name="read1" type="data" format="fastqsanger.gz,fastqsanger" label="Single-end data"/>
+                <param name="read1" type="data" format="fastq,fastq.gz,fastqsanger.gz,fastqsanger" label="Single-end data"/>
```
wm75, line 165: This isn't what I meant:
`fastqsanger` is a subtype of `fastq`. If you're going to allow `fastq`, then that'd imply `fastqsanger` already.
but what I really meant was: whatever format the user passes in, should be the format that the tool outputs.

tools/bbtools/bbnorm.xml
tuncK: a) GALAXY_MEMORY_MB is handled in line 56-59. Job should fail if you need more memory than allowed.
b) Yes, good catch, it should have been _JAVA_OPTIONS
c) Parsing the 50% warning would only get triggered if you exceed the limit. I am unhappy with the possibility that if a galaxy instance consists of 2 very different types of nodes and if two files A & B get distributed to them  during parallel processing, they are not necessarily on equal footing. There might be some bias at the kmer frequency histograms, I dont know how/how much it exactly would impact the DAG construction. I want the memory to be deterministically controlled value for that particular instance rather than being determined by what was available at the run time.

wm75: So here's the catch: there could be cases where an admin has not configured memory limits in any way, but the job's constrained at run time by actual available memory. Catching the warning emitted by bbnorm would be able to handle the situation, while your prior lookup would fail.

> I am unhappy with the possibility that if a galaxy instance consists of 2 very different types of nodes and if two files A & B get distributed to them during parallel processing, they are not necessarily on equal footing.

Agreed, but your check won't fully protect against this happening because memory could be scarce at run time (see above).

Digging again into the stderr output of bbnorm, I'm finding lines like this one:
`Made hash table:  	hashes = 3   	 mem = 43.97 MB   	cells = 23.05M   	used = 58.766%`
emitted for every pass (but I think later passes have lower mem requirements, so checking the first such line should be sufficient).
Maybe parsing this line and just looking at the usage % is enough?

wm75: Oh, and for local testing I'm serving a modified version of the wrapper with e.g.:
`_JAVA_OPTIONS="-Xmx100M -Xms25M" bbnorm.sh t=4 -Xmx500M`
This lets me play with different settings by just changing one line in the xml without having to restart planemo. The above config, for example, will demonstrate that _JAVA_OPTIONS takes precedence over bbnorm's -Xmx option.
Hope this helps!

tools/bbtools/bbnorm.xml
bgruening: ping @tuncK

tuncK:
After some more discussion with @wm75 and @bgruening, I removed the memory "micro-management". Now if you do not pass anything, it should default to 4GB. Advantage is much simpler code, disadvantage is less numerical stability. This still stayed: if hash table occupancy reaches the critical threshold of 60%, the tool generates an warning, the wrapper catches & raises a fatal error.

Please take one more look and merge / tell me what else needs fixing.

bgruening (approved):
@wm75 ready to go?

tools/bbtools/bbnorm.xml
bernt-matthias: This one seems still open

tuncK: `<collection name="output_pair" type="paired" format_source="reads_collection" label="${tool.name} on ${on_string} (normalised reads)">`

Is this "format_source" field not what he means?

bernt-matthias: Seems good (github confused me again by showing the outdated diff) .. just to be sure, by allowing `fastq` you are actually allowing `fastqsanger`, `fastqillumina`,  `fastqcssanger`, `fastqsolexa` ... does this tool really speak all these fastq dialects? 

If so I would suggest to make the `ftype` attributes in the test in and outputs specific, i.e. not use `fastq` in the tests but the specific dialect .. in order to check if `format_source` works as expected. 

xref https://github.com/galaxyproject/galaxy/issues/15558

tuncK: Thanks for the catch. No it does not, at least, I do not see a clear indication of that in their docs. I now hence limited to fastqsanger.

PR #5132, opened 2023-02-14
Add a safe guard to integron 

bgruening:
This is needed because of https://github.com/gem-pasteur/Integron_Finder/issues/90 which results in a directory with multi-million of files.

### DIFF
--- a/tools/integron_finder/integron_finder.xml
+++ b/tools/integron_finder/integron_finder.xml
@@ -35,7 +35,9 @@
         && mv Results_Integron_Finder_* Results_Integron_Finder
     ]]></command>
     <inputs>
-        <param type="data" name="sequence" format="fasta" label="Replicon file" help="Replicon can be entire chromosome, contif, PCR fragments..." />
+        <param type="data" name="sequence" format="fasta" label="Replicon file" help="Replicon can be entire chromosome, contif, PCR fragments...">
+             <validator type="expression" message="Integron Finder has problems with large multi FASTA files. Please assembly your sequences or split them up into files smaller 10.000 sequences."><![CDATA[value is not None and value.metadata.sequences <= 10000]]></validator>
+        </param>
         <param name="local_max" argument="--local-max" type="boolean" checked="false" truevalue="--local-max" falsevalue="" label="Thorough local detection" help="This option allows a more sensitive search. I will be slower (dependant on the number of hits) if integrons are found, but will be as fast if nothing is detected and will not increase the false positive rate." />
 	      <param name="type_replicon" type="select" optional="true" label="Default replicons topology" help="Set the default topology for replicons, linear, circular (deault: no topology)">
 	          <option value="--linear">linear (--linear)</option>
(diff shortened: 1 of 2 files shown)


bgruening:
@pimarin can you maybe have a look at this one?

pimarin:
I will check yes !

pimarin:
So you check that length of input won't be so big for integron due to number tmp file need ?

bgruening:
Yes, we only allow fasta files with less than 10.000 sequences. Which will already create 30-000 tp 40-000 files in one directory.

pimarin:
Oky it's okay to me, test pass so it's okay, maybe an issue to the integronfinder to change it coulbe be better for galaxy in the futur

bgruening:
Commented on an already existing issue: https://github.com/gem-pasteur/Integron_Finder/issues/90

pimarin:
Oky it's good

bgruening:
@bernt-matthias can you please review and maybe merge? Thanks.

tools/integron_finder/integron_finder.xml
```diff
@@ -35,7 +35,9 @@
         && mv Results_Integron_Finder_* Results_Integron_Finder
     ]]></command>
     <inputs>
-        <param type="data" name="sequence" format="fasta" label="Replicon file" help="Replicon can be entire chromosome, contif, PCR fragments..." />
+        <param type="data" name="sequence" format="fasta" label="Replicon file" help="Replicon can be entire chromosome, contif, PCR fragments...">
```
bernt-matthias, line 38: What kind of problems does integron finder have with large files?

Do you have an upstream issue that we could link here?

pimarin: integron finder make 1 tmp directory per analyzed contigs from the fasta file, applied to metagenome or huge genome it make thousand of tmp directories

pimarin: for a smallest genome test (e faecalis) i have 100 directories x 10 file into them, for a contamined genome I obtained 1708 directories

bernt-matthias: Just wondering if we really need the hard limit (bu I would be fine with it).

Maybe we can add doc (to the parameters help or the tool help) how to deal with larger data.

For your reference: https://integronfinder.readthedocs.io/en/latest/user_guide/tutorial.html#for-big-data-people

bgruening: > Do you have an upstream issue that we could link here?

See the PR description ;)

Its the pure amount of small files that are created, we had 10.000.000 in one directory. We can discuss if we should limit to 10.000 or 100.000 but we need to limit it imho.

bernt-matthias (approved):

bernt-matthias:
Should we have bumped the version? I stopped deployment...

bgruening:
We did with VERSION_SUFFIX correct?

bernt-matthias:
My bad. Will restart

Issue #2013, opened 2018-07-31
featureCounts -  SAF (Simplified Annotation Format) or Bed as annotation file 


rgilsbach:
Would it be possible to use the "SAF (Simplified Annotation Format)" of featureCounts in Galaxy like on command line? 
SAF format is "GeneID Chr Start End Strand" according to the docs.
Alternatively, a standard Bed6 would be even nicer.

PR #5138, opened 2023-02-16
featurecounts update to 2.0.3

tuncK:
New featurecounts requires paired-end input to be explicitly marked by a -p flag. Auto-detection of PE without the flag makes  it error out, necessitating the changes below.

FOR CONTRIBUTOR:


tools/featurecounts/featurecounts.xml
```diff
@@ -1,16 +1,15 @@
...
+        <token name="@TOOL_VERSION@">2.0.3</token>
+        <token name="@VERSION_SUFFIX@">0</token>
     </macros>
     <xrefs>
         <xref type="bio.tools">subread</xref>
     </xrefs>
     <requirements>
         <requirement type="package" version="@TOOL_VERSION@">subread</requirement>
-        <requirement type="package" version="1.11">samtools</requirement>
-        <requirement type="package" version="8.31">coreutils</requirement>
```
mvdbeek, line 13: Why did you drop the coreutils requirement ? I think we added this for the paste command, which is present in busybox, but I don't know if that works consistently across older versions.

tuncK: To shrink container size only, it does not interfere with anything per se. I observe that paste is used in "tabdel_medium" case, which is covered by the tests. Tests pass without coreutils, so I dropped it. If there is some reason that I am not aware of,, happy to update it v. 8.31->9.1.

mvdbeek: Fair enough, I guess when we added it it failed tests without.

tools/featurecounts/featurecounts.xml
```diff
@@ -1,16 +1,15 @@
-<tool id="featurecounts" name="featureCounts" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="20.05">
+<tool id="featurecounts" name="featureCounts" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="22.01">
```
mvdbeek, line 1:
```suggestion
<tool id="featurecounts" name="featureCounts" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="20.09">
```
We've not added anything new to the profile versions since then, and we know this tool worked back on 20.05. We should probably extend https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tool-profile to say "for new tools". For existing tools I think it only makes sense to bump the profile version if there's anything to be gained for the tool.

bgruening: The tool dev usually do not know that. Should the IUC recommend one lower-bound version by default?

bernt-matthias:
Hi @tuncK we have quite a few open issues regarding featurecounts .. maybe you can have a look and close some of them :) 

https://github.com/galaxyproject/tools-iuc/issues?q=is%3Aissue+is%3Aopen+featurecounts

bernt-matthias:
But if you don't have the time this is good to go from my side...

tuncK:
> Hi @tuncK we have quite a few open issues regarding featurecounts .. maybe you can have a look and close some of them :)
> 
> https://github.com/galaxyproject/tools-iuc/issues?q=is%3Aissue+is%3Aopen+featurecounts

Addressed https://github.com/galaxyproject/tools-iuc/issues/2013

mvdbeek (approved):
I'm a big fan of small changes, if you want to take this out of WIP I'm happy to merge

tools/featurecounts/featurecounts.xml
```diff
@@ -171,19 +193,27 @@
...
                 <param name="reference_gene_sets"
                        format="gff,gtf,gff3"
                        type="data"
                        label="Gene annotation file"
                        help="The program assumes that the provided annotation file is in GTF format. Make sure that the gene annotation file corresponds to the same reference genome as used for the alignment">
                 </param>
             </when>
+            <when value="history_tabular">
+                <param name="reference_gene_sets"
+                       format="tabular"
```
mvdbeek, line 211: Maybe do just bed6 ? I don't think we want to allow all tabular datasets, that'll be not great for people doing support or people with a lot of tabular datasets in their history.

tuncK: Strictly speaking, the tool does not accept bed files, but rather has its own annotation input format (.saf = .tsv with 5 columns): https://subread.sourceforge.net/featureCounts.html

To make bed inputs possible, I used awk to convert .bed to .saf. Should not be the end of the world, but maybe we would be better off if users provided their tabular files instead of .bed?

mvdbeek: what you do in the command block is ok, but I don't think we have a reasonable way to restrict tabular files in the dataset selection, that is why I prefer bed or bed6

mvdbeek:
> user-uploaded genome which might not set the dbkey very consistently.

Like I said in that issue, that's a big decision we shouldn't make here. I am not comfortable with that change, I think it will increase the amount of bug reports coming in. If someone else thinks it's fine then go for it.

bernt-matthias:
My preference would be to roll back to 69bc517311a55ebeb5e7af17d6d512ca48e52481

- saf seems not a common format, or? if we want to support it my suggestion would be to find out if its 0/1 based and if the stop position is included and then create a Galaxy datatype. Seems not worth the effort, or?
- regarding the dbkey: @mvdbeek is right: this should be treated in Galaxy, e.g https://github.com/galaxyproject/galaxy/pull/12679

bernt-matthias:
I can formulate a reply to the two issues.

tools/featurecounts/featurecounts.xml
```diff
@@ -44,23 +49,17 @@
             $read_filtering_parameters.primary
             $read_filtering_parameters.ignore_dup
 
-            -t '$extended_parameters.gff_feature_type'
-            -g '$extended_parameters.gff_feature_attribute'
-                $extended_parameters.summarization_level
+            #if $anno.anno_select=="history":
```
bernt-matthias, line 52: This should also work for `"cached"`, or? Seems also to be gtf/gff https://github.com/galaxyproject/tools-iuc/blob/0ef45e4e6a9f2836b8c6c625db23fa7210b7c3ff/tools/featurecounts/tool-data/gene_sets.loc.sample#L1

tuncK:
Alright, how about now = revert back to https://github.com/galaxyproject/tools-iuc/commit/69bc517311a55ebeb5e7af17d6d512ca48e52481 + add back disabling attribute selectors if no GTF provided?

While playing with it, I found those hidden settings to be misleading because the built-in files are 5-column tables (saf) without any of these attributes ("This argument is only applicable when isGTFAnnotationFile is TRUE."). This means it actually ignores users prompt to count genes, transcripts, introns but always counts exons and reports gene_id's. As such, results obtained with the built-in annotation and history-provided annotation numerically differ.

tools/featurecounts/featurecounts.xml
tuncK: Arrgh, yes you are right. Should be non-builtin.

bernt-matthias:
> Alright, how about now = revert back to https://github.com/galaxyproject/tools-iuc/commit/69bc517311a55ebeb5e7af17d6d512ca48e52481 + add back disabling attribute selectors if no GTF provided?

`#if $anno.anno_select!="builtin":` would sond good to me.

tools/featurecounts/featurecounts.xml
```diff
@@ -177,6 +177,27 @@
                     </options>
                     <validator type="no_options" message="A cached annotation file is not available for the genome build associated with the selected input file"/>
                 </param>
+                <param name="gff_feature_type"
```
bernt-matthias, line 180: It would be cool to solve this with a macro instead of duplicating code, or

tuncK: Hmm, I did not think about it. Will do.

tuncK: Done

tools/featurecounts/featurecounts.xml
```diff
@@ -1,23 +1,34 @@
...
+        <edam_topic>topic_0622</edam_topic> <!-- Genomics -->
+        <edam_topic>topic_0091</edam_topic> <!-- Bioinformatics -->
+    </edam_topics>
+    <edam_operations>
+        <edam_operation>operation_0415</edam_operation> <!-- Nucleic acid feature detection -->
+        <edam_operation>operation_0226</edam_operation> <!-- Annotation -->
+    </edam_operations>
     <xrefs>
         <xref type="bio.tools">subread</xref>
     </xrefs>
```
bernt-matthias, line 23: Apparently featurecounts is now in bio.tools:

```suggestion
    <xrefs>
        <xref type="bio.tools">featurecounts</xref>
    </xrefs>
```
Having the `edam_topics`/`edam_operations` and the bio.tools xref is redundant. I would remove the former. 

In particular since the edam operations seems to be different than the edam terms suggested by bio.tools

-  topic_3168 Read summarisation. 
- operation_3793 Read summarisation

Also if you go for edam, then I would suggest to go for 1 operation and 1 topic, or. @bgruening what would happen in the tool menu if a tool is annotated  with multiple terms?

tuncK: If multiple terms are present, it seems to be shown under all categories.

bgruening: Multiple terms are ok imho. But we should prefer bio.tools ID and if we are not happy with this annotation fix it in bio.tools, this way all communities gets an improved annotation.

bernt-matthias (approved):

bernt-matthias:
Still good for you @mvdbeek ?

mvdbeek:
thanks @tuncK!

Issue #5146, opened 2023-02-23
Add mean option to bedtools coverage


pimarin:
Bedtools coverage command has an option `-mean` which can calculate the mean coverage of specific region given by the bed file.

I will open a PR with just adding this option

bernt-matthias:
https://github.com/galaxyproject/tools-iuc/pull/5147

bernt-matthias:
Thanks @pimarin

PR #5147, opened 2023-02-23
bedtool add mean option to calculate mean depth of specific bed regions

pimarin:
Just add the `-mean options` which simplify the output to extract the mean depth of specific regions


bernt-matthias (commented):
Looks good. Bump is needed

bernt-matthias:
Fixes https://github.com/galaxyproject/tools-iuc/issues/5146

bgruening (approved):

PR #5149, opened 2023-02-23
Quast: support assemby (individual or co) modes

bebatut:
FOR CONTRIBUTOR:
* [X]  I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [ ]  License permits unrestricted use (educational + commercial)
* [ ]  This PR adds a new tool or tool collection
* [ ]  This PR does something else (explain below)

---

With the current implementation of QUAST, we do not support the mapping  (when reads are given) of raw data on contigs generated by assembly run on individual samples (because of the `multiple=True` and loops on inputs)

It should work like this: 
image

But in reality, currently, we have 2 use cases with individual assembly:
- With raw reads in paired collection:
     
    image

     This case was really computational expensive as raw reads were mapped on each contig files in the collection

- With reads in forward and reverse collections:

     image

To solve this, I added a conditional for the assembly mode:
- If individual assembly, I removed all `multiple=True` and for loops on inputs, so QUAST runs several times in each element in collection
- If co-assembly, I kept the `multiple=True` and put loops on inputs, so QUAST runs once on all elements together

### DIFF
--- a/tools/quast/macros.xml
+++ b/tools/quast/macros.xml
@@ -15,7 +15,7 @@
         </xrefs>
     </xml>
     <xml name="gene_thresholds">
-        <param name="gene_thresholds" argument="--gene-thresholds" type="text" value="0,300,1500,3000" label="Comma-separated list of thresholds (in bp) for gene lengths to find with a finding tool"/>
+        <param argument="--gene-thresholds" type="text" value="0,300,1500,3000" label="Comma-separated list of thresholds (in bp) for gene lengths to find with a finding tool"/>
     </xml>
     <xml name="citations">
         <citations>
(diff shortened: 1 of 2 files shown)


gallardoalba:
I reviewed the changes, but mapping the reads independently probably has not so much sense except when evaluating metagenomes. Could this change in the functionality be depended on the metagenomic mode? Otherwise I think in most cases users will need to merge the collections of reads in advance.

gallardoalba (approved):

bebatut:
I do not understand why it is failing there

gallardoalba:
> I do not understand why it is failing there

It is related with that problem https://github.com/galaxyproject/planemo/pull/1356

PR #5162, opened 2023-02-27
Update ivar suite of tools to v1.4

wm75:
FOR CONTRIBUTOR:

- implements new -m default in ivar trim
- removes max value restriction on -m option in ivar consensus after upstream fix
- works around messed up ivar trim stdout/stderr in new version

### DIFF
--- a/tools/ivar/ivar_consensus.xml
+++ b/tools/ivar/ivar_consensus.xml
@@ -21,8 +21,6 @@
     ]]></command>
     <inputs>
         <param name="input_bam" type="data" format="bam" label="Bam file" help="Aligned reads, to trim primers and quality"/>
-        <!-- Warning: integer params in the following define a max of 255 intentionally
-        because the underlying C++ code of ivar defines them as uint8 and does not check for overflow! -->
         <param name="min_qual" argument="-q" type="integer" min="0" max="255" value="20" label="Minimum quality score threshold to count base"/>
         <param name="min_freq" argument="-t" type="float" min="0" max="1" value="0.0" label="Minimum frequency threshold">
             <help>
@@ -36,7 +34,7 @@
             </help>
         </param>
         <param name="min_indel_freq" argument="-c" type="float" min="0" max="1" value="0.8" label="Minimum indel frequency threshold" help="Same meaning as the &quot;Minimum frequency threshold&quot; above, but applied to indels."/>
-        <param name="min_depth" argument="-m" type="integer" min="1" max="255" value="10" label="Minimum depth to call consensus"/>
+        <param name="min_depth" argument="-m" type="integer" min="1" value="10" label="Minimum depth to call consensus"/>
         <param name="depth_action" type="select" label="How to represent positions with coverage less than the minimum depth threshold">
             <option value="-k">Drop from output (-k)</option>
             <option value="-n N" selected="true">Represent as N (-n N)</option>
--- a/tools/ivar/ivar_trim.xml
+++ b/tools/ivar/ivar_trim.xml
@@ -151,13 +151,11 @@
                     <has_size value="4023965" delta="1000"/>
                 </assert_contents>
             </output>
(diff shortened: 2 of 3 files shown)


bernt-matthias (approved):
Looks good to me. Was just wondering if instead of the hidden parameters some logic should be added to the command. Potential problem with the hidden parameters is that they might accidentally be overwritten in tests. 

But I'm totally fine with either way

wm75:
Thanks @bernt-matthias, that's fairly trivial. I'm still hopeful that ivar 1.4.1 will get released rather soon and fix SAM output piping. So for me it's ok to have this PR here remain open for a couple of days.
I'll either bump to the next release then, or just implement your suggestion.

wm75:
@cmaceves no rush at all, but what's your approximate timeline for a 1.4.1 release with https://github.com/andersen-lab/ivar/issues/157 fixed?

cmaceves:
> @cmaceves no rush at all, but what's your approximate timeline for a 1.4.1 release with [andersen-lab/ivar#157](https://github.com/andersen-lab/ivar/issues/157) fixed?

hi! we're looking for a push to 1.4.1 no later than next week.

bgruening:
@cmaceves any news about version 1.4.1? Thanks :)

cmaceves:
hi @bgruening! we just pushed a new release, it might be a bit until it goes through bioconda release but the master branch has the piping bug addressed and is functional for streaming!

bgruening:
We automatized the bioconda thing to a [large degree](http://m.genome.cshlp.org/content/early/2023/02/23/gr.276963.122.full.pdf) so we can probably take care of that - create packages, containers and get this PR here in.

Thanks a lot @cmaceves!

wm75:
nad I guess it's due to your changes around https://github.com/andersen-lab/ivar/blob/2e982a8383c9ef0b3ccdbe04ae5a23a89b713a28/src/trim_primer_quality.cpp#L659

where the continue now causes an endless loop.

cmaceves:
Okay, @wm75 thanks for pointing this out, I'll dig into it!

wm75:
> Looks good to me. Was just wondering if instead of the hidden parameters some logic should be added to the command. Potential problem with the hidden parameters is that they might accidentally be overwritten in tests.
> 
> But I'm totally fine with either way

@bernt-matthias I've implemented your suggestion now.
Let's get this in with v1.4 for now.

wm75:
@bgruening looks a bit as if 1.4.1 gets chosen when requirement is 1.4. Can I somehow require 1.4 exactly?

wm75:
> @bgruening looks a bit as if 1.4.1 gets chosen when requirement is 1.4. Can I somehow require 1.4 exactly?

Aha, specifying 1.4.0 seems to have done the trick.

bernt-matthias:
> Aha, specifying 1.4.0 seems to have done the trick.

This is surprising, isn't it? I always assumed that the specified version is exact...
The xml doc defines: `For requirements of type package this value defines a specific version of the tool dependency.`

@wm75 how did you find out which version is installed?

bgruening:
Not really, if we specify python=3 we also want the latest Python version.
The package was called 1.4.0, so specifying the exact version is fixing it. Or did I understand something wrong.

wm75:
The package version is just 1.4 with no minor version. I found yesterday that conda accepts also 1.4.0.
The confusing aspect is imo that 1.4 *is* an exact version id and the resolver still goes looking for newer minor releases. Maybe this particular edge-case behavior should be considered a bug?

wm75:
@bernt-matthias I was diagnosing the version only by behavior of the tests, which hang again with 1.4 as the requirement, but passed with 1.4.0.

bgruening:
Oh there is no https://anaconda.org/bioconda/ivar/files 1.4.0 ... I guess it's at least worth talking to the conda people if that is what it should be.

PR #5175, opened 2023-03-06
Updating tools/metaphlan from version 4.0.5 to 4.0.6

planemo-autoupdate:
Hello! This is an automated update of the following tool: **tools/metaphlan**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

I have updated tools/metaphlan from version 4.0.5 to 4.0.6.

**Project home page:** https://github.com/biobakery/MetaPhlAn/releases

### DIFF
--- a/tools/metaphlan/macros.xml
+++ b/tools/metaphlan/macros.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <macros>
-    <token name="@TOOL_VERSION@">4.0.5</token>
+    <token name="@TOOL_VERSION@">4.0.6</token>
     <token name="@VERSION_SUFFIX@">0</token>
     <token name="@PROFILE@">22.05</token>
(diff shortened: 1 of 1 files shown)


bernt-matthias:
Probably something for the data manager:  The GTDB taxonomic assignment for the vOct22 database is now available.

bgruening:
Do you think we need to update the DM?

bernt-matthias:
> Do you think we need to update the DM?

Seems so.

tuncK:
> Probably something for the data manager: The GTDB taxonomic assignment for the vOct22 database is now available.

It was actually released in Dec 2022 but the .tar was broken. We had already addressed it in the past DM revision, it seems to work as is, as far as I can tell.

Can you please merge this or https://github.com/galaxyproject/tools-iuc/pull/5218 ?

bernt-matthias:
https://github.com/galaxyproject/tools-iuc/pull/5218

PR #5181, opened 2023-03-08
Update CAT tools v5.0.3 -> 5.2.3

gallardoalba:
FOR CONTRIBUTOR:

Also this PR fixes a bug in cat bin.

### DIFF
--- a/tools/cat/cat_bins.xml
+++ b/tools/cat/cat_bins.xml
@@ -6,6 +6,7 @@
     <expand macro="requirements" />
     <expand macro="version_command" />
     <command detect_errors="exit_code"><![CDATA[
+    #import re
     #set $bin_dir = None
     #if len($mags) > 1:
         #set $bin_dir = 'inputs'
--- a/tools/cat/cat_bins.xml
+++ b/tools/cat/cat_bins.xml
@@ -15,7 +15,7 @@
             ln -s '$mag' '${bin_dir}/${input_identifier}.FASTA' &&
         #end for
     #else
-        #set $bin_file = $mags[0].element_identifier
+        #set $bin_file = re.sub('[^\s\w\-]', '_', str($mags[0].element_identifier))
         ln -s $mags[0] '$bin_file' &&
     #end if
     CAT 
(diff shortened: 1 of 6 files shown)


tools/cat/cat_bins.xml
```diff
@@ -11,14 +11,17 @@
         #set $bin_dir = 'inputs'
         mkdir -p $bin_dir &&
         #for mag in $mags:
-            ln -s '$mag' $bin_dir/ &&
+            ln -s '$mag' '${bin_dir}/${mag.element_identifier}.dat' &&
```
bgruening, line 14: the dat is not needed I think.

bgruening: do we need to strip here whitespaces and such?

gallardoalba: > the dat is not needed I think.

Yes the dat is required I think because the input can have different extensions (fna, fasta, FASTA, etc). Otherwise I can just add a different extension (perhaps to add just the *.fasta extension has more sense).

gallardoalba: > do we need to strip here whitespaces and such?

Yes, I think so; I'll include that.

tools/cat/cat_bins.xml
```diff
@@ -11,14 +11,18 @@
         #set $bin_dir = 'inputs'
         mkdir -p $bin_dir &&
         #for mag in $mags:
-            ln -s '$mag' $bin_dir/ &&
+            #set input_identifier = re.sub('[^\s\w\-]', '_', str($mag.element_identifier))
+            ln -s '$mag' '${bin_dir}/${input_identifier}.FASTA' &&
         #end for
+    #else
+        #set $bin_file = $mags[0].element_identifier
```
bgruening, line 18: some strip here?

gallardoalba: Ah, yes sorry.

bgruening (approved):

gallardoalba:
Strange, all tests pass when running it in local.

bgruening:
> /tmp/tmpkhw0nqri/job_working_directory/000/2/tool_script.sh: line 10: CAT: command not found

Do you test with --biocontainers?

gallardoalba:
> > /tmp/tmpkhw0nqri/job_working_directory/000/2/tool_script.sh: line 10: CAT: command not found
> 
> Do you test with --biocontainers?

I just tried by using --biocontainers, and yes, it fails also in the local instance.

bgruening:
What is going on here. Galaxy does not find the container if I understand the logs correctly.

bgruening:
Thanks @gallardoalba

PR #5182, opened 2023-03-08
yahs: added option for Omni-C enzyme-free prep

abueg:
hello! 👋🏼 

added an option to the yahs tools under the "preconfigured enzymes" to select Dovetail Omni-C, which is an enzyme-free prep, which means just running yahs without the restriction enzyme argument

FOR CONTRIBUTOR:

### DIFF
--- a/tools/yahs/yahs.xml
+++ b/tools/yahs/yahs.xml
@@ -259,7 +259,10 @@
                 <param name="bfile" value="test.bed"/>
             </conditional>
             <conditional name="enzyme_conditional">
-                <param name="enzyme_options" value="omnic"/>
+                <param name="enzyme_options" value="preconfigured"/>
+                <conditional name="enzyme_conditional">
+                    <param name="enzyme_options" value="omnic"/>
+                </conditional>
             </conditional>
             <param name="log_out" value="yes"/>
             <output name="log_file" ftype="txt">
--- a/tools/yahs/yahs.xml
+++ b/tools/yahs/yahs.xml
@@ -39,8 +39,10 @@
                     -e 'GATC'
                 #else if $function.enzyme_conditional.preconfigured_enzymes == 'arima1'
                     -e 'GATC,GANTC'
-                #else
+                #else if $function.enzyme_conditional.preconfigured_enzymes == 'arima2'
                     -e 'GATC,GANTC,CTNAG,TTAA'
+                #else if $function.enzyme_conditional.preconfigured_enzymes == 'omnic'
+
                 #end if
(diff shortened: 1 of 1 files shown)


tools/yahs/yahs.xml
```diff
@@ -248,6 +251,20 @@
             </output>
             <!-- COMMAND:   yahs test.fasta test.bed -r 50000,100000,150000,2000000,1000000 -a test.agp -o test_1 -->
         </test>
+        <!-- TEST 7: omnic prep -->
```
bernt-matthias, line 254: Shouldn't the new test set the value of `preconfigured_enzymes` to `omnic`?

abueg: you are completely right, sorry about that 😅 fixing now!

tools/yahs/yahs.xml
```diff
@@ -248,6 +251,23 @@
...
+        <test expect_num_outputs="6">
+            <conditional name="function">
+                <param name="function_select" value="yahs"/>
+                <param name="fasta" value="test.fasta"/>
+                <param name="bfile" value="test.bed"/>
+            </conditional>
+            <conditional name="enzyme_conditional">
+                <param name="enzyme_options" value="omnic"/>
+            </conditional>
```
bernt-matthias, line 263:
```suggestion
            <conditional name="enzyme_conditional">
                <param name="enzyme_options" value="preconfigured"/>
                <conditional name="enzyme_conditional">
                    <param name="enzyme_options" value="omnic"/>
                </conditional>
            </conditional>
```

Otherwise the new option is still not tested. Also needs to be changed in test 3 and 4.

bernt-matthias (commented):
isn't setting `enzyme_options="not_specified"` identical to the case that you introduce here?

tools/yahs/yahs.xml
abueg: they are identical, as they both result in no `-e` flag/argument getting passed. i wanted to have OmniC as an option among the other enzymes, though, for users who go into it with OmniC in mind, because i feel they might look for that option instead of `not specified`? though if it makes things needlessly complicated, i could edit the text and direct people to use `not specified`? 🤔

bernt-matthias: Sounds reasonable. 

@bgruening do you have someone in mind who might have an opinion on this.

bgruening: @abueg is the expert! I trust her and would merge it.

bgruening (approved):

bernt-matthias (approved):
Then lets merge. Thanks a lot.

PR #5184, opened 2023-03-10
update proteinortho v6.2.3 and some changes to the xml

pmjklemm:
update to 6.2.0

FOR CONTRIBUTOR:

### DIFF
--- a/tools/proteinortho/proteinortho.xml
+++ b/tools/proteinortho/proteinortho.xml
@@ -44,7 +44,6 @@
         proteinortho 
             --project=result
             --cpus="\${GALAXY_SLOTS:-4}"
-            --ram="\${GALAXY_MEMORY_MB:-16000}"
             #if $more_options.selfblast:
                 $more_options.selfblast
             #end if
--- a/tools/proteinortho/proteinortho.xml
+++ b/tools/proteinortho/proteinortho.xml
@@ -129,9 +129,34 @@
         </conditional>
     </inputs>
     <outputs>
-        <data name="blastgraph" format="tabular" label="${tool.name} on ${on_string}: RBH graph" from_work_dir="result.blast-graph"/>
-        <data name="proteinortho" format="tabular" label="${tool.name} on ${on_string}: orthology-groups" from_work_dir="result.proteinortho.tsv"/>
-        <data name="proteinorthograph" format="tabular" label="${tool.name} on ${on_string}: orthology-pairs" from_work_dir="result.proteinortho-graph"/>
+        <data name="blastgraph" 
+              format="tabular" 
+              label="${tool.name} on ${on_string}: RBH graph" 
+              from_work_dir="result.blast-graph">
+            <actions>
+                <action name="column_names" type="metadata"
+                    default="seqidA,seqidB,evalue_ab,bitscore_ab,evalue_ba,bitscore_ba"/>
+            </actions>
+        </data>
+        <data name="proteinortho" 
+              format="tabular" 
+              label="${tool.name} on ${on_string}: orthology-groups" 
+              from_work_dir="result.proteinortho.tsv">
+            <actions>
+                <action name="column_names" type="metadata"
+                    default="species,genes,alg.-conn.,${','.join([ f.element_identifier for f in $input_files ])}"/>
+            </actions>
+        </data>
+        <data name="proteinorthograph" 
+              format="tabular" 
+              label="${tool.name} on ${on_string}: orthology-pairs" 
+              from_work_dir="result.proteinortho-graph">
+            <actions>
+                <action name="column_names" type="metadata"
+                    default="seqidA,seqidB,evalue_ab,bitscore_ab,evalue_ba,bitscore_ba"/>
+            </actions>
+        </data>
+        </data>
     </outputs>
     <tests>
         <test expect_num_outputs="3"> <!-- test normal -->
--- a/tools/proteinortho/proteinortho.xml
+++ b/tools/proteinortho/proteinortho.xml
@@ -52,13 +52,13 @@
                 $more_options.singles
             #end if
             --p=$p
-            --e=$evalue
+            --e=$more_options.evalue
             --conn=$conn
             #if $more_options.cov:
                 --cov=$more_options.cov
             #end if
-            #if $more_options.sim:
-                --sim=`LC_NUMERIC=C awk "BEGIN {printf \"%.2f\",$more_options.sim/100}"`
+            #if $sim:
+                --sim=`LC_NUMERIC=C awk "BEGIN {printf \"%.2f\",$sim/100}"`
             #end if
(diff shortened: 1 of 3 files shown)


bernt-matthias:
The artifacts are here: https://github.com/galaxyproject/tools-iuc/actions/runs/4386446633

You can find them if you go to any of the tests and there go to the summary page.

pmjklemm:
Thank you! 
I screened through the JSON output and found `NameMapper.NotFound: cannot find 'evalue'`, I will try to investigate this.

tools/proteinortho/proteinortho.xml
```diff
@@ -100,11 +100,11 @@
...
+        <param argument="--sim" type="integer" value="95" min="0" max="100" label="Minimal reciprocal similarity in %" help="This and --evalue are main parameters for the generation of the reciprocal best hit graph. 1 = only the best reciprocal hits are reported, 0 = all possible reciprocal blast matches (within the E-value cutoff) are reported."/>
+        <param argument="--conn" type="float" value="0.1" min="0." max="1." label="Minimal algebraic connectivity" help="This is the main parameter for the clustering step. Choose larger values then more splits are done, resulting in more and smaller clusters. A value of 0 corresponds to no clustering."/>
         <section name="more_options" title="Additional Options" expanded="False">
+            <param argument="--evalue" type="float" value="0.001" min="0" label="E-value threshold of the blast algorithm" help="Larger values results in more false positives (connections between proteins)."/>
```
bernt-matthias, line 106: For consistency I would use `$more_options.evalue` in the command block (even if `$evalue` should work as well).

bernt-matthias:
> I screened through the JSON output 

The html output is easier to read (if you click on failed / errored tests below the table the failing tests are shown).

tools/proteinortho/proteinortho.xml
pmjklemm: ahh ok, so since these options are nested in the more_options sections, the env var needs this too, I will add this to the xml

pmjklemm: Do the test also need this? So e.g. `<param name="evalue" value="1"/>` needs to be `<param name="more_options.evalue" value="1"/>` ?

pmjklemm:
OK, i managed to pass all test when removing M.faa. The test only fails using `diamond makedb` (using different version) with the error  `Option is not permitted for this workflow: ignore-warnings`. This error is neither thrown on my computer (linux) nor on a osx system ... Very strange, I will try to investigate further

tools/proteinortho/proteinortho.xml
```diff
@@ -44,21 +44,20 @@
         proteinortho 
             --project=result
             --cpus="\${GALAXY_SLOTS:-4}"
-            --ram="\${GALAXY_MEMORY_MB:-16000}"
```
pmjklemm, line 47: --ram is not supported anymore in proteinortho

bernt-matthias:
Cool. Looks good to me.

For the problem with the citations I would suggest to check if they also appear if you start a local instance with `planemo serve`.

I was also wondering if it would be better to include all the requirements in the bioconda recipe? But we don't have to do it now. Advantage would be that they would be included in the container that is built by bioconda.. And we could avoid building yet another container over here.

bernt-matthias:
FYI: the linter warns (for summary) : `WARNING: Command template contains TODO text`

pmjklemm:
The requirements are optional and not mandatory (only diamond+lapack are required).
So, I do not want to include all of these alternative tools. 

I can try to `planemo serve` but I will take some time to set this up.

Thanks for mentioning the TODO in summary, I will work on resolving this!

bernt-matthias (approved):
I checked the citations with `planemo serve`. They look good. So the problem seems to be at usegalaxy.eu. @bgruening can you check if something is wrong with the citations of proteinortho at the public server?

One last optional suggestion: you could set the column names metadata for the output tables: https://github.com/galaxyproject/tools-iuc/blob/f6ccf9e6d01368c832110742b6f6534ff3b3b6a9/tools/featurecounts/featurecounts.xml#L458

Leave this up to you. Just tell me if I should merge.

pmjklemm:
`planemo serve` does not produce a localhost:9000 endpoint for me, not sure what is wrong. You just call `planemo serve` in the root directory of the tools-iuc right?

I was not aware that you can put variables in the `column_names`, therefore I just omitted these. I will add those to the metadata, very nice!

bernt-matthias:
guess you would need to add the metadata test in the output test macro: [test_outputs](https://github.com/galaxyproject/tools-iuc/blob/2e8c0484b91a3bb84167ec658d48ab101d2cc6cb/tools/proteinortho/proteinortho.xml#L6)

tools/proteinortho/proteinortho.xml
```diff
@@ -4,7 +4,6 @@
         <import>proteinortho_macros.xml</import>
         <xml name="test_outputs">
             <output name="proteinortho">
-                <metadata name="column_names" value="species,genes,alg.-conn.,${','.join([ f.element_identifier for f in $input_files ])}"/>
```
bernt-matthias: I guess you could keep the metadata tag but you would need to replace `${','.join([ f.element_identifier for f in $input_files ])}` with a specific list of columns for the respective test. If this list is not constant we can parametrize the macro. 

Let me know if you need help with this.

bernt-matthias:
Hi @pmjklemm I have a few more commits over here https://github.com/bernt-matthias/tools-iuc/tree/tools/proteinortho for some reason I can not push to your branch or open a PR against your fork (probably because you forked from the autoupdate fork).


If you like you can cherry pick them to here. 

Mostly stricter tests regarding column and row numbers (please take check test 3 .. row numbers seem to change from call to call .. so I added delta). 

Successful linting of the `comment` attribute of `has_n_columns` depends on  https://github.com/galaxyproject/galaxy/pull/15791 .. I hope that we can merge that in time ..

pmjklemm:
Yes, I tried to push to the autoupdate branch of the last PR of proteinortho and realized that this is not possible.

Your additions look very good, I did take them all! And no hurry, we can wait on #15791 and merge afterwards.

bernt-matthias (approved):
Still looks good. But we also still have the problem with the linter. 

Btw. you can always trigger CI by closing and reopening a PR.

bernt-matthias:
The linter problem should be fixed once https://github.com/galaxyproject/galaxy/pull/15933 and https://github.com/galaxyproject/galaxy/pull/15873 are merged and new python packages have been released. I'm not sure when this will happen .. people seem busy with the 23.0 release and courses. 

Tell me if its pressing. Then we can merge and deploy manually.

pmjklemm:
> Tell me if its pressing. Then we can merge and deploy manually.

It is not urgent, I can wait. From time to time I will check if it merges, so that is no problem (:

PR #5185, opened 2023-03-12
add -long option to fa2twobit

bgruening:
FOR CONTRIBUTOR:

### DIFF
--- a/data_managers/data_manager_twobit_builder/data_manager/twobit_builder.py
+++ b/data_managers/data_manager_twobit_builder/data_manager/twobit_builder.py
@@ -26,11 +26,14 @@ def get_id_name(params, dbkey, fasta_description=None):
     return sequence_id, sequence_name
 
 
-def build_twobit(data_manager_dict, fasta_filename, params, target_directory, dbkey, sequence_id, sequence_name):
+def build_twobit(data_manager_dict, fasta_filename, params, target_directory, dbkey, sequence_id, sequence_name, long):
     twobit_base_name = "%s.2bit" % (sequence_id)
     twobit_filename = os.path.join(target_directory, twobit_base_name)
 
-    args = ['faToTwoBit', fasta_filename, twobit_filename]
+    args = ['faToTwoBit']
+    if long:
+        args.append('-long')
+    args.extend([fasta_filename, twobit_filename])
     tmp_stderr = tempfile.NamedTemporaryFile(prefix="tmp-data-manager-twobit-builder-stderr")
     proc = subprocess.Popen(args=args, shell=False, cwd=target_directory, stderr=tmp_stderr.fileno())
     return_code = proc.wait()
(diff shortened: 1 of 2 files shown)


data_managers/data_manager_twobit_builder/data_manager/twobit_builder.py
```diff
@@ -71,6 +74,8 @@ def main():
     parser.add_option('-f', '--fasta_filename', dest='fasta_filename', action='store', type="string", default=None, help='fasta_filename')
     parser.add_option('-d', '--fasta_dbkey', dest='fasta_dbkey', action='store', type="string", default=None, help='fasta_dbkey')
     parser.add_option('-t', '--fasta_description', dest='fasta_description', action='store', type="string", default=None, help='fasta_description')
+    parser.add_option('-l', '--long', dest='long', action="store_true", default=False, help='For big genomes you need to pass the -long option.')
```
bernt-matthias, line 77: Could not find any docs that `-long` exists for `fato2bit`? Wondering what is "big"? Can we set it automatically based on the file size?

bgruening: The long option was given in the error message. I also did not find in the documentation :( 
So no idea. With the option, it worked.

data_managers/data_manager_twobit_builder/data_manager/twobit_builder.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="twobit_builder_data_manager" name="TwoBit" tool_type="manage_data" version="0.0.4" profile="19.05">
+<tool id="twobit_builder_data_manager" name="TwoBit" tool_type="manage_data" version="0.0.5" profile="19.05">
```
bernt-matthias, line 1: Do we want to use `377` as version?

bernt-matthias (commented):
From the sources:

```
  "   -long          use 64-bit offsets for index.   Allow for twoBit to contain more than 4Gb of sequence. \n"
  "                  NOT COMPATIBLE WITH OLDER CODE.\n"
```
So this leaves two questions to me: 

1. will data table entries generated with `--long` be usable by older tools? Otherwise we might want to consider to version the tool data table entries, or?
2. Should we just use `--long` always?

bgruening:
> So this leaves two questions to me:
>  *   will data table entries generated with --long be usable by older tools? Otherwise we might want to consider to version the tool data table entries, or?
>  *  Should we just use --long always?

I have no idea :(

bernt-matthias:
Then it seems reasonable to leave it as optional argument for now and collect some experience with it... 

Anything else you want to change here?

bernt-matthias:
Can you remove the version here: https://github.com/galaxyproject/tools-iuc/blob/main/data_managers/data_manager_twobit_builder/data_manager_conf.xml xref https://github.com/galaxyproject/galaxy/pull/12255 (maybe we should just do this for all DMs in one push .. otherwise it will be copy pasted for years)

bgruening:
No, please merge or adopt as you like.

bernt-matthias (approved):

PR #5193, opened 2023-03-14
Datamash 1.1.0->1.8 and fix

tuncK:
FOR CONTRIBUTOR:

1) Updated datamash 1.1.0 -> 1.9
2) Current tool errors out if the input matrix is too big. This seems to be inherent to datamash itself and can happen even if 1 << input size < RAM (`tool_script.sh: line 27: 1772740 Bus error (core dumped) datamash`) I propose splitting the matrix into chunks, transposing the chunks and then combining the results, if the input is big.

### DIFF
--- a/tools/datamash/datamash-transpose.xml
+++ b/tools/datamash/datamash-transpose.xml
@@ -3,19 +3,57 @@
     <macros>
         <import>macros.xml</import>
     </macros>
-    <expand macro="requirements" />
-    <expand macro="stdio" />
+    <edam_topics>
+        <edam_topic>topic_3570</edam_topic> <!-- Pure math / linear algebra -->
+    </edam_topics>
+    <edam_operations>
+         <!-- <edam_operation>operation_1234</edam_operation> -->
+    </edam_operations>
+    <expand macro="requirements"/>
+    <expand macro="stdio"/>
     <command><![CDATA[
-        datamash transpose
-        @FIELD_SEPARATOR@
-        < $in_file > $out_file
+        #set size_threshold_MB = 1024
+        file_size_MB=\$(du -m $in_file | cut -f 1) &&
+        if [ \$file_size_MB -le $size_threshold_MB ]; then
+            datamash transpose @FIELD_SEPARATOR@ < $in_file > $out_file;
+        else
+            ## Input matrix is very big: divide and conquer
+            ## If the input file is very big, datamash runs out of memory (much earlier than file size ~ available RAM.
+            ## Split into managable chunks of row vectors, transpose the chunks and juxtapose column vector chunks.
+            echo Huge matrix detected &&
+            num_lines=\$(wc -l $in_file | cut -d " " -f 1) &&
+            lines_per_chunk=\$(( $size_threshold_MB * \$num_lines / \$file_size_MB )) &&
+            split -l \${lines_per_chunk} $in_file split_input_ &&
+            for chunk in \$(ls split_input*); do
+                datamash transpose @FIELD_SEPARATOR@ < \$chunk > \${chunk}_transposed;
+            done &&
+            paste split_input_*_transposed > $out_file;
+        fi
     ]]></command>
-    <expand macro="inputs_outputs" />
+    <expand macro="inputs_outputs"/>
     <tests>
-        <test>
-            <param name="in_file" value="datamash_transpose_input.txt" />
-            <output file="datamash_transpose_output.txt" name="out_file" />
+        <test expect_num_outputs="1">
+            <param name="in_file" value="datamash_transpose_input.txt"/>
+            <output file="datamash_transpose_output.txt" name="out_file"/>
         </test>
+        <!-- Test for transposing an extremely big input matrix
+         Disabled to keep the repository size reasonable.
+        For testing, manually download a pathological in- and output from:
+        https://usegalaxy.eu/u/tunc/h/very-big-scrna-matrix
+        -->
+        <!--
+        <test>
+            <param name="in_file" value="big.tabular"/>
+            <output file="transposed_big.tabular" name="out_file"/>
+        </test>
+        -->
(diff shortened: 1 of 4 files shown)


tools/datamash/datamash-transpose.xml
```diff
@@ -3,19 +3,57 @@
...
+    </edam_operations>
+    <expand macro="requirements"/>
+    <expand macro="stdio"/>
     <command><![CDATA[
-        datamash transpose
-        @FIELD_SEPARATOR@
-        < $in_file > $out_file
+        #set size_threshold_MB = 1024
+        file_size_MB=\$(du -m $in_file | cut -f 1) &&
+        if [ \$file_size_MB -le $size_threshold_MB ]; then
```
wm75, line 17: I think it would be preferable to do the size check in cheetah. Given how exotic the usecase is, the goal should be to keep the command line for the regular case unaltered.

tools/datamash/datamash-transpose.xml
```diff
@@ -3,19 +3,57 @@
...
+            ## Split into managable chunks of row vectors, transpose the chunks and juxtapose column vector chunks.
+            echo Huge matrix detected &&
+            num_lines=\$(wc -l $in_file | cut -d " " -f 1) &&
+            lines_per_chunk=\$(( $size_threshold_MB * \$num_lines / \$file_size_MB )) &&
+            split -l \${lines_per_chunk} $in_file split_input_ &&
+            for chunk in \$(ls split_input*); do
+                datamash transpose @FIELD_SEPARATOR@ < \$chunk > \${chunk}_transposed;
+            done &&
+            paste split_input_*_transposed > $out_file;
+        fi
```
wm75, line 31: I care a lot less about the special case, so essentially whatever works is fine with me, but:

Looking at https://git.savannah.gnu.org/gitweb/?p=datamash.git;a=blob;f=src/text-lines.h;h=048f8141362dcebbbd93919f34a2e9d5fb5503b4;hb=HEAD#l31 the memory overhead you're observing may come from the fact that the code does not only allocate memory for lines, but also a pointer into that memory per field/column. So in your example file with > 26,000 rather short field values per line, you may well require more memory for those pointers than for the actual line content.
This suggests that a more complex formula for calculating the number of chunks may be required? Maybe your approximation is completely sufficient though as a rule of thumb and, in that case, don't overengineer but just go with what's working.

tuncK: I saw this last one, but it says "-C" is not available on all versions of split. Do I need to worry about this?

bgruening: not if you require a split version, it that tool in the coreutils package?

tools/datamash/datamash-transpose.xml
tuncK: Done.

tools/datamash/datamash-transpose.xml
tuncK: While testing, it turned out that `-C` does not throw an error if the content of 1 line is bigger than the chunk size but rather splits the lines (i.e. column of the prospective output) resulting in a numerically incorrect output (shuffles rows) for sparse matrix inputs. I hence reverted back to `-l`.

tuncK:
Good catch, thanks @wm75. When I have time, I probably will try to fix datamash's source somehow. In addition to the additional 8B/cell that you  pointed out, there is also the full storage in memory of input (https://git.savannah.gnu.org/gitweb/?p=datamash.git;a=blob;f=src/datamash.c;h=65d6e7eb1dafac8a2490a8dd48c63f4396290adf;hb=HEAD#l744) and output. These, and potentially some others, collectively increase the memory demand to ~5 fold of the input size.

One more look and a merge will be appreciated for the time being.

wm75 (approved):
Looks good to me. Thanks @tuncK !

tools/datamash/datamash-transpose.xml
tuncK: Yes, it already happened while trying to transpose a matrix with long headers (Favouritismus_organismus_gutA_fo39_Gchr1234_ENSBML012345678910...), but mostly 0 (1B) entries. :)

I like fatal errors over non-fatal. In the -l 0 case, the tool errors out in -C, it goes ahead and scrambles. A way out is of course could be to check if there are any split_input_xyz files of size larger than size_threshold_MB, but that increases the bash code amount.

tuncK:
Added @wm75's latest idea of specifying number of chunks instead.

tools/datamash/datamash-transpose.xml
```diff
@@ -3,19 +3,57 @@
...
+        #set size_threshold_MB = 1024
+        #if $file_size_MB <= $size_threshold_MB:
+            datamash transpose @FIELD_SEPARATOR@ < $in_file > $out_file
+        #else:
+            ## Input matrix is very big: divide and conquer
+            ## If the input file is very big, datamash runs out of memory (much earlier than file size ~ available RAM.
+            ## Split into managable chunks of row vectors, transpose the chunks and juxtapose column vector chunks.
+            #set num_chunks = 1 + int(file_size_MB/size_threshold_MB)
+            echo Huge matrix detected, processing in $num_chunks chunks. &&
+            split -n l/$num_chunks \${lines_per_chunk} $in_file split_input_ &&
```
wm75, line 26:
```suggestion
            split -n l/$num_chunks $in_file split_input_ &&
```
I assume?

tuncK: Yes, good catch and surprised why the other one worked at all.

wm75: Really nice otherwise. Line count's gone and, if I'm not mistaken, this is even more robust than the `split -l` way cause neither datamash nor paste seem to have issues with empty files (in case you specify more chunks than there are lines in the file).

PR #5196, opened 2023-03-17
Add scenic.R

pcm32:
FOR CONTRIBUTOR:

This PR adds the SCENIC R package https://github.com/aertslab/SCENIC .

### DIFF
--- a/tools/scenic/scenic.R
+++ b/tools/scenic/scenic.R
@@ -1,0 +1,43 @@
+library(SCENIC)
+library(SCopeLoomR)
+library(optparse)
+
+# Set up command line options
+option_list <- list(
+  make_option("-i", "--input-file", dest="input_file", type="character", 
+              help="Input expression matrix file (in tabular format)"),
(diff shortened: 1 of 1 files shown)


pcm32:
Might be better than against the python tool though... it probably also reads AnnData out of the box.

tools/scenic/macros.xml
```diff
@@ -0,0 +1,9 @@
+<macros>
+  <token name="@TOOL_VERSION">1.1.2</token>
+  <xml name="requirements">
+        <requirements>
+            <requirement type="package" version="@TOOL_VERSION@">no-package-for-r-scenic</requirement>
```
bgruening, line 5: Should I create one for you?

pcm32: I suspect that we should better do this with the new python scenic rewrite, which seems to be what is going to be maintained. What do you think?

bgruening: Makes sense! python for the win!

PR #5199, opened 2023-03-21
Reduce number of github action runs

wm75:
FOR CONTRIBUTOR:

Wonder what you think about this and whether you'd see unwanted side effects:
The idea here is to run the pr workflow only on PRs that are ready for review or on pushes to the main branch.

This has the following effects:
- when a draft PR gets opened, all tests will be skipped until the PR is marked as ready for review, which will then directly trigger runs of all of them
- when a PR gets converted to a draft, this will stop any tests to run on any new commits to it
- tests will not run in duplicate anymore for pushes to PRs

Taken from: https://engineering.leanix.net/blog/halve-your-github-actions-bill/

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -1,5 +1,10 @@
 name: Galaxy Tool Linting and Tests for push and PR
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review]
 env:
   GALAXY_FORK: galaxyproject
   GALAXY_BRANCH: release_22.05
@@ -21,6 +26,7 @@ jobs:
   setup:
     name: Setup cache and determine changed repositories
     runs-on: ubuntu-latest
+    if: github.event_name != 'pull_request' || !github.event.pull_request.draft
     outputs:
       galaxy-head-sha: ${{ steps.get-galaxy-sha.outputs.galaxy-head-sha }}
       repository-list: ${{ steps.discover.outputs.repository-list }}
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -3,8 +3,9 @@ on:
   push:
     branches:
       - main
+      - master
   pull_request:
-    types: [opened, reopened, synchronize, ready_for_review]
+    types: [opened, reopened, synchronize, ready_for_review, review_requested]
 env:
   GALAXY_FORK: galaxyproject
   GALAXY_BRANCH: release_22.05


wm75:
> * tests will not run in duplicate anymore for pushes to PRs

which is not the case anyway anymore I guess for PRs based on an up-to-date main branch?

wm75:
> > * tests will not run in duplicate anymore for pushes to PRs
> 
> which is not the case anyway anymore I guess for PRs based on an up-to-date main branch?

Ah, which currently happens when the PR is based on a branch of this repo instead of a fork.

.github/workflows/pr.yaml
```diff
@@ -1,5 +1,10 @@
 name: Galaxy Tool Linting and Tests for push and PR
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review]
```
bernt-matthias, line 7: Then pushes to non-draft PRs won't trigger the workflow at all and we need to draft-undraft or close/open them which seems a bit awkward, or?

We could add a slash command.

But from the perspective of a reviewer I just want to look at the finished tests and not need to trigger tests and wait for them to finish.

Personally I have to admit that I often use the github infrastructure for the tests (from an eco perspective it does not matter where the energy is spent) which allows me to work on other things (eg other tools).

.github/workflows/pr.yaml
```diff
@@ -21,6 +26,7 @@ jobs:
   setup:
     name: Setup cache and determine changed repositories
     runs-on: ubuntu-latest
+    if: github.event_name != 'pull_request' || !github.event.pull_request.draft
```
bernt-matthias, line 29: I don't understand this. Won't this stop all runs except push to master?

bernt-matthias (commented):
Like the idea to reduce our eco footprint. But I'm not sure how big the impact will be. I guess the cumulative time is the same order of magnitude as the CI workflow... But maybe in this respect the constant factor matters? 

Furthermore I'm hesitant for the following reasons
- my own usage patter for IUC's GA 😅
- contributors will stumble over this non-standard behavior and I'm afraid fast this will cause some extra work for us (educating them, documenting this.. If, then we should do this in the PR template).
- maybe it's also not a good idea to use the draft state. But an extra label would be better?

.github/workflows/pr.yaml
wm75: > Then pushes to non-draft PRs won't trigger the workflow at all and we need to draft-undraft or close/open them which seems a bit awkward, or?

No, this will continue working just fine as before.

.github/workflows/pr.yaml
wm75: No, this will allow runs if the trigger event is not a PR change, but a direct push to main **or** if the PR is not a draft PR.

wm75:
> Like the idea to reduce our eco footprint. But I'm not sure how big the impact will be. I guess the cumulative time is the same order of magnitude as the CI workflow... But maybe in this respect the constant factor matters?
> 

I do not understand this part. Why do you think that the cumulative time is the same order of magnitude? Potentially, you could update a draft PR with dozens of individual commits without ever triggering a CI run, then do it once when things are ready for review.

> Furthermore I'm hesitant for the following reasons
> 
>     * my own usage patter for IUC's GA sweat_smile
> 
>     * contributors will stumble over this non-standard behavior and I'm afraid fast this will cause some extra work for us (educating them, documenting this.. If, then we should do this in the PR template).
> 

Yes, this is exactly what I have in mind as the next step :-)

>     * maybe it's also not a good idea to use the draft state. But an extra label would be better?

Agreed that using the draft state might not be the ideal solution for every use case, but it's just the most straightforward to implement. The idea would be to start with this, gather some experience on how this works in practice, then see if we need a more complex system with more granularity.

Let me sketch how I imagine this to be used:
- in the PR template suggest that opening the PR as a draft PR might be a good idea (and explain the eco aspect)
- if you're sure the PR doesn't need much dicussion, but is *just* about technical review, **and** you have fully prepared the content locally, use a normal PR and have CI run or remove the draft state right after looking at the diff one last time.
- in all other cases, open a draft PR and ask for conceptual feedback first, or keep pushing updates until you're ready for review
  for me, that would actually apply quite often cause I find myself pushing commits regularly because:
  - I want to go home, but better have the current state of my work online
  - I'm going to switch devices and just want to sync changes between local repos
  - I know right away that I need feedback on my proposed changes
  - I know how my brain works and that I'll spot my own mistakes only a second after I've pushed to github
  So I would often start out in draft state, then change that a little later (from back at home, after thinking a little while about my own chnages, after getting initial positive feedback).
- educate reviewers that a draft state does not *strictly* mean no reviewing, but that the author may have asked for feedback on the draft
- we could also think about returning to draft state if there are extensive reviewer comments that you'd want to address one-by-one, then switch back out of it when you're ready for a new round of reviewing.

An additional next step I'd like to take is to have the planemo autoupdate bot use draft PRs. There specifically one of the following often applies:
- you find that the upstream tool update only affects command line functionality not relevant for the Galaxy wrapper
- nobody has time to review the autoupdate PR before the next version of the upstream tool gets released and the bot updates the PR anyway
- you find there are many new command line options so a more complex PR is required anyway
In all these cases, CI runs are just wasteful. Instead you could, when you really decide to review the bot PR, just remove the draft state and, yes, wait a little for CI to finish.

.github/workflows/pr.yaml
bernt-matthias: > > Then pushes to non-draft PRs won't trigger the workflow at all and we need to draft-undraft or close/open them which seems a bit awkward, or?
> No, this will continue working just fine as before.

OK. So the default is `opened, reopened, synchronize` and you add `ready_for_review` (which happens when the draft state is removed. Maybe add `review_requested`?

.github/workflows/pr.yaml
bernt-matthias: So I guess `synchronize` is when there is a push to the PR branch?

bernt-matthias:
Wondering if [skip ci](https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/) might be a better solution for some of the use cases. Question is what is the better default.

.github/workflows/pr.yaml
wm75: Indeed: https://github.com/orgs/community/discussions/24567

wm75:
> Wondering if [skip ci](https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/) might be a better solution for some of the use cases. Question is what is the better default.

So one advantage of relying on the draft status over everything else I thought of so far is that you cannot merge a draft. This means that we cannot accidentally merge a PR for which the tests haven't been run.
I don't think skip ci offers that protection?

bernt-matthias:
> I don't think skip ci offers that protection?

We require that tests pass, or? I would then assume that this requires the tests to pass for the last commit.

bernt-matthias:
Just comes to my mind that using `[skip ci]` in commit messages is also "dangerous", because we have to take care to remove it in the commit message of the merge commit (otherwise deployment is skipped) ..

bernt-matthias (approved):
I'm willing to try as long as we document it in the PR template.

wm75:
> I'm willing to try as long as we document it in the PR template.

Thanks! I'll add a commit for it.
Also would like to see what happens if the hanging Check workflow success job times out. Currently this uses `always()` as a condition so I might have to add a draft status check there, too.

.github/workflows/pr.yaml
```diff
@@ -1,5 +1,11 @@
 name: Galaxy Tool Linting and Tests for push and PR
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+      - master
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review, review_requested]
```
mvdbeek, line 8:
```suggestion
```

.github/workflows/pr.yaml
```diff
@@ -21,6 +27,7 @@ jobs:
   setup:
     name: Setup cache and determine changed repositories
     runs-on: ubuntu-latest
+    if: github.event_name != 'pull_request' || !github.event.pull_request.draft
```
mvdbeek, line 30:
```suggestion
```

mvdbeek (changes requested):
Tests should run on draft. I see no reason to slow down review and merging to wait for the tests once you turn them into ready for review.

bgruening:
> Tests should run on draft. I see no reason to slow down review and merging to wait for the tests once you turn them into ready for review.

So your argument is, that the speed of review is reduced? Is that really crucial here?
If a user wants to run tests constantly on every commit you can simply not use the Draft mode as we do now. However, we save some energy via this simple enhancement and a bit of user education and we should care about that imho.

mvdbeek:
Yes. Also, draft is work in progress, of course you want to run tests on that. If you dont want that, use skip ci.

mvdbeek:
> If a user wants to run tests constantly on every commit you can simply not use the Draft mode as we do now.

This is what draft mode is for, to prevent premature merges and see if things work. We don't have to keep running tests for outdated commits and we don't need to run tests on branches, which is a good change in this PR , but I think not running tests on draft is going too far. Because our time is valuable and limited and also fueled by a ton of energy. I'm also gonna say the few hours per month those small CI instances are running is so minor compared to what our HPC and cloud resources are consuming on an every day basis.

wm75:
I do not agree with you @mvdbeek on the perceived size of impacts (even a minor saving on energy is a good thing, and, yes, most such savings will come at a little bit of inconvenience), but I understand that and how you can have a different view on the changes here and can weigh their pros and cons differently.
I've opened #5212 with just the push trigger changes for now.

PR #5219, opened 2023-03-28
Do not initiate tool tests if no tool was touched

wm75:
FOR CONTRIBUTOR:

The intention here is to prevent the entire tool linting and tests workflow from running if the only files changed in a PR or a merge commit are in folders that do not contain tool-associated files.

This is kind of straightforward using paths-ignore, but needs a bit of extra work because of the repo's required status check (see https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore and the note on status checks therein).

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -1,10 +1,20 @@
 name: Galaxy Tool Linting and Tests for push and PR
 on:
   pull_request:
+    paths-ignore:
+      - '.github/**'
+      - 'deprecated/**'
+      - 'docs/**'
+      - '*'
   push:
     branches:
       - main
       - master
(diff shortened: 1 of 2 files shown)


wm75:
Admittedly not a big win since most PRs would not benefit at all from it, but would also come at no downsides - if I can get it working correctly.

wm75:
The 1 successful check here demonstrates the effect of the PR.
https://github.com/wm75/tools-iuc/pull/6 - a PR against the PR branch here - shows that if a PR changes a file outside the paths-ignore list, it still triggers the regular tests.

mvdbeek (approved):

PR #5242, opened 2023-04-11
Circos and PR workflow fixes

bernt-matthias:
FOR CONTRIBUTOR:


.github/workflows/pr.yaml
```diff
@@ -334,6 +334,12 @@ jobs:
       with:
         path: ~/.cache/pip
         key: pip_cache_py_${{ matrix.python-version }}_gxy_${{ needs.setup.outputs.galaxy-head-sha }}
+    - name: Check if all test chunks succeeded
```
nsoranzo, line 337: I think you also need to change line 319 above to:

```
    if: ${{ always() && needs.setup.outputs.repository-list != '' }}
```

so it is not skipped if one of the `test` runs is skipped.

bernt-matthias:
> I think you also need to change line 319 above

Lets try :) Thanks for the suggestion.

Wondering if it might be a solution to forbid `skipped` here https://github.com/galaxyproject/tools-iuc/blob/1d443e73d2eb888660bbbc7af198f5bcca9c1a70/.github/workflows/pr.yaml#L432 (and maybe in all other tasks of this job except for the file size task).

Somehow I remember that we added this only to allow for PR that do not change tools .. but this might be better solved by https://github.com/galaxyproject/tools-iuc/pull/5219

bernt-matthias:
xref https://github.com/galaxyproject/tools-iuc/commit/4cc19b21e0bf4d6c3dabfef1a5ebad0558a98887

hexylena (approved):
looks fine for me

wm75:
> Somehow I remember that we added this only to allow for PR that do not change tools .. but this might be better solved by https://github.com/galaxyproject/tools-iuc/pull/5219

Not quite, I'm afraid. #5219 only serves as a shortcut for obvious cases of "no tool can possibly be affected by this". You can still have changes *witihin* tool directories that are not supposed to trigger tests. So #5219 can act as a prefilter that gets simple cases out of the way fast, but you still need the more fine-grained logic of the planemo action for everything that passes that first check.

bernt-matthias:
> You can still have changes witihin tool directories that are not supposed to trigger tests.

Do you have an example?

tools/circos/macros.xml
```diff
@@ -8,13 +8,14 @@
...
         <requirement type="package" version="0.6.6">bcbiogff</requirement>
-        <requirement type="package" version="1.72">biopython</requirement>
+        <!-- tiles-from-gff3.py and text-from-gff3.py needs UnknownSeq which is removed in 1.81  -->
+        <requirement type="package" version="1.79">biopython</requirement>
         <requirement type="package" version="0.3.18">pybigwig</requirement>
         <requirement type="package" version="0.23">circos-tools</requirement>
-        <requirement type="package" version="3.4">grep</requirement>
-        <requirement type="package" version="5.1.0">gawk</requirement>
+        <requirement type="package" version="3.10">grep</requirement>
+        <!-- <requirement type="package" version="5.1.0">gawk</requirement> -->
```
nsoranzo, line 18:
```suggestion
```

bernt-matthias: done

nsoranzo (approved):

bernt-matthias:
And the last question before merging: should we bump? I would say no...

PR #5255, opened 2023-04-19
Update metaphlan: error on alternative analysis types

tuncK:
Addresses the Cheetah error encountered if a non-default analysis type is chosen:

...
  File "/opt/galaxy/venv/lib64/python3.8/site-packages/Cheetah/Template.py", line 1053, in __unicode__
  File "cheetah_DynamicallyCompiledCheetahTemplate_1681888750_8410845_53299.py", line 401, in respond
NameMapper.NotFound: cannot find 'tax_lev' while searching for 'analysis.analysis_type.tax_lev.tax_lev'

FOR CONTRIBUTOR:

### DIFF
--- a/tools/metaphlan/metaphlan.xml
+++ b/tools/metaphlan/metaphlan.xml
@@ -192,16 +192,18 @@ metaphlan
 mv 'bowtie2out' '$bowtie2out'
 #end if
 
-#if $analysis.analysis_type.tax_lev.tax_lev == 'a' and $analysis.analysis_type.tax_lev.split_levels
-&&
-mkdir 'split_levels'
-&&
-python '$__tool_directory__/formatoutput.py'
-    split_levels
-    --metaphlan_output '$output_file'
-    --outdir 'split_levels'
-    $out.legacy_output
 
+#if $analysis.analysis_type.t in ['rel_ab', 'rel_ab_w_read_stats']
+    #if $analysis.analysis_type.tax_lev.tax_lev == 'a' and $analysis.analysis_type.tax_lev.split_levels
+        &&
+        mkdir 'split_levels'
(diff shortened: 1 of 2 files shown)


tools/metaphlan/metaphlan.xml
```diff
@@ -1054,6 +1057,34 @@ python '$__tool_directory__/formatoutput.py'
...
+                    <param name="db_selector" value="cached"/>
+                    <param name="cached_db" value="test-db-20210409"/>
+                </conditional>
+            </section>
+            <section name="analysis">
+                <conditional name="analysis_type">
+                    <param name="t" value="marker_ab_table"/>
+                </conditional>
+            </section>
+            <output name="output_file" ftype="tabular" file="SRS014464-Anterior_nares-legacy-abundances.tabular" compare="sim_size">
```
bernt-matthias, line 1080: Could you add a test for the `levels` output / adapt all the tests if necessary.

I guess the filter for this output is also wrong .. I would guess that it might in the outputs for every test because it does not check for the value of `analysis["analysis_type"]["t"]`.

bernt-matthias (commented):
Good finding. Could you check the `levels` output?

tools/metaphlan/metaphlan.xml
tuncK: You suspicion was right, I fixed the filter. All previous tests were using the default case, so they are not impacted. That levels collection is absent is already under testing implicitly by "expect_num_outputs=x", I updated the new test accordingly.

bernt-matthias: Cool. Thanks

bernt-matthias (approved):

PR #5262, opened 2023-04-20
Updating tools/bwameth from version 0.2.6 to 0.2.7

planemo-autoupdate:

I have updated tools/bwameth from version 0.2.6 to 0.2.7.

### DIFF
--- a/tools/bwameth/bwameth.xml
+++ b/tools/bwameth/bwameth.xml
@@ -78,7 +78,7 @@ bwameth.py
 #else:
     $read1 $read2
 #end if
-| samtools sort -@\${GALAXY_SLOTS:-4} -T "\${TMPDIR:-.}" -O bam -o output.bam
+| samtools sort --no-PG -l 0 -T "\${TMPDIR:-.}" -O bam | samtools view --no-PG -O bam -@ \${GALAXY_SLOTS:-1}  -o output.bam
 
     ]]></command>
     <inputs>
--- a/tools/bwameth/bwameth.xml
+++ b/tools/bwameth/bwameth.xml
@@ -154,7 +154,7 @@ bwameth.py
 
 **What it does**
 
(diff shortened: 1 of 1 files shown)


tools/bwameth/bwameth.xml
```diff
@@ -78,7 +78,8 @@ bwameth.py
 #else:
     $read1 $read2
 #end if
-| samtools view --no-PG -u - | samtools sort --no-PG -@ "\${GALAXY_SLOTS:-4}" -T "\${TMPDIR:-.}" -O bam -o output.bam -
+| samtools sort -@\${GALAXY_SLOTS:-4} -T "\${TMPDIR:-.}" -O bam -o output.bam
```
bernt-matthias, line 81: Please undo this change and maybe add the explanation from here https://github.com/galaxyproject/tools-iuc/blob/3e7bf5ae62de3520635d75e3825701960b9722e4/tools/hisat2/hisat2.xml#L341

Maybe use exactly the command line from over there: 

`| samtools sort --no-PG -l 0 -T "\${TMPDIR:-.}" -O bam | samtools view --no-PG -O bam -@ \${GALAXY_SLOTS:-1}  -o output.bam`

bgruening:
I did, but now I have `sort: unrecognized option '--no-PG'`

bernt-matthias:
> I did, but now I have `sort: unrecognized option '--no-PG'`

Might bwameth come with a very old samtools version? 

Anyway, maybe we should remove it anyway, we recently came to the conclusion that we should not use it: https://github.com/galaxyproject/tools-iuc/pull/5985#discussion_r1590388227

bernt-matthias:
Can we easily find out which samtools verison is used here? bioconda seems to have no pinning on samtools. I would expect something recent .. seems fishy.

bgruening:
> (planemo) bag@bag:~/projects/code/tools-iuc/tools/jbrowse2$ docker run -i -t --rm quay.io/biocontainers/bwameth:0.2.7--pyh7cba7a3_0 bash
root@3b599841503d:/# samtools --version
samtools 1.6
Using htslib 1.6
Copyright (C) 2017 Genome Research Ltd.

bernt-matthias (approved):
If you like we can proceed. 

samtools 1.6 is 7 years old (container only 1 year) if we want something newer we could bump the container, but this would delay things here. 

Up to you :)

bgruening:
I tried, but realisticially will not have time to work on this "tab" in the next month.

Issue #4425, opened 2022-02-25
DESeq2: enable custom defined size factor


gallardoalba:
A user in Ghelp asked about the possibility of enabling custom-defined size factors: [Ghelp post](https://help.galaxyproject.org/t/how-to-set-up-user-defined-size-factor-in-deseq2/7593).

Orange220:
Yes, enabling custom-defined size factor is useful in spike-in normalization experiments. It seems to haven't been added in the latest version (Galaxy Version 2.11.40.7+galaxy2). Looking forward to having it.

Orange220:
To clarify, what we need to custom-defined, is not the"Method for estimateSizeFactors", but the value of the SizeFactor.

Issue #3048, opened 2020-05-22
betaPrior in DEseq2 wrapper


drosofff:
The betaPrior parameter is [exposed](https://github.com/galaxyproject/tools-iuc/blob/master/tools/deseq2/deseq2.R#L267) in the DESeq2.R script, but not in the XML wrapper.

As a result, the Galaxy DESeq2 is constitutively `betaPrior` "not off". Is it an omission or for statistical consistency ?

PR #5294, opened 2023-05-09
Update Deseq2 to v. 1.38

tuncK:
FOR CONTRIBUTOR:

I  wish there would be a way to fix the displayed version numbers to something that follows the version of deseq2 itself. Displaying "2.11.40.8" is misleading.

### DIFF
--- a/tools/deseq2/deseq2.R
+++ b/tools/deseq2/deseq2.R
@@ -36,12 +36,6 @@ options(show.error.messages = FALSE, error = function() {
   q("no", 1, FALSE)
 })
 
-# we need that to not crash galaxy with an UTF8 error on German LC settings.
-# As if "loc <- Sys.setlocale("LC_MESSAGES", "en_US.UTF-8")" is not needed anymore.
-# Sometimes some tests break, but not 100% reproducible.
-#loc <- Sys.setlocale("LC_MESSAGES", "en_US.UTF-8")
-
-
 library("getopt")
 library("tools")
 options(stringAsFactors = FALSE, useFancyQuotes = FALSE)
--- a/tools/deseq2/deseq2.R
+++ b/tools/deseq2/deseq2.R
@@ -37,8 +37,7 @@ options(show.error.messages = FALSE, error = function() {
 })
 
 # we need that to not crash galaxy with an UTF8 error on German LC settings.
-# The line below breaks test #11 and as if not needed anymore?
-# loc <- Sys.setlocale("LC_MESSAGES", "en_US.UTF-8")
+# As if "loc <- Sys.setlocale("LC_MESSAGES", "en_US.UTF-8")" is not needed anymore breaks test #11
 
 
 library("getopt")
--- a/tools/deseq2/deseq2.R
+++ b/tools/deseq2/deseq2.R
@@ -37,7 +37,9 @@ options(show.error.messages = FALSE, error = function() {
 })
 
 # we need that to not crash galaxy with an UTF8 error on German LC settings.
-loc <- Sys.setlocale("LC_MESSAGES", "en_US.UTF-8")
+# The line below breaks test #11 and as if not needed anymore?
+# loc <- Sys.setlocale("LC_MESSAGES", "en_US.UTF-8")
+
 
 library("getopt")
 library("tools")
@@ -69,7 +71,7 @@ spec <- matrix(c(
   "outlier_replace_off", "a", 0, "logical",
   "outlier_filter_off", "b", 0, "logical",
   "auto_mean_filter_off", "c", 0, "logical",
-  "beta_prior_off", "d", 0, "logical",
+  "use_beta_priors", "d", 0, "logical",
   "alpha_ma", "A", 1, "numeric",
   "prefilter", "P", 0, "logical",
   "prefilter_value", "V", 1, "numeric"
@@ -312,12 +314,15 @@ if (is.null(opt$auto_mean_filter_off)) {
 }
 
 # shrinkage of LFCs
-if (is.null(opt$beta_prior_off)) {
-  beta_prior <- TRUE
-} else {
+if (is.null(opt$use_beta_priors)) {
   beta_prior <- FALSE
-  if (verbose) cat("beta prior off\n")
+  if (verbose)
+    cat("Applied default - beta prior off\n")
+} else {
+  beta_prior <- opt$use_beta_priors
 }
+sprintf("use_beta_prior is set to %s", beta_prior)
+
 
 # dispersion fit type
 if (is.null(opt$fit_type)) {
--- a/tools/deseq2/deseq2.R
+++ b/tools/deseq2/deseq2.R
@@ -229,7 +229,7 @@ if (!is.null(opt$esf)) {
     } else {
         sf_table <- read.table(opt$esf)
         # Sort the provided size factors just in case the order differs from the input file order.
-        merged_table <- merge(sample_table, sf_table, by.x=0, by.y=1, sort=F)
+        merged_table <- merge(sample_table, sf_table, by.x = 0, by.y = 1, sort = FALSE)
         sf_values <- as.numeric(unlist(merged_table[5]))
         "sizeFactors"(dds) <- sf_values
 
--- a/tools/deseq2/deseq2.R
+++ b/tools/deseq2/deseq2.R
@@ -37,7 +37,9 @@ options(show.error.messages = FALSE, error = function() {
 })
 
 # we need that to not crash galaxy with an UTF8 error on German LC settings.
-# As if "loc <- Sys.setlocale("LC_MESSAGES", "en_US.UTF-8")" is not needed anymore breaks test #11
+# As if "loc <- Sys.setlocale("LC_MESSAGES", "en_US.UTF-8")" is not needed anymore.
+# Sometimes some tests break, but not 100% reproducible.
+loc <- Sys.setlocale("LC_MESSAGES", "en_US.UTF-8")
 
 
 library("getopt")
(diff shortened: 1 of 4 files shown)


tuncK:
Not much change in DeSeq2 itself. I started this after noticing that the DeSeq2 version being used is very hard to figure out for a typical user.

Now addressing: https://github.com/galaxyproject/tools-iuc/issues/3048. I added a boolean switch into the wrapper to turn beta priors on/off.

Looking at https://www.bioconductor.org/packages/release/bioc/manuals/DESeq2/man/DESeq2.pdf, "In
versions >=1.16, the default is set [betaPrior] to FALSE, and shrunken LFCs are obtained
afterwards using lfcShrink". Galaxy wrapper's default is betaPrior = true. We should switch I guess?

bernt-matthias:
> I started this after noticing that the DeSeq2 version being used is very hard to figure out for a typical user.

True .. at least from the tool version. Does the `version_command` is of any help for the user?

> Galaxy wrapper's default is betaPrior = true. We should switch I guess?

IMO yes.

tools/deseq2/deseq2.R
```diff
@@ -37,7 +37,9 @@ options(show.error.messages = FALSE, error = function() {
 })
 
 # we need that to not crash galaxy with an UTF8 error on German LC settings.
-loc <- Sys.setlocale("LC_MESSAGES", "en_US.UTF-8")
+# The line below breaks test #11 and as if not needed anymore?
```
bgruening, line 40: I assume but are not sure, that we added this to fix the `,` vs `.` problem.

tuncK: Leaving it as is causes "Error in (function (x)  : attempt to apply non-function" in R somehow on my VM.

bgruening: Interesting, this is unexpected. But yes maybe that is not a problem anymore.

bgruening: Thanks for including a recent profile!

tuncK:
Addressed https://github.com/galaxyproject/tools-iuc/issues/4425, added an option to provide custom size factors as a tabular from the history.

bernt-matthias (approved):

tools/deseq2/deseq2.xml
```diff
@@ -105,6 +109,9 @@ Rscript '${__tool_directory__}/deseq2.R'
     #if $advanced_options.auto_mean_filter_off:
         -c
     #end if
+    #if $advanced_options.use_beta_priors:
+        -d
+    #end if
```
bernt-matthias, line 114: This may be simplified to 

```suggestion
    $advanced_options.use_beta_priors
```

tools/deseq2/deseq2.xml
```diff
@@ -207,28 +224,31 @@ Rscript '${__tool_directory__}/deseq2.R'
...
                 </param>
                 <when value="-P">
-                    <param name="prefilter_value" type="integer" min="0" value="10" label="Pre-filter value" help="Keep only rows that have at least N reads total." />
+                    <param name="prefilter_value" type="integer" min="0" value="10" label="Pre-filter value" help="Keep only rows that have at least N reads total."/>
                 </when>
                 <when value=""/>
             </conditional>
+            <param name="use_beta_priors" type="boolean" truevalue="1" falsevalue="0" checked="false"
+                label="Use beta priors"
+                help="Whether or not to put a zero-mean normal prior on the non-intercept coefficients"/>
```
bernt-matthias, line 233: If this is changed to 

```
            <param name="use_beta_priors" type="boolean" truevalue="-d" falsevalue="" checked="false"
                label="Use beta priors"
                help="Whether or not to put a zero-mean normal prior on the non-intercept coefficients"/>
```

Or? May be applicable to other booleans as well. 

Implement at will :)

tuncK: Done.

tuncK:
I am stuck with "planemo test" errors and I dont know how to proceed:

1) On my end, 14 out of 14 pass with or without --biocontainers flag.
2) Here on GH, tests #6, 11,12,13 fail with the very same complaint "Error in (function (x)  : attempt to apply non-function"
3) Also on GH, if I let the locale setting to be performed tests #6,7,10,12 fail instead.

Do any of you have any idea what might be off here? That one of the included conda recipes had an unnoticed dependency on something local and we need to rebuild?

bernt-matthias:
Can you call

```
GALAXY_SLOTS=2 planemo test...
```

locally? This should be the only difference...

bernt-matthias:
The container is built  from scratch for each CI run. Maybe you can also try to remove your local image.

Changes in conda are always a possibility.

tuncK:
@bernt-matthias Whatever that irreproducible issue was, is now gone. If you do not have any other suggestions, can you please merge this?

tuncK:
"Incomplete read while reading from https://nodejs.org/download/release/v18.12.1/node-v18.12.1-linux-x64.tar.gz" As if a re-run would fix it?

bernt-matthias:
I restarted it.

bgruening:
restarted and fixed

lldelisle:
The beta parameter change is a big change. We are running a training and using different versions of DESeq2 gives very different results!
How can we warn the users?

lldelisle:
Here is the MA plot with the new default (right) or the beta option to true (like before)
image

tuncK:
Yes, it is supposed to have a significant impact on the results. You can make it behave like in the past by toggling "use beta priors" to true.

pavanvidem:
I did a quick test run with and without the beta priors option on our tutorial data. Yes, the plots look different but the final DE genes set is almost the same. I found 951 out of 954 significant DE genes (`pdj <=0.05`) in common. The main difference is in fold changes which is expected because we do not shrink the fold changes anymore. Previous versions have 113 genes with `abs(FC)>1`. With the current version, we have 204 genes. All 113 genes from the previous version are a subset of the current version. If we increase the FC threshold maybe we get close to the same genes set as before but we need to find a reasonable argumentation for this. Probably `betaPrior=True` makes more sense because, in our Galaxy tool, we do not shrink the fold changes which can be done either with the `lfcShrink` function or `betaPrior=True`. Please refer to this post https://support.bioconductor.org/p/95695/

tuncK:
I have a poor understanding of the theoretical difference between the two sets of p-values. Also here, he says "my current recommendation would be to use the p-values from un-shrunken LFC and then use the shrunken LFC for visualization or ranking of genes. This is the table you get with default DESeq => results => lfcShrink."

My understanding is that one had better introduce the shrinkage function (or even offer multiple options) into deseq2.R, but still report the p-values with beta-prior=False rather than simply recommending the users to set betaPrior=True (which we could also set as the default within the wrapper). If someone could confirm the statistical theory behind it, I would be happy to implement it.

pavanvidem:
Yes, it is even better to call `lfcShrink` after calling `results`. It is also shown in the [vignette](https://bioconductor.org/packages/release/bioc/vignettes/DESeq2/inst/doc/DESeq2.html#differential-expression-analysis) that they used the shrunk FCs for plotting only. I wonder why it was not implemented like this in our wrapper. If we follow the vignette by adding  `lfcShrink` after `results`,  we can remove the `betaPrior` param completely from the Galaxy tool. `DESeq` function generates unshruken FCs by default because `betaPrior` defaults to FALSE. This is the result table we output. Then use the results of `lfcShrink` for plotting.  

In summary, we have to revert back to the version without `betaPrior` option and add `lfcShrink` only before `plotMA` call.

We will still get different results than in the tutorial but the plots will look the same.

bernt-matthias:
> In summary, we have to revert back to the version without betaPrior option and add lfcShrink only before plotMA

I would keep `betaPrior` (it was requested https://github.com/galaxyproject/tools-iuc/issues/3048) and add a  `lfcShrink` call to our R script. 

> We will still get different results than in the tutorial but the plots will look the same.

Thats nice indeed .. but not a strong argument :)

tuncK:
>> In summary, we have to revert back to the version without betaPrior option and add lfcShrink only before plotMA

> I would keep betaPrior (it was requested https://github.com/galaxyproject/tools-iuc/issues/3048) and add a lfcShrink call to our R script.

I think so too. Pavan's suggestion would lead to double-shrink, would not it?

> We will still get different results than in the tutorial but the plots will look the same.

Does anyone here understand the math behind why I am to plot a different p-value rather than the one I am to tabulate? If nobody has a good explanation about it, I think we should contact the authors for clarification.

pavanvidem:
Sure, we can keep it if it was requested. If we set it to FALSE by default, then wouldn't lead to double shrinkage.

The shrinkage applies to MA-plot only which does not plot any p-values but FCs. From the vignette: "It is more useful to visualize the MA-plot for the shrunken log2 fold changes, which remove the noise associated with log2 fold changes from low count genes without requiring arbitrary filtering thresholds.". That makes sense. It affects mainly the genes with a very low mean expression. They usually have very high fold changes which is often just noise. You do not see much difference in FC shrinkage for highly expressed genes.

tuncK:
The FC part is clear to me, p-value is not. Which set of p-values are we going to apply the cutoff's to call DE vs non-DE genes at the end? In the MA-plot above, for instance, you labelled DEs in blue vs gray. But this binary classification is before or after shrinkage and why?

pavanvidem:
I don't think there will be any changes in p-values after shrinkage. The p-values change only when you switch `betaPrior` of `DESeq` function. If you run `DESeq` function followed by  `lfcShrink`, you probably get the same p-values. I haven't tested this though.

bernt-matthias:
Can someone please summarize  this discussion in an issue?

pavanvidem:
Sure, added an issue https://github.com/galaxyproject/tools-iuc/issues/5552

PR #5301, opened 2023-05-11
Update get_organelle_from_reads.xml to handle paired collections correctly 

o-william-white:
Hello, 

Today I found that get_organelle_from_reads.xml was not able to work with "paired_collections" as input. 

I was able to find a couple of potential issues: 
- spelling of "in1_name1" 
- the paired_collection paramter name "fastq_input1" did not match the name used in the command block "paired_input".  

I thought this might fix my issue but I still get the error below: 

```
NameMapper.NotFound: cannot find 'paired_input' while searching for 'fastq_input.paired_input.forward.ext'
```

@bernt-matthias can you see where i might have gone wrong here? 

Fix errors with paired collection parameters

FOR CONTRIBUTOR:

### DIFF
--- a/tools/getorganelle/get_organelle_from_reads.xml
+++ b/tools/getorganelle/get_organelle_from_reads.xml
@@ -249,14 +249,14 @@
                 </assert_contents>
             </output>
             <output_collection name="path_sequence_list" type="list" count="2">
-                <element name="a" ftype="fasta">
+                <element name="embplant_pt.K115.complete.graph1.1" ftype="fasta">
                     <assert_contents>
-                        <has_n_lines n="100"/>
+                        <has_n_lines n="2"/>
                     </assert_contents>
                 </element>
-                <element name="b" ftype="fasta">
+                <element name="embplant_pt.K115.complete.graph1.2" ftype="fasta">
                     <assert_contents>
-                        <has_n_lines n="100"/>
+                        <has_n_lines n="2"/>
                     </assert_contents>
                 </element>
             </output_collection>
@@ -292,7 +292,7 @@
             </output>
         </test>
         <!-- Ensure paired collection works; also check output filters -->
-        <test expect_num_outputs="2">
+        <test expect_num_outputs="3">
             <param name="config_dir" value="getorganelle_refdata"/>
             <conditional name="fastq_input">
                 <param name="fastq_input_selector" value="paired_collection"/>
@@ -304,7 +304,7 @@
                 </param>
             </conditional>
             <param name="F" value="embplant_pt" />
-            <param name="advanced" value="advanced" />
+	    <param name="advanced" value="simple" />
             <assert_stdout>
                 <has_text text="Thank you!" />
             </assert_stdout>
--- a/tools/getorganelle/get_organelle_from_reads.xml
+++ b/tools/getorganelle/get_organelle_from_reads.xml
@@ -278,6 +278,59 @@
                 </assert_contents>
             </output>
         </test>
+        <!-- Ensure paired collection works -->
+        <test expect_num_outputs="8">
+            <param name="config_dir" value="getorganelle_refdata"/>
+            <param name="fastq_input_selector" value="paired_collection"/>
+            <param name="paired_input">
+                <collection type="paired">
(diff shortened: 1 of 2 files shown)


bernt-matthias:
Also con't spot a problem. Could you add a test for this case (e.g. just take the existing input and change select of the conditional and the fastq input to a paired collection).

o-william-white:
Hi @bernt-matthias thanks for the suggestion, I added the test using a paired collection but it passed when I ran it locally.

o-william-white:
[tool_test_output.html.txt](https://github.com/galaxyproject/tools-iuc/files/11462937/tool_test_output.html.txt)

o-william-white:
Html of test attached, I had to change the file ending to .txt so it could be uploaded. 

Can you see an issue with my test? It seems to pass, but when I run the same data using the Galaxy GUI, I get the same error as before.

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -278,6 +278,59 @@
                 </assert_contents>
             </output>
         </test>
+        <!-- Ensure paired collection works -->
+        <test expect_num_outputs="8">
+            <param name="config_dir" value="getorganelle_refdata"/>
+            <param name="fastq_input_selector" value="paired_collection"/>
+            <param name="paired_input">
```
bgruening, line 285: this needs to be inside a conditional

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -53,7 +53,7 @@
             -1 '${$in1_name}' -2 '${$in2_name}'
         #end if
         #if str($fastq_input.fastq_input_selector) == "paired_collection":
-            -1 '${$in1_name1}' -2 '${$in2_name}'
+            -1 '${$in1_name}' -2 '${$in2_name}'
```
bgruening, line 56: Mh, those selectors are wrong I think ... its either `${foo}` or `$foo`. I'm not aware that `${$foo}` will work?

o-william-white: Ah that's a silly mistake! Thanks for spotting!

tools/getorganelle/get_organelle_from_reads.xml
o-william-white: Would you be able to point me to an example of this? I used fastp https://github.com/o-william-white/tools-iuc/blob/master/tools/fastp/fastp.xml as an example to set this up.

tools/getorganelle/get_organelle_from_reads.xml
bernt-matthias: Odd that this worked at all. But, this is in all branches of the if statement...

bernt-matthias:
@o-william-white can you check if the version of my latest commit works as you would expect?

o-william-white:
Hi @bernt-matthias many thanks for the changes, much appreciated. 

I made some additional changes to the tests and both pass now when I use planemo. 

Think this looks a lot better!

tools/getorganelle/get_organelle_from_reads.xml
```diff
@@ -292,7 +292,7 @@
             </output>
         </test>
         <!-- Ensure paired collection works; also check output filters -->
-        <test expect_num_outputs="2">
+        <test expect_num_outputs="3">
```
bernt-matthias, line 295: Can you adapt the test to check for the content of the 3rd output .. or such that test has 2 outputs again?

o-william-white: Test adapted, does this look ok?

bernt-matthias (approved):

bernt-matthias (changes requested):
Please bump the tool version

o-william-white:
Hi @bernt-matthias, I checked the getorganelle version and it is using the most up to date version. Do you mean galaxy tool version? Where would I change this?

bernt-matthias (approved):

bernt-matthias:
Like so https://github.com/galaxyproject/tools-iuc/pull/5301/commits/ce2b9a1b77f81217e47ba4ffd296a7bf506d1902 would be the IUC way to bump the galaxy tool version.

o-william-white:
Awesome, thanks for explaining!

PR #5309, opened 2023-05-16
humann: use versioned metaphlan cached reference

bernt-matthias:
and also allow to use legacy data. 

We have to do this because the old datatable may contain entries that are not usable without the `--mpa3` flag. Alternatively we could filter legacy entries. 

I could also imagine to add a boolean flag allowing for legacy metaphan data from the history, but I don't care so much.

ping @tuncK 

FOR CONTRIBUTOR:


tools/humann/humann.xml
tuncK, line 1: https://github.com/biobakery/humann/blob/3e2fe6b3f1dcd203ad42bf40daee464299932609/humann/config.py#L305

If my understanding is correct, there is a hard-coded path referenced in humann to Oct22. So as if .loc vs _versioned.loc is not restrictive enough?

bernt-matthias:
> Would it be a very very bad idea if we decouple the data tables of humann and metaphlan completely from each other?

What exactly do you mean with "decouple"?

> One side of me also thinks that we should maybe remove/hide from our instance older versions of humann3.x / metaphlan4.x after merging this

We should try to keep them .. reproducibility! But if possible clean data tables from entries that don't work.

tuncK:
> > Would it be a very very bad idea if we decouple the data tables of humann and metaphlan completely from each other?
> 
> What exactly do you mean with "decouple"?
> 
So currently, before and after your PR, humann and metaphlan share the same data table. This works if all metaphlan SGB DBs in that list are compatible with humann 3+. But given hardcoding in humann/config.py, I am suspicious that it is not the case.


> > One side of me also thinks that we should maybe remove/hide from our instance older versions of humann3.x / metaphlan4.x after merging this
> 
> We should try to keep them .. reproducibility! But if possible clean data tables from entries that don't work.
>
The problem is, in the current setting these old versions actually error out, leaving users confused. Compatible DBs are missing / not identifiable from their names. But fatal errors might actually be a blessing: if it ends up calculating some number, I am not sure anymore under what conditions it is accurate. 

A workaround could be to delete 3.x/4.x  DBs on usegalaxy.eu except v. 3.7. If users opt for an outdated version, it becomes the their responsibility to figure out which DBs to download and provide from the history?

bernt-matthias:
Do you think it would help to file an upstream issue? I think we need to know which versions of the two data tables and versions of the tool are compatible....

tuncK:
In April 2023, I had this email exchange with devs and ended up hard-coding the DB, because it sounded as if the tool accepts more DBs without any fatal error, but its implications on the numerical accuracy are uncertain:

"This hard-coding is because we've taken extra steps to allow HUMAnN 3 and its pangenomes to work with MetaPhlAn 4's SGB-based taxonomy profiles (the upcoming HUMAnN 4 will use the SGBs natively, but is still in development). Specifically we need to generate a file that tells us which SGBs correspond to which HUMAnN 3 pangenomes. We made such a file for the Jan21 pangenomes and released it alongside HUMAnN 3.5. To use the recently updated SGBs (Oct22) will require an update on this file, and we are waiting on input from the MetaPhlAn group to help construct it. I expect it will be released as a HUMAnN 3.7. In the meantime I would stick with the previous MetaPhlAn SGB release + HUMAnN 3.6."

bernt-matthias:
@tuncK do you have any suggestion how we could continue here?

tuncK:
> Do you think it would help to file an upstream issue? I think we need to know which versions of the two data tables and versions of the tool are compatible....

Such incompatibilities were causing problems lately and are not caught by tool-specific tests we have in place. So my student @jozh2008 has been developing a new tool that can test compatibility between installed tools/DBs systematically in the past few months.

@jozh2008, do you think we are now ready to give this real life problem a try?

paulzierep:
I see, assuming the DB source servers are back up, I could try to bump this PR directly to the latest version using the versioned metaphlan DB. 
Since humann 3.9 uses metaphlan 4.2.0 internally, we would need to pin the DB to a version compatible with 4.2.0.

PR #5313, opened 2023-05-22
Updating tools/pangolin from version 4.2 to 4.3

planemo-autoupdate:

I have updated tools/pangolin from version 4.2 to 4.3.

### DIFF
--- a/tools/pangolin/pangolin.xml
+++ b/tools/pangolin/pangolin.xml
@@ -1,7 +1,7 @@
 <tool id="pangolin" name="Pangolin" version="@TOOL_VERSION@+galaxy0" profile="20.01">
     <description>Phylogenetic Assignment of Outbreak Lineages</description>
     <macros>
-        <token name="@TOOL_VERSION@">4.2</token>
+        <token name="@TOOL_VERSION@">4.3</token>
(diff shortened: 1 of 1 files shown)


bernt-matthias:
Need to remove the pangolearn mode.. It seems.

Does this tool really need so many requirements?

pvanheus:
I recall some of these requirements were added because of a problem with a certain bioconda build and can probably be removed - @wm75 do you remember the details?

We should definitely add a note to the effect that pangoLEARN mode is deprecated.

wm75:
All those requirements are just pinning core dependencies of pangolin's lineage assignment logic. The bioconda recipe doesn't do that, but it means two pangolin environments created at different times with pangolin itself at the same version could produce rather different results. For the Galaxy wrapper we opted for more reproducibility though it means more maintenance effort: whenever you want to update the wrapper you should try a fresh conda install of pangolin to get the versions that would be installed of all requirements, then update them in the wrapper.

I have just updated the corresponding comment in the wrapper a bit to make the importance of the core dependencies clearer in #5398 .

wm75:
Closing this now as completed in #5391 .

PR #5015, opened 2022-12-19
Add new tool: frameshift_deletions_checks from smallgenomeutilities

moeezmalik:
### About the tool

This pull request adds the tool frameshift_deletions_checks from the repository smallgenomeutilities. The smallgenomeutilities are a collection of scripts that is useful for dealing and manipulating NGS data of small viral genomes.

The tool frameshift_deletions_checks detects frameshifts indels and stop codons in the provided viral consensus sequences and generate a report of the event.

#### Diclaimer
The above mentioned descriptions about the repository and the tool are taken directly from the authors of the tool.

### Tests
There are 4 tests in total for the tool. The tests focus on optional parameters that can be provided when running the tool. 

### Resources

- [smallgenomeutilites Github Repo](https://github.com/cbg-ethz/smallgenomeutilities)
- [smallgenomeutilites Bioconda Package](https://bioconda.github.io/recipes/smallgenomeutilities/README.html)

### For Contributor
* [x] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [x] - License permits unrestricted use (educational + commercial)
* [x] - This PR adds a new tool or tool collection
* [ ] - This PR updates an existing tool or tool collection
* [ ] - This PR does something else (explain below)

PR #5335, opened 2023-05-30
Adding tool frameshift_deletions_checks from smgu

DrYak:
This pull request adds the tool frameshift_deletions_checks from the repository smallgenomeutilities. 
It combines my own learning of planemo tool and galaxy tool XML in general, with previous work by @moeezmalik  on PR #5015.

Among other, this new PR:
 - handles the different input format (pysam/HTSlib support BAM as well as CRAM)
 - handles temporary files gracefully (avoid the maffs output going into the read-only input directory).
 - much tinier CRAM file for test
 
Background:

The [smallgenomeutilities](https://github.com/cbg-ethz/smallgenomeutilities/) are a collection of scripts that is useful for dealing and manipulating NGS data of small viral genomes.

The tool frameshift_deletions_checks detects frameshifts indels and stop codons in the provided viral consensus sequences and generate a report of the event. (This is useful when submitting to repositories such as GISAID or ENA whose automatic quality checking might complain about frameshifts. This tools will help diagnose the (data-supported) reason of the frameshifts).

(Dislaimer: I am part of the V-pipe devs working on smallgenomeutilites repository)

FOR CONTRIBUTOR:


tools/smallgenomeutilities/frameshift_deletions_checks.xml
```diff
@@ -0,0 +1,93 @@
+<tool id="smgu_frameshift_deletions_checks" name="Frameshift Deletions Checks" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@">
+    <description>Produces a report about frameshifting indels in a consensus sequences</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="xrefs"/>
+    <requirements>
+        <requirement type="package" version="@TOOL_VERSION@">smallgenomeutilities</requirement>
+    </requirements>
+    <!-- <version_command>frameshift_deletion_checks &#x002D&#x002Dversion</version_command> -->
```
bgruening, line 10: that is not working?

tools/smallgenomeutilities/frameshift_deletions_checks.xml
```diff
@@ -0,0 +1,93 @@
...
+    <command detect_errors="exit_code">
+    <![CDATA[
+#if $input.is_of_type("cram"):
+ echo "is CRAM" &&
+ ln -vs "$input" 'input.${input.ext}' &&
```
bgruening, line 15: please use single-quotes instead of double quotes

tools/smallgenomeutilities/frameshift_deletions_checks.xml
DrYak: ...yet. smgu is still oldschool "_version are a bunch of static strings sprinkled all over the code base, most of the out of date_". 
For v0.4.0, I plan to introduce version-from.git tag. This will automatically make `--version` always work.
(Done that for LolliPop already)

tools/smallgenomeutilities/frameshift_deletions_checks.xml
```diff
@@ -0,0 +1,93 @@
...
+    --consensus='consensus.${consensus.ext}'
+    --reference='reference.${reference.ext}'
+    --genes='$genes'
+    --output='$report' 
+#if $orf1ab is not None:
+    --orf1ab='$orf1ab'
+#end if
+#if $english is not None:
+    $english
+#end if
```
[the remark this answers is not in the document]
DrYak: Okay, I might mis understood that when an `optional=True` tool has an option not set, that option is instead set to None. I'll fix it.

tools/smallgenomeutilities/frameshift_deletions_checks.xml
```diff
@@ -0,0 +1,93 @@
...
+#if $zero_based is not None:
+    $zero_based
+#end if
+]]>
+    <!-- ##cores \${GALAXY_SLOTS:-4} -->
+    </command>
+    <inputs>
+        <param argument="--consensus" type="data" format="fasta" label="Consensus" help="Fasta file containing the sample's consensus sequence (majority, with indels)" />
+        <param argument="--input" type="data" format="bam,cram" label="Input BAM" help="Input BAM file with sample's sequencing reads, aligned against the reference" />
+        <param argument="--reference" type="data" format="fasta" label="Reference" help="Fasta file containing the reference sequence (used during alignment) to compare against" />
```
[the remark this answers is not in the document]
DrYak: Nope. We rely on biopython SeqIO under the hood and that one doesn't do on-the-fly decompression.
I am merely compressing the test data to save space, and then letting planemo handle decompression for us during testing.

tools/smallgenomeutilities/frameshift_deletions_checks.xml
```diff
@@ -0,0 +1,91 @@
+<tool id="smgu_frameshift_deletions_checks" name="Frameshift Deletions Checks" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@">
+    <description>Produces a report about frameshifting indels in a consensus sequences</description>
```
bernt-matthias, line 2: Sequences - > sequence

Shorter description would be nice. Keep in mind that name + description is displayed in the tool menu.

tools/smallgenomeutilities/frameshift_deletions_checks.xml
```diff
@@ -0,0 +1,91 @@
+<tool id="smgu_frameshift_deletions_checks" name="Frameshift Deletions Checks" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@">
```
bernt-matthias, line 1: Please add profile

tools/smallgenomeutilities/frameshift_deletions_checks.xml
```diff
@@ -0,0 +1,91 @@
...
+    </macros>
+    <expand macro="xrefs"/>
+    <requirements>
+        <requirement type="package" version="@TOOL_VERSION@">smallgenomeutilities</requirement>
+    </requirements>
+    <!-- once we have version-from-git-tag in 0.4.0: <version_command>frameshift_deletion_checks &#x002D&#x002Dversion</version_command> -->
+    <command detect_errors="exit_code">
+    <![CDATA[
+#if $input.is_of_type("cram"):
+ echo 'is CRAM' &&
```
bernt-matthias, line 14: Remove echo

tools/smallgenomeutilities/frameshift_deletions_checks.xml
```diff
@@ -0,0 +1,91 @@
...
+ ln -vs '$input' 'input.${input.ext}' &&
```
bernt-matthias, line 15: What is -v doing?

tools/smallgenomeutilities/frameshift_deletions_checks.xml
```diff
@@ -0,0 +1,91 @@
...
+        <param argument="--reference" type="data" format="fasta" label="Reference" help="Fasta file containing the reference sequence (used during alignment) to compare against" />
+        <param argument="--genes" type="data" format="gff" label="Genes GFF" help="GFF file listing genes positions on the reference sequence" />
+        
+        <param argument="--english" type="boolean" truevalue="--english" falsevalue="--no-english" checked="true" optional="true" label="Write Summary Diagnosis?" help="If checked writes english summary diagnosis." />
+        <param argument="--zero-based" type="boolean" truevalue="--zero-based" falsevalue="" checked="false" optional="true" label="Use 0-based Sequence Positions?" help="Use 0-based (python) instead of 1-based (standard) seq positions" />
+        <param argument="--orf1ab" type="text" value="cds-YP_009724389.1" optional="true" label="ORF1AB" help="CDS ID for the full Orf1ab CDS, comprising the ribosomal shift. In the GFF this CDS should consist of 2 entries with the same CDS ID due to the partial overlap caused by the ribosomal shift at translation time."/>
```
bernt-matthias, line 51: Maybe add a validator?

tools/smallgenomeutilities/frameshift_deletions_checks.xml
```diff
@@ -0,0 +1,91 @@
...
+ln -vs '$consensus' 'consensus.${consensus.ext}' &&
+ln -vs '$reference' 'reference.${reference.ext}' &&
+
+frameshift_deletions_checks 
+    --input='input.${input.ext}'
+    --consensus='consensus.${consensus.ext}'
+    --reference='reference.${reference.ext}'
+    --genes='$genes'
+    --output='$report' 
+#if $orf1ab is not None:
```
bernt-matthias, line 33: Text parameters are `'' ` if left empty.

Use single quotes around text parameters.

tools/smallgenomeutilities/frameshift_deletions_checks.xml
```diff
@@ -0,0 +1,91 @@
...
+    --orf1ab='$orf1ab'
+#end if
+    '$english'
+#if $zero_based is not None:
```
bernt-matthias, line 37: You can remove the if here and just use `$zero_based`

tools/smallgenomeutilities/frameshift_deletions_checks.xml
```diff
@@ -0,0 +1,91 @@
...
+    $zero_based
+#end if
+]]>
+    <!-- ##cores \${GALAXY_SLOTS:-4} -->
+    </command>
+    <inputs>
+        <param argument="--consensus" type="data" format="fasta" label="Consensus" help="Fasta file containing the sample's consensus sequence (majority, with indels)" />
+        <param argument="--input" type="data" format="bam,cram" label="Input BAM" help="Input BAM file with sample's sequencing reads, aligned against the reference" />
```
bernt-matthias, line 45: Do you require position sorted bam?

tools/smallgenomeutilities/frameshift_deletions_checks.xml
DrYak: Will try  something smaller.

tools/smallgenomeutilities/frameshift_deletions_checks.xml
bernt-matthias: This would be `unsorted.bam` (meaning that no assumptions on the sorting are made .. ie arbitrary bam is allowed). 

Then there are 

- `bam`, i.e. coordiante sorted
- `qname_input_sorted.bam` and
- qname_sorted.bam

tools/smallgenomeutilities/frameshift_deletions_checks.xml
DrYak: Long term plan  would be to have the tool autodetect overlapping CDSes and deduce ribosomal shifts in a generic manner. I wouldn't lose too much time on creating an input validator at this stage, and concentrate dev efforts (mostly me... :roll_eyes: ) on making my colleagues' prototype more generic.

tools/smallgenomeutilities/frameshift_deletions_checks.xml
DrYak: okay, I definitely completely misunderstood how optional work.

tools/smallgenomeutilities/frameshift_deletions_checks.xml
DrYak: Oh, got it: https://galaxyproject.org/support/sort-your-inputs/
- `bam` is for position-sorted.
- `unsorted.bam` is when no sorting is required.
So current wording is correct.

tools/smallgenomeutilities/frameshift_deletions_checks.xml
DrYak: What's the quoting rule, does galaxy correctly escape ` ' `  when passing them?
If that's the case, then no matter what the use types, the arbitrary string will correctly be passed to python's arg parse.
If that's not the case, yes we're then at risk of somebody typing `' & do_something_nefarious & true '` and I should at least use a regex like `[^\']*`

bernt-matthias: Default is defined here https://github.com/galaxyproject/galaxy/blob/290592b20f7450beb799cdd79016089ec2ab00c7/lib/galaxy/tools/parameters/sanitize.py#LL45C89-L45C89

Guess single quoting the parameter value in the command section would be safe.

tools/smallgenomeutilities/frameshift_deletions_checks.xml
bernt-matthias: maybe redirect debug output to stderr?

tools/smallgenomeutilities/frameshift_deletions_checks.xml
DrYak: Okay, this is the point where I need expert help.

The way I read [the documentation](https://docs.galaxyproject.org/en/master/dev/schema.html#character-presets):

>  This tag set is used to replace the basic parameter sanitization with custom directives.
> …
> default (equal to string.ascii_letters + string.digits + " -=_.()/+*^,:?!")

i.e. if I don't touch and over-ride  anything, there's already a basic sanitizer, and the default set of allowed characters doesn't include any quote (`'` nor `"`) nor any backtick-lick not expansion (no <tt>\`</tt>, no `${}`, and  `()` only tolerated without `$`), and removes some command chaining (`&&`, `||`) and lacks ways to further escape stuff (no ` \ `). Still has globing capability (`*` and `?`) and a few other (`! `, `;`) but these are thwarted simply by quoting.

So am I correct that the default behavior is good enough (in the context of Python arg parse that can swallow any arbitrary string as long as it's quoted property as in the current Cheetah command template?)

It seems to be  corroborated by my testing:
putting ` ' ` in this parameter, gets detected by the sanitizer and escaped to `__sq__` (aka _single quote_) instead, removing any possibility to break out of the quoted parameter.

bernt-matthias: You are right. Missed that you already single quote the parameter. Sorry for the noise.

tools/smallgenomeutilities/frameshift_deletions_checks.xml
DrYak: Done.

bernt-matthias (approved):

bernt-matthias:
Excellent contribution. Thanks @DrYak

PR #5346, opened 2023-06-06
Add meningotype (tool for Nesseria meningtidis typing)

pvanheus:
FOR CONTRIBUTOR:


tools/meningotype/strip_header.py
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env python3
+
+import sys
+
+if __name__ == '__main__':
+    next(sys.stdin)  # skip first line
+    for line in sys.stdin:
+        sys.stdout.write(line)
```
bgruening, line 8: can this be replace by `tail -n -1`

pvanheus: I am deliberately trying to avoid using `tail` and similar, so as to not add a dependency on `coreutils`.

pvanheus:
So a comment on design after lots of commits to debug [this problem](https://github.com/bioconda/bioconda-recipes/pull/41426): the tool does 3 things besides running the underlying tool:

1. Optionally update the databases used by the tool
3. Check for ERROR text in the tool output
4. Optionally trim the header line of the output

All of these are performed using Python scripts because the underlying tool is already written in Python and thus additional dependencies can be avoided by using Python scripts instead of other command line tools.

Unfortunately due to a bug in the `mlst` bioconda recipe, the current recipe requires `coreutils`. This can be removed when the next release of `meningotype` and thus a working container is available.

tools/meningotype/meningotype.xml
```diff
@@ -0,0 +1,233 @@
...
+    </macros>
+    <requirements>
+        <requirement type="package" version="@TOOL_VERSION@">meningotype</requirement>
+        <!-- coreutils dep added because of a bug in the meningotype 0.8.5 container - it doesn't have `file` -->
+        <requirement type="package" version="9.0">coreutils</requirement>
+    </requirements>
+    <version_command>meningotype --version</version_command>
+    <command detect_errors="aggressive"><![CDATA[
+        #if $db_source == 'network'
+            python3 '$__tool_directory__'/update_db.py &&
```
bgruening, line 14:
```suggestion
            python3 '$__tool_directory__/update_db.py' &&
```

tools/meningotype/meningotype.xml
```diff
@@ -0,0 +1,233 @@
...
+        #end if
+        #if $input.source == "single"
+            #set $input_data = "'" + $input.fasta.name.replace(' uncompressed', '') + "'"
```
bgruening, line 17: Why is this name-mangling needed?
And please use element_identifier instead of `.name`

tools/meningotype/meningotype.xml
```diff
@@ -0,0 +1,233 @@
...
+            #for $el in $input.fastas
+                #set $name = $el.name.replace(' uncompressed', '')
+                ln -s '$el' '$name' &&
+
+                #set $input_data = $input_data + " '" + $name + "'"
+            #end for
+        #end if
+        meningotype 
+        #if $db_source == "network"
+            --db \$(pwd)/db
```
bgruening, line 30:
```suggestion
            --db ./db
```

bgruening:
I think we should use standard tools as much as we can. Less code to maintain is good and outweighs some additional dependencies imho.
This is not a blocker from my side.

tools/meningotype/meningotype.xml
pvanheus: The `meningotype` tool uses the filename to create a sample ID in its output. Since it does not natively support .fasta.gz, Galaxy automatically uncompresses the incoming FASTA and appends `' uncompressed'` to the name. To preserve the original dataset name, I strip this off.

pvanheus:
> I think we should use standard tools as much as we can. Less code to maintain is good and outweighs some additional dependencies imho. This is not a blocker from my side.

I'm not sure which approach is best but I will be engaging with the upstream tool authors to make as much code I have added to the wrapper as possible obsolete. E.g. I hope to make my `update_db.py` script unnecessary.

bgruening (approved):

bgruening:
Thanks @pvanheus

PR #5356, opened 2023-06-12
Pharokka update

paulzierep:
FOR CONTRIBUTOR:

Update tool to use data table with version column

### DIFF
--- a/tools/pharokka/tool-data/pharokka_db.loc.sample
+++ b/tools/pharokka/tool-data/pharokka_db.loc.sample
@@ -2,11 +2,11 @@
 #to use a pharokka DB folder. The pharokka_db.loc 
 #file needs this format (longer white space is the TAB character):
 
-#<unique_build_id>		<dbkey>		<display_name>		<DB_folder_path>
+#<unique_build_id>  <dbkey> <version>   <display_name>  <DB_folder_path>
 
 # for example:
 
-# pharokka_db	pharokka_db	Pharokka Database v1.2.0	/data/pharokka_db
+# pharokka_db	pharokka_db	v1.2.0	Pharokka Database v1.2.0	/data/pharokka_db
 
 # To retrieve the complete DB look at https://github.com/gbouras13/pharokka or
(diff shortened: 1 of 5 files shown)


tools/pharokka/tool-data/pharokka_db.loc.sample
```diff
@@ -2,11 +2,11 @@
 #to use a pharokka DB folder. The pharokka_db.loc 
 #file needs this format (longer white space is the TAB character):
 
-#<unique_build_id>		<dbkey>		<display_name>		<DB_folder_path>
+#<unique_build_id>  <dbkey> <version>   <display_name>  <DB_folder_path>
```
bgruening, line 5: is that on par with the DM?

paulzierep: yes:
https://github.com/galaxyproject/tools-iuc/blob/main/data_managers/data_manager_pharokka_database_fetcher/tool_data_table_conf.xml.sample#L1-L7

bernt-matthias: As far as I know we can't change data tables, ie their columns...

paulzierep: Now that we modified the table on EU to make the tool work, we should probably leave it as is atm and only work on the DM to match that table... is that a correct assumption ?
@bernt-matthias what would the approach if something in the table would need to be changed, i.e. if an updated tool requires more data table columns ?

bernt-matthias: I guess: create a new data table. Then admins may move old entries to the new data table and create new entries in the new column.

As far as I know the behavior of Galaxy in the case is completely undefined. I guess in the best case this "just" breaks old tool versions (which will use the content of the version column as display name and the display name as path). If new columns are added in the middle it definitely breaks things. Not sure what happens if columns are appended. 

Not sure if adding a new column to the end might work (in the future https://github.com/galaxyproject/galaxy/issues/16260).

PR #5360, opened 2023-06-17
adding gzipped files formates to inputs

EngyNasr:
FOR CONTRIBUTOR:
 - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
 - This PR updates an existing tool or tool collection


bgruening:
@EngyNasr please add a test for those filetypes. So we can check that kraken can handle them.

tools/krakentools/extract_kraken_reads.xml
```diff
@@ -129,7 +129,31 @@
...
     </outputs>
     <tests>
-        <!-- test Kraken2 input, single input -->
+    <!-- test Kraken2 input, single input -->
+        <test expect_num_outputs="1">
+            <param name="input_1" value="extract_kraken_reads/R1.fastqsanger.gz" ftype="fastqsanger.gz" />
+            <param name="library|type" value="single" />
+            <param name="results" value="extract_kraken_reads/kraken2.results" ftype="tabular" />
+            <param name="taxid" value="11176" />
+            <output name="output_1" file="extract_kraken_reads/out1.k2.11176.fa" decompress="true" ftype="fasta.gz"  />
```
bgruening, line 138: the path indicates a uncompressed fasta file

bgruening: @bernt-matthias do you know if decompress=true and works here?

tools/krakentools/extract_kraken_reads.xml
```diff
@@ -129,7 +129,31 @@
...
+    <!-- test Kraken2 input, single input -->
+        <test expect_num_outputs="1">
+            <param name="input_1" value="extract_kraken_reads/read1.fastq" ftype="fastq" />
+            <param name="library|type" value="single" />
+            <param name="results" value="extract_kraken_reads/kraken2.results" ftype="tabular" />
+            <param name="taxid" value="11176" />
+            <output name="output_1" file="extract_kraken_reads/extractoutput.fasta.gz" decompress="true" ftype="fasta.gz"  />
+        </test>
+    <!-- test Kraken2 input, single input -->
+        <test expect_num_outputs="1">
```
bgruening, line 149: Those two last tests are more or less identical, would a compressed paired end test be more useful?

EngyNasr:
tool does not handle them you are right, I get the same errors with the tests as I got before, I will try to decompress internally maybe

tools/krakentools/extract_kraken_reads.xml
EngyNasr: I changes the idea a little bit so I removed the test

EngyNasr:
I need to explain why I needed that in the first place: 

For the foodborne preprocessing workflow that takes a collection of samples, there is a Porechop step which output a collection of fastqsanger.gz files. Which I use afterwords to filter the Kraken2 output and keep only the non hosts sequences. To so so I either use Kraken Extract Reads tool or Filter sequence by ID. These two tools works perfectly with gz files as it does an internal unzipping conversion, however when these tools are used within the workflow this internal unzipping doesnot occur. Meaning when I run the tool within the workflow it doesnot work but without the workflow it works. So I am updating the tool wrapper to do this unzipping

tools/krakentools/extract_kraken_reads.xml
bernt-matthias: This should try to decompress the dataset in the history and the test files. If these files are already, then there is nothing to do.

https://github.com/galaxyproject/galaxy/blob/0cdcc6430af894a6f218063bf449f6ef172ab4e9/lib/galaxy/tool_util/verify/__init__.py#L276

which uses

https://github.com/galaxyproject/galaxy/blob/0cdcc6430af894a6f218063bf449f6ef172ab4e9/lib/galaxy/util/compression_utils.py#L99

tools/krakentools/extract_kraken_reads.xml
```diff
@@ -31,7 +31,18 @@
             #set temp_output_1 = 'output_1.fastq'
             #set temp_output_2 = 'output_2.fastq'
         #end if
-
+        #if $library.type == 'paired':
+            find $library.input_1 -name "*.gz" -exec gunzip {} \;
```
bernt-matthias, line 35: Why calling an extra `find`?

Typically `gunzip` will unzip in the folder of the input. This is not possible here and should fail.

Maybe JUST similar to this  `gunzip -c INPUT_PATH > SOME_TMP_PATH`?

bernt-matthias: Also the params need to add the zipped formats to the format attribute.

bernt-matthias: A test with zipped input would be good to have

bernt-matthias (commented):
Allowing zipped input is a good idea. This saves galaxy from doing the implicit conversion.

But it would also be good to know why this does not work in workflows?

tools/krakentools/extract_kraken_reads.xml
```diff
@@ -13,15 +13,38 @@
 
         ##set input
         #if $library.type == 'paired':
-            #set input_1 = $library.input_1
-            #set input_2 = $library.input_2
+            #set full_ext = $library.input_1.datatype.file_ext
+            #set temp_input_1 = $library.input_1
+            #set temp_input_2 = $library.input_2
+            #if $full_ext.endswith("gz")
+                gunzip $temp_input_1
```
wm75, line 20: This cannot work because
1. you cannot write outside the tool's job working directory (but gunzip used like this will extract to the same location that Galaxy decided to store the input file in)
2. gunzip above will write to a new file, not overwrite the original one (but delete it) so $temp_input_1 would not exist anymore when the kraken tool tries to access it later.

Try to use the syntax @bernt-matthias suggested in an earlier comment.

wm75:
> 
> But it would also be good to know why this does not work in workflows?

@bernt-matthias it does work (I think) with the conversion step in the red box in the screenshot, you just need to think of adding it. Is that correct @EngyNasr?

tools/krakentools/extract_kraken_reads.xml
EngyNasr: I have tried all combinations I am getting the same error

paulzierep:
@EngyNasr Now that I think about it, the tool can handle gzipped input, therefore I think the gunzip is actually not needed, see my PR (https://github.com/EngyNasr/tools-iuc/pull/1).

> > But it would also be good to know why this does not work in workflows?
> 
> @bernt-matthias it does work (I think) with the conversion step in the red box in the screenshot, you just need to think of adding it. Is that correct @EngyNasr?

I think the main issue here is, that the tools works as expected using the tool alone or in a WF with single files (galaxy decompresses automatically the input since the input is in fasta.gz, but the tool can only handle fasta); in a collection the conversion is not done ... the global problem is, that the test with planemo are successful, but later the tool fails using a collection. Adding 'fastq.gz,fasta.gz' as input format avoids the automatic decompression and then the tool can be correctly tested with planemo also. But I think @EngyNasr  has also encountered other tools, where only the switch to collections allowed us to spot that problem. So either the automatic decompression should be part of collection behavior or at least planemo should not do automatic decompression.

bernt-matthias:
> in a collection the conversion is not done

Does the tool consume the collection or is the tool mapped over the collection (i.e. a separate job is created for each collection element)?

EngyNasr:
> in a collection the conversion is not done

a separate job is created for each element of the collection

EngyNasr:
Shall we merge it ? I really hope this can solve the workflow problem.

@wm75, I have updated the IWC upload [PR](https://github.com/galaxyproject/iwc/pull/182) as per the comments except the decompress comment. is it possible to merge the IWC PR as a first version for now and after the tool works I will update the workflow and upload a newer version to the IWC?

bernt-matthias (approved):
Looks good to me.

tools/krakentools/extract_kraken_reads.xml
```diff
@@ -13,13 +13,42 @@
...
         #if $library.type == 'paired':
-            #set input_1 = $library.input_1
-            #set input_2 = $library.input_2
+            #set full_ext = $library.input_1.datatype.file_ext
+            #set temp_input_1 = $library.input_1
+            #set temp_input_2 = $library.input_2
+
+            ## tool needs a file ending with gz to work
+            #if $full_ext.endswith("gz"):
+                ln -s $temp_input_1 'temp_input_1.gz' &&
```
wm75, line 22: You're quoting the wrong part here and everywhere below: the literal string is safe, but the variable is not.

tools/krakentools/extract_kraken_reads.xml
```diff
@@ -13,13 +13,42 @@
 
         ##set input
         #if $library.type == 'paired':
-            #set input_1 = $library.input_1
-            #set input_2 = $library.input_2
+            #set full_ext = $library.input_1.datatype.file_ext
+            #set temp_input_1 = $library.input_1
```
wm75, line 17: Maybe all those `temp_input` variables can now be reverted back to just `input` ones?

tools/krakentools/extract_kraken_reads.xml
```diff
@@ -13,13 +13,42 @@
 
         ##set input
         #if $library.type == 'paired':
-            #set input_1 = $library.input_1
-            #set input_2 = $library.input_2
+            #set full_ext = $library.input_1.datatype.file_ext
```
wm75, line 16: You can save those `full_ext` assignments and use `$library.input_1.datatype.file_ext.endswith("gz")` and the like directly below.

wm75 (approved):

Issue #5366, opened 2023-06-20
FASTQ to FASTA tool: gzipped output has fasta datatype


neoformit:
Just popped this out of https://github.com/galaxyproject/tools-iuc/issues/5333.

This is for the fastx_toolkit tool: https://github.com/galaxyproject/tools-iuc/tree/main/tool_collections/fastx_toolkit/fastq_to_fasta

**Issue:** The tool attempts to infer the FASTA output compression (none, gz or bz) from the input datatype. The tool does compress the data correctly, but always sets the output datatype as `fasta` resulting in a broken/weird dataset.

I like the suggestion from @bernt-matthias :+1: 
> My preference would be to have zipping/unzipping controlled by a select with the following options:
>
> - same as input (default)
> - zipped
> - unzipped
>
> So we have the best of both worlds. Unfortunately this brings a bit of "heavy" logic for determining the output format. Could be implemented with tool provided metadata (galaxy.json).
> 
> Otherwise my (weak) preference would be "same as input"...
> 
> What about fastqsanger.bz
> 
> Is this a thing? :)

... so will implement that.

I'm working on a solution for inferring the output datatype that uses two outputs (one `fasta`, one `fasta.gz`) with a filter on each. Something like:

```xml
        <data name="output" format="fasta" metadata_source="input">
            <filter>
                compress_output == 'no'
                or compress_output == 'infer'
                and input.is_of_type('fasta', 'fastqsanger', 'fastqsolexa', 'fastqillumina')
            </filter>
        </data>
        <data name="output" format="fasta.gz" metadata_source="input">
            <filter>
                compress_output == 'yes'
                or compress_output == 'infer'
                and input.is_of_type(
                    'fasta.gz',
                    'fastqsanger.gz',
                    'fastqsolexa.gz',
                    'fastqillumina.gz',
                    'fastqsanger.bz2',
                    'fastqsolexa.bz2',
                    'fastqillumina.bz2',
                )
            </filter>
        </data>
```

Looks somewhat hacky to me, but it's the only way I could see to make the `format` attribute dynamic (unless there is some magic to be done with metadata as Matthias suggests). At least it's fairly readable.

neoformit:
@bernt-matthias

bernt-matthias:
Might work. Worth a try. Maybe you can simplify `input.is_of_type(...)` to `input.ext.endswith('.gz') or input.ext.endswith('.bz2')`

Alternatively `galaxy.json` like here: 
[...]

PR #5367, opened 2023-06-21
Fastq to fasta fix

neoformit:
Fix `fastx_toolkit/fastq_to_fasta`. Resolves https://github.com/galaxyproject/tools-iuc/issues/5366

- Add a select input so the user can specify whether they want compressed FASTA output [ `infer` (default), `yes`, `no` ].
- Add some tests


neoformit:
@bernt-matthias

bernt-matthias:
Yes? Guess tool test and linting problems need to be fixed :)

neoformit:
Sorry should have checked those... my planemo was out of date :(

tool_collections/fastx_toolkit/fastq_to_fasta/fastq_to_fasta.xml
```diff
@@ -9,7 +9,11 @@
 $SKIPN
 $RENAMESEQ
 -v @FQQUAL@
-@GZIP@ > '$output'
+#if $compress_output == 'yes' or $compress_output == 'infer' and $input.ext.endswith('.gz') or $input.ext.endswith('.bz2'):
```
bernt-matthias, line 12: Should be fine by evaluation order, but I think it makes it more readable, but up to you

```suggestion
#if $compress_output == 'yes' or ($compress_output == 'infer' and ($input.ext.endswith('.gz') or $input.ext.endswith('.bz2'))):
```

bernt-matthias (approved):

tool_collections/fastx_toolkit/fastq_to_fasta/fastq_to_fasta.xml
```diff
@@ -49,45 +45,45 @@ $RENAMESEQ
...
+            <output name="output" file="fastq_to_fasta-out2.out" expect_num_outputs="1" />
         </test>
         <test>
             <!-- 6 Force FASTQ-To-FASTA.GZ -->
             <param name="input" value="fastq_to_fasta-in2.fastq" ftype="fastqsanger" />
             <param name="SKIPN" value="no"/>
             <param name="RENAMESEQ" value="yes"/>
             <param name="compress_output" value="yes"/>
-            <output name="output" file="fastq_to_fasta-out2.out.gz" />
+            <output name="output" file="fastq_to_fasta-out2.out.gz" expect_num_outputs="1" />
```
bgruening: is that correct? expect_num_outputs should be on the top `<test expect_num_outputs="1">` level.

bgruening (approved):

tool_collections/fastx_toolkit/fastq_to_fasta/fastq_to_fasta.xml
bernt-matthias: Nah.

bernt-matthias: I would have expected that the xsd linter raises an error...

neoformit: Yeah that was just me being half asleep on the job, think I fixed this up?

bernt-matthias:
@neoformit @bgruening just realized that this broke the other tools of the fastx_toolkit (due to the symlinked macros file). https://github.com/galaxyproject/tools-iuc/issues/5412

How to proceed?

neoformit:
I'll revert the linked macro file back to prev version and replace the linked macros.xml in fastq_to_fasta with the updated/fixed version. That at least keeps everything working for now. I haven't checked whether the faulty logic in fastq_to_fasta can also be found across the suite, and should perhaps be fixed globally?

neoformit:
Actually I just reverted to the old linked macros.xml as it's forwards compatible with my changes to fastq_to_fasta.

PR #5372, opened 2023-06-25
update cojac version to 0.9

PlushZ:
FOR CONTRIBUTOR:


wm75:
@PlushZ I've decided that it's time to rewrite all the bed /amplicon info file handling code spreading now across the ivar suite of tools and here. My plan is to create a PyPI package for this that ivar and cojac can reuse, but for now I just pasted my first version of that code here. We can always replace it later.

PlushZ (commented):
@ wm75 is it a good idea to replace this option from 2 files with only one ARTIC primer scheme file? We can avoid asking a user for an extra amplicon_info file since your viramp_hub PyPI package can generate it.

tools/cojac/cooc_mutbamscan.xml
```diff
@@ -32,8 +39,26 @@ cooc-mutbamscan
...
+                <option value="single_file">Amplicon definitions formatted for
+                    Cojac</option>
+                <option value="two_files">ARTIC network primer scheme and amplicon grouping
+                file</option>
+            </param>
+            <when value="single_file">
+                <param name="bed_file" type="data" format="bed" label="Amplicon definitions file"
+                       help="BED dataset providing the positions of the amplicons on the reference sequence. Cojac: mutbamscan expects coordinates excluding the primer sequences."/>
+            </when>
+            <when value="two_files">
```
PlushZ, line 54: @wm75 is it a good idea to replace this option from 2 files with only one ARTIC primer scheme file? We can avoid asking a user for an extra amplicon_info file since your viramp_hub PyPI package can generate it.

wm75: Yes, that's the idea, but you'd want to keep the two files option as a fallback because with exotic primer schemes the autodetection of the primer/amplicon mapping by viramp-hub might fail.

tools/cojac/cooc_mutbamscan.xml
```diff
@@ -1,96 +1,146 @@
...
-cooc-mutbamscan
-    -a '${identifier}.bam'
-    -b 'bed_file'
-    -m '$vocdir'
-    -y cooc.yaml
-    -j cooc.json
-    #if $amplicons_file.choice == 'build'
-        -A amplicons.yaml
+## prepare aligned reads inputs
+while IFS=\$(printf '\t') read -r identifier batch data_path index_path; do
```
bgruening, line 16: You are using Cheetah to create a config file and then iterate here over in bash. Wouldn't it be better to create a bash file in the config-file with cheetah, that you just need to execute here. A bash script that has `mkdir ... && ln -s ` included?

tools/cojac/cooc_mutbamscan.xml
```diff
@@ -1,96 +1,146 @@
...
+                    <param name="choice" type="select" label="Format of amplicon definitions">
+                        <option value="insert_scheme">Amplicon insert scheme (Cojac-native format)</option>
+                        <option value="primer_scheme">Primer scheme (with autodetection of grouping into amplicons</option>
+                        <option value="primer_scheme_plus">Primer scheme and explicit amplicon grouping information</option>
+                    </param>
+                    <when value="insert_scheme">
+                        <param name="insert_bed" type="data" format="bed" label="Amplicon insert scheme" help="BED dataset providing the positions of the amplicon inserts (i.e. excluding the primer sequences) on the reference sequence."/>
+                    </when>
+	                <when value="primer_scheme">
+		                <param name="primer_bed" type="data" format="bed" label= "Primer scheme" help="The tool will attempt to autodetect the relationship between primers and amplicons based on the primer names."/>
```
bgruening, line 82:
```suggestion
		                <param name="primer_bed" type="data" format="bed" label="Primer scheme" help="The tool will attempt to autodetect the relationship between primers and amplicons based on the primer names."/>
```

tools/cojac/cooc_mutbamscan.xml
```diff
@@ -116,18 +166,75 @@ cooc-mutbamscan
...
+            <param name="mapped_input" value="tbam11.bam"/>
+            <conditional name="spec_layout">
+                <param name="mode" value="combined"/>
                 <param name="in_amp" value="amplicons111.yaml"/>
             </conditional>
-            <param name="output_files" value="yaml,tabular"/>
+            <section name="out_opts">
+                <param name="output_formats" value="yaml,tabular"/>
+            </section>
+            <output name="cooc_yaml" ftype="yaml">
```
bgruening, line 179: Wondering if we could use those JSON tests here: https://docs.galaxyproject.org/en/latest/dev/schema.html#id177

tools/cojac/cooc_mutbamscan.xml
wm75: Yes, I could, but would two configfiles really be simpler than one and a bit of bash magic?

tools/cojac/cooc_mutbamscan.xml
wm75: Hmm, looks interesting, but the json produced by the tool is nested like this:
`{"tbam11.bam": {"76_om1": {"sites": {"1": 68, "2": 4}, "muts": {}}, "81_om1": {"sites": {"2": 211}, "muts": {"1": 209}}}}`
and it doesn't read like the search for a property will go into nested structures, which would mean we'd have to use the entire outer value in the test, which then is very similar to a simple `has_line`.

tools/cojac/cooc_mutbamscan.xml
bgruening: For admins its also easier to read config files than bash magic ;)

wm75: Ok, looks more readable with two configfiles :+1:

bgruening (approved):

PR #5376, opened 2023-06-28
fixed no bins in output bug

Ralwoss:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/das_tool/das_tool.xml
+++ b/tools/das_tool/das_tool.xml
@@ -83,9 +83,9 @@ DAS_Tool
         </data>
         <collection name="bins" type="list" label="${tool.name} on ${on_string}: Bins">
             <filter>output['write_bins']['write_bins'] != ""</filter>
-            <discover_datasets pattern="^(?!unbinned\.fa)((?P&lt;designation&gt;.*)\.fa)" format="fasta" directory="outputs_DASTool_bins" />
+            <discover_datasets pattern="^(?!unbinned\.fa)((?P&lt;designation&gt;.*)\.fa)" format="fasta" directory="outputs_DASTool_bins"/>
           </collection>
-        <data name="unbinned_contigs" format="fasta" from_work_dir="outputs_DASTool_bins/unbinned.fa" label="${tool.name} on ${on_string}: Unbinned contigs" >
+        <data name="unbinned_contigs" format="fasta" from_work_dir="outputs_DASTool_bins/unbinned.fa" label="${tool.name} on ${on_string}: Unbinned contigs">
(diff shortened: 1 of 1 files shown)


tools/das_tool/das_tool.xml
```diff
@@ -83,7 +83,7 @@ DAS_Tool
         </data>
         <collection name="bins" type="list" label="${tool.name} on ${on_string}: Bins">
             <filter>output['write_bins']['write_bins'] != ""</filter>
-            <discover_datasets pattern=".*?\.(?P&lt;designation&gt;.*)\.fa" format="fasta" directory="outputs_DASTool_bins" />
```
bernt-matthias, line 86: Do you know the motivation for the old Regex? Or how those filenames are constructed? 

Maybe the idea was to allow for some prefix ending with a dot... 

Closer to the original idea might be 
`(?:.*\.)?(?P<designation>.*)\.fa`

However, we definitely should have a test case covering this. Seems that this is missing.

Ralwoss: Yes, thank you. The file names come from the input file names, so there are cases with prefixes. It is fixed

tools/das_tool/das_tool.xml
```diff
@@ -83,7 +83,7 @@ DAS_Tool
         </data>
         <collection name="bins" type="list" label="${tool.name} on ${on_string}: Bins">
             <filter>output['write_bins']['write_bins'] != ""</filter>
-            <discover_datasets pattern="(?P&lt;designation&gt;.*)\.fa" format="fasta" directory="outputs_DASTool_bins" />
+            <discover_datasets pattern="^(?!unbinned\.fa)((?:.*\.)?(?P&lt;designation&gt;.*)\.fa)" format="fasta" directory="outputs_DASTool_bins" />
```
bernt-matthias, line 86: Is the non-capturing group:  `(?:.*\.)?` really needed.

For instance if there are two files, `a.x.fa` and `b.x.fa` they would both yield the same designation, i.e. `x`. I guess this should be done only if such a conflict can not occur. Can we exclude this possibility?

Otherwise, maybe just

```suggestion
            <discover_datasets pattern="^(?!unbinned\.fa)((?P&lt;designation&gt;.*)\.fa)" format="fasta" directory="outputs_DASTool_bins" />
```

Ralwoss: Yes, I removed the prefix filtering now. This way it is less likely to have conflicting designations

bebatut (approved):
Thanks @Ralwoss for the fix

Issue #5377, opened 2023-06-29
DADA2 creates faulty output with `fastqsanger` input


neoformit:
A user reported an issue with `dada2_filterAndTrim` - this is another case where the tool infers output datatype based on the input and this now seems to fail (could it be a 23.0 bug?).

- Input: `fastqsanger`
- Output: `fastqsanger` datatype assigned to gzipped `fastqsanger` data

I created an additional test which catches the above issue (not published, and requires existing test data to be gunzipped):

```xml
<!-- paired data in paired collection with uncompressed input -->
<test expect_num_outputs="4">
    <conditional name="paired_cond">
        <param name="paired_select" value="paired"/>
        <param name="reads">
            <collection type="paired">
                <element name="forward" value="F3D0_S188_L001_R1_001.fastq" ftype="fastqsanger"/>
                <element name="reverse" value="F3D0_S188_L001_R2_001.fastq" ftype="fastqsanger"/>
            </collection>
        </param>
    </conditional>
    <output_collection name="paired_output" type="paired" count="2">
        <element name="forward" value="filterAndTrim_F3D0_R1.fq" ftype="fastqsanger" />
        <element name="reverse" value="filterAndTrim_F3D0_R2.fq" ftype="fastqsanger" />
    </output_collection>
    <output name="outtab" value="F3D0_S188_L001_R1_001.tab" ftype="tabular" lines_diff="2">
        <assert_contents>
            <has_text text="745" />
            <has_text text="750" />
        </assert_contents>
    </output>
</test>
```

Fails with:

```
Collection element forward of collection paired_output:  different than expected, difference (using diff):
( /tmp/tmpml2jfb_6filterAndTrim_F3D0_R1.fq v. /tmp/tmphk8tihkzfilterAndTrim_F3D0_R1.fq )
Binary data detected, not displaying diff
```

(Should be text fastq, not binary)

bernt-matthias:
Fixed by https://github.com/galaxyproject/tools-iuc/pull/5378

PR #5378, opened 2023-06-29
Dada2 - fix for uncompressed inputs

neoformit:

Resolves https://github.com/galaxyproject/tools-iuc/issues/5377

- Add test to cover uncompressed `fastqsanger` input
- Compression determined by input datatype


bernt-matthias (commented):
Good catch. 

FYI: My original intention was to always have compressed output (because smaller .. but I guess this is not that important for amplicon data). 

Please bump the tool version.

neoformit:
I guess we can keep that logic and remove the `format_source="reads"` from outputs?

tools/dada2/dada2_dada.xml
```diff
@@ -98,11 +98,11 @@ dada_result <- dada(derep, err,
         <param name="selfconsist" type="boolean" checked="false" truevalue="TRUE" falsevalue="FALSE" label="Alternate between sample inference and error rate estimation until convergence"/>-->
     </inputs>
     <outputs>
-        <data name="dada" format="dada2_dada">
+        <data name="dada" format="dada2_dada" label="${tool.name} on ${on_string}">
             <filter>batch_cond['batch_select']=="yes"</filter>
         </data>
-        <collection name="data_collection" type="list">
-            <discover_datasets pattern="(?P&lt;name&gt;.+)\.dada2_dada" format="dada2_dada" directory="output" />
+        <collection name="data_collection" type="list" label="${tool.name} on ${on_string}: outputs">
```
bernt-matthias, line 104: Was this a requirement of the linter? Wondering if it should be, since the outputs are mutually exclusive and the same name should be fine (but it would be hard/impossibe for the linter to find this out .. and just allowing arbitrary names for outputs with filters might be to loose).

bernt-matthias:
> I guess we can keep that logic and replace format_source="reads" with format="fastqsanger.gz" in the outputs?

I leave this up to you.

tools/dada2/dada2_dada.xml
neoformit: Yep was linted out. I agree - this shouldn't fail linting but the logic would be too complex. It's not so hard to make different labels here, but in another case this might be really annoying (i.e. 10 mutually exclusive filtered outputs).

neoformit:
> I leave this up to you.

Ok let's leave as in then, `fastqsanger` -> `fastqsanger`.

bernt-matthias (approved):

bernt-matthias:
Good to go, or?

neoformit:
Yep, send it

PR #5386, opened 2023-07-05
Upload lineagespot wrappers

inabbioinformatics:
FOR CONTRIBUTOR:


tools/lineagespot/lineagespot_wrapper.xml
```diff
@@ -0,0 +1,83 @@
...
+    </inputs>
+    <outputs>
+        <data name="lineage_hits" format="txt" />
+        <data name="lineage_report" format="txt" />
+        <data name="variants_table" format="txt" />
+    </outputs>
+    <tests>
+        <test expect_num_outputs="3">
+            <param name="in_vcf" value="${__tool_directory__}/test-data/extdata/vcf-files"/>
+            <param name="in_gff3" value="${__tool_directory__}/test-data/extdata/NC_045512.2_annot.gff3"/>
```
PlushZ, line 41: In tests instead of using the absolute path value '${__tool_directory__}/test-data/extdata/NC_045512.2_annot.gff3', we should use the name of the file that is in the test-data folder as input value, so here is 'NC_045512.2_annot.gff3'

tools/lineagespot/lineagespot_wrapper.xml
```diff
@@ -0,0 +1,83 @@
...
+        <param type="float" name="in_threshold" value="0.8" label="Kmer size to count"/>
+    </inputs>
+    <outputs>
+        <data name="lineage_hits" format="txt" />
+        <data name="lineage_report" format="txt" />
+        <data name="variants_table" format="txt" />
+    </outputs>
+    <tests>
+        <test expect_num_outputs="3">
+            <param name="in_vcf" value="${__tool_directory__}/test-data/extdata/vcf-files"/>
```
PlushZ, line 40: in tests, we should provide the specific file name, we cannot provide a directory.

tools/lineagespot/lineagespot_wrapper.xml
```diff
@@ -0,0 +1,83 @@
...
+        <requirement type="package" version="1.14.6">data.table</requirement>
+        <requirement type="package" version="1.5.0">stringr</requirement>
+        <requirement type="package" version="1.4.4">httr</requirement>
+        <requirement type="package" version="4.2.0">utils</requirement>
+        <requirement type="package" version="2.26.0">BiocStyle</requirement>
+        <requirement type="package" version="1.4.0">RefManageR</requirement>
+        <requirement type="package" version="2.20">rmarkdown</requirement>
+        <requirement type="package" version="1.42">knitr</requirement>
+        <requirement type="package" version="3.1.6">testthat</requirement>
+        <requirement type="package" version="1.2.0">lineagespot</requirement>
```
PlushZ, line 18: in requirements, we should mention the name of a package in conda because Galaxy uses conda to install them, instead of 'lineagespot' use 'bioconductor-lineagespot'. This is applicable to all requirements

tools/lineagespot/lineagespot_wrapper.xml
```diff
@@ -0,0 +1,83 @@
+<tool id="lineagespot" name="lineagespot" version="1.2.0">
+    <description> Identifies SARS-CoV-2 related mutations based on a single (or a list) of variant(s) file(s).</description>
+    <requirements>
+        <requirement type="package" version="4.2.0">R</requirement>
+        <requirement type="package" version="1.44.0">VariantAnnotation</requirement>
+        <requirement type="package" version="1.20.3">getopt</requirement>
```
PlushZ, line 6: Galaxy uses conda to install packages. For example, in conda, the package name is 'r-getopt' instead of 'getopt'.

tools/lineagespot/lineagespot_wrapper.xml
```diff
@@ -0,0 +1,83 @@
...
+        <requirement type="package" version="1.5.0">stringr</requirement>
+        <requirement type="package" version="1.4.4">httr</requirement>
+        <requirement type="package" version="4.2.0">utils</requirement>
+        <requirement type="package" version="2.26.0">BiocStyle</requirement>
+        <requirement type="package" version="1.4.0">RefManageR</requirement>
+        <requirement type="package" version="2.20">rmarkdown</requirement>
+        <requirement type="package" version="1.42">knitr</requirement>
+        <requirement type="package" version="3.1.6">testthat</requirement>
+        <requirement type="package" version="1.2.0">lineagespot</requirement>
+    </requirements>
```
PlushZ, line 19: When specifying the requirements, it is recommended to include only the essential dependencies. If a package you depend on already includes another package as a requirement, there is no need to list it again in the requirements section.

tools/lineagespot/lineagespot_wrapper.xml
PlushZ: This applies to other files in your tests

tools/lineagespot/lineagespot_wrapper.xml
inabbioinformatics: lineagespot is designed so that parameter in_vcf requires a whole repository of vcf files

tools/lineagespot/lineagespot_wrapper.xml
inabbioinformatics: in a previous comment you mentioned that if in requirements a package exists in bioconda-recipe we can remove it..So now in requirements are only those that are not included in bioconductor

tools/lineagespot/lineagespot_wrapper.xml
bgruening: @inabbioinformatics Galaxy works with Object Stores, there is no concept of a folder in Object Stores. If the tool requires a folder you need to create it in your command section, e.g.

```
ln -s '$file1' ./foo/file.vcf &&
ln -s '$file2' ./foo/file.vcf &&
```

And the tests are then single files, e.g.

`<param name="in_vcf" value="vcf-file1.vcf,vcf-file2.vcf,vcf-file3.vcf"/>`

inabbioinformatics: Thank you for the comment. So just to clarify, is it right to create in command section the vcf files that are prerequisites with command "ln -s"                                                                                                           e.g ln -s 'SampleA_freebayes_ann.vcf' ./test-data/extdata/vcf-files/SampleA_freebayes_ann.vcf &&
        ln -s 'SampleB_freebayes_ann.vcf' ./test-data/extdata/vcf-files/SampleB_freebayes_ann.vcf &&
        ln -s 'SampleC_freebayes_ann.vcf' ./test-data/extdata/vcf-files/SampleC_freebayes_ann.vcf &&                                                                                                                
        and afterwards could we use multiple vcf files in the same parameter at once ?                                                                                                                e.g  <param name="in_vcf" value="SampleA_freebayes_ann.vcf,SampleB_freebayes_ann.vcf,SampleC_freebayes_ann.vcf"/>

tools/lineagespot/lineagespot_wrapper.xml
```diff
@@ -0,0 +1,92 @@
...
+        ln -s 'B.1.1.7.txt' ./test-data/extdata/ref/B.1.1.7.txt &&
+        ln -s 'B.1.1.7+S_E484K.txt' ./test-data/extdata/ref/B.1.1.7+S_E484K.txt &&
+        ln -s 'B.1.351.2.txt' ./test-data/extdata/ref/B.1.351.2.txt &&
+        ln -s 'B.1.351.3.txt' ./test-data/extdata/ref/B.1.351.3.txt &&
+        ln -s 'B.1.351.txt' ./test-data/extdata/ref/B.1.351.txt &&
+        ln -s 'B.1.617.2.txt' ./test-data/extdata/ref/B.1.617.2.txt &&
+        ln -s 'P.1.1.txt' ./test-data/extdata/ref/P.1.1.txt &&
+        ln -s 'P.1.2.txt' ./test-data/extdata/ref/P.1.2.txt &&
+        ln -s 'P.1.txt' ./test-data/extdata/ref/P.1.txt &&
+        Rscript '${__tool_directory__}/lineagespot_verbose.R --in_vcf ${__tool_directory__}/test-data/extdata/vcf-files --in_gff3 ${__tool_directory__}/test-data/extdata/NC_045512.2_annot.gff3 --in_ref ${__tool_directory__}/test-data/extdata/ref --in_voc "B.1.617.2, B.1.1.7, B.1.351, P.1" --in_threshold 0.8'
```
wm75, line 33: So just to clarify: lineagespot expects a *folder* (with one or several vcfs inside) passed through the `--in_vcf` option, correct?

Then I guess what @bgruening meant is this:
1. the first line in your `<inputs>` section lets the user select a vcf dataset from their history.
2. the path to this user-selected dataset will, in the `<command>` section, be accessible via `$in_vcf` (the name given to the param prefixed with a $ sign)
3. Galaxy will set some working directory for the `<command>` section to be run in and will also remove that entire directory again when the job is done (we call it the job working directory, you don't have to handle it in any way)
4. You need to create a directory right inside the job working directory (since it's the current working directoy a simple `mkdir whateverdir` will do
5. You then symlink the user-selected `$in_vcf` into that new directory
6. and finally your call the lineagespot script with `--in_vcf whateverdir` to have it discover the symlinked file

There's no need to use the test-data anywhere in the command section, it's for the `<tests>` section only. The `<command>` section references the params defined in `<inputs>`, which are set by the user from the Galaxy UI.

Another minor technical issue is that you're currently quoting the entire line after `Rscript`, which would pass everything as a single parameter.

tools/lineagespot/lineagespot_wrapper.xml
```diff
@@ -0,0 +1,92 @@
...
+    ]]></command>
+    <inputs>
+        <param type="data" name="in_vcf" format="vcf" label="vcf file(s)" />
```
wm75, line 36: To allow not just one input vcf dataset, but multiple ones, you'd use the `multiple="true"` attribute.
In the `<command>` section you can then use a cheetah (python-like) for loop to iteratively generate symlinks for all of them, similar to what's explained here: https://docs.galaxyproject.org/en/latest/dev/schema.html#id43

tools/lineagespot/lineagespot_wrapper.xml
```diff
@@ -0,0 +1,83 @@
+<tool id="lineagespot" name="lineagespot" version="1.2.0">
```
PlushZ, line 1: It is recommended to use variables @TOOL_VERSION@ and @VERSION_SUFFIX@ for version instead of hardcoding the version. So here is better to write version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" and add short macros in the wrapper:
```
    <macros>
        <token name="@TOOL_VERSION@">1.2.0</token>
    </macros>
```

tools/lineagespot/lineagespot_wrapper.xml
PlushZ: Awesome, also include 'bioconductor-lineagespot' requirement, with this requirement galaxy will install lineagespot. Use variable for tool version in the wrapper. Something like:
`<requirement type="package" version="@TOOL_VERSION@">bioconductor-lineagespot</requirement>`

tools/lineagespot/lineagespot_verbose.R
```diff
@@ -0,0 +1,80 @@
+## How to execute this tool
+
+# A command-line interface to lineagespot for use with Galaxy
+#
+# The following arguments are required:
+#
+#   'in_vcf'  a character vector of paths to VCF files object from Galaxy lineagespot/test-data 
+#   'in_gff3' a character vector of path to GFF3 file containing SARS-CoV-2 gene coordinates object from Galaxy lineagespot/test-data 
+#   'in_ref'  a character vector of path to a folder containing lineage reports object from Galaxy lineagespot/test-data 
+#   'in_voc'  a character vector containing the names of the lineages of interest
```
wm75, line 10: Question: how does lineagespot know about these names?
Are they taken from the *ref filenames*?

wm75: If so, that would be a major problem cause Galaxy doesn't know about these names since, internally, it stores uploaded datasets under totally unrelated identifiers.

inabbioinformatics: @wm75  Indeed, Lineagespot gathers information about these names by extracting them from the input filenames, such as vcf files and ref files. For instance, if the filename is 'sample_956265.vcf,' Lineagespot will identify and include the name 'sample_956265' in the generated output tables.

wm75 (approved):

wm75:
planemo lint is failing with: `Error '404 Client Error: Not Found for url: https://github.com/galaxyproject/tools-iuc/blob/main/tools/lineagespot/convert_lineage_defs.py'`.
A chicken and egg problem :)

tools/lineagespot/convert_lineage_defs.py
```diff
@@ -0,0 +1,163 @@
...
+if len(sys.argv) < 2:
+    sys.exit('Please run with -h / --help for help.')
+
+args = parser.parse_args()
+
+for definitions_file in os.listdir(args.input):
+    # In constellations, the only reliable way to get the lineage name is from
+    # the file name by stripping the .json suffix from it and dropping the
+    # leading 'c' (e.g. cBA.5.json holds the definition for lineage BA.5).
+    lineage_name = definitions_file[1:-5]
```
bgruening, line 149: would it be safer to strip the `c` and the `.json` instead of relying on the size of the string?

tools/lineagespot/lineagespot_wrapper.xml
```diff
@@ -0,0 +1,220 @@
...
+        <token name="@TOOL_VERSION@">1.4.0</token>
+        <token name="@VERSION_SUFFIX@">0</token>
+    </macros>
+    <requirements>
+        <requirement type="package" version="4.3.1">r-base</requirement>
+        <requirement type="package" version="1.20.3">r-getopt</requirement>
+        <requirement type="package" version="@TOOL_VERSION@">bioconductor-lineagespot</requirement>
+        <requirement type="package" version="3.11">python</requirement>
+    </requirements>
+    <stdio>
```
bgruening, line 13: remove and use detect_errors instead?

tools/lineagespot/lineagespot_wrapper.xml
```diff
@@ -0,0 +1,220 @@
+<tool id="lineagespot" name="lineagespot" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@">
+    <description>identifies SARS-CoV-2 lineages contributing to metagenomic samples from per-sample variant files</description>
+    <macros>
+        <token name="@TOOL_VERSION@">1.4.0</token>
+        <token name="@VERSION_SUFFIX@">0</token>
+    </macros>
+    <requirements>
+        <requirement type="package" version="4.3.1">r-base</requirement>
```
bgruening, line 8: r-base can be remove

tools/lineagespot/lineagespot_wrapper.xml
```diff
@@ -0,0 +1,220 @@
+<tool id="lineagespot" name="lineagespot" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@">
```
bgruening, line 1: add a profile

tools/lineagespot/lineagespot_wrapper.xml
```diff
@@ -0,0 +1,220 @@
...
+        <param type="data" name="in_vcf" format="vcf" multiple="true" label="Per-sample variant calling data" />
+        <conditional name="voc_source">
+            <param name="choice" type="select"
+                   label="Source of lineage definitions"
+                   help="Lineagespot detects lineage evidence based on lineage mutation patterns defined in tool-specific definition files. The Galaxy tool also has experimental support for reading lineage definitions from cached constellations data (see tool help below).">
+                <option value="custom">Collection of lineagespot-specific definitions in history</option>
+                <option value="cached">Lineage definitions from cached constellations data</option>
+            </param>
+            <when value="custom">
+                <param name="collection" type="data_collection" collection_type="list" format="tabular" label="Collection of lineage definitions" />
```
bgruening, line 69: Is there any reason why we can not use type=data with multiple=true?

tools/lineagespot/lineagespot_wrapper.xml
```diff
@@ -0,0 +1,220 @@
...
+
+*Genome feature annotations*
+
+To learn about the position of the genes mentioned in VCF ANN fields and the lineage definitions on the genome, the tool requires an additional genome feature annotation input in gff3 format.
+You should normally just use the offered built-in genome annotation file.
+A custom annotation file would only be required if you've mapped your sequencing data to a different reference genome than NC_045512.2 that isn't positionally identical to it.
+    ]]></help>
+    <citations>
+        <citation type="doi">10.1038/s41598-022-06625-6</citation>
+    </citations>
```
bgruening, line 219: if you like you can add here author's information https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-creator-person

tools/lineagespot/lineagespot_wrapper.xml
wm75: I thought I'd play it safe to avoid future issues with dependency versions by pinning r-base to a specific version.
But you'd prefer this to be removed and we just trust conda?

tools/lineagespot/lineagespot_wrapper.xml
wm75: User experience! tabular is just a very generic datatype and non-expert users might be overwhelmed by the number of choices from their history. Requiring a *collection* of tabular datasets helps (we can, for example, offer a public history with such collections, that users can just copy over to start an analysis) and also captures the concept better: you have a set of lineage definitions, or you don't, but you're not just using some arbitrary datasets from your history as lineage definitions.

tools/lineagespot/convert_lineage_defs.py
wm75: the script now ignores files that don't start and end in the expected strings.

tools/lineagespot/lineagespot_wrapper.xml
wm75: added xrefs for now :)

PR #5389, opened 2023-07-10
Remove VCF as viable format

JasperO98:
Gene.iobio does not accept `vcf` as it has to find the `tbi` file which is only generated from the `vcf_bgzip` format.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/geneiobio/geneiobio-iframe.xml
+++ b/tools/geneiobio/geneiobio-iframe.xml
@@ -1,4 +1,4 @@
-<tool id="gene_iobio_display_generation_iframe" name="gene.iobio visualisation" version="4.7.1"+galaxy1>
+<tool id="gene_iobio_display_generation_iframe" name="gene.iobio visualisation" version="4.7.1+galaxy1">
 	<description>analyses VCFs for single and trio analysis to find causative variants using gene.iobio's public server</description>
 	<command>cat '$index' | tr -d '\n' > '$outfile'</command>
 	<configfiles>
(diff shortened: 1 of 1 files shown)


bernt-matthias:
Could we generate the needed file on the fly in the tool?

JasperO98:
Can we still retrieve a URL from that, if so then I can add it.

mvdbeek:
> Could we generate the needed file on the fly in the tool?

No, we can't change the input of a tool, which is what would be required for this to work. This would work if this was a display application, but I'm not sure that is better than the tool here.

mvdbeek (approved):
Looks good to me, but can you bump up the tool version on line 1 ? Not sure why this is version 4.7.1, but if that reflects something about iobio the new version should be `4.7.1+galaxy1`, otherwise you can pick any version higher than 4.7.1

JasperO98:
> Looks good to me, but can you bump up the tool version on line 1 ? Not sure why this is version 4.7.1, but if that reflects something about iobio the new version should be `4.7.1+galaxy1`, otherwise you can pick any version higher than 4.7.1

I've corrected the version to `4.7.1+galaxy1`. 4.7.1 reflects the version of gene.iobio when we implemented this tool.

tools/geneiobio/geneiobio-iframe.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="gene_iobio_display_generation_iframe" name="gene.iobio visualisation" version="4.7.1">
+<tool id="gene_iobio_display_generation_iframe" name="gene.iobio visualisation" version="4.7.1"+galaxy1>
```
mvdbeek, line 1:
```suggestion
<tool id="gene_iobio_display_generation_iframe" name="gene.iobio visualisation" version="4.7.1+galaxy1">
```

PR #5410, opened 2023-07-25
RNA star: Added custom junction type

tuncK:
In the current setting, STAR RNA errors out if a gtf file is provided without any exons. Some organisms, such as SC2, can have unusual transcription that can split reads across different functional units. I added a button to make provision of the name of the genetic elements to use as junctions possible.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/rgrnastar/macros.xml
+++ b/tools/rgrnastar/macros.xml
@@ -5,7 +5,7 @@
     the index versions in sync, but you should manually update @IDX_VERSION_SUFFIX@ -->
     <!-- STAR version to be used -->
     <token name="@TOOL_VERSION@">2.7.10b</token>
-    <token name="@VERSION_SUFFIX@">3</token>
+    <token name="@VERSION_SUFFIX@">4</token>
     <token name="@PROFILE@">21.01</token>
     <!-- STAR index version compatible with this version of STAR
     This is the STAR version that introduced the index structure expected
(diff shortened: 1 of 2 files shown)


tuncK:
@wm75 This is my proposed fix to the GTF issue on STAR.

bernt-matthias:
Ping @wm75

tools/rgrnastar/macros.xml
```diff
@@ -122,6 +123,12 @@
             --runThreadN \${GALAXY_SLOTS:-4}
             ## in bytes
             --limitGenomeGenerateRAM \$((\${GALAXY_MEMORY_MB:-31000} * 1000000))
+
+            #if $varExists('refGenomeSource.GTFconditional.sjdbGTFfeatureExon'):
```
lldelisle, line 127: For me it makes more sense to put the three SJDBOPTIONS at the same place. Either all outside the if/else with `$varExists` or all inside.

tools/rgrnastar/macros.xml
```diff
@@ -142,6 +149,12 @@
             #end if
         #end if
     #end if
+
+    #if $varExists('refGenomeSource.GTFconditional.sjdbGTFfeatureExon'):
```
lldelisle, line 153: same here

lldelisle (commented):
Thanks for this new option. Do you think this should be included in the data_manager as well?

tuncK:
> Thanks for this new option. Do you think this should be included in the data_manager as well?
We could, but I suggest we dont. I bet that not many people will use this non-exon junction option anyways, apart from when one needs to try out viruses or bacteria with this tool for method development studies.

If we include, then we also need to keep track of whether the cached DB with annotation was prepared with exons, transcripts or ORFs on a separate DM column. We could introduce one more data table format with that, but as if it is not worth having 3 table versions floating around and causing extra system admin work.

tools/rgrnastar/macros.xml
tuncK: Done

tools/rgrnastar/macros.xml
tuncK: Sorry, no. I am OoO today, will work on the 3rd issue (another test) and let you know.

tools/rgrnastar/rg_rnaStar.xml
```diff
@@ -628,6 +628,39 @@ with Cufflinks if your sequences come from an unstranded library preparation.">
                 <metadata name="column_names" value="GeneID,Counts_unstrand,Counts_firstStrand,Counts_secondStrand" />
             </output>
         </test>
+        <!-- test splices sites at other features other than exons -->
```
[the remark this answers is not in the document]
tuncK: Fair point. To avoid increasing the total number of tests & files too much, I added modified gtf's. This results in a "no exon present" fatal error in the test, unless this new field is correctly functions such that "fakexons" are used as the joining units instead of "exons". As "fakexons" coordinates are the same as human exons, the same output files can be referred.

Hope this addresses the shortcoming you pointed out. Ready for your review.

lldelisle (approved):
Perfect, thanks

bgruening (approved):
Thanks everyone. Great teamwork!

PR #5411, opened 2023-07-25
bwa-mem2: expose the `-K` parameter for batch size

abueg:
FOR CONTRIBUTOR:

hello! 👋🏼 

this PR exposes the `-K` parameter in bwa-mem2, which is described in the help as: `-K INT        process INT input bases in each batch regardless of nThreads (for reproducibility) []`

normally, the program just sets this parameter scaled to the number of threads given, but we've been having some memory issues on our local infrastructure where we want to still give the program a good amount of cores but not have the memory necessarily scale up, too, so this lets us set it lower

### DIFF
--- a/tools/bwa_mem2/bwa-mem2.xml
+++ b/tools/bwa_mem2/bwa-mem2.xml
@@ -64,7 +64,6 @@ bwa-mem2 mem
     #if str( $analysis_type.io_options.io_options_selector ) == "set":
         -T '${analysis_type.io_options.T}'
         -h '${analysis_type.io_options.h}'
-        -K '${analysis_type.io_options.K}'
         ${analysis_type.io_options.a}
         ${analysis_type.io_options.C}
         ${analysis_type.io_options.V}
@@ -72,6 +71,9 @@ bwa-mem2 mem
         ${analysis_type.io_options.M}
         ${analysis_type.io_options.five}
         ${analysis_type.io_options.q}
+        #if str( $analysis_type.io_options.batch_size_options.batch_size_selector ) == "set_batch_size":
+            -K '${analysis_type.io_options.batch_size_options.K}'
+        #end if
     #end if
 
 #end if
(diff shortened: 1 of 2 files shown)


bgruening:
Unfortunately, the new planemo lint finds a few errors. Do you have time to work on this or should I have a look?

tools/bwa_mem2/bwa-mem2.xml
```diff
@@ -252,6 +255,18 @@ bwa-mem2 mem
                         <param name="V" type="boolean" truevalue="-V" falsevalue="" label="Output the reference FASTA header in the XR tag" help="-C"/>
                         <param name="Y" type="boolean" truevalue="-Y" falsevalue="" label="Use soft clipping for supplementary alignments" help="-Y; By default, BWA-MEM uses soft clipping for the primary alignment and hard clipping for supplementary alignments" />
                         <param name="M" type="boolean" truevalue="-M" falsevalue="" label="Mark shorter split hits of a chimeric alignment in the FLAG field as 'secondary alignment' instead of 'supplementary alignment'" help="-M; For Picard&lt;1.96 compatibility" />
+                        <conditional name="batch_size_options">
```
bernt-matthias, line 258: The whole conditional might be replaced by an optional integer parameter, or?

bernt-matthias (commented):
Thanks for spending time on this.

Could you also take care of the linter warnings (which make the linter fail)?

mvdbeek:
> Do you have time to work on this or should I have a look?

I think we as the IUC should fix warnings resulting from us being more strict. And we should not block deployment on linting errors (or really anything ... if it's approved and merged it should be uploaded IMHO)

tools/bwa_mem2/bwa-mem2.xml
abueg: oh, I had initially tried it with an integer parameter but not as optional so it wouldn't let me run the tool without putting something in the box (when the default behavior is that the flag doesn't have a defined integer), let me try it with an optional parameter now!

bernt-matthias: +1 You just need to be careful with the if statement if you allow 0 as a value. Guess you do not want this and have a min attribute, or?

abueg: finished making it an optional integer instead of conditional 👍🏼 i tried putting in "0" as a value and it ran fine with default behavior, so i think it is OK without a minimum?

abueg:
i tried looking at the `Lint tool-list` logs to see what was happening, and i see L402 says it failed linting, but i don't see an error message, just warnings about preferred syntax. is that what needs to be fixed?

mvdbeek:
> just warnings about preferred syntax. is that what needs to be fixed?

yes ... but I would call this part optional. I don't think it's reasonable to completely refactor the tool when all you want to do is expose a new parameter. I think we can merge your changes without applying all the lint fixes.

@bernt-matthias can we put the fail level back to error and fix the warnings separately ? this is a lot of noise for contributors.

bernt-matthias:
>  i don't see an error message, just warnings about preferred syntax. is that what needs to be fixed?

Indeed. Seems to be conditionals that use a boolean instead of a select and tests needing `expect_num_outputs`. 

As @mvdbeek said. Totally up to you. 

> Can we put the fail level back to error?

My preferred choice would be to disable linting during merging.

> and fix the warnings separately ?

Experience shows that this is unlikely to happen. Considering the CI tool tests that fail for months and IUC members don't find the time to fix. But maybe I'm wrong.

So my ideal solution is to ask contributors if they could find the time to fix. But I'm certainly not expecting that they do.

> this is a lot of noise for contributors.

True.

abueg:
if it is OK, my preference would be for me not to refactor the tool, i don't have much bandwidth for it now plus i would be afraid of messing something up refactoring the conditionals 😅

bernt-matthias (approved):

mvdbeek:
> Considering the CI tool tests that fail for months and IUC members don't find the time to fix.

I think that's a different discussion. Most of the tests I see failing are failing because an online resource changed or went away or some other minor thing that would be nice to fix, but isn't critical. And then there are tests that are failing because our testing is getting more strict over time ... but this, as the linting, is our choice as IUC maintainers.

Yes, noone has dedicated time to fix tests for things they're not using, but if the list gets too big we can surely organize an event to trim down the list. The same is true for linting, or missing containers and so on.

mvdbeek:
@abueg last thing we need is a version bump. Can you change https://github.com/galaxyproject/tools-iuc/blob/main/tools/bwa_mem2/macros.xml#L5 to 1 instead of 0 ?

abueg:
> last thing we need is a version bump. Can you change https://github.com/galaxyproject/tools-iuc/blob/main/tools/bwa_mem2/macros.xml#L5 to 1 instead of 0 ?

oh yes, my bad -- just pushed one with version bump!

mvdbeek:
I've uploaded the tool manually

bernt-matthias:
Thanks @abueg and @mvdbeek

mvdbeek:
> Attention: deployment skipped!


that's to be expected, i've deployed manually

abueg:
thank you, everyone!! 🙏🏼 🎊

PR #5413, opened 2023-07-27
multigsea: conda requirements are lower case

bernt-matthias:
Otherwise planemo monitor does not build container.  This worked in the tool tests because of a bug in Galaxy https://github.com/galaxyproject/galaxy/issues/16435

FOR CONTRIBUTOR:

### DIFF
--- a/tools/multigsea/macros.xml
+++ b/tools/multigsea/macros.xml
@@ -1,23 +1,23 @@
 <macros>
     <token name="@TOOL_VERSION@">1.8.0</token>
-    <token name="@SUFFIX_VERSION@">0</token>
+    <token name="@SUFFIX_VERSION@">1</token>
     <token name="@PROFILE@">20.05</token>
     <xml name="requirements">
         <requirements>
             <requirement type="package" version="@TOOL_VERSION@">bioconductor-multigsea</requirement>
             <requirement type="package" version="2.2.1">r-argparse</requirement>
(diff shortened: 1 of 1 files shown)


bernt-matthias:
Problem is here: 

https://github.com/galaxyproject/planemo-monitor/actions/runs/5663072587/job/15344172867

```
Target [CondaTarget[bioconductor-org.Dm.eg.db,version=3.16.0]] is not available in best practice channels - skipping
Target version [3.16.0] for package [bioconductor-org.Dm.eg.db] is not available in best practice channels - please specify the full version
```

which might be because `conda search --override-channels --channel conda-forge --channel bioconda --channel defaults --json --platform linux-64 bioconductor-org.Dm.eg.db` fails .. but strangely it works locally for me .. have not understood this completely

bernt-matthias:
> And that'll just work fine once we release new packages, or am I missing something?

What I'm wondering is that IUC and planemo-monitor are using `23.0`. But https://github.com/galaxyproject/galaxy/issues/16435 has been introduced in 23.1. So the fix of @nsoranzo won't fix this, or?

mvdbeek:
> 23.0

23.0 didn't change anything, why do you want to change this now ?

bernt-matthias:
I'm just saying that at the moment (with 23.0) planemo monitor (`planemo container_register`) fails due to the capitalization.

I have not found a way to debug this, yet. If you have an idea, I would be interested :)

Issue #5412, opened 2023-07-27
fastx_toolkit broken


bernt-matthias:
https://github.com/galaxyproject/tools-iuc/issues/5333 broke the fastx_toolkit

only one tool with a symlinked macros file was changed. tests did not run for the other tools.

- revert / restore missing macros?
- convert tool suite such that no symlinked macros exist?

PR #5415, opened 2023-07-27
Revert macros.xml

neoformit:
Fixes https://github.com/galaxyproject/tools-iuc/issues/5412 by reverting to old macros.xml

### DIFF
--- a/tool_collections/fastx_toolkit/fastx_renamer/fastx_renamer.xml
+++ b/tool_collections/fastx_toolkit/fastx_renamer/fastx_renamer.xml
@@ -1,4 +1,4 @@
-<tool id="cshl_fastx_renamer" name="Rename sequences" version="@VERSION@+galaxy1" >
+<tool id="cshl_fastx_renamer" name="Rename sequences" version="@VERSION@+galaxy@VERSION_SUFFIX@" >
     <description></description>
     <macros>
         <import>macros.xml</import>
--- a/tool_collections/fastx_toolkit/fastx_renamer/fastx_renamer.xml
+++ b/tool_collections/fastx_toolkit/fastx_renamer/fastx_renamer.xml
@@ -1,4 +1,4 @@
-<tool id="cshl_fastx_renamer" name="Rename sequences" version="@VERSION@+galaxy@VERSION_SUFFIX@" >
+<tool id="cshl_fastx_renamer" name="Rename sequences" version="@VERSION@+galaxy@VERSION_SUFFIX@" profile="22.05" >
     <description></description>
     <macros>
         <import>macros.xml</import>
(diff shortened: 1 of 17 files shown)


bernt-matthias:
So... Tests do not run (because only macro is changed).

Wondering if this is fine? I guess so. The tool that has been changed by the previous pr works anyway.. And the other ones have not been modified in the TS.

Or?

mvdbeek:
Can you bump the version or is that also completely contained in the macro ?

bernt-matthias:
I checked three tools: they had a hardcoded version like `1.0.1`.

neoformit:
Hmmm... should these tools have `+galaxyN` versions? I can't bump `version="1.0.2"`?

bernt-matthias:
Jep, just bump those versions.

neoformit:
Hmm ok, I can do that, but isn't it misleading if the tool form says `1.0.2` when the fastx_toolkit version is actually `0.0.14`?

mvdbeek:
You can't go backwards though, that'll mess with the latest version ordering. So the best thing to do is bump the +galaxyN suffix until the time upstream eventually catches up.

bernt-matthias:
It is, but we can't change it :(

neoformit:
I think we agreed on this :point_up:

bernt-matthias:
Can you do the same for all tools of the suite? Ideally adding a `VERSION_SUFFIX` token to the macros.

neoformit:
`fastx_renamer` already has `+galaxy1` suffix, so I have set `VERSION_SUFFIX=1` and annotated the entire suite with `version="x.x.x+galaxy@VERSION_SUFFIX@"`. Is that ok?

neoformit:
@bernt-matthias I am not sure how to drop the `from_file` attribute that planemo lint is complaining about.
Apparently it is deprecated in favour of [from_data_table](https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-inputs-param-options) but I'm not sure if this change will break current installations? I assume `fastx_clipper_sequences.txt.sample` needs to be converted to `fastx_clipper_sequences.loc.sample` for this to work?

bernt-matthias:
Leave it as it is.

Warnings are mainly for newly added tools. I will try to modify our workflow such that we can merge and deployment still happens.

bernt-matthias:
Good to merge?

neoformit:
I'm happy :+1:

tool_collections/fastx_toolkit/fastx_renamer/fastx_renamer.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="cshl_fastx_renamer" name="Rename sequences" version="@VERSION@+galaxy1" >
+<tool id="cshl_fastx_renamer" name="Rename sequences" version="@VERSION@+galaxy@VERSION_SUFFIX@" >
```
bgruening, line 1: this here had a version suffix of 1 and not its falling back to 0

bgruening:
When we touch those tools maybe we can add a profile version and bump the build number to 2?

tool_collections/fastx_toolkit/fastx_renamer/fastx_renamer.xml
neoformit: Good catch, I meant to set the version suffix to 1, not 0.

neoformit:
Shall I just add a profile version to them now?

bgruening:
I would say so. And bump to version 2 so all of them get a update.

neoformit:
Done

bernt-matthias:
Thanks @neoformit

PR #5424, opened 2023-08-02
Add --threads flag to bellerophon command

cat-bro:
FOR CONTRIBUTOR:

bellephoron  has a [--threads](https://github.com/davebx/bellerophon/blob/main/bellerophon/cli.py#L12) flag but it isn't used. The `GALAXY_SLOTS` variable is only used for the subsequent samtools command.

### DIFF
--- a/tools/bellerophon/bellerophon.xml
+++ b/tools/bellerophon/bellerophon.xml
@@ -2,7 +2,7 @@
     <description>chimeric reads from Arima Genomics</description>
     <macros>
         <token name="@TOOL_VERSION@">1.0</token>
-        <token name="@VERSION_SUFFIX@">0</token>
+        <token name="@VERSION_SUFFIX@">1</token>
     </macros>
     <requirements>
         <requirement type="package" version="@TOOL_VERSION@">bellerophon</requirement>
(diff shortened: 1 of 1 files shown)


bgruening:
Looks good to me, but looking at the source I don't think it will do much :(

https://github.com/davebx/bellerophon/blob/main/bellerophon/__init__.py

cat-bro:
For 24GB input in a pair of BAM files, the bellephoron job was initially taking 100 minutes with 1 CPU. After we raised the number of allocated CPUs from 1 to 8, the job completed in 80 minutes. After adding the --threads flag for the bellephoron command this was reduced to 70 minutes. Not as much of an effect as I had hoped, but definitely worth adding this.

cat-bro:
Admins will not need to kill any bellerophon jobs if the tool is configured to run on one CPU. For galaxies that running it with >1 CPUs for an inefficient speedup, this change does make it less inefficient.

bernt-matthias (approved):
I'm really curious how just allowing 8CPUs (i.e. without the `--threads` flag) can change the run time. In my opinion, it can only be random fluctuations. 

> this change does make it less inefficient

Still with a parallel efficiency of approx 18% enabling more than 1 thread is a waste of energy. 

Anyway, I'm approving it since it allows admins to limit the number of used CPUs explicitly to 1.

cat-bro:
this was already multithreaded - the threads were being used only for the samtools command and not the bellerophon command. If you are going to use threads, might as well use it for both, right? thanks for merging @bernt-matthias

PR #5426, opened 2023-08-02
Mapseq

paulzierep:
FOR CONTRIBUTOR:


tools/mapseq/mapseq.xml
```diff
@@ -0,0 +1,160 @@
+<tool id="mapseq" name="MAPseq" version="2.1.1+galaxy0">
```
bgruening, line 1:
```suggestion
<tool id="mapseq" name="MAPseq" version="2.1.1+galaxy0" profile="21.05">
```

tools/mapseq/mapseq.xml
```diff
@@ -0,0 +1,160 @@
...
+    <description>fast and accurate sequence read classification tool designed to assign taxonomy and OTU classifications to ribosomal RNA sequences.</description>
```
bgruening, line 2: Can we shorten that?

```suggestion
    <description>sequence read classification designed to assign taxonomy and OTU classifications</description>
```

tools/mapseq/mapseq.xml
```diff
@@ -0,0 +1,160 @@
...
+    <requirements>
+        <requirement type="package" version="2.1.1">mapseq</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+
+        ## the DB should contain only one fasta and txt file
+        #if $ref_db.db_source == "cached":
+            ls -la ${ref_db.db_cached.fields.path} &&
```
bgruening, line 10: remove?

tools/mapseq/mapseq.xml
```diff
@@ -0,0 +1,160 @@
...
+            ln -s ${ref_db.db_cached.fields.path}/*.fasta db.fasta &&
+            ln -s ${ref_db.db_cached.fields.path}/*.txt taxonomy.txt &&
```
bgruening, line 12: single quotes are needed, I hope that works
```suggestion
            ln -s '${ref_db.db_cached.fields.path}'/*.txt taxonomy.txt &&
```

bgruening: please the others as ell

tools/mapseq/mapseq.xml
```diff
@@ -0,0 +1,160 @@
...
+                <param type="data" name="taxonomy" label="Taxonomy file" format="tabular" />
+                <param type="data" name="mscluster" label="Database cluster" format="txt" optional="true" /> 
+            </when>
+        </conditional>
+               
+        <param argument="-tophits" type="integer" label="number of reference sequences to include in alignment phase" 
+               value="20" help="(-tophits)" />
+               
+        <param argument="-topotus" type="integer" label="number of internal reference otus to include in alignment phase" 
+               value="10" help="(-topotus)" />
```
bgruening, line 60:
```suggestion
               value="10" help="" />
```

This is done automatically when you use `argument`

tools/mapseq/mapseq.xml
```diff
@@ -0,0 +1,160 @@
...
+                </param>
+            </when>
+            <when value="history">
+                <param type="data" name="database" label="Database file (FASTA format)" format="fasta" />
+                <param type="data" name="taxonomy" label="Taxonomy file" format="tabular" />
+                <param type="data" name="mscluster" label="Database cluster" format="txt" optional="true" /> 
+            </when>
+        </conditional>
+               
+        <param argument="-tophits" type="integer" label="number of reference sequences to include in alignment phase"
```
bgruening, line 56: min/max?

tools/mapseq/mapseq.xml
```diff
@@ -0,0 +1,160 @@
...
+            <param name="sequences" value="ERR2237853_MERGED_FASTQ.fasta"/>
+            <param name="minscore" value="10"/>
+            <output name="classifications" file="ERR2237853_MERGED_FASTQ.mapseq" sort="true"/>
+        </test>
+    </tests>
+
+    <help><![CDATA[
+    
+MAPseq v1.2.6
+by Joao F. Matias Rodrigues, Thomas S. B. Schmidt, Janko Tackmann, and Christian von Mering
```
bgruening, line 114: this is the citation

tools/mapseq/mapseq.xml
```diff
@@ -0,0 +1,160 @@
...
+Institute of Molecular Life Sciences, University of Zurich, Switzerland
+Matias Rodrigues JF, Schmidt TSB, Tackmann J, and von Mering C (2017) MAPseq: highly efficient k-mer search with confidence estimates, for rRNA sequence analysis. Bioinformatics. http://doi.org/10.1093/bioinformatics/btx517
+
+Usage:
```
bgruening, line 118: this is not useful for a Galaxy tool

tools/mapseq/mapseq.xml
```diff
@@ -0,0 +1,160 @@
+<tool id="mapseq" name="MAPseq" version="2.1.1+galaxy0">
+    <description>fast and accurate sequence read classification tool designed to assign taxonomy and OTU classifications to ribosomal RNA sequences.</description>
+    <requirements>
+        <requirement type="package" version="2.1.1">mapseq</requirement>
```
bernt-matthias, line 4: Use a TOOL_VERSION macro token. Here an in few tool version attribute.

tools/mapseq/mapseq.xml
```diff
@@ -0,0 +1,160 @@
...
+        </param>       
+    </inputs>
+
+    <outputs>
+        <data format="tabular" name="classifications" label="Classification results"/>
+    </outputs>
+
+    <tests>
+        <test expect_num_outputs="1">
+            <param name="db_source" value="history" />
```
bernt-matthias, line 86: Can you add a test for cached?

tools/mapseq/mapseq.xml
```diff
@@ -0,0 +1,160 @@
...
+               value="20" help="(-tophits)" />
+               
+        <param argument="-topotus" type="integer" label="number of internal reference otus to include in alignment phase" 
+               value="10" help="(-topotus)" />
+               
+        <param argument="-minscore" type="integer" label="minimum score cutoff to consider for a classification, should be reduced when searching very small sequences, i.e.: primer search" 
+               value="30" help="(-minscore)" />
+               
+        <param argument="-minid1" type="integer" label="minimum number of shared kmers to consider hit in second phase kmer search" 
+               value="1" help="(-minid1)" />
```
bernt-matthias, line 66: Repeating the argument in parentheses is not necessary. Galaxy does this automatically (of you use argument... Which you do 👍)

tools/mapseq/mapseq.xml
```diff
@@ -0,0 +1,160 @@
...
+            </when>
+        </conditional>
+               
+        <param argument="-tophits" type="integer" label="number of reference sequences to include in alignment phase" 
+               value="20" help="(-tophits)" />
+               
+        <param argument="-topotus" type="integer" label="number of internal reference otus to include in alignment phase" 
+               value="10" help="(-topotus)" />
+               
+        <param argument="-minscore" type="integer" label="minimum score cutoff to consider for a classification, should be reduced when searching very small sequences, i.e.: primer search"
```
bernt-matthias, line 62: Label should be as short as possible. Put all explanations to help

bgruening:
@RZ9082 do you take this one over?

RZ9082:
> @RZ9082 do you take this one over?

Yes, just did the review update and merged the mapseq2biom tool as an option.

bgruening:
Thanks a lot @RZ9082!

@bernt-matthias Paul is on vacation, does the data table look ok for you? The layout? Do we always need to use a dbkey? It feels to be like this makes less sense here.

tools/mapseq/mapseq2biom.pl
bernt-matthias, line 1: Where is the peel script coming from? Are we allowed to distribute it?

tools/mapseq/tool_data_table_conf.xml.test
```diff
@@ -0,0 +1,7 @@
+<tables>
+    <!-- Location of MGnify database files -->
```
bernt-matthias, line 2: Comment refers to the wrong tool, or?

tools/mapseq/tool_data_table_conf.xml.sample
```diff
@@ -0,0 +1,7 @@
+<tables>
+    <!-- Location of MGnify database files -->
+    <table name="mapseq_db" comment_char="#">
+        <columns>value, dbkey, version, name, path</columns>
```
bernt-matthias, line 4: I would reorder the columns to value, name,... And remove dbkey.

tools/mapseq/mapseq.xml
```diff
@@ -0,0 +1,279 @@
...
+    <inputs>
+        <param type="data" name="sequences" format="fasta" label="Input sequences" />
+        <conditional name="ref_db">
+            <param name="db_source" type="select" label="Use cached database or database from history" help="">
+                <option value="cached">Cached database</option>
+                <option value="history">From history</option>
+            </param>
+            <when value="cached">
+                <param name="db_cached" type="select" label="Using built-in mapseq DB" help="">
+                    <options from_data_table="mapseq_db">
```
bernt-matthias, line 69: Are the correct columns used, if you do not specify them?

tools/mapseq/mapseq2biom.pl
RZ9082: This script is from the MGnify pipline v5.0:

https://github.com/EBI-Metagenomics/pipeline-v5/blob/master/tools/RNA_prediction/mapseq2biom/mapseq2biom.pl

The repo has Apache-2.0 license

tools/mapseq/tool_data_table_conf.xml.test
RZ9082: In this case, the test files are snippets from the database files MGnify uses for pipeline v5.0.

tools/mapseq/tool_data_table_conf.xml.sample
RZ9082: done.

tools/mapseq/mapseq.xml
RZ9082: also done:)

tools/mapseq/mapseq2biom.pl
bernt-matthias: Is there maybe a conda package containing it?

bernt-matthias: Then we have to include the licence file if we want to include it here. Maybe also have a reference to the original source in the help of or a readme file?

bgruening: We don't need the license file but can annotate it in the `<tool license="">` field isn't it?

bernt-matthias: I'm not an expert, but https://github.com/EBI-Metagenomics/pipeline-v5/blob/a83ee883bb3c7480010fa952939fac771491ddf4/LICENSE#L94

RZ9082: I added source and license of mapseq2biom to the help section.

RZ9082: Please inform me if you have any further improvement suggestions!

tools/mapseq/mapseq.xml
```diff
@@ -0,0 +1,298 @@
...
+            <param name="krona_input" value="yes"/>
+            <param name="otu_table" value="mapseq_db/test.otu" />
+            <output name="classifications" file="sequences.mapseq" sort="true"/>
+            <output name="otu_tsv" file="mapseq2biom/tab-output_including_taxIDs.tabular" />
+            <output name="otu_tsv_notaxid" file="mapseq2biom/tab-output_without_taxIDs.tabular" />
+            <output name="krona_format" file="mapseq2biom/krona_input.tabular" />
+        </test>
+        <test expect_num_outputs="1">
+            <param name="db_source" value="cached" />
+            <param name="db_cached" value="MGnify LSU Reference Database 0.1" />
```
bgruening, line 192: the value should be `mapseq_db` the unique ID from the tool_data_table_conf.xml.test file isn't it?

RZ9082: Yes, you're right. Thanks

bgruening: Ok, but why did the tests then pass? there is something fishy here, isn't it?

RZ9082: > Ok, but why did the tests then pass? there is something fishy here, isn't it?

when writing the cached tests I took inspiration from this tool: https://github.com/galaxyproject/tools-iuc/blob/65fa34f035934b4e9c1fbeb5dffe00b12e6a0f32/tools/ucsc_blat/blat.xml , where the display name is passed as the value for the test.

To check the matter i have just modified the test output to make fail/mismatch, and ran the test twice, one time passing the unique ID and second time passing the display name. Both of them failed.
So could it be that both unique ID and display name are accepted?

RZ9082: So after testing several other tools from the iuc repo now (passing name instead of unique_build_id),
I could say passing the name also works, it did not affect the results of the tests!!

bernt-matthias: The name and the value should work for all select parameters.

bernt-matthias: We also have a test for this in the galaxy code https://github.com/galaxyproject/galaxy/blob/9877fed8023e06e256beb30d66e4da7d29abb7ef/test/functional/tools/dbkey_filter_input.xml#L46

RZ9082: Thanks, i wasn't quite sure about that.

tools/mapseq/mapseq.xml
```diff
@@ -0,0 +1,298 @@
...
+
+        <param argument="-outfmt" type="select" label="Output format">
+            <option value="simple">simple</option>
+            <option value="confidences">confidences</option>
+        </param>
+    </inputs>
+
+    <outputs>
+        <data format="tabular" name="classifications" label="Classification results"/>
+        <data name="otu_tsv" format="tabular" label="tab-output including taxIDs">
```
TomHarrop, line 134: Soudn't this be a biom format?

RZ9082: Do you mean 'BIOM1' and 'BIOM2' formats?

This is actually a part of the MGnify amplicon pipeline, which I'm porting to Galaxy. The output of 'mapseq2biom' is neither JSON nor HDF5; it is a tab-separated values file. The step following 'mapseq2biom' in the workflow would involve converting it to JSON and HDF5 using 'biom-convert.'

I hope this answers your question.

bgruening: So the script `mapseq2biom` does not really convert the files to BIOM?

If so can and should we not include the `biom-convert` step also in this tool. I would prefer that tools output a standard format by default and not offering some wired tabular files that is useful users.

Does that make sense?

paulzierep: No, it does not really convert to BIOM1/2 format. I think the naming is not ideal (by the original authors :)); the script does indeed produce a Taxon Table (reads per taxon similar to kraken2 and metaphlan output). This output is sometimes referred to as the classic biom format (I think therefore the confusion). This output can be converted to biom1/2 format using biom-convert downstream.
I think we should just change the name of the optional output to taxon table and explain in the help, that it can be used as input for biom-convert. I would not add the biom-convert step here, since biom-convert has more functionality (e.g. add metadata) and the user might want to work first with the taxon table to create 
* relative abundances
* sum up specific taxa levels
* join multiple tables before converting to biom format  

OK ?

tools/mapseq/mapseq.xml
```diff
@@ -0,0 +1,298 @@
...
+        <param type="data" name="sequences" format="fasta" label="Input sequences" />
+        <conditional name="ref_db">
+            <param name="db_source" type="select" label="Use cached database or database from history" help="">
+                <option value="cached">Cached database</option>
+                <option value="history">From history</option>
+            </param>
+            <when value="cached">
+                <param name="db_cached" type="select" label="Using built-in mapseq DB" help="">
+                    <options from_data_table="mapseq_db">
+                        <column name="value" index="0" />
```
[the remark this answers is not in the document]
paulzierep: I will adjust the DM in any case

paulzierep (commented):
Use otu table instead of biom,
minor text changes

tools/mapseq/mapseq.xml
bgruening: Ok.

bgruening (approved):

bgruening:
Thanks, very cool!

PR #5432, opened 2023-08-07
Updating tools/humann from version 3.7 to 3.8

planemo-autoupdate:
Hello! This is an automated update of the following tool: **tools/humann**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

I have updated tools/humann from version 3.7 to 3.8.

**Project home page:** http://huttenhower.sph.harvard.edu/humann

### DIFF
--- a/tools/humann/macros.xml
+++ b/tools/humann/macros.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <macros>
-    <token name="@TOOL_VERSION@">3.7</token>
+    <token name="@TOOL_VERSION@">3.8</token>
     <token name="@VERSION_SUFFIX@">0</token>
     <token name="@PROFILE@">22.01</token>
     <xml name="edam_ontology">


bgruening:
ping @tuncK

tuncK:
Looks like no new DB was released, and this is supposed to work as is. I guess you can merge this PR as is.

On the other hand, they changed the quantification algorithm slightly regarding what happens if there are multiple enzymatic pathways achieving the same thing. They also updating a mapping file here: https://github.com/biobakery/humann/commit/26e63c134487f3f6e3402a3ceba966756988a811#diff-3c447f69ab06fe9cf807a0956c442c4532ec957221815f3003498d2f395dc552. So our tests pass, but the numerical results that one gets with real world datasets might slightly differ between v3.8 and v3.7.

bgruening:
Results have to change, hopefully to the better and more correct... Galaxy is versioning the tool so users have to actively opt in to an update - and therefore to changes.

Thanks a lot for looking at the changes.

PR #5436, opened 2023-08-07
Add bedops sort

mvdbeek:
which should be a little faster than gnu sort and can be memory limited (which bedtools sort is not)

FOR CONTRIBUTOR:


tools/bedops/sort-bed.xml
```diff
@@ -0,0 +1,63 @@
+<tool id="bedops-sort-bed" name="sort-bed" version="2.4.41" profile="22.01">
```
bernt-matthias, line 1: Version token would be good.

bernt-matthias, line 1: Can you add "bedops" to the name?

tools/bedops/sort-bed.xml
```diff
@@ -0,0 +1,63 @@
...
+    <command><![CDATA[
+sort-bed 
+--max-mem "\${GALAXY_MEMORY_MB:-1024}M"
+--tmpdir "\${_GALAXY_JOB_TMP_DIR:-.}"
+$unique
+$duplicates
+#for file in $inputs:
+    '$file'
+#end for
+> $output
```
bernt-matthias, line 14: Output should also be quoted.

tools/bedops/sort-bed.xml
```diff
@@ -0,0 +1,63 @@
...
+    ]]></command>
+    <inputs>
+        <param name="inputs" type="data" format="bed" multiple="true"/>
```
bernt-matthias, line 17: Is multiple intended?

tools/bedops/.shed.yml
```diff
@@ -0,0 +1,13 @@
+categories:
```
bernt-matthias, line 1: Hoping that there will more tools from bedops.. How about defining the shed yaml file as suite?

mvdbeek: When we need it ?

tools/bedops/sort-bed.xml
mvdbeek: Yes

tools/bedops/sort-bed.xml
mvdbeek: I don't like that it doesn't match the command line, but I see we also do that with bedtools :(

bgruening (approved):

PR #5437, opened 2023-08-08
Add unequal column table sanitizer tool

paulzierep:
FOR CONTRIBUTOR:

I am not soo sure about the name, good and catchy ideas welcome !


wm75:
The Cut1 tool (e.g. https://usegalaxy.eu/root?tool_id=Cut1) can do most of this. From the help: "If you specify more columns than actually present - empty spaces will be filled with dots".
If you tell it to **Cut columns**: `c1,c2,c3,c4,c5,c6,c7,c8,c9` from your `test-data/input.tsv` it works nicely.
You just can't specify the fill value.

wm75:
For this use case, Cut1 requires you too know the number of columns in the longest line, but then only reuires one pass over the data instead of two.

bgruening:
I would also check the table-compute tool.

tools/unequal_column_sanitizer/unequal_column_sanitizer.xml
```diff
@@ -0,0 +1,69 @@
...
+Output
+~~~~~~
+ === ============ ============== ==============
+ 139 Bacteria     NA             NA            
+ 3   Bacteria     Actinobacteria Actinobacteria
+ 1   Bacteria     Bacteroidetes  NA
+ === ============ ============== ==============
+
+The output can be used for tools that perform tabluar operations such as grouping, split_file_on_column, filter_tabular and table_compute,
+that otherwise fail due to unequal column numbers.
```
wm75, line 63: That's true, unfortunately, Table compute can't even read the file.

wm75:
toolshed.g2.bx.psu.edu/repos/devteam/column_maker/Add_a_column1/2.0 can also add missing columns (see the last example in the help at https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/devteam/column_maker/Add_a_column1/2.0), but I wouldn't recommend using it for multiple missing columns.

Cut1 seems like the best existing choice.

bgruening:
By coincidence, @fubar2 has wrapped a similar tool, he mentioned two features that would be cool to add to this PR here.

* keeping (optional) empty rows
* deal with headers in input files (What does currently happens if you add input files with headers)

paulzierep:
I agree, the Cut1 approach seems the best option at the moment, however it is difficult for the user to find Cut1, since most tabular tools fail with this input; it is fair to assume that Cut1 would as well. 
This tool would not require knowing the number of columns in advance, which I think might be advantageous sometimes.
I will add the ideas from @fubar2 as well.

PR #5438, opened 2023-08-08
added cmsearch-deoverlap tool

RZ9082:
FOR CONTRIBUTOR:

cmsearch-deoverlap removes overlapping hits to models in the same clan in a single --tblout  cmsearch output file.


tools/cmsearch_deoverlap/cmsearch-deoverlap.xml
```diff
@@ -0,0 +1,46 @@
...
+    <tests>
+        <test>
+            <param name="clan_information" value="ribo.claninfo" />
+            <param name="cmsearch_matches" value="1.cmsearch.tblout" />
+            <output name="deoverlapped_matches" file="1.cmsearch.tblout.deoverlapped">
+            </output>
+        </test>
+    </tests>
+    <help><![CDATA[
+Usage:
```
paulzierep, line 29: I think the part can be removed.

tools/cmsearch_deoverlap/cmsearch-deoverlap.xml
```diff
@@ -0,0 +1,46 @@
...
+    </requirements>
+    <command><![CDATA[
+        ln -s '$clan_information' clan_info;
+        ln -s '$cmsearch_matches' cmsearch_matches;
+        LC_ALL=C perl '$__tool_directory__/cmsearch-deoverlap.pl' --clanin clan_info cmsearch_matches;
+        mv *.deoverlapped '$deoverlapped_matches'
+
+    ]]></command>
+    <inputs>
+        <param type="data" name="clan_information" format="txt" label="clan information"/>
```
paulzierep, line 14: A small help text would be good.

paulzierep (commented):
some text changes

tools/cmsearch_deoverlap/cmsearch-deoverlap.xml
```diff
@@ -0,0 +1,60 @@
+<tool id="cmsearch_deoverlap" name="CMsearch deoverlap" version="0.08+galaxy0">
```
bgruening, line 1: please add a profile="21.05"

tools/cmsearch_deoverlap/cmsearch-deoverlap.xml
```diff
@@ -0,0 +1,60 @@
...
+    <description>removes lower scoring overlaps from cmsearch --tblout files</description>
```
bgruening, line 2:
```suggestion
    <description>removes lower scoring overlaps from cmsearch results</description>
```

bgruening (commented):
@RZ9082 @paulzierep this script has no parameter and is very specific to one exact tool/outputs. Should we add this as option to cmsearch directly?

tools/cmsearch_deoverlap/cmsearch-deoverlap.pl
```diff
@@ -0,0 +1,495 @@
...
+# hits 1 and 4 no longer overlap.
+#
+use strict;
+use warnings;
+use Getopt::Long;
+use File::Basename;
+
+my $in_tblout  = "";   # name of input tblout file
+
+my $usage;
```
paulzierep, line 41: Can we add the sort hit (bit score / E-value), clanin or not, and maxkeep / not max keep as parameter ?

paulzierep (commented):
add parameters

paulzierep:
> @RZ9082 @paulzierep this script has no parameter and is very specific to one exact tool/outputs. Should we add this as option to cmsearch directly?

I think there are 3 parameters that should be added, see my review.
I had similar thoughts about integrating it in cmsearch, but there could be cases where one would want to use it on old data, and therefore I would prefer to have it separately.

paulzierep:
> Please use latest version from original repo
> 
> https://github.com/nawrockie/cmsearch_tblout_deoverlap/blob/master/cmsearch-deoverlap.pl
> 
> Please ask there if they can add a license otherwise we can not distribute it: https://choosealicense.com/no-permission/
> 
> May fit better to @bgruenings rnateam repo

Interesting, if EBI-Metagenomics copied it into their repo: https://github.com/EBI-Metagenomics/pipeline-v5 and added a license, we cannot use it since the original does not have one, right ?

bernt-matthias:
> we cannot use it since the original does not have one, right ?

No. Just ask the maintainers of the original repo to add a license.

PR #5463, opened 2023-08-29
Annotatemyids bt

pavanvidem:
FOR CONTRIBUTOR:


bernt-matthias (approved):

bgruening:
Adding ew dBs is that very sustainable as the container grows and grows. Do you know if we can install a DB on the fly in the home directory?

I will merge, but we need to find a way how we can install this package as part of the tool - installing NOT into the system packages which are write only but in the current working dir.

bernt-matthias:
Maybe even a data manager creating an [renv](https://rstudio.github.io/renv/articles/renv.html).

pavanvidem:
Thanks.  It was requested by a user. DM with renv seems like a solution.

PR #5464, opened 2023-08-30
Add hybpiper

TomHarrop:
Adding the tool hybpiper

---

FOR CONTRIBUTOR:


bgruening:
@TomHarrop to trigger tests you need to add a .shed.yml file.

tools/hybpiper/hybpiper.xml
```diff
@@ -22,7 +21,12 @@
         --allow_gene_removal
         hybpiper.ctl
     #elif str( $job_conditional.hybpiper_job ) == "assemble":
-        ln -s \$_GALAXY_JOB_TMP_DIR hp_out
+        #if str($job_conditional.prefix) != '':
+            #set sample_prefix = str($job_conditional.prefix)
+        #else:
+            #set sample_prefix = str($R1.name).split('.')[0]
+        #end if
+        ln -s \$_GALAXY_JOB_TMP_DIR '${sample_prefix}'
```
bgruening, line 29: Do not user Galaxy internals in tools, tools are supposed to be Galaxy-independent if possible.
You could use $TEMP or any other POSIX compatible location. Galaxy will the set those variables for you.

In this case I'm wondering why we are not using simply the CWD, so the directory in which Galaxy puts us before executing the script.

```
     mkdir ./temp_dir &&
     --prefix ./temp_dir &&
     --directory ./temp_dir 
```

tools/hybpiper/hybpiper.xml
```diff
@@ -0,0 +1,221 @@
...
+        <conditional name="job_conditional">
+            <param name="hybpiper_job" type="select" label="Type of hybpiper run">
+                <option value="check_and_fix_targetfile">Check and fix targetfile</option>
+                <option value="assemble">Assemble target loci</option>
+                <option value="stats">Extract sequences and/or stats from Hybpiper runs</option>
+            </param>
+            <when value="check_and_fix_targetfile">              
+            </when>
+            <when value="assemble">
+                <param name="R1" type="data" format="fastq" label="R1 read file"/>
```
bernt-matthias, line 94: Is this `fastq` (ie at all fastq variants supported) or `fastqsanger`?

tools/hybpiper/hybpiper.xml
```diff
@@ -0,0 +1,221 @@
...
+                <param name="R2" type="data" format="fastq" label="R2 read file" />
+                <param name="unpaired" optional="true" type="data" format="fastq" label="Unpaired reads" help="If you kept the unpaired reads from adaptor trimming, you can use them here." />
+                <param argument="--prefix" optional="true" type="text" value="" label="Sample name" help="Provide a name for this sample. By default, the name of the R1 read file will be used.">
+                    <sanitizer invalid_char="">
```
bernt-matthias, line 98: I think the sanitizer is redundant. The validator seems not to allow any chars that would be sanitized.

tools/hybpiper/hybpiper.xml
```diff
@@ -0,0 +1,221 @@
...
+                <param name="output_type" type="select" multiple="true" label="Choose Hybpiper summary outputs">
+                    <option value="stats">Stats</option>
+                    <option value="heatmap">Heatmap</option>
+                    <option value="sequences">Sequences</option>
+                </param>
+            </when>
+        </conditional>
+    </inputs>
+    <outputs>
+        <collection type="list" name="output_targetfile" label="${tool.name} check and fix output on ${on_string}">
```
bernt-matthias, line 117: All outputs should have labels. Otherwise all have the same default label.

tools/hybpiper/hybpiper.xml
TomHarrop: Removed, thanks for explaining

tools/hybpiper/hybpiper.xml
TomHarrop: Thanks, I always miss that :S

tools/hybpiper/hybpiper.xml
TomHarrop: Fixed

tools/hybpiper/hybpiper.xml
TomHarrop: Thanks, added

tools/hybpiper/hybpiper.xml
```diff
@@ -0,0 +1,295 @@
...
+        </output_collection>
+    </test>
+
+    <!-- test2: assemble with paired collection -->
+    <test expect_num_outputs="1">
+        <param name="targetfile_dna" value="test_targets.fasta.gz"/>
+        <conditional name="job_conditional">
+            <param name="hybpiper_job" value="assemble"/>
+            <param name="paired_input">
+                <collection type="paired" name="NZ874">
```
TomHarrop, line 216: Hi @bgruening @bernt-matthias thanks for your feedback, this is almost ready but I need some help with this test.

Setting `name` on the collection causes a linting error (The attribute 'name' is not allowed).

If i I remove `name`, `element_identifier` is set to `Unnamed Collection`  , hybpiper chokes on the space, and the test fails.

Is there a way to set the `element_identifier` in `<test>` -> `<collection> `?

bernt-matthias: Setting the collection name seems not possible at the moment. Could be worth to open an issue for this. But the question is: do the tests depend on that? I guess they should not.

TomHarrop: Thanks, I opened an issue: https://github.com/galaxyproject/galaxy/issues/16654

It does work (element_identifier resolves to NZ874), it's just linting that fails.

For now, I've added a check for special characters and set this test to `expect_failure="true"` in https://github.com/galaxyproject/tools-iuc/pull/5464/commits/2eeb313a6003fbfe070e0a3620da4b5a7f7b0f61

TomHarrop:
Thanks again for the help, I think this is ready for review / merge now. :tada:

tools/hybpiper/hybpiper.xml
```diff
@@ -0,0 +1,351 @@
...
+HybPiper assemblies. HybPiper relies on the directory hierarchy it creates for each
+sample during assembly. The hierarchy is based on the name of the
+sample, which you provide to Galaxy as the identifier in the collection.
+
+Using paired collections
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you have your sequencing reads in individual datasets, you can easily organise them into a paired
+collection. See the Galaxy training material on `using dataset
+collections <https://training.galaxyproject.org/training-material/topics/galaxy-interface/tutorials/collections/tutorial.html>`__
```
bgruening, line 303: can you please use the more stable pURLs https://gxy.io/GTN:T00146

bgruening:
```suggestion
collections <https://gxy.io/GTN:T00146>`__
```

tools/hybpiper/hybpiper.xml
```diff
@@ -0,0 +1,351 @@
...
+                    <action name="column_names" type="metadata" default="Name,NumReads,ReadsMapped,PctOnTarget,GenesMapped,GenesWithContigs,GenesWithSeqs,GenesAt25pct,GenesAt50pct,GenesAt75pct,GenesAt150pct,ParalogWarningsLong,ParalogWarningsDepth,GenesWithoutStitchedContigs,GenesWithStitchedContigs,GenesWithStitchedContigsSkipped,GenesWithChimeraWarning,TotalBasesRecovered" />
+                </actions>
+            </data>
+            <data name="stats_supercontig" label="Hybpiper statistics (supercontig)" format="tabular" from_work_dir="stats.supercontig.tsv">
+                <actions>
+                    <action name="column_names" type="metadata" default="Name,NumReads,ReadsMapped,PctOnTarget,GenesMapped,GenesWithContigs,GenesWithSeqs,GenesAt25pct,GenesAt50pct,GenesAt75pct,GenesAt150pct,ParalogWarningsLong,ParalogWarningsDepth,GenesWithoutStitchedContigs,GenesWithStitchedContigs,GenesWithStitchedContigsSkipped,GenesWithChimeraWarning,TotalBasesRecovered" />
+                </actions>
+            </data>
+            <data name="seqlengths_gene" label="Assembled sequence lengths (gene)" format="tabular" from_work_dir="seq_lengths.gene.tsv">
+            </data>
```
bgruening, line 190:
```suggestion
            <data name="seqlengths_gene" label="Assembled sequence lengths (gene)" format="tabular" from_work_dir="seq_lengths.gene.tsv"/>
```

tools/hybpiper/hybpiper.xml
```diff
@@ -0,0 +1,351 @@
...
+            </when>
+    
+            <when value="stats">
+                <param name="hybpiper_results" type="data_collection" collection_type="list" format="tar" multiple="true" label="Results from Hybpiper assemble runs" />
+                <param name="stats_type_select" type="select" label="Choose statistics to report" display="checkboxes" multiple="true">
+                    <option value="gene" selected="true">Gene</option>
+                    <option value="supercontig">Supercontig</option>
+                </param>
+                <param name="heatmap" type="boolean" checked="false" label="Produce a heatmap for each of the selected statistics" />
+                <param name="sequence_type_select" type="select" display="checkboxes" label="Choose sequences to extract" multiple="true">
```
bgruening, line 151: is this one optional or not, is it possible to select none of them?

tools/hybpiper/hybpiper.xml
```diff
@@ -0,0 +1,351 @@
...
+    
+        <conditional name="job_conditional">
+            <param name="hybpiper_job" type="select" label="Type of hybpiper run">
+                <option value="check_and_fix_targetfile">Check and fix targetfile</option>
+                <option value="assemble" selected="true">Assemble target loci</option>
+                <option value="stats">Extract sequences and/or stats from Hybpiper runs</option>
+            </param>
+    
+            <when value="check_and_fix_targetfile">              
+            </when>
```
bgruening, line 138:
```suggestion
            <when value="check_and_fix_targetfile"/>
```

tools/hybpiper/hybpiper.xml
TomHarrop: They are all optional. I marked them optional in [28bcb4490](https://github.com/galaxyproject/tools-iuc/pull/5464/commits/28bcb4490d2fc964378cb6e3cb17d812d019956d).

Although if the user really deselects everything there won't be any outputs in Galaxy.

TomHarrop:
Thanks for the review @bgruening

tools/hybpiper/hybpiper.xml
bgruening: We should avoid that, correct? Thats why I have ask. To make sure the users has at least one output.

bgruening: Sorry, if I confused you :(

bgruening: but then we simply make it optional=false?

bgruening: I mean the output select?

TomHarrop: OK, I made stats `optional=false` in https://github.com/galaxyproject/tools-iuc/pull/5464/commits/022a3b2f5299671aab60dfe1e036b1d73e6dcbb0.

stats will be generated no matter what else the user selects. I wanted to avoid it because the underlying software is so slow but it's probably an edge case.

TomHarrop: Hmm can't use checkboxes in that case. I think this would make the tool worse than having no output if all the output is de-selected. The user has to actually deselect everything since I have  `selected="true"` on the usual options.

Apart from making it a single param, is there a way to say the user has to select at least one item from `$job_conditional.stats_type_select` or `$job_conditional.sequence_type_select`?

bgruening:
Damn the linting was not happy with checkbox and optional=false. Are you happy with the last commit. Then we can merge I think

TomHarrop:
I would personally just revert [022a3b2](https://github.com/galaxyproject/tools-iuc/pull/5464/commits/022a3b2f5299671aab60dfe1e036b1d73e6dcbb0) but I can live with it either way.

bgruening:
Ok, please revert.

https://github.com/galaxyproject/tools-iuc/blob/c083399ddb31b026aaab1d600d8789bf56cf4d4b/tools/homer/macros.xml#L49

This you could try. A validator with type="no_option", maybe that works?

TomHarrop:
Sorry, I had an urgent appointment with a bibimbap last night :rice: 

I think the confusion is that there are **two optional params** here. The user can select nothing from either of the selectors, but has to select at least one output across the two params. So the validator didn't work either. 

I'm doing this because the **setup** for extracting stats (selector one) or extracting sequences (selector two) is expensive and it could be run on hundreds of samples at once. So it's good that the user can do both in one tool run. BUT, the procedure for extracting those things is also slow, so it's good that the user can exclude one if they don't want that output.

I've got the effect by adding an output that Galaxy expects if the user deselects everything. Then there is a check in the `<command>` and an error will be visible in the history. Added a test for that too. Hopefully that does it!

Thanks again @bgruening

bgruening:
> Ok, please revert.
> 
> https://github.com/galaxyproject/tools-iuc/blob/c083399ddb31b026aaab1d600d8789bf56cf4d4b/tools/homer/macros.xml#L49
> 
> This you could try. A validator with type="no_option", maybe that works?

Have you tried this validator?

TomHarrop:
Yes

TomHarrop:
I can't use the validator because there are two params. This is what it's doing:

## 1. This works. Stats are optional.

screenshot-localhost_9090-2023 09 22-13_55_47

## 2. This works. Sequences are optional.

screenshot-localhost_9090-2023 09 22-13_56_15

## 3. This should fail (and does, with "ERROR: No outputs selected")

screenshot-localhost_9090-2023 09 22-13_56_29

screenshot-localhost_9090-2023 09 22-14_08_01

bgruening (approved):

bgruening:
Thanks @TomHarrop!

PR #5471, opened 2023-09-08
featurecounts: take unsorted.bam

bernt-matthias:
if sam is allowed also unsorted.bam (i.e. any bam) should be. also fixed the test data ftypes to what they actually are (should save a few implicit conversions in the testing).

From the manual: 

> If the input contains location-sorted paired-end reads, featureCounts will automatically
re-order the reads to place next to each other the reads from the same pair before counting
them. Sometimes name-sorted paired-end input reads are not compatible with featureCounts
(due to for example reporting of multi-mapping results) and in this case featureCounts will
also automatically re-order them.

FOR CONTRIBUTOR:


tools/featurecounts/featurecounts.xml
```diff
@@ -137,7 +137,7 @@
         <param name="alignment"
                type="data"
                multiple="false"
-               format="bam,sam"
+               format="unsorted.bam,sam"
```
bgruening, line 140: can we also add `bam` here, just for clarity ... I don't think is very intuitive  that `unsorted.bam` also includeds sorted bam.

bernt-matthias: We could certainly do this since it would not change anything.

But it would be better to fix this on the Galaxy side. At the moment the allowed datatypes are only shown if the history does not contain such a dataset or if the input is optional. Otherwise a dataset is "loaded" and the list of possible datatypes is not shown anymore. 

Maybe they should be shown next to the label together with a tooltip explaining the datatype.

bgruening: That's sounds more like a UI issue. Let's add here BAM as  well.

bernt-matthias, line 140:
```suggestion
                format="unsorted.bam,bam,sam"
```

bgruening (approved):

PR #5476, opened 2023-09-11
Update shovill tool to accept fastq and fastq.gz

pimarin:
Shovill seem to failed when input is indicated as fast or fastq.gz while it can work with this input.
I just add the fastq as input format

### DIFF
--- a/tools/shovill/shovill.xml
+++ b/tools/shovill/shovill.xml
@@ -2,7 +2,7 @@
     <description>Faster SPAdes assembly of Illumina reads</description>
     <macros>
         <token name="@TOOL_VERSION@">1.1.0</token>
-        <token name="@VERSION_SUFFIX@">1</token>
+        <token name="@VERSION_SUFFIX@">2</token>
     </macros>
     <xrefs>
         <xref type="bio.tools">shovill</xref>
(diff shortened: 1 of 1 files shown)


bernt-matthias:
Why do you use fastq as format for your data and not fastqsanger or fastqillumina?

Fastq is essentially an unspecified fastq which should be used if the underlying tool can work with both phred encodings. Is this the case here?

pimarin:
The tool failed when we submitted classical fastq or fastq.gz illumina reads (while shovill command line accept this input type). If I force to change data input as fastqsanger it's okay but not good 
In the wrapper only `fastqsanger,fastqsanger.gz,fastqsanger.bz2` were provided, while in other assembler (for example spades) or tools which analyse fastq files, `fastq and fastq.gz` are also indicated.

bernt-matthias:
Then we should explicitly add `fastqillumina` and the zipped variants. 

`fastq` would also allow [`fastqcssanger`](https://github.com/galaxyproject/galaxy/blob/3c6028352405545e484b7ab6467f3572deef1468/lib/galaxy/datatypes/sequence.py#L972C12-L972C12) which is probably not desired.

bernt-matthias:
But I also have to add that the fact that a tool accepts both types may not be the same as that the tool correctly works with both. 

Most of the time tools accepting both types either 

- ignore qualities anyway
- have a flag to set the phred score
- or detect internally (which is seldom I think)

pimarin:
Humm you  right, Shovill explicitaly work with ` Illumina paired-end reads`as the author said. So it's better to remove the `fastqsanger and fastq` like option and only add `fastqillumina ` ? but I need to force the datatype when a fastq file is used (even if it's fastq from illumina)

bernt-matthias:
> Humm you right,

Ironically I can't agree with this statement. If you check [their test data](https://github.com/tseemann/shovill/tree/master/test) they have fastq files with all quality values being `2` which is not possible in `fastqillumina`, but it is in `fastqsanger`. So they likely can deal with `fastqsanger`.

Maybe it's best to just open an issue at their repo and ask if shovill can deal with fastq data of any quality encoding.

bgruening:
They can for sure deal with fastqsanger.

pimarin:
I don't have any answer from my issue on shovill repo, so it's better to keep only fastqsanger as previously write on the wrapped. So my PR is not necessary.

bgruening (approved):

PR #5483, opened 2023-09-15
Add tsebra

rlibouba:
FOR CONTRIBUTOR:


bgruening (commented):
You test files are too big. Howver you could just remove them as you do not need them currently. You only test with asserts.

However, I think we should use them, at least a part of it. It wil be good with future updates to see (in a future diff) what has changed and if the GTFs still look good.

You could remove most of the lines of your test-files and then use `contains`

> file="mode_mp_output1.gff" ftype="gff" compare="contains"/>

https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-tests-test-output

bgruening:
The rest looks really great @rlibouba!

tools/tsebra/tsebra.xml
```diff
@@ -69,7 +69,7 @@
             <param name="ignore_transcrpts" value="false"/>
             <param name="filter_exon" value="false"/>
             <output name="output" ftype="gtf" file="output.gtf" lines_diff="2"/>
-            <output name="output_tab" ftype="txt" file="output_tab.txt" lines_diff="2"/>
+            <output name="output_tab" ftype="txt" file="output_tab.txt" compare="sim_size" delta="100"/>
```
bgruening, line 72: please do not compare text files via sim_size
Use asserts instead. https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-tests-test-output-assert-contents

rlibouba:
I'm having difficulty with the second test, in particular with the second output generated by the --score_tab parameter. In one column, the rounding of the numbers is not always the same, which causes the test to fail. When I run the tests locally with planemo test, the tests pass

tools/tsebra/tsebra.xml
```diff
@@ -0,0 +1,88 @@
...
+    #end if 
+    
+    -o '$output' 
+
+    ]]></command>
+    <inputs>
+        <param name="gtf_file" argument="-g" type="data" format="gtf" optional="true" label="List of gene prediciton files"/>
+        <param name="keep_gtf" argument="-k" type="data" format="gtf" optional="true" label="Gene prediction list" help="All transcripts from these gene sets will be included in the output"/>
+        <param name="hint_file" argument="-e" type="data" format="gff" optional="true" label="List of files containing extrinsic evidence"/>
+        <param name="cfg_file" argument="-c" type="data" format="txt" optional="true" label="Configuration file" help="Find the recommanded parameter at config/default.cfg"/>
```
bgruening, line 39: All 4 inputs are optional, in theory a user can start this tool without inputs, I think this should not happen.

Do you have a smart idea here @rlibouba ?

tools/tsebra/tsebra.xml
```diff
@@ -42,8 +38,8 @@
         <param name="hint_file" argument="-e" type="data" format="gff" optional="true" label="List of files containing extrinsic evidence"/>
         <param name="cfg_file" argument="-c" type="data" format="txt" optional="true" label="Configuration file" help="Find the recommanded parameter at config/default.cfg"/>
         <param name="score_tab" argument="-s" type="boolean" checked="false" label="Print the transcript scores as a table to the specified file"/>
-        <param name="ignore_transcrpts" type="boolean" label="Ignore the phase transcripts while detecting clusters of overlapping transcripts"/>
-        <param name="filter_exon" type="boolean" label="Filter out all single-exon genes"/>
+        <param argument="--ignore_tx_phase" type="boolean" truevalue="--ignore_tx_phase" falsevalue="" label="Ignore the phase transcripts while detecting clusters of overlapping transcripts"/>
```
bgruening, line 41: @rlibouba this is a general trick to handle booleans. It makes your commandline section easier, see above, the if is not needed and its more readable imho.

tools/tsebra/tsebra.xml
rlibouba: Hi Björn, thank you for your return. I've suggested something for that. What do you think?

bgruening:
Do you need help with the failing test?

rlibouba:
Yes, I'd like to. Thank you. I don't understand the error output in relation to the choice of tests. The numbers in the penultimate column are completely different.

bgruening:
grafik

Looks like the test file is not up to date. You could use asserts and not file comparisons if this helps.

bgruening (approved):

bgruening:
Thanks @rlibouba

bgruening:
@rlibouba running it on master resulted in: `galaxy.tool_util.verify INFO: ## files diff on '/tmp/tmp1g96p8z2output_tab.txt' and '/tmp/tmpahhbf4n7output_tab.txt': lines_diff = 0, found diff = 278`

Is this output a bit random? Do you need to sort it?

rlibouba:
I've created a PR for the update. All tests passed. There were no major changes to the files. 
Here's the link: https://github.com/galaxyproject/tools-iuc/pull/5543

PR #5492, opened 2023-09-20
Simplify and fix trinity scratch dir handling

mvdbeek:
This is currently failing on usegalaxy.org with

```
cp: -r not specified; omitting directory 'trinity_out_dir/Trinity.tmp.fasta.salmon.idx'
```

we now use the same logic throughout tool regardless of whether an admin has set `TRINITY_SCRATCH_DIR`.
We should **never** introduce these tool specific variables, this should've just used $TMPDIR and then the admin / tpv can set this to special values if needed.
The upside is that this now gets tested.

FOR CONTRIBUTOR:


natefoo:
> We should never introduce these tool specific variables, this should've just used $TMPDIR and then the admin / tpv can set this to special values if needed.

Agreed, but this was much more tedious before TPV since it meant another destination or some custom implementation in a dynamic rule.

natefoo (approved):

PR #5493, opened 2023-09-21
Fix tmp dir for picard

mvdbeek:
The TMP_DIR option doesn't do anything (anymore ?), and so java for whatever reason just decides to use /tmp instead of whatever is configured on the host via $TMPDIR.

Since we're not mounting in the hosts' tmpdir anymore (good!) we may not have a writable /tmp, so that fails (not so good ... we should possibly mount /tmp if none of the other TMP variables are /tmp).

FOR CONTRIBUTOR:


mvdbeek:
@bernt-matthias do you know why this is a problem ?

```
WARNING: Conditional [read_group_lb_conditional] first param of type="boolean" is discouraged, use a select
```

is it cause you can't add a third option ? It seems relatively silly to turn this into a select if it's clearly boolean.

bgruening:
This is an IUC guideline: https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#booleans

bgruening (approved):

mvdbeek:
I see, it's specifically

> Options that are conditionally hidden (using the <conditional> element) should use a select param type and not a boolean. The user may not expect a boolean checkbox to change the content of a form.

PR #5512, opened 2023-09-30
Update EdgeR: Provide optional formula and contrasts file for multiple contrasts

pcm32:
FOR CONTRIBUTOR:

This PR makes EdgeR fully workflow compatible by adding a contrasts file option. It also adds an optional formula field for more complex modelling.

### DIFF
--- a/tools/edger/edger.xml
+++ b/tools/edger/edger.xml
@@ -72,10 +72,10 @@ Rscript '$__tool_directory__/edger.R'
     -F '$formula'
 #end if
 
-#if $cinfo:
-    -C '$cinfo'
+#if $contrasts.contrastOpt == 'file':
+    -C '$contrasts.cinfo'
 #else: 
-    -C '${ ','.join( ['%s' % $x.contrast for x in $rep_contrast] ) }'
+    -C '${ ','.join( ['%s' % $x.contrast for x in $contrasts.rep_contrast] ) }'
 #end if
 
 #if $f.filt.filt_select == 'yes':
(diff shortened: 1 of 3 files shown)


pcm32:
Unfortunately the validator on the contrast's repeat will force a conditional I guess...

pcm32:
@bgruening is it acceptable to rely on the user to remove any previous repetition of the contrasts repeat when using the file? I prefer this solution since this means that the topology doesn't change, and whomever might be using this on their pipelines won't see them broken when updating to this newer version.

tools/edger/edger.xml
```diff
@@ -176,13 +184,20 @@ cp '$outReport.files_path'/*.tsv output_dir/
             </when>
             <when value="no"/>
         </conditional>
+        <!-- Optional formula -->
+        <param name="formula" type="text" optional="true" label="Formula for linear model" help="An optional formula for the EdgeR linear model, this will override the use of the fields in factors as a simple sum. The formula can only use elements available in the factors file. This needs to be exactly as EdgeR expect the formula, ie. `~ 0 + factor_A + factor_B:factor_C`. See EdgeR documentation for more details.">
+            <sanitizer>
+                <valid initial="string.printable"/>
```
bgruening, line 190: printable looks dangerous. You could escape the `-F '      '`

bgruening:
> @bgruening is it acceptable to rely on the user to remove any previous repetition of the contrasts repeat when using the file? I prefer this solution since this means that the topology doesn't change, and whomever might be using this on their pipelines won't see them broken when updating to this newer version.

You mean to keep the min=1 as before?
Why are you not putting both in a conditional? Use contrasts or formulas. Or is this not working?

pcm32:
It is mostly to avoid breaking people's workflows when they change the tool version. In the current version, there is no conditional there, so the repeat is directly exposed to the workflow, on the new version with a conditional before the repeat, I guess this would break connections. But if you prefer with a conditional at the expense of that, I can do it like that.

tools/edger/edger.xml
pcm32: What is the most lenient we can go without being injection dangerous. I would like to be able to express formulas of the type `~ 0 + factor_a:factor_b + factor_b + factor_a` ideally at least.

pcm32: Would something like this be acceptable:

https://github.com/galaxyproject/tools-iuc/blob/dcaf8046840f163143075b276dd75909d344ec3a/tools/ancombc/macros.xml#L12 ?

pcm32: Ping @bgruening :-)

pcm32:
@bgruening do you feel that the current state satisfies your comments? Thanks for taking the time to look into this.

tools/edger/edger.xml
```diff
@@ -176,13 +184,35 @@ cp '$outReport.files_path'/*.tsv output_dir/
             </when>
             <when value="no"/>
         </conditional>
+        <!-- Optional formula -->
+        <param name="formula" type="text" optional="true" label="Formula for linear model" help="An optional formula for the EdgeR linear model, this will override the use of the fields in factors as a simple sum. The formula can only use elements available in the factors file. This needs to be exactly as EdgeR expect the formula, ie. `~ 0 + factor_A + factor_B:factor_C`. See EdgeR documentation for more details.">
+            <sanitizer>
+                <valid initial="string.printable">
+                    <remove value="&apos;"/>
```
bgruening, line 191: @hexylena maybe you can look at this as well.

I feel like removing `\` would be good as well.

bgruening (approved):
Yes, I think this is good to go. I ping Helena, she is way better at estimating security concerns.

tools/edger/edger.xml
hexylena: I agree, good to flag that, it's very unlikely to be used in a factor expression in my experience. I'd imagine `[a-zA-Z0-9_+@$.: -]` is probably a better expression, you shouldn't need `!`, `.$@` are probably useful, `<>` probably not. It's definitely compounded by how complex R permits variable names to be. 

string.printable includes a lot

```
'0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ!"#$%&\'()*+,-./:;<=>?@[\\]^_`{|}~ \t\n\r\x0b\x0c'
```

hexylena: But if you want to go the restriction rather than explicitly allowing, to avoid user support issues, then probably don't need the advanced whitespace (\t\n\r...) and yea ' \ are definitely not necessary, everything else should just generate R errors. I'm not seeing the formula be eval'd anywhere 

(but i've never done an R CTF challenge so, maybe there's techniques I'm not aware of)

pcm32: I have also seen people doing things like `~ log(factor1) + factor2:factor3` as well, so calling arbitrary functions in the models.

hexylena: @pcm32 Ahh ok, then we keep `()`, I hadn't seen `log` before, that's good to know. Then maybe we also need `*/+-`

I guess it should look something like:

```xml
<sanitizer invalid_char="">
  <valid initial="string.letters,string.digits">
    <add value="_" />
    <add value="-" />
    <add value="+" />
    <add value="*" />
    <add value=":" />
    <add value=" " />
    <add value="(" />
    <add value=")" />
    <add value="@" />
    <add value="$" />
  </valid>
</sanitizer>
```

pcm32:
Thanks @bgruening and @hexylena for the review and suggestions, I have added the new sanitizer scheme suggested by Helena, plus a couple of additional characters (~, / for divisions and ^ I think). Hopefully this can be merged then.

hexylena (approved):

PR #5530, opened 2023-10-20
coverm: fix usage of cached reference genomes

bernt-matthias:
when iterating over dynamic selects one can not access
the fields property, but name and path column need to
be set manually for the select options, xref https://github.com/galaxyproject/galaxy/pull/16885

also adapts a test to check if code works

this never worked since there is obviously no element_identifier
attribute even if `.fields.path` would exist.

FOR CONTRIBUTOR:


bgruening (approved):
@bebatut you want to look about this one?

PR #5531, opened 2023-10-20
maf_stats: remove unused parameter

bernt-matthias:
No idea what the tool does (the generated command line and the python script are super confusing) .. but species is not used.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/maf_stats/maf_stats.xml
+++ b/tools/maf_stats/maf_stats.xml
@@ -109,7 +109,7 @@ Alternatively, you can request only summary information for a set of intervals:
     panTro2      30390      0.2353
     ========  ===========  ========
 
-where **coverage** is the number of nucleotides divided by the total length of the provider intervals.
+where **coverage** is the number of nucleotides divided by the total length of the intervals provided in the BED file.
     </help>
     <citations>
         <citation type="doi">10.1093/bioinformatics/btr398</citation>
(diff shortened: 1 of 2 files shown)


tools/maf_stats/maf_stats.xml
```diff
@@ -87,40 +76,40 @@
...
+    ========  ===========  ========
+    #species  nucleotides  coverage
+    ========  ===========  ========
+    hg18         30639      0.2372
+    rheMac2      7524       0.0582
+    panTro2      30390      0.2353
+    ========  ===========  ========
 
-        where **coverage** is the number of nucleotides divided by the total length of the provided intervals.
+where **coverage** is the number of nucleotides divided by the total length of the provider intervals.
```
bernt-matthias, line 112: I think provided was correct. Maybe more explicit like so:

```suggestion
where **coverage** is the number of nucleotides divided by the total length of the intervals provided in the BED file.
```

bgruening:
@nekrut @martenson do you remember what this tool was supposed to do? And why the species filter is not used?

I guess it makes sense to filter species from a huge MAF file.

bernt-matthias:
Guess this was intended to work similar to https://github.com/galaxyproject/tools-iuc/blob/main/tools/interval2maf/interval2maf.py .. 

Not sure how to proceed here. I have no time to implement it (an given the IMO horrible code I also would rather not touch it), but leaving an unused parameter seems also like no option.

bgruening:
Agree. Happy to merge when @nekrut or @martenson don't object or don't respond:-)

bgruening (approved):

PR #4908, opened 2022-11-07
Updating tools/phyloseq from version 1.38.0 to 1.46.0

planemo-autoupdate:
Hello! This is an automated update of the following tool: **tools/phyloseq**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

I have updated tools/phyloseq from version 1.38.0 to 1.42.0.

**Project home page:** https://www.bioconductor.org/packages/release/bioc/html/phyloseq.html

For any comments, queries or criticism about the bot, not related to the tool being updated in this PR, please create an issue [here](https://github.com/planemo-autoupdate/autoupdate/issues/new).

PR #5532, opened 2023-10-20
update xml file and add binary

SalihaZ:
despite the validity of the planemo test, phyml does not work on all the instances of galaxy and on galaxy planemo as well.
I added the binary with the xml file.
Could you please check if it works?
Thanks in advance


bernt-matthias:
What is the actual problem?

bernt-matthias:
Dear @SalihaZ adding a binary is not an option (in my opinion). Feel free to open an issue describing the actual problem or make this an update of phyloseq wo adding a binary here (see also https://github.com/galaxyproject/tools-iuc/pull/4908).

SalihaZ:
Thanks for your reply. Actually when running phyml (on galaxy planemo and on other servers), I got the error: "command not found". I had to specify on the xml file, the path to the binary and it worked.
Is there another solution then adding the binary?

bernt-matthias:
`planemo test` on phyml works for me (see also  https://github.com/galaxyproject/tools-iuc/pull/5536 .. which proves this). Have you setup conda (or docker) correctly?

Can you share a link to the history where you observed an error on a public instance?

bernt-matthias:
https://github.com/galaxyproject/tools-iuc/issues/5535

SalihaZ:
The `planemo test` worked on phyml but not when I've run the `planemo serve`. 
I've just checked and it's now working on both planemo and Fr instance. Many thanks for you help

bernt-matthias:
In serve you may be able to install the dependencies manually in the admin interface....? 

Thanks for the update.

SalihaZ:
I've been using Galaxy for not very long. Yes indeed, I just discovered that it is possible to manage dependencies on the admin interface.

PR #5533, opened 2023-10-20
update ribowaltz with an extra option

pavanvidem:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/ribowaltz/ribowaltz.R
+++ b/tools/ribowaltz/ribowaltz.R
@@ -21,6 +21,7 @@ spec <- matrix(c(
   "cds_coverage_info", "Z", 1, "character",
   "psite_info_rdata", "O", 0, "character",
   "refseq_sep", "s", 0, "character",
+  "indel_threshold", "t", 0, "integer",
   "params_duplicate_filterting", "d", 0, "character",
   "params_peridiocity_filterting", "l", 0, "character",
   "params_custom_filterting", "c", 0, "character",
@@ -48,7 +49,7 @@ if (opt$refseq_sep == "") {
   sep <- NULL
 }
 # convert alignments in BAM files into list of data tables
-reads_list <- bamtolist(bamfolder = opt$bamdir, annotation = annotation_dt, refseq_sep = sep)
+reads_list <- bamtolist(bamfolder = opt$bamdir, annotation = annotation_dt, refseq_sep = sep, indel_threshold=opt$indel_threshold)
 
 library("jsonlite")
 # remove duplicate reads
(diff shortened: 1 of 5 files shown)


pavanvidem:
Adding an extra param. No new tests are needed.

bgruening:
Can you please fix the lining?

bgruening:
Still one r lint error.

tools/ribowaltz/ribowaltz.R
```diff
@@ -21,6 +21,7 @@ spec <- matrix(c(
   "cds_coverage_info", "Z", 1, "character",
   "psite_info_rdata", "O", 0, "character",
   "refseq_sep", "s", 0, "character",
+  "indel_threshold", "t", 0, "integer",
```
bernt-matthias, line 24: the 3rd column is the default or? Should be 5 then...

bernt-matthias (commented):
> Adding an extra param. No new tests are needed.

Why? :)

tools/ribowaltz/ribowaltz.R
pavanvidem: I think it means is optional or required option.

bernt-matthias: Right, but according to https://cran.r-project.org/web/packages/getopt/getopt.pdf

```
Column 3: Argument mask of the flag. An integer. Possible values: 0=no argu-
ment, 1=required argument, 2=optional argument.
```

this should be 1 or 2. Maybe you can fix this across the matrix?

Also, this highlights the need for tests! We should make sure that the argument has an effect if set to a non-default value (and we should have done this for all parameters).

I also tend to become lazy and skip tests and docs .. and have to remind myself that code that is not tested usually does not work. Also makes the life of reviewers easier .. and only the life of the CI harder.

bernt-matthias: > I think it means ...

I also often tend to make assumptions based on my intuition .. and most of the time I'm wrong :)

bernt-matthias: Maybe print the content of `opt` in the R code (maybe to stderr) and check if the parameters are set correctly by using an `assert_stderr`.

pavanvidem: You're correct. I initially copied it from DESeq2 wrapper which I thought was well tested. So now we have to fix this in many other R scripts too.

bernt-matthias: Actually, I'm wrong https://github.com/trevorld/r-getopt/issues/3 .. LOL 

So all good :)

bgruening (approved):
I guess a test can not hurt, but looks good to me.

bgruening (approved):

PR #5550, opened 2023-10-27
FastK: adding new tool

abueg:
FOR CONTRIBUTOR:

Hello! 👋🏼 

This PR adds the k-mer counting/manipulation tool [FastK](https://github.com/thegenemyers/FASTK) as a galaxy tool. There 3 tests using different sets of parameters 👍🏼 

The license permits use as long as the license text is reproduced in the binary, which I believe the conda package respects, here is the license: https://github.com/thegenemyers/FASTK/blob/master/LICENSE

Thank you @astrovsky01 for all your help with this!! 🙏🏼


bernt-matthias:
`.shed.yml` is missing.

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,115 @@
+<tool id="fastk" name="FastK" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="20.01">
+    <description>A k-mer counter for high-quality assembly datasets.</description>
+    <macros>
+        <token name="@TOOL_VERSION@">1.0</token>
+        <token name="@VERSION_SUFFIX@">0</token>
+    </macros>
+    <requirements>
+        <requirement type="package" version="1.0.0">fastk</requirement>
```
bernt-matthias, line 8: Use `@TOOL_VERSION@` here.

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,115 @@
...
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        mkdir -p outfiles/tmpfiles &&
```
bernt-matthias, line 11: I guess you can also run without creating `outfiles`.

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,115 @@
...
+        ln -s '$infile' input.fasta &&
+        FastK input.fasta
+        #set kmer_size=$kmer_size
+        -k$kmer_size 
+        #if $sorted_table.sorted_table_presence == 'yes':
+            #if $sorted_table.type_sorted_table.sorted_table_options == 'default_sorted_table':
+                -t 
+            #elif $sorted_table.type_sorted_table.sorted_table_options == 'cutoff_sorted_table':
+                #set sorted_table_cutoff=$sorted_table.type_sorted_table.sorted_table_cutoff
+                -t$sorted_table_cutoff
```
bernt-matthias, line 22: No space after `-t`?

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,115 @@
...
+        <token name="@TOOL_VERSION@">1.0</token>
+        <token name="@VERSION_SUFFIX@">0</token>
+    </macros>
+    <requirements>
+        <requirement type="package" version="1.0.0">fastk</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        mkdir -p outfiles/tmpfiles && 
+        cd outfiles &&
+        ln -s '$infile' input.fasta &&
```
bernt-matthias, line 13: Infile can be also other formats than fasta.

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,115 @@
...
+                    </when>
+                </conditional>
+                <section title="Advanced" name="advanced">
+                    <param name="tabex_threshold" label="Tabex count threshold" type="integer" value="5" min="1"/>
+                </section>
+            </when>
+        </conditional>
+    </inputs>
+    <outputs>
+        <data name="fastk_out" format="tar" from_work_dir="outfiles/fastk.tar" label="${tool.name} on ${on_string}: FastK hist files"/>
```
bernt-matthias, line 61: Maybe a collection would be better than a tar?

bernt-matthias (commented):
Some first comments.

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,115 @@
...
+                -t$sorted_table_cutoff 
+            #end if
+        #end if
+        -T\${GALAXY_SLOTS:-1} -Noutput -Ptmpfiles
+        #if $sorted_table.sorted_table_presence == 'yes':
+            && Tabex output.ktab -t${sorted_table.advanced.tabex_threshold} LIST > tabex.txt
+            && mv tabex.txt ..
+        #end if
+        && mv input.fasta ..
+        && ls -lah
```
bgruening, line 31:
```suggestion
```

tools/fastk/fastk.xml
abueg: no, there is no space after the parameters for this tool

tools/fastk/fastk.xml
abueg: the output files won't be needed/used on their own, only ever as a folder with those files inside them, so that was the rationale for tar

tools/fastk/fastk.xml
abueg: true, that was my short-sightedness from just using this with FASTAs, will add the other supported formats as described in the program's readme 👍 thank you!

bgruening:
The linting fails because of a missing citation. You can add a bibtex and just cite the GitHub repo if there is no proper citation yet.

The input fasta file is unfortunately too big, can you reduce the size?

bernt-matthias:
> The input fasta file is unfortunately too big, can you reduce the size?

If its available remotely we can use `location="URL"` .. (CI will work as soon as we have a new planemo version with https://github.com/galaxyproject/planemo/pull/1388)

tools/fastk/fastk.xml
abueg: ok so what i am trying to figure out now is, the program seems to decide how to run based on the given file extension, so i gzipped the test data to try to make it smaller, and then i had to change the CMD to `fasta.gz` so that it would properly run on that. i'm going to mark this PR as draft for now

abueg:
added an if/elif part to the start of the tool for input file extension detection, as the tool relies on the input file's extension to determine how to run. this made some of the later parts messy so i moved the tool out of `outfiles` as a wd and am working outside it instead. this current version is passing the tests locally and i will work on adding the other file types into the CMD now 👍🏼 (ty @bgruening for the `infile.ext` tip!) 

PS can i use a `case` statement in the CMD insead of if/elif/elif/elif?

tools/fastk/fastk.xml
bernt-matthias: I understand. But then we have one optional output `tabex_hist` and one output that "won't be needed/used ...". So what is supposed to be used as output?

abueg:
> Still archives may also include date/time which might cause the difference.

is there a preferred way of handling this, like setting an arbitrary `--mtime` or? sorry, we were modelling this off of what the [meryl](https://github.com/galaxyproject/tools-iuc/blob/main/tools/meryl/meryl.xml) tool does since it works similarly in that its output is a folder and then the tool operates on the whole folder...

and thank you for the tip about `planemo test` 🙏🏼

tools/fastk/fastk.xml
abueg: oh, sorry about the confusion -- when k-mer counting, there should always be an output of the `output.hist` file. optionally, if the user specifies the `-k` option, then the outputs will include `output.ktab` and a number of hidden `.output.ktab.n` files, with `n` corresponding to the number of threads given to the run. these hidden files are needed to be in the same folder as the non-hidden `output.ktab` file in order for subsequent commands (e.g., `tabex` or `histex`) to work. 

so to sum, there will always be at `output.hist` as an output, and optionally other outputs (`.output.ktab.n` files) that might be produced, and can't be used on their own

bernt-matthias:
> is there a preferred way of handling this

With `assert_contents` you can express quite a lot assumptions on archive members and their contents: https://docs.galaxyproject.org/en/master/dev/schema.html#has-archive-member

tools/fastk/fastk.xml
bernt-matthias: Still I don't see if there is always an output of the Galaxy tool that can be used (within Galaxy).

bgruening:
its green :)

Do you want to take this out of Draft?

tools/fastk/fastk.xml
```diff
@@ -100,7 +100,7 @@
             <param name="infile" value="input01.fasta.gz"/>
             <output name="fastk_out" ftype="tar">
                 <assert_contents>
-                    <has_size value="266240" delta="1000" />
+                    <has_archive_member path="./outfiles/output.hist" />
```
bernt-matthias, line 103: With `<has_archive_member path=".*" n="10"/>` you could also add an assumption on the number of files in the archive, which would be neat. 

Additionally you could also make assumptions on the content by including more assumptions, like so:

```
<has_archive_member path="./outfiles/output.hist">
  <not_has_text text="EDK72998.1"/>
</has_archive_member>
```

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,148 @@
+<tool id="fastk" name="FastK" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="20.01">
+    <description>A k-mer counter for high-quality assembly datasets.</description>
```
bernt-matthias, line 2:
```suggestion
    <description>A k-mer counter for high-quality assembly datasets</description>
```

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,148 @@
...
+            #elif $sorted_table.type_sorted_table.sorted_table_options == 'cutoff_sorted_table':
+                #set sorted_table_cutoff=$sorted_table.type_sorted_table.sorted_table_cutoff
+                -t$sorted_table_cutoff 
+            #end if
+        #end if
+        -T\${GALAXY_SLOTS:-1} -Noutfiles/output -Poutfiles/tmpfiles
+        #if $sorted_table.sorted_table_presence == 'yes':
+            && Tabex outfiles/output.ktab -t${sorted_table.advanced.tabex_threshold} LIST > tabex.txt
+        #end if
+        && tar -c -f fastk.tar ./outfiles/
```
bernt-matthias, line 63: Will the `tar` also include `tmpfiles/`? How about the hidden files that you mentioned. Is this intended?

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,148 @@
...
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        mkdir -p outfiles/tmpfiles && 
+        #if $infile.ext == "fasta":
+            ln -s '$infile' ./input.fasta &&
+            #set INPUTFILE="input.fasta"
+        #elif $infile.ext == "fasta.gz":
+            ln -s '$infile' ./input.fasta.gz &&
+            #set INPUTFILE="input.fasta.gz"
+        #elif $infile.ext == "fastq":
```
bernt-matthias, line 32: This should use is `$infile.is_of_type("fastq")`. Since `fastq` has derived types like `fastqsanger` (in which case `ext` would be `fastqsanger`).

Analogous for `fastq.gz` and `unsorted.bam`.

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,148 @@
...
+            #end if
+        #end if
+        -T\${GALAXY_SLOTS:-1} -Noutfiles/output -Poutfiles/tmpfiles
+        #if $sorted_table.sorted_table_presence == 'yes':
+            && Tabex outfiles/output.ktab -t${sorted_table.advanced.tabex_threshold} LIST > tabex.txt
+        #end if
+        && tar -c -f fastk.tar ./outfiles/
+    ]]></command>
+    <inputs>
+        <param name="infile" type="data" format="fasta,fasta.gz,fastq,fastq.gz,cram,bam,sam" label="Input file" />
```
bernt-matthias, line 66: Use `unsorted.bam` instead of `bam`. In Galaxy `bam` means a coordinate sorted BAM file. `unsorted.bam` makes no assumptions on the sorting.

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,148 @@
...
+        <param name="kmer_size" type="integer" min="5" max="50" value="40" label="K-mer size" help="Default k-mer size is 40." />
+        <conditional name="sorted_table">
+            <param name="sorted_table_presence" type="select" label="Sorted table selection" help="Do you want a sorted table of all canonical k-mers and their counts?">
+                <option value="no">No</option>
```
bernt-matthias, line 70: Still wondering: What is the use of this tools if the user selects `No`? The tool will only produce a tar, which can't be used in Galaxy, isn't it?

Instead of yes/no, shouldn't the user select from at least one of the tools fastk provides:

- [Histex](https://github.com/thegenemyers/FASTK#histex): Display a FastK histogram or convert to 1-code
- [Tabex](https://github.com/thegenemyers/FASTK#tabex): List, Check, find a k‑mer in a FastK table, or convert to 1-code
- [Profex](https://github.com/thegenemyers/FASTK#profex): Display a FastK profile or convert to 1-code
- [Logex](https://github.com/thegenemyers/FASTK#logex): Combine kmer,count tables with logical expressions & filter with count cutoffs
- [Symmex](https://github.com/thegenemyers/FASTK#symmex): Produce a symmetric k-mer table from a canonical one

If you agree, but don't have the time or need for all of them then maybe prepare the tool such that it can be easily extended in this way?

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,148 @@
...
+                <conditional name="type_sorted_table">
+                    <param name="sorted_table_options" type="select" label="Sorted table presence" help="Do you want to specify a cut-off?">
+                        <option value="default_sorted_table">default (1)</option>
+                        <option value="cutoff_sorted_table">specify cutoff</option>
+                    </param>
+                    <when value="default_sorted_table"/>
+                    <when value="cutoff_sorted_table">
+                        <param name="sorted_table_cutoff" type="integer" min="2" value="10" label="Sorted table cutoff value"/>
+                    </when>
+                </conditional>
```
bernt-matthias, line 84: Instead of this, maybe:

```suggestion
           <param name="sorted_table_cutoff" type="integer" min="2" optional="true" label="Sorted table cutoff value"/>
```

Plus the necessary change in the command section (ie check if a value is given).

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,148 @@
...
+            <when value="no"/>
+            <when value="yes">
+                <conditional name="type_sorted_table">
+                    <param name="sorted_table_options" type="select" label="Sorted table presence" help="Do you want to specify a cut-off?">
+                        <option value="default_sorted_table">default (1)</option>
+                        <option value="cutoff_sorted_table">specify cutoff</option>
+                    </param>
+                    <when value="default_sorted_table"/>
+                    <when value="cutoff_sorted_table">
+                        <param name="sorted_table_cutoff" type="integer" min="2" value="10" label="Sorted table cutoff value"/>
```
bernt-matthias, line 82: The user does not know what the `Sorted table` is. Needs explanation.

How does this interact with `tabex_threshold`?

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,148 @@
...
+                    journal = {GitHub repository},                                               
+                    url = {https://github.com/thegenemyers/FASTK},                        
+                    }                                                                    
+                </citation>
+            </citations>
+        </xml>
+    </macros>    
+    <requirements>
+        <requirement type="package" version="@TOOL_VERSION@">fastk</requirement>
+    </requirements>
```
bernt-matthias, line 23:
```suggestion
    <macros>
        <import>macros.xml</import>
    </macros>
    <expand macro="requirements" />
```

tools/fastk/macros.xml
```diff
@@ -0,0 +1,24 @@
+<macros>
+    <token name="@TOOL_VERSION@">1.0.0</token>
+    <token name="@GALAXY_TOOL_VERSION@">galaxy</token>
```
bernt-matthias, line 3:
```suggestion
```

tools/fastk/fastk.xml
abueg: the tar is intended to include the hidden files, correct 👍🏼

tools/fastk/fastk.xml
abueg: i considered adding # of files, but for the test involving sorted tables, the number of files generated can change because the number of hidden files made is dependent on how many threads is given to the program -- so it's dependent on `$GALAXY_SLOTS`... though i guess for the test i can specify `-T1` for one thread so it's consistent? 

for content assumption, the `hist` and `ktab` files are binaries, so i didn't think the `has_text`/`not_has_text` assumptions would work, but please correct me if wrong!

abueg:
> its green :)
> 
> Do you want to take this out of Draft?

yay green! 🟢 i think i will still keep in draft for now as it seems there is still some work to do? my apologies for that! 😭 does keeping it in draft prevent unnecessary tests being run? if not then i guess i will take it out of draft if it doesn't change much either way

tools/fastk/fastk.xml
abueg: would i put a placeholder in the other option `elif`s? or just have `fastk` and `tabex` implemented?

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,168 @@
...
+            <when value="Logex">
+            </when>
+            <when value="Symmex">
+            </when>
+        </conditional>
+    </inputs>
+    <outputs>
+        <data name="fastk_out" format="tar" from_work_dir="fastk.tar" label="${tool.name} on ${on_string}: FastK hist files"/>
+        <data name="tabex_hist" format="txt" from_work_dir="tabex.txt" label="${tool.name} on ${on_string}: Tabex output">
+            <filter>sorted_table['sorted_table_presence'] == "yes"</filter>
```
abueg, line 111: ah i need to change this because i added the `operation_type`/`command_type` layer!

tools/fastk/fastk.xml
astrovsky01: @bernt-matthias each of the sub-tools for fastk use the fastk tar as an input. The original intention was to have this be a suite where you would run fastk and the output file would be useable by any of the other tools without needing to rerun fastk itself. We can add a separate output for the .hist files aside from the tar file, but the intent was to standardize an input for the histex, tabex, etc. tools to use

tools/fastk/fastk.xml
astrovsky01: Several of the subsequent tools take an input that ran without the sorted table cutoff flag, which is why we made it optional

bernt-matthias:
> my apologies for that!

Nothing to apologize for! Thanks for your work.

abueg:
hi @bernt-matthias & @bgruening !! i've put together a flowchart of inputs/outputs of the tools in this suite and how they work within (and occasionally without) -- sorry about the delay!

i hope this is helpful for figuring out how to proceed with wrapping the tool(s)? the gist is that FastK and Logex will produce binary files that are only readable within this suite (and the MerquryFK tool), but Tabex and Histex have human-readable output. Histex's output additionally can be read into GenomeScope 2.0, which is already wrapped (so we might not need to wrap GeneScopeFK, as it operates on the same output as GenomeScope 2.0 afaik)... 

image

(there are additional commands available from the FastK suite, but I do not have the need for them at the moment)

bernt-matthias:
So, it appears to me that it would be good to have the intermediate binary files as output. Then we can reuse them as input for more than one program. Also gives options in terms of scalability.

For this, we would need new datatypes. For `ktab` one may subclass from the new directory datatype (?) alternatively a composite datatype might work (but difficult to use in tests). In the simplest case we can just add them to the `datatypes.xml.sample` file. But, this may be realized earliest with the next Galaxy release. If we can't wait we could just use binary/directory .. for now.

mvdbeek:
> But, this may be realized earliest with the next Galaxy release.

Why ? We've always supported adding datatypes to the current stable release.

bernt-matthias:
> Why ? We've always supported adding datatypes to the current stable release.

Cool. I repeatedly forget :)

abueg:
im still alive i promise

tools/fastk/fastk.xml
abueg: added

bgruening:
@abueg is this ready for review?

abueg:
@bgruening I think it is ok to review now, last commits were fixing some issues with testing! I  might add more things in a couple weeks, but I think this should work right now for the k-mer counting & hist generation!!

tools/fastk/fastk.xml
abueg: edited!

tools/fastk/fastk.xml
abueg: edited!

tools/fastk/fastk.xml
abueg: i think the `outfiles` dir makes tar'ing the files down the line easier, because i just `tar -c -f fastk.tar ./outfiles/` ?

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,169 @@
...
+                <option value="FastK">FastK: count k-mers</option>
+                <option value="Histex">Histex: display a FastK histogram</option>
+                <option value="Tabex">Tabex: list, check, or find a k-mer in a FastK table</option>
+                <option value="Profex">Profex: display a FastK profile</option>
+                <option value="Logex">Logex: combine k-mer,count tables with logical expressions and filter with count cutoffs</option>
+                <option value="Symmex">Symmex: produce a symmetric k-mer table from a canonical one</option>
+            </param>
+            <when value="FastK">
+                <param name="kmer_size" type="integer" min="5" max="50" value="40" label="K-mer size" help="Default k-mer size is 40." />
+                <conditional name="sorted_table">
```
bernt-matthias, line 73: Do you think it would be a good idea to flatten the nested conditionals and just have one with 3 options:

- No
- Yes with default threshold
- Yes with manual threshold

?

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,169 @@
...
+            </when>
+            <when value="Logex">
+            </when>
+            <when value="Symmex">
+            </when>
+        </conditional>
+    </inputs>
+    <outputs>
+        <data name="fastk_out" format="tar" from_work_dir="fastk.tar" label="${tool.name} on ${on_string}: FastK files"/>
+        <data name="fastk_hist" format="binary" from_work_dir="outfiles/output.hist" label="${tool.name} on ${on_string}: FastK hist" />
```
bernt-matthias, line 110: Those two outputs will also need a filter, or?

tools/fastk/fastk.xml
bernt-matthias: > though i guess for the test i can specify -T1

I would suggest to avoid this

> for content assumption, the hist and ktab files are binaries, so i didn't think the has_text/not_has_text assumptions would work, but please correct me if wrong!

Indeed. Then maybe `has_size` is your friend :)

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,169 @@
...
+                                <param name="sorted_table_cutoff" type="integer" min="2" value="10" label="Sorted table cutoff value"/>
+                            </when>
+                        </conditional>
+                        <section title="Advanced" name="advanced">
+                            <param name="tabex_threshold" label="Tabex count threshold" type="integer" value="5" min="1"/>
+                        </section>
+                    </when>
+                </conditional>
+            </when>
+            <when value="Histex">
```
bernt-matthias, line 96: Do you plan to implement this as a single tool?

tools/fastk/test-data/test02.tabex.txt
```diff
@@ -0,0 +1,311 @@
+Opening 40-mer table with 502,597 entries
```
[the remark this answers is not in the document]
abueg: it's part of the output from the `Tabex output.ktab LIST > output.ktab.txt` part, to have a human-readable output because the `output.ktab` file is binary
image

tools/fastk/fastk.xml
abueg: added `has_size` assertions!

tools/fastk/fastk.xml
abueg: sorry, do you mean implement `Histex` as a single tool, or implement all these others as a single tool with FastK?

tools/fastk/fastk.xml
abueg: ah you mean a filter for if `FastK` was the selected tool?

tools/fastk/fastk.xml
bernt-matthias: I was just wondering. Both would be fine.

bgruening: We can maybe help with splitting the tools later. Lets go with what is here for now.

tools/fastk/fastk.xml
bgruening: I like that idea. @abueg does this make sense?

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,172 @@
...
+    <inputs>
+        <param name="infile" type="data" format="fasta,fasta.gz,fastq,fastq.gz,cram,unsorted.bam,sam" label="Input file" />
+        <conditional name="operation_type">
+            <param name="command_type" type="select" label="Operation type selector" help="Select a type of operation">
+                <option value="FastK">FastK: count k-mers</option>
+                <option value="Histex">Histex: display a FastK histogram</option>
+                <option value="Tabex">Tabex: list, check, or find a k-mer in a FastK table</option>
+                <option value="Profex">Profex: display a FastK profile</option>
+                <option value="Logex">Logex: combine k-mer,count tables with logical expressions and filter with count cutoffs</option>
+                <option value="Symmex">Symmex: produce a symmetric k-mer table from a canonical one</option>
```
bgruening, line 69:
```suggestion
                <!--option value="Histex">Histex: display a FastK histogram</option>
                <option value="Tabex">Tabex: list, check, or find a k-mer in a FastK table</option>
                <option value="Profex">Profex: display a FastK profile</option>
                <option value="Logex">Logex: combine k-mer,count tables with logical expressions and filter with count cutoffs</option>
                <option value="Symmex">Symmex: produce a symmetric k-mer table from a canonical one</option-->
```

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,172 @@
...
+            <when value="Histex">
+            </when>
+            <when value="Tabex">
+            </when>
+            <when value="Profex">
+            </when>
+            <when value="Logex">
+            </when>
+            <when value="Symmex">
+            </when>
```
bgruening, line 105:
```suggestion
            <!--when value="Histex">
            </when>
            <when value="Tabex">
            </when>
            <when value="Profex">
            </when>
            <when value="Logex">
            </when>
            <when value="Symmex">
            </when-->
```

bgruening (commented):
@abueg if we comment out the other functions it is ready to go.

bernt-matthias (commented):
I would suggest to make this a suite and use one tool per subcommand. The change should be equally simple compared to the suggestions made by @bgruening

This make scheduling much easier (guess the steps need different amounts of resources).

bernt-matthias:
Guess this will be superseeded by https://github.com/galaxyproject/tools-iuc/pull/5965

abueg:
ACK sorry yes @SaimMomin12 has very kindly taken over the wrappers for this suite!!

abueg:
closed as, has been mentioned, is superseded by this PR: https://github.com/galaxyproject/tools-iuc/pull/5965

Issue #5099, opened 2023-01-26
cutadapt: paired parameter inconsistency


bernt-matthias:
- `Optional separate quality cutoff for Read 2` is show only for paired end input (which makes sense)
- `Specify a minimum/maximum length for reverse reads (R2)` (and included params) is in the `Filter Options` section (irrespective of the input being paired or not)

Should be unified.

PR #5577, opened 2023-11-03
cutadapt improvements

bernt-matthias:
fixes https://github.com/galaxyproject/tools-iuc/issues/5099 

While working on this I have noted some more room for improvement. special attention is needed here IMO:

- the `internal` parameter has been dropped. while the implementation was probably not good anyway (the set value applied to all adapters which was likely not intended), the current implementation does not allow adding the `X` for builtin adapters.

Maybe I find the time to add some missing parameters ... 

FOR CONTRIBUTOR:


tools/cutadapt/macros.xml
```diff
@@ -1,7 +1,7 @@
 <macros>
-    <token name="@TOOL_VERSION@">4.4</token>
+    <token name="@TOOL_VERSION@">4.5</token>
     <token name="@VERSION_SUFFIX@">0</token>
-    <token name="@FASTQ_TYPES@">fastq.gz,fastq,fasta</token>
+    <token name="@FASTQ_TYPES@">fastqsanger.bz2,fastqsanger.gz,fastqsanger,fasta</token>
```
mvdbeek, line 4: Are you sure that cutadapt doesn't rewrite quality scores ?

bernt-matthias: I'm about to add a fix for this which will append `--quality-base=64` to the CLI for `fastaillumina` and `fastqsolexa`.

Seems that Illumina input gives Illumina output. There is a [test case in the cutadapt sources](https://github.com/marcelm/cutadapt/blob/e04cc32e392e1cbe0c518b4e0637cdf03533d440/tests/test_commandline.py#L252). Test data [before](https://github.com/marcelm/cutadapt/blob/v4.5/tests/data/illumina64.fastq) and [after](https://github.com/marcelm/cutadapt/blob/v4.5/tests/cut/illumina64.fastq). 

I will add these tests here using the new `location` attribute .. which would need https://github.com/galaxyproject/planemo/pull/1388 to pass here :) I just don't understand the problem there.

mvdbeek: Please don't use location in the IUC, especially if the input file is not large.

bernt-matthias: > Please don't use location in the IUC, especially if the input file is not large or it is easy to downsample.

Why is that? Here it becomes particular handy because I could even use `@TOOL_VERSION@` in the link :)

mvdbeek: Why expose ourselves to disappearing urls for no reason ?

bgruening (approved):

bernt-matthias:
Do you want to have another look on this @mvdbeek ?

tools/cutadapt/cutadapt.xml
```diff
@@ -39,15 +39,18 @@
...
 #end if
 
-#if $input_1.is_of_type("fastq.gz", "fastqsanger.gz"):
-    #set ext = ".fq.gz"
-#else if $input_1.is_of_type("fastq.bz2", "fastqsanger.bz2"):
-    #set ext = ".fq.bz2"
-#else if $input_1.is_of_type('fasta'):
-    #set ext = ".fa"
-#else:
+#if $input_1.ext.startswith("fastq"):
```
mvdbeek, line 42: This prevents adding any hypothetical fastq datatype that can't be processed by cutadapt. I think we'd be better off with the `is_of_type` check, which you can still combine with the `.endswith` check below ?

mvdbeek (approved):

tools/cutadapt/cutadapt.xml
bernt-matthias:
```suggestion
#if $input_1.ext.is_of_type("fastq"):
```

mvdbeek:
thanks a lot @bernt-matthias!

lldelisle:
Hi guys,
I've run this tool on my old instance (22.01) and it failed because the `cut` argument has moved from the section `library.r1` to the section `read_mod_options` but the command line uses `str($cut)` so my instance says:
```
  File "/data/galaxy/galaxy/server/lib/galaxy/jobs/runners/__init__.py", line 248, in prepare_job
    job_wrapper.prepare()
  File "/data/galaxy/galaxy/server/lib/galaxy/jobs/__init__.py", line 1202, in prepare
    self.command_line, self.version_command_line, self.extra_filenames, self.environment_variables = tool_evaluator.build()
  File "/data/galaxy/galaxy/server/lib/galaxy/tools/evaluation.py", line 448, in build
    global_tool_logs(self._build_command_line, config_file, "Building Command Line")
  File "/data/galaxy/galaxy/server/lib/galaxy/tools/evaluation.py", line 77, in global_tool_logs
    raise e
  File "/data/galaxy/galaxy/server/lib/galaxy/tools/evaluation.py", line 73, in global_tool_logs
    return func()
  File "/data/galaxy/galaxy/server/lib/galaxy/tools/evaluation.py", line 465, in _build_command_line
    command_line = fill_template(command, context=param_dict, python_template_version=self.tool.python_template_version)
  File "/data/galaxy/galaxy/server/lib/galaxy/util/template.py", line 107, in fill_template
    raise first_exception or e
  File "/data/galaxy/galaxy/server/lib/galaxy/util/template.py", line 80, in fill_template
    return unicodify(t, log_exception=False)
  File "/data/galaxy/galaxy/server/lib/galaxy/util/__init__.py", line 1080, in unicodify
    value = str(value)
  File "/data/galaxy/galaxy/venv/lib/python3.8/site-packages/Cheetah/Template.py", line 1053, in __unicode__
  File "cheetah_DynamicallyCompiledCheetahTemplate_1716409893_2057257_87444.py", line 608, in respond
NameMapper.NotFound: cannot find 'cut'
```
Would you mind to fix this or change the profile version?
Thanks.

mvdbeek:
Which version is that happening on @lldelisle ? This seems to be fixed in the latest version at least ?

lldelisle:
Maybe I just ran an iwc rna-seq workflow and it was version 4.4+galaxy1

PR #5578, opened 2023-11-04
upgrade from SemiBin to SemiBin2

Ralwoss:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/semibin/bin.xml
+++ b/tools/semibin/bin.xml
@@ -11,7 +11,7 @@
     <command detect_errors="exit_code"><![CDATA[
 #import re
 @FASTA_FILES@
-SemiBin bin
+SemiBin2 bin
     --input-fasta 'contigs.fasta'
     --data '$data'
 #if $mod.select == 'history'
@@ -22,10 +22,13 @@ SemiBin bin
     --output 'output'
     --threads \${GALAXY_SLOTS:-1}
     --minfasta-kbs $minfasta_kbs
-    $no_recluster
+#if $mode.select == 'single' or $mode_fasta.mode_select == 'co'
+    $write_pre_reclustering_bins
+#end if
     --max-node $max_node
     --max-edges $max_edges
     --random-seed $random_seed
+    --compression none
     @MIN_LEN@
     --orf-finder '$orf_finder'
     ]]></command>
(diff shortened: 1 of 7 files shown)


tools/semibin/bin.xml
```diff
@@ -50,20 +53,20 @@ SemiBin bin
...
-        <expand macro="no-recluster"/>
+        <expand macro="write_pre_reclustering_bins"/>
     </inputs>
     <outputs>
-        <collection name="output_recluster_bins" type="list" label="${tool.name} on ${on_string}: Reconstructed bins after reclustering">
-            <filter>not no_recluster</filter>
-            <discover_datasets pattern=".*?\.(?P&lt;designation&gt;.*).fa" format="fasta" directory="output/output_recluster_bins" />
+        <collection name="output_bins" type="list" label="${tool.name} on ${on_string}: Reconstructed bins after reclustering">
+            <filter>not write_pre_reclustering_bins</filter>
+            <discover_datasets pattern="(?P&lt;designation&gt;.*).fa((.gz)|(.xz)|(.bz2))?" format="fasta" directory="output/output_bins" />
```
bgruening, line 61: can we control the compression format of the output?

The `format=` needs to change according to that.

tools/semibin/concatenate_fasta.xml
```diff
@@ -14,15 +14,17 @@
     #set $identifier = re.sub('[^\s\w\-\\.]', '_', str($e.element_identifier))
     #if $e.ext.endswith(".gz")
 gunzip -c '$e' > '${identifier}.fasta' &&
+echo $identifier
```
bgruening, line 17: remove?

tools/semibin/semibin.xml
```diff
@@ -158,49 +166,10 @@ ls output
...
-                        <has_text text=">g2k_0"/>
-                    </assert_contents>
-                </element>
-                <element name="2" ftype="fasta">
-                    <assert_contents>
-                        <has_text text=">g3k_0"/>
-                    </assert_contents>
-                </element>-->
-            </output_collection>
+            <output_collection name="output_bins" count="0"/>
```
bgruening, line 172: no output in this collection?

bgruening: If this is true can we not create this collection at all - with a filter on the outputs?

tools/semibin/bin.xml
bgruening: Can we set `    --compression none` or gz?

Ralwoss: It is not possible to change the compression, I forgot to remove that

tools/semibin/semibin.xml
Ralwoss: That is the main output of the tool, so it should not be empty, but the test casesreturn empty results since the upgrade to SemiBin2

bgruening: Isn't that indicating a bug somewhere?

Ralwoss: I tried it on Gitpod via planemo and on my local machine vie SemiBin2 directly and the output was the same. If it is a bug, it should be in SemiBin directly. It could also be possible that SemiBin2 is filtering the bins more strictly.

Ralwoss: I tried a test case without output with a different, slightly larger dataset and there were bins found, so it most likely is that the test data is not suited for SemiBin2 anymore

bgruening: Is "slightly larger dataset" still under 1MB so we can use it here? :)

Ralwoss: unnfortunately not, the contig file is nearly 2 and the corresponding bam is nearly 6MB

bgruening: Ok. Please keep in mind that we will have soon a `location` feature. Which can take URL pointing to larger datasets. We should use this next time we update this tool.

Thanks @Ralwoss

bgruening (approved):

PR #5637, opened 2023-11-15
Scanpy reset default colours for ploting

pavanvidem:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/scanpy/inspect.xml
+++ b/tools/scanpy/inspect.xml
@@ -66,11 +66,11 @@ sc.pp.calculate_qc_metrics(
     adata=adata,
     expr_type='$method.expr_type',
     var_type='$method.var_type',
-    #if str($method.qc_vars) != ''
+    #if $method.qc_vars
         #set $qc_vars = [str(x.strip()) for x in str($method.qc_vars).split(',')]
     qc_vars=$qc_vars,
     #end if
-    #if str($method.percent_top) != ''
+    #if $method.percent_top
         #set $percent_top = [int(x.strip()) for x in str($method.percent_top).split(',')]
         percent_top=$method.percent_top,
     #end if
@@ -122,10 +122,10 @@ sc.tl.score_genes_cell_cycle(
 sc.pp.neighbors(
     adata=adata,
     n_neighbors=$method.n_neighbors,
-    #if str($method.n_pcs) != ''
+    #if $method.n_pcs
     n_pcs=$method.n_pcs,
     #end if
-    #if str($method.use_rep) != ''
+    #if $method.use_rep
     use_rep='$method.use_rep',
     #end if
     knn=$method.knn,
@@ -138,7 +138,7 @@ sc.pp.neighbors(
 sc.tl.rank_genes_groups(
     adata=adata,
     groupby='$method.groupby',
-    #if str($method.groups) != ''
+    #if $method.groups
     #set $group=[x.strip() for x in str($method.groups).split(',')]
     groups=$group,
     #end if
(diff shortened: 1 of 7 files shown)


tools/scanpy/macros.xml
```diff
@@ -861,7 +868,7 @@ with open('anndata_info.txt','w', encoding='utf-8') as ainfo:
         <param argument="legend_fontsize" type="integer" optional="true" value="" label="Legend font size" help=""/>
     </xml>
     <token name="@CMD_param_legend_fontsize@"><![CDATA[
-    #if str($method.plot.legend_fontsize) != ''
```
bernt-matthias, line 864: I think this syntax is needed for optional integer (or float) parameters since `''`×and `0` evaluate to `False`.

tools/scanpy/macros.xml
```diff
@@ -1,7 +1,7 @@
 <macros>
     <token name="@version@">1.9.6</token>
     <token name="@profile@">22.05</token>
-    <token name="@galaxy_version@"><![CDATA[@version@+galaxy0]]></token>
```
bernt-matthias, line 4: Please bump all tools. Maybe introduce a `@VERSION_SUFFIX@` token.

tools/scanpy/macros.xml
```diff
@@ -309,6 +310,7 @@ with open('anndata_info.txt','w', encoding='utf-8') as ainfo:
         <option value="winter_r">winter_r</option>
     </xml>
     <xml name="matplotlib_pyplot_colormap">
+        <option value="">Default</option>
```
bernt-matthias, line 313: This just moves this option to the macro? Instead of this you might also make the select optional (up to you).

tools/scanpy/inspect.xml
```diff
@@ -368,9 +364,9 @@ sc.pp.sqrt(
                 <param argument="n_genes" type="integer" min="0" value="100" label="The number of genes that appear in the returned tables" help=""/>
                 <conditional name="tl_rank_genes_groups_method">
                     <param argument="method" type="select" label="Method">
-                        <option value="t-test">t-test</option>
+                        <option value="t-test" selected="true">t-test</option>
```
bernt-matthias, line 367: Why change the default? Do you follow an upstream default?

tools/scanpy/macros.xml
pavanvidem: Done

tools/scanpy/macros.xml
pavanvidem: Probably, optional attribute is easy to maintain. I added it.

tools/scanpy/inspect.xml
pavanvidem: Yes, changed in the latest version.

tools/scanpy/macros.xml
```diff
@@ -662,7 +662,7 @@ with open('anndata_info.txt','w', encoding='utf-8') as ainfo:
         <option value="YellowGreen">YellowGreen</option>
     </xml>
     <xml name="param_matplotlib_pyplot_edgecolors">
-        <param argument="edgecolors" type="select" label="Edge color of the marker" help="">
+        <param argument="edgecolors" type="select" optional="true" label="Edge color of the marker" help="">
```
bernt-matthias, line 665: What will happen if the user selects no option? Same as if `"none"` is selected? If so we should remove this option, or?

pavanvidem: Thanks for the catch. From the [documentation](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.scatter.html), both of them are allowed. In my testing, both gave the same results. I removed `none`. BTW, plural `linewidths` and `edgecolors` passed to the function are not working. I fixed them [here](https://github.com/galaxyproject/tools-iuc/pull/5637/files#diff-b2ced7d32e78ae8f951fa61b63504b1dabd42806aeaefdb1ea7f1298ae1c375fR713-R715).

bernt-matthias (approved):
Good to go?

bgruening:
He said yes from the other room ;)

PR #5649, opened 2023-11-22
Update Hapo-G from 1.3.6 to 1.3.7

wm75:
FOR CONTRIBUTOR:


wm75:
Turns out the previous failures occurred because Hapo-G runs samtools sort multithreaded depending on `--threads`, but uses a fixed deafault memory assignment of 5G per thread instead of scaling it accordingly.
It expects the user to do this themselves via `--samtools-mem`.

Issue #4100, opened 2021-10-21
Fix tool section ordering


bernt-matthias:
These are trivial paper cuts:

- [ ] tools/circos/circos.xml
  - [ ] Best practice violation [stdio] elements should come before [version_command]
- [ ] tools/emboss_5/emboss_seqret.xml
  - [ ] Best practice violation [code] elements should come before [stdio]
- [ ] tools/fasta_nucleotide_color_plot/fasta_nucleotide_color_plot.xml
  - [ ] Best practice violation [macros] elements should come before [requirements]
- [ ] tools/ngmlr/ngmlr.xml
  - [ ] Best practice violation [description] elements should come before [requirements]
- [ ] tools/snp-dists/snp-dists.xml
  - [ ] Best practice violation [description] elements should come before [macros]
- [ ] tools/tximport/tximport.xml
  - [ ] Best practice violation [description] elements should come before [macros]

Issue #4094, opened 2021-10-20
Fix conditionals with missing when blocks or missing options


bernt-matthias:
For conditionals each option should have a corresponding when block and vice versa. The former case should be less critical but also easier to fix: 

Here is the list of tools that currently suffer from this (please put your name / link to the PR next to the tool if you are working on this):

- [ ] tools/barcode_splitter/barcode_splitter.xml
  - Conditional [nameinterface] no <when /> block found for select option 'forward'
  - Conditional [nameinterface] no <when /> block found for select option 'reverse'
  - Conditional [nameinterface] no <when /> block found for select option 'forwardindex'
  - Conditional [nameinterface] no <when /> block found for select option 'reverseindex'
- [ ] tools/gecko/gecko.xml
  - Conditional [options] no <when /> block found for select option 'noalignments'
  - Conditional [options] no <when /> block found for select option 'alignments'
- [ ] tools/moabs/moabs.xml
  - Conditional [bsmap_mismatch] no <when /> block found for select option '0'
  - Conditional [doComp] no <when /> block found for select option '1'

And here are the potentially more difficult cases:

- [ ] tools/qiime/qiime_core/pick_otus.xml
  - Conditional [method] no <option /> found for when block 'mothur'
  - Conditional [method] no <option /> found for when block 'blast'
  - Conditional [method] no <option /> found for when block 'sumaclust'
  - Conditional [method] no <option /> found for when block 'cdhit'
- [ ] tools/qiime/qiime_core/assign_taxonomy.xml
  - Conditional [methodcond] no <option /> found for when block 'rdp'
  - Conditional [methodcond] no <option /> found for when block 'rtax'
- [ ] tools/scanpy/plot.xml
  - Conditional [method] no <option /> found for when block 'pl.rank_genes_groups_violin'

bernt-matthias:
Covered by lint_skip

Issue #4093, opened 2021-10-20
Fix output cheetah placeholders


bernt-matthias:
Many tools use output names with invalid cheetah placeholders. Valid is `^[a-zA-Z_]\w*$`. 

A valid placeholder

- starts with a character or underscore
- and is followed by a sequence of alpha numeric characters or underscores

The following tools violate this (by starting with numbers or including dots). This should be fixed (please put your name / link to the PR next to the tool if you are working on this):

- [x] tools/codeml/codeml.xml https://github.com/galaxyproject/tools-iuc/pull/4216
  - 2ngdn
  - 2ngds
  - 2ngt
  - 4fold
- [ ] tools/khmer/extract-partitions.xml
  - groups-of-partitions
- [ ] tools/medaka/medaka_variant.xml
  - out_round_1_phased.vcf
- [ ] tools/mothur/chimera.slayer.xml
  - slayer.chimeras
- [ ] tools/mothur/biom.info.xml
  - tax.summary
- [ ] tools/mothur/chimera.uchime.xml
  - uchime.chimeras
- [ ] tools/mothur/chimera.check.xml
  - chimeracheck.chimeras
- [ ] tools/mothur/chimera.bellerophon.xml
  - bellerophon.chimeras
  - bellerophon.accnos
- [ ]  tools/mothur/chimera.pintail.xml
  - pintail.chimeras
- [ ] tools/mothur/chimera.vsearch.xml
  - vsearch.chimeras
- [ ] tools/mothur/chimera.ccode.xml
  - ccode.chimeras
  - ccode.accnos
- [ ] tools/mothur/align.check.xml
  - align.check
- [ ] tools/mummer4/dnadiff.xml
  - 1delta
  - 1coords

bernt-matthias:
Covered by lint_skip

Issue #4091, opened 2021-10-20
Fix duplicated options


bernt-matthias:
There are many tools that have select parameters with duplicated values or names, which probably needs to be fixes. It would be interesting to see if there are valid cases among those.

In addition there are a few cases that mix defining options with 'option' children elements, a 'options' element or the 'dynamic_options' attribute .. here we need to verify if this actually works, e.g. with a test. If so we need to adapt the linter.

Here is the list of tools that currently suffer from this (please put your name / link to the PR next to the tool if you are working on this):

- [ ] tool_collections/bamtools/bamtools_filter/bamtools-filter.xml
  - bam_property_selector has multiple options with the same text content
- [ ] tools/augustus/augustus.xml https://github.com/galaxyproject/tools-iuc/pull/4813
  - organism has multiple options with the same text content
  - organism has multiple options with the same value
- [ ] tools/bcftools/bcftools_view.xml
  - types has multiple options with the same text content
  - exclude_types has multiple options with the same text content
- [ ] tools/bedtools/groupbyBed.xml
  - operation has multiple options with the same text content
  - operation has multiple options with the same value
- [ ] tools/busco/busco.xml
  - augustus_species has multiple options with the same text content
  - augustus_species has multiple options with the same value
- [ ] tools/colibread/discosnp_RAD.xml
  - coverage_options has multiple options with the same text content
- [ ] tools/colibread/discosnp_pp.xml
  - coverage_options has multiple options with the same text content
- [x] tools/deseq2/deseq2.xml https://github.com/galaxyproject/tools-iuc/pull/4188
  - output_selector has multiple options with the same text content
- [ ] tools/emboss_5/emboss_twofeat.xml https://github.com/galaxyproject/tools-iuc/pull/4734
  - overlap has multiple options with the same text content
- [ ] tools/enasearch/enasearch_search_data.xml
  - sortfields has multiple options with the same text content
  - sortfields has multiple options with the same text content
  - sortfields has multiple options with the same text content
  - sortfields options have to be defined by either 'option' children elements, a 'options' element or the 'dynamic_options' attribute.
  - sortfields has multiple options with the same text content
  - sortfields has multiple options with the same text content
  - sortfields has multiple options with the same text content
  - sortfields has multiple options with the same text content
  - sortfields has multiple options with the same text content
  - sortfields has multiple options with the same text content
  - sortfields has multiple options with the same text content
  - field has multiple options with the same text content
  - sortfields has multiple options with the same text content
[...]

PR #5650, opened 2023-11-22
Drop lint from preventing deployment

mvdbeek:
and we should really take a good look at the signal we're sending by applying linting rules to PRs for things that are already broken. It does increase the barrier to contribution significantly and leads to PRs that do more than one thing and that's bad IMO.

FOR CONTRIBUTOR:

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -381,7 +381,7 @@ jobs:
   # deploy the tools to the toolsheds (first TTS for testing)
   deploy:
     name: Deploy
-    needs: [setup, lint, combine_outputs]
+    needs: [setup, combine_outputs]
     if: ${{ (github.ref == 'refs/heads/master' || github.ref == 'refs/heads/main' ) && github.repository_owner == 'galaxyproject' }}
     runs-on: ubuntu-latest
     strategy:


mvdbeek:
What we do in the Galaxy code base is fix whatever new linting rule (thing eslint, ruff, black) we're applying before making it a requirement, we should follow that same principle here.

mvdbeek:
> I understand the frustration, but I think we need a better solution.

Yes, see https://github.com/galaxyproject/tools-iuc/pull/5650#issuecomment-1823412600.

> My perception is that IUC does not have the resources or interest (obviously each of us is interested only in a subset of the tools and cares more about those than others) in fixing existing problems

Not every IUC member needs to care about all tools, but I know you and I are not alone in this, to that non-exhaustive list I would add at least @nsoranzo and @bgruening (apologies if I missed anyone there). This is the reason for my continued involvement here, I otherwise really don't care about any particular tool. Now another question is the timeline of introducing major changes affecting contributors.
You can't enable a whole barrage of things and then push it onto someone just fixing a bug in an existing tool. That's not how any project that I am aware of handles this.

> So why shouldn't we point them to other things that could be fixed and kindly ask/nudge them to do so?

Because that's how you make it seem that contributing here is always a huge effort. As you said that:

>  linter warnings are very easy to fix

... **for us**

We sure have the resources to fix all linting errors on a per rule basis, but that needs to be coordinated and prioritized.

> Would be fine for me as long as the commits are atomic.

it is not fine for me at all, and it's quite problematic when every bugfix potentially comes with a new bug. atomicity means nothing if you can't install those commits.

> https://github.com/galaxyproject/tools-iuc/issues/4100

enforcing this is IMHO a waste of our time, it has zero impact and is the last thing I would do after everything important has been addressed. I hope that's not anything that currently fails the CI.


> https://github.com/galaxyproject/tools-iuc/issues/4094

hard for me to see why that's important but I give you that it may hide some bugs

> https://github.com/galaxyproject/tools-iuc/issues/4093

that seems important, but do we lint for this ?

> https://github.com/galaxyproject/tools-iuc/issues/4091

unclear how that's supposed to be fixed and why it is a problem. Is there a linter for it ?


Would you be onboard with this plan:
 - roll back any warning level failures for everything that currently fails
 - prioritize the rules we want to fix
 - fix the tools for that new rule
 - make violations to the rule a CI error

bernt-matthias:
> > #4093
> 
> that seems important, but do we lint for this ?

Yes, https://github.com/galaxyproject/galaxy/blob/1ed254ff72b1c6c4399e0d47dbc7716bc68ff01c/lib/galaxy/tool_util/linters/outputs.py#L50 .. but super hard to fix since it affects `mothur` which fails tests for ages.

> Would you be on board with this plan:
> 
>     * roll back any warning level failures for everything that currently fails

I only would like to do this for linter messages that are not valid, i.e. false positives of the linter.  Or even better just fix the linter if possible. For me, the problem with rolling back all of them is 

1. the problem is to large. Out of the linter messages we have in the galaxy code base the tools in IUC trigger
  - approx 1/3 of the unique error messages 
  - more than half of the unique warning messages 
2. it affects all tool repositories and in particular the language server. If we would disable all those warnings tool development quality might suffer quite a bit.

We certainly have to work on this. My suggestion for the meantime would be:

- revert to use `--error-level error` instead of `warn` in the PR workflow (to get a better contributor experience .. you certainly have a good point here)
- add a linter job to weekly CI (to create awareness and track our progress)

>     * prioritize the rules we want to fix
>     * fix the tools for that new rule

This part I fully support.  For a start, I just ran `shed_lint` to get an overview of the extent of the problem. At the moment we have 

- 157 linter errors [shed_lint_error.txt](https://github.com/galaxyproject/tools-iuc/files/13449090/shed_lint_error.txt) corresponding to 21 unique messages
- 1580 linter warning [shed_lint_warn.txt](https://github.com/galaxyproject/tools-iuc/files/13449091/shed_lint_warn.txt) corresponding to 35 unique messages

If I'm not wrong we have a total of 53 unique warning messages in our linters and 76 error messages (determined by just  `egrep "error|(warn(" lib/galaxy/tool_util/linters/*py`).

Wondering how to discuss/prioritize those most efficiently. I could imagine a shared document that everyone can reorder/comment on might be more useful than a GitHub issue, or? I'm also sure some of the problems can also be solved by automatic means. 

I think what we could need on the long run is a code for each linter message (like the pep8/flake codes), such that the repos / users can select which rules should apply. Maybe also restructure the linter code, such that we have 1 linter function per message.

bernt-matthias:
> `[noci]` seems fine in that case ? 

I agree, but wrt your point on atomic comits / PRs this also would result in a non-installable intermediate tool version. But this is certainly fine for me.

> I do not understand the argument. Take one rule, fix it, done. There isn't really any viable alternative

The question is what you meant by "roll back any warning level failures ". I understood to remove / disable the linter rules, i.e. in `tool-util`. I'm totally fine with fixing all this :)  

> Only two items here would qualify as an actual error in my book

Those two are certainly of the highest priority. My problem is that we do not have a definition of what an error is. I guess that your definition of an error is everything that makes the tool run fail. My definition would be more strict: everything that makes any part of the tool fail would qualify as an error, e.g. `filter needs to define a ref attribute` renders the filter dysfunctional, I would also consider things that invalidate tests as an error.

But this discussion in large parts can be avoided if we have the possibility for a more fine grained selection. 

> this must happen before we can reasonably clean up large categories, like the warning linters. 

+1

I guess, we can just add a linter code [here](https://github.com/galaxyproject/galaxy/blob/34d4e3902ee5984accb50e6d8bb5b7a8533103d5/lib/galaxy/tool_util/lint.py#L80C9-L80C10) and think of a naming scheme. Could be just numbers like for PEP or a few capital letters that identify each message. Then restructure the code such that each error/warning code corresponds to one linter class. Some of the linters might even have a `fix()` method that could auto-fix the problems :)  

> I was under the impression that we can do that already with --skip, is that not the case ?

We can only skip whole linters, e.g. the `inputs` linter, but not separate linter messages. When we restructured the linter code we should be able to do this with the current planemo code. 

Would my plan for the meantime be acceptable for you, i.e. use error as fail level and install weekly linter CI?

mvdbeek:
> The question is what you meant by "roll back any warning level failures ".

our fail level is now warning for PRs (https://github.com/galaxyproject/tools-iuc/blob/main/.github/workflows/pr.yaml#L132), this isn't reasonable **currently**. We can either keep it at warning and skip the ones that are currently not fixed across tools-iuc, or move it back to error. If that isn't visible enough we can use the `planemo lint` >> $GITHUB_STEP_SUMMARY and list this as things to check for reviewers.

> But this discussion in large parts can be avoided if we have the possibility for a more fine grained selection.

yes, I agree, that's a very valuable thing to do.

bernt-matthias:
> yes, I agree, that's a very valuable thing to do.

I started .. open for discussion https://github.com/galaxyproject/galaxy/pull/17081

pcm32:
I think that it would be very good that this keeps moving forward, I very often these days think twice on whether to make changes to old tools because of the effort it involves to please the linter on so many things that are irrelevant for the tool to keep working as it has been so far.

bernt-matthias:
@pcm32 I think way forward will be the possibility to enable/disable linter rules in planemo. This will be possible soonish. The ingredients are 

- https://github.com/galaxyproject/galaxy/pull/17081
- https://github.com/galaxyproject/planemo/pull/1420 (requires https://github.com/galaxyproject/planemo/pull/1441)

Then we can disable all rules that are not fulfilled repo-wide (and systematically work on them).

@mvdbeek could you find the time to do a new release (or pre replease) of the python packages for 24.0 such that we can progress on planemo?

mvdbeek:
Newer than https://pypi.org/project/galaxy-tool-util/24.0.0/ ?

bernt-matthias:
I remember that most of the problems in the [planemo 24.0 PR](https://github.com/galaxyproject/planemo/pull/1441) should be fixed in https://github.com/galaxyproject/galaxy/pull/17899

I was assuming that we need new packages to fix this in the PR. Am I wrong?

Is there a way to speed this up (e.g. by having a conditional requirement in planemo's requirement.txt that installs galaxy packages from git if in CI)?

mvdbeek:
I've created a new minor release, should have new packages once https://github.com/galaxyproject/galaxy/actions/runs/8924916026/job/24512355542 completes.

mvdbeek:
And yes, you can also install from git, `pip install 'git+https://github.com/galaxyproject/galaxy.git@release_24.0#egg=galaxy-tool-util&subdirectory=packages/tool_util'`

bernt-matthias:
https://github.com/galaxyproject/tools-iuc/pull/6184

PR #5679, opened 2023-12-13
pretext map allow for map quality 0

astrovsky01:
FOR CONTRIBUTOR:

Current version of pretext map evaluates setting a minimum mapping quality as 0 to be false, and therefore sets it to the default of 10, despite 0 being a valid entry. This fixes that bug

### DIFF
--- a/tools/pretext/.shed.yml
+++ b/tools/pretext/.shed.yml
@@ -13,6 +13,6 @@ auto_tool_repositories:
   name_template: "{{ tool_id }}"
   description_template: "Wrapper for Pretext: {{ tool_name }}"
 suite:
-  name: "pretext_suite"
+  name: "suite_pretext"
   description: "A suite of tools from the Pretext suite for contact mapping"
   type: repository_suite_definition
--- a/tools/pretext/pretext_map.xml
+++ b/tools/pretext/pretext_map.xml
@@ -2,7 +2,7 @@
     <description>converts SAM or BAM files into genome contact maps</description>
     <macros>
         <token name="@TOOL_VERSION@">0.1.9</token>
-        <token name="@VERSION_SUFFIX@">0</token>
+        <token name="@VERSION_SUFFIX@">1</token>
     </macros>
     <requirements>
         <requirement type="package" version="@TOOL_VERSION@">pretextmap</requirement>
@@ -23,7 +23,7 @@
             --sortby name
             --sortorder $sorting.sortorder
         #end if
-        #if $map_qual:
+        #if $map_qual or $map_qual == 0:
             --mapq $map_qual
         #end if
         #if $filter.filter_type == "in":
--- a/tools/pretext/pretext_map.xml
+++ b/tools/pretext/pretext_map.xml
@@ -23,7 +23,7 @@
             --sortby name
             --sortorder $sorting.sortorder
         #end if
-        #if $map_qual or $map_qual == 0:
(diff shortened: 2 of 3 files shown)


tools/pretext/pretext_map.xml
```diff
@@ -23,7 +23,7 @@
             --sortby name
             --sortorder $sorting.sortorder
         #end if
-        #if $map_qual:
+        #if $map_qual or $map_qual == 0:
```
bgruening, line 26:
```suggestion
        #if str($map_qual):
```

Is I think the proper way to do it.
Please check other integers as well.

astrovsky01: This was the only integer in the tool

astrovsky01:
@bgruening lint issues are with the other pretext tool. Would you prefer I fix here or in a separate PR?

bgruening:
Please fix it here :)

tools/pretext/.shed.yml
```diff
@@ -13,6 +13,6 @@ auto_tool_repositories:
   name_template: "{{ tool_id }}"
   description_template: "Wrapper for Pretext: {{ tool_name }}"
 suite:
-  name: "pretext_suite"
+  name: "suite_pretext"
```
bernt-matthias, line 16: Can we rename the suite without breaking lineage .. probably not

tools/pretext/pretext_snapshot.xml
```diff
@@ -43,12 +43,12 @@
             <when value="png"/>
             <when value="bmp"/>
             <when value="jpeg">
-                <param name="jpegQuality" argument="--jpegQuality" type="integer" label="JPEG quality factor" value="80" min="0" max="100" help="An integer between 1 and 100, default 80. Larger values result in increased image quality and file size."/>
+                <param argument="--jpegQuality" type="integer" label="JPEG quality factor" value="80" min="0" max="100" help="An integer between 1 and 100, default 80. Larger values result in increased image quality and file size."/>
```
bernt-matthias, line 46: Also update the help which says "An integer between 1 and 100"

tools/pretext/pretext_snapshot.xml
```diff
@@ -2,7 +2,7 @@
         <description>image generator for Pretext contact maps</description>
     <macros>
         <token name="@WRAPPER_VERSION@">0.0.3</token>
```
bernt-matthias, line 4: If you like you could change this to `TOOL_VERSION` .. then we are sure that the autoupdate covers these tools.

tools/pretext/.shed.yml
astrovsky01: I can switch it back. I figured it would cause problems to switch, but it will also throw errors on updates because of linting if we don't, right?

bgruening: This is only changing to suite TS repo isn't it? This one should no one use eitherway?

bernt-matthias: > errors on updates because of linting

Do you have an example?

LOL: I use them all the time. But I'm not using pretext. So fine for me.

bgruening (approved):
Good for me.

tools/pretext/.shed.yml
astrovsky01: Yeah, it threw problems with the first commit on this PR

PR #5683, opened 2023-12-15
CI: replace `lintr` by `styler` and fix all R code

bernt-matthias:
`styler` is the R code linter/styler from tidyverse. in contrast to `lintr` it can also fix the found problems (added a script that can be used for checking and fixing). Main advantage is that we can apply this script once for all R scripts and do not have to bother contributors with errors that they are not responsible for.

In the testbump https://github.com/galaxyproject/tools-iuc/pull/5683/commits/1947881cce28c29f8e037a54355c9feacac394ed one can see it in action https://github.com/galaxyproject/tools-iuc/actions/runs/7225652632/job/19689585872?pr=5683

I would suggest **[skip-ci]**, i.e. not bump or deploy the changed tools.
There are two linter errors left (URLs: for egsea I contacted the website admins, mageck will be fixed [here](https://github.com/galaxyproject/tools-iuc/pull/5684)) which should be fine if we agree on skip-ci.

Do we want to run the lintr job also in weekly CI? 

If we want to pin the version of the installed packages (i.e. of styler) we need to use (since setup-r-dependencies does not really support setting the version)
```
    - name: Install packages
      run: |
        R -e 'install.packages("argparse", version="X.Y.Z", repos="http://cran.rstudio.com/")'
        R -e 'install.packages("styler", version="A.B.C", repos="http://cran.rstudio.com/")'
```

FOR CONTRIBUTOR:
* [x] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [x] - License permits unrestricted use (educational + commercial)
* [ ] - This PR adds a new tool or tool collection
* [ ] - This PR updates an existing tool or tool collection
* [x] - This PR does something else (explain below)

PR #5684, opened 2023-12-15
fix all XML schema errors

bernt-matthias:
discovered here https://github.com/galaxyproject/galaxy/pull/17081

FOR CONTRIBUTOR:


tools/ampvis2/export_fasta.xml
```diff
@@ -34,7 +34,7 @@
             <output name="output" ftype="fasta">
                 <assert_contents>
                     <has_text text=">OTU_2397; k__Bacteria; p__Chloroflexi; c__Anaerolineae; o__Anaerolineales; f__Anaerolineaceae; g__uncultured; s__"/>
-                    <has_size size="106282"/>
+                    <has_size value="106282"/>
```
bernt-matthias, line 37:
```suggestion
                    <has_size value="125478"/>
```

data_managers/data_manager_build_kraken2_database/data_manager/kraken2_build_database.xml
```diff
@@ -1,8 +1,9 @@
 <?xml version="1.0"?>
 <tool id="kraken2_build_database" name="Kraken2" tool_type="manage_data" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
+    <description>database builder</description>
     <macros>
         <token name="@TOOL_VERSION@">2.1.2</token>
-        <token name="@VERSION_SUFFIX@">0</token>
+        <token name="@VERSION_SUFFIX@">1</token>
```
mvdbeek, line 6:
```suggestion
        <token name="@VERSION_SUFFIX@">0</token>
```

mvdbeek: Ah, I guess the clean param actually does have some effect.

data_managers/data_manager_diamond_database_builder/data_manager/data_manager_diamond_database_builder.xml
```diff
@@ -15,10 +15,10 @@
             help="This will be shown to the user in the database selectbox."/>
         <param type="text" name="database_id" value="" label="Database ID" help="This needs to be unique. No whitespaces allowed. Use underscores." />
         <conditional name="reference_source">
-            <param name="reference_source_selector" type="select" label="Choose the source for the reference genome">
+            <param name="reference_source_selector" type="select" label="Choose the source for the reference genome" help="For History it's recommended to archive the original FASTA file as well.">
```
mvdbeek, line 18:
```suggestion
            <param name="reference_source_selector" type="select" label="Choose the source for the reference genome">
```
This is just plain confusing and probably good that this wasn't even shown.

tools/ampvis2/export_fasta.xml
```diff
@@ -34,7 +34,7 @@
             <output name="output" ftype="fasta">
                 <assert_contents>
                     <has_text text=">OTU_2397; k__Bacteria; p__Chloroflexi; c__Anaerolineae; o__Anaerolineales; f__Anaerolineaceae; g__uncultured; s__"/>
-                    <has_size size="106282"/>
+                    <has_size value="125478"/>
```
mvdbeek, line 37: Isn't that surprising ? Was this failing before ?

mvdbeek (approved):
Looks good! I'd suggest we merge this before this becomes too large ? I don't think all the lint failures need to be fixed before we merge.

tools/ampvis2/export_fasta.xml
bernt-matthias: > Was this failing before ?

No

> Isn't that surprising ? 

A bit. But the reason is that the assertion does nothing if it does not receive a value. This is because The numeric comparisons of assertions all end up here: https://github.com/galaxyproject/galaxy/blob/573b9024808b2e1583f49f82c5c7f14d541900f7/lib/galaxy/tool_util/verify/asserts/_util.py#L36 

We could (and maybe should) change this, such that either `n` or at least one of `min`, `max` or `delta` are given.

mvdbeek: ahh, i missed the mismatch between value and size again. yeah, that makes sense.

bernt-matthias:
> Looks good! I'd suggest we merge this before this becomes too large ?

100%

>  I don't think all the lint failures need to be fixed before we merge.

Hope to find time to get rid of them in small portions. In particular the R linter errors I try to tackle here https://github.com/galaxyproject/tools-iuc/pull/5683 .. review would be appreciated :)

PR #5695, opened 2023-12-27
Jbrowse2

fubar2:
FOR CONTRIBUTOR:
* [X] - I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [X] - License permits unrestricted use (educational + commercial)
* [X] - This PR adds a new tool or tool collection

This is a very stripped-down working prototype, repurposing @hexylena's JBrowse1-in-Galaxy for JBrowse2. 

Available for testing from the [toolshed](https://toolshed.g2.bx.psu.edu/view/fubar/jbrowse2dev/af3fba6c21aa)

Most things are working in testing - screenshot shows bed bam vcf maf bigwig gff sample data tracks
image

and HiC
image

- JB1 is EOL.
- JB2 configuration json is wildly different from JB1. 
- JB2 defaults are sensible enough for many non-complicated situations.
  - for many workflows like the [TreeValGal](https://github.com/fubar2/treeval_gal) project, this will work very well as is
  - some challenges using jbrowse1 with very dense tracks in huge VGP data. 
- Only the easy part is working for the basic tracks, recycling Helena's core code, and adding hic.
- URLs are used wherever easy to do to minimise the size of JBrowse2 files folder so these are much smaller than jb1 where all were copied. Not practicable for huge files like hic. JBrowse2 lets galaxy offload existing data and sequence files to nginx range downloads.
- A _lot of work_ remains, to figure out and implement all the new track configuration options in the UI. 
   - Could not find any human readable documentation on these - need to use the browser to find track setting minutae. 
   - Nor was there anything on changing to different codon schemes that I could find. So that's a big loss - but it might be buried somewhere. The new architecture is plugins to do stuff so might need one for codons.
  - SPARQL, REST remain to be implemented.

PR #5697, opened 2024-01-08
JBrowse2: #5695 rebased with @abretaud (#3997)

fubar2:

#### Merges #5695 derived from @hexylena JBrowse1 tool with most of the improvements made by @abretaud #3997 such as adding track controls to the tool form. 

Sad I didn't know about #3997 until after repeating a lot of the same updates - but that PR included all the track formatting UI form code that #5695 was missing.

Merge adds:
- track json without calls to jbrowse cli - can be more finely controlled.
- added tests for some common track types
- working default session 
- Form has track options including colour/styling for different track types. Expanded UI sample image shown below.
- MAF tracks will work when bcbio-gff is updated when https://github.com/chapmanb/bcbb/pull/141 is accepted

Removes some #3997 tests failing because looking for JB1 elements - need updating. 
Fixes testing in biocontainers - different path in Conda so now use `jbrowse create`. 

Available for testing from the [toolshed](https://toolshed.g2.bx.psu.edu/view/fubar/jbrowse2/37b2c26c7b9b)

Screenshot shows test bed bam vcf maf bigwig gff sample data tracks
image

and HiC
image

- SPARQL and paf remain to be implemented.
- Needs a test for each track input type

### expanded UI example:
image


hexylena:
It will take some time to review. Would it be possible to re-base your work atop #3997 and retain the commits? @abretaud has been working on this for a long time, it would be a pity to lose those commits and authorship attribution fully. It would really help the reviewing process, it'd be clearer what was new, relative to the work Anthony has already done, so he could say whether the changes made sense or what might still be needed.

tools/jbrowse2/servejb2.py
```diff
@@ -0,0 +1,159 @@
+#!/usr/bin/env python3
+
+# spec: simplest python web server with range support and multithreading that takes root path,
+# port and bind address as command line arguments; by default uses the current dir as webroot,
+# port 8000 and bind address of 0.0.0.0
+# borrowed from https://github.com/danvk/RangeHTTPServer
```
hexylena, line 6: you must reproduce the license in this file then.

tools/jbrowse2/static/images/bam.png
hexylena, line 1: this is a jb1 screenshot

tools/jbrowse2/static/images/blast.png
hexylena, line 1: this rendering is from the BlastXML plugin which is jb1 only, not sure if we migrate that or replace the rendering. Perhaps just using the MAF rendering is the better choice, and providing a new converter, blatsxml to maf.

tools/jbrowse2/static/images/sections.png
hexylena, line 1: more jb1

tools/jbrowse2/test-data/index.html
```diff
@@ -0,0 +1,112 @@
...
+  <head>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+    <title>JBrowse</title>
+    <link rel="stylesheet" type="text/css" href="css/genome.css">
+</head>
+<body>
+
+    <script type="text/javascript">
+            // jshint unused: false
+            var dojoConfig = {
```
hexylena, line 12: dojo is also jb1, so, this index is wrong.

thus also the test is wrong or this isn't being tested, both of which are sub-optimal.

tools/jbrowse2/maf2bed.pl
hexylena, line 1: this probably needs a license header due to importing it from the plugin

tools/jbrowse2/jbrowse2.xml
```diff
@@ -0,0 +1,1062 @@
...
+A JBrowse2 history item can be opened by viewing it (the "eye" icon).
+They can also be downloaded as archives ("floppy disk" icon) to share and for local viewing.
+One extra step is required before they can be viewed. A local python web server must be started using a script included in each archive.
+Unzip the archive (*unzip [filename].zip*) and change to the first level directory. It contains a file named "servejb2.py"
+
+Assuming you have python3 installed, running:
+
+*python3 servjb2.py*
+
+will serve the unarchived JBrowse2 configuration, so it can be browsed by pointing a web browser to localhost:8080
```
hexylena, line 967: oof. This is unfortunate. Maybe we rename it just `server.py` or something more clear, it's in a jb2 folder it doesn't need to be named jb2.

tools/jbrowse2/jbrowse2.xml
```diff
@@ -0,0 +1,1062 @@
...
+Compared to JBrowse1-in-Galaxy, there is no support for alternative codons for unusual genomes,
+and detailed track styling is not yet implemented. Send code.
+JBrowse1 development has now ceased in favour of JBrowse2.
+
+Use and local viewing
+=====================
+
+A JBrowse2 history item can be opened by viewing it (the "eye" icon).
+They can also be downloaded as archives ("floppy disk" icon) to share and for local viewing.
+One extra step is required before they can be viewed. A local python web server must be started using a script included in each archive.
```
hexylena, line 960:
```suggestion
One extra step is required before they can be viewed. A local web server must be run in order to access the contents, for example using an existing webserver, or the included python script in each archive.
```

tools/jbrowse2/macros.xml
```diff
@@ -0,0 +1,509 @@
...
+            <requirement type="package" version="1.82">biopython</requirement>
+            <requirement type="package" version="0.7.0">bcbio-gff</requirement>
+            <requirement type="package" version="1.19">samtools</requirement>
+            <requirement type="package" version="6.0.1">pyyaml</requirement>
+            <requirement type="package" version="1.11">tabix</requirement>
+            <requirement type="package" version="4.6.0">findutils</requirement>
+            <yield/>
+        </requirements>
+    </xml>
+    <token name="@DATA_DIR@">\$GALAXY_JBROWSE_SHARED_DIR</token>
```
hexylena, line 26: Is this not used?

tools/jbrowse2/macros.xml
```diff
@@ -0,0 +1,509 @@
...
+    <token name="@WRAPPER_VERSION@">galaxy0</token>
+    <token name="@ATTRIBUTION@"><![CDATA[
+**Attribution**
+This Galaxy tool relies on the JBrowse, maintained by the GMOD Community. The Galaxy wrapper is developed by the IUC
```
hexylena, line 30:
```suggestion
This Galaxy tool relies on the JBrowse2, maintained by the GMOD Community. The Galaxy wrapper is developed by the IUC
```

tools/jbrowse2/jbrowse2.py
```diff
@@ -0,0 +1,1451 @@
...
+            "-n",
+            "JBrowse2 in Galaxy",
+            "--target",
+            self.config_json_file,
+            "-v",
+            " LinearGenomeView",
+        ]
+        if self.debug:
+            log.info("### calling set-default-session with cmd=%s" % "  ".join(cmd))
+        self.subprocess_check_call(cmd)
```
hexylena, line 513: logging is already handled by the check_call wrapper, why duplicate it ehre?

tools/jbrowse2/jbrowse2.py
```diff
@@ -0,0 +1,1451 @@
...
+        from https://github.com/cmdcolin/maf2bed
+        Note: Both formats start with a MAF as input, and note that your MAF file should contain the species name and chromosome name
+        e.g. hg38.chr1 in the sequence identifiers.
+        need the reference id - eg hg18, for maf2bed.pl as the first parameter
+        """
+        mafPlugin = {
+            "plugins": [
+                {
+                    "name": "MafViewer",
+                    "url": "https://unpkg.com/jbrowse-plugin-mafviewer/dist/jbrowse-plugin-mafviewer.umd.production.min.js",
```
hexylena, line 597: please pin the version, otherwise this won't be reproducible (from an end user perpsective)

tools/jbrowse2/jbrowse2.py
```diff
@@ -0,0 +1,1451 @@
...
+            "favicon.ico",
+            "robots.txt",
+            "umd_plugin.js",
+            "version.txt",
+            "test_data",
+        ]:
+            cmd = ["rm", "-rf", os.path.join(self.outdir, fn)]
+            self.subprocess_check_call(cmd)
+        cmd = ["cp", os.path.join(INSTALLED_TO, "servejb2.py"), self.outdir]
+        self.subprocess_check_call(cmd)
```
hexylena, line 1278: I think we need to go back to how we did it in jb1 -  a copy of the source code included in the conda package that was copied based on the jbrowse source directory.

Per their description of this command:

> This fetches the latest version of jbrowse-web and unzips it to a folder named "jbrowse2" from github (https://github.com/GMOD/jbrowse-components/releases), you could run this step manually if you wanted to instead.

which makes the job unreproducible, jbrowses generated at different times could behave differently. I'd much rather see a copy of the source baked into the conda package like we used to, to ensure it's reproducible. I get that this was easier though.

hexylena (changes requested):
Some comments in line, i'm really excited for jb2, but there's a lot of small details to get the tool optimal.

abretaud:
I still have to look into details but for now:

> URLs are used wherever easy to do to minimise the size of JBrowse2 files folder so these are much smaller than jb1 where all were copied. (from #5695)

That can be a problem, ie the output dataset is no longer self-containing but depends on the existence/sharing of other ones, that can be problematic. And I have a use case where I use galaxy to generate jbrowse archives that I can download/deploy on another server.
Having at least an option for that would be helpful

The default session / track styling was still largely WIP, I remember I wasn't sure of the proper way to implement this, and opened a discussion there: https://github.com/GMOD/jbrowse-components/discussions/3568
I don't know if there wer changes in jb code to make this easier since my last try

tools/jbrowse2/servejb2.py
fubar2: ok - good point - copy of apache2 added

tools/jbrowse2/static/images/blast.png
fubar2: that's the gff display using jb1 code modified to write json. Better rendering is not something I know anything about but at least it does produce a track...so can be improved if somone knows how.

On MAF - awaiting Brad Chapman accepting a PR to fix bcbio-gff for latest biopython - once that's in, the maf code basically from JBI will work well. AFAIK broken and not currently on the track menu in JB1.

tools/jbrowse2/macros.xml
fubar2: not used - uses the jbrowse create command to avoid path differences between conda and biocontainers.

tools/jbrowse2/jbrowse2.py
fubar2: Agree with the downside - need to adjust for biocontainers - will take a look.

tools/jbrowse2/static/images/bam.png
fubar2: all were just copied - now removed from repo and help until the help is redone.

tools/jbrowse2/test-data/index.html
fubar2: No longer used. Removed.

fubar2:
No longer used as too unreliable - depending on the history being published and the server being anonymously accessible - otherwise breaks badly.

> I still have to look into details but for now:
> 
> The default session / track styling was still largely WIP, I remember I wasn't sure of the proper way to implement this, and opened a discussion there: [GMOD/jbrowse-components#3568](https://github.com/GMOD/jbrowse-components/discussions/3568) I don't know if there wer changes in jb code to make this easier since my last try

tools/jbrowse2/jbrowse2.py
fubar2: was following the author's recipe of course - good point. Will fix.

fubar2:
>Would it be possible to re-base your work atop https://github.com/galaxyproject/tools-iuc/pull/3997 and retain the commits?

@hexylena: done. Interesting experience.

tools/jbrowse2/maf2bed.pl
fubar2: same apache2_license.txt is now referenced in that header.

tools/jbrowse2/jbrowse2.xml
fubar2: this is probably not needed but as you will. Renamed to webserver.py since that's what it does...

tools/jbrowse2/jbrowse2.py
fubar2: left over from some bug ridden anxiety or other. 
Many eyeballs required to get this stuff right ...fixed.

tools/jbrowse2/jbrowse2.py
fubar2: now pinned with
`JB2REL="v2.10.0"` for use in the -t parameter for release tag

tools/jbrowse2/jbrowse2.xml
fubar2: change committed - thanks..

tools/jbrowse2/jbrowse2.xml
hexylena: Let's do this instead: [redbean2](https://justine.lol/redbean2/)

- runs on most platforms we'll care about (windows/linux/MacOS)
- single static file
- can serve files
- it's a polyglot binary, the same binary works across all those platforms.
- can automatically open the index page in the user's browser.

it's 2.3M just stuff it in the tool directory and copy from there to the output dir for "full" outputs.

Running `./redbean.com -D . ` will serve the local directory, just include the following as `.init.lua` and she'll be right.

```
-- /.init.lua is loaded at startup in redbean's main process
HidePath('/usr/share/zoneinfo/')
HidePath('/usr/share/ssl/')

-- open a browser tab using explorer/open/xdg-open
LaunchBrowser('/index.html')

function OnServerListen(fd, ip, port)
   unix.setsockopt(fd, unix.SOL_TCP, unix.TCP_SAVE_SYN, true)
   return false
end

function OnClientConnection(ip, port, serverip, serverport)
   syn, synerr = unix.getsockopt(GetClientFd(), unix.SOL_TCP, unix.TCP_SAVED_SYN)
end

-- this intercepts all requests if it's defined
function OnHttpRequest()
   if GetHeader('User-Agent') then
      Log(kLogInfo, "client is running %s and reports %s" % {
             finger.GetSynFingerOs(finger.FingerSyn(syn)),
             VisualizeControlCodes(GetHeader('User-Agent'))})
   end
   if HasParam('magic') then
      Write('<p>\r\n')
      Write('OnHttpRequest() has intercepted your request<br>\r\n')
      Write('because you specified the magic parameter\r\n')
      Write('<pre>\r\n')
      Write(EscapeHtml(LoadAsset('/.init.lua')))
      Write('</pre>\r\n')
   else
      Route() -- this asks redbean to do the default thing
   end
   SetHeader('Server', 'redbean!')
end
```

hexylena:
>  An, um, interesting experience since I've never tried it before.

did a bang up job, cheers. really appreciate it!

> There's lots more to do and it needs feedback from users to improve, but arguably useful enough to expose as is so we can get that user feedback.

yeah, i mean, it sounds like there's a lot of pressure from one side to get this shipped, it could be installed just on that server in the current as-is state for better "real life" testing before we get it finalised since, yeah it's a massive tool.

tools/jbrowse2/jbrowse2.py
```diff
@@ -0,0 +1,1457 @@
...
+import subprocess
+import tempfile
+import xml.etree.ElementTree as ET
+from collections import defaultdict
+
+logging.basicConfig(level=logging.INFO)
+log = logging.getLogger("jbrowse")
+TODAY = datetime.datetime.now().strftime("%Y-%m-%d")
+GALAXY_INFRASTRUCTURE_URL = None
+JB2REL = "v2.10.0"
```
abretaud, line 22: I'm sure we'll forget to update it here :grimacing:
Could you pass it in the command line so that we can use the defines in the macro?

tools/jbrowse2/jbrowse2.py
```diff
@@ -0,0 +1,1457 @@
+#!/usr/bin/env python
+# change to accumulating all configuration for config.json based on the default from the clone
```
abretaud, line 2: isn't it a leftover comment?

tools/jbrowse2/jbrowse2.py
```diff
@@ -0,0 +1,1457 @@
...
+        self.genome_names = []
+        self.trackIdlist = []
+        self.tracksToAdd = []
+        self.config_json = {}
+        self.config_json_file = os.path.join(outdir, "config.json")
+        self.clone_jbrowse()
+
+    def subprocess_check_call(self, command, output=None):
+        if output:
+            if self.debug:
```
abretaud, line 390: Can you use `logging.basicConfig` at the top of the script to show or not debug messages?

tools/jbrowse2/jbrowse2.py
```diff
@@ -0,0 +1,1457 @@
...
+                style_data,
+            ]
+        }
+
+    def process_genomes(self):
+        assemblies = []
+        for i, genome_node in enumerate(self.genome_paths):
+            genome_name = genome_node["meta"]["dataset_dname"].strip()
+            if len(genome_name.split()) > 1:
+                genome_name = genome_name.split()[0]
```
abretaud, line 452: We'll surely get name collisions here, and meaningless names too, we need a better name cleaning method.
There was some code for that in my PR, I guess you can reuse it somehow

tools/jbrowse2/jbrowse2.py
```diff
@@ -0,0 +1,1457 @@
...
+            uri: 'https://s3.amazonaws.com/jbrowse.org/genomes/GRCh38/fasta/GRCh38.fa.gz',
+            faiLocation:
+            uri: 'https://s3.amazonaws.com/jbrowse.org/genomes/GRCh38/fasta/GRCh38.fa.gz.fai',
+            gziLocation:
+            uri: 'https://s3.amazonaws.com/jbrowse.org/genomes/GRCh38/fasta/GRCh38.fa.gz.gzi',
+        Cool will not be likely to be a good fit - see discussion at https://github.com/GMOD/jbrowse-components/issues/2438
+        """
+        tId = trackData["label"]
+        # can be served - if public.
+        # dsId = trackData["metadata"]["dataset_id"]
```
abretaud, line 557: can be deleted if I udnerstand correctly, we're back to copying data now?

tools/jbrowse2/jbrowse2.py
```diff
@@ -0,0 +1,1457 @@
...
+                self.add_maf(
+                    dataset_path,
+                    outputTrackConfig,
+                )
+            elif dataset_ext == "bigwig":
+                self.add_bigwig(
+                    dataset_path,
+                    outputTrackConfig,
+                )
+            elif dataset_ext == "bam":
```
abretaud, line 1128: cram is missing, it's very similar to bam, I had a add_xam for both in my PR IIRC

tools/jbrowse2/jbrowse2.py
```diff
@@ -0,0 +1,1457 @@
...
+        tool_id=metadata["tool_tool_id"],
+        # tool_version=metadata['tool_tool_version'],
+    )
+    return metadata
+
+
+class JbrowseConnector(object):
+    def __init__(self, outdir, genomes):
+        self.debug = False
+        self.usejson = True
```
abretaud, line 375: Is there a reason to keep both methods?

abretaud (changes requested):
Ok, I've made a first review of the code, with a few comments/changes to do. Can you look into it before merging?
In the meantime I'd like to also make a few tests locally

tools/jbrowse2/macros.xml
```diff
@@ -0,0 +1,513 @@
...
+        </edam_topics>
+        <edam_operations>
+            <edam_operation>operation_0573</edam_operation>
+            <edam_operation>operation_0564</edam_operation>
+        </edam_operations>
+    </xml>
+    <xml name="requirements">
+        <requirements>
+            <requirement type="package" version="@TOOL_VERSION@">jbrowse2</requirement>
+            <requirement type="package" version="1.82">biopython</requirement>
```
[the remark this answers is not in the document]
fubar2: Those calls are still being deprecated - just back until then - and need to undo corresponding fixes to the python script. 
Prefer waiting for https://github.com/chapmanb/bcbb/pull/141 before altering at all, because @bgruening had to manually adjust the biocontainer last time!

fubar2:
Now somewhat simplified again, but seems to be working with the test data I have.

#### Track details, like colours, can only be configured by the user in the track browser - not on the tool form.

Making them configurable requires hacking the internal _default state_. 
It is not documented, so requires reverse engineering and it has no long term stable guarantees.
It can be done as people find time and users make demands, but not by me.

#### Note to reviewers

Please don't ask me to fix the things listed below as part of a review, because I can't. This needs people who understand paf, cram and sparql which is not me. Jbrowse2 is a very substantial undertaking, and as far as I can tell, this PR is now sufficiently useful to have available as a starting point, to which others can contribute the bits they know and care about.

#### Things remaining:

- testing/fixing PAF synteny tracks - the code is there but untested as I have no data or clue;
- testing/fixing SPARQL tracks - ditto
- modifying the working bam code for cram input. 
  - Whether they are the same function is up to whoever has cram data to test and PR. 
- update to 2.10.1 and bcbbio-gff to restore MAF tracks - awaiting Brad merging a PR to deal with BioPython deprecations
  - Biopython has temporarily restored the deprecation with a warning but only until next release - it will all go then so we need Brad's help. AFAIK @bgruening has to fiddle with the biocontainer? Better to burden him only once, rather than several times in a row?

tools/jbrowse2/jbrowse2.py
fubar2: PR please. I cannot do this as I have no data or experience.

tools/jbrowse2/jbrowse2.py
fubar2: yes. just a note to myself. It took a while to figure out so preserved. I think it will be useful but have not had time to make it a reliable alternative input source - it requires duplicating a lot of jbrowse2 internal logic for index file and other accessories.

tools/jbrowse2/jbrowse2.py
fubar2: Now uses a hash for the file name. 
The genome name really matters a lot for MAF where it must match precisely...

tools/jbrowse2/jbrowse2.py
fubar2: all gone now

tools/jbrowse2/readme.rst
```diff
@@ -0,0 +1,117 @@
...
+the visualizations. An example used during development/testing can be seen in
+`test-data/*/test.xml`. The format is in no way rigorously defined and is
+likely to change at any time! Beware. ;)
+
+History
+=======
+
+- 2.10.0+galaxy2
+
+    - UPDATED to JBrowse 2.10.0
```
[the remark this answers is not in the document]
fubar2: done

tools/jbrowse2/readme.rst
```diff
@@ -0,0 +1,117 @@
...
+
+The Galaxy tool interface writes out a xml file which is then used to generate
+the visualizations. An example used during development/testing can be seen in
+`test-data/*/test.xml`. The format is in no way rigorously defined and is
+likely to change at any time! Beware. ;)
+
+History
+=======
+
+- 2.10.0+galaxy2
```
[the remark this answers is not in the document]
fubar2: see note below. let's wait for Brad and the bcb gff fix for updating dependencies.
I'm done here. If you want to change the pinned version and test it, a PR would be welcomed.

tools/jbrowse2/apache2_license.txt
```diff
@@ -0,0 +1,201 @@
+Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
```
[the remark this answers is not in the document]
fubar2: note added to the specific code.

tools/jbrowse2/macros.xml
```diff
@@ -0,0 +1,513 @@
...
+        </sanitizer>
+    </xml>
+
+
+    <xml name="input_conditional" token_label="Track Data" token_format="data">
+        <param label="@LABEL@" format="@FORMAT@" name="annotation" type="data" multiple="True"/>
+    </xml>
+    <xml name="citations">
+        <citations>
+        <citation type="doi">10.1101/gr.094607.109</citation>
```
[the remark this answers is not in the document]
fubar2: Might as well have both for the same price.

tools/jbrowse2/jbrowse2.py
fubar2: no but someone who wants to is more than welcome to.

tools/jbrowse2/jbrowse2.py
fubar2: probably is. there are quite a few. Some are useful, others not.

tools/jbrowse2/jbrowse2.xml
fubar2: A PR would be favourite. It's probably not needed if https://github.com/GMOD/jbrowse-components/discussions/4161#discussioncomment-8024748 works.

hexylena:
> Track details, like colours, can only be configured by the user in the track browser - not on the tool form.

That's unfortunate, but no, let's not undertake modifying internal state, that sounds fraught. We *should* consider flagging this upstream, since it was a nice feature in the past and gave users a bit of control, but also maybe we can live without it. (I'm saying: I used it and loved it; not sure anyone else did.)

> Advice is wonderful, but I cannot sustain all the work of implementing it.

I mean, yeah...  this is why it took us so long to get here, none of us have been able to allocate the time and energy to implementing such a big change. @abretaud had done all the work up until here, I would have loved to help but haven't been able to find free time. It doesn't help that for many of our use cases, jb1 is still sufficient (especially e.g. apollo.), and there wasn't an immediate need for jb2 even though it's a massive improvement in many many ways.

Sorry you had to find out the hard way :( doubly so, for learning about the existing PR after expending so much energy.

> Please pitch in if you think this tool is worth having.

We definitely do, but "things we think are worth having" and "science-capitalism allocates us available bandwidth to work" are two very, very separate axes.

> Jbrowse2 is a very substantial undertaking, and as far as I can tell, this PR is now sufficiently useful to have available as a starting point

yes, we can look towards merging. I guess you'll have to give @abretaud and I some time to install and test it (but cases like this are also *why* EU has https://github.com/usegalaxy-eu/temporary-tools/, to ship testing tools so they can use it immediately and/or debug a pre-release tool.)

> testing/fixing SPARQL tracks - as above

not a priority.

fubar2:
> and I some time to install and test it
That would be wonderful when you can find time. No more I can do at this point.

bgruening:
@hexylena would it help if I deploy it on EU as test tool?

hexylena:
@bgruening that would be a fantastic help!!

bgruening:
@hexylena @abretaud here you go: https://usegalaxy.eu/?tool_id=jbrowse2

fubar2:
whoopsie

fubar2:
not intended

fubar2:
This is being reopened as a new PR as some git repairs went bad.

hexylena:
Danke sehr @bgruening

hexylena:
https://usegalaxy.eu/u/helena-rasche/h/jbrowse2-testing looks like some issues unfortunately,  not sure what's happening there.

fubar2:
> https://usegalaxy.eu/u/helena-rasche/h/jbrowse2-testing looks like some issues unfortunately, not sure what's happening there.

dependencies were not installed so jbrowse not found :(
I just ran it on EU - please try again?

bgruening:
please try again, works now for me.

hexylena:
success-ish

is it expected that no track data is rendering for me? (chrome + firefox on linux)
https://usegalaxy.eu/datasets/4838ba20a6d8676556a41e2bf6a52c7d/preview?__vkey__=1705659579999

jbrowse with uniformly empty tracks

fubar2:
When I choose the *other* NC contig:
image

hexylena:
Ah that does look better. That absolutely needs to be fixed. Where is `genome.fa` coming from? Is it a URL parameter we can pass based on the first provided contig?

(Also can we re-open this PR, the git history was fine here, the new one less so.)

fubar2:
> (Also can we re-open this PR, the git history was fine here, the new one less so.)

I managed to make a real mess of the branch with some maladroit repair work.

hexylena:
> No idea where that empty genome.fa is coming from. It's not a built in feature that I know of.

No, just a single file, the `genome.fa` is the name of the input file, so, galaxy must be passing that in somehow. It's being referenced in many places as the "assembly".

image

I think this will also be relevant when we want to display multiple genome assemblies together for the synteny view. E.g. we'll need to add a "grape" and "peach" assembly separately, with separate genomes and separate tracks.

I suspect we need a top level `repeat` in the tool to support this use case, where we can give each assembly a name, and then select a genome/fasta file to go with that assembly, as well as tracks. It'll not be relevant for most people, but they can just leave the 'default' assembly there and won't have to modify anything.

I'm flagging this specifically because it will make different versions of the tool very hard to switch between, I think we need to add this now, so when we use in the future, people will be able to "rerun with newer version" without having to completely reconfigure their entire jbrowse. You can set min=1/max=1 on the repeats for now, so you don't have to implement any of the internals for supporting/adding multiple repeats, but I think the options + data need to be there to make users' lives easier as time goes on, and when they want to use a newer version of the wrapp.er 

> Do you have >1 genome selected on the tool form? Never tried that.

this would generate multiple JBrowses.

> I managed to make a real mess of the branch with some maladroit repair work.

Ah, gotcha.

> Review

I miss the option to set big wig scaling (local vs global), but I guess this needs to be filed upstream. And I miss the multibigwig plugin

and some other comments

- [ ] track groupings do not seem to be used, they should be.
- [ ] I don't understand the purposes of the colours. I changed mine, nothing seemed to change in the output. Maybe they aren't passed correctly? I'd like to use them that's a cool feature. 
- [ ] as far as i can tell font size also isn't taking effect. 
- [ ] why does setting the default location break the session? I am no longer loaded directly into jb2 with the track view, instead I'm in a "create a new session" view where i have to select that yes, I really want that file.  Should this be passed via the URL or something instead to prevent that? I just wanted to load a specific location.
      image

filed/commented on upstream:

- https://github.com/GMOD/jbrowse-components/issues/1765
- https://github.com/GMOD/jbrowse-components/issues/4173
- https://github.com/GMOD/jbrowse-components/issues/4174


Since you have chosen to continue using the trackXML that we devised, I'd like to propose a very small tool that does a jb1 to jb2 "upgrade", it should take the jbrowse1 HTML output (then look on it's files_path for the galaxy.xml) and pass the galaxy.xml directly to the new configuration script. **If** your script is properly handling missing fields (by ignoring or using a sensible default), then it could be a very easy and cool feature to motivate users to try out the new jb2.

fubar2:
@hexylena WRT the genome.fa issue - "genome.fa" is the name of the file in galaxy.xml I see so that part is correct and probably not related to the template name `genome`. Actual reference compressed/index filenames are hashed now to avoid collision as suggested, but the name matters a lot for MAF - must be the primary genome in the maf chromosome identifiers - that may need some fixing but that's why it is used as the name.

That said, the original jb1 code around line 1170 looks like it should pick up the general options default position and fix the disconnect between the refname and the genome name.
I tried setting them in my sample and AFAIK it generates a config.json that seems to as described in the documentation at https://jbrowse.org/jb2/docs/config_guides/default_session/ but does not seem to do the needful. 
Could you please double check in a downloaded archive if the config.json is right but not working?

fubar2:
Grrr. Now that I've fixed that, the genome name is wierd. Hang about.

fubar2:
> why does setting the default location break the session? 

It doesn't now - but your default location has to be something like `NC_00913.3:1..8999` and that period in the contig name was breaking the old regexp. Pretty sure that will fail in jbrowse1 with that regexp? Added a period and it seems to work for my VGP example.

hexylena:
> Much remains to do. Question is, when is a big complex tool ready to expose for proper testing?
> So far, only three of us have tried it AFAIK.

I would be open to shipping a first version when:

- [ ] top level is wrapped in an "assembly" repeat, *in planning for* future multiple genome builds, without **any** expectation of that working now/PAF/etc.
- [x] it opens on the first contig

the rest can be skipped. The checklist in my last issue, yeah, they're issues but they can wait, none of them are experience breaking. I do not expect synteny to be done soon, that will probably need some discusison with jbrowse regarding CLI tools for configuration that into a a good user experience in galaxy.

hexylena:
Oh, good work, colours are there, and it's opening on the locaion correctly!

image

That is fascinating behaviour though, it really subsets the genome, it doesn't just focus the entire genome's view there, it gives you an actual subset that you cannot scroll past the end of. Note the disabled zoom out box, you cannot zoom out. Hilarious.

Ok, then I change my sine qua non list to be:

- [ ] initial location renamed to "subset to display to new users", because it isn't an initial location, it's an initial subset. You have to change your viewed contig to fix that.
- [ ] Wrap it in an (unused) repeat for "assembly" with min=1/max=1.

hexylena:
`re.search(r"^(\w+):(\d+)\.+(\d+)$",`

that regex is quite restrictive, what about `re.search(r"^([^:]+):(\d+)\.+(\d+)$"`?

hexylena:
>  The paf code has been rewritten and so has the genome maker, so syntenic tracks can add genomes. The jbrowseConnector now has a list of genomes for that purpose. It keeps the default genome for non-synteny tracks like vcf that need a genome. It is set when the first assembly is added. May need rethinking but that's what's there. Problem is I have neither a paf file nor a clue about synteny, to test the new synteny track.

I mean, for my use case it would look like this:

- i have a reference genome
- I have a couple of related genomes I want to visualise in synteny
- as a user, I want to go to jbrowse and add my assembly + track, and then additional assemblies with their own datasets, all from my own data.

so, we'll need the ability to provide multiple "assemblies" eventually to the jb2 tool.

I don't think we can do the synteny on a decent time frame now, so, again let's just wrap it all in one big repeat for assemblies, and then in the future we'll tackle that in a good way. I just feel quite certain we'll want to support multiple assemblies, and if the tool already had the XML element for it, it won't break re-run for our users.

fubar2:
>  top level is wrapped in an "assembly" repeat, in planning for future multiple genome builds, without any expectation of that working now/PAF/etc.

The code is already a loop for a list, but the XML does not allow multiselect.
 
```
def process_genomes(self):
 
       for genome_node in self.genome_paths:
```
The code isn't changed much so `genome` list at jbrowseConnector construction items will all be compressed and indexed into the browser. The very first one will be used when a vcf or other track wants a genome. I'm not clear what to change - what's the use case? Can a user put tracks with two or more different reference sequences to good use other than synteny? How do the tracks know which reference they are supposed to be related to? Other than synteny, not sure how multiple genomes works. Even MAF all use the same reference.

For synteny, JB2 expects a paf and a list of genomes. It adds any new genomes ready for a track to the list from the paf form choice, ready for display - untested.

fubar2:
> so, we'll need the ability to provide multiple "assemblies" eventually to the jb2 tool.

You can. When a paf track is specified, it includes another genome. That gets indexed and added to the directory ready for tracks if the code works right which it probably does not.
image

fubar2:
Ah. Here's what I got by rerunning the most recent job in that history, setting the `default location` (name has been changed to subset as requested...) to `NC_000913.3:1..5000`

Remember, `genome.fa` is just your reference or assembly dbkey for MAF or similar uses. 
The display has to be a start..end of one specific contig or chromosome in that assembly.

image

hexylena (changes requested):
My previous review(s) have been looking only at the functionality of the UI/UX as a user, we've reached an agreement on that for all intents and purposes:

- I've ceded a default view as an upstream issue, and *all* of the other potential things I flagged, it's fine, don't do them.
- (except for the repeat I want wrapping the entire inputs section to support 'assemblies' as a top level concept to make it more future proof. If you cannot add this, tell me, I will once we get there. Ignoring for the test cases, it will be a 4 line diff.)

Now I'm looking at the code

Since the rebasing hasn't happened yet (or rather it did, and then was subsequently un-done), I'm looking primarily at a diff of the two branches, to see what's actually changed (`git diff -w pr-3997:tools/jbrowse2/jbrowse2.py pr-5697:tools/jbrowse2/jbrowse2.py`, if you have two appropriately named branches on hand.)

jbrowse.xml diff, compared to #3997

- mostly things that are supported in anthony's branch, are removed (cram)
- some new test cases added (yay!)
- most of the help text deleted (shrug. I don't think anyone was ever reading it.)

jbrowse.py diff

- clearly run through black (yay)
- all of the tracks are manually configured via JSON, rather than using the built in `add-track` command. (This is not good.)
- categories support lost
- assemblies support lost (rather, assemblies are made per-provided .fa file which is not desirable from my view.)

You've added some scripts and utilities and nice stuff (cool! it's nice! the webserver I think would be more generally useful as redbean, but I also don't think anyone will actually use this, so, I do not care, just leave it as is.)

That said, I do not think I can accept this PR, I am -1 on it as it stands.

Anthony's code is significantly cleaner, it has native support for assemblies (in the .py at least, already handling an edge case I noticed of naming collisions when testing myself with the CLI), and more features like CRAM support. I want all of that work to still be included. I can't accept it being dropped. Nor can I accept you asking him to PR his changes into your branch, when it should be the other way around.

The diff over what is currently present in @abretaud's PR is maybe interesting, but the replacement of `add-track` with more brittle json configuration is also sub-optimal, though I suspect a relic of your first PR where you mention using pointing inputs at URLs to avoid copies (I would have rejected that too, sorry, it will fail in far too many environments.). The result however, is that it relies too much on jbrowse2 internals when there's no need and they provide really nice CLI tools, whose cli interface is going to be more stable over time, and doesn't require reverse engineering their internal configuration, which makes adding new formats a cinch because they're already supported by the underlying add-track function.

*I really do want whatever useful changes you've made!!* I'm really happy you're motivated. I've probably killed a lot of that motivation, and I'm sorry for that, but I have my standards for what I approve, and I'd like to see _all_ of our goals met; you, me, GGA, Anthony. You've done great work getting the dependencies updated and worrying about a lot of the little details that needed to be done, Thank you.

Still, I feel really sad at how much code has been written, how *many* commits, how many adventures you've had with git, how many PRs that have been opened and closed, how much stress was generated for all of us. It's not fun for anyone, I'm sorry it happened, that sucks.

As you've told me, this is not a pleasant review process, it's not a pleasant feeling, and I can completely understand and empathise there, I think **all** three of us feel that way. I really had hoped that after the first discussion, this could have been rebased on @abretaud 's PR, and we could review the you know, 5 commits or whatever, and merge it. Instead we've had a lot of miscommunications that lead to long discussions (assemblies), and multiple branches opened and closed, dozens of commits made that have to be re-reviewed, and that has been really unfortunate. I don't know how we could work around that, if you have suggestions here for next time I am very open to what could change.

The way I see to move forward is to:

- open a PR to anthony's branch
- format his code with black as one commit, with no other changes, to make diffs clearer.
- add in the bits of your code that are missing to his, as _one_ big commit. I didn't want you to lose all of your commits, but, maybe rebasing is too complicated due to the very divergent histories.
- leave it, make no more changes, we will review that and try and merge it quickly for you.

Perhaps I shouldn't've suggested rebasing, perhaps I should've suggested something like this from the beginning.

fubar2:
@hexylena Thanks!
A lot to digest. 
Might be best to go back to #3997 and get it working properly, unless there's a better suggestion that doesn't involve me doing all the remediation needed for this one.

fubar2:
Closing this PR here and moving to https://github.com/usegalaxy-eu/temporary-tools/tree/master/jbrowse2 if anyone wants to contribute. Synteny is now available - see below.

At this point, it seems that at least two things are required that I can not or will not do to satisfy review:

- rebasing this PR to a previous independent pull request
  - #3997 lay idle in a non-working state for most of the year before I started on this code.
  - Annoying I didn't find it before I started, but that's how things have worked out.
  - Got a rebase done after much pfaffing about somehow, and then later screwed up the branch history totally. 
  - A PR to correct the historical record and port any missing fixes from all that work would be welcomed. 
  - I am in favour, but am neither motivated nor competent to do that work. 
- calls to the jbrowse CLI are required to replace most of the generated json. 
  - Generated json works well but at some point, the CLI did not, which is why it was swapped out.
  - The CLI requires less code, so simpler and cleaner approach *if* it can be made to work but otherwise same functionality.
  - The CLI for default session generation has long been known non-functional for at least a year, so json is needed.
  - assemblies seem to need json - even the old PR used it.
  - I have no specific objections to reverting to the CLI where it can be made to function so a PR would be welcomed.
  - I have no motivation to do that work. 
 
image

hexylena:
> Will continue development at [usegalaxy-eu/temporary-tools@master/jbrowse2](https://github.com/usegalaxy-eu/temporary-tools/tree/master/jbrowse2?rgh-link-date=2024-01-26T00%3A53%3A54Z) if anyone wants to contribute.

EU's temporary tools repo is **not** intended for outside contributors. It's just used for deploying existing repo (from elsewhere) to EU. I would recommend against it.

It is really unfortunate that you can't contribute on top of Anthony's PR. I guess the two of us will cherry-pick from your branch when we find time to continue that PR.

And in the end we will have two competing jbrowse2 implementations. I don't think that's the best result for the community, but ok, maybe that's the best solution in this case.

> swap out functional generated json for the CLI. 

If it's not always reliable, I think those are upstream issues! The whole community would benefit from us reporting.

> Multiple separate sets of assemblies and their tracks are mandatory for approval. 

I get that you don't have a use case, but I do, I'm sorry we can't reach agreement on that. 

for the record again, I was _not asking for full support for multiple assemblies_, but only a **2 line diff** (ok some some more to fix up test cases but all changes not requiring a lot of work), adding a repeat element I could help fill in later, that would help users in the future.

Thank you for the time spent on this discussion, I wish you much success with your implementation.

hexylena:
I would recommend changing your toolshed API key

https://github.com/galaxyproject/tools-iuc/pull/5697/files#diff-98cd68d15f58349113186306e07f32d7ffdbe4bd5adeb65d25baf7f1389b2c0e

fubar2:
>for the record again, I was not asking for full support for multiple assemblies, but only a 2 line diff (ok some some more to fix up test cases but all changes not requiring a lot of work), adding a repeat element I could help fill in later, that would help users in the future.

If I understand correctly, @hexylena mandates a non-functional repeat, so that when someone who cares gets around to adding all the code needed to make it work, future users will be able to rerun old jobs. Until then, the tool must remain broken, and users filling in the non-functional repeat elements will find they have zero impact. 

The reviewer insists that because I As a consequence of my refusal to break working code for purposes I don't understand, it seems there will be redundant independent re-development from a broken old PR.

Is there anything else that needs to be fixed? If not, why not approve this as is, and submit a PR to make it do exactly what you need when you find time.

abretaud:
Giving my point of view now as I didn't have much time to follow closely enough before. I have to say I'm very troubled to see this unfortunate turn of events.

First I'm 100% happy to see interest in a jbrowse2 tool, and can't wait to see it working. I need it for different projects, I have no interest in blocking or delaying anything.

I'm guilty of not finishing what I started in #3997, but I've always been completely open to anyone contributing to it, that's exactly why I opened a WIP PR. I think the jb code has also evolved in good ways since then, probably solving a few of the problems I encountered.

It's unfortunate that you hadn't seen #3997 before you started, but it was clearly not hidden, it was evoked at last GCC and various annotation meetings, and on GGA matric channel. But ok, that's how things have worked out, we can't change this now.

So you opened your new PR out of nowhere and have been super pushy to get it reviewed and merged quick. @hexylena and I have spent time reviewing it, although we had other plans. Now what I see is that :
- apart from a first rebase that was then lost in opened/closed PR and forced pushes, you haven't made more efforts to integrate your work with what had already been done in #3997. I know this is a huge PR, and it's not easy to do, but that's how it works: when you contribute you need to adapt to what is existing.
- you declined many of our comments with "A PR from someone more xxx is required" : again, you propose a contribution, you can ask politely for help but you can't just ignore other peoples' use cases
- I don't see how these comments were so hard to address, given that @hexylena proposed to focus on a shortlist, and that for CRAM it works essentially as BAM files and there's working code in #3997  (coded it in like 30min in my PR without ever using CRAM format before), and adding a top level repeat is maybe boring (renaming fields) but without any programming challenge
- you end up with an attack against @hexylena who have spent hours reviewing/testing your work and trying to find solutions constructively
- and finally you seem to continue working on your forked tool on a repo that has never been made for that, which will only bring confusion to us contributors, and to Galaxy users ("wait, which jbrowse2 tool did you use?").

I don't like any of this. I understand that there's pressure to get something out quick, but it's no reason for this kind of turn of events.

So yes, I'm troubled. I'm confident that the latest work from @hexylena will allow us to merge #3997 soon, I just hope that we can integrate your valuable work at some point.

fubar2:
This has indeed turned unpleasant but please understand I submitted a functional, useful PR before your work was pointed out. No matter what, I refuse to be bullied into making a change someone else wants that has no impact on what I am busy working on. 

As always, submit a PR, or do whatever you need to do. 
There are [10 track types currently working](https://github.com/fubar2/temporary-tools/tree/nohash/jbrowse2), and when the IUC reproduces that work, I'll use it. Who wrote it is all the same to me.

PR #5706, opened 2024-01-10
Extend `map_param_value` to datasets

bernt-matthias:
The current state allows to check 

- if a dataset is not given (by keeping also `from` empty + a default/fallback value)
- if its the same as another .. which probably is not useful

Altogether this allows to map optional datasets to True/False if the dataset is given or not.

I decided to go for an extension of the map value tool (and not a separate tool for which I would have copied the tool more or less (using macros of course)). You are right that the tool is already quite complex, but my impression is that the change does not make it worse. If you disagree I can still go for a separate tool.

Next step would be to do the same for collections, or? 

FOR CONTRIBUTOR:


mvdbeek (approved):
Ah, awesome, that's actually much simpler than I thought it would be!

PR #5713, opened 2024-01-11
pick_value: allow for an optional default

bernt-matthias:
the use case is to connect an optional workflow input to a mandatory tool input (the tool is conditionally executed only of a dataset is given - which is implemented via the map_value tool functionality https://github.com/galaxyproject/tools-iuc/pull/5706).

the optional WF input can not be connected to the mandatory tool input. with the pick value tool we can transform the optional input in a non-optional one, before this change a default value needed to be specified, i.e. an (empty) file needed to be provided somehow. now the tool does this automatically if no default is given.

potentially nice is that now for all input types the default parameter is optional.

FOR CONTRIBUTOR:


bgruening:
@bernt-matthias this needs a rebase

bernt-matthias:
This does not work without https://github.com/galaxyproject/galaxy/pull/17271

Closing this, because the current state of the tool also works once we have https://github.com/galaxyproject/galaxy/pull/17271.

PR #5722, opened 2024-01-18
update bracken to v2.9  and add subspecies level in filter (S1 and S2 level)

pimarin:
I just updated the tool version and add the level options S1 and S2.


bgruening (approved):
Awesome! You rock @pimarin!

Issue #5703, opened 2024-01-10
Names of log files from Trimmomatic don't correctly reflect inputs


AnnaSyme:
Issue with Trimmomatic output log files: 

Summary: can outputs be renamed differently? 

Running on Galaxy Australia, Trimmomatic version 0.36.6

Inputs: Paired-end reads: 2 collections (one of R1 files, one of R2 files)

Output required: Trimmed reads, and a collection of log files (not trimlog files). 

(Edited to update with a more general problem with output file naming, not only relevant to output log files)

Issue: when input files are named in the format `sample123_R1.fastqsanger` and `sample123_R2.fastqsanger`, output file names all take on the former name only, e.g. `sample123_R1.fastqsanger`

I think it would be better if outputs (where they refer to both R1 and R2 files) could be labelled without the R1 extension, such as `sample123.fastqsanger`

This would be clearer for the user, and also for downstream tools (such as plots in MultiQC, which are thus also labelled as being R1 files, when they are actually relating to both R1 and R2 files).

AnnaSyme:
I don't know if this is related to the issue about element identifiers? https://github.com/galaxyproject/tools-iuc/issues/1595

posting in case it might be.

bernt-matthias:
Guess the problem only occurs if `Paired-end (two separate input files)` is used (and not `Paired-end (as collection)`?

AnnaSyme:
I'm not sure @bernt-matthias, but in this case yes the input is `Paired-end (two separate input files)` (except that each of these inputs is, itself, a collection). I'll give it a try with a single input collection.

AnnaSyme:
Update: this tools is in a different repo - thanks @neoformit for pointing this out! - However it seems that it may be transferred here to the iuc quite sooon so perhaps any changes can take place here when it does.

neoformit:
Looks like it's [merged to IUC](https://github.com/galaxyproject/tools-iuc/pull/5698) now, I will take a look at this today.

neoformit:
Anna and I just had a chat about this. It seems more like a Galaxy Collections bug than tool wrapper issue:

![image](https://github.com/galaxyproject/tools-iuc/assets/42562517/0246723d-c231-48c0-9a3d-2b9f7e61d2ce)

1. The Trimmomatic job was submitted with paired reads in two collections: R1 and R2
2. The job outputs four collections: Unpaired R1, Unpaired R2, Paired R1, Paired R2
[...]

PR #5728, opened 2024-01-23
Fix trimmomatic outputs

neoformit:
* [x ] - This PR updates an existing tool or tool collection

Fixes https://github.com/galaxyproject/tools-iuc/issues/5703 by adding `default_identifier_source` to outputs.


bgruening:
We would need a version bump here.

bernt-matthias (approved):
Cool. We can merge when bumped.

We should open an issue stating that this needs to be done for some more tools.

PR #5731, opened 2024-01-25
update recentrifuge version to v1.13.2

pimarin:
Simple update to the last version of recentrifuge, do not update with the 1.13.0 version which has a bug


bgruening (approved):

bgruening:
Merci!

PR #5737, opened 2024-01-27
Upgrade htseq-count tool and fix the "(no feature output)" for compatibility with multiQC tool.

drosofff:
FOR CONTRIBUTOR:
---

I came to this fix because in its current state, the tool return a dataset "...(no feature)" which is the only one that can be used by multiQC, but then produces scaring results since the metrics "aligned reads" is missing in the dataset.

My changes:

- [x] upgrade htseq conda to "2.0.5"
- [x] upgrade samtools conda to "1.19.2"
- [x] add gawk 5.3.0 as requirement for reproducibility
- [x] add coreutils 9.4 to run csplit (see below)
- [x] drop one use of awk to split aligned and other counts. Use csplit instead, thank @wm75 !
- [x] drop the --format option of htseq since it is now autodetected
- [x] As format is autodetected, format all intermediary inputs as bam files
- [x] hardcode the gff/gtf reference file with a gtf extension to enforce outputs without quoted genes
- [x]  add bash code to sum aligned counts and print it in '$othercounts'
- [x] `$othercounts` becomes `$allcounts` for consistency
- [x] Adapt test output names accordingly
- [x] change the output name from ...(no feature) to ...: summary
- [x] removed the 2nd test that was a duplicate of the first one
- [x] revoved redundante 'name' attributes in inputs
- [x] bump tool version to 2.0.5+galaxy0 (using macro tokens)
- [x] updated README.rst, except for the "Galaxy Wrapper License" part


tools/htseq_count/htseq-count.xml
```diff
@@ -203,7 +203,7 @@
             <param name="ref_file" value="htseq-test_reference.fasta" />
             <output name="counts" file="htseq-test_counts.tsv" />
             <output name="othercounts" file="htseq-test_othercounts.tsv" />
-            <output name="samoutfile" file="htseq-test_samout.bam" />
+            <output name="samoutfile" file="htseq-test_samout.bam" compare="sim_size" delta_frac="0.03" />
```
bgruening, line 206: BAM files can be compared via normal diff. Galaxy will unpack BAM to SAM before doing the diff.

drosofff: yes, I was missing that in that case, the attribute `ftype` must be specified as `bam`

drosofff:
Should I change the variable name `othercounts` to `allcounts`, to be more meaningful @bgruening ?

bgruening:
> Should I change the variable name `othercounts` to `allcounts`, to be more meaningful @bgruening ?

Please do. It's a big version bump so we can cleanup some stuff. Also please use the latest IUC guidelines in version numbering. This will enable our bot to create new update PRs for this tool.

drosofff:
> Also please use the latest IUC guidelines in version numbering. This will enable our bot to create new update PRs for this tool.

Do you mean [there](https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#) @bgruening ?

bgruening:
Yes

drosofff:
I think the PR is ready to merge now. If not, please advise @bgruening

wm75:
A couple of remarks:
- the quoting of feature names seems to be a GFF version/GTF issue

  from my local testing, if I pass the `htseq-test.gff` file with a `.gtf` suffix explictly, then there's no quoting. Linking the gff input dataset to a properly named file according to its datatype might be a better solution than the sed

- a **much** more readable way than the awk script writing to stdout/stderr alternatively, would be a simple `| csplit -q - /^__/` (since the extra lines are all grouped at the end of htseq-count's output. This will write the first part of the output to a file `xx00`, the second part to `xx01`

- the htseq-count help says `--format  Type of <alignment_file> data. DEPRECATED: file format is detected automatically. This option is ignored.`

  Haven't tried but that would mean we shouldn't put the option on the command line anymore.

- If the format is indeed auto-detected and can be BAM, we should likely not have samtools sort write SAM, but BAM instead to save on disk storage.

- by now, the most recent upstream htseq version is 2.0.5, and it would be nice to target this here.

tools/htseq_count/htseq-count.xml
```diff
@@ -60,9 +65,12 @@
     --format=sam
     '$name_sorted_alignment_filename'
 
-    "$gfffile" | awk '{if ($1 ~ "no_feature|ambiguous|too_low_aQual|not_aligned|alignment_not_unique") print $0 | "cat 1>&2"; else print $0}'
+    "$gfffile" | awk '{if ($1 ~ "no_feature|ambiguous|too_low_aQual|not_aligned|alignment_not_unique|aligned") print $0 | "cat 1>&2"; else print $0}'
```
wm75, line 68: I'm not sure I understand this change. Is a lne with `aligned` ever produced by htseq-count, or did you intend to append your echo output to the original output?

tools/htseq_count/htseq-count.xml
```diff
@@ -1,11 +1,16 @@
...
+        <token name="@PROFILE@">23.0</token>
+    </macros>
     <xrefs>
         <xref type="bio.tools">htseq</xref>
     </xrefs>
     <requirements>
-        <requirement type="package" version="0.9.1">htseq</requirement>
-        <requirement type="package" version="1.7">samtools</requirement>
+        <requirement type="package" version="2.0.3">htseq</requirement>
+        <requirement type="package" version="1.19.2">samtools</requirement>
```
wm75, line 13: since different implementations of `awk` can differ in surprising ways, can you require `gawk` explicitly here?

tools/htseq_count/htseq-count.xml
drosofff: good point, it is the trace of an abandoned strategy...

drosofff:
> * from my local testing, if I pass the `htseq-test.gff` file with a `.gtf` suffix explictly, then there's no quoting.

Could you detail a little bit @wm75 ? in my local testing I still have quoting even pushing explicitly a .gtf suffix

wm75:
> > * from my local testing, if I pass the `htseq-test.gff` file with a `.gtf` suffix explictly, then there's no quoting.
> 
> Could you detail a little bit @wm75 ? in my local testing I still have quoting even pushing explicitly a .gtf suffix

Sure, I simply did a `cp test-data/htseq-test.gff test-data/htseq-test.gtf`, and while `htseq-count --mode=union --stranded=yes --minaqual=10 --type='exon' --idattr='gene_id'   --order=name --format=sam 'name_sorted_alignment.sam' htseq-test.gff` gives quoted feature names, `htseq-count --mode=union --stranded=yes --minaqual=10 --type='exon' --idattr='gene_id'   --order=name --format=sam 'name_sorted_alignment.sam' htseq-test.gtf` doesn't.

(These are essentially the command lines of wrapper test 1, but with a pregenerated name_sorted_alignment.sam.)

wm75:
Nice,  I've opened a [PR](https://github.com/ARTbio/tools-iuc/pull/1) against your topic branch with some additional suggestions.

wm75:
@bernt-matthias do you have an idea why the `samtools view -Su -t '${reference_fasta_filename}.fai'` dance is necessary twice - once for SAM input, once for the optional SAM output?

I assume the first is though to cover the case of a headerless input SAM,  but htseq-count seems to produce headered SAM output so why passing this through samtools view again?

Also the first use assumes the ref index is available, but typically it isn't (only if the user selects the additional BAM output. This path is only safe because samtools view silently continues if the index passed via `-t` doesn't exist.

bernt-matthias:
> do you have an idea why

No :(

From the samtools view docs:

> `-S` :  Ignored for compatibility with previous samtools versions. Previously this option was required if input was in SAM format, but now the correct format is automatically detected by examining the first few characters of input. 

> Also the first use assumes the ref index is available, but typically it isn't (only if the user selects the additional BAM output). This path is only safe because samtools view silently continues if the index passed via -t doesn't exist.

Isn't this always produced by: https://github.com/galaxyproject/tools-iuc/blob/26de736d8fad34eff7ad6ebd144041c05d1c21ba/tools/htseq_count/htseq-count.xml#L37

wm75:
Ok, I think I've figured out how SAM input and the optional reference really have to be handled. Feel free to merge my new [PR](https://github.com/ARTbio/tools-iuc/pull/2).

wm75 (approved):
Thanks @drosofff. I don't have any other ideas/objections.

@bgruening @bernt-matthias ready to merge from my side ...

drosofff:
Thanks @wm75 the wrapper is much better with your modifs !
Funnily enough, I never use htseq-count :-) I needed it for a training to transcriptome analysis...

bgruening:
Thanks everyone!

PR #5758, opened 2024-02-09
Add biotools for spades tools

bebatut:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/spades/macros.xml
+++ b/tools/spades/macros.xml
@@ -1,6 +1,6 @@
 <macros>
     <token name="@TOOL_VERSION@">3.15.5</token>
-    <token name="@VERSION_SUFFIX@">0</token>
+    <token name="@VERSION_SUFFIX@">1</token>
     <xml name="requirements">
         <requirements>
             <requirement type="package" version="@TOOL_VERSION@">spades</requirement>
@@ -504,10 +504,6 @@ export OMP_THREAD_LIMIT=\${GALAXY_SLOTS:-4} &&
             </param>
             <when value="true">
                 <expand macro="input_files_paired" format="fastq,fastq.gz,fastqsanger.gz,fasta,fasta.gz" label="FASTA/FASTQ file(s)" help="@HELP@"/>
-                <!--<param name="library_number" type="select" label="The samples belong to the same library" help="If the reads have been generated from the sample sample, it means that they belong to the same library.">
-                    <option value="true" selected="true">True</option>
-                    <option value="false">False</option>
-                </param>-->
             </when>
             <when value="false"/>
         </conditional>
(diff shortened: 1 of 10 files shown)


tools/spades/macros.xml
```diff
@@ -504,10 +504,10 @@ export OMP_THREAD_LIMIT=\${GALAXY_SLOTS:-4} &&
             </param>
             <when value="true">
                 <expand macro="input_files_paired" format="fastq,fastq.gz,fastqsanger.gz,fasta,fasta.gz" label="FASTA/FASTQ file(s)" help="@HELP@"/>
-                <param name="library_number" type="select" label="The samples belong to the same library" help="If the reads have been generated from the sample sample, it means that they belong to the same library.">
```
bgruening, line 507: I guess we need a version bump here.

bernt-matthias: Please just remove code instead of commenting. Can you explain why this is removed?

tools/spades/metaviralspades.xml
```diff
@@ -52,6 +55,7 @@ spades.py --metaviral
         </section>
         <expand macro="kmer"/>
         <expand macro="phred"/>
+        <expand macro="operation_mode"/>
```
bernt-matthias, line 58: Has this been forgotten before? Add a test?

tools/spades/macros.xml
bebatut: I commented to see if it was passing the tests and I forgot to remove it :(
The `input_files_paired` macro has already the `input_files_paired` param

tools/spades/metaviralspades.xml
bebatut: It seems it has been forgotten before. It was used in the output filters and also in the tests

bernt-matthias: Then we should also use `$operation_mode` on the command line -- as some others do. It might be worth to check if all tools actually support setting this .. depending on your time.

bernt-matthias: For two incompatibility is documented https://github.com/ablab/spades/blob/3eed982f35e4021439b3b97012898505543afb14/README.md?plain=1#L457

bebatut: but not for `metaviralspades` here. 
I did not plan to spend much time on it. I just wanted to add the bio.tool reference 😅 
To avoid an issue there (and leave that for an update of the tool), what about removing `$operation_mode` in `metaviralspades` from the param and test for now?

bernt-matthias: Maybe we are lucky and it just works if you add it to the command :) ... then it's even less work

bernt-matthias (approved):
Thanks a lot @bebatut

PR #5785, opened 2024-02-20
Add colabfold

astrovsky01:
Adds colabfold tool. Testing output requires data caches, so the tests here just check for command line content
FOR CONTRIBUTOR:


tools/colabfold/colabfold_msa.xml
```diff
@@ -0,0 +1,127 @@
...
+    #if $custom_template:
+        mkdir template_dir &&
+        #for $file in $custom_template:
+            ln -s $file "template_dir/${file.element_identifier}.pdb" &&
+        #end for
+    #end if   
+
+    #if $query_type.select_query_type == "fasta":
+        #set input_file = re.sub('[^\w\-_\.]', '_', str($query_type.input.element_identifier)) + '.fasta'
+        ln -s $query_type.input "$input_file" &&
```
bgruening, line 39: Sorry, I keep nagging you here, but please use single-quotes ;) and please remember :)

tools/colabfold/colabfold_msa.xml
```diff
@@ -0,0 +1,127 @@
+<tool id="colabfold_msa" name="colabfold_msa" version="@VERSION@+galaxy@VERSION_SUFFIX@" profile="21.05">
+    <description>This tool runs the alphafold step of the collabfold tool in Galaxy</description>
```
bgruening, line 2: the description in the help section reads different "        Generate MSAs for the alphafold step of Colabfold"

tools/colabfold/colabfold_msa.xml
```diff
@@ -0,0 +1,127 @@
...
+    <macros>
+        <token name="@VERSION@">1.5.3</token>
+        <token name="@CUDA_VERSION@">12.2.2</token>
+        <token name="@VERSION_SUFFIX@">1</token>
```
bgruening, line 6: 0

tools/colabfold/colabfold_msa.xml
```diff
@@ -0,0 +1,127 @@
...
+        <!-- Non functional in current release of colabfold, planned to expose in next one -->
+        <!-- <param name="pairing_strategy" label="Pairing strategy" type="select">
+            <option value="greedy">Greedy: MSA sequences should only be paired if the same species exist in at least two MSAs </option>
+            <option value="complete">Complete: MSA sequences should only be paired if the same species exist in all MSAs</option>
+        </param> -->
+        <param argument="--templates" label="Query PDB templates from the MSA server" type="boolean" truevalue="--templates" falsevalue=""/>
+        <param name="custom_template" label="List of pdb files to provide the prediciton as custom templates" type="data" format="pdb" multiple="true" optional="true"/>
+    </inputs>
+    <outputs>
+        <data name="output" format="tar" from_work_dir="output.tar" label="${tool.name} on ${on_string}: tar file"/>
```
bgruening, line 90: can we come up here with a specific datatype?

tools/colabfold/colabfold_alphafold.xml
```diff
@@ -0,0 +1,189 @@
...
+        <token name="@VERSION@">1.5.3</token>
+        <token name="@CUDA_VERSION@">12.2.2</token>
+        <token name="@VERSION_SUFFIX@">1</token>
+    </macros>
+    <requirements>
+        <container type="docker"> ghcr.io/sokrypton/colabfold:@VERSION@-cuda@CUDA_VERSION@</container>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[ 
+    #import os
+    ln -s $input input.tar &&
```
bgruening, line 13: quotes are missing here at all

tools/colabfold/colabfold_alphafold.xml
```diff
@@ -0,0 +1,189 @@
+<tool id="colabfold_alphafold" name="colabfold_alphafold" version="@VERSION@+galaxy@VERSION_SUFFIX@" profile="21.05">
+    <description>This tool runs the alphafold step of the collabfold tool in Galaxy</description>
+    <macros>
+        <token name="@VERSION@">1.5.3</token>
+        <token name="@CUDA_VERSION@">12.2.2</token>
+        <token name="@VERSION_SUFFIX@">1</token>
+    </macros>
+    <requirements>
+        <container type="docker"> ghcr.io/sokrypton/colabfold:@VERSION@-cuda@CUDA_VERSION@</container>
```
bgruening, line 9:
```suggestion
        <container type="docker">ghcr.io/sokrypton/colabfold:@VERSION@-cuda@CUDA_VERSION@</container>
```

tools/colabfold/colabfold_alphafold.xml
```diff
@@ -0,0 +1,189 @@
+<tool id="colabfold_alphafold" name="colabfold_alphafold" version="@VERSION@+galaxy@VERSION_SUFFIX@" profile="21.05">
+    <description>This tool runs the alphafold step of the collabfold tool in Galaxy</description>
+    <macros>
```
bgruening, line 3: should this be in a shared marco?

astrovsky01: No, because the tool is all under a single command, this intentionally splits it into two to allow running in different locations. The params are shared across both parts, but only applicable to this one

tools/colabfold/colabfold_alphafold.xml
```diff
@@ -0,0 +1,189 @@
...
+    tar -xmf input.tar --strip-components 1 -C input_data &&
+    mkdir output &&
+    colabfold_batch
+    #if $num_recycles.set_num_recycles == "manual"
+        --num-recycle $num_recycles.num_recycles
+    #end if
+    #if $recycle_tolerance.set_early_stop == "manual":
+        --recycle-early-stop-tolerance $recycle_tolerance.recycle_early_stop_tolerance
+    #end if
+    #if $num_ensemble:
```
bernt-matthias, line 24: You need to check for `str(param)!="" `, since 0 is evaluated to false.

tools/colabfold/colabfold_alphafold.xml
```diff
@@ -0,0 +1,189 @@
...
+        <param argument="--random-seed" label="Set seed" type="integer" min="0" optional="true"/>
+        <param argument="--num-seeds" label="Number of seeds" type="integer" min="0" optional="true" help="Number of seeds to try iterated based on random seed"/>
+        <param argument="--num-models" label="Number of models to use for structure prediction" type="select" help="Reducing the number of models speeds up the prediction but results in lower quality">
+            <option value="1">1</option>
+            <option value="2">2</option>
+            <option value="3">3</option>
+            <option value="4">4</option>
+            <option value="5">5</option>
+        </param>
+        <param name="max_msa" label="Max msa" type="select" help="Enable dropouts and increase number of seeds to sample predictions from uncertainty of the model. Decrease to increase uncertainity">
```
bernt-matthias, line 101: No idea what the label or the values are supposed to do. Can this be improved?

tools/colabfold/colabfold_alphafold.xml
```diff
@@ -0,0 +1,189 @@
...
+        --num-relax $amber.num_relaxed
+    #end if
+    $output_options.save_all
+    $output_options.save_recycles
+    $output_options.save_single_representations
+    $output_options.save_pair_representations
+    input_data
+    output
+    && cd output
+    && mv *.a3m output.a3m
```
TomHarrop, line 51: Can we test if these outputs are getting picked up by Galaxy? The current test is only checking the command line.

tools/colabfold/colabfold_msa.xml
```diff
@@ -0,0 +1,127 @@
...
+                <param name="select_query_type" value="a3m"/>
+                <param name="input" value="test.a3m"/>
+            </conditional>
+            <param name="pair_mode" value="paired"/>
+            <param name="templates" value="--templates"/>
+            <param name="custom_template" value="test.pdb,test_2.pdb"/>
+            <assert_command>
+                <has_text text="colabfold_batch --msa-only --pair-mode paired --templates --custom-template-path template_dir"/>
+            </assert_command>        
+        </test>
```
TomHarrop, line 119: Can we test the outputs, e.g. 

```
<test expect_num_outputs="1">
    <conditional name="query_type">
        <param name="select_query_type" value="fasta"/>
        <param name="input" value="test.fasta"/>
        <conditional name="select_db">
            <param name="use_db" value="yes"/>
            <param name="msa_mode" value="mmseqs2_uniref"/>
        </conditional>
    </conditional>
    <output name="output" file="test2.tar"/>
</test>
```
or even just check the file size?

tools/colabfold/colabfold_alphafold.xml
```diff
@@ -0,0 +1,189 @@
...
+        </collection>
+        <collection name="json_files" type="list" format="json" label="${tool.name} on ${on_string}: JSON predictions">
+            <discover_datasets format="json" pattern="__name_and_ext__" directory="output/json_out"/>
+        </collection>
+        <collection name="pdb" type="list" format="pdb" label="${tool.name} on ${on_string}: PDB predictions">
+            <discover_datasets format="pdb" pattern="__name_and_ext__" directory="output/pdb_out"/>
+        </collection>
+        <collection name="pickle" type="list" format="pickle" label="${tool.name} on ${on_string}: Pickle file outputs">
+            <discover_datasets format="pickle" pattern="__name_and_ext__" directory="output/pickle_out"/>
+            <filter>output_options['save_all'] == "--save-all"</filter>
```
wm75, line 141:
```suggestion
            <filter>output_options['save_all']</filter>
```

see https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-outputs-data-filter

tools/colabfold/colabfold_alphafold.xml
```diff
@@ -0,0 +1,189 @@
...
+            <option value="256:512">256:512</option>
+            <option value="64:128">64:128</option>
+            <option value="32:64">32:64</option>
+            <option value="16:32">16:32</option>
+        </param>
+        <param argument="--use-dropout" label="Use dropouts" type="boolean" truevalue="--use-dropout" falsevalue="" help="Activate dropouts during inference to sample from uncertainity of the models."/>
+        <conditional name="amber">
+            <param name="use_amber" label="Use amber" type="select" help="Use amber for structure refinement">
+                <option value="yes">Use amber</option>
+                <option value="no">Don't use Amber</option>
```
wm75, line 113:
```suggestion
                <option value="no">Don't use amber</option>
```

tools/colabfold/colabfold_msa.xml
```diff
@@ -0,0 +1,127 @@
...
+        <container type="docker"> ghcr.io/sokrypton/colabfold:@VERSION@-cuda@CUDA_VERSION@</container>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[ 
+    #import re
+    #if $custom_template:
+        mkdir template_dir &&
+        #for $file in $custom_template:
+            ln -s $file "template_dir/${file.element_identifier}.pdb" &&
+        #end for
+    #end if
```
wm75, line 35: Is a <configfile> a more elegant solution for setting up the templates?

bernt-matthias: And otherwise `element_identifier` needs cleaning (maybe even if an configfile is used).

tools/colabfold/colabfold_alphafold.xml
```diff
@@ -0,0 +1,189 @@
...
+        </collection>
+        <!-- <collection name="npy" type="list" format="npy" label="${tool.name} on ${on_string}: Numpy embeddings">
+            <discover_datasets format="npy" pattern="__name_and_ext__" directory="output/npy_out"/>
+            <filter>output_options['save_single_representations'] == "save-single-representations" or output_options['save_pair_representations'] == "save-pair-representations"</filter>
+        </collection> -->
+        <data name="a3m_out" format="a3m" from_work_dir="output/output.a3m" label="${tool.name} on ${on_string}: a3m file"/>
+    </outputs>
+    <tests>
+        <test>
+            <param name="input" value="input.tar"/>
```
wm75, line 151: I'm confused by this test .tar. What's inside it looks to me like the *output* of colabfold_alphafold, not like an input for it.

tools/colabfold/colabfold_msa.xml
```diff
@@ -0,0 +1,127 @@
...
+                <expand macro="db_selector"/>
+            </when>
+            <when value="a3m">
+                <param name="input" type="data" format="a3m" label="Query sequence"/>
+            </when>
+        </conditional>
+        <param name="pair_mode" label="Pair mode" type="select">
+            <option value="unpaired_paired">Attempt to pair sequences from the same operon within the genome</option>
+            <option value="paired">Only used sequences that were successfully paired</option>
+            <option value="unpaired">Generated seperate MSA for each protein</option>
```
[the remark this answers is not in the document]
wm75: and separate

tools/colabfold/colabfold_alphafold.xml
astrovsky01: It's a small tar file, since the output is not testing for content but for command line, and empty files are not allowed

tools/colabfold/colabfold_msa.xml
```diff
@@ -0,0 +1,127 @@
...
+            <param name="templates" value="--templates"/>
+            <param name="custom_template" value="test.pdb,test_2.pdb"/>
+            <assert_command>
+                <has_text text="colabfold_batch --msa-only --pair-mode paired --templates --custom-template-path template_dir"/>
+            </assert_command>        
+        </test>
+    </tests>
+    <help><![CDATA[
+        Generate MSAs for the alphafold step of Colabfold
+    ]]></help>
```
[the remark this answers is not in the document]
astrovsky01: The docs are strange on this one -- it's originally intended to be run from a notebook with all parameters being entered or pre-filled out. This version doesn't have docs per-se, and the help text/explanations from the original notebook are what I've mostly used for the param descriptions.

tools/colabfold/colabfold_alphafold.xml
```diff
@@ -0,0 +1,189 @@
...
+        <token name="@VERSION_SUFFIX@">1</token>
+    </macros>
+    <requirements>
+        <container type="docker"> ghcr.io/sokrypton/colabfold:@VERSION@-cuda@CUDA_VERSION@</container>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[ 
+    #import os
+    ln -s $input input.tar &&
+    mkdir input_data &&
+    tar -xmf input.tar --strip-components 1 -C input_data &&
```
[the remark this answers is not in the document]
astrovsky01: I think I am going to need to make a colabfold_tar datatype like @bgruening suggested. The whole directory of the outputs from the msa step is the true input for the alphafold step

tools/colabfold/colabfold_alphafold.xml
astrovsky01: On this tool specifically, we cannot. It needs cached data which cannot be used in the planemo checks on github

tools/colabfold/colabfold_alphafold.xml
bernt-matthias: I would expect that the tool is failing then (is the Galaxy tool detecting tool failures correctly) .. we can still test for the command line?

bernt-matthias:
Just stumbled over https://academic.oup.com/bioinformatics/article/39/1/btac749/6839971 .. wondering if the collabfold output could be used for this as is.

astrovsky01:
@bernt-matthias That's using a different part of colabfold than we use here, specifically Search instead of Batch. Part of why I made it a suite is that there will be different future versions of the tool depending on where it's being run, and that is one of them. If that is made into a tool, the output should be useable, though, from what I understand

astrovsky01:
Also, the shed error that's happening here is not occurring locally, and I'm unsure how to fix

tools/colabfold/colabfold_alphafold.xml
```diff
@@ -0,0 +1,160 @@
...
+            <filter>output_options['save_all']</filter>
+        </collection>
+        <collection name="npy" type="list" format="npy" label="${tool.name} on ${on_string}: Numpy embeddings">
+            <discover_datasets format="npy" pattern="__name_and_ext__" directory="output/npy_out"/>
+            <filter>output_options['save_single_representations'] or output_options['save_pair_representations']</filter>
+        </collection>
+        <data name="a3m_out" format="a3m" from_work_dir="output/output.a3m" label="${tool.name} on ${on_string}: a3m file"/>
+    </outputs>
+    <tests>
+        <test expect_exit_code="1">
```
bernt-matthias, line 123: The linter expects `expect_num_outputs` here. Can this be added, or does it make no sense in combination with `expect_exit_code` (`!="0"`)?

In the later case: its a warning that we could ignore .. and I should fix the linter.

astrovsky01: This part of the tool shouldn't be running, and is failing how I would expect it to, as it needs access to a cache that is mounted to the container, so I don't think that the expect_num_outputs would make sense here

bernt-matthias: The outputs should be created anyway. So I guess we can still use it. Would be great if we can verify this here. 

I see. I'm wondering if we can move this tool (and a few others like `gtdbtk`) that require large reference data (I guess its the alphafold reference data) to a special repo that has access to cvmfs in the tool tests (like we do in iwc). 

If carefully used it might be worth to try it. I remember that @mvdbeek always did not like to enable this for the general repo .. I have the feeling that having proper tests may outweight the downsides.

astrovsky01: In that case, I can add expected outputs for the discovered collections. And yes, it's the alphafold reference data

astrovsky01:
Current failure is on the msa step is due to a new param I added to standardize naming. Easily fixed, but requires minor testing first

astrovsky01:
@bernt-matthias The fail for the msa step is the same as it was before, but I can't even tell what's going wrong with the alphafold step, it looks like it's just shutting down immediately?

bernt-matthias:
Do you know how much memory is used in the msa step test?

astrovsky01:
@bernt-matthias @bgruening the most I have found that gives a hint is `Tool colabfold_alphafold created job None (16.176 ms)` in the test log

bernt-matthias:
Can this be reproduced locally? Often this indicates a problem with setting the metadata of the output.

astrovsky01:
@bernt-matthias  I'm having trouble updating the docker image locally to the newest version, despite it working fine on test, so I haven't actually been able to test the newest version of the tool locally

astrovsky01:
This fix seemed to have worked. The fail for the msa is my fault for filenames and I'm fixing quickly. The alphafold step messages are not errors, but the fact that it failed seems to mean that it's not respecting the expect_error_code tag still

bernt-matthias:
Would be interesting to see if `expect_failure="true"` helps .. and if this is the case which of the other asserts (exit code, commandline, ...) still works, i.e. is checked.

bernt-matthias:
> This part of the tool shouldn't be running, and is failing how I would expect it to, as it needs access to a cache that is mounted to the container

Could you detail a bit how this is supposed to be setup in Galaxy? How about non-containerized installations? And how about my (wild) idea with tool tests using cvmfs?

astrovsky01:
1. I'll try the expect failure, though it is telling me I can't have both expect_failure and expect_num_outputs
2. These tools will be on servers and will be limited for people to run by the admins to prevent overloading the msa server it is running against. The second version of this tool will run on hardware we have recently purchased. This tool is meant much less for a local user. The cvmfs test is a good idea and I want to include in in the next version, which will come quickly, since it's meant to run slightly differently, though versions are good and useful to have

bernt-matthias:
> These tools will be on servers 

Yep. Any Galaxy server should be supported, or?

> and will be limited for people to run by the admins to prevent overloading the msa server it is running against. 

Setting up this MSA server should be documented as well as the the necessary job configuration, or?

astrovsky01:
> Yep. Any Galaxy server should be supported, or?
It requires playbook setup for the docker command, so major servers and ones with admins who can work with that 
> Setting up this MSA server should be documented as well as the the necessary job configuration, or?
Not for this, no. The MSA server isn't running on our hardware Colabfold runs on a server our of Korea, where the MSA data is kept, and that is why there are concerns about overwhelming the server. That step is why this is so much faster than just alphafold. It also requires an extremely large amount of RAM to be held. The US server will have access to resources that will allow an MSA server for Galaxy, which will come in the second version of the tool.

astrovsky01:
Is there a chance that the issue here is that I screwed up on the datatype and it's not seeing it as an archive? When I ls the directory in the command, the files are there and named properly for the msa step that is failing

bernt-matthias:
> Is there a chance that the issue here is that because it can't find the colab.tar datatype, it's not seeing it as an archive? When I ls the directory in the command, the files are there and named properly for the msa step that is failing

Maybe an easy test is to run the test with `tar`. If it works we have to backport the datatype to 24.0.

bernt-matthias:
> The MSA server isn't running on our hardware Colabfold runs on a server our of Korea, where the MSA data is kept, and that is why there are concerns about overwhelming the server.

IMO this is why we should document that the usage needs to be restricted and how this is / can be done. Because every Galaxy can and probably will install it. So it's good (but maybe not sufficient) if org is limiting usage.

astrovsky01:
> Maybe an easy test is to run the test with `tar`. If it works we have to backport the datatype to 24.0.

It didn't find the files, and they definitely look like they're there when I ls. Am I missing something on has_archive_member?

tools/colabfold/colabfold_msa.xml
```diff
@@ -0,0 +1,115 @@
...
+    ## --pair-strategy $pairing_strategy
+    $templates
+    #if $custom_template:
+        --custom-template-path template_dir
+    #end if
+    --jobname-prefix "galaxy"
+    "$input_file"
+    output.colab &&
+    && ls output.colab
+    tar -cf output.colab.tar output.colab
```
bernt-matthias, line 40:
```suggestion
    ls output.colab &&
    tar -cf output.colab.tar output.colab
```

tools/colabfold/colabfold_msa.xml
```diff
@@ -0,0 +1,115 @@
...
+        #end if
+    #end if
+    --pair-mode $pair_mode
+    ## --pair-strategy $pairing_strategy
+    $templates
+    #if $custom_template:
+        --custom-template-path template_dir
+    #end if
+    --jobname-prefix "galaxy"
+    "$input_file"
```
bernt-matthias, line 37: Use single quotes (except if you use bash variables).

astrovsky01: FIxed locally, will be reflected in the next push

bernt-matthias:
Can you try `./galaxy_0_all/msa.sh` (or `.*galaxy_0_all/msa.sh` ). 

From the docs: 

> full paths of the members may be prefixed with ./

astrovsky01:
> full paths of the members may be prefixed with ./

I'd seen both used in iuc so I wasn't sure. One sec

tools/colabfold/colabfold_msa.xml
```diff
@@ -0,0 +1,115 @@
...
+            <option value="paired">Only used sequences that were successfully paired</option>
+            <option value="unpaired">Generate separate MSA for each protein</option>
+        </param> 
+        <!-- Non functional in current release of colabfold, planned to expose in next one -->
+        <!-- <param name="pairing_strategy" label="Pairing strategy" type="select">
+            <option value="greedy">Greedy: MSA sequences should only be paired if the same species exist in at least two MSAs </option>
+            <option value="complete">Complete: MSA sequences should only be paired if the same species exist in all MSAs</option>
+        </param> -->
+        <param argument="--templates" label="Query PDB templates from the MSA server" type="boolean" truevalue="--templates" falsevalue=""/>
+        <param name="custom_template" label="List of PDB files to provide the prediciton as custom templates" type="data" format="pdb" multiple="true" optional="true"/>
```
bernt-matthias, line 67: Would it make sense to re-use the pdb files that should be available from alphafold reference data?

astrovsky01: I don't think so. My understanding is that this is telling the tool a starting place from data you have, rather than a large number of files

tools/colabfold/colabfold_msa.xml
```diff
@@ -0,0 +1,115 @@
...
+                    <param name="use_db" value="yes"/>
+                    <param name="msa_mode" value="mmseqs2_uniref"/>
+                </conditional>
+            </conditional>
+            <assert_command>
+                <has_text text="colabfold_batch --msa-only --msa-mode mmseqs2_uniref --pair-mode unpaired_paired"/>
+            </assert_command>
+            <output name="output">
+                <assert_contents>
+                    <has_archive_member path="galaxy_0_all/msa.sh"/>
```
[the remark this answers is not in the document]
astrovsky01: This was also intended to test specific naming schema that was added via a hard coded naming parameter. I can add in the extra test for number of files, too, once I get this current one to work, but I see those as testing different parts

astrovsky01:
Woo passing! Before any merge, though, I wanted to ask for clarification about the expect_failure. Does that still test the command line against the assert_command I have, or is it just expecting a non-zero error code? I can't tell.

astrovsky01:
@bernt-matthias @bgruening intentionally breaking the command assertion failed, which means the test is doing what I thought it was, and the tool is ready

bernt-matthias:
> intentionally breaking the command assertion failed

Thanks for testing this. Can you test for the exit code?

astrovsky01:
> Thanks for testing this. Can you test for the exit code?

It's failing with exit code 1, which is what the test is specifying, but the test itself is still failing despite that

tools/colabfold/colabfold_alphafold.xml
```diff
@@ -122,7 +122,7 @@
         <data name="a3m_out" format="a3m" from_work_dir="output/output.a3m" label="${tool.name} on ${on_string}: a3m file"/>
     </outputs>
     <tests>
-        <test expect_failure="true">
+        <test expect_exit_code="1">
```
bernt-matthias, line 125: I meant

```suggestion
        <test  expect_failure="true" expect_exit_code="1">
```

astrovsky01:
The linter will not allow both expect_num_outputs and expect_fail, so this is what I can do with the test

tools/colabfold/colabfold_alphafold.xml
```diff
@@ -0,0 +1,158 @@
+<tool id="colabfold_alphafold" name="colabfold alphafold" version="@VERSION@+galaxy@VERSION_SUFFIX@" profile="21.05">
```
bernt-matthias, line 1: The name is supposed to be lower case? (fine with me)

tools/colabfold/colabfold_alphafold.xml
```diff
@@ -0,0 +1,158 @@
...
+    <description>AlphaFold step of colabfold in Galaxy</description>
```
bernt-matthias, line 2:
```suggestion
    <description>AlphaFold step of colabfold</description>
```

Still kind of super redundant with the tool name. The tool menu will show `colabfold alphafold: AlphaFold step of colabfold`

tools/colabfold/colabfold_alphafold.xml
```diff
@@ -0,0 +1,158 @@
...
+    #end if
+    #if str($advanced.recycle_early_stop_tolerance)!="":
+        --recycle-early-stop-tolerance $advanced.recycle_early_stop_tolerance
+    #end if
+    #if $advanced.num_ensemble:
+        --num-ensemble $advanced.num_ensemble 
+    #end if
+    #if $random_seed:
+        --random-seed $advanced.random_seed 
+    #end if
```
bernt-matthias, line 25: Should be `$advanced.random_seed` and also be tested for `str($advanced.random_seed)!=""`? 

The above test for just `num_ensemble` (without the `str`) should be fine (because we can not have 0), but you could still do it for consistency.

tools/colabfold/colabfold_alphafold.xml
```diff
@@ -0,0 +1,158 @@
...
+    #if $advanced.num_seeds:
```
bernt-matthias, line 26: Same. And check also below.

tools/colabfold/colabfold_alphafold.xml
```diff
@@ -0,0 +1,158 @@
...
+        --num-seeds $advanced.num_seeds 
+    #end if    
+    #if $advanced.num_models:
+        --num-models $advanced.num_models
+    #end if
+    $advanced.use_dropout
+    #if $advanced.max_msa:
```
bernt-matthias, line 33: This seems to be a non-optional parameter. So no if needed. Or should it be optional (the commented `auto` option indicates this).

tools/colabfold/colabfold_msa.xml
```diff
@@ -0,0 +1,115 @@
...
+    <expand macro="requirements" />
+    <command detect_errors="exit_code"><![CDATA[ 
+    #import re
+
+    ## Symlinking and formatting
+    #if $custom_template:
+        mkdir template_dir &&
+        #for $file in $custom_template:
+            #set input_file = re.sub('[^\w\-_\.]', '_', str($file.element_identifier)) + '.pdb'
+            ln -s $file "template_dir/${file.element_identifier}.pdb" &&
```
bernt-matthias, line 16: This also needs single quotes.

tools/colabfold/colabfold_msa.xml
```diff
@@ -0,0 +1,115 @@
...
+    --pair-mode $pair_mode
+    ## --pair-strategy $pairing_strategy
+    $templates
+    #if $custom_template:
+        --custom-template-path template_dir
+    #end if
+    --jobname-prefix 'galaxy'
+    '$input_file'
+    output.colab &&
+    ls output.colab &&
```
bernt-matthias, line 39:
```suggestion
```

tools/colabfold/macros.xml
```diff
@@ -0,0 +1,37 @@
+<macros>
+        <token name="@VERSION@">1.5.5</token>
```
bernt-matthias, line 2: Can you remove one indentation level for the tokens.

bernt-matthias (commented):
Gave it one last round. Sorry.

bernt-matthias (approved):

bernt-matthias:
Thanks for the efforts

bernt-matthias:
Anything left from your side @bgruening?

bgruening (approved):

bernt-matthias:
We have this one again:

```
Unexpected HTTP status code: 500: {"err_msg": "Metadata may have been defined for some items in revision '66cd8f416122'.  Correct the following problems if necessary and reset metadata.<br/><b>colabfold_alphafold.xml<\/b> - parameter 'num_models': the attribute 'value' must be set for non optional parameters<br/>"}
```

we had this on matrix recently, but forgot the conclusion...

bernt-matthias:
Should actually be allowed https://github.com/galaxyproject/galaxy/pull/16966 ..

bernt-matthias:
Quick fix is to set `value=""`.

astrovsky01:
I set it to optional here https://github.com/galaxyproject/tools-iuc/pull/5908

PR #5787, opened 2024-02-21
compleasm: Fix the lineage list problem

rlibouba:
Hello, 
I've opened this PR to correct a problem I noticed once the tool was available on usegalaxy.eu. The list with the different lineages did not appear. I made some modifications (using busco as an example) to correct this.


tools/compleasm/compleasm.xml
```diff
@@ -11,7 +11,7 @@
     -a '$input'
     -o galaxy
     --mode $mode
-    -l '$busco_database'
```
abretaud, line 14: Hum, lineage_dataset is a subdir of busco_database right?

rlibouba: lineage_dataset corresponds to the list of lineages defined in macros.xml. Perhaps the name of the parameter isn't chosen correctly?

bgruening:
ping @abueg 
I guess this is the error you are seeing.

abretaud (approved):
Nice, should be ok now

bgruening:
Its failing with:

> OSError: [Errno 30] Read-only file system: '/home/runner/work/tools-iuc/tools-iuc/tools/compleasm/test-data/test-db/busco_downloads/lineages/file_versions.tsv.tmp'

I guess we need to copy this file into the JWD?

abretaud (approved):

Issue #5812, opened 2024-03-04
fastp is running forever


bgruening:
We have a few fastp jobs that are running forever. All of them have this in the logs ...

```
galaxy@sn06:/data/jwd02f/main/066/794/66794680$ cat outputs/tool_stderr 
Detecting adapter sequence for read1...
No adapter detected for read1

Detecting adapter sequence for read2...
No adapter detected for read2

ERROR: sequence and quality have different length:
@A00730:926:HWF77DSX7:2:1232:20636:17660 1:N:0:ACGCACCT+CCTTCACC
CAGGG


WARNNIG: different read numbers of the 24507 pack
Read1 pack size: 200
Read2 pack size: 1000

```

Should we turn this into an error?

PR #5813, opened 2024-03-04
update fastp and fix linting and testing issues

bgruening:
fixes: https://github.com/galaxyproject/tools-iuc/issues/5812

FOR CONTRIBUTOR:


bgruening:
Even our test data had this bug, I included a test that should fail.

bernt-matthias (approved):

tools/fastp/fastp.xml
```diff
@@ -25,19 +25,19 @@
     #set $in2_name = re.sub('[^\w\-\s]', '_', str("%s_%s" % ($single_paired.paired_input.name, "R2"))) + $ext
     #set out1 = $output_paired_coll.forward
     #set out2 = $output_paired_coll.reverse
-    ln -s '$in1' '$in1_name' &&
-    ln -s '$in2' '$in2_name' &&
+    ln -sf '$in1' '$in1_name' &&
```
mvdbeek, line 28: Is that the fix to your issue, did you need the force for resubmitted  jobs ? That seems like it would need a framework level fix, it doesn't seem like something we should fix within tools.

bernt-matthias: Right, just to be sure, resubmitted jobs are supposed to get a clean working dir, or? 

Then the `-f` could only "help" with name collisions of the inputs (for which it might be better to fail).

bgruening: Nope, this was not the fix.

Resubmission from Galaxy gets a clean job_working_dir ... I think / hope / have seen it :) ... but job schedulers can also resubmit jobs. In this case we do not get a clean job dir I think.

Either way, this was not the fix, the update was the fix. The latest version is supposed to fail when the fastq files do not match up. I recommend updating this tool. For us, a few fastp jobs are blocking slots for days...

bernt-matthias: Right, eu uses htcondor retry/resubmit 

> but job schedulers can also resubmit jobs. In this case we do not get a clean job dir I think.

most likely not. 

Maybe it would be cleaner to add a `rm -rf working/*` to the script referred to by `environment_setup_file`? The `environment_setup_file` is sourced in the job script (`galaxy_xyz.sh`), i.e. it would run with every submission.

PR #5831, opened 2024-03-11
Added bio.tools ref to miniprot

RZ9082:
This PR links miniprot to its bio.tools [entry](https://bio.tools/miniprot).


bebatut (approved):
Thanks @RZ9082

PR #5832, opened 2024-03-11
Added bio.tools ref to pycoqc

RZ9082:
This PR links pycoqc to its bio.tools [entry](https://bio.tools/pycoqc).


bebatut (approved):
Thanks @RZ9082

PR #5843, opened 2024-03-11
Add bio.tool for krocus

rlibouba:
Hi all,

This PR links the tools to its bio.tools entry: https://bio.tools/krocus
It will be useful to get EDAM ontology annotations

Have a good day!

### DIFF
--- a/tools/krocus/krocus.xml
+++ b/tools/krocus/krocus.xml
@@ -33,7 +33,7 @@
             </options>
         </param>
         <section name="opt_args" title="Optional Arguments">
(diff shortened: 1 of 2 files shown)


tools/krocus/krocus.xml
```diff
@@ -33,7 +33,7 @@
             </options>
         </param>
         <section name="opt_args" title="Optional Arguments">
-            <param name="filtered_reads" type="boolean" truevalue="" falsevalue="" checked="false" label="Save filtered reads ?"/>
```
bgruening, line 36: any reason why you removed the falsevalue? I don't think it will do any harm, just wondering.

PR #5853, opened 2024-03-12
Add bio.tools ref to kraken

RZ9082:
This PR links kraken to its bio.tools [entry](https://bio.tools/kraken).


bebatut (approved):
Thanks @RZ9082

PR #5860, opened 2024-03-13
Add bio.tools ref to crisprstudio

RZ9082:
This PR links crisprstudio to its bio.tools [entry](https://bio.tools/crisprstudio).


bebatut (approved):
Thanks @RZ9082

PR #5863, opened 2024-03-13
Explicitly set tmp dir for fasterq-dump

mvdbeek:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/sra-tools/fasterq_dump.xml
+++ b/tools/sra-tools/fasterq_dump.xml
@@ -2,6 +2,7 @@
     <description>format from NCBI SRA</description>
     <macros>
         <import>macros.xml</import>
+        <token name="VERSION_SUFFIX">1</token>
     </macros>
     <expand macro="edam_ontology"/>
     <expand macro="bio_tools"/>
@@ -13,7 +14,9 @@
     @CONFIGURE_RETRY@
     @SET_ACCESSIONS@
     while [ \$SRA_PREFETCH_ATTEMPT -le \$SRA_PREFETCH_RETRIES ] ; do
-        fasterq-dump "\$acc" -e \${GALAXY_SLOTS:-1}
+        fasterq-dump "\$acc"
+        -e \${GALAXY_SLOTS:-1}
+        -t \${TMPDIR}
         --seq-defline '$adv.seq_defline'
         --qual-defline '+'
         $adv.split


tools/sra-tools/fasterq_dump.xml
```diff
@@ -13,7 +14,9 @@
     @CONFIGURE_RETRY@
     @SET_ACCESSIONS@
     while [ \$SRA_PREFETCH_ATTEMPT -le \$SRA_PREFETCH_RETRIES ] ; do
-        fasterq-dump "\$acc" -e \${GALAXY_SLOTS:-1}
+        fasterq-dump "\$acc"
+        -e \${GALAXY_SLOTS:-1}
+        -t \${TMPDIR}
```
bernt-matthias, line 19: Maybe 

```suggestion
        -t \${TMPDIR:./}
```
or 

```suggestion
        -t \${TMPDIR:\$_GALAXY_JOB_TMP_DIR}
```

mvdbeek: We're always setting it, so the fallback seems confusing ?

tools/sra-tools/fasterq_dump.xml
```diff
@@ -2,6 +2,7 @@
     <description>format from NCBI SRA</description>
     <macros>
         <import>macros.xml</import>
+        <token name="VERSION_SUFFIX">1</token>
```
bernt-matthias, line 5: Does this "overwrite" the token in macros.xml? 

Is this a good idea?

mvdbeek: yes. why should we bump every tool if only one has changed ?

mvdbeek:
```
Downloading accession: SRR11953971...
Failed to call external services.
Prefetch attempt 1 of 3 exited with code 1
Failed to call external services.
Prefetch attempt 2 of 3 exited with code 1
Failed to call external services.
Prefetch attempt 3 of 3 exited with code 1
```
that's fun ...

tools/sra-tools/fasterq_dump.xml
bernt-matthias: I understand the argument, but IMO this is easily missed when updating the macros.

mvdbeek: Isn't that handled by the bot ? What's the process to just bump one tool ?

bernt-matthias: Sometimes people bump manually :)

mvdbeek: ok, so now what ? I can drop `@VERSION_SUFFIX@ ` and put a literal 1 if that's any better

bernt-matthias:
Seems that this did not work. At least the [toolshed](https://toolshed.g2.bx.psu.edu/view/iuc/sra_tools/1ec814014652) still reports this as `galaxy0`. Also the complete repo including all three tools has been updated, or? Was a bit surprised that fasterq-dump can't be found at all as a single tool.

PR #5867, opened 2024-03-15
Add sam/bam/cram output options to nucmer

fubar2:
FOR CONTRIBUTOR:

Mummer is old and github untouched for >4 years but still seems popular and has a potential use for self-homology mapping in the VGP. Nucmer is the nucleotide mapper. The default output is "delta" format but bam/cram alignment formats would be more useful downstream. This PR adds output conversion to ~sam~bam/cram as an option. 

Nucmer now has `--sam-long` and `--sam-short` output options. 
Both produce _**sam headers with spaces instead of tabs**_ so samtools complains that the header cannot be read unless they are fixed. 

Update: Removed all the sam and the sam-short based outputs and added a routine samtools calmd step to prevent downstream Mummer sam format errors.

~Cannot get gnuplot tests to pass CI here so unless there are strong objections or an easy fix, will remove all the plotting functions and encourage downstream visualisation of the bam/cram with JBrowse~


fubar2:
oh dear. nucmer is fine but need to fix the rest of the broken xml and tests.

fubar2:
Complaining about missing libgl - which is a low level graphic subsystem dependent library.

The plots are not very useful IMHO but trying adding https://anaconda.org/conda-forge/mesa-libgl-devel-cos7-x86_64 
Will remove all plots if that doesn't fix the CI tests.

fubar2:
Can anyone suggest an easy and working fix for this libGL issue with gnuplot?
Cannot find any other IUC tools with gnuplot as a dependency. 
Perhaps this is why.
Otherwise, does anybody object to the plots being removed here? 
They are crude compared to using JBrowse for example.

bgruening:
I know how to fix it. Can you remove the libgl dep please again. I need 1-2 once I get up ;-)

fubar2:
@bgruening - removed so you can do the needful...
Thanks!

tools/mummer4/nucmer.xml
```diff
@@ -31,52 +38,124 @@
...
+                && tail  -n +3 outsam.sam > outsambody
+                && samtools dict reference.fa > outsamhead
+                && cat outsambody >> outsamhead
+                #if $outform.out_format in ['bam-short', 'bam-long']:
+                    && samtools view -b -o bam_out.bam outsamhead
+                    && samtools sort bam_out.bam -o outsam
+                    && samtools index outsam
+                #else if $outform.out_format in ['cram-short', 'cram-long']:
+                    && samtools view -C -o cram_out.cram  -T reference.fa  outsamhead && mv cram_out.cram outsam
+                    && samtools index outsam
```
bgruening, line 80:
```suggestion
```

tools/mummer4/nucmer.xml
```diff
@@ -31,52 +38,124 @@
...
                     @MUMMER_GNUPLOT_MANUAL@
+                #end if
+            #else:
+                && tail  -n +3 outsam.sam > outsambody
+                && samtools dict reference.fa > outsamhead
+                && cat outsambody >> outsamhead
+                #if $outform.out_format in ['bam-short', 'bam-long']:
+                    && samtools view -b -o bam_out.bam outsamhead
+                    && samtools sort bam_out.bam -o outsam
+                    && samtools index outsam
```
bgruening, line 77:
```suggestion
```

bgruening (commented):
@fubar2 the output is now BAM correct? Should we also change the wording in the wrapper and talk always from BAM?

Also what is the difference between short and long? Should we offer only long?

tools/mummer4/nucmer.xml
fubar2: Thanks - I thought I added that because otherwise bam failed - testing without now.

fubar2:
> @fubar2 the output is now BAM correct?

or cram.

> Also what is the difference between short and long? Should we offer only long?

```The original output format of nucmer, the delta format, contains only the minimum information necessary to quickly recreate the alignment. It contains the name of the matching sequences, the length of the match, number of errors and positions of indels. An important addition in MUMmer4 is option to produce output in SAM format, one of the most widely used formats for alignments of NGS data [[10](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1005944#pcbi.1005944.ref010)]. This also allows the MUMmer4 output to be used in any of the numerous tools that require SAM files as input. Nucmer4 supports two different options for SAM format output. With --sam-short, nucmer4 reports only the name of the matching sequence, length, and CIGAR string (which reports the indel positions). With --sam-long, it additionally reports the MD string (which specifies the mismatching positions), the sequence and, if applicable, the quality values of the matching sequence. The long format is more expensive to compute and it generates larger output files, but this option allows nucmer4 to match the behavior of other aligners such as Bowtie2 or BWA.```

**Comment by fubar2** [comment]
I'd leave both but up to you.
Have added that text to the help.

File: tools/mummer4/nucmer.xml [hunk 3]
```diff
@@ -31,142 +38,237 @@
...
+       <test expect_num_outputs="1">
+            <param name="advanced" value="defaults" />
+            <param name="makeshort" value="1" />
+            <conditional name="outform">
+                <param name="out_format" value="bam-short" />
+            </conditional>
+            <param name="seq_input" value="yes" />
+            <param name="reference_sequence" ftype="fasta" value="human_aqp3.fasta"/>
+            <param name="query_sequence" ftype="fasta" value="mouse_aqp3.fasta" />
+           <output name="sam_output" ftype="bam" compare="sim_size" value="out.bam" />
```
bgruening, line 223:
```suggestion
            <output name="sam_output" ftype="bam" value="out.bam" ftype="bam" />
```

Galaxy should be smart enough to uncompress the BAM and compare the SAM via diff

tools/mummer4/nucmer.xml
```diff
@@ -31,142 +38,237 @@
...
+        <test expect_num_outputs="1">
+            <param name="advanced" value="defaults" />
+            <param name="makeshort" value="0" />
+            <conditional name="outform">
+                <param name="out_format" value="cram-long" />
+            </conditional>
+            <param name="seq_input" value="yes" />
+            <param name="reference_sequence" ftype="fasta" value="human_aqp3.fasta"/>
+            <param name="query_sequence" ftype="fasta" value="mouse_aqp3.fasta" />
+           <output name="sam_output" ftype="cram" compare="sim_size" value="out.cram" />
```
bgruening, line 234:
```suggestion
            <output name="sam_output" ftype="cram" compare="sim_size" value="out.cram" />
```

bgruening (commented):
There is an empty SAM file in this PR tools/mummer4/test-data/out.sam

tools/mummer4/nucmer.xml
fubar2: It might be but those changes make planemo lint complain so I've reverted....

```
(.venv) ross@pn50:~/rossgit/tools-iuc/tools/mummer4$ planemo lint nucmer.xml                                                                                 
galaxy.util ERROR: Error parsing file /home/ross/rossgit/tools-iuc/tools/mummer4/nucmer.xml                                                                  
  File "/home/ross/rossgit/galaxytf231/.venv/lib/python3.10/site-packages/galaxy/util/__init__.py", line 305, in parse_xml                                   
    tree = etree.parse(str(fname), parser=parser)                                                                                                            
  File "src/lxml/etree.pyx", line 3541, in lxml.etree.parse
  File "src/lxml/parser.pxi", line 1879, in lxml.etree._parseDocument
  File "src/lxml/parser.pxi", line 1905, in lxml.etree._parseDocumentFromURL
  File "src/lxml/parser.pxi", line 1808, in lxml.etree._parseDocFromFile
  File "src/lxml/parser.pxi", line 1180, in lxml.etree._BaseParser._parseDocFromFile
  File "src/lxml/parser.pxi", line 618, in lxml.etree._ParserContext._handleParseResultDoc
  File "src/lxml/parser.pxi", line 728, in lxml.etree._handleParseResult
  File "src/lxml/parser.pxi", line 657, in lxml.etree._raiseParseError
  File "/home/ross/rossgit/tools-iuc/tools/mummer4/nucmer.xml", line 223
lxml.etree.XMLSyntaxError: Attribute ftype redefined, line 223, column 79
Error loading tool with path /home/ross/rossgit/tools-iuc/tools/mummer4/nucmer.xml
  File "/home/ross/rossgit/galaxytf231/.venv/lib/python3.10/site-packages/galaxy/tool_util/loader_directory.py", line 103, in _load_tools_from_path
    tool_element = loader_func(possible_tool_file)
  File "/home/ross/rossgit/tools-iuc/tools/mummer4/nucmer.xml", line 223
lxml.etree.XMLSyntaxError: Attribute ftype redefined, line 223, column 79
Could not lint /home/ross/rossgit/tools-iuc/tools/mummer4/nucmer.xml due to malformed xml.

```

bgruening: Oh, sorry, I added a new ftype but it has ftype already. Just remove the second.

wm75:
> with --sam-long, it additionally reports the MD string (which specifies the mismatching positions), the sequence and, if applicable, the quality values of the matching sequence. The long format is more expensive to compute and ...

@fubar2 can you run the long format output through samtools calmd please to see if there's a warning about existing MD tags that got changed?
I wouldn't trust a tool that doesn't even get the sam header right to calculate MD correctly.
If everything is fine, then maybe offer only the long version, otherwise short cause you'd have to run it through samtools calmd anyway.

fubar2:
> @fubar2 can you run the long format output through samtools calmd please to see if there's a warning about existing MD tags that got changed?

The samtools calmd tool returns a file of the same size, with no visible warning on both short and long sam outputs 
@wm75: @bgruening: I have no expertise for making this decision so do not have useful opinions to offer in this debate. 
If it makes sense to run calmd routinely,  or to get rid of the short form outputs please just tell me what you want or feel free to repair....

fubar2:
Testing with a larger input: calmd does report and fix some NM and MD errors in the sam long format so will add it as a prophylactic step.

> [bam_fillmd1] different NM for read 'sacCer3_gold_BK006938.2': 202 -> 1706
[bam_fillmd1] different MD for read 'sacCer3_gold_BK006938.2': '33^a23a4t1c0t2c4c5c4c1^ct2t1t2c4g16c17g18t1c8c7^t2g17c4t
...

fubar2:
@bgruening @wm75 
Now with calmd run to fix Mummer sam format and an executive decision to remove all short-sam. 
Added a test for sam output too.

fubar2:
and added the required output file for the sam test.....

tools/mummer4/nucmer.xml
```diff
@@ -31,142 +34,223 @@
...
         ]]>
-    </command> 
+    </command>
     <inputs>
         <param name="reference_sequence" type="data" format="fasta" label="Reference Sequence" help="FastA or multi-FastA" />
         <param name="query_sequence" type="data" format="fasta" label="Query Sequence" help="FastA or multi-FastA" />
+        <conditional name="outform">
+            <param name="out_format" type="select" label="Output format" help="Select delta format if a plot is needed. Jbrowse is a good choice to view cram and bam tracks">
+                <option value="bam-long">bam format</option>
+                <option value="sam-long">sam format</option>
```
bgruening, line 89: I would remove SAM altogether, to be honest.

tools/mummer4/nucmer.xml
```diff
@@ -31,142 +34,223 @@
...
+                <option value="cram-long">cram format</option>
+                <option value="delta">Mummer delta format - allows plots</option>
+            </param>
+            <when value="delta">
+                <conditional name="mumplot" >
+                    <param name="plot" type="select" label="If the output file is in delta (not sam) format, make a 2-D dotplot of the input sequences? (mummerplot)" >
```
bgruening, line 95:
```suggestion
                    <param name="plot" type="select" label="Create a 2-D dotplot of the input sequences?" >
```

tools/mummer4/nucmer.xml
```diff
@@ -31,142 +34,223 @@
...
+                    $outform.mumplot.snp
+                    #if $outform.mumplot.range.custom == 'yes':
+                        -x [$outform.mumplot.range.min_x:$outform.mumplot.range.max_x]
+                        -y [$outform.mumplot.range.min_y:$outform.mumplot.range.max_y]
                     #end if
                     'out.delta'
                     @MUMMER_GNUPLOT_MANUAL@
+                #end if
+            #else:
+                && tail  -n +3 outsam.sam > outsambody
```
bgruening, line 67: It is always just 3 lines as header?

bgruening (commented):
Vers nice Ross, I only have three small comments/questions. Lets wait for @wm75 and then get it in.

tools/mummer4/nucmer.xml
fubar2: Always two broken rows. 
That works correctly strangely enough - an example:
```
(.venv) ubuntu@ross-newgrt:/evol/galaxytf23.1.3$ cat tsconfig.json 
{
    "extends": "./client/tsconfig.json",
}
(.venv) ubuntu@ross-newgrt:/evol/galaxytf23.1.3$ tail -n +3 tsconfig.json 
}
(.venv) ubuntu@ross-newgrt:/evol/galaxytf23.1.3$ 

```

tools/mummer4/nucmer.xml
fubar2: If nobody else has a specific use case for sam I'll remove it - I don't need it...

fubar2: ok. sam gone - tested and pushed.

fubar2:
@wm75 @bgruening: saved a few percent run time wasted in a copy, but noticing that `samtools sort` on the 5GB output I'm seeing takes > 5% (30 minutes of about 600 on my test server) - is it a good investment for all future jobs despite both having indexes?

tools/mummer4/nucmer.xml
```diff
@@ -31,142 +34,206 @@
...
+            #else:
+                && samtools dict reference.fa > outsamhead
+                && tail -n +3 outsam.sam >> outsamhead
+                && samtools sort -o outsamheadsort outsamhead
+                && samtools calmd outsamheadsort reference.fa > outsamcalm
+                #if $outform.out_format == 'bam-long':
+                    && samtools view -b -o outsam outsamcalm
+                #else if $outform.out_format == 'cram-long':
+                    && samtools view -C -o outsam -T reference.fa outsamcalm
+                #end if
```
wm75, line 75:
```suggestion
                && samtools view -T reference.fa -u -x ^ outsam.sam | samtools sort -@ \${GALAXY_SLOTS:-1} -T "\${TMPDIR:-.}" |
                #if $outform.out_format == 'bam-long':
                    -u | samtools calmd -b --threads {GALAXY_SLOTS:-1} - reference.fa > outsamcalm
                #else if $outform.out_format == 'cram-long':
                    -O cram --reference reference.fa -o outsamcalm
                #end if
```

what about this instead?
untested, but likely working.

wm75: - the samtools view command adds the required header and removes the MD and NM tags calculated by nucmer
- samtools sort gets called with multiple threads and with the admin supplied tempdir for efficiency
- samtools calmd is required only for bam; cram format autocalculates MD/NM on decoding (http://www.htslib.org/doc/samtools-calmd.html)

wm75: Otherwise, this looks good to me :-)

fubar2: Will test and get this suggestion merged later today - thanks @wm75

fubar2: @wm75: 

> the samtools view command adds the required header and removes the MD and NM tags calculated by nucmer

Unfortunately samtools view refuses to even read the broken header input so won't remove it. 
It seems there's no choice but to manually remove it. Will continue to experiment...

fubar2: This works I think - thanks @wm75 . 
Frustrating that we won't see this bogus sam and other warts fixed upstream unless we can find someone willing to take it on as a project.

fubar2: I just realised there's no need for `-x ^` since we just made decent headers!!

wm75: > Frustrating that we won't see this bogus sam and other warts fixed

A pity, yes, but at least you've come up with an acceptable workaround :+1: 

You should either drop the `samtools dict` or the `samtools view` though to avoid redundancy.
`samtools view` should work after the header is removed by `tail` and the intention behind `-x ^` was to suppress the warnings from samtools calmd.

If you don't need MD/NM for vgp, you could also just drop samtools calmd and produce BAM without these tags. Users who need them and cannot go with cram could still run samtools calmd as a separate step of their WF.

Any of these options are fine with me, just pick what seems most practical for you.

fubar2: @wm75: How about we go with: 

```
#else:
&& samtools dict reference.fa > outsamhead
&& tail -n +3 outsam.sam >> outsamhead
&& samtools sort  -@ \${GALAXY_SLOTS:-1} -T "\${TMPDIR:-.}" outsamhead |
#if $outform.out_format == 'bam-long':
    samtools calmd -b --threads {GALAXY_SLOTS:-1} - reference.fa > outsam
#else if $outform.out_format == 'cram-long':
    samtools view -C --reference reference.fa -o outsam -
#end if
```

wm75 (approved):
Yes, that's an option, too :-)

fubar2:
Thanks for all your help, @bgruening and @wm75.

PR #5869, opened 2024-03-15
Replace EDAM by bio.tools in usher

bebatut:
This PR links usher to its bio.tools [entry](https://bio.tools/usher).


bebatut:
Thanks @bgruening for fixing

PR #5891, opened 2024-03-20
Check InterProScan seqtype

TomHarrop:
FOR CONTRIBUTOR:
---

This tool is causing a bit of support work when users input nucleotide sequences but select protein as the sequence type.

This greps for anything that is not an IUPAC nucleotide and checks the result against the selected seqtype.

It won't always be correct because a protein consisting only of amino acid residues that are in the nucleotide alphabet is valid (but probably not common). It could also take a long time to search through valid nucleotide input. I could limit it to the first 1000 lines or something?

Ping @igormakunin

### DIFF
--- a/tools/interproscan/interproscan.xml
+++ b/tools/interproscan/interproscan.xml
@@ -39,7 +39,6 @@ if [ \$detected_seqtype != '$seqtype' ]; then
     exit 1 ;
 fi;
 
-
 ## Now run interproscan
 interproscan.sh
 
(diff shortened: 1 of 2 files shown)


TomHarrop:
screenshot-localhost_9090-2024 03 20-13_31_32

---

screenshot-localhost_9090-2024 03 20-13_31_09

---

@neoformit please LMK if you have any comments on the UX or anything else.

neoformit:
Looks good to me, is it possible to add "detected nucleotide when you selected protein" to the error message, just to be a bit more specific for the user?

TomHarrop:
@neoformit that would be better, but the values of the seqtype variable are "p" and "n" so it makes the command section a bit nastier. See https://github.com/galaxyproject/tools-iuc/pull/5891/commits/7a2535904df9c001be95b776150b0829074e691d

neoformit:
Since you already have the `if grep -q '[^[:space:]]' <<< \${match}; then` line, yeah I think this is worth it. Looks good, thanks Tom!

bgruening:
Thanks! @TomHarrop can you look at the failing lint, please?

How is interproscan failing and how fast? We could also add a `stdio`, catch this error and provide your better error description.

Some thoughts for the future, because this is now coming up more and more often:

* do we need separate datatypes for nt, and prot
* do we need a library/script that checks the content and can be shared?
* should this be part of the fasta sniffer (it will have performance implications)

TomHarrop:
Thanks @bgruening . The lint failure is because of InterProScan's non-standard version numbering.

> How is interproscan failing and how fast? We could also add a `stdio`, catch this error and provide your better error description.

I believe it doesn't fail but creates huge jobs that run for days, and there is a similar problem with blastp. My understanding is that  @igormakunin has to cancel jobs manually to keep the queue moving. Hopefully he can comment.

---

>   do we need separate datatypes for nt, and prot

I don't think there is any technical difference in FASTA that we can rely on, e.g. "ATGC" is a valid peptide string.

---
 
>   do we need a library/script that checks the content and can be shared?

This seems like a good option to me. I could look into that if you point me in the right direction.

---

>    should this be part of the fasta sniffer (it will have performance implications)

As above we might not detect them reliably, but I don't know how the sniffer works under the hood.

neoformit:
> do we need a library/script that checks the content and can be shared?

Ages ago I did start developing a Python lib `fastkit` for that reason: https://github.com/neoformit/fastkit

It has a validator for DNA, protein, sequence count etc. and some common formatting for FASTA files. Could also implement for FASTQ but haven't yet. It raises stderr that results in user-friendly error messages in Galaxy.

Maybe there are existing libraries but I don't know of one that does this exactly. It wraps BioPython etc. under the hood.

neoformit:
> should this be part of the fasta sniffer (it will have performance implications)

This seems like a nice solution but the logic for checking and raising error messages still has to be re-implemented in every wrapper, right?

bgruening:
@TomHarrop how do we proceed here? Should we merge what you have here?

TomHarrop:
@bgruening I guess so.

It's not ideal because the data still gets copied to the runner before the check runs. But I don't recall hearing about any other solution in the last 6 months so I assume this is still a problem :)

bgruening:
Can you rebase and fix the linting ? Thanks!

TomHarrop:
Yikes what happened here?

Don't merge this, I must have rebased off the wrong branch.

TomHarrop:
Thanks @bgruening this should be OK now.

bgruening:
@abretaud is that ok with you?

+1 from my side.

PR #5915, opened 2024-03-27
Pairtools split (Minor Changes and Updated test files)

SaimMomin12:


tools/pairtools/split.xml
```diff
@@ -0,0 +1,40 @@
...
+            --output-sam '$output_file'
+            --nproc-in \${GALAXY_SLOTS:-4}
+            --nproc-out \${GALAXY_SLOTS:-4}
+    ]]></command>
+    <inputs>
+        <param name="pairsam_path" type="data" format="tabular" label="Input .pairsam file" help="Input .pairsam file."/>
+    </inputs>
+    <outputs>
+        <data name="output_pairs" format="tabular" label="${tool.name} on ${on_string}: .pairs"/>
+        <data name="output_file" format="sam,bam" label="${tool.name} on ${on_string}: .bam"/>
```
bgruening, line 20: outputs can not have two output formats. It's either SAM or BAM. It seems to be SAM here, which raises the question if this tool should not also call samtools view and convert SAM to BAM to just output BAM. To save space.

tools/pairtools/split.xml
```diff
@@ -0,0 +1,40 @@
...
+    </outputs>
+    <tests>
+        <!--Test default parameters-->
+        <test expect_num_outputs="2">
+            <param name="pairsam_path" value="output_dedup_pairs_markdups.pairsam"/>
+            <output name="output_pairs" ftype="tabular" file="output_pairs_split.pairs" lines_diff="10"/>
+            <output name="output_file" ftype="sam,bam" file="output_pairs_split.bam" lines_diff="10"/>
```
bgruening, line 27: that this passes linting is surprising ... it should only be one filetype

diegomics: Let's keep it on BAM, right? In the command, if the path ends with .bam, the output is compressed into a bam file, otherwise is sam

bgruening: Oh nice, yes, then least keep BAM.
Q, does the BAM file needs to be coordinate sorted or name sorted?

SaimMomin12: Okay, I'll change it to BAM

tools/pairtools/split.xml
```diff
@@ -0,0 +1,40 @@
...
+    <command detect_errors="exit_code"><![CDATA[
+        pairtools split
+            '$pairsam_path'
+            --output-pairs '$output_pairs'
+            --output-sam ./output.bam
+            --nproc-in \${GALAXY_SLOTS:-4}
+            --nproc-out \${GALAXY_SLOTS:-4}
+    ]]></command>
+    <inputs>
+        <param name="pairsam_path" type="data" format="4dn_pairsam" label="Input .pairsam file" help="Input .pairsam file."/>
```
bgruening, line 16:
```suggestion
        <param name="pairsam_path" type="data" format="4dn_pairsam" label="Input pairsam file"/>
```

tools/pairtools/split.xml
```diff
@@ -0,0 +1,40 @@
...
+        <test expect_num_outputs="2">
+            <param name="pairsam_path" value="output_dedup_pairs_markdups.pairsam"/>
+            <output name="output_pairs" ftype="4dn_pairs" file="output_pairs_split.pairs" lines_diff="10"/>
+            <output name="output_file" ftype="qname_sorted.bam" file="output_pairs_split.bam" lines_diff="20"/>
+        </test>
+    </tests>
+    <help><![CDATA[
+        **Pairtools split**
+
+        Restore a .sam file from sam1 and sam2 fields of a .pairsam file. Create a .pairs file without sam1/sam2 fields.
```
bgruening, line 33:
```suggestion
        Restore a SAM/BAM file from sam1 and sam2 fields of a .pairsam file. Create a .pairs file without sam1/sam2 fields.
```

bgruening: Does this sentence indicate that 2 SAM/BAM files are created?

diegomics: It means that from the pairsam file, two files are created:
File 1: .sam or .bam (depending on the extension of the output, we are using bam, so only bam is created)
File 2: .pairs
Also, one thing that I do in my scripts, is, after this step, compress (gz) the .pairs file

bgruening (approved):

SaimMomin12:
Closed #5932

PR #5920, opened 2024-03-28
minor bug fix for pretext graph to function in workflows

astrovsky01:
There appears to be an inconsistency in how the workflow editor reads input formats vs the tool form, where the accidental space left in the formats section of the inputs functioned ok on a tool form, but not in the workflow editor, which didn't recognize the bedgraph format as a valid connection. Fixing the space here and adding an issue on the galaxy repo for the inconsistency problem. Not bumping version because it's a minor bug fix of the wrapper, not the tool

### DIFF
--- a/tools/pretext/pretext_graph.xml
+++ b/tools/pretext/pretext_graph.xml
@@ -23,7 +23,7 @@
         -o output.pretext
     ]]></command>
     <inputs>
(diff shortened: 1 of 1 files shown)


bgruening (approved):

bgruening:
Thanks @astrovsky01

PR #5927, opened 2024-04-02
Fixed a minor test issue for hapog tool 

SaimMomin12:


wm75:
hmm, not sure whether hiding real output changes behind a lines_diff is a good idea.

SaimMomin12:
Hi @wm75,

Based on your suggestion I verified if the input.bam used in the failing test contains secondary alignments - Accordingly there seems to be no secondary alignments in the same.  (`samtools view -f 0x100 input.bam`)

In the scenario, when I update the test-data with the same input.bam the tests do pass without the need of specifying lines_diff.

How do we go forward with it?

wm75:
The point is: if the test change is really just due to 0x900 reads getting removed before the analysis, then that's an intended change, we can update the test input file and be done with it.
But if it's something else going on here, then we should find out and possibly report upstream.

SaimMomin12:
> ok, I haven't been specific enough, but the version diff I linked above shows that the new hapo-g version removes reads with the 0x900 flag combination menaing secondary and supplementary alignments. Are reads meeting this criterion in the input.bam then?

No, there are no reads in the input bam file that meet the criteria of secondary and supplementary alignments as far as I see.

wm75:
hmm, doesn't it say so right there:
> 141 + 0 supplementary

??

wm75:
> But we do have some supplementary reads in the input bam.

Ah missed that sentence, sorry.

Then, yes, try to remove these manually and use that cleaned input with the previous version of hapo-g to see if then it matches the new version's output. If so, we're fine.

wm75 (approved):
Cool, thanks for investigating!

PR #5928, opened 2024-04-02
interaction:fixed effects and corrections modified

renu-pal:
FOR CONTRIBUTOR:
[ *] I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
[ *] This PR updates an existing tool or tool collection

### DIFF
--- a/tools/maaslin2/maaslin2.xml
+++ b/tools/maaslin2/maaslin2.xml
@@ -12,6 +12,7 @@
 ## can only return indices with type="data_column" 
 ## using awk so that the file is only parsed on command line execution
 
+#if $fixed_effects
 #set idx = []
 #for $i in $fixed_effects:
     #silent idx.append(f'${i}')
@@ -20,7 +21,10 @@
 
 fixed_effects=`awk -v OFS=',' -F"\t" 'NR == 1 { print $idx_for_awk}' '$input_metadata'` &&
 echo 'Assigned fixed effects as:' \$fixed_effects &&
+#end if
 
+
+#if $random_effects
 #set idx = []
 #for $i in $random_effects:
     #silent idx.append(f'${i}')
@@ -29,6 +33,7 @@ echo 'Assigned fixed effects as:' \$fixed_effects &&
 
 random_effects=`awk -v OFS=',' -F"\t" 'NR == 1 { print $idx_for_awk}' '$input_metadata'` &&
 echo 'Assigned random effects as:' \$random_effects &&
+#end if
 
 ln -s '$input_data' 'input_data.tsv'
 &&
@@ -79,9 +84,9 @@ cd outputFolder && mkdir -p figures/ && cp *.pdf figures
     <inputs>
         <param name="input_data" type="data" format="tabular" label="Data (or features) file"/>
         <param name="input_metadata" type="data" format="tabular" label="Metadata file"/>
-        <param argument="--fixed_effects" type="data_column" data_ref="input_metadata" use_header_names="true"  multiple="true" optional="true" label="Interactions: Fixed effects" help="The fixed effects for the model, comma-delimited for multiple effects" />
+        <param argument="--fixed_effects" type="data_column" data_ref="input_metadata" use_header_names="true"  multiple="true" optional="true" label="Interactions: Fixed effects" help="The fixed effects for the model, comma-delimited for multiple effects, Default value: All " />
         		
-        <param argument="--random_effects" type="data_column" data_ref="input_metadata" use_header_names="true" multiple="true" optional="true" label="Random effects" help="The random effects for the model,  comma-delimited for multiple effects" />
+        <param argument="--random_effects" type="data_column" data_ref="input_metadata" use_header_names="true" multiple="true" optional="true" label="Random effects" help="The random effects for the model,  comma-delimited for multiple effects, Default: None" />
         	
        
         	
@@ -108,7 +113,7 @@ cd outputFolder && mkdir -p figures/ && cp *.pdf figures
                 <option value="NEGBIN">NEGBIN</option>
                 <option value="ZINB">ZINB</option>
             </param>
-            <param argument="--correction" type="select" value="BH" multiple="true" optional="true" label="Correction" help="The correction method for computing  the q-value ">
+            <param argument="--correction" type="select" value="BH" optional="true" label="Correction" help="The correction method for computing  the q-value ">
                 <option value="holm">Holm</option>
                 <option value="hochberg">Hochberg</option>
                 <option value="hommel">Hommel</option>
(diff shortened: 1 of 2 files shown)


tools/maaslin2/maaslin2.xml
```diff
@@ -55,7 +55,7 @@ cd outputFolder && mkdir -p figures/ && cp *.pdf figures
     <inputs>
         <param name="input_data" type="data" format="tabular" label="Data (or features) file"/>
         <param name="input_metadata" type="data" format="tabular" label="Metadata file"/>
-        <param argument="--fixed_effects" type="select" multiple="true" optional="true" label="Interactions: Fixed effects" help="The fixed effects for the model, comma-delimited for multiple effects">
+        <param argument="--fixed_effects" type="text" multiple="true" optional="true" label="Interactions: Fixed effects" help="The fixed effects for the model, comma-delimited for multiple effects">
```
bgruening, line 58: Why should this be a text paraeter? We should probably fix the help instead?

tools/maaslin2/maaslin2.xml
```diff
@@ -87,7 +87,14 @@ cd outputFolder && mkdir -p figures/ && cp *.pdf figures
...
             </param>
-            <param argument="--correction" type="text" value="BH" optional="true" label="Correction" help="The correction method for computing  the q-value"/>
+            <param argument="--correction" type="select" value="BH" multiple="true" optional="true" label="Correction" help="The correction method for computing  the q-value">
+            <option value="holm">holm</option>
+            <option value="hochberg">hochberg</option>
+            <option value="hommel">hommel</option>
+            <option value="bonferroni">bonferroni</option>
+            <option value="BH">BH</option>
+            <option value="BY">BY</option>
+            </param>
```
bgruening, line 97:
```suggestion
            <param argument="--correction" type="select" value="BH" multiple="true" optional="true" label="Correction" help="The correction method for computing  the q-value">
                 <option value="holm">holm</option>
                 <option value="hochberg">hochberg</option>
                 <option value="hommel">hommel</option>
                 <option value="bonferroni">bonferroni</option>
                 <option value="BH">BH</option>
                 <option value="BY">BY</option>
            </param>
```

bgruening (commented):
Thanks @renu-pal.

One question inline. Please also bump the tool version.

tools/maaslin2/maaslin2.xml
renu-pal: We can do that now but if more association factors get added in the future, we might have to make changes again as we are doing now. So instead take it as a text parameter.

paulzierep: In fact, the `fixed_effects` as well as the `random_effects` should be of type: `data_column` (see: https://github.com/galaxyproject/tools-iuc/blob/4e3719a2c09e11afd9069bb520de11e3a0f02c72/tools/bedtools/overlapBed.xml#L17 for an example).
Since for both multiple columns from the input can be chosen. The original optional values are nonsense, since they hard-coded the column names of the test data.

tools/maaslin2/maaslin2.xml
```diff
@@ -87,7 +87,14 @@ cd outputFolder && mkdir -p figures/ && cp *.pdf figures
                 <option value="NEGBIN">NEGBIN</option>
                 <option value="ZINB">ZINB</option>
             </param>
-            <param argument="--correction" type="text" value="BH" optional="true" label="Correction" help="The correction method for computing  the q-value"/>
+            <param argument="--correction" type="select" value="BH" multiple="true" optional="true" label="Correction" help="The correction method for computing  the q-value">
+            <option value="holm">holm</option>
+            <option value="hochberg">hochberg</option>
+            <option value="hommel">hommel</option>
+            <option value="bonferroni">bonferroni</option>
+            <option value="BH">BH</option>
```
paulzierep: Can you add the full names for BH and BY as well, maybe also add a link to the explanation to the help and check if they all work, maybe add a test or two as well

paulzierep (commented):
Thanks a lot @renu-pal ! Ones this is fixed, the tool is finally functional !
Ones they update the bioconda recipe: https://github.com/bioconda/bioconda-recipes/pull/46915 we can also update the wrapper, but lets do this one first !

paulzierep:
> Thanks @renu-pal.
> 
> One question inline. Please also bump the tool version.

@renu-pal that means you need to increase the `VERSION_SUFFIX` by 1: https://github.com/renu-pal/tools-iuc/blob/ea393cd20a2651a554f59c8afa1620b5792951f8/tools/maaslin2/macros.xml#L4

The `TOOL_VERSION` can be increased to match the latest bioconde recipe ones its updated: https://forum.biobakery.org/t/update-maaslin2-bioconda-recipe/6830 and https://github.com/bioconda/bioconda-recipes/pull/46915

tools/maaslin2/maaslin2.xml
renu-pal: I am getting column numbers instead of names after using type="data_column" . Do we have any other type where we can get names of columns directly instead of numbers?

bgruening: Have you tried setting numerical=false? https://docs.galaxyproject.org/en/latest/dev/schema.html#id51

renu-pal: setting numerical="false" did not work instead setting use_header_names="true" worked and UI shows column names now but the test cases are failing. Do not understand why it is still accepting  numerical values as valid options for tests. So working on that now. 

[tool_test_output.json](https://github.com/galaxyproject/tools-iuc/files/14912085/tool_test_output.json)

bgruening: I could be wrong, but the use_header_names="true" option is really just cosmetic and for the UI, you still need to give numbers in the tests.

Very good idea to use use_header_names="true"!!!

paulzierep: hi @renu-pal it was a surprise for me, that there is not easy way to get the column_names available for the command line from Galaxy, however after some fideling I found a way to get them from the input file:
```
## get column names of fixed and random effect from the input file, since galaxy 
## can only return indices with type="data_column" 

## get header
#set $input = open(str($input_metadata), 'r')
#set $header = $input.readlines()[0].split('\t')

## get fixed effects
#set $fixed_effects_val = []
#for $i in $fixed_effects:
    #silent $fixed_effects_val.append($header[int($i)])
#end for
#set $fixed_effects = ','.join($fixed_effects_val)

## get random effects
#set $random_effects_val = []
#for $i in $random_effects:
    #silent $random_effects_val.append($header[int($i)])
#end for
#set $random_effects = ','.join($random_effects_val)
```

Maybe let the user know in the help, that the column names must be in the first line of the table or modify the code to skip comments when parsing the header

paulzierep: forget about that, we need to get the column names in the command line not the cheetah code...

paulzierep: So update, this way, using awk the file is only parsed on script execution 
```
## get column names of fixed and random effect from the input file, since galaxy 
## can only return indices with type="data_column" 
## using awk so that the file is only parsed on command line execution

#set idx = []
#for $i in $fixed_effects:
    #silent idx.append(f'${i}')
#end for
#set idx_for_awk = ','.join(idx)

fixed_effects=`awk -v OFS=',' -F"\t" 'NR == 1 { print $idx_for_awk}' '$input_metadata'` &&
echo 'Assigned fixed effects as:' \$fixed_effects &&

#set idx = []
#for $i in $random_effects:
    #silent idx.append(f'${i}')
#end for
#set idx_for_awk = ','.join(idx)

random_effects=`awk -v OFS=',' -F"\t" 'NR == 1 { print $idx_for_awk}' '$input_metadata'` &&
echo 'Assigned random effects as:' \$random_effects &&
```

tools/maaslin2/macros.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <macros>
-    <token name="@TOOL_VERSION@">0.99.12</token>
+    <token name="@TOOL_VERSION@">1.7.3</token>
```
paulzierep, line 3: @atm we are still at 0.99.12, no ?

tools/maaslin2/maaslin2.xml
```diff
@@ -7,6 +7,29 @@
     <expand macro="xrefs"/>
     <expand macro="requirements"/>
     <command detect_errors="exit_code"><![CDATA[
+    
+## get column names of fixed and random effect from the input file, since galaxy 
+## can only return indices with type="data_column" 
+## using awk so that the file is only parsed on command line execution
+
+#set idx = []
```
paulzierep, line 15: @renu-pal you said `
Yh it did a bit, but it's returning the complete list in case of 'random effect " when user does not choose any option `
Could you try to fix that here, i.e. add an if statement, that does not choose any columns if no columns are specified by the user - and also add a test for it ...?

paulzierep (commented):
Please update to work with no columns selected

tools/maaslin2/maaslin2.xml
renu-pal: okay, working on that

renu-pal:
warning : the correction methods  other than BH and BY will not work due to older version of maaslin.

bgruening:
@renu-pal the newer version of https://anaconda.org/bioconda/maaslin2/files is online.

renu-pal:
Thanks for the update @bgruening ! I will bump the version to 1.16.0 but will need some time to push the updated version as test cases are failing because of the version update. So I need to look into that and make necessary changes.

paulzierep:
Since for the current version the wrapper is broken (not logically usable), would it make sense to merge this version first and then bump to the next version in a new PR ? @bgruening any preference ?

tools/maaslin2/maaslin2.xml
```diff
@@ -245,7 +278,7 @@ cd outputFolder && mkdir -p figures/ && cp *.pdf figures
         <test expect_num_outputs="5">
             <param name="input_data" value="HMP2_taxonomy.tsv"/>
             <param name="input_metadata" value="HMP2_metadata.tsv"/>
-            <param name="fixed_effects" value="diagnosis,dysbiosisnonIBD"/>
+            <param name="fixed_effects" value="c4: diagnosis,c9: dysbiosisnonIBD"/>
```
bgruening, line 281: @mvdbeek does this look like a regression in Galaxy to you? I thought that either the column name or the `c4` should work, but in combination, it looks wired to me.

renu-pal: The combination is not necessary. I mentioned header names in test cases to make it easier for us  to recognize which values we are taking. I will change it to just numbers.

tools/maaslin2/maaslin2.xml
```diff
@@ -304,6 +337,100 @@ cd outputFolder && mkdir -p figures/ && cp *.pdf figures
...
+                <assert_contents>
+                    <has_text text="feature"/>
+                    <has_n_lines n="8092"/>
+                    <has_n_columns n="9"/>
+                </assert_contents>
+            </output>
+            <output name="significant_results">
+                <assert_contents>
+                    <has_text text="subject"/>
+                    <has_n_lines n="213"/>
```
bgruening, line 370:
```suggestion
                    <has_n_lines n="216" delta="5"/>
```

renu-pal:
Ready to go from my end too!!

tools/maaslin2/maaslin2.xml
```diff
@@ -347,6 +474,9 @@ Output
         - It only includes associations with q-values <= to the threshold.
     - Data frame with residuals for each feature (R data file)
         - This file contains a data frame with residuals for each feature.
+        
+Correction methods to compute the q-value : https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/p.adjust
```
paulzierep, line 478: If this is called for the correction, why can it only take BH / BY ? Or is this fixed now ?

renu-pal: Can you check this link.
https://forum.biobakery.org/t/invalid-p-value-correction-options/3321

It seems there was some issue from their end. So I believe if we update the version, it should work but  I am not sure if they fixed the bug. Any way to check that ?

paulzierep: OK good to know, then the wrapper is fine as is, but you can remove this text from the help section (it will confuse the user since the option is not there); and put it back when we update the tool :)

renu-pal: Okay, will update that

renu-pal:
I have updated the correction help. Please let me know if anything else needs to be updated.

paulzierep:
Looks good from my side, can we merge it @bgruening ?

bgruening:
Thanks @renu-pal! Great work!

renu-pal:
Thanks for the guidance @paulzierep and @bgruening :)

PR #5937, opened 2024-04-09
Add `$GALAXY_SLOTS` and `$GALAXY_MEMORY_MB` to bbmerge

natefoo:
Worth noting that setting `-Xmx` via `$_JAVA_OPTIONS` in the job env as many admins do still overrides this. I pulled the clever bit of handling for that from bbnorm.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/bbtools/bbmerge.xml
+++ b/tools/bbtools/bbmerge.xml
@@ -39,7 +39,12 @@
     ln -s '${read2}' '${read2_file}' &&
 #end if
 
-bbmerge.sh
+if [[ "\${_JAVA_OPTIONS}" != *-Xmx* && "\${JAVA_TOOL_OPTIONS}" != *-Xmx* ]]; then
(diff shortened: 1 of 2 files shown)


bgruening (approved):

tools/bbtools/bbmerge.xml
```diff
@@ -39,7 +39,12 @@
     ln -s '${read2}' '${read2_file}' &&
 #end if
 
-bbmerge.sh
+if [[ "\${_JAVA_OPTIONS}" != *-Xmx* && "\${JAVA_TOOL_OPTIONS}" != *-Xmx* ]]; then
```
bernt-matthias, line 42: I'm a bit scared of the unquoted `*` characters.

If we use the same code in different tools we should use a macro token.

natefoo: This is in a bash double bracket evaluation which I believe is safe (patterns are not matched)?

bernt-matthias: `shellcheck` also does not complain. should be fine.

PR #5945, opened 2024-04-12
Fix containerized merqury

mvdbeek:
Might be faster than https://github.com/bioconda/bioconda-recipes/pull/47201

FOR CONTRIBUTOR:

### DIFF
--- a/tools/merqury/test-data/output_02_qv.tabular
+++ b/tools/merqury/test-data/output_02_qv.tabular
@@ -1,1 +1,1 @@
-assembly	0	1037953		
+assembly	0	1037953	+inf	0
--- a/tools/merqury/macros.xml
+++ b/tools/merqury/macros.xml
@@ -11,7 +11,9 @@
             <requirement type="package" version="@TOOL_VERSION@">merqury</requirement>
             <requirement type="package" version="1.34">tar</requirement>
             <requirement type="package" version="4.6.0">findutils</requirement>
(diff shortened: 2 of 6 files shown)


natefoo:
I'm +1 on not bumping the version since the previous version is broken with containers.

Also just for reference the coreutils bump is necessary since 8.32 was from defaults, does not exist in conda-forge, and no longer exists in defaults.

bgruening:
I take care of the bioconda one

natefoo:
In this case we can drop the addition of awk and sed requirements but will still need the coreutils change to build a new mulled container?

bgruening:
Yes. Will do ... do you need it today or can we wait for the bot until tm?

natefoo:
I think we can wait for the bot, thanks!

mvdbeek:
> I'm +1 on not bumping the version since the previous version is broken with containers.

This makes it much harder to tell what version of a tool was run if users run into this on other servers. We should really always bump the version.

tools/merqury/test-data/output_02_qv.tabular
```diff
@@ -1 +1 @@
-assembly	0	1037953		
+assembly	0	1037953	+inf	0
```
mvdbeek, line 1: It's normal that this changes, it just means that the bug is fixed. The problem is that the test data was wrong initially.

bgruening (approved):

PR #5955, opened 2024-04-16
das tool: do not allow empty labels

bernt-matthias:
just use element identifiers if not given. Seems to be what the tool does by default: https://github.com/cmks/DAS_Tool/blob/bb0954f4f45a2340e9cb9b896d412e477c6c49fc/src/DAS_Tool.R#L398

No idea what the `--labels` argument is good for. But with `--labels ''` the program just fails. So this might be a solution.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/das_tool/das_tool.xml
+++ b/tools/das_tool/das_tool.xml
@@ -7,18 +7,25 @@
     <expand macro="requirements"/>
     <expand macro="version"/>
     <command detect_errors="exit_code"><![CDATA[
-#set $bins = ""
-#set $labels = ""
-#set $sep = ""
+#import re
+
+#set $bins = []
+#set $labels = []
 #for $i, $s in enumerate($binning)
-    #set $bins += "%s%s" %($sep, $s.bins)
-    #set $labels += "%s%s" %($sep, $s.labels)
-    #set $sep = ","
+    #silent $bins.append(str($s.bins))
+    #if $s.labels != ''
+        #silent $labels.append(str($s.labels))
+    #else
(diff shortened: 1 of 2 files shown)


tools/das_tool/das_tool.xml
```diff
@@ -46,7 +53,15 @@ DAS_Tool
         <param argument="--contigs" type="data" format="fasta" label="Contig sequences"/>
         <repeat name="binning" title="Bins" min="1">
             <param argument="--bins" type="data" format="tabular" label="Contigs-to-bin table" help="Tabular with two columns: contig-IDs and bin-IDs. Fasta_to_Contigs2Bin can be used to  Converts genome bins in fasta format to Contigs-to-bin table"/>
-            <param argument="--labels" type="text" value="" label="Name of binning prediction tool used to generate the table"/>
+            <param argument="--labels" type="text" value="" label="Name of binning prediction name" help="If left empty the identifier of the contig-to-bin table is used. Only alphanumeric characters, dash, underscore and dor are allowed. Other characters are replaced by underscore.">
+                <sanitizer invalid_char="_">
```
bgruening, line 57: _ is invalid and then you add it again?

bernt-matthias: `invalid_char`  specifies the character used as a replacement for invalid characters. `_` is a valid char. 

Idea is that this does the same as the regexp that is applied to the `element_identifier`s in the cheetah code (which I took from what we use all the time).

bgruening:
bump?

PR #5963, opened 2024-04-22
update bwa

nilchia:
FOR CONTRIBUTOR:


tools/bwa/bwa.xml
```diff
@@ -77,6 +74,9 @@
             </when>
         </xml>
     </macros>
+    <xrefs>
+        <xref type="bio.tools">bwa</xref>
+    </xrefs>
```
wm75, line 79: applies to bwa-mem, too. You can turn this into a macro and use in both wrappers.

nilchia: @wm75 thanks for the comments. I've added them.

tools/bwa/bwa.xml
```diff
@@ -1,9 +1,6 @@
 <?xml version="1.0"?>
-<tool id="bwa" name="Map with BWA" version="@VERSION@.5">
+<tool id="bwa" name="Map with BWA" version="@TOOL_VERSION@">
```
bgruening, line 2:
```suggestion
<tool id="bwa" name="Map with BWA" version="@TOOL_VERSION@" profile="22.05">
```

nilchia: Thanks :tulip:

macros/read_group_macros.xml
```diff
@@ -138,11 +138,13 @@
...
     </token>
     <xml name="read_group_auto_name_conditional">
-        <param name="do_auto_name" type="boolean" label="Auto-assign" help="Use dataset name or collection information to automatically assign this value" checked="no" />
-        <when value="true">
-        </when>
-        <when value="false">
-            <yield />
+        <param name="do_auto_name" type="select" label="Auto-assign" help="Use dataset name or collection information to automatically assign this value">
+            <option value='true'>Yes</option>
+            <option value='false' selected="true">No</option>
```
wm75, line 143: the value values should be double-quoted.

wm75: maybe that's the reason why you have to set it currently to false explicitly in the test? That shouldn't be necessary.

You could, in addition swap the order of the options - the first one will be the default (so you can avoid the `selected="true"`.

tools/bwa/bwa-mem.xml
```diff
@@ -282,67 +280,68 @@ bwa mem
...
-        <test>
+        <test expect_num_outputs="1">
             <param name="reference_source_selector" value="history" />
             <param name="ref_file" ftype="fasta" value="bwa-mem-mt-genome.fa"/>
             <param name="fastq_input_selector" value="paired"/>
             <param name="fastq_input1" ftype="fastqsanger" value="bwa-mem-fastq1.fq"/>
             <param name="fastq_input2" ftype="fastqsanger" value="bwa-mem-fastq2.fq"/>
             <param name="analysis_type_selector" value="illumina"/>
-            <output name="bam_output" ftype="bam" file="bwa-mem-test1.bam" lines_diff="2" />
+            <output name="bam_output" ftype="bam" file="bwa-mem-test1.bam" lines_diff="4" />
```
wm75, line 290: how comes you have to increase the lines_diff on all tests when you've already updated all the test-data?

macros/read_group_macros.xml
nilchia: Oh ;) yes, that was it.
Thanks, I didn't know that :+1:

tools/bwa/bwa-mem.xml
wm75: So, the MANY differences could well be because of additional XB read tag written by the new version (https://github.com/lh3/bwa/releases/tag/v0.7.18) and that's fine, of course.

The fact that you still have to use a higher lines_diff after updating the test-data, however, suggests that there is now an additional unreproducible line in the output. My guess is that this comes from the samtools update because recent versions of samtools are much stricter about adding their own action as a PG header line.

I think it would be good to observe the failing tests once to know what the cause is, then decide whether that's fine. It very likely is, but we better make sure.

nilchia: Thanks for the comment :tulip: 
I'll push with `line_diff="2"` so we can check the test.

bgruening:
```
Output bam_output:  different than expected, difference (using diff):
( /tmp/local_bam_converted_to_sam_w_nk6u4t.sam v. /tmp/history_bam_converted_to_sam_tedi7j7g.sam )
--- local_file
+++ history_data
@@ -1,7 +1,7 @@
 @HD	VN:1.5	SO:coordinate
 @SQ	SN:gi|251831106|ref|NC_012920.1|	LN:16569
-@PG	ID:bwa	PN:bwa	VN:0.7.18-r1243-dirty	CL:bwa sampe localref.fa first.sai second.sai /tmp/tmpqufi7n5b/files/5/8/0/dataset_580445f9-a94a-4867-974a-83315838926a.dat /tmp/tmpqufi7n5b/files/2/8/0/dataset_280885eb-4e10-447b-8dd1-275aef841462.dat
-@PG	ID:samtools	PN:samtools	PP:bwa	VN:1.20	CL:samtools sort -@1 -T /tmp/tmpqufi7n5b/tmp -O bam -o /tmp/tmpqufi7n5b/job_working_directory/000/11/outputs/dataset_be28ce69-155e-45b3-8c20-58d5d1a4847b.dat
+@PG	ID:bwa	PN:bwa	VN:0.7.18-r1243-dirty	CL:bwa sampe localref.fa first.sai second.sai /tmp/tmpa_72_4_v/files/e/d/6/dataset_ed65bf2b-5b72-4e85-8767-9c3baf81a597.dat /tmp/tmpa_72_4_v/files/4/3/c/dataset_43cab61d-282a-4616-bdf5-1a9ee7e97cde.dat
+@PG	ID:samtools	PN:samtools	PP:bwa	VN:1.20	CL:samtools sort -@4 -T /tmp/tmpa_72_4_v/tmp -O bam -o /tmp/tmpa_72_4_v/job_working_directory/000/7/outputs/dataset_231c2a5d-42ff-40c0-86e5-37c9481a4bab.dat
 M01368:8:000000000-A3GHV:1:1104:11572:8264	117	gi|251831106|ref|NC_012920.1|	1	0	*	=	1	0	CGTAGGTGCGATAAATAATAGGATGAGGCAGGAATCAAAGACAGATACTGCGACATAGGGTGCTCCGGCTCCAGCGTCTCGCAATGCTATCGCGTGCACACCCCCCAGACGAAAATACCAAATGCATGGAGAGCTCCCGTGAGTGGTTAATAGGGTGATAGACCTGTGATCCATCGTGATGTCTTATTTAAGGGGAACGTGTGGGCTATTTAGGTTTTATGACCCTGAAGTAGGAACCAGATGTCGGATAC	9FEFAFE?EFFB/FFFFB/BBFFBF:.BFB;//FFFBFFFBBFFFFAD=;9;FFFFEA?FE=>;>CEFFFFFFFDDFFCFFEFFFFDFFAFFFFGFDGB-GGCHEGEGHGHHHHHGGHHHHHHGHHHHFHHHGEGEGGHGHHGHHHHHHHEHHHHHHHHHHHGHHFFHHHGGGGGHHHHHHHGHHHHHHHHHHHHGFGGGGHHHHEHFFGHHGHCGFGHHHHGHHGHHGGGGGGGGGGGFCCCABFBCBAB
 M01368:8:000000000-A3GHV:1:1104:11572:8264	153	gi|251831106|ref|NC_012920.1|	1	37	251M	=	1	0	GATCACAGGTCTATCACCCTATTAACCACTCACGGGAGCTCTCCATGCATTTGGTATTTTCGTCTGGGGGGTGTGCACGCGATAGCATTGCGAGACGCTGGAGCCGGAGCACCCTATGTCGCAGTATCTGTCTTTGATTCCTGCCTCATCCTATTATTTATCGCACCTACGTTCAATATTACAGGCGAACATACTTACTAAAGTGTGTTAATTAATTAATGCTTGTAGGACATGATAATAACAATTGAATG	BF;//BF;/////.A/FFFFBBFFFA;/?FDGDGFFBA.BFFGFEGGGGGGGGGGFB9:C-C.GGGC@GGGGHHGFGGGGHHHHGHGGGGFHFGGGEHFEGFCGGGHHFCGHFGHEGFGGHHHHHHFHEHHGHHHHHHGFGGGHHHGHHHHHFGHHHGGHGGHGEGGHHHGGHHHHHHHHHHGGGDHHHHHGHHHFHHHHHHGFHHHFFHHHHHHHHHHHHHHHHHHHHGGGE5GGGGGFFFFFFFCBBAA	XT:A:U	NM:i:2	SM:i:37	AM:i:0	X0:i:1	X1:i:0	XM:i:2	XO:i:0	XG:i:0	MD:Z:72A160A17
 M01368:8:000000000-A3GHV:1:1104:23670:4161	89	gi|251831106|ref|NC_012920.1|	1	37	1S250M	=	1	0	GGATCACAGGTCTATCACCCTATTAACCACTCACGGGAGCTCTCCATGCATTTGGTATTTTCGTCTGGGGGGTGTGCACGCGATAGCATTGCGAGACGCTGGAGCCGGAGCACCCTATGTCGCAGTATCTGTCTTTGATTCCTGCCTCATCCTATTATTTATCGCACCTACGTTCAATATTACAGGCGAACATACTTACTAAAGTGTGTTAATTAATTAATGCTTGTAGGACATAATAATAACAATTGAAT	BFB/F/FBBFFB/FAA/.FBFB;/./FFF9.AEGEGGGFABBBAGFFB00EEFFFBGG?GDCGEGGGGGGGHHHHGGGGGFFGFHFFDGGGHFGDFDCF@/GGGGGHGHGGHGBHHGGGEGHHGHFHGHGEGDFFGHHHFGGGEHHHHHHGHHHHHGGGGHGGGHHGHHHHGHHHHHGHHHFGGGGGHHHHHHHFGGHHHHHHGGFHHHHHHHHHHGHHHHHHHHHGGBBGCGGGGGGGFFFFFFFBAAAA	XT:A:U	NM:i:1	SM:i:37	AM:i:0	X0:i:1	X1:i:0	XM:i:5	XO:i:1	XG:i:1	MD:Z:72A177

**Comment by wm75** [comment]
Yes,

```
-@PG	ID:bwa	PN:bwa	VN:0.7.18-r1243-dirty	CL:bwa mem -t 1 -v 1 localref.fa /tmp/tmpe4tqptnr/files/1/e/5/dataset_1e5e438c-9045-4bb7-801c-2925a3b977b7.dat
-@PG	ID:samtools	PN:samtools	PP:bwa	VN:1.20	CL:samtools sort -@1 -T /tmp/tmpe4tqptnr/tmp -O bam -o /tmp/tmpe4tqptnr/job_working_directory/000/7/outputs/dataset_4e8b44cb-3da1-4b78-98ef-cbb4021dfaf7.dat
+@PG	ID:bwa	PN:bwa	VN:0.7.18-r1243-dirty	CL:bwa mem -t 4 -v 1 localref.fa /tmp/tmp7pc3ar41/files/8/a/d/dataset_8ad90aa3-318f-4277-8ddb-de65a1b1a5b6.dat
+@PG	ID:samtools	PN:samtools	PP:bwa	VN:1.20	CL:samtools sort -@4 -T /tmp/tmp7pc3ar41/tmp -O bam -o /tmp/tmp7pc3ar41/job_working_directory/000/7/outputs/dataset_a22cb5ef-88bc-4ac1-af4f-7324f04e1381.dat
```

what I expected. That `samtools sort` PG line wasn't added by the previous samtools version (just checked with old bwa-mem tool output).

No problem then, you can just switch back to `4`. Maybe add a comment to the first of these to explain that it's about the PG lines written by bwa and samtools sort.

**Comment by wm75** [comment]
Just looked at your changes with planemo serve now and there is one issue:

image

If you're selecting Yes for any of the dropdowns, you cannot tell anymore, which part of the read group information they correspond to (because the text field carrying the info is not displayed anymore). I think we need to find a solution for this.

**Comment by wm75** [comment]
Maybe a tokenized macro that lets you pass in the name of what's getting auto assigned. Sorry, that one's a bit complicated.

**Comment by wm75** [comment]
Much better, but one last request (sorry for so many of them :sweat_smile: ) that could improve this further:

image

When you look at this, then more logical labels for the select boxes would be: "Set read group identifier (ID)", "Set sample name (SM)", ...
with options:
- Auto-assign
- Set manually

and the help would read: Dataset name or collection information will be used for determining the auto-assigned value.

For the manual case, the label of the text boxes could be abbreviated to "Value for ID", "Value for SM", etc.

Sorry again, but readability issues with this section have plagued trainings for many years so this is the chance to get a real improvement in.

**Comment by wm75** [comment]
> > No problem then, you can just switch back to 4. Maybe add a comment to the first of these to explain that it's about the PG lines written by bwa and samtools sort.
> 
> Alternatively use `--no-PG` ?

Yes, but here it's probably fine to have all the actions of the wrapper documented in the final header?

**Comment by nilchia** [comment]
@wm75 Yes that makes more sense. Thank you.

No please, It helps me to learn new tricks. I like it :wink:

**Comment by bernt-matthias** [comment]
> > > No problem then, you can just switch back to 4. Maybe add a comment to the first of these to explain that it's about the PG lines written by bwa and samtools sort.
> > 
> > 
> > Alternatively use `--no-PG` ?
> 
> Yes, but here it's probably fine to have all the actions of the wrapper documented in the final header?

Yes, maybe we should come to a IUC wide consensus here. We do this quite often. For sorting is just a consequence of the Galaxy definition of bam being sorted.

**Comment by wm75** [comment]
Thanks a lot @nilchia, this looks like a real improvement!

I've pushed an additional commit, which updates the tool help for bwa and bwa-mem.

@bernt-matthias @bgruening you may want to take a closer look at this one.

File: macros/read_group_macros.xml [hunk 5]
```diff
@@ -57,15 +57,15 @@
 #if $use_rg
     #if $rg_param('read_group_id_conditional') is None
         #set $rg_id = $rg_auto_name
-    #elif $rg_param('read_group_id_conditional').do_auto_name
```
bernt-matthias, line 60: Quiet a few tools use these macros. We need to bump all of them

`grep read_group_macros.xml tools -r`

```
tools/picard/picard_AddOrReplaceReadGroups.xml:        <import>read_group_macros.xml</import>
tools/bwa_mem2/macros.xml:    <import>read_group_macros.xml</import>
tools/bwa_mem2/bwa-mem2.xml:        <import>read_group_macros.xml</import>
tools/bwa/bwa_macros.xml:    <import>read_group_macros.xml</import>
tools/bwa/bwa.xml:        <import>read_group_macros.xml</import>
tools/bwa/bwa-mem.xml:        <import>read_group_macros.xml</import>
```

Pay special attention to tests, where the used values might have been different from `true` and `false`.

macros/read_group_macros.xml
```diff
@@ -137,32 +137,34 @@
...
         <conditional name="read_group_id_conditional">
-            <expand macro="read_group_auto_name_conditional">
+            <expand macro="read_group_auto_name_conditional" label="Set read group identifier (ID)">
                 <expand macro="read_group_id_param" />
             </expand>
         </conditional>
     </xml>
     <xml name="read_group_sm_param">
-        <param name="SM" type="text" value="" label="Read group sample name (SM)" help="This value should be descriptive. Use pool name where a pool is being sequenced" />
+        <param name="SM" type="text" value="" label="Value for SM" help="This value should be descriptive. Use pool name where a pool is being sequenced" />
```
bernt-matthias, line 163: I'm a fan of short labels. But besides for ID I think the abbreviations need explanations. Moving them to the might be a compromise.

macros/read_group_macros.xml
```diff
@@ -137,32 +137,34 @@
...
+        <param name="do_auto_name" type="select" label="@LABEL@" help="Dataset name or collection information will be used for determining the auto-assigned value.">
+            <option value="false">Set manually</option>
+            <option value="true">Auto-assign</option>
+        </param>
+        <when value="true"></when>
+        <when value="false"> 
+            <yield/>
         </when>
     </xml>
     <xml name="read_group_id_param">
```
bernt-matthias, line 150: Could all these conditionals be replaced by a simple text field that is allowed to be empty. Where the case that the user leaves the field empty equals to the case where the conditional select is false.

macros/read_group_macros.xml
```diff
@@ -171,13 +173,13 @@
          as per Picard.
     -->
     <xml name="read_group_sm_param_required">
-        <param name="SM" type="text" value="" label="Read group sample name (SM)" optional="false" help="This value should be descriptive. Use pool name where a pool is being sequenced">
+        <param name="SM" type="text" value="" label="Value for SM" optional="false" help="This value should be descriptive. Use pool name where a pool is being sequenced">
             <validator type="empty_field" />
         </param>
     </xml>
     <xml name="read_group_sm_required_conditional">
```
bernt-matthias, line 180: I can not see any difference between `read_group_sm_required_conditional` and `read_group_sm_required_conditional`, or am I wrong?

tools/bwa/bwa-mem.xml
```diff
@@ -169,7 +167,7 @@ bwa mem
         <expand macro="read_group_conditional" />
 
         <conditional name="analysis_type">
-            <param name="analysis_type_selector" type="select" label="Select analysis mode">
+            <param name="analysis_type_selector" type="select" label="Select analysis mode" help="Please note that minimap2 is recommended over BWA as the aligner for long-read or contig data, for which it outperforms BWA in speed and typically in accuracy (see tool help below).">
```
bernt-matthias, line 170: I do not see how this comment relates to this parameter.

bernt-matthias (commented):
Thanks for the update. 

I have the impression that the `read_group_macros.xml` could be improved. But I would suggest to move any changes to this file to a separate PR since it has effects to some more tools (would be very welcome).

macros/read_group_macros.xml
bgruening: Those tools would not be pushed to the TS if we merge this PR, does it?

bernt-matthias: If we do not bump them they won't be pushed. But we should definitely bump them if we change this file.

wm75: Ok, but then *that* could be done in a separate PR?

bernt-matthias: This would be be my suggestion. We can ignore the linter warnings due to the boolean conditional parameter.

macros/read_group_macros.xml
wm75: The explanations are given right above in the tool interface, it's just that the macros are so nested that it's not obvious in the xml.

tools/bwa/bwa-mem.xml
bernt-matthias: Maybe just: `Please check the help if you want to align long-read or contig data` otherwise its redundant?

It would also be good if we would give a citation for this recommendation.

macros/read_group_macros.xml
nilchia: I think the wrong macro is called for the required one. It should be like:
```
    <xml name="read_group_sm_required_conditional">
        <conditional name="read_group_sm_required_conditional">
            <expand macro="read_group_auto_name_conditional" label="Set read group sample name (SM)">
                <expand macro="read_group_sm_param_required" />
            </expand>
        </conditional>
    </xml>
```
?

macros/read_group_macros.xml
wm75: sounds nice - but wouldn't work for the optional values, where, I think, empty already means do not put this tag into the RG record.

bernt-matthias:
Implemented the macro changes over here https://github.com/galaxyproject/tools-iuc/pull/5986

bgruening:
@nilchia can you please undo changes to macros/read_group_macros.xml.

bgruening (approved):

PR #5965, opened 2024-04-23
Updated FASTK tool wrapper

SaimMomin12:
FOR CONTRIBUTOR:

Hi,

Based on the last FASTK tool wrapper developed by @abueg in #5550 I have made necessary changes and updated the wrapper. The idea forward would be to create standalone suite of tools (Logex,Histex,etc) that exist within FASTK.

Best,
Saim


tools/fastk/.shed.yml
```diff
@@ -0,0 +1,10 @@
+categories:
```
bgruening, line 1: please create here a suite based shed.yml file

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,123 @@
...
+        <param name="kmer_size" type="integer" min="5" max="50" value="40" label="Enter desired k-mer size" help="Default: 40" />
+        <conditional name="sorted_table">
+            <param name="sorted_table_option" type="select" label="Sort table" help="Do you want a sorted table of all canonical k-mers and their counts? The sorted table is sorted lexicographically on the k-mer where a &lt; c &lt; g &lt; t.">
+                <option value="no">No</option>
+                <option value="yes_with_default">Yes, Default sorted </option>
+                <option value="yes_with_custom">Yes, Custom sorted </option>
+            </param>
+            <when value="no"/>
+            <when value="yes_with_default">
+                <param name="tabex_threshold_for_default" label="Tabex count threshold" help="-t: Trim all k-mers with counts less than threshold (Default: 5)" type="integer" value="5" min="1"/>
```
bgruening, line 57:
```suggestion
                <param name="tabex_threshold_for_default" argument="-t" type="integer" value="5" min="1" label="Tabex count threshold" help="Trim all k-mers with counts less than threshold" />
```

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,123 @@
...
+            </when>
+            <when value="yes_with_custom">
+                <param name="sorted_table_cutoff" type="integer" min="2" value="10" label="Enter sorted table cutoff value"/>
+                <param name="tabex_threshold_for_custom" label="Tabex count threshold" help="-t: Trim all k-mers with counts less than threshold. (Default: 5)" type="integer" value="5" min="1"/>
```
bgruening, line 61: see above, maybe move to a macro?

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,123 @@
...
+            </when>
+        </conditional>
+    </inputs>
+    <outputs>
+        <data name="fastk_out" format="tar" from_work_dir="fastk.tar" label="${tool.name} on ${on_string}: FastK files"/>
+        <data name="fastk_hist" format="binary" from_work_dir="outfiles/output.hist" label="${tool.name} on ${on_string}: FastK hist" />
```
bgruening, line 67: we should add a proper Galaxy datatype for this one

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,123 @@
...
+            </output>
+            <output name="tabex_hist" value="test03.tabex.txt"/>
+        </test>       
+    </tests>
+    <help><![CDATA[
+        FastK is a k‑mer counter that is optimized for processing high quality DNA assembly data sets such as those produced with an Illumina instrument or a PacBio run in HiFi mode.
+    
+        FastK can produce the following outputs:
+
+        1. a histogram of the frequency with which each k‑mer in the data set occurs.
```
bgruening, line 117: I find the term histogram very confusing when its a binary.

tools/fastk/fastk.xml
bernt-matthias: https://github.com/galaxyproject/galaxy/pull/17265?

tools/fastk/fastk.xml
SaimMomin12: Both parameters are independent of each other, so I think its better to be here?

tools/fastk/fastk.xml
SaimMomin12: The generated `.hist` (histogram) file in binary format can be used later in other tools downstream

tools/fastk/fastk.xml
bgruening: ok

tools/fastk/fastk.xml
bgruening: Thats true, but a user expects an image if you read "histogram" maybe? That is what I find confusing.

bgruening: Technically its correct, I'm just worried about user perception ...

SaimMomin12: Hmm you are correct. So maybe we can reframe this as follows;

 1. a (.hist) file that can be used downstream for generating histogram of the frequency for occurence of each k‑mer in the dataset.

tools/fastk/.shed.yml
```diff
@@ -0,0 +1,12 @@
+name: fastk
+owner: iuc
+categories:
+- Assembly
+description: "FastK: A K-mer counter (for HQ assembly data sets)"
+long_description: FastK is a k‑mer counter that is optimized for processing high-quality DNA assembly data sets such as those produced with an Illumina instrument or a PacBio run in HiFi mode.
+homepage_url: https://github.com/thegenemyers/FASTK
+remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/master/tools/fastk
+type: unrestricted
+suite:
```
bgruening, line 10: this does not seem to be a proper suite file? some autogenerate tags are missing.

bgruening:
Other than fixing the shed.yml I think the first tool is read to be shipped.

tools/fastk/.shed.yml
SaimMomin12: Do you have some example suite file where the missing tags can be compared to?

bgruening: https://github.com/galaxyproject/tools-iuc/blob/b8bc892a6d5eeae0b78c33058b532fcd52208179/tools/mageck/.shed.yml#L11

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,123 @@
...
+                <param name="sorted_table_cutoff" type="integer" min="2" value="10" label="Enter sorted table cutoff value"/>
+                <param name="tabex_threshold_for_custom" argument="-t" type="integer" value="5" min="1" label="Tabex count threshold" help="Trim all k-mers with counts less than threshold"/>
+            </when>
+        </conditional>
+    </inputs>
+    <outputs>
+        <data name="fastk_out" format="tar" from_work_dir="fastk.tar" label="${tool.name} on ${on_string}: FastK files"/>
+        <data name="fastk_hist_out" format="fastk_hist" from_work_dir="outfiles/output.hist" label="${tool.name} on ${on_string}: FastK hist" />
+        <data name="tabex_hist" format="txt"  label="${tool.name} on ${on_string}: Tabex output">
+            <filter> ( sorted_table['sorted_table_option'] == 'yes_with_default' or sorted_table['sorted_table_option'] == 'yes_with_custom' )</filter>
```
bernt-matthias, line 69: Just `sorted_table['sorted_table_option'] != 'no'`?

tools/fastk/fastk.xml
```diff
@@ -0,0 +1,123 @@
...
+            -t 
+        #elif $sorted_table.sorted_table_option == 'yes_with_custom':
+            -t$sorted_table_cutoff 
+        #end if
+        -T\${GALAXY_SLOTS:-8} -Noutfiles/output -Poutfiles/tmpfiles
+        #if $sorted_table.sorted_table_option == 'yes_with_default':
+            && Tabex outfiles/output.ktab -t${sorted_table.tabex_threshold_for_default} LIST > '$tabex_hist'
+        #elif $sorted_table.sorted_table_option == 'yes_with_custom':
+            && Tabex outfiles/output.ktab -t${sorted_table.tabex_threshold_for_custom} LIST > '$tabex_hist'
+        #end if
```
bernt-matthias, line 43: I mixing `FastK` and `Tabex` in one tool what we want (just wondering)? I think we should at least document it in the help, i.e. which output is produced by which program .. ? 

Also the use of the `argument` attribute (which btw seems to be missing for most parameters) is then difficult, since one does not know for which program the parameter is an argument.

bernt-matthias (approved):
Apart from the small change to the help this looks good to go from my side. Thanks a lot to all who contributed.

bgruening:
Thanks, everyone!

SaimMomin12:
Hi @astrovsky01, yes the earlier day I realized that the ktab files are necessary while I was working on `Logex` tool. I will figure out a way to retain these files and update the wrapper accordingly.

Issue #5983, opened 2024-05-02
Bowtie overcommits CPU ressources


bernt-matthias:
`bowtie` seems to be extremely efficient, i.e. `X` `GALAXY_SLOTS` are used the CPU usage is most of the time larger than `X`*100% (if checked in `top`). This seems to be because the `-p` parameter seems to control the number of (additional) alignment threads (probably plus 1 control thread). 

Additionally we pipe the result to `samtools` using potentially  `GALAXY_SLOTS`  more slots (in practice most of the time less).

In summary I would suggest to remove the pipes (to the cost of having an additional temporary file.)

PR #5985, opened 2024-05-03
bowtie2: do not overcommit

bernt-matthias:
fixes https://github.com/galaxyproject/tools-iuc/issues/5983

by using a pipe bowtie2 and samtools run in parallel. since already bowtie2 alone uses more CPU than the assigned ones (less than 1 core) we should not do this, but run them serially.
background had a user running 2 fastq files (each 2.6GB) against a file containing 500k contigs. Running bowtie there was extremely efficient (tested up to 40 cores), i.e. for `GALAXY_SLOTS`=X the CPU usage (according to top) was `(X+1)*100%` .. probably because X alignment + 1 control thread are used.  

Alternatively we could do [this](https://github.com/galaxyproject/tools-iuc/blob/921fa90438b4ce2e9cf9c4129568f4036c318d82/tools/hisat2/hisat2.xml#L351), but I do not see a reason why we should.

also properly redirects `stderr`.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/bowtie2/bowtie2_wrapper.xml
+++ b/tools/bowtie2/bowtie2_wrapper.xml
@@ -307,10 +307,10 @@ bowtie2
 ## output file
 #if str( $sam_options.sam_options_selector ) == "no" or (str( $sam_options.sam_opt ) == "false" and str($sam_options.reorder) == ''):
     > alignment.sam
-    && samtools sort --no-PG -@\${GALAXY_SLOTS:-2} -T "\${TMPDIR:-.}" -O bam -o '$output' alignment.sam
+    && samtools sort -@\${GALAXY_SLOTS:-2} -T "\${TMPDIR:-.}" -O bam -o '$output' alignment.sam
 #else if $sam_options.reorder:
     > alignment.sam
-    && samtools view --no-PG -b -o '$output' alignment.sam
+    && samtools view -b -o '$output' alignment.sam
 #else:
     > '$output'
 #end if
--- a/tools/bowtie2/bowtie2_wrapper.xml
+++ b/tools/bowtie2/bowtie2_wrapper.xml
@@ -12,6 +12,8 @@
     </requirements>
     <version_command>bowtie2 --version</version_command>
     <command detect_errors="exit_code"><![CDATA[
+## Use pipefail if available to quit with first non-zero exit code
+set -o | grep -q pipefail && set -o pipefail;
 ## prepare bowtie2 index
 #set index_path = ''
 #if str($reference_genome.source) == "history":
@@ -306,11 +308,18 @@ bowtie2
 
 ## output file
 #if str( $sam_options.sam_options_selector ) == "no" or (str( $sam_options.sam_opt ) == "false" and str($sam_options.reorder) == ''):
-    > alignment.sam
-    && samtools sort -@\${GALAXY_SLOTS:-2} -T "\${TMPDIR:-.}" -O bam -o '$output' alignment.sam
+    ## Convert SAM output to sorted BAM
+    ## using the two pipe stages has the following effect
+    ## - mapping and sorting run in parallel, during this time sort produces
+    ##   presorted temporary files but does not produce output (hence
+    ##   view does not run)
+    ## - once mapping is finished sort will start to merge the temporary
+    ##   files (which should be fast also on a single thread) gives the
+    ##   sorted output to view which only compresses the files (now
+    ##   using full parallelism again)
+    | samtools sort -l 0 -T "\${TMPDIR:-.}" -O bam | samtools view --no-PG -O bam -@ \${GALAXY_SLOTS:-1} -o '$output'
 #else if $sam_options.reorder:
-    > alignment.sam
-    && samtools view -b -o '$output' alignment.sam
+    | samtools view --no-PG -b -o '$output' 
 #else:
     > '$output'
 #end if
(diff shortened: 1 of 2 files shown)


bwlang:
One problem with the non piping is that you have to write to the disk one more time... that can be very expensive for some environments. Maybe there is a way to partition the thread usage so that samtools can use some of the available resources.

tools/bowtie2/bowtie2_wrapper.xml
```diff
@@ -303,14 +301,16 @@ bowtie2
...
 #if $save_mapping_stats
-    2> '$mapping_stats'
+    2> >(tee '$mapping_stats' >&2) 
 #end if
 
 ## output file
 #if str( $sam_options.sam_options_selector ) == "no" or (str( $sam_options.sam_opt ) == "false" and str($sam_options.reorder) == ''):
-    | samtools sort --no-PG -@\${GALAXY_SLOTS:-2} -T "\${TMPDIR:-.}" -O bam -o '$output'
+    > alignment.sam
+    && samtools sort --no-PG -@\${GALAXY_SLOTS:-2} -T "\${TMPDIR:-.}" -O bam -o '$output' alignment.sam
```
bwlang, line 310: why no-PG... i think it's good to have the headers that describe the tools that worked on a bam in place.

wm75:
> Alternatively we could do [this](https://github.com/galaxyproject/tools-iuc/blob/921fa90438b4ce2e9cf9c4129568f4036c318d82/tools/hisat2/hisat2.xml#L351), but I do not see a reason why we should.

Seems you have given the solution for HISAT2 quite some thought. What makes you think it's *not* the way to go here?

wm75:
the one extra thread needed by samtools sort or view could be subtracted from the ones used by bowtie2 with logic like that used in the samtools collection of tools: https://github.com/galaxyproject/tools-iuc/blob/57fa4dcee2aaa115df4a5b7173a3a8fb889bc5a3/tool_collections/samtools/macros.xml#L157-L160

tools/bowtie2/bowtie2_wrapper.xml
bwlang: i see that complexity for testing, but i think that bams produced by galaxy do not only stay within galaxy.
E.g. they get submitted to SRA as data associated with scientific publications.

bernt-matthias:
> Seems you have given the solution for HISAT2 quite some thought. What makes you think it's not the way to go here?

It just seemed to be to complicated.  Give me some time I will change it. Probably I would go for `GALAXY_SLOTS - 1` threads for bowtie2 (here I will need to make sure that the result will be at least 1, since bowtie [seems](https://github.com/BenLangmead/bowtie2/blob/a43fa6f43f54989468a294967898f85b9fe4cefa/bt2_build.cpp#L315) not to allow 0).

tools/bowtie2/bowtie2_wrapper.xml
wm75: Yes, we should come to an IUC-wide decision with this, and I would go with @bwlang's arguments for recording everything in the header.

One case that comes to my mind immediately where --no-PG seems justified to me is with the reheader tool itself because you use it explicitly to shape the header like you think it's best - and even there we only make it optional to omit the additiona @PG record (https://github.com/galaxyproject/tools-iuc/blob/57fa4dcee2aaa115df4a5b7173a3a8fb889bc5a3/tool_collections/samtools/samtools_reheader/samtools_reheader.xml#L19-L21).

wm75: Here's another case of what I'd consider a valid use of --no-PG:
https://github.com/galaxyproject/tools-iuc/blob/57fa4dcee2aaa115df4a5b7173a3a8fb889bc5a3/tool_collections/bamtools/bamtools_split/bamtools_split_ref.xml#L13

bamtools_split_ref first gets the header of the BAM it's operating on so that it can add it back to "empty" split results that don't have reads in them.

bwlang: agreed - those seem reasonable to me.  I can't think of a case where view is important to the analysis history of a file - it would be implied if we had some analysis tool that only takes sam for example.

wm75: right, so @bernt-matthias in your HISAT2 sorting solution:
https://github.com/galaxyproject/tools-iuc/blob/57fa4dcee2aaa115df4a5b7173a3a8fb889bc5a3/tools/hisat2/hisat2.xml#L351
I'd say that the `--no-PG` with samtools view is fine (since it's an implementation detail of how you write the result of sorting), but the samtools sort should be added as a PG record.

wm75: I guess when you use samtools view for read filtering, that's when it deserves its own record?

tools/bowtie2/bowtie2_wrapper.xml
```diff
@@ -113,12 +113,22 @@ set -o | grep -q pipefail && set -o pipefail;
     ln -f -s '${library.input_1}' ${read1} &&
 #end if
 
+
+## compute number of threads to be used for bowtie2
+## the bowtie parameter -p specifies the number of alignment threads to use (in
+## addition to a control thread) # just using GALAXY_SLOTS will lead to
+## overcommiting ressources (in particular because there may be a samtools view
```
wm75, line 120:
```suggestion
## overcommiting ressources (in particular because there may be a samtools sort
```

wm75 (approved):
Looks like a good solution to me!

PR #5986, opened 2024-05-03
update readgroup macros

bernt-matthias:
was originally suggested here https://github.com/galaxyproject/tools-iuc/pull/5963 but better done separately .. with bumping all tools using the macros

Just noticed that `bowtie2` uses more or less the same macros. Should we work toward that also bowtie2 uses these macros? Or the other way round?

FOR CONTRIBUTOR:

### DIFF
--- a/macros/read_group_macros.xml
+++ b/macros/read_group_macros.xml
@@ -57,7 +57,7 @@
 #if $use_rg
     #if $rg_param('read_group_id_conditional') is None
         #set $rg_id = $rg_auto_name
-    #elif $rg_param('read_group_id_conditional').do_auto_name
+    #elif $rg_param('read_group_id_conditional').do_auto_name == "true"
         #set $rg_id = $rg_auto_name
     #else
         #set $rg_id = str($rg_param('read_group_id_conditional').ID)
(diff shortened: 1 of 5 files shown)


wm75:
So if there are no changes required to any tool to adjust for the macros update, then why not do it as part of the bwa PR?

Other tools would get the changes on their next update naturally, and we don't create additional wrapper versions.

wm75:
I see, the picard tools are a single repo so we would need to make sure  that #5963 gets merged before #5297 and that the latter bumps the add or replace rg tool.

bernt-matthias:
> So if there are no changes required to any tool 

It's a change to the tool :) But alongside with https://github.com/galaxyproject/tools-iuc/pull/5297#issuecomment-2095433475 I will refrain from this here.

wm75:
Sorry @bernt-matthias if that came across wrong. I'm also not entirely sure which is the better way to go here,I'd just like to get the bwa update merged this week and, if possible without breaking other stuff :-)

wm75:
> > So if there are no changes required to any tool
> 
> It's a change to the tool :) But alongside with [#5297 (comment)](https://github.com/galaxyproject/tools-iuc/pull/5297#issuecomment-2095433475) I will refrain from this here.

Sure, what I meant is none of the tools here will require additional work in a separate PR to make it compatible with the new macro version.

bernt-matthias:
> that came across wrong

no worries :) .. all good. 

IMO Best would be to just revert the changes to the macro and ignore the linter warnings. Deployment will ignore all warnings.

PR #6000, opened 2024-05-08
Chewbbaca update

nilchia:
FOR CONTRIBUTOR:

This update resolves two issues:
1. chewBBACA AlleleCall checks the file extensions. So it does not recognize a FASTA file in the history without an extension.
2. ExtractCgMLST works fine with a normal chewBBACA pipeline but if the user gets the allele profiles from another package or wants to analyze the joined_profile results, there would be a problem because it only takes AlleleCall output. With this update, the user can define the inputs manually.


tools/chewbbaca/AlleleCall.xml
```diff
@@ -8,7 +8,7 @@
         mkdir 'input' &&
         mkdir 'schema' &&
         #for $file in $input_file
-        ln -sf '$file' 'input/${file.element_identifier}' &&
+        ln -sf '$file' 'input/${file.element_identifier}.fa' &&
```
bgruening, line 11: Do we need the element identifier here? Or can we just name it input/foo.fa

nilchia: AlleleCall accepts multiple input FASTAs. I think in that case it overwrites the inputs. Am I right?

bgruening: You can use foo_1 foo_2 please search in this repo for enumerate ... We do that often

nilchia: Cool!
I did it with:
```
        #for i,file in enumerate($input_file)
        ln -sf '$file' 'input/fasta_${i}.${file.ext}' &&
        #end for
```

One thing, the result of AlleleCall is a tabular file with row names corresponding to the FASTA file names.
So the result would be something like:
image
 
rather than:
image

bgruening: yeah, thats bad. Ok lets keep the element_identifier then, but please sanatize them as here for example: https://github.com/galaxyproject/tools-iuc/blob/1056d5cc83082613f5d92b22b4e531fa9d505b91/tools/picard/picard_macros.xml#L33

bgruening (approved):

PR #6026, opened 2024-05-23
update VirHEAT to v0.7, add -a and -s

PlushZ:
FOR CONTRIBUTOR:


PlushZ:
Required new virheat version updated in bioconda

tools/virheat/virheat.xml
```diff
@@ -72,6 +84,12 @@ virheat
         </conditional>
         <param argument="--sort" type="boolean" truevalue="--sort" falsevalue="--no-sort" checked="false" label="Sort sample names alphanumerically"/>
         <param argument="--min-cov" type="integer" min="1" max="1000" value="20" label="Display mutations covered at least x time"/>
+        <repeat name="scores" title="Scores" help="Include visualizations of additional scores (e.g., MAVE scores for binding affinity, expression level, antibody escape, etc.) mapped to mutations on the heatmap.">
+            <param name="scores_file" type="data" format="csv" optional="false" label="CSV file" help="CSV file containing scores"/>
```
bgruening, line 88: How does such a file look like?

tools/virheat/virheat.xml
```diff
@@ -72,6 +84,12 @@ virheat
...
+            <param name="pos_col" type="text" optional="false" label="Mutation position column" help="The name of the column in the CSV file that contains the mutation positions in classic notation (e.g., T430Y)"/>
```
bgruening, line 89: Can you use here https://docs.galaxyproject.org/en/latest/dev/schema.html#data-column

bgruening (commented):
Bioconda is working again.

tools/virheat/virheat.xml
PlushZ: > How does such a file look like?

it's the one that we import from maveDB, like [this](https://github.com/PlushZ/tools-iuc/blob/update-virheat/tools/virheat/test-data/MaveBindRBD.csv)

bgruening: Is this clear to the user? Do you want to add this somewhere?

PlushZ: It can be any csv file with mutation positions and scores, even if the user manually does this. The only requirement is 2 columns: position and score. I think it should be clear.

tools/virheat/virheat.xml
PlushZ: it works both with tsv and csv, but doesn't work with tabular, tabular does not fill the column_names metadata and therefore the data_meta filter would not work

bgruening (approved):

PR #6041, opened 2024-05-29
rename all config dirs to avoid conflicts when running parallel tasks…

bdartigues:
… in galaxy workflow

FOR CONTRIBUTOR:


tools/dimet/dimet_abundance_plot.xml
```diff
@@ -9,13 +9,13 @@
...
     <command detect_errors="exit_code"><![CDATA[
-    @INIT_CONFIG@
+    @INIT_CONFIG_AB_PLOT@
     @INIT_ABUNDANCE_PLOT@
     @INIT_ABUNDANCE_PLOT_CONDITIONS@
     @INIT_TIMEPOINTS@
     @INIT_ENRICHMENT_METABOLITES@
     HYDRA_FULL_ERROR=1 python -m dimet
-        -cp '$__new_file_path__/config'
+        -cp '$__new_file_path__/config_ab_plot'
```
bgruening, line 18: please remind me why are we using `$__new_file_path` here and not just `./config_temp/config_ab_plot`? So the current unique working directory every job gets?

bdartigues: I don't remember exactly why we have chosen $__new_file_path ? Do you think it is better to use ./config_temp/

bernt-matthias: `new_file_path` is the same for all jobs (its configured Galaxy instance wide), i.e. all tools can access `'$__new_file_path__/config_ab_plot'` and in particular all jobs for `dimet_abundance_plot` will use the same path. This might be an advantage if there is some sort of caching going on. Otherwise the suggestion of @bgruening seems better.

tools/dimet/macros.xml
```diff
@@ -597,6 +597,79 @@
...
+    <token name="@INIT_CONFIG_TIMECOURSE@"><![CDATA[
+    mkdir -p '$__new_file_path__/config_timecourse' &&
+    touch '$__new_file_path__/config_timecourse/config.yaml' &&
+    ]]></token>
+    <token name="@REMOVE_CONFIG_TIMECOURSE@"><![CDATA[
+    && rm -r '$__new_file_path__/config_timecourse'
+    ]]></token>
+
+
+
```
bernt-matthias:
```suggestion
```

bernt-matthias (approved):

tools/dimet/dimet_abundance_plot.xml
bdartigues: Thanks for the changes but I thought about changing the configuration directory because when I run a workflow from isocor to tracegroomer and to all DIMet tools, I get errors due to using the same configuration directory at runtime. Using hydra and the config.yaml file causes this error and I thought the best solution was to create an individual configuration directory for each tool. I'm going to replace the $__new_file_path__/ with ./ but I think you should keep a specific configuration directory for each tool to avoid this problem. what do you think?

Translated with DeepL.com (free version)

bernt-matthias: Hi @bdartigues. In galaxy each job has a job working directory. `.config` will be created within this directory. Hence it's not needed to name them differently since no jobs (or tools) share the jow working dir.

bdartigues: Oh my bad.. So the problem was due to another reason.... Ok i will keep your changes. Many thnaks for your help;-)

tools/dimet/macros.xml
bdartigues: @bernt-matthias I don't know what's happening, i got a warning with numpy version and in our bioconda recipe the maximal version is 1.24.0.
```
/usr/local/lib/python3.11/site-packages/scipy/__init__.py:155: UserWarning: A NumPy version >=1.18.5 and <1.26.0 is required for this version of SciPy (detected version 1.26.4
  warnings.warn(f"A NumPy version >={np_minversion} and <{np_maxversion}"
```

bernt-matthias: Actually the minimal numpy version is 1.26.4 https://github.com/bioconda/bioconda-recipes/blob/87bc681c355a1fea82de77b0242e64d09bbb7837/recipes/dimet/meta.yaml#L32C20-L32C21

Guess the scipy & numpy pinnings need to be fixed.

bernt-matthias:
The last changes make no sense to me. The new_file_path thing is to my understanding a global directory where data might get mixed up during different tool runs. If this makes the tests pass (and my understanding is correct) something goes horribly wrong.

bdartigues:
In fact, this is the procedure that was accepted during the previous pull request, so I thought it corresponded to the expected standards. I've tried replacing $__new_file_path__/config with ./config but it doesn't work, no file is output. Is it possible that a warning due to the version of scipy not matching that of numpy could be preventing output files from being produced? After testing locally with planemo serve, I don't have any output files.

bdartigues:
So I tested again with config, and now it seems to work. i just remove the -cp option of hydra and i changed the hydra.run.dir that pointed the working directory to a different directory than the config current directory and that caused the error. Many tha lks for your remarks. I think now it is clean and fixed

bgruening (approved):

PR #6054, opened 2024-06-05
Column-join: fix fill character for files containing only a header

RZ9082:
FOR CONTRIBUTOR:

@bernt-matthias


bernt-matthias:
I do not understand the problem yet. Can you describe 

- the inputs
- the expected output
- the problem with the current output?

RZ9082:
> I do not understand the problem yet. Can you describe
> 
>     * the inputs
> 
>     * the expected output
> 
>     * the problem with the current output?

Yeah. for example If you have 2 tabular inputs:

- [input1](https://github.com/galaxyproject/tools-iuc/blob/02d63808406637a5b6512e68660581827f2e73f1/tools/collection_column_join/test-data/input1_fill_char_test.tabular) with some data.
- [input2](https://github.com/galaxyproject/tools-iuc/blob/02d63808406637a5b6512e68660581827f2e73f1/tools/collection_column_join/test-data/input2_fill_char_test.tabular) containing only a header.

The join command does not fill the empty cells with the fill character:
Screenshot from 2024-06-05 16-48-35

But actually this should be the expected [output](https://github.com/galaxyproject/tools-iuc/blob/02d63808406637a5b6512e68660581827f2e73f1/tools/collection_column_join/test-data/output_fill_char_test.tabular).

bernt-matthias:
If I get it correctly the problem is that we remove the header first. With the header `join` could use it to determine the number of columns, without the header join does not know how many columns are there.

```
       -o FORMAT
              obey FORMAT while constructing output line
 ```


```
FORMAT is one or more comma or blank sepa‐
       rated specifications, each being 'FILENUM.FIELD' or '0'.  Default FORMAT outputs the join field, the remaining fields from FILE1, the remaining fields from FILE2, all separated by CHAR.   If  FORMAT
       is the keyword 'auto', then the first line of each file determines the number of fields output for each line.
```

So we could use the header to determine the format. But seems not trivial with bash.

PR #6055, opened 2024-06-05
fix DM of plasmidfiner?

bgruening:
discovered by @natefoo ... is that the proper fix? Probably needs a bump.


bernt-matthias (approved):
At least consistent to all other data table config files. 

Bump can not hurt.

PR #6061, opened 2024-06-07
Chewbbaca update on PrepExternal and CreateSchema

nilchia:
FOR CONTRIBUTOR:

This PR updates 2 chewBBACA tools:
* 1- **CreateSchema**: It solves the same issue we had with `chewbbaca_AlleleCall` (chewBBACA couldn't check the FASTA file extension.
* 2- **PrepExternalSchema**: This tool was optimized to work well with chewbbaca_CreateSchema output but it does not make sense as if the user already has used that tool, The `PrepExternalSchema` is not needed anymore. (We also have error reports from this tool)


tools/chewbbaca/CreateSchema.xml
```diff
@@ -61,6 +63,16 @@
                 </assert_contents>
             </output>
         </test>
+        <test expect_num_outputs="1">
+            <param name="input_file" value="GCA_000007265.1_ASM726v1_genomic"/>
```
bernt-matthias, line 67: Is there an extension missing in the filename. Can you add a comment on what the intension of the test is?

bernt-matthias (commented):
We need to bump the tool version.

tools/chewbbaca/CreateSchema.xml
nilchia: Yes. The problem is that this tool does not check the file type and looks for the file extension.
So if you have a FASTA file in your history with the name "File" and not "File.fa" it will give an error and say there is no FASTA file.
So I changed the XML so that on line 12 when creating a symlink a file extension is added.
I added this test to see if this works with a FASTA file with no extension

PR #6075, opened 2024-06-11
Genome coverage for bam, bed, gff as a bigwig with optional len file

fubar2:
apropos https://github.com/galaxyproject/galaxy/pull/18356

Use case is VGP workflows where the reference assembly is constantly being improved until it is finished. A custom genome could work but would need to be manually revised for every assembly update. Possible that this would always be done manually, but fragile and likely to lead to failures from a missing contig in a new assembly so this PR allows a chromosome length file to be supplied. 

This amalgamates and fixes some built-in converters. Based on CONVERTER_bed_gff_or_vcf_to_bigwig_0 - Test added and dependencies updated to latest versions.

Advantage is an optional len file so contigs and lengths are correct at each workflow run for an assembly where dbkey makes no real sense yet.

Note that this estimates a bigwig of genome coverage for the intervals in a bed or gff3 - it ignores bed file scores - depth of coverage is what is wanted here. 

A bed track can be used to inspect scores but this is where the bed has millions of rows and is just too big for scores to be useful as JBrowse2 tracks.

FOR CONTRIBUTOR:


fubar2:
Darn. I cannot get an inbuilt hg38 test to work.
Is the recipe available anywhere?
The other three tests are fine. Advice appreciated.

bernt-matthias (commented):
Regarding the error `tool-data/shared/ucsc/chrom/hg38.len` is just not there in Galaxy by default. Seems that there are non at all.

I would suggest to add ucsc tools to the ucsc folder here. 

Wondering if we should implement this in separate tools (and maybe join as workflow).

fubar2:
Finally got the tests working.

@bernt-matthias:
> Wondering if we should implement this in separate tools (and maybe join as workflow).

If I thought that, I'd be offering 3 different tools! This makes sense to me as is:

 1. the code is relatively simple 
 2. this will make my life a lot easier for VGP JBrowse2 generating workflows where bigwig is going to be the most widely used track format because they have huge multimillion row gff and bed, and 50GB bams that are far too big and slow to be tracks.
 3. there is already converter that does 3 different inputs - it's broken but it exists - would you argue for breaking it up?
 4. this fixes CONVERTER_bed_gff_or_vcf_to_bigwig_0, but fixes the broken gff conversion and replaces (broken) vcf with bam - so it does what 3 existing built-in converters do.
 5. This is for assemblies where there is no useful dbkey.

tools/bbgbigwig/bam_bed_gff_to_bigwig.xml
```diff
@@ -0,0 +1,106 @@
...
+    python '$__tool_directory__/gff_to_bed_converter.py' < '$input1' > input2 && 
+#else:
+    ln -s '$input1' input2 &&
+#end if
+#if $input1.ext == "bam":
+    bedtools genomecov -bg -split -ibam input2 | 
+#else
+    bedtools genomecov -bg -i input2 -g ./CHROMFILE |
+#end if
+LC_COLLATE=C sort -k1,1 -k2,2n > temp.bg &&
```
bgruening, line 27: Should we use bedrolls sort here?

bgruening (approved):

tools/bbgbigwig/bam_bed_gff_to_bigwig.xml
fubar2: Sure - what's that?
Suggest code - happy to adjust if it's a worthwhile improvement?

bgruening: Lol sorry. Mobile... bedtools sort ... :-)

fubar2: Ha ha - I actually RTM:
```
Disclaimer: it should be noted that bedtools sort is merely a convenience utility, as the UNIX sort utility will sort BED files more quickly while using less memory. For example, UNIX sort will sort a BED file by chromosome then by start position in the following manner:

sort -k 1,1 -k2,2n a.bed
chr1 1   10
chr1 80  180
chr1 750 10000
```

bgruening:
Awesome! thanks

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/9522222084

fubar2:
> Unexpected HTTP status code: 500: {"err_msg": "Metadata may have been defined for some items in revision '498748c87252'.  Correct the following problems if necessary and reset metadata.<br/><b>bam_bed_gff_to_bigwig.xml<\/b> - This file refers to a file named <b>dbkeys.loc<\/b>.  Upload a file named <b>dbkeys.loc.sample<\/b> to the repository to correct this error.<br/>"}
Failed to update repository contents for repository 'bbgbigwig' on the main Tool Shed.
Repository metadata updated successfully for repository 'bbgbigwig' on the main Tool Shed.

Will add the suggested sample shortly?

PR #6078, opened 2024-06-12
Add missing format attributes

mvdbeek:
We can find them with
```
planemo lint -s command,tool_xsd,tests,general,help,inputs,citations,xml_order  --recursive
```
If we don't set an output format the hda extension becomes null in the database, and we fail at various places in the galaxy codebase. We can make Galaxy more resilient to it, but this is an easy fix here. Galaxy issue is here: https://github.com/galaxyproject/galaxy/issues/18383

FOR CONTRIBUTOR:

### DIFF
--- a/tools/khmer/extract-partitions.xml
+++ b/tools/khmer/extract-partitions.xml
@@ -33,7 +33,7 @@ output
     <outputs>
         <data name="distribution" format="txt" from_work_dir="output.dist" label="${tool.name} on ${on_string}: Partition size distribution" />
         <collection name="groups-of-partitions" type="list" format="fasta">
-            <discover_datasets pattern="__name__" directory="output" />
+            <discover_datasets pattern="__name__" directory="output" format="fasta"/>
         </collection>
     </outputs>
     <tests>
--- a/tools/khmer/extract-partitions.xml
+++ b/tools/khmer/extract-partitions.xml
@@ -23,7 +23,7 @@ output
     </command>
     <inputs>
         <expand macro="input_sequences_filenames" />
-        <param argument="--max-size" name="max_size" type="integer" label="Max group size" value="1000000"
+        <param argument="--max-size" type="integer" label="Max group size" value="1000000"
             help="No more than this many number of sequences will be stored in each output"/>
         <param name="min_partition_size" type="integer" label="Min partition size" value="5"
             help="The minimum partition size worth keeping (--min-partition-size/-m)" />
@@ -32,7 +32,7 @@ output
     </inputs>
     <outputs>
         <data name="distribution" format="txt" from_work_dir="output.dist" label="${tool.name} on ${on_string}: Partition size distribution" />
-        <collection name="groups-of-partitions" type="list">
+        <collection name="groups-of-partitions" type="list" format="fasta">
             <discover_datasets pattern="__name__" directory="output" />
         </collection>
     </outputs>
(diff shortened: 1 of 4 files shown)


tools/mothur/chimera.check.xml
```diff
@@ -67,8 +67,8 @@ echo 'chimera.check(
     <outputs>
         <expand macro="logfile-output"/>
         <data name="chimeracheck.chimeras" format="txt" from_work_dir="fasta.chimeracheck.chimeras" label="${tool.name} on ${on_string}: chimeracheck.chimeras"/>
-        <collection name="images" type="list" label="${tool.name} on ${on_string}: SVG images">
-            <discover_datasets pattern="(?P&lt;designation&gt;.*)\.chimeracheck\.svg"/>
+        <collection name="images" type="list" label="${tool.name} on ${on_string}: SVG images" format="svg">
```
bernt-matthias, line 70: Is it needed for the `collection` and `discover_datasets`?

tools/khmer/extract-partitions.xml
```diff
@@ -23,7 +23,7 @@ output
     </command>
     <inputs>
         <expand macro="input_sequences_filenames" />
-        <param argument="--max-size" name="max_size" type="integer" label="Max group size" value="1000000"
+        <param argument="--max-size" type="integer" label="Max group size" value="1000000"
```
bernt-matthias, line 26: Could you add `ftype` to the test outputs for the cases that have tests ?

mvdbeek: what's the point of that ? we're adding a required attribute, and the framework dictates that the output must be fasta.

tools/mothur/chimera.check.xml
bernt-matthias: The workflow editor argument is good. And anyway the redundancy does not hurt.

But I guess the `xsd` might not be 100% correct here. The `collection` `format` is the default that may be overwritten by the `format` attribute of `discover_datasets` (and probably also by patterns). See [this](https://github.com/bernt-matthias/galaxy/blob/454e755f064485ae4ab15542a2401b08e424be60/test/functional/tools/collection_creates_pair_format.xml#L13) test. If you agree I will fix the xsd accordingly.

mvdbeek: The test you linked doesn't cover this, since it's not using afaict.

bernt-matthias:
Agreed. How should we handle this in the light of broken mothur. Do we want a deployment to TS? Havin the khmer changes in a separate PR might simplify the problem a bit. 

That's something that we might want to think about (independent of the PR): 

```
docker: [DEPRECATION NOTICE] Docker Image Format v1 and Docker Image manifest version 2, schema 1 support is disabled by default and will be removed in an upcoming release. Suggest the author of quay.io/biocontainers/mothur:1.39.5--h284faf4_4 to upgrade the image to the OCI Format or Docker Image manifest v2, schema 2. More information at https://docs.docker.com/go/deprecated-image-specs/.
```

mvdbeek:
What problem do you mean ? Yes, this should be deployed.

bernt-matthias:
The problem that we do not have successful tests for mothur for ages.

mvdbeek:
I'm not sure I see a problem there. It's not that the PR influences tool execution one way or another, so if we block deployment on tests we can just upload the changes by hand.

mvdbeek:
Can we merge this ?

bernt-matthias:
Sure. Sorry.

mvdbeek:
Thanks! Uploading the repositories now.

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/9547078323

PR #6080, opened 2024-06-12
Update Stringtie wrapper to expose the -s option (and change tests)

lldelisle:
FOR CONTRIBUTOR:

While I was checking the tests I realized that most of them were not checking that the parameter used (`fraction`) had an impact because there was no gene with multiple transcripts. That's why I propose a new input bam and updated the tests.


null
bgruening, line 1: Not sure, what this here is, can you try to remove it please?

lldelisle: Oops

PR #6088, opened 2024-06-14
Create dbkeys_loc.sample

fubar2:
@mvdbeek:
Seems that tool-data is no longer a thing?
Removed it and put an empty dbkey_loc.sample into test-data/ 
Seeing if this fixes the deployment problem - but needs to be deployed (!)


fubar2:
Will close this and try again - planemo does not like an empty tool_data_conf.xml.sample?
>Unexpected HTTP status code: 500: {"err_msg": "Error attempting to parse file tool_data_table_conf.xml.sample: Document is empty, line 1, column 1 (tool_data_table_conf.xml.sample, line 1)"}

PR #6089, opened 2024-06-15
Fix bbgbigwig deployment lack of dbkey_loc.sample

fubar2:
@mvdbeek:
added a dbkey_loc.sample to test-data to try to fix that deployment problem?
tool-data seems to be MIA in the IUC repo.
Planemo does not like an empty tool_data_table_conf.xml.sample
update to closed PR #6088

### DIFF
--- a/tools/bbgbigwig/test-data/dbkeys.loc.sample
+++ b/tools/bbgbigwig/test-data/dbkeys.loc.sample
@@ -1,0 +1,2 @@
+#<dbkey>	<display_name>	<len_file_path>
+hg38	hg38	testing.len
--- a/tools/bbgbigwig/test-data/dbkeys.loc.sample
+++ b/tools/bbgbigwig/test-data/dbkeys.loc.sample
@@ -1,2 +1,0 @@
-#<dbkey>	<display_name>	<len_file_path>
-hg38	hg38	testing.len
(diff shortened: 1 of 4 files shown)


tools/bbgbigwig/tool_data_table_conf.xml.sample
```diff
@@ -2,6 +2,7 @@
     <!-- Locations of dbkeys and len files under genome directory -->
     <table name="__dbkeys__" comment_char="#">
         <columns>value, name, len_path</columns>
-        <file path="tool-data/dbkeys.loc" />
+        <file path="test-data/dbkeys.loc" />
```
bgruening, line 5: Should the really point to test not tool-data?

fubar2: Yes - fixed.
The deployment failed *probably* because tool-data had never been added :(

fubar2:
@bgruening: hang about. figured out a different but related problem - retesting the deployment now - a test I don't normally run..

tools/bbgbigwig/test-data/dbkeys.loc.sample
```diff
@@ -0,0 +1,2 @@
+#<dbkey>	<display_name>	<len_file_path>
```
bernt-matthias, line 1: This file is not needed.

fubar2: ok. removed - thanks @bernt-matthias

bgruening:
Thanks!

PR #6100, opened 2024-06-20
Fix HiCExplorer wrappers

lldelisle:
I realize that these 2 wrappers are currently modifying inputs. This breaks reproducibility!

### DIFF
--- a/tools/hicexplorer/hicMergeMatrixBins.xml
+++ b/tools/hicexplorer/hicMergeMatrixBins.xml
@@ -11,8 +11,8 @@
             --matrix 'matrix.$matrix_h5_cooler.ext'
             --numBins $numBins
             $runningWindow
-            --outFileName 'matrix.$matrix_h5_cooler.ext'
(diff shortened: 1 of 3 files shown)


lldelisle:
@bgruening, this bug really needs to be fixed, this is really annoying to have wrappers that modify inputs...
The tests of 'hicbuildmatrix' are failing because HiCExplorer is not compatible with pandas 2.0. I guess, this should be fixed in the docker quay.io/biocontainers/hicexplorer:3.7.2--pyhdfd78af_1, no?

bgruening:
I think we should fix the hicexplorer bioconda recipe and pin it to pandas<2. If this was done, we need to rebuild a new container via the multi-tool container repo of biocontainers. Sorry on my mobile will try to add links later or fix it.

lldelisle:
The thing is that HiCExplorer is 3.7.4 now. So do you think we should update the wrappers to 3.7.4?

PR #6121, opened 2024-07-02
Make bed features from nonparametric continuous extreme bigwig value regions. 

fubar2:
FOR CONTRIBUTOR:
* [*] I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [*] License permits unrestricted use (educational + commercial)
* [*] This PR adds a new tool or tool collection

## Make bed features for continuous bigwig outlier high or low value regions

For building some of the [NIH MARBL T2T assembly polishing](https://github.com/marbl/training) tools as Galaxy workflows for the VGP. Idea is to identify regions of very low or high coverage, or of sparse or dense dinucleotide repeat regions for example, as bed features that can be combined into a single, summary "issues" or "anomalies" track for a new (VGP or other) assembly. Bigwig regions as long or longer than `minwin` (10 works well) where all values are at or above (below) a defined quantile cutoff such as 0.99 (0.01) are written to a bed file. 

Both high and low features combined, and high-only/low-only bed files can be created. Multiple bigwigs can be combined and bed features have the provided bigwig label as annotation - only makes sense if they share the same reference so all the features can be displayed. 

The quantile cut point values are estimated and applied over each chromosome, then turned into contiguous regions of `minwin` length using a few lines of [borrowed numpy code](http://gregoryzynda.com/python/numpy/contiguous/interval/2019/11/29/contiguous-regions.html). Makes it look easy and is very fast.

Sample images below from the JBrowse2 tool and Arang's HG002 human assembly paternal haplotype. Wrote a new JB2 plugin to show bed feature scores high as red and low as blue, in the sample images below. 

Whole paternal chromosome 14 view - bigwig is top track, "issues" bed is lower track:
image

Zooming in to start of the chromosome, showing details of that long region of low coverage and peaks of high coverage:
image

Some obvious differences in the red/blue bed region distributions between chromosomes 14,15,21,22 and Y in this "all contigs" view - explanations welcomed!

image

[pybigtools](https://github.com/jackh726/bigtools) is used for the bigwig interface. 

### Why not use DeepTools windowed mean values for this purpose?
bigwig coverage values have a very long right tail and look more like a poisson than a normal distribution so a  parametric approach is problematic for guaranteeing some level of control for type 1 errors (false positives) in the context of billions of bigwig values. Using an empirical windowed quantile cut-off this way, the probability of 10 consecutive values at or above a 99th centile cutoff by chance gives 0.01<sup>10</sup> as a generalisable false positive rate. See warning about the low cutoff in sparse truncated distributions such as coverage

### Quantiles per chromosome rather than quantiles for the whole bigwig 
Not feasible to hold all contigs in the entire decoded bigwig in RAM to estimate quantiles. It may be
better to sample across all chromosomes so as not to lose any systematic differences between them - the current method will hide those differences unfortunately. Sampling might be possible. Looking at the actual quantile values across a couple of test bigwigs suggests that there is not much variation between chromosomes but there's an optional tabular report to check them for each input bigwig.


tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
```diff
@@ -0,0 +1,285 @@
...
+                        if seg[1] - seg[0] >= self.bedwin:
+                            bedlo.append((chr, seg[0], seg[1], '%s_lo' % (bwlabel), -1))
+                bwmean = np.mean(bw)
+                bwstd = np.std(bw)
+                bwmax = np.max(bw)
+                nrow = np.size(bw)
+                bwmin = np.min(bw)
+                restab.append('%s\t%s\t%d\t%f\t%f\t%f\t%f\t%f\t%f' % (bwlabel,chr,nrow,bwmean,bwstd,bwmin,bwmax,self.bwtop,self.bwbot))  
+        stable = '\n'.join(restab)      
+        print(stable)
```
bgruening, line 151: can be removed

tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
```diff
@@ -0,0 +1,285 @@
...
+#end if
+#if $qlo:
+--qlo '$qlo'
+#end if
+#if $tableout == "create":
+ --tableoutfile '$tableoutfile'
+#end if
+]]></command>
+  <configfiles>
+    <configfile name="runme"><![CDATA[#raw
```
bgruening, line 51: If you use configfiles you can hardcode the params into the scrip and avoid using argparse. If you want to use argparse I would store the script next to the tool as python file so we get the lining for free.

tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
```diff
@@ -0,0 +1,285 @@
...
+            bedf.write('\n'.join(beds))
+            bedf.write('\n')
+        print('Wrote %d bed regions to %s' % (len(bed), bedfname))
+
+    def makeBed(self):
+        bedhi = []
+        bedlo = []
+        bwlabels = self.bwlabels
+        bwnames = self.bwnames
+        print('bwnames=', bwnames, "bwlabs=", bwlabels)
```
bgruening, line 116: remove?

tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
```diff
@@ -0,0 +1,285 @@
...
+    a('-l', '--qlo',default=None, type=float)
+    a('-i', '--qhi',default=None, type=float)
+    a('-w', '--bigwig', nargs='+')
+    a('-n', '--bigwiglabels', nargs='+')
+    a('-o', '--bedouthilo', default=None, help="optional high and low combined bed")
+    a('-u', '--bedouthi', default=None, help="optional high only bed")
+    a('-b', '--bedoutlo', default=None, help="optional low only bed")
+    a('-t', '--tableoutfile', default=None)
+    args = parser.parse_args()
+    print('args=', args)
```
bgruening, line 179: remove?

tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
```diff
@@ -0,0 +1,285 @@
...
+    <param name="minwin" type="integer" value="10" label="Minimum continuous bases to count as a high or low bed feature" help="Actual run length will be found and used for continuous features as long or longer."/>
+    <param name="qhi" type="float" value="0.99" label="Quantile cutoff for a high region - 0.99 will cut off at or above the 99th percentile" help="" optional="true"/>
+    <param name="qlo" type="float" value="0.01" label="Quantile cutoff for a low region - 0.01 will cut off at or below the 1st percentile." help="" optional="true"/>
+    <param name="outhilo" type="boolean" truevalue="true" falsevalue="false"  checked="true" label="Output both high and low regions to a bed file" help="At least one up to 3 outputs. This one is best for JBrowse2"/>
+    <param name="outhi" type="boolean" truevalue="true" falsevalue="false" checked="false" label="Output only high regions to a bed file" help="Choose if you only care about high regions"/>
```
bgruening, line 195:
```suggestion
    <param argument="--bedouthilo" type="boolean" truevalue="--bedouthilo" falsevalue="" checked="false" label="Output only high regions to a bed file" help="Choose if you only care about high regions"/>
```

Why not using something like that?

tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
```diff
@@ -0,0 +1,285 @@
...
+      <filter>outhilo</filter>
+    </data>
+    <data name="bedouthi" format="bed" label="High bed" hidden="false">
+      <filter>outhi</filter>
+    </data>
+    <data name="bedoutlo" format="bed" label="Low bed" hidden="false">
+      <filter>outlo</filter>
+    </data>
+    <data name="tableoutfile" format="tabular" label="Contig statistics" hidden="false">
+      <filter>options['tableout'] == "create"</filter>
```
bgruening, line 214: Oha, why is "options" here needed?

tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
```diff
@@ -0,0 +1,285 @@
...
+        sys.exit(2)
+    findOut(args)
+#end raw]]></configfile>
+  </configfiles>
+  <inputs>
+    <param name="bigwig" type="data" optional="false" label="Bigwig file(s) to process" help="If more than one, MUST all use the same reference sequence to be displayable. Feature names will include the bigwig label." format="bigwig" multiple="true"/>
+    <param name="minwin" type="integer" value="10" label="Minimum continuous bases to count as a high or low bed feature" help="Actual run length will be found and used for continuous features as long or longer."/>
+    <param name="qhi" type="float" value="0.99" label="Quantile cutoff for a high region - 0.99 will cut off at or above the 99th percentile" help="" optional="true"/>
+    <param name="qlo" type="float" value="0.01" label="Quantile cutoff for a low region - 0.01 will cut off at or below the 1st percentile." help="" optional="true"/>
+    <param name="outhilo" type="boolean" truevalue="true" falsevalue="false"  checked="true" label="Output both high and low regions to a bed file" help="At least one up to 3 outputs. This one is best for JBrowse2"/>
```
bgruening, line 194: to prevent failing the tool when nothing is selected, could we use a 4-option select param with multiple=true but also optional=false?

tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
```diff
@@ -0,0 +1,285 @@
...
+ **What it Does**
+ 
+ Takes one or more bigwigs mapped to the same reference and finds all the minimum window sized or greater continuous regions above or below an upper and lower quantile cutoff.
+ A window size of 10 works well, and quantiles set at 0.01 and 0.99 will generally work well.
+
+ For JBrowse2 bed tracks, putting both high and low extreme regions in one bed file is usually best since they can now appear coloured red/blue.
+
+ If you only need to see either the consistently high or low regions, use one of those as the output instead.
+
+ At least one bed output must be selected or the tool will fail.
```
bgruening, line 278: with the 4-option select box this can not happen anymore, see above

tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
```diff
@@ -0,0 +1,285 @@
+<tool name="bigwig_outlier_bed" id="bigwig_outlier_bed" version="0.01" profile="22.05">
```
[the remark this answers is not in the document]
bgruening: can we give it a better name than "bigwig_outlier_bed". Removing the `_`, maybe "Filter bigwig outliers"? or "extremes"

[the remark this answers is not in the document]
fubar2: yes of course. Should say it makes beds. Should say it uses bigwigs. Something about consistently extreme segments. Filter is apt but redundant maybe.
How about `Bigwig extremes to bed features` ?

tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
fubar2: yes!

tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
fubar2: funny you should suggest that. I even have a branch - noargparse. 
Oddly this works more conveniently and I never got the noargparse version to pass tests so gave up.

tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
fubar2: sure

tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
fubar2: Because that's what the docs say works - and it does :)

tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
fubar2: Sure.

tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
fubar2: Good idea. Will do shortly.

tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
fubar2: Style and taste. _De gustibus non disputandem est_ but of course.

tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
fubar2: yes. will do.

tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
fubar2: ah - no - that's for conditionals. Fixed!

fubar2:
@bgruening: Name ok?  Do not have energy to figure out why the argparse-less version would never pass tests.

fubar2:
It passes tests locally with the samples in the repo, but fails in CI. 
Trying biocontainers test - forgot that's what CI uses...

fubar2:
Strange. Passes here using biocontainers too.
No idea what the problem is.

bgruening:
grafik


You can see the error report here:

https://github.com/galaxyproject/tools-iuc/actions/runs/9776636408?pr=6121

tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
```diff
@@ -0,0 +1,345 @@
...
+        self.bedouthilo=args.bedouthilo
+        self.tableoutfile = args.tableoutfile
+        self.bedwin = args.minwin
+        self.qhi = args.qhi
+        self.qlo = args.qlo
+        nbw = len(args.bigwig)
+        nlab = len(args.bigwiglabels)
+        if nlab < nbw:
+          self.bwlabels += ['Nolabel']*(nbw-nlab)
+        print("bwlabels=", self.bwlabels, "bw=", self.bwnames)
```
bgruening, line 93: Do you want to keep all those prints?

fubar2: no - thanks!

tools/bigwig_outlier_bed/bigwig_outlier_bed.xml
mvdbeek: Can we instead place this as a normal script next to the tool ?
The big advantage of that is that we run through all the normal code linting and we get syntax highlighting and all on github.

fubar2: @mvdbeek: Good idea, thanks! 
 
- Separated the Python script out to a file; 
- Added a `required_files` entry for Pulsar operation.

tools/bigwig_outlier_bed/.shed.yml
```diff
@@ -0,0 +1,20 @@
+categories:
+- Sequence Analysis
+description: pybigtools and numpy code to find continuous runs above a high or below a low quantile cutpoint in bigwig files
+homepage_url: https://github.com/fubar2/bigwig_outlier_bed
```
bgruening, line 4:
```suggestion
homepage_url: https://github.com/galaxyproject/tools-iuc/tree/master/tools/bigwig_outlier_bed
```

This should point to https://github.com/jackh726/bigtools but this wrapper is not really for https://github.com/jackh726/bigtools ... its just using it. So we use the IUC URL since the code lives here.

tools/bigwig_outlier_bed/.shed.yml
```diff
@@ -0,0 +1,20 @@
...
+  low quantile.
+
+  Quantile values recommended are 0.01 and 0.99. They are calculated for each chromosome and vary a bit.
+  Ideally should be estimated over the entire assembly but not feasible without sampling due to RAM hoggery.
+  Minimum window size of 10 will give a very, very low risk of random false positives at about 0.01**10 for 0.01
+  quantile cutoff for example.
+
+name: bigwig_outlier_bed
+owner: iuc
+remote_repository_url: https://github.com/jackh726/bigtools
```
bgruening, line 19:
```suggestion
remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/master/tools/bigwig_outlier_bed
```

This should point to the repo of the wrapper.

bgruening (approved):

PR #6123, opened 2024-07-03
Quick fix for fastq-dump reads name issue

diegomics:
In **fastq-dump**, the current default for --defline-seq is `@$sn[_$rn]/$ri`
The program default is `@$ac.$si $ri length=$rl`
Maybe the program default is too much, but missing the accession can produce reads named 1/1, 2/1, 3/1 in some cases. This is problematic if you have several fastq with reads named the same.
I'm proposing bringing back the accession at the beginning to avoid this. It should be `@$ac.$sn[_$rn]/$ri`

Similarly, for **fasterq-dump**, now it is `@$sn/$ri` and it should be `@ac.$sn/$ri`

### DIFF
--- a/tools/sra-tools/fasterq_dump.xml
+++ b/tools/sra-tools/fasterq_dump.xml
@@ -73,7 +73,7 @@
     <inputs>
         <expand macro="input_conditional"/>
         <section name="adv" title="Advanced Options" expanded="False">
-            <expand macro="defline" defline_param="--seq-defline" defline_default="@$sn/$ri"/>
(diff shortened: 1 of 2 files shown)


SaimMomin12:
Hi @diegomics, 
Thanks for bringing this into our notice, could you also update the test data files for the tool which will allows the above the CI/CD tests to pass?

SaimMomin12:
Hi @diegomics, I have updated the test dataset for this PR [here.](https://github.com/galaxyproject/tools-iuc/pull/6260)

Thanks for the contribution!

PR #6161, opened 2024-07-17
cutadapt: replace adapter sanitizer by validator

bernt-matthias:
I had a user who used spaces in the adapter (for better visual orientation). these were silently replaced by X.

I think replacing the sanitizer by a validator is better, because more explicit behavior.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/cutadapt/macros.xml
+++ b/tools/cutadapt/macros.xml
@@ -38,6 +38,7 @@
         #end if
     </token>
     <xml name="adapter_sanitizer">
+        <sanitizer sanitize="false" />
         <validator type="regex"><![CDATA[[0-9ATCGURYSWKMBDHVNX{}^$.()]+$]]></validator>
     </xml>
     <xml name="adapter_conditional" tokens="adapter_type,argument">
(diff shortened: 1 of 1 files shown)


tools/cutadapt/macros.xml
```diff
@@ -38,34 +38,7 @@
...
-                <add value="{"/><!--specify repeats .. needs digits which are added as default-->
-                <add value="}"/>
-                <add value="^"/><!--anchoring-->
-                <add value="$"/>
-                <add value="."/>
-                <add value="("/><!--specify what to keep-->
-                <add value=")"/>
-            </valid>
-        </sanitizer>
+        <validator type="regex"><![CDATA[[0-9ATCGURYSWKMBDHVNX{}^$.()]+$]]></validator>
```
wm75, line 41: The default valid characters are string.ascii_letters + string.digits + `" -=_.()/+*^,:?!"` so I guess you'll have to say at least `<sanitizer sanitize="false" />`.

bernt-matthias: True. Good catch. Thanks

wm75 (approved):
Looks very reasonable to me. Thanks!

PR #6113, opened 2024-06-28
Scanpy plot new function

heylf:
FOR CONTRIBUTOR:
* [X] I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [x] License permits unrestricted use (educational + commercial)
* [ ] This PR adds a new tool or tool collection
* [X] This PR updates an existing tool or tool collection
* [ ] This PR does something else (explain below)

PR #6162, opened 2024-07-19
Update Scanpy to 1.10, preparing for GTA 

pavanvidem:
FOR CONTRIBUTOR:


bgruening:
@pavanvidem not sure if you have seen it, but we also need a rebase here.

nilchia:
mnnpy conda recipe needs update

bgruening:
Please rebase this PR

bgruening:
What do you mean with mnnoy update? What is broken?

nilchia:
All tests pass now :D
I'll rebase the PR and resolve the conflicts next.

nilchia:
I removed `sc.pl.dpt_groups_pseudotime()` because it has problem in upstream source code.
It is incompatible with `pandas` version I think. The way `scanpy` tries to index a `pandas.series` object is not supported by `pandas`.
```
  File "<stdin>", line 1, in <module>
  File "/home/nilchia/miniconda3/envs/scanpy/lib/python3.12/site-packages/legacy_api_wrap/__init__.py", line 80, in fn_compatible
    return fn(*args_all, **kw)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/nilchia/miniconda3/envs/scanpy/lib/python3.12/site-packages/scanpy/plotting/_tools/__init__.py", line 276, in dpt_groups_pseudotime
    timeseries_subplot(
  File "/home/nilchia/miniconda3/envs/scanpy/lib/python3.12/site-packages/scanpy/plotting/_utils.py", line 140, in timeseries_subplot
    X = X[:, None]
        ~^^^^^^^^^
  File "/home/nilchia/miniconda3/envs/scanpy/lib/python3.12/site-packages/pandas/core/series.py", line 1153, in __getitem__
    return self._get_with(key)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/nilchia/miniconda3/envs/scanpy/lib/python3.12/site-packages/pandas/core/series.py", line 1163, in _get_with
    return self._get_values_tuple(key)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/nilchia/miniconda3/envs/scanpy/lib/python3.12/site-packages/pandas/core/series.py", line 1203, in _get_values_tuple
    disallow_ndim_indexing(result)
  File "/home/nilchia/miniconda3/envs/scanpy/lib/python3.12/site-packages/pandas/core/indexers/utils.py", line 341, in disallow_ndim_indexing
    raise ValueError(
ValueError: Multi-dimensional indexing (e.g. obj[:, None]) is no longer supported. Convert to a numpy array before indexing instead.
>>> 
```

bgruening:
Try using numpy <2.0. Maybe that helps.

nilchia:
my package versions are these:
```
numpy                     1.26.4          py312heda63a1_0    conda-forge
pandas                    2.2.2           py312h1d6d2e6_1    conda-forge
scanpy                    1.10.2             pyhd8ed1ab_0    conda-forge
```

nilchia:
I did rebase and resolved conflicts locally but still, here it said there was a conflict :(

I reset back to my last commit.

nilchia:
The tests added from #6113 need a proper input file I guess.
error:
```
KeyError: "Could not find keys '['C1orf54', 'CLEC9A', 'CPNE3', 'DNASE1L3', 'HSP90B1', 'ID2', 'IGLL1', 'IRF8', 'LGALS1', 'NASP', 'PCNA', 'PPIB', 'RNASE6', 'SEC11C', 'SNX3', 'TK1']' in columns of `adata.obs` or in adata.var_names."
```

tools/scanpy/cluster_reduce_dimension.xml
```diff
@@ -616,16 +569,37 @@ sc.tl.embedding_density(
...
+            </section>
+            <output name="hidden_output">
+                <assert_contents>
+                    <has_text_matching expression="sc.tl.paga"/>
+                    <has_text_matching expression="groups='paul15_clusters'"/>
+                    <has_text_matching expression="use_rna_velocity=False"/>
+                    <has_text_matching expression="model='v1.2'"/>
+                </assert_contents>
+            </output>
+            <output name="anndata_out" file="tl.paga.neighbors_gauss_braycurtis.recipe_weinreb17.paul15_subsample.h5ad" ftype="h5ad" compare="sim_size"/>
```
bgruening, line 591: fyi: h5ad is an hdf5 file and for those you can use asserts: https://docs.galaxyproject.org/en/latest/dev/schema.html#assertions-for-h5-data

So IMHO it would make much more sense to assert the content of h5 files everywhere.

tools/scanpy/filter.xml
```diff
@@ -505,32 +579,55 @@ df.to_csv('marker.tsv', sep='\t', index=True)
...
+            </section>
+            <output name="hidden_output">
+                <assert_contents>
+                    <has_text_matching expression="sc.pp.scrublet"/>
+                    <has_text_matching expression="sim_doublet_ratio=2.0"/>
+                    <has_text_matching expression="expected_doublet_rate=0.05"/>
+                    <has_text_matching expression="n_prin_comps=5"/>
+                </assert_contents>
+            </output>
+            <output name="anndata_out" file="pp.scrublet.krumsiek11.h5ad" ftype="h5ad"/>
```
bgruening, line 629: use asserts here?

tools/scanpy/filter.xml
```diff
@@ -267,9 +330,30 @@ df.to_csv('marker.tsv', sep='\t', index=True)
...
-                <param argument="groupby" type="text" value="" label="The key of the observation grouping to consider (e.g., celltype)" help="">
+                <param argument="groupby" type="text" value="" label="The key of the observation grouping to consider (e.g., celltype)">
+                    <expand macro="sanitize_query"/>
+                </param>
+            </when>
+            <when value="pp.scrublet">
+                <param argument="batch_key" type="text" value="" optional="true" label="Batch key for the concatenate">
                     <expand macro="sanitize_query" />
                 </param>
+                <param argument="sim_doublet_ratio" type="float" value="2.0" label="Number of doublets to simulate relative to the number of observed transcriptomes"/>
```
bgruening, line 341: do you think you can add max/min to a lot of those floats and ints? Is that feasible?

tools/scanpy/filter.xml
```diff
@@ -267,9 +330,30 @@ df.to_csv('marker.tsv', sep='\t', index=True)
...
+                <param name="knn_dist_metric" type="select" label="Distance metric used when finding nearest neighbors">
+                    <expand macro="distance_metric_options"/>
+                </param>
+                <param argument="normalize_variance" type="boolean" truevalue="True" falsevalue="False" checked="true" label="normalize the data such that each gene has a variance of 1"/>
+                <param argument="log_transform" type="boolean" truevalue="True" falsevalue="False" checked="false" label="Whether to use log1p() to log-transform the data prior to PCA"/>
+                <param argument="mean_center" type="boolean" truevalue="True" falsevalue="False" checked="true" label="If True, center the data such that each gene has a mean of 0"/>
+                <param argument="n_prin_comps" type="integer" value="30" label="Number of principal components used to embed the transcriptomes prior to k-nearest-neighbor graph construction"/>
+                <param argument="use_approx_neighbors" type="boolean" truevalue="True" falsevalue="None" checked="false" label="Use approximate nearest neighbor method (annoy) for the KNN classifier"/>
```
bgruening, line 352: True/None is here correct? And not True/False

tools/scanpy/inspect.xml
```diff
@@ -1,258 +1,298 @@
...
-#else if $method.method == "tl.marker_gene_overlap"
+#else if str($method.method) == "tl.marker_gene_overlap":
 reference_markers = {}
 #for $i, $s in enumerate($method.reference_markers)
     #set $list=[x.strip() for x in str($s.values).split(',')]
 reference_markers['$s.key'] = $list
 #end for
 
-sc.tl.marker_gene_overlap(
+# Temporary fix for the issue with writing Pandas dataframes to h5ad
```
bgruening, line 231: which issues? Do you have a link? When can we remove it?

tools/scanpy/inspect.xml
```diff
@@ -495,15 +554,35 @@ sc.pp.sqrt(
...
+            </section>
+            <output name="hidden_output">
+                <assert_contents>
+                    <has_text_matching expression="sc.pp.calculate_qc_metrics"/>
+                    <has_text_matching expression="expr_type='counts'"/>
+                    <has_text_matching expression="var_type='genes'"/>
+                    <has_text_matching expression="qc_vars=\['mito', 'negative'\]"/>
+                </assert_contents>
+            </output>
+            <output name="anndata_out" file="pp.calculate_qc_metrics.sparce_csr_matrix.h5ad" ftype="h5ad" compare="sim_size"/>
```
bgruening, line 576: asserts?

tools/scanpy/inspect.xml
```diff
@@ -766,47 +817,51 @@ sc.pp.sqrt(
...
-                <has_text_matching expression="key='rank_genes_groups'"/>
-                <has_text_matching expression="method='overlap_count'"/>
+                <has_text_matching expression="marker_gene_overlap"/>
             </assert_stdout>
+            <output name="hidden_output">
+                <assert_contents>
+                    <has_text_matching expression="sc.tl.marker_gene_overlap"/>
+                </assert_contents>
+            </output>
             <output name="anndata_out" file="tl.marker_gene_overlap.pbmc68k_reduced.h5ad" ftype="h5ad" compare="sim_size"/>
```
bgruening, line 854: Asserts?

tools/scanpy/macros.xml
```diff
@@ -107,13 +85,13 @@ with open('anndata_info.txt','w', encoding='utf-8') as ainfo:
...
     </xml>
     <xml name="numeric_distance_metric_options">
+    	<option value="euclidean">euclidean</option>
         <option value="braycurtis">braycurtis</option>
         <option value="canberra">canberra</option>
         <option value="chebyshev">chebyshev</option>
         <option value="cityblock">cityblock</option>
         <option value="correlation">correlation</option>
         <option value="cosine">cosine</option>
-        <option value="euclidean" selected="true">euclidean</option>
```
bgruening, line 116: I like to have the selected=true - the explicit form is better imho.

tools/scanpy/macros.xml
```diff
@@ -679,166 +522,505 @@ header=None
...
+    <token name="@CMD_PRETTIFY_STDOUT@"><![CDATA[
+| sed -r '1 s|AnnData object with (.+) = (.*)\s*|\1: \2|g' | sed "s|'||g"  | sed -r 's|^\s*(.*):\s(.*)|[\1]\n-    \2|g' | sed 's|, |\n-    |g'
+    ]]>
+    </token>
+    <token name="@CMD_READ_INPUTS@"><![CDATA[
+adata = sc.read_h5ad('anndata.h5ad')
+    ]]>
+    </token>
+    <token name="@CMD@"><![CDATA[
+cp '$adata' 'anndata.h5ad' &&
```
bgruening, line 561: is a cp still needed or can we use ln -s here.

tools/scanpy/macros.xml
```diff
@@ -679,166 +522,505 @@ header=None
...
+with open('anndata_info.txt','w', encoding='utf-8') as ainfo:
+    print(adata, file=ainfo)
+    ]]>
+    </token>
+    <token name="@CMD_PARAM_PLOT_INPUTS@"><![CDATA[
+    adata,
+    save='.$format',
+    ]]>
+    </token>
+    <token name="@CMD_SHOW@"><![CDATA[
```
bgruening, line 586: The token is call CMD_SHOW but its show=None? That is funny :)

tools/scanpy/macros.xml
```diff
@@ -679,166 +522,505 @@ header=None
...
+    ]]>
+    </token>
+    <!-- command macros -->
+
+
+
+    <!-- param and cmd macros -->
+    <xml name="param_standard_scale">
+        <param argument="standard_scale" type="select" label="Standardize a dimension between 0 and 1" help="Each variable or observation is subtracted by the minimum and divided each by its maximum.">
+            <option value="None">No standardization</option>
```
bgruening, line 613: selected=true

tools/scanpy/macros.xml
```diff
@@ -679,166 +522,505 @@ header=None
...
+            </param>
+            <when value="all"/>
+            <when value="custom">
+                <param argument="var_names" type="text" value="" label="List of variables to plot" help="They should be a valid subset of 'adata.var_names', and separated by comma">
+                    <expand macro="sanitize_query"/>
+                </param>
+            </when>
+            <when value="customfile">
+                <param argument="var_names" type="data" format="tabular" label="List of variables to plot" help="This should be a tsv where row = group (e.g. celltypes) and columns = variables."></param>
+                <param name="header" type="select" label="Header in the list of markers?">
```
bgruening, line 664: I guess this could be boolean?

tools/scanpy/macros.xml
```diff
@@ -679,166 +522,505 @@ header=None
...
+    </xml>
+    <xml name="param_vmin">
+        <param argument="vmin" type="float" value="" optional="true" label="Minimum value to normalize luminance data" help="If not set, it is inferred from the data and other keyword arguments"/>
+    </xml>
+    <xml name="param_vmax">
+        <param argument="vmax" type="float" value="" optional="true" label="Maximum value to normalize luminance data" help="If not set, it is inferred from the data and other keyword arguments"/>
+    </xml>
+    <xml name="section_matplotlib_pyplot_scatter">
+        <section name="matplotlib_pyplot_scatter" title="Parameters for matplotlib.pyplot.scatter">
+            <!--<param argument="marker" type="select" label="Marker style">
```
bgruening, line 784: is this still commented out? Why?

tools/scanpy/macros.xml
```diff
@@ -847,144 +1029,105 @@ header=None
...
-    <token name="@CMD_param_legend_fontsize@"><![CDATA[
-    #if str($method.plot.legend_fontsize) != ''
-    legend_fontsize=$method.plot.legend_fontsize,
-    #end if
-    ]]>
-    </token>
     <xml name="param_legend_fontweight">
-        <param argument="legend_fontweight" type="select" label="Legend font weight" help="">
+        <param argument="legend_fontweight" type="select" label="Legend font weight">
+            <option value="bold">bold</option>
```
bgruening, line 1091: selected=true

tools/scanpy/macros.xml
```diff
@@ -679,166 +522,505 @@ header=None
...
+        <section name="matplotlib_pyplot_pcolor" title="Parameters for matplotlib.pyplot.pcolor">
+            <param argument="cmap" type="select" optional="true" label="Color palette">
+                <expand macro="seaborn_color_palette_options"/>
+            </param>
+            <expand macro="param_matplotlib_pyplot_edgecolors"/>
+            <expand macro="param_alpha"/>
+            <param argument="snap" type="boolean" truevalue="True" falsevalue="False" checked="false" label="Snap the mesh to pixel boundaries?"/>
+        </section>
+        <expand macro="param_standard_scale"/>
+        <param argument="vmin" type="float" value="" optional="true" label="The value representing the lower limit of the color scale" help="Values smaller than vmin are plotted with the same color as vmin."/>
```
bgruening, line 891: Is that one different to https://github.com/galaxyproject/tools-iuc/pull/6162/files#diff-b2ced7d32e78ae8f951fa61b63504b1dabd42806aeaefdb1ea7f1298ae1c375fR776 the other macro?

tools/scanpy/macros.xml
```diff
@@ -997,303 +1140,336 @@ header=None
...
+    min_logfoldchange=$method.min_logfoldchange,
     #end if
-    snap=$method.matplotlib_pyplot_pcolor.snap
     ]]>
     </token>
-    <xml name="pl_stacked_violin">
+
+    <xml name="param_density_norm">
+        <param argument="density_norm" type="select" label="Method used to scale the width of each violin">
+        	<option value="area">area: each violin will have the same area</option>
```
bgruening, line 1403: indentaion

tools/scanpy/macros.xml
```diff
@@ -997,303 +1140,336 @@ header=None
...
-            <option value="var">Standardization on variable</option>
-            <option value="obs">Standardization on observation</option>
-        </param>
-        <expand macro="seaborn_violinplot"/>
-    </xml>
-    <token name="@CMD_pl_stacked_violin@"><![CDATA[
+        <expand macro="param_standard_scale"/>
+        <expand macro="params_seaborn_violinplot"/>
+        <param argument="yticklabels" type="boolean" truevalue="True" falsevalue="False" checked="false" label="Set to view the y tick labels"/>
+        <param argument="vmin" type="float" value="" optional="true" label="The value representing the lower limit of the color scale" help="Values smaller than vmin are plotted with the same color as vmin."/>
```
bgruening, line 1421: Is that the same as: 
https://github.com/galaxyproject/tools-iuc/pull/6162/files#diff-b2ced7d32e78ae8f951fa61b63504b1dabd42806aeaefdb1ea7f1298ae1c375fR776

tools/scanpy/plot.xml
```diff
@@ -486,57 +527,58 @@ sc.pl.rank_genes_groups_stacked_violin(
                 <option value="pl.diffmap">Embeddings: Scatter plot in Diffusion Map basis, using 'pl.diffmap'</option>
                 <option value="pl.draw_graph">Embeddings: Scatter plot in graph-drawing basis, using 'pl.draw_graph'</option>
                 <option value="pl.embedding_density">Embeddings: Plot the density of cells in an embedding (per condition), using 'pl.embedding_density'</option>
-                <option value="pl.dpt_groups_pseudotime">Branching trajectories and pseudotime, clustering: Plot groups and pseudotime, using 'pl.dpt_groups_pseudotime'</option>
+                <!-- <option value="pl.dpt_groups_pseudotime">Branching trajectories and pseudotime, clustering: Plot groups and pseudotime, using 'pl.dpt_groups_pseudotime'</option> -->
```
bgruening, line 530: is this still out? please add a comment why?

tools/scanpy/plot.xml
```diff
@@ -939,91 +1034,113 @@ sc.pl.rank_genes_groups_stacked_violin(
...
+                    <has_text_matching expression="sort_order=True"/>
+                    <has_text_matching expression="projection='2d'"/>
+                    <has_text_matching expression="legend_fontsize=1"/>
+                    <has_text_matching expression="legend_fontweight='normal'"/>
+                    <has_text_matching expression="frameon=True"/>
+                    <has_text_matching expression="size=1.0"/>
+                    <has_text_matching expression="title='A title'"/>
+                </assert_contents>
+            </output>
+            <output name="out_png" file="pl.scatter.pbmc68k_reduced.png" ftype="png" compare="sim_size"/>
```
bgruening, line 1113: there are nowadays proper image-assertions ... so you don't need to compare the size.

tools/scanpy/plot.xml
```diff
@@ -1110,337 +1221,294 @@ sc.pl.rank_genes_groups_stacked_violin(
...
+                  <has_text_matching expression="use_raw=None"/>
+                  <has_text_matching expression="num_categories=7"/>
+                  <has_text_matching expression="dendrogram=True"/>
+              </assert_contents>
+          </output>
+        </test>
+        
+        <!-- test 8 -->
+        <test expect_num_outputs="2">
+            <param name="adata" value="pp.filter_genes_dispersion.krumsiek11-seurat.h5ad"/>
```
bgruening, line 1288: indentation

tools/scanpy/plot.xml
```diff
@@ -1110,337 +1221,294 @@ sc.pl.rank_genes_groups_stacked_violin(
...
                   <has_text_matching expression="groupby='cell_type'"/>
                   <has_text_matching expression="log=False"/>
-                  <has_text_matching expression="use_raw=False"/>
+                  <has_text_matching expression="use_raw=None"/>
                   <has_text_matching expression="num_categories=7"/>
                   <has_text_matching expression="dendrogram=True"/>
                   <has_text_matching expression="swap_axes=True"/>
               </assert_contents>
           </output>
+          <output name="out_png" file="pl.stacked_violin_pp.filter_genes_dispersion.krumsiek11-seurat.png" compare="sim_size"/>
```
bgruening, line 1424: assertions ?

tools/scanpy/remove_confounders.xml
```diff
@@ -1,89 +1,183 @@
...
    #set $keys = [str(x.strip()) for x in str($method.keys).split(',')]
    keys=$keys,
+   n_jobs = int(os.getenv("GALAXY_SLOTS", 4)),
    copy=False)
 
-#else if $method.method == "pp.mnn_correct"
-    #for i, filepath in enumerate($methods.extra_adata)
-adata_$i = ad.read('$filepath')
-    #end for
+## #else if str($method.method) == 'external.pp.mnn_correct':
```
bgruening, line 33: please add a comment why this is not included and when should it be included again ...

tools/scanpy/remove_confounders.xml
```diff
@@ -102,96 +196,287 @@ sc.pp.combat(
...
             </output>
-            <output name="anndata_out" file="pp.regress_out.krumsiek11.h5ad" ftype="h5ad" compare="sim_size"/>
+            <output name="anndata_out" file="pp.regress_out.batch.h5ad" ftype="h5ad" compare="sim_size"/>
         </test>
-        <!--<test expect_num_outputs="2">
-            < test 2 >
-            <param name="adata" value="krumsiek11.h5ad" />
+
+        <!-- test 2 -->
+        <!-- <test expect_num_outputs="2">
```
bgruening, line 311: Why is this test out, please add a comment

tools/scanpy/remove_confounders.xml
```diff
@@ -102,96 +196,287 @@ sc.pp.combat(
...
+                    <has_text_matching expression="batch_key='batch'"/>
+                    <has_text_matching expression="use_rep='X_pca'"/>
+                    <has_text_matching expression="use_annoy=True"/>
+                    <has_text_matching expression="annoy_n_trees=10"/>
+                    <has_text_matching expression="neighbors_within_batch=3"/>
+                    <has_text_matching expression="n_pcs=10"/>
+                    <has_text_matching expression="set_op_mix_ratio=0.5"/>
+                </assert_contents>
+            </output>
+            <output name="anndata_out" file="external.pp.bbknn_annoy.krumsiek11.h5ad" ftype="h5ad" compare="sim_size"/>
```
bgruening, line 374: asserts

tools/scanpy/cluster_reduce_dimension.xml
nilchia: Sure, Will do

tools/scanpy/filter.xml
nilchia: There is no information about that in the documentation.
Even for the plotting arguments that are passed to matplotlib or seaborn, I did not find min/max values

tools/scanpy/filter.xml
nilchia: Yes. For example [here](https://github.com/scverse/scanpy/blob/b6193502e11b84fc1b4a011ee9cf08a19da22ebf/src/scanpy/preprocessing/_scrublet/__init__.py#L55C5-L55C46):
use_approx_neighbors: bool | None = None,

And also in the documentation the default is None

tools/scanpy/inspect.xml
nilchia: It is in the source [code](https://github.com/scverse/scanpy/blob/b6193502e11b84fc1b4a011ee9cf08a19da22ebf/src/scanpy/tools/_marker_gene_overlap.py#L167).

It is related to pandas.
It seems they are going to [deprecate](https://github.com/pandas-dev/pandas/issues/16529) `inplace=True` for pandas.

tools/scanpy/macros.xml
nilchia: Isn't the first option selected as default?
I thought maybe it would be easier to maintain that way.

tools/scanpy/macros.xml
nilchia: That should work. I'll change it.

tools/scanpy/macros.xml
nilchia: It is from matplotlib and I think It is not used. Usually default "o" is used.

But I can add it. No problem,

@pavanvidem Is there any other reason you commented out this param?

tools/scanpy/macros.xml
nilchia: No :)
I'll fix it. thanks.

tools/scanpy/inspect.xml
bgruening: Please link this into your comments, so future contributors knows what it is about.

tools/scanpy/macros.xml
bgruening: everything that is commented out should have a comment in such complex tools, or remove it. Future contributors might not know the reason things have been commented out.

tools/scanpy/cluster_reduce_dimension.xml
nilchia: This assert has the same problem as the anndata metadata.
for example:
```
            <output name="anndata_out" ftype="h5ad">
                <assert_contents>
                    <has_h5_keys keys="X,layers,obs,obsm,obsp,uns,var,varm,varp"/>
                    <has_h5_attribute key="obs" value="louvain"/>
                    <has_h5_attribute key="uns" value="louvain"/>
                </assert_contents>
            </output>
```
fails with this error:
```
Not a HDF5 file or H5 attributes do not match:
	[('encoding-type', 'anndata'), ('encoding-version', '0.1.0')]

	(obs : louvain)
```

nilchia: `has_h5_keys` works fine and I can assert keys but `has_h5_attribute` does not work for the attributes.

tools/scanpy/macros.xml
nilchia: @pavanvidem I have removed this. Please let me know if you think this is needed.

tools/scanpy/macros.xml
nilchia: Yeah! Right.
Thanks!

tools/scanpy/cluster_reduce_dimension.xml
nilchia: I wrote an issue #18767

tools/scanpy/macros.xml
nilchia: That does not work because the output is overwritten to it.
With Simlink it gave this error:
`OSError: [Errno 30] Unable to synchronously create file (unable to open file: name = 'anndata.h5ad', errno = 30, error message = 'Read-only file system', flags = 13, o_flags = 242)`

bgruening: Add a comment please.

tools/scanpy/macros.xml
pavanvidem: I don't think people ever use it. The default works just fine. It's fine to remove.

nilchia:
Please don't merge. I'll add h5 assert.

tools/scanpy/macros.xml
```diff
@@ -679,166 +522,491 @@ header=None
...
+    ]]>
+    </token>
+    <token name="@CMD_IMPORTS@"><![CDATA[
+import scanpy as sc
+import pandas as pd
+import numpy as np
+    ]]>
+    </token>
+    <token name="@CMD_ANNDATA_WRITE_OUTPUTS@"><![CDATA[
+adata.write_h5ad('anndata.h5ad')
```
pavanvidem, line 577:
```suggestion
adata.write_h5ad('anndata.h5ad', compression='gzip')
```
Please replace in all places where we save anndata to disk. It saves disk space and does not hurt while reading.

nilchia:
I added the h5 assert.
However, for some functions, this assertion is not a good test because the function does not add any layer or key to the h5ad file but only changes the contents of the keys.
Functions:
 `pp.downsample_counts`
`filter_marker`
`pp.sqrt`
`external.pp.magic` with t=-1
`pp.regress_out`
`pp.combat`

nilchia:
in filter.xml, the real output of `filter_marker` is the tabular file and not the adata, right?

nilchia:
`.. ERROR: Error '403 Client Error: Forbidden for url: https://www.biorxiv.org/content/10.1101/298430v1' accessing https://www.biorxiv.org/content/10.1101/298430v1`

What is wrong with the link? It was the same error with the previous link.

bgruening:
No idea, but the preprint is published by now and you should link the final paper https://www.nature.com/articles/nbt.4314

pavanvidem:
> in filter.xml, the real output of `filter_marker` is the tabular file and not the adata, right?

Yes, it is is. Good catch! It also updates the anndata but tabular file is the actual output to test. Thanks!

pavanvidem:
> I added the h5 assert. However, for some functions, this assertion is not a good test because the function does not add any layer or key to the h5ad file but only changes the contents of the keys. Functions: `pp.downsample_counts` `filter_marker` `pp.sqrt` `external.pp.magic` with t=-1 `pp.regress_out` `pp.combat`

Yes, all those functions only update the count matrix, not the annotations. If we don't want to save the test data then check whether .X exists.

nilchia:
I added test for those. Is it good?

tools/scanpy/macros.xml
```diff
@@ -679,166 +524,491 @@ header=None
...
+    <token name="@CMD_IMPORTS@"><![CDATA[
+import scanpy as sc
+import pandas as pd
+import numpy as np
+    ]]>
+    </token>
+    <token name="@CMD_ANNDATA_WRITE_OUTPUTS@"><![CDATA[
+adata.write_h5ad('anndata.h5ad', compression='gzip')
+with open('anndata_info.txt','w', encoding='utf-8') as ainfo:
+    print(adata, file=ainfo)
```
pavanvidem, line 581: Here we are writing anndata info to a file and then to stdout. So for all tests where there are no changes in the annotations, you can also check the matrix size in stdout. Something like:
```
<assert_stdout>
    <has_text_matching expression="336 × 11"/>
</assert_stdout>
```

nilchia: Do you mean instead of the `min() max() ...` ?
But for those functions the matrix size does not change.
Only the values of adata.X is changed.

I used this assert for `filter_cells/genes` and `downsample` that this data size changes

nilchia:
@pavanvidem @bgruening 
Should I add/modify anything here? or is it ready to merge?

pavanvidem:
Can you please check if this was fixed? https://github.com/galaxyproject/tools-iuc/issues/6302

Otherwise, good to merge.

nilchia:
Actually, I didn't understand what was the issue there. Did the config file have a problem? or the adata was not defined?

bgruening:
`NameError: name 'adata' is not defined`, can you check if you still have this variable somewhere and if you have a test for this?

So many things have changed, I guess we can proceed.

nilchia:
Almost all functions use this variable. I guess we don't have any problems with that.

pavanvidem:
Agree. Let's proceed. The error was in the plot tool. None of the functions change the anndata object. So it should not affect the analysis. The worst thing that can happen is that no plot is generated which we can investigate later.

bgruening (approved):

bgruening:
Awesome!

pavanvidem:
Thanks, @nilchia and @bgruening for helping out!!

PR #5793, opened 2024-02-26
Updating tools/metaphlan from version 4.0.6 to 4.1.1

gxydevbot:
Hello! This is an automated update of the following tool: **tools/metaphlan**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

I have updated tools/metaphlan from version 4.0.6 to 4.1.0.

**Project home page:** https://github.com/biobakery/MetaPhlAn/releases

For any comments, queries or criticism about the bot, not related to the tool being updated in this PR, please create an issue [here](https://github.com/planemo-autoupdate/autoupdate/issues/new).

PR #6183, opened 2024-07-27
Bump metaphlan

bernt-matthias:
replaces https://github.com/galaxyproject/tools-iuc/pull/5793

I'm not entirely sure if we should implement the subsampling parameters: 

- the make the tool form more complex/error prone (in the current we do not match the paired parameters with paired inputs, and they are only allowed for fastqinput)
- maybe there are better external tools for the job, but
  - then we have the intermediate output
  - and we also can not (or its difficult) to subsample the mapping results (maybe with 2 metaphlan runs)

FOR CONTRIBUTOR:


tools/metaphlan/metaphlan.xml
```diff
@@ -118,7 +118,7 @@ python '$__tool_directory__/customizemetadata.py'
 
 metaphlan
 #if $inputs.in.selector == "raw"
-    '$file_path'
+    $file_path
```
bgruening, line 121: are you sure about that one?

tools/metaphlan/metaphlan.xml
```diff
@@ -351,6 +399,8 @@ python '$__tool_directory__/formatoutput.py'
                 label="Scale relative abundances to the number of reads mapping to known clades in order to estimate unknowness?"/>
             <param name="krona_output" type='boolean' checked="false" truevalue='true' falsevalue='false' label="Output for Krona?"/>
         </section>
+        <!-- enabling this is tests will allow metaphlan to download reference data (we do this only with the smallish TOY DB) -->
```
bgruening, line 402:
```suggestion
        <!-- enabling this in tests will allow metaphlan to download reference data (we do this only with the smallish TOY DB) -->
```

tools/metaphlan/test-data/test_database_versioned.loc
bgruening, line 1: I guess those files should be called .test by conversion.

bernt-matthias: > I guess those files should be called .test by conversion.

I guess you mean https://github.com/galaxyproject/tools-iuc/blob/5ce2989223313aaa8dccb4fb2be7f4cce88a70d0/tools/metaphlan/tool_data_table_conf.xml.test? Can't remember that this is none for the loc files. But I also don't care :)

bgruening: https://github.com/search?q=repo%3Agalaxyproject%2Ftools-iuc%20loc.test&type=code

Its also better to grep or distinguish on the hard drive.

tools/metaphlan/metaphlan.xml
bgruening: I see why you are doing it ... its a bit hard to read, but ok.

This is missing? https://github.com/galaxyproject/tools-iuc/pull/6183/files#diff-8cacfb8eaa2b96a3a171b27052b6f317fd58f37ab1dc0e9d5f3e8d2d8550304eR48

bernt-matthias:
Ready from my side. The only thing that I still do not like here is the `multiple` option in the raw_in selector. But I'm hesitant to change the conditional.

bgruening (approved):

tools/metaphlan/metaphlan.xml
```diff
@@ -854,16 +1046,18 @@ python '$__tool_directory__/formatoutput.py'
...
                 </assert_contents>
             </output>
-        </test>
-        <test expect_num_outputs="2">
+            <assert_stderr>
+                <has_text text="Downloading" negate="true"/>
+            </assert_stderr>
+        </test> -->
+        <!-- bowtie2out, cached DB -->
+        <!-- <test expect_num_outputs="2">
```
bgruening, line 1054: Good from my side, but I guess you want to take alle those tests in again?

PR #6193, opened 2024-07-29
add flexynesis

nilchia:
FOR CONTRIBUTOR:

This PR adds [flexynesis](https://github.com/BIMSBbioinfo/flexynesis) to Galaxy


nilchia:
The problem is that the `data_types` argument should be the name of the omics file (E.g. gex here).
but when the value is set to "gex" it thinks that it is a file and not a text. So it gives: `--data_types {'src': 'hda', 'id': '55d23f6c34fa3eae'}` for example.

nilchia:
I added another omics file and that problem is solved.

nilchia:
why does it use `diff` for "job.feature_importance" element?
I did not set `compare="diff"`

bgruening:
You specified `file` so you want to compare the file with something ... and the default comparison method is diff. Remove the `file`.

nilchia:
Thanks @bgruening !

tools/flexynesis/flexynesis.xml
```diff
@@ -0,0 +1,322 @@
...
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="edam"/>
+    <expand macro="requirements"/>
+    <command detect_errors="exit_code"><![CDATA[
+        #if str($license_type) == 'academic' or str($license_type) == 'commercial':
+            mkdir -p 'input/test' &&
+            mkdir -p 'input/train' &&
+            mkdir -p 'output' &&
```
bgruening, line 12:
```suggestion
            mkdir -p input/test input/train output &&
```

tools/flexynesis/macros.xml
```diff
@@ -0,0 +1,133 @@
...
+                        <option value="RandomSurvivalForest">RandomSurvivalForest</option>
+                    </param>
+                    <when value="DirectPred"></when>
+                    <when value="GNN">
+                        <param argument="--gnn_conv_type" type="select" label="GNN convolution type" help="Which graph convolution type to use.">
+                            <option value="GC">GC</option>
+                            <option value="GCN">GCN</option>
+                            <option value="SAGE">SAGE</option>
+                        </param>
+                        <param argument="--string_organism" type="integer" value="9606" label="string organism" help="STRING DB organism id. Default is 9606 (Homo Sapiens)." />
```
bgruening, line 76: Maybe add here a select with optional integer overwrite? And include most command organisms?

tools/flexynesis/macros.xml
```diff
@@ -0,0 +1,133 @@
...
+                        <param argument="--string_node_name" type="select" label="String node name" help="">
+                            <option value="gene_name">Gene name</option>
+                            <option value="gene_id">Gene id</option>
+                        </param>
+                    </when>
+                    <when value="MultiTripletNetwork"></when>
```
bgruening, line 82:
```suggestion
                    <when value="MultiTripletNetwork"/>
```

bgruening: This could be proposed as a galaxy-language-server enhancement.

tools/flexynesis/macros.xml
```diff
@@ -0,0 +1,133 @@
...
+                    <when value="RandomForest"></when>
+                    <when value="SVM"></when>
+                    <when value="RandomSurvivalForest"></when>
+                </conditional>
+                <param argument="--target_variables" type="text" label="Target variables" help="Which variables in 'clin.csv' to use for predictions, comma-separated if multiple.">
```
bgruening, line 87: Do you know this param type: https://docs.galaxyproject.org/en/latest/dev/schema.html#data-column

tools/flexynesis/macros.xml
```diff
@@ -0,0 +1,133 @@
...
+            <when value="cm_train">
+                <param argument="--model_class" type="select" label="Model class" help="The kind of model class to instantiate">
+                    <option value="CrossModalPred">CrossModalPred</option>
+                </param>
+                <param argument="--input_layers" type="text" label="Input layers" help="Which data types to use as input/encoded layers, comma-separated if multiple.">
+                    <sanitizer invalid_char="">
+                        <valid initial="string.printable"></valid>
+                    </sanitizer>
+                </param>
+                <param argument="--output_layers" type="text" label="Output layers" help="Which data types to use as output/decoded layers, comma-separated if multiple.">
```
bgruening, line 119: I assume its quite hard for a user to know what this input here is. Any idea?

tools/flexynesis/macros.xml
```diff
@@ -0,0 +1,133 @@
...
+            <param argument="--early_stop_patience" type="integer" min="-1" value="10" label="Early stop patience" help="How many epochs to wait when no improvements in validation loss are observed. Set to -1 to disable early stopping." />
+            <param argument="--hpo_iter" type="integer" min="1" value="100" label="Hyperparameter Iterations" help="Number of iterations for hyperparameter optimisation." />
+            <param argument="--hpo_patience" type="integer" min="0" value="10" label="Hyperparameter patience" help="How many hyperparameter optimisation iterations to wait for when no improvements are observed. Set to 0 to disable early stopping." />
+            <param argument="--use_cv" type="boolean" truevalue="--use_cv" falsevalue="" checked="false" label="Cross validation" help="If set, a 5-fold cross-validation training will be done. Otherwise, a single training on 80 percent of the dataset is done. " />
+            <param argument="--use_loss_weighting" type="boolean" truevalue="True" falsevalue="False" checked="true" label="Loss weighting" help="Whether to apply loss-balancing using uncertainty weights method." />
```
bgruening, line 40:
```suggestion
            <param argument="--use_loss_weighting" type="boolean" truevalue="--use_loss_weighting" falsevalue="" checked="true" label="Loss weighting" help="Whether to apply loss-balancing using uncertainty weights method." />
```

tools/flexynesis/flexynesis.xml
```diff
@@ -0,0 +1,322 @@
...
+                        --variance_threshold $variance_threshold
+                        --correlation_threshold $correlation_threshold
+                        --subsample $subsample
+                        --features_min $features_min
+                        --features_top_percentile $features_top_percentile
+                        --data_types $data_types
+                        --log_transform $log_transform
+                        --early_stop_patience $early_stop_patience
+                        --hpo_patience $hpo_patience
+                        --use_loss_weighting $use_loss_weighting
```
bgruening, line 52:
```suggestion
                        $use_loss_weighting
```

tools/flexynesis/flexynesis.xml
```diff
@@ -0,0 +1,322 @@
...
+                        $use_cv
+                        $evaluate_baseline_performance
+                        $disable_marker_finding
+        #end if
+    ]]></command>
+    <inputs>
+        <conditional name="license" label="LICENSE">
```
bgruening, line 59: You can copy this one here https://github.com/galaxyproject/tools-iuc/blob/6da96d865a3a557cfa3ad09e1cfa830519e73748/tools/meme/fimo.xml#L164

tools/flexynesis/macros.xml
bgruening: Change all booleans to this please.

tools/flexynesis/macros.xml
nilchia: That would be nice. I think only "homo sapiens" and "mus musculus" are needed.

tools/flexynesis/macros.xml
nilchia: The `--input_layers` and `--output_layers` should be set to one or more modalities (set in `--data_types`)
Maybe a better help?
I'll try

tools/flexynesis/flexynesis.xml
nilchia: I think for this it does not work.
Because it should be: `--use_loss_weighting True` or `--use_loss_weighting False`

but for `--disable_marker_finding` it works.

```
 --use_loss_weighting (str) – Whether to apply loss-balancing using uncertainty weights method. Choices are ['True', 'False']. Default is 'True'.

 --disable_marker_finding: (bool) - If set, marker discovery after model training is disabled.
 ```

bgruening: Ok, please create upstream issues about this inconsistency. 
And then you can still use the trick just adopt the truevalue="--use_loss_w True"

nilchia: Nice trick :exploding_head:

tools/flexynesis/macros.xml
nilchia: For this param I have to specify `data_ref=""` but here both `$test` and `$train` have multiple inputs.
Is it possible to specify "clin.csv" from one of these inputs?

tools/flexynesis/macros.xml
nilchia: Better?

tools/flexynesis/macros.xml
bgruening: Ok, too bad, ok leave it like that.

PR #6194, opened 2024-07-30
Add PACU SNP phylogeny workflow

BertBog:
FOR CONTRIBUTOR:

To render the report, the tools must be added to the allow list.


tools/pacu/macros.xml
```diff
@@ -0,0 +1,38 @@
...
+            journal = {Journal of clinical microbiology},
+            language = {eng},
+            month = {mar},
+            pages = {e0157623},
+            pmid = {38441926},
+            title = {{Closing the gap: Oxford Nanopore Technologies R10 sequencing allows comparable results to Illumina
+            sequencing for SNP-based outbreak investigation of bacterial pathogens.}},
+            year = {2024}
+            }
+        </citation>
```
bgruening, line 35:
```suggestion
        <citation type="doi">10.1128/jcm.01576-23</citation>
```

tools/pacu/pacu_map.xml
```diff
@@ -0,0 +1,113 @@
...
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro='xrefs'/>
+    <expand macro="requirements" />
+    <stdio>
+        <exit_code range="1:" level="fatal"/>
+    </stdio>
+    <expand macro="version_command" />
+    <command>
```
bgruening, line 14:
```suggestion
    <expand macro="version_command" />
    <command detect_errors="exit_code">
```

tools/pacu/pacu_map.xml
```diff
@@ -0,0 +1,113 @@
...
+        #if str($fastq_tech.read_type) == 'illumina':
+            --fastq-illumina $fastq_tech.fastq_illumina_1 $fastq_tech.fastq_illumina_2
+            --fastq-illumina-names "$fastq_tech.fastq_illumina_1.name" "$fastq_tech.fastq_illumina_2.name"
+        #elif str($fastq_tech.read_type) == 'ont':
+            --fastq-ont $fastq_tech.fastq_ont
+            --fastq-ont-name "$fastq_tech.fastq_ont.name"
+        #end if
+
+        ## Output
+        --output $bam
```
bgruening, line 33:
```suggestion
        --output '$bam'
```

Please quote all data and text inputs.

tools/pacu/pacu_map.xml
```diff
@@ -0,0 +1,113 @@
...
+        --ref-fasta-name "$ref_fasta.name"
+
+        ## FASTQ input
+        --read-type $fastq_tech.read_type
+        #if str($fastq_tech.read_type) == 'illumina':
+            --fastq-illumina $fastq_tech.fastq_illumina_1 $fastq_tech.fastq_illumina_2
+            --fastq-illumina-names "$fastq_tech.fastq_illumina_1.name" "$fastq_tech.fastq_illumina_2.name"
+        #elif str($fastq_tech.read_type) == 'ont':
+            --fastq-ont $fastq_tech.fastq_ont
+            --fastq-ont-name "$fastq_tech.fastq_ont.name"
```
bgruening, line 29: please use single-quotes

tools/pacu/pacu_map.xml
```diff
@@ -0,0 +1,113 @@
...
+        #end if
+
+        ## Output
+        --output $bam
+
+        ## Working directory
+        --dir-working \$_GALAXY_JOB_TMP_DIR
```
bgruening, line 36: you can simply use `./` Galaxy provides you a clean working dir

tools/pacu/pacu_map.xml
```diff
@@ -0,0 +1,113 @@
...
+
+        ## Other options
+        $trim
+        --threads 4
```
bgruening, line 40: please use GALAXY_SLOTS here

tools/pacu/pacu_snp.xml
```diff
@@ -0,0 +1,132 @@
+<tool id="pacu" name="PACU" version="@VERSION@">
```
bgruening, line 1: can you please use our recommendation for versioning https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tool-versions 

and add a profile version?

tools/pacu/pacu_snp.xml
```diff
@@ -0,0 +1,132 @@
...
+    <description>Workflow for whole genome sequencing based phylogeny of Illumina and ONT R9/R10 data.</description>
```
bgruening, line 2:
```suggestion
    <description>whole genome sequencing based phylogeny of Illumina and ONT R9/R10 data</description>
```

trying to shorten it a bit

tools/pacu/pacu_snp.xml
```diff
@@ -0,0 +1,132 @@
...
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro='xrefs'/>
+    <expand macro="requirements"/>
+    <stdio>
+        <exit_code range="1:" level="fatal"/>
+    </stdio>
+    <expand macro="version_command"/>
+    <command>
```
bgruening, line 12:
```suggestion
    <expand macro="version_command"/>
    <command detect_errors="exit_code">
```

tools/pacu/pacu_snp.xml
```diff
@@ -0,0 +1,132 @@
...
+        <![CDATA[
+        PACU_galaxy
+
+        ## Input files
+        --ref-fasta $input.ref_fasta
+        --ref-fasta-name "$input.ref_fasta.name"
+
+        #for $bam in $input.bam_ilmn:
+            #if $bam:
+                --bam-ilmn $bam "$bam.name"
```
bgruening, line 22: please single-quotes

tools/pacu/pacu_snp.xml
```diff
@@ -0,0 +1,132 @@
...
+    <stdio>
+        <exit_code range="1:" level="fatal"/>
+    </stdio>
+    <expand macro="version_command"/>
+    <command>
+        <![CDATA[
+        PACU_galaxy
+
+        ## Input files
+        --ref-fasta $input.ref_fasta
```
bgruening, line 17: single-quotes

tools/pacu/pacu_snp.xml
```diff
@@ -0,0 +1,132 @@
...
+            </section>
+            <output name="html" ftype="html">
+                <assert_contents>
+                    <has_text text="PACU report"/>
+                </assert_contents>
+            </output>
+        </test>
+    </tests>
+    <help>
+        **Warning:** At least four input BAM files are required for bootstrapping (these can be a mix of Illumina / ONT
```
bgruening, line 125: A data param has a min="4" argument that you can use: https://docs.galaxyproject.org/en/latest/dev/schema.html#id48

tools/pacu/pacu_map.xml
```diff
@@ -0,0 +1,113 @@
+<tool id="pacu_map" name="PACU - mapping" version="@VERSION@">
+    <description>
+        Helper tool to map ONT / Illumina reads to a reference genome for the PACU workflow
```
bgruening, line 3:
```suggestion
        Map ONT / Illumina reads to a reference genome for PACU
```

shorten ...

tools/pacu/pacu_map.xml
```diff
@@ -0,0 +1,113 @@
+<tool id="pacu_map" name="PACU - mapping" version="@VERSION@">
```
bgruening, line 1: please add profile ...

bgruening (commented):
Thanks @BertBog! Please add a shed.yml e.g. https://github.com/galaxyproject/tools-iuc/blob/main/tools/ivar/.shed.yml

I also added some comments inline.

tools/pacu/pacu_map.xml
BertBog: I tried, but the tool currently requires an absolute path for the working directory. I'll fix this when I update the PACU package to the next version, but for now I'll keep the `\$_GALAXY_JOB_TMP_DIR` to make the wrapper work, if that's OK?

tools/pacu/pacu_snp.xml
BertBog: The tool needs four input datasets in total, but this can be a combination of the bam-ont and bam-ilmn options. So forcing a minimum of 4 will not solve the problem.

BertBog:
Hey @bgruening 
Thanks a lot for the quick feedback! I have addressed your comments. Let me know if I still need to adapt some things.

tools/pacu/pacu_map.xml
bgruening: You can use `\$PWD` I think to get to the current working dir.

BertBog:
> The tests are unfortunately failing. Please see at the bottom of the page here: https://github.com/galaxyproject/tools-iuc/actions/runs/10164187651?pr=6194
> 
> bedtools is missing as dependency is one error.
> 
> You can test this locally with `planemo test --biocontainers ./tool/dir/`

Ok thanks, good to know that I can run the tests in mulled containers. The tests should work now. I missed the dependency issue due to a global `bedtools' installation.
I updated the source to fix the renaming issue with the `PACU_map' script and added the missing dependency to the bioconda recipe.

tools/pacu/pacu_snp.xml
```diff
@@ -0,0 +1,132 @@
...
+
+        #for $bam in $input.bam_ilmn:
+            #if $bam:
+                --bam-ilmn $bam '$bam.name'
+            #end if
+        #end for
+
+        #for $bam in $input.bam_ont:
+            #if $bam:
+                --bam-ont $bam '$bam.name'
```
bgruening, line 25:
```suggestion
                --bam-ont '$bam' '$bam.element_identifier'
```

tools/pacu/pacu_snp.xml
```diff
@@ -0,0 +1,132 @@
...
+            #end if
+        #end for
+
+        ## Phages
+        #if str($input.mask_phages.mask_phages_selector) == 'yes':
+            --ref-bed $input.mask_phages.bed_phages
```
bgruening, line 31:
```suggestion
            --ref-bed '$input.mask_phages.ref_bed'
```
This one? Please add a test for this.

tools/pacu/pacu_snp.xml
```diff
@@ -0,0 +1,132 @@
...
+            <param argument="--bam-ont" type="data" format="bam" multiple="true" optional="true" label="BAM files (ONT)"
+                   help="Make sure the reads are mapped to the reference selected above."/>
+            <conditional name="mask_phages">
+                <param name="mask_phages_selector" type="select" label="Mask (pro-)phage regions">
+                    <option value="no" selected="true">No</option>
+                    <option value="yes">Yes</option>
+                </param>
+                <when value="yes">
+                    <param argument="--ref-bed" type="data" format="bed"
+                           label="BED file with (pro-)phage regions in the reference genome"/>
```
bgruening, line 73: You could also just make this input file optional="true" and get rid of the conditional.

tools/pacu/pacu_map.xml
```diff
@@ -0,0 +1,106 @@
...
+        --ref-fasta-name '$ref_fasta.name'
+
+        ## FASTQ input
+        --read-type $fastq_tech.read_type
+        #if str($fastq_tech.read_type) == 'illumina':
+            --fastq-illumina $fastq_tech.fastq_illumina_1 $fastq_tech.fastq_illumina_2
+            --fastq-illumina-names '$fastq_tech.fastq_illumina_1.name' '$fastq_tech.fastq_illumina_2.name'
+        #elif str($fastq_tech.read_type) == 'ont':
+            --fastq-ont $fastq_tech.fastq_ont
+            --fastq-ont-name '$fastq_tech.fastq_ont.name'
```
bgruening, line 24: use element_identifier, see below

tools/pacu/pacu_map.xml
```diff
@@ -0,0 +1,106 @@
...
+    <expand macro='xrefs'/>
+    <expand macro="requirements" />
+    <expand macro="version_command"/>
+    <command detect_errors="exit_code">
+        <![CDATA[
+        PACU_map
+
+        ## FASTA input
+        --ref-fasta '$ref_fasta'
+        --ref-fasta-name '$ref_fasta.name'
```
bgruening, line 15: element_identifier see below

bgruening:
Cool @BertBog! Its green! I added a few more comments, we are getting there. Thanks again!

tools/pacu/pacu_snp.xml
BertBog: nice one, updated

tools/pacu/pacu_snp.xml
BertBog: fixed & test added

tools/pacu/pacu_snp.xml
```diff
@@ -0,0 +1,132 @@
...
+        <![CDATA[
+        PACU_galaxy
+
+        ## Input files
+        --ref-fasta '$input.ref_fasta'
+        --ref-fasta-name '$input.ref_fasta.name'
+
+        #for $bam in $input.bam_ilmn:
+            #if $bam:
+                --bam-ilmn $bam '$bam.name'
```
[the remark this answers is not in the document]
BertBog: The tools also uses the name because it generates an output report and `my_input_a.bam` and `my_input_b.bam` are easier to link back to the original data compared to `datasets_001.dat`, dataset_002.dat`, etc. 
See for example in this screenshot:
image

tools/pacu/pacu_snp.xml
```diff
@@ -0,0 +1,132 @@
...
+    <command detect_errors="exit_code">
+        <![CDATA[
+        PACU_galaxy
+
+        ## Input files
+        --ref-fasta '$input.ref_fasta'
+        --ref-fasta-name '$input.ref_fasta.name'
+
+        #for $bam in $input.bam_ilmn:
+            #if $bam:
```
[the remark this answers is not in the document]
BertBog: I just tested and if the input is empty, it returns `None` instead of an empty list.
`--bam-ont None 'None'`
So I think this check needs to stay.

From the docs:
"Perhaps counter-intuitively, a multiple="true" data parameter requires at least one data input. If optional="true" is specified, this condition is relaxed and the user is allowed to select 0 datasets. Unfortunately, if 0 datasets are selected the resulting value for the parameter during Cheetah templating (such as in a command block) will effectively be a list with one None-like entity in it."

tools/pacu/pacu_snp.xml
```diff
@@ -0,0 +1,132 @@
+<tool id="pacu" name="PACU" version="@VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
```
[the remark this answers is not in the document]
BertBog: Changed to pacu_snp

tools/pacu/pacu_snp.xml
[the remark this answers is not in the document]
bgruening: Most of the times Galaxy names are looking like "Tool B on Data A". Is that useful for you? Maybe ask explicitly for a name in that case? Or optionally and fallback to the element_identifier if no name is provided?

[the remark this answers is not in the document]
BertBog: It's a bit tricky. I'm not sure that adding names as parameters is convenient because it's a multiple input field. Using the element identified or name allows the users to rename their datasets (if needed) to the label they want in the report. I could add an additional explanation to the help section in the wrapper?

One way around this problem is that the `PACU_map` tool writes a custom tag in the BAM file with the sample name derived from the input FASTQ files. So for the typical PACU analysis with the following steps, the sample name would be derived from the original FASTQ input:
- Upload raw FASTQ files
- Run `PACU_map` to trim the reads and map them to a reference genome
- Run `PACU` on the `PACU_map` output BAM files

However, if users were to generate the BAM files in a different way, the names would indeed be `Tool X on Y`. But they can be renamed manually before using the `PACU_snp` tool.

We've been running the tool for a while on our internal Galaxy instances and it seems to work quite well as is, but I'm open to changing it if you think another solution is more convenient for the users.

tools/pacu/pacu_snp.xml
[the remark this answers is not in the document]
BertBog: It returns `[None]`, so I think `if $bam` still returns True instead of False (since the list is not empty).

BertBog:
Hey @bgruening just checking in. I was wondering if there are any changes needed or if there’s any additional information I can provide on the latest changes.
Please let me know if you need any further details or if there’s anything specific you’d like me to address.

bgruening (approved):
No, this is now great. It just slipped ... sorry! Thanks for the ping.

PR #6198, opened 2024-07-31
Update inspect.xml

nilchia:
FOR CONTRIBUTOR:

when using adata with genes < 500 (The case in image-based spatial transcriptomic) the user need to specify the percent_top argument.
This PR fixes the syntax error in percent_top

### DIFF
--- a/tools/scanpy/inspect.xml
+++ b/tools/scanpy/inspect.xml
@@ -72,7 +72,7 @@ sc.pp.calculate_qc_metrics(
     #end if
     #if $method.percent_top
         #set $percent_top = [int(x.strip()) for x in str($method.percent_top).split(',')]
-        percent_top=$method.percent_top,
+        percent_top=($method.percent_top),
     #end if
     inplace=True)
 


bgruening:
Please bump the version number. The prefix galaxy+X

tools/scanpy/inspect.xml
```diff
@@ -72,7 +72,7 @@ sc.pp.calculate_qc_metrics(
     #end if
     #if $method.percent_top
         #set $percent_top = [int(x.strip()) for x in str($method.percent_top).split(',')]
-        percent_top=$method.percent_top,
+        percent_top=($method.percent_top),
```
pavanvidem, line 75:
```suggestion
        percent_top=$percent_top,
```

nilchia: thanks!

pavanvidem (approved):

tools/scanpy/inspect.xml
```diff
@@ -72,7 +72,7 @@ sc.pp.calculate_qc_metrics(
     #end if
     #if $method.percent_top
         #set $percent_top = [int(x.strip()) for x in str($method.percent_top).split(',')]
-        percent_top=$method.percent_top,
+        percent_top=$percent_top,
```
bgruening, line 75:
```suggestion
        percent_top=($percent_top),
```

Isn't that needed?

pavanvidem: `$percent_top` already has an integer list

nilchia: `$percent_top` will give `[50,100,200]` for example.
before, `$method.percent_top` was called which would give `50,100,200`

for example we checked `qc_vars` in here:
```
    #if $method.qc_vars
        #set $qc_vars = [str(x.strip()) for x in str($method.qc_vars).split(',')]
    qc_vars=$qc_vars,
    #end if
```
it will give `qc_vars=['mito', 'ERCC'],` for example

bgruening: Having a test for this case would be good I think.

bgruening:
Thanks!

PR #6207, opened 2024-08-05
fastqc: set tmpdir

bernt-matthias:
xref https://github.com/galaxyproject/galaxy/pull/18628

FOR CONTRIBUTOR:


tools/fastqc/rgFastQC.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="fastqc" name="FastQC" version="0.73+galaxy0">
+<tool id="fastqc" name="FastQC" version="0.74+galaxy0">
```
mvdbeek, line 1: The 0.73.galaxy0 in the diff here is bizarre! https://github.com/galaxyproject/tools-iuc/blob/main/tools/fastqc/rgFastQC.xml#L1

```suggestion
<tool id="fastqc" name="FastQC" version="0.74+galaxy1" profile="23.0">
```

Off-topic: Is there any reason why there are still questions about when tools should be bumped ?
We should finally say that it is **always**, we can only have nice things if we consistently bump for every change.

bernt-matthias: > Off-topic: Is there any reason why there are still questions about when tools should be bumped ?

Nope. Was quite sure that I did :) IMO it did not matter if we bump the version or galaxyX here. But no idea where the 0.73 came from.

> We should finally say that it is always, we can only have nice things if we consistently bump for every change.

+1

bgruening: I also would say, let's bump always!
Tool versions are still a bit too expensive in my view, but that's a different problem.

bgruening (approved):

PR #6213, opened 2024-08-07
Fix ampvis2 NA in metadata error

paulzierep:
If there are NA values in the metadata, the subset tool fails with:  
```
Error in if (classes[[name]] == "character" && all(data$metadata[[name]] ==  : 
  missing value where TRUE/FALSE needed
```
Fixed by setting the values to "NA".

@bernt-matthias any suggestion ?


tools/ampvis2/macros.xml
```diff
@@ -49,6 +49,7 @@ echo $(R --version | grep "R version")", ampvis2 version" $(R --vanilla --slave
                     ## the additional column in the ist file can the be used
                     ## to filter SampleIDs from select inputs
                     ## (check for character columns only .. since data in the column is otherwise converted to the corresponding type .. which fails)
+                    data\$metadata[is.na(data\$metadata)] <- "NA" ##the comparison fails if NA values are present, see https://stackoverflow.com/questions/7355187/error-in-if-while-condition-missing-value-where-true-false-needed
```
bernt-matthias, line 52: Should we do this in the for loop? 

What is `data\$metadata[is.na(data\$metadata)]` actually doing? Guess some subsetting. Intuitively I would have expected the negation of `is.na`  (but its R, i.e. it might be unintuitive) .. also my question would be if we need to modify anything else in `data`.

bernt-matthias (commented):
Thanks for the initiative. Would be cool to add a corresponding test.

tools/ampvis2/macros.xml
paulzierep: * You are right, this can be fixed outside the loop
* `data\$metadata[is.na(data\$metadata)]` selects all cells which are NA
* I think the data can be untouched, as only the metadata is changed, but if it should be changed to as string as "NA" or undefined or something -  I am not sure - the user should probably be warned that the metadata will be modified

bgruening:
@bernt-matthias ok from your side?

bernt-matthias:
Yep. I enabled automerge. Can someone restart tests. I only have my mobile until end of the month.

bernt-matthias:
Thanks @paulzierep

PR #6214, opened 2024-08-07
Add new hierachical clustering tool based on scipy

wm75:
A tool like this was requested by @dbrites

FOR CONTRIBUTOR:


wm75:
One thing this version does not support atm is *multiple* heights or n_clusters values, which, on the command line would result in multi-column tabular output with the cluster assignments for all different settings.
Let me know if that seems of interest.

tools/clustering_from_distmat/clustering_from_distmat.xml
```diff
@@ -0,0 +1,102 @@
...
+                <option value="height">Yes, and use distance threshold to divide into clusters</option>
+            </param>
+            <when value="dendrogram-only" />
+            <when value="n-cluster">
+                <param name="n_cluster" type="integer" value="5" min="1" label="How many clusters to divide into?" />
+                <param name="generate_dendrogram" type="boolean" label="Produce also the dendrogram of clustering results" />
+            </when>
+            <when value="height">
+                <param name="height" type="float" value="5.0" label="Distance threshold for clusters to be reported" />
+                <param name="generate_dendrogram" type="boolean" label="Produce also the dendrogram of clustering results" />
```
bgruening, line 42: This param is not used in the CLI

wm75: Yes, this is difficult (and not so important) to implement in the cli.
Instead this is used only to filter the outputs on the Galaxy side to avoid clutter in the history.
Not a good idea?

tools/clustering_from_distmat/.shed.yml
```diff
@@ -0,0 +1,6 @@
+name: clustering_from_distmat
+owner: iuc
+description: Distance matrix-based hierarchical clustering using SciPy
+remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/master/tools/clustering_from_distmat/
```
bgruening, line 4:
```suggestion
remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/master/tools/clustering_from_distmat/
long_description: Distance matrix-based hierarchical clustering using SciPy
homepage_url: https://github.com/galaxyproject/tools-iuc/tree/master/tools/clustering_from_distmat/
```

bgruening (approved):

wm75:
We should maybe have column names for the cluster assignment output. Let me add this still before merging.

tools/clustering_from_distmat/test-data/test_assignment_average_h18.tsv
```diff
@@ -0,0 +1,6 @@
+sample	cluster_id_h18.0
```
bgruening, line 1: why is this here h18.0 a float and below an integer? Is anything wired going on?
Any both files are identical besides the header...

wm75: h18.0 is the height at which the tree got cut in order to produce the cluster IDs below, and height can be a float.
The results are otherwise identical because I chose --height and -n so that both tests give the same result. So before I added now the column names only a single test file was needed.

PR #6218, opened 2024-08-09
ci.yaml and pr.yaml update

heylf:
FOR CONTRIBUTOR:
* [X] I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [ ] License permits unrestricted use (educational + commercial)
* [ ] This PR adds a new tool or tool collection
* [ ] This PR updates an existing tool or tool collection
* [X] This PR does something else (explain below)

PR #6228, opened 2024-08-12
Revert "ci.yaml and pr.yaml update"

mvdbeek:
Reverts galaxyproject/tools-iuc#6218

With planemo 0.75.25 we should be getting pillow and tifffile, making this unnecessary.


mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/10485651756

PR #6242, opened 2024-08-14
Change the url to the right files in gtdbtk DM

SantaMcCloud:
FOR CONTRIBUTOR:
* [x] I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [x] License permits unrestricted use (educational + commercial)
* [ ] This PR adds a new tool or tool collection
* [x] This PR updates an existing tool or tool collection
* [ ] This PR does something else (explain below)

PR #6244, opened 2024-08-15
Update GTDB-tk DM

paulzierep:
FOR CONTRIBUTOR:

This PR adds:
* a test case that checks all URLs used if they exist
* more explicit distinction between metadata and main DB
* Test case for both metadata types as spotted here: https://github.com/galaxyproject/tools-iuc/pull/6242

### DIFF
--- a/data_managers/data_manager_gtdbtk_database_installer/data_manager/gtdbtk_database_installer.py
+++ b/data_managers/data_manager_gtdbtk_database_installer/data_manager/gtdbtk_database_installer.py
@@ -9,7 +9,7 @@ import sys
 import tarfile
 from datetime import datetime
 from urllib.parse import urlparse
-from urllib.request import Request, urlopen
+from urllib.request import HTTPError, Request, urlopen
 
 # rather provide the urls based on the release, less error potential for the admins !
 urls = {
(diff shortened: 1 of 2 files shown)


paulzierep:
One thing that should be done to verify that the full DB also works:
Download all full DBs locally and check the content, if they contain a folder or raw content. 
This cannot be done in the wrapper since the DB is too large. But maybe @SantaMcCloud you could

* use the jupytool
* download all DBs (take like 2 Hours each) using `wget https://data.gtdb.ecogenomic.org/releases/release202/202.0/auxillary_files/gtdbtk_r202_data.tar.gz`, 
* document the content with something like
```
tar -xvzf gtdbtk_r202_data.tar.gz -C some_custom_folder_name
tree some_custom_folder_name
```

And add the output to the help of the DM. I suspect, that the structure is not as we expect it, at least in the 220 version, since that DM failed on EU.
This can be done independently of this PR !

bgruening (approved):

SantaMcCloud:
Didn't done what was written here: https://github.com/galaxyproject/tools-iuc/pull/6244#issuecomment-2290885557, but I just write this comment now to tell that the DM works fine, and all Metadata are usable for my wrapper!

PR #6263, opened 2024-08-22
Update InferCNV: control number of threads, downgrade profile and sparse matrix support

pcm32:
FOR CONTRIBUTOR:

Minor change in the wrapper to control number of threads. It also downgrades the profile for wider compatibility.


pcm32:
I will test installation as is first to see if the API based installation with higher profile than the instance works.

tools/infercnv/infercnv.xml
```diff
@@ -53,6 +55,9 @@
                                 num_ref_groups = NULL, # keep this as default
                                 ref_subtract_use_mean_bounds = TRUE, # keep this as default
 
+                                # Control number of threads for multiprocessesor use
+                                #num_threads = num_threads_galaxy_slots,
```
bgruening, line 59:
```suggestion
                                #num_threads = num_threads_galaxy_slots,
```

You need to put the `os.environ.get('GALAXY_SLOTS', 1)` into this script.

pcm32: You mean I should grab the GALAXY_SLOT from R at level with `Sys.getenv('GALAXY_SLOT')` or use the cheetah construct `$os.environ.get('GALAXY_SLOTS', 1)` directly here? Won't I still need the cheetah `import os` which might be breaking things? Thanks.

bgruening: You can not use cheetah, you need to get the envs from within the script. In that case your R script.

pcm32: I see... I found a potential way in another tool, tried before I saw your response, if it doesn't work will switch to the R way.

pcm32:
...finally....

tools/infercnv/infercnv.xml
pcm32: Moved now to full R logic for grabbing GALAXY_SLOTS. Thanks!

pcm32:
Looking in more detail, this tool is receiving a non-sparse matrix in tsv format, that is prohibitive for large datasets. The tool should instead receive a sparse matrix format, ideally matrix market or a seurat object. Otherwise this means exploding very large matrices.

bgruening:
LGTM. Do you want to add the matrix-market ...?

pcm32:
Yes, almost there with the sparse matrix support.

pcm32:
I'm once again somehow offending the Cheetah gods... 

```
Job in error state.. tool_id: infercnv, exit_code: None, stderr: .
Job in error state.. tool_id: infercnv, exit_code: None, stderr: .
Output collection 'infercnv_results': failed to find identifier 'infercnv.heatmap_thresholds' in the tool generated elements []
```

new test passes, but previous ones don't - could you please advice @bgruening ? Thanks.

tools/infercnv/infercnv.xml
```diff
@@ -9,6 +9,12 @@
             ## extract gene order from gtf file
             python '$__tool_directory__/gtf_to_position_file.py' --attribute_name $attribute '$gtf_file' 'gene_order.tabular' &&
         #end if
+        #if $input_mat.mtx and $input_mat.mtx_genes and $input_mat.mtx_barcodes:
```
nilchia, line 12:
```suggestion
        #if str($input_mat.format) == "mtx":
```

tools/infercnv/infercnv.xml
```diff
@@ -19,9 +25,16 @@
             # load library
             library('infercnv')
 
+            #if $input_mat.mtx and $input_mat.mtx_genes and $input_mat.mtx_barcodes:
```
nilchia, line 28:
```suggestion
            #if str($input_mat.format) == "mtx":
```

nilchia (changes requested):

pcm32:
Thanks @nilchia for the suggestions!

nilchia:
You're welcome @pcm32 !
Thanks for the update.

nilchia (approved):

tools/infercnv/infercnv.xml
```diff
@@ -169,7 +190,20 @@
...
+            <param name="format" type="select" label="Matrix format">
+                <option value="tsv">Tabular (non-sparse matrix)</option>
+                <option value="mtx" selected="true">10x sparse matrix</option>
+            </param>
+            <when value="tsv">
+                <param name="raw_mtx" type="data" format="tabular" label="Raw counts matrix" help="The matrix of genes (rows) vs. cells (columns) containing the raw counts" />
+            </when>
+            <when value="mtx">
+                <param name="mtx" type="data" format="mtx,tabular" label="Raw counts matrix (matrix market)" help="The matrix of genes (rows) vs. cells (columns) containing the raw counts" />
+                <param name="mtx_genes" type="data" format="tsv,tabular" label="Genes file" help="The genes file for matrix market matrix, a two column symbol - symbol or gene_id - symbol tab separated file" />
```
bgruening, line 203: tabular and tsv? I guess tabular is enough: https://github.com/galaxyproject/galaxy/blob/e9c6a071c73ec764ddb0963c87b0e4e829138872/lib/galaxy/datatypes/tabular.py#L1512

pcm32: Agreed. Added txt in exchange, sometimes on upload some of those files end up as text and then is less hassle to use the tool/pipeline.

bgruening (approved):

PR #6267, opened 2024-08-22
Bump gemini_load.xml wrapper version

chambm:
It looks like the last wrapper update to this tool did not bump the version as it should have.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/gemini/gemini_load.xml
+++ b/tools/gemini/gemini_load.xml
@@ -1,4 +1,4 @@
-<tool id="gemini_@BINARY@" name="GEMINI @BINARY@" version="@VERSION@+galaxy2">
+<tool id="gemini_@BINARY@" name="GEMINI @BINARY@" version="@VERSION@+galaxy3">
     <description>Loading a VCF file into GEMINI</description>
     <expand macro="bio_tools"/>
     <macros>


wm75:
which wrapper update are you referring to exactly? The last one only adjusted the wrapper tests afaict and that alone is not considered a reason to bump the version.

chambm:
https://github.com/galaxyproject/tools-iuc/commit/f7bdf08922aaf4119aefe7041e754a69cf64aebd

I hadn't noticed that was only changes to tests. However the macro file also changed. And I think the most important thing is there are two revisions listed with the same version in the toolshed. That doesn't seem like it should ever happen?
https://toolshed.g2.bx.psu.edu/repository?repository_id=9a30a14600b75038

The reason I made the PR is when I tried to install the newest revision it got stuck on "New":
image

wm75:
Generally, installing different revisions with identical tool versions works fine, e.g. see the situation on .eu below:

Screenshot from 2024-09-09 10-47-27

The rule applied for iuc wrappers is that if a change is purely technical and doesn't affect functionality of the wrapper or the underlying command line tool, then we don't do a version bump.

PR #6278, opened 2024-08-26
Chnage name in .shed  to match with the wrapper

SantaMcCloud:
FOR CONTRIBUTOR:


SantaMcCloud:
@bgruening this sould fix the problem with including the tool. Sorry i didnt know it should be the .shed name thought it should be the name of the wrapper

SantaMcCloud:
fix it

tools/gtdb_to_taxdump/.shed.yml
```diff
@@ -1,4 +1,4 @@
-name: gtdb_to_taxdump
+name: ncbi_gtdb_map
```
bgruening, line 1: I don't understand. The folder ist called "gtdb_to_taxdump" and the github repo of the upstream tool is also called "gtdb_to_taxdump". Why renaming?

mvdbeek: It would also break updating the tool from one version to another, you'd have to insert it as a new tool in workflows or when used via the rerun functionality.

SantaMcCloud:
I thiught about since there was an error for adding the tool to eu. Now in my second thought this PR is unnecessary sorry for opening it and the confusions.

tools/gtdb_to_taxdump/.shed.yml
bgruening: I think I got it now. The tool id is not inline with the repo name ... which would be nice for convenience. The tool was just released over the weekend - so I assume we can still change it - or stick to the tool ID forever.

https://github.com/galaxyproject/tools-iuc/blob/af516fb4fa4dfa42920ff9006a47f939fa1e86d2/tools/gtdb_to_taxdump/ncbi-gtdb_map.xml#L1

PR #6285, opened 2024-08-29
Update Samtools collection to 1.20 

fubar2:

Update the samtools collection to 1.20

Fixed the samtools_split problem - it was outputing sam instead of bam... and the samtools_merge problem - the error message has changed.

Seeing if this passes - all tests upated


tool_collections/samtools/samtools_view/samtools_view.xml
```diff
@@ -300,6 +302,21 @@
...
                         </when>
                     </conditional>
+                    <conditional name="cond_expr">
+                        <param name="select_expr" type="select" label="Filter by expression">
+                            <option value="no" selected="True">No</option>
+                            <option value="yes">Filter using an expression (see manual)</option>
+                        </param>
+                        <when value="no"/>
+                        <when value="yes">
+                            <param name="expression" type="text" argument="-e" label="Filter by expression - for example sclen&gt;0 will filter all soft clipped reads" help="See Samtools manual for Filter expression syntax">
```
bgruening, line 312: Is a negation possible something like !chrM ...

fubar2: Never tried. 
It is powerful - the sanitizer is like grep's so arguably could be tightened.

fubar2: from man samtools, @bgruening - some missing but yes, ! can be used for sam fields.

>FILTER EXPRESSIONS
       Filter expressions are used as an on-the-fly checking of incoming SAM, BAM or CRAM records, discarding records that do not match the specified expression.

       The language used is primarily C style, but with a few differences in the precedence rules for bit operators and the inclusion of regular expression matching.

       The operator precedence, from strongest binding to weakest, is:

       Grouping        (, )             E.g. "(1+2)*3"
       Values:         literals, vars   Numbers, strings and variables
       Unary ops:      +, -, !, ~       E.g. -10 +10, !10 (not), ~5 (bit not)
       Math ops:       *, /, %          Multiply, division and (integer) modulo
       Math ops:       +, -             Addition / subtraction
       Bit-wise:       &                Integer AND
       Bit-wise        ^                Integer XOR
       Bit-wise        |                Integer OR
       Conditionals:   >, >=, <, <=
       Equality:       ==, !=, =~, !~   =~ and !~ match regular expressions
       Boolean:        &&, ||           Logical AND / OR

       Expressions  are  computed  using  floating  point mathematics, so "10 / 4" evaluates to 2.5 rather than 2.  They may be written as integers in decimal or "0x" plus hexadecimal, and
       floating point with or without exponents.However operations that require integers first do an implicit type conversion, so "7.9 % 5" is 2 and "7.9 & 4.1" is equivalent to "7  &  4",
       which  is  4.  Strings are always specified using double quotes.  To get a double quote in a string, use backslash.  Similarly a double backslash is used to get a literal backslash.
       For example ab\"c\\d is the string ab"c\d.

       Comparison operators are evaluated as a match being 1 and a mismatch being 0, thus "(2 > 1) + (3 < 5)" evaluates as 2.  All comparisons involving undefined (null) values are  deemed
       to be false.

       The  variables are where the file format specifics are accessed from the expression.  The variables correspond to SAM fields, for example to find paired alignments with high mapping
       quality and a very large insert size, we may use the expression "mapq >= 30 && (tlen >= 100000 || tlen <= -100000)".  Valid variable names and their data types are:

       endpos               int            Alignment end position (1-based)
       flag                 int            Combined FLAG field
       flag.paired          int            Single bit, 0 or 1
       flag.proper_pair     int            Single bit, 0 or 2
       flag.unmap           int            Single bit, 0 or 4
       flag.munmap          int            Single bit, 0 or 8
       flag.reverse         int            Single bit, 0 or 16
       flag.mreverse        int            Single bit, 0 or 32
       flag.read1           int            Single bit, 0 or 64
       flag.read2           int            Single bit, 0 or 128
       flag.secondary       int            Single bit, 0 or 256
       flag.qcfail          int            Single bit, 0 or 512
       flag.dup             int            Single bit, 0 or 1024
       flag.supplementary   int            Single bit, 0 or 2048
       hclen                int            Number of hard-clipped bases
       library              string         Library (LB header via RG)
       mapq                 int            Mapping quality
       mpos                 int            Synonym for pnext
       mrefid               int            Mate reference number (0 based)
       mrname               string         Synonym for rnext
       ncigar               int            Number of cigar operations
       pnext                int            Mate's alignment position (1-based)
       pos                  int            Alignment position (1-based)
       qlen                 int            Alignment length: no. query bases
       qname                string         Query name
       qual                 string         Quality values (raw, 0 based)
       refid                int            Integer reference number (0 based)
       rlen                 int            Alignment length: no. reference bases
       rname                string         Reference name
       rnext                string         Mate's reference name
       sclen                int            Number of soft-clipped bases
       seq                  string         Sequence
       tlen                 int            Template length (insert size)
       [XX]                 int / string   XX tag value

       Flags are returned either as the whole flag value or by checking for a single bit.  Hence the filter expression flag.dup is equivalent to flag & 1024.

       "qlen" and "rlen" are measured using the CIGAR string to count the number of query (sequence) and reference bases consumed.  Note "qlen" may not exactly  match  the  length  of  the
       "seq" field if the sequence is "*".

       "sclen"  and  "hclen"  are the number of soft and hard-clipped bases respectively.  The formula "qlen-sclen" gives the number of sequence bases used in the alignment, distinguishing
       between global alignment and local alignment length.

       "endpos" is the (1-based inclusive) position of the rightmost mapped base of the read, as measured using the CIGAR string, and for mapped reads is equivalent  to  "pos+rlen-1".  For
       unmapped reads, it is the same as "pos".

       Reference names may be matched either by their string forms ("rname" and "mrname") or as the Nth @SQ line (counting from zero) as stored in BAM using "tid" and "mtid" respectively.

       Auxiliary  tags  are  described  in square brackets and these expand to either integer or string as defined by the tag itself (XX:Z:string or XX:i:int).  For example [NM]>=10 can be
       used to look for alignments with many mismatches and [RG]=~"grp[ABC]-" will match the read-group string.

       If no comparison is used with an auxiliary tag it is taken simply to be a test for the existence of that tag.  So [NM] will return any record containing an NM tag, even if that  tag
       is  zero (NM:i:0).  In htslib <= 1.15 negating this with ![NM] gave misleading results as it was true if the tag did not exist or did exist but was zero.  Now this is strictly does-
       not-exist.  An explicit exists([NM]) and !exists([NM]) function has also been added to make this intention clear.

       Similarly in htslib <= 1.15 using [NM]!=0 was true both when the tag existed and was not zero as well as when the tag did not exist.  From 1.16 onwards all comparison operators  are
       only true for tags that exist, so [NM]!=0 works as expected.

       Some  simple functions are available to operate on strings.  These treat the strings as arrays of bytes, permitting their length, minimum, maximum and average values to be computed.
       These are useful for processing Quality Scores.

       length(x)   Length of the string (excluding nul char)
       min(x)      Minimum byte value in the string
       max(x)      Maximum byte value in the string
       avg(x)      Average byte value in the string

       Note that "avg" is a floating point value and it may be NAN for empty strings.  This means that "avg(qual)" does not produce an error for records that have both seq and qual of "*".
       NAN values will fail any conditional checks, so e.g. "avg(qual) > 20" works and will not report these records.  NAN also fails all equality, < and > comparisons,  and  returns  zero
       when given as an argument to the exists function.  It can be negated with !x in which case it becomes true.

       Functions that operate on both strings and numerics:

       exists(x)      True if the value exists (or is explicitly true).
       default(x,d)   Value x if it exists or d if not.

       Functions that apply only to numeric values:

fubar2:
Two broken tests need help please?
Failed Tests

    ❌ samtools_markdup (Test # 5)

    Problems:

        Standard error of the job different than expected
        Expected line '[markdup] error: no MC tag. Please run samtools fixmate on file first.' in output ('Fatal error: Exit code 1 (Error)
        samtools markdup: error, no MC tag. Please run samtools fixmate on file first.

        samtools markdup: error, unable to assign pair hash key.

        ')

and 

<markdown-accessiblity-table data-catalyst="">
 
--


</markdown-accessiblity-table>
<details open=""><summary>Failed Tests</summary>
<ul>
<li>
<details><summary>❌ samtools_markdup (Test # 5)</summary></details>
</li>
<li>
<details open=""><summary>❌ samtools_split (Test # 1)</summary><div>
<p><strong>Problems</strong>:</p>
<ul>
<li>
<div class="snippet-clipboard-content notranslate position-relative overflow-auto"><pre class="notranslate"><code class="notranslate">Job in error state.. tool_id: samtools_split, exit_code: None, stderr: .
</code></pre><div class="zeroclipboard-container position-absolute right-0 top-0">
    </div></div></li></ul></div></details></li></ul></details>
<div class="snippet-clipboard-content notranslate position-relative overflow-auto"><pre class="notranslate"><code class="notranslate">Job in error state.. tool_id: samtools_split, exit_code: None, stderr: .
</code></pre></div>

bgruening:
I can help with the tests, can you add this advanced syntax to the tool help, or link it out and maybe add some more tests for the syntax?

fubar2:
@bgruening - thanks for the help. 

- added one more filter test - like your question about removing chrM - it works but does not adjust the header according to the docs.
- Managed to add some help
   - markdown is a terrible choice here - can we think about adding an option for some less awful markup in tool help please?

fubar2:
@bgruening: all tests passing now with samtools 1.20

tool_collections/samtools/samtools_view/samtools_view.xml
```diff
@@ -312,9 +312,7 @@
                             <param name="expression" type="text" argument="-e" label="Filter by expression - for example sclen&gt;0 will filter all soft clipped reads" help="See Samtools manual for Filter expression syntax">
                                 <sanitizer invalid_char="">
                                     <valid initial="string.printable">
-                                    <remove value=" "/>
                                     <remove value="'"/>
-                                    <remove value='"'/>
```
bgruening, line 317: Can those be added back? " Are not needed for the chromosomes correct?

bgruening (approved):
Awesome, thanks a lot! One small comment.

tool_collections/samtools/samtools_view/samtools_view.xml
fubar2: Without `"` cannot write 'qname == "foo"'
`'` is removed only.

fubar2:
@bgruening: thanks! 
That was an interesting experience: more work than I expected for an old collection like samtools. 1.51 was April 2022...

PR #3945, opened 2021-09-13
Updating tools/bcftools from version 1.15.1 to 1.24

ghost:
Hello! This is an automated update of the following tool: **tools/bcftools**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

Please see https://github.com/planemo-autoupdate/autoupdate for more information.

PR #6293, opened 2024-09-02
bcftools changes

bgruening:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/bcftools/bcftools_call.xml
+++ b/tools/bcftools/bcftools_call.xml
@@ -237,7 +237,7 @@ For example: --prior-freqs REF_AN,REF_AC
             <param name="output_type" value="v" />
             <output name="output_file">
                 <assert_contents>
-                    <has_text text="MinDP" />
+                    <has_text text="MIN_DP" />
                     <has_text text="DP4=2,4,8,11;MQ=49" />
                 </assert_contents>
             </output>
--- a/tools/bcftools/bcftools_call.xml
+++ b/tools/bcftools/bcftools_call.xml
@@ -1,4 +1,4 @@
-<tool name="bcftools @EXECUTABLE@" id="bcftools_@EXECUTABLE@" version="@TOOL_VERSION@+galaxy5" profile="@PROFILE@">
+<tool name="bcftools @EXECUTABLE@" id="bcftools_@EXECUTABLE@" version="@TOOL_VERSION@+galaxy0" profile="@PROFILE@">
     <description>SNP/indel variant calling from VCF/BCF</description>
     <macros>
         <token name="@EXECUTABLE@">call</token>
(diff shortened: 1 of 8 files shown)


nsoranzo (approved):

nsoranzo:
Still 4 tests failing unfortunately.

bernt-matthias:
https://github.com/galaxyproject/tools-iuc/pull/3945

bgruening:
@SaimMomin12 probably useful for the pangenoms

bgruening:
There are a lot of useful open PR and issues that we should address here as well. E.g. the vcf.gz for inputs.

wm75:
The failing `norm` test looks like it's uncovering a bug in recent bcftools versions (https://github.com/samtools/bcftools/issues/2476).

wm75:
The `mendelian` plugin was replaced with a new `mendelian2` plugin in version 1.17, which explains another two failing tests.

There have also been heavy changes around bcftool consensus over the last couple of versions, but I still need to understand all the details.

wm75:
Ok, this might finally be in a state that we can merge. Things that we may still want to address (but perhaps in separate PRs) are:
- https://github.com/samtools/bcftools/issues/2476 is still open, but would be very complicated to fix on the Galaxy side. I simply rewrote the test for now to get it to pass.
- there are likely still plenty of subcommand options that have been added or extended since v1.15 and that are not handled by this update
- The help on nearly all subcommands could be better

My suggestion would be that with every update from now on, we should at least check one subcommand and see if we're missing any options for it or if we can improve its help in any way.

tools/bcftools/bcftools_call.xml
```diff
@@ -22,25 +19,23 @@
             #if str($section.genotypes.novel_rate_ins):
               #silent $novel_rate.append(str($section.genotypes.novel_rate_ins))
             #end if
-            #if len($novel_rate) > 0:
+            #if len($novel_rate) &gt; 0:
```
bgruening:
```suggestion
            #if len($novel_rate) > 0:
```

tools/bcftools/bcftools_norm.xml
```diff
@@ -69,34 +61,30 @@ $atomize
...
-        <expand macro="macro_fasta_ref" />
-        <param name="check_ref" type="select" display="radio"
-        label="When any REF allele does not match the reference genome base"
-        help="Warnings about REF mismatches will be emitted to the standard
-        error (stderr) stream, and it is recommended to check there for
-        problems if you choose not to exit with an error immediately upon
-        encountering a mismatch.">
+        <expand macro="macro_input"/>
+        <expand macro="macro_fasta_ref"/>
+        <param name="check_ref" type="select" display="radio" label="When any REF allele does not match the reference genome base" help="Warnings about REF mismatches will be emitted to the standard         error (stderr) stream, and it is recommended to check there for         problems if you choose not to exit with an error immediately upon         encountering a mismatch.">
```
wm75:
```suggestion
        <param name="check_ref" type="select" display="radio" label="When any REF allele does not match the reference genome base" help="Warnings about REF mismatches will be emitted to the standard error (stderr) stream, and it is recommended to check there for problems if you choose not to exit with an error immediately upon encountering a mismatch.">
```

bgruening:
+1 from my side!

mvdbeek:
Deployment status: **skipped**

https://github.com/galaxyproject/tools-iuc/actions/runs/19828830794

mvdbeek:
Deployment status: **skipped**

https://github.com/galaxyproject/tools-iuc/actions/runs/19828830794

wm75:
Weird, it's specifically all the plugin tests that are failing now, while this has never happened in the PR CI.

wm75:
Not so weird after all, just all plugins have alphabetically adjacent names so run one after another, and their was a conda connection issue :-)

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/19828830794

PR #6307, opened 2024-09-05
Update ncbi fcs gx

richard-burhans:
Updating NCBI FCS GX.. Removed the ncbi_fcs_gx_config table and added those fields to the ncbi_fcs_gx_databases table.  Made sure both the tool and the data manager are using version 0.5.4.

FOR CONTRIBUTOR:


data_managers/data_manager_ncbi_fcs_gx_database_downloader/tool_data_table_conf.xml.test
```diff
@@ -1,7 +1,7 @@
 <tables>
     <!-- Locations of NCBI FCS GX databases -->
     <table name="ncbi_fcs_gx_databases" comment_char="#">
-        <columns>value, source_manifest, name</columns>
+        <columns>value, name, source_manifest, use_source_manifest, phone_home, phone_home_label, node_cache_dir, local_manifest</columns>
```
bgruening, line 4: We can not change the columns in a already existing data-table. I think you need to create a new one. e.g. `ncbi_fcs_gx_databases_ext`

tools/ncbi_fcs_gx/ncbi_fcs_gx.xml
```diff
@@ -59,9 +59,9 @@ GX_NUM_CORES=\${GALAXY_SLOTS:-2}
                 <option value="clean">Clean genome</option>
             </param>
             <when value="screen">
-                <!-- value, name, use_source_manifest, phone_home, phone_home_label, node_cache_dir -->
+                <!-- value, description, source_manifest, use_source_manifest, phone_home, phone_home_label, node_cache_dir, local_manifest -->
                 <param name="config_tag" type="select" label="Database">
-                    <options from_data_table="ncbi_fcs_gx_config">
+                    <options from_data_table="ncbi_fcs_gx_databases">
                         <filter type="sort_by" name="sorted_description" column="1"/>
```
bernt-matthias, line 65: `name` should not be allowed. Is there a linter complaining?

data_managers/data_manager_ncbi_fcs_gx_database_downloader/data_manager/data_manager_ncbi_fcs_gx_database_downloader.xml
```diff
@@ -7,22 +7,40 @@
...
     ]]></command>
     <inputs>
         <param name="tag" type="text" label="Unique identifier for this database"/>
-        <param name="source_manifest" type="text" label="Source Manifest"/>
+        <param name="description" type="text" label="Description for this database"/>
+        <param name="source_manifest" type="text" label="Manifest file for this database"/>
+        <param name="use_source_manifest" type="boolean" label="Should the tool use the source manifest"/>
+        <param name="phone_home" type="boolean" label="Should phone home be enabled"/>
+        <param name="phone_home_label" type="text" label="Phone home label"/>
+        <param name="node_cache_dir" type="text" label="Directory to copy database to local node"/>
```
bernt-matthias, line 31: Can you explain what this is doing? I can not see that anything is copied in the data manager.

data_managers/data_manager_ncbi_fcs_gx_database_downloader/data_manager/data_manager_ncbi_fcs_gx_database_downloader.xml
```diff
@@ -7,22 +7,40 @@
...
+    --output_file '$output_file'
+    --output_dir '$output_file.extra_files_path'
     ]]></command>
     <inputs>
         <param name="tag" type="text" label="Unique identifier for this database"/>
-        <param name="source_manifest" type="text" label="Source Manifest"/>
+        <param name="description" type="text" label="Description for this database"/>
+        <param name="source_manifest" type="text" label="Manifest file for this database"/>
+        <param name="use_source_manifest" type="boolean" label="Should the tool use the source manifest"/>
+        <param name="phone_home" type="boolean" label="Should phone home be enabled"/>
```
bernt-matthias, line 29: The `phone*` entries seem to be unused in the tool. Do we need them? What are they good for?

tools/ncbi_fcs_gx/ncbi_fcs_gx.xml
richard-burhans: No linter is complaining about the `name="sorted_description"`  in ` <filter type="sort_by" name="sorted_description" column="1"/>`.  I'll remove `name="sorted_description"` from the `filter`.  FYI: It's also found in the following:

```
tools/dimet/macros.xml
tools/mothur/get.lineage.xml
tools/mothur/remove.lineage.xml
```

data_managers/data_manager_ncbi_fcs_gx_database_downloader/tool_data_table_conf.xml.test
richard-burhans: Thanks!  I'll rename it.

data_managers/data_manager_ncbi_fcs_gx_database_downloader/data_manager/data_manager_ncbi_fcs_gx_database_downloader.xml
richard-burhans: I've added some documentation to the Help section that hopefully clears some of this up.  The Galaxy Admin chooses a manifest file (for example: https://ftp.ncbi.nlm.nih.gov/genomes/TOOLS/FCS/database/latest/all.manifest).  The data manager then uses [sync_files.py](https://github.com/ncbi/fcs-gx/blob/release/scripts/sync_files.py)  from the [ncbi-fcs-gx](http://bioconda.github.io/recipes/ncbi-fcs-gx/README.html) bioconda package to download a local copy  of the database described in the manifest.  It then adds the location of the local manifest file to the `ncbi_fcs_gx_databases_ext` table.

The data manager then parses the downloaded [taxa.tsv](https://ftp.ncbi.nlm.nih.gov/genomes/TOOLS/FCS/database/latest/all.taxa.tsv) file to create a lookup table mapping GX divisions to descriptions that will be seen and selected by the user.

data_managers/data_manager_ncbi_fcs_gx_database_downloader/data_manager/data_manager_ncbi_fcs_gx_database_downloader.xml
richard-burhans: The NCBI FCS GX tool uses [run_gx.py](https://github.com/ncbi/fcs-gx/blob/release/scripts/run_gx.py) which includes [code to send analytics back to NCBI](https://github.com/ncbi/fcs-gx/blob/release/scripts/run_gx.py#L79-L115).  The `phone_home` entry determines if we are sending analytics to NCBI.  The `phone_home_label` entry is a label/identifier sent along with the analytics to NCBI to aid with aggregation.

data_managers/data_manager_ncbi_fcs_gx_database_downloader/data_manager/data_manager_ncbi_fcs_gx_database_downloader.xml
bgruening: This parameter is optional? Its an implementation details, a TMP_DIR? In that case, I would not even expose it.

The problem I see is that we put a path in a config-file that might be node/cluster dependent. But we have heterogenous clusters most of the time. The path might be destination dependent. Usually, we work with those use-cases by setting $TEMP or related envs via Galaxy for a particular tool.

My question: Do we really need this parameter and the value set in the loc file, or can we just go with the default value (assuming that is some TMP dir).

bgruening:
@bernt-matthias can you have a look at this one? It LGTM.

data_managers/data_manager_ncbi_fcs_gx_database_downloader/data_manager/data_manager_ncbi_fcs_gx_database_downloader.xml
richard-burhans: I think we can  use TMPDIR for this.  It needs to be fast, ideally pinned in memory and always present.  At usegalaxy.org we'll set this to /dev/shm.

data_managers/data_manager_ncbi_fcs_gx_database_downloader/tool-data/ncbi_fcs_gx_databases_ext.loc.sample
```diff
@@ -0,0 +1,3 @@
+#tag|description|source_manifest|use_source_manifest|phone_home|phone_home_label|local_manifest
+#latest|Full GX Database|https://ftp.ncbi.nlm.nih.gov/genomes/TOOLS/FCS/database/latest/all.manifest|0|1|usegalaxy.org|/big/data/dir/ncbi_fcs_gx_databases_ext/latest/all.manifest
+#test|Test GX Database|https://ftp.ncbi.nlm.nih.gov/genomes/TOOLS/FCS/database/test-only/test-only.manifest|0|1|usegalaxy.org|/big/data/dir/ncbi_fcs_gx_databases_ext/test/test-only.manifest
```
bgruening, line 3: why are here | this should all be tabs, isn't it? I would also appreciate some more comments about what phone_home and phone_home_label means here. A few admins might look at the sample file.

tools/ncbi_fcs_gx/ncbi_fcs_gx.xml
```diff
@@ -6,17 +6,20 @@
...
     #end if
-    mkdir -p '$mode.config_tag.fields.node_cache_dir' &&
-    sync_files.py get --mft '$manifest_pathname' --dir '$mode.config_tag.fields.node_cache_dir' > /dev/null 2>&1 &&
+    ## skip checksumming the database if the db_state_file exists
+    if [ ! -e '\${TMPDIR}/sync-files-completed.txt' ]; then
+        ## copy data to local storage
+        mkdir -p '\${TMPDIR}' &&
+        sync_files.py get --mft '$manifest_pathname' --dir '\${TMPDIR}' > /dev/null 2>&1 &&
+        touch '\${TMPDIR}/sync-files-completed.txt';
+    fi &&
```
bgruening, line 20: Its late here, maybe I'm missing something. But if the `if` condition does not hold true, then the `&&` is the first command, isn't it? And this would crash?

data_managers/data_manager_ncbi_fcs_gx_database_downloader/tool-data/ncbi_fcs_gx_databases_ext.loc.sample
richard-burhans: I've replaced the pipes with tabs and added a comment explaining the phone_home and phone_home_label fields.

tools/ncbi_fcs_gx/ncbi_fcs_gx.xml
richard-burhans: The `if ... fi` is the first "command" (notice we have `if` and not `#if`) and `run_gx.py `is the second command.

I did find an interesting bug or a misunderstanding I have with bash. The code `if [ 1 -eq 1 ]; then echo "one still equals one"; fi` works as expected.  What happens when you prefix the `if` with an environment variable like this: `ONE=1 if [ 1 -eq 1 ]; then echo "one still equals one"; fi`?  As it turns out, this is what you get: `bash: syntax error near unexpected token 'then'`.  The following works but seems clunky:  `ONE=1 bash -c 'if [ 1 -eq 1 ]; then echo "one still equals one"; fi'`

bgruening (approved):
Thanks @richard-burhans

PR #6308, opened 2024-09-06
Update to Virannot 0.1.0

marieBvr:
FOR CONTRIBUTOR:

Fix BLASTXML parsing.
OTU module create a fasta file for each pfam motif.


tools/virAnnot/virAnnot_otu.xml
```diff
@@ -73,9 +73,9 @@
     </outputs>
     <tests>
         <test>
-            <param name="blast_files" value="otu_s1_tblastx.tab,otu_s2_tblastx.tab"/>
```
bgruening, line 76: can those old files be removed?

tools/virAnnot/otu.py
```diff
@@ -355,7 +360,9 @@ def _get_stats(options, hits_collection):
         worksheet.write(row, column + 2, 'contigs_list')
         row = 1
         # column = 0
+        f_nucc = open(file_fasta_nucc, "w+")
```
bgruening, line 363: You can use `with open(file_fasta_nucc, "w+") as f_nucc` here. Its the new recommended way in python. This way you also don't need to close() the handle.

tools/virAnnot/virAnnot_otu.xml
marieBvr: You are right, it is done.

tools/virAnnot/otu.py
marieBvr: The change has been made, for this line and another one.
Thanks.

bgruening:
Thanks @marieBvr

PR #6337, opened 2024-09-17
Update gtdb-tk

paulzierep:
FOR CONTRIBUTOR:

`column="2"` starts counting on 0
with the new `loc.test` file, you can test this behavior and the regex when serving locally


bernt-matthias:
Few points: 

- we need to bump
- do you think it would be a good idea to use the column name instead (should work according to the docs). I think it would be better.
- Have you tested this locally?

paulzierep:
Column name works. I have tested it locally in the sense, that serving the tool with planemo allows selecting the test DB. But I did not test with a full DB, since I do not have the space on my disc :)

bgruening (approved):

PR #6343, opened 2024-09-20
Update name2taxid wrapper

SantaMcCloud:
FOR CONTRIBUTOR:


SantaMcCloud:
I change the select column from data_ref to integer, since it can happen that you have multiple files as input and all files share the same column as input. With data_ref it can not work since my workflow generate a collection which stop it there since it has now data as reference for the column or at least it still throws an error if you try to use it this way.

SantaMcCloud:
@bgruening could you merge this quick to have this in the bot update scope?

bgruening:
Same here, please explain us what is the motivation and what is the error that you are seeing. Maybe share the workflow/history.

SantaMcCloud:
I found a way to handle this error here, so I will close the PR since it is not needed anymore

PR #6350, opened 2024-09-24
Enhance seqtk_mergefa: FASTA-only output and improved ambiguity handling

dianichj:
*Bug Fix*
Tool merges FASTA/Q files into a FASTA output and considers the quality threshold for FASTQ files when merging.

1. Clarified the -m option to handle ambiguous bases and conflicts (e.g., N and other IUPAC codes).
2. Improved help documentation with clearer examples and explanations.
3. Refined input parameter labels for better clarity and consistency.

FOR CONTRIBUTOR:
* This PR adds a new tool or tool collection
* This PR does something else (explain below)


wm75:
Sorry @dianichj if our offline discussion of this wasn't clear enough, but that's not what the change was meant to look like.
The tool *can* merge fastq inputs - it's just that the result will be fasta in all cases. So instead of restricting the input formats, you should fix the possible output formats.

dianichj:
> Sorry @dianichj if our offline discussion of this wasn't clear enough, but that's not what the change was meant to look like. The tool _can_ merge fastq inputs - it's just that the result will be fasta in all cases. So instead of restricting the input formats, you should fix the possible output formats.

Pavan and I are testing if indeed the tool is able to generate FASTQ Files from the original tool documentation. So far we have encountered some issues. We will be able to confirm later on if the tool is for both FASTA and FASTQ or only FASTA =)!

PR #6358, opened 2024-09-25
Improve idr_download to support download of the full hyperstack

lldelisle:
FOR CONTRIBUTOR:
* [X I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.

I needed to add a new dependency as pylibtiff does not allow to write hyperstacks.
I used the asserts for images and the image_diff as much as I could.

### DIFF
--- a/tools/idr_download/idr_download_by_ids.py
+++ b/tools/idr_download/idr_download_by_ids.py
@@ -10,8 +10,8 @@ from tempfile import TemporaryDirectory
 import numpy
 from libtiff import TIFF
 from omero.cli import cli_login
-from omero.constants.namespaces import NSBULKANNOTATIONS  # noqa
-from omero.gateway import BlitzGateway  # noqa
+from omero.constants.namespaces import NSBULKANNOTATIONS
+from omero.gateway import BlitzGateway
 from tifffile import imwrite
 
 
--- a/tools/idr_download/idr_download_by_ids.py
+++ b/tools/idr_download/idr_download_by_ids.py
@@ -10,7 +10,7 @@ from tempfile import TemporaryDirectory
 import numpy
 from omero.cli import cli_login
 from omero.constants.namespaces import NSBULKANNOTATIONS
-from omero.gateway import BlitzGateway, _ImageWrapper
+from omero.gateway import _ImageWrapper, BlitzGateway
 from tifffile import imwrite
 
 
--- a/tools/idr_download/idr_download_by_ids.py
+++ b/tools/idr_download/idr_download_by_ids.py
@@ -10,11 +10,11 @@ from tempfile import TemporaryDirectory
 import numpy
 from omero.cli import cli_login
 from omero.constants.namespaces import NSBULKANNOTATIONS
-from omero.gateway import BlitzGateway
+from omero.gateway import BlitzGateway, _ImageWrapper
 from tifffile import imwrite
 
 
-def warn(message, image_identifier, warn_skip=False):
+def warn(message: str, image_identifier: str, warn_skip: bool = False) -> None:
     """Print an error `message` to stderr and
     - prefix with the `image_identifier`
     - suffix with 'Skipping download!' if `warn_skip` is True
@@ -38,11 +38,11 @@ def warn(message, image_identifier, warn_skip=False):
     )
 
 
-def find_channel_index(image, channel_name):
+def find_channel_index(image: _ImageWrapper, channel_name: str) -> int:
     """Identify the channel index from the `image` and the `channel_name`
 
     Args:
-        image (omero.gateway._ImageWrapper): image wrapper on which the channel should be identified
+        image (_ImageWrapper): image wrapper on which the channel should be identified
         channel_name (string): name of the channel to look for
 
     Returns:
@@ -65,11 +65,11 @@ def find_channel_index(image, channel_name):
     return -1
 
 
-def get_clipping_region(image, x, y, w, h):
+def get_clipping_region(image: _ImageWrapper, x: int, y: int, w: int, h: int) -> list[int]:
     """Check `x`, `y` and adjust `w`, `h` to image size to be able to crop the `image` with these coordinates
 
     Args:
-        image (omero.gateway._ImageWrapper): image wrapper on which region want to be cropped
+        image (_ImageWrapper): image wrapper on which region want to be cropped
         x (int): left x coordinate
         y (int): top y coordinate
         w (int): width
@@ -80,7 +80,7 @@ def get_clipping_region(image, x, y, w, h):
         ValueError: if the x or y coordinates are larger than the width or height of the image.
 
     Returns:
-        list of int: new x, y, width, height adjusted to the image
+        list[int]: new [x, y, width, height] adjusted to the image
     """
     # If the (x, y) coordinate falls outside the image boundaries, we
     # cannot just shift it because that would render the meaning of
@@ -108,11 +108,11 @@ def get_clipping_region(image, x, y, w, h):
     return [x, y, w, h]
 
 
-def confine_plane(image, z):
+def confine_plane(image: _ImageWrapper, z: int) -> int:
     """Adjust/Confine `z` to be among the possible z for the `image`
 
     Args:
-        image (omero.gateway._ImageWrapper): image wrapper for which the z is adjusted
+        image (_ImageWrapper): image wrapper for which the z is adjusted
         z (int): plane index that need to be confined
 
     Returns:
@@ -127,11 +127,11 @@ def confine_plane(image, z):
     return z
 
 
-def confine_frame(image, t):
+def confine_frame(image: _ImageWrapper, t: int) -> int:
     """Adjust/Confine `t` to be among the possible t for the `image`
 
     Args:
-        image (omero.gateway._ImageWrapper): image wrapper for which the t is adjusted
+        image (_ImageWrapper): image wrapper for which the t is adjusted
         t (int): frame index that need to be confined
 
     Returns:
@@ -146,18 +146,18 @@ def confine_frame(image, t):
     return t
 
 
-def get_image_array(image, tile, z, c, t):
+def get_image_array(image: _ImageWrapper, tile: list[int], z: int, c: int, t: int) -> numpy.ndarray:
     """Get a 2D numpy array from an `image` wrapper for a given `tile`, `z`, `c`, `t`
 
     Args:
-        image (omero.gateway._ImageWrapper): image wrapper from which values are taken
-        tile (list of int): [x, y, width, height] where x,y is the top left coordinate of the region to crop
+        image (_ImageWrapper): image wrapper from which values are taken
+        tile (list[int]): [x, y, width, height] where x,y is the top left coordinate of the region to crop
         z (int): plane index
         c (int): channel index
         t (int): frame index
 
     Returns:
-        numpy array of 2 dimensions: image values of the selected area
+        numpy.ndarray: image values of the selected area (2 dimensions)
     """
     pixels = image.getPrimaryPixels()
     try:
@@ -170,14 +170,14 @@ def get_image_array(image, tile, z, c, t):
     return selection
 
 
-def get_full_image_array(image):
+def get_full_image_array(image: _ImageWrapper) -> numpy.ndarray:
     """Get a 5D numpy array with all values from an `image` wrapper
 
     Args:
-        image (omero.gateway._ImageWrapper): image wrapper from which values are taken
+        image (_ImageWrapper): image wrapper from which values are taken
 
     Returns:
-        numpy array of 5 dimensions: image values in the TZCYX order
+        numpy.ndarray: image values in the TZCYX order (5 dimensions)
     """
     # The goal is to get the image in TZCYX order
     pixels = image.getPrimaryPixels()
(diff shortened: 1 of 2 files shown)


tools/idr_download/idr_download_by_ids.py
```diff
@@ -215,33 +245,107 @@ def download_image_data(
...
+                        archive.add(fname, os.path.basename(fname))
+                        os.remove(fname)
+                except Exception as e:
+                    if skip_failed:
+                        # respect skip_failed on unexpected errors
+                        warn(str(e), image_warning_id, warn_skip=True)
+                        continue
+                    else:
+                        raise
+
```
kostrykin, line 294: There are no blank lines before the `if` block above and the `else` block below:
```suggestion
```

tools/idr_download/idr_download_by_ids.py
```diff
@@ -102,60 +100,94 @@ def get_image_array(image, tile, z, c, t):
...
-    coord=(0, 0), width=0, height=0, region_spec='rectangle',
-    skip_failed=False, download_tar=False, omero_host='idr.openmicroscopy.org', omero_secured=False, config_file=None
+    download_full=False,
+    channel=None,
+    z_stack=0,
+    frame=0,
+    coord=(0, 0),
+    width=0,
+    height=0,
+    region_spec="rectangle",
```
kostrykin, line 148: I think `region_spec` is ignored if `download_original` or `download_full`, is that intended?

tools/idr_download/idr_download_by_ids.py
```diff
@@ -400,77 +463,82 @@ def _center_to_ul(center_x, center_y, width, height):
...
     p.add_argument(
-        '-c', '--channel',
-        help='name of the channel to retrieve data for '
-             '(note: the first channel of each image will be downloaded if '
-             'left unspecified)'
+        "--download-full",
+        dest="download_full",
+        action="store_true",
+        help="download the full image on omero",
+    )
```
kostrykin, line 484: Shouldn't `--download-original` and `--download-full` be mutually exclusive?

If so, the interface of `download_image_data` ideally should enforce the mutuality.

tools/idr_download/idr_download_by_ids.py
```diff
@@ -102,60 +100,94 @@ def get_image_array(image, tile, z, c, t):
...
+    coord=(0, 0),
+    width=0,
+    height=0,
+    region_spec="rectangle",
+    skip_failed=False,
+    download_tar=False,
+    omero_host="idr.openmicroscopy.org",
+    omero_secured=False,
+    config_file=None,
 ):
```
kostrykin, line 154: I think that due to the length of this function, a docstring describing the individual parameters would be very helpful.

tools/idr_download/idr_download_by_ids.xml
```diff
@@ -19,8 +19,9 @@
         <edam_operation>operation_3443</edam_operation>
     </edam_operations>
     <requirements>
-        <requirement type="package" version="5.10.1">omero-py</requirement>
-        <requirement type="package" version="0.4.2">pylibtiff</requirement>
+        <requirement type="package" version="5.19.4">omero-py</requirement>
+        <requirement type="package" version="0.6.1">pylibtiff</requirement>
```
kostrykin, line 23: Is `pylibtiff` required? The module is not used in the wrapper code, neither is it a [direct dependency of omero-py](https://github.com/ome/omero-py/blob/4fd72efaa8cdbec9da98e1d28678af16878a86ba/setup.py#L208), could be a transitive dependency though.

tools/idr_download/idr_download_by_ids.xml
```diff
@@ -135,9 +138,11 @@
         <conditional name="image_region">
             <param name="original" type="select" label="Which images do you want to download?">
                 <option value="TIFF" selected="true">Exported TIFF (single channel, single stack)</option>
+                <option value="full" >Whole TIFF (all channels, all frames, all stacks)</option>
```
kostrykin, line 141:
```suggestion
                <option value="full">Whole TIFF (all channels, all frames, all stacks)</option>
```

kostrykin (commented):
Much thanks, a few comments inside!

tools/idr_download/idr_download_by_ids.py
lldelisle: Yes, it is intended. If you download the original file you cannot crop it because you may not be able to open it. If you download full, we could allow to crop. I thought, people could crop with another tool but if you think it is better to be able to crop in this tool, I can implement it.

tools/idr_download/idr_download_by_ids.py
lldelisle: Indeed

tools/idr_download/idr_download_by_ids.py
lldelisle: Good point, I will write the args differently.

tools/idr_download/idr_download_by_ids.xml
lldelisle: It was used in the original version but I agree we could get rid of it.

tools/idr_download/idr_download_by_ids.py
```diff
@@ -44,7 +42,7 @@ def find_channel_index(image, channel_name):
             if p[0] == "Channels":
                 channels = p[1].replace(" ", "").split(";")
```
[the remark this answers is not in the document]
kostrykin: >  I don't know if the issue is the galaxywrapper that do not accept parameters with space...
Actually this shouldn't be a problem, as long as the parameters are escaped with quotes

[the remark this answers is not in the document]
lldelisle: In fact this block is only used if the channel name cannot be found. The first part accepts spaces:
```
    channel_name = channel_name.lower()
    for n, channel in enumerate(image.getChannelLabels()):
        if channel_name == channel.lower():
            return n
```

tools/idr_download/idr_download_by_ids.py
```diff
@@ -44,7 +42,7 @@ def find_channel_index(image, channel_name):
...
                 for n, c in enumerate(channels):
-                    for value in c.split(':'):
+                    for value in c.split(":"):
                         if channel_name == value.lower():
```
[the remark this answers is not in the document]
lldelisle: If fact above you have `channel_name = channel_name.lower()`

tools/idr_download/idr_download_by_ids.py
```diff
@@ -102,60 +100,94 @@ def get_image_array(image, tile, z, c, t):
...
 
+def get_full_image_array(image):
+    # The goal is to get the image in TZCYX order
+    pixels = image.getPrimaryPixels()
+    tzclist = list(
+        product(
+            range(image.getSizeT()), range(image.getSizeZ()), range(image.getSizeC())
+        )
+    )
+    zctlist = [(z, c, t) for (t, z, c) in tzclist]
```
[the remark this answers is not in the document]
kostrykin: Just to be clear that nothing unexpected happens here.

You feed the `zctlist` into `pixels.getPlanes`, where `zctlist` is structured like
```
[(0, 0, 0),
 (0, 1, 0),
 (0, 2, 0),
 ...
 (1, 0, 0),
 (1, 1, 0),
 (1, 2, 0),
 ...
 (2, 0, 0),
 (2, 1, 0),
 (2, 2, 0),
 ...
 (0, 0, 1),
 (0, 1, 1),
 (0, 2, 1),
 ...
 (1, 0, 1),
 (1, 1, 1),
 (1, 2, 1),
 ...]
```
Is this really intended? Because to me it naively looks like the slices that you get will be in the wrong order.

tools/idr_download/idr_download_by_ids.py
kostrykin: I think it's fine as it is, maybe just add a comment, that the argument is ignored in specific cases.

lldelisle: OK

tools/idr_download/idr_download_by_ids.py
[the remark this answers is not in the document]
kostrykin: You're right, I missed that!

tools/idr_download/idr_download_by_ids.py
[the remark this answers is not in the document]
kostrykin: Yes, but the second block still searches for the channel name. Although I don't grasp what exactly the code does, I see that the code path will diverge if the channel name contains whitespaces. Wondering whether this is intended.

tools/idr_download/idr_download_by_ids.py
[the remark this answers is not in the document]
lldelisle: OK I will recheck carefully and come back to you.

tools/idr_download/idr_download_by_ids.py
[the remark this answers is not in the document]
lldelisle: I don't know. I would not touch this.

tools/idr_download/idr_download_by_ids.py
lldelisle: I added a dockstring. Is it fine?

tools/idr_download/idr_download_by_ids.py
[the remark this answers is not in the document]
kostrykin: I pinged the authors of this code section in https://github.com/lldelisle/tools-iuc/commit/81da9b4e7b2324e8f4cfec152171312b7a2b1226, maybe they still remember.

I see that this issue is rather tangential to this PR, so I'm fine with keeping it as-is for the moment.

tools/idr_download/idr_download_by_ids.py
kostrykin: Thanks! Looks overly good to me.

I'd suggest to stick to some standardized formatting though, so that the docstrings get rendered nicely when hovering functions in VS code or other IDEs and so on.

lldelisle: I used a VS code extension (https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring) and for me it renders nicely but tell me which standards I should follow or how could I improve the docstrings.

tools/idr_download/idr_download_by_ids.py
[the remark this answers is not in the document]
lldelisle: I checked and tested with an image with multiple C, Z and T and my reasoning is correct.
The zctlist means:
```
z=0,c=0,t=0
z=0,c=1,t=0
z=0,c=2,t=0
...
z=1,c=0,t=0
z=1,c=1,t=0
z=1,c=2,t=0
...
z=2,c=0,t=0
z=2,c=1,t=0
z=2,c=2,t=0
...
z=0,c=0,t=1
z=0,c=1,t=1
z=0,c=2,t=1
...
z=1,c=0,t=1
z=1,c=1,t=1
z=1,c=2,t=1
...
z=2,c=0,t=1
z=2,c=1,t=1
z=2,c=2,t=1
...
z=0,c=0,t=2
z=0,c=1,t=2
z=0,c=2,t=2
...
z=1,c=0,t=2
z=1,c=1,t=2
z=1,c=2,t=2
...
z=2,c=0,t=2
z=2,c=1,t=2
z=2,c=2,t=2
...
```
This order is: all t=0 first, then all t=1, then all t=2... and for a given t, all z=0, then all z=1 etc...
Then when I reshape with numpy it gives:
```

[[['z=0,c=0,t=0' 'z=0,c=1,t=0' 'z=0,c=2,t=0' ...]
  ['z=1,c=0,t=0' 'z=1,c=1,t=0' 'z=1,c=2,t=0' ...]
  ['z=2,c=0,t=0' 'z=2,c=1,t=0' 'z=2,c=2,t=0' ...]
  ...]

 [['z=0,c=0,t=1' 'z=0,c=1,t=1' 'z=0,c=2,t=1'...]
  ['z=1,c=0,t=1' 'z=1,c=1,t=1' 'z=1,c=2,t=1'...]
  ['z=2,c=0,t=1' 'z=2,c=1,t=1' 'z=2,c=2,t=1'...]
  ...]

 [['z=0,c=0,t=2' 'z=0,c=1,t=2' 'z=0,c=2,t=2'...]
  ['z=1,c=0,t=2' 'z=1,c=1,t=2' 'z=1,c=2,t=2'...]
  ['z=2,c=0,t=2' 'z=2,c=1,t=2' 'z=2,c=2,t=2'...]
  ...]
  
  ...]
```
Which is exactly the TZCYX order I wanted.

tools/idr_download/idr_download_by_ids.py
lldelisle: Apparently the default format is google. You want me to change it to [sphinx](https://github.com/NilsJPWerner/autoDocstring/blob/HEAD/docs/sphinx.md) or [sphinx in one line](https://github.com/NilsJPWerner/autoDocstring/blob/HEAD/docs/one-line-sphinx.md)?

kostrykin: I don't want to claim that this is a necessity, just a suggestion!

Like for example:

```python
    """Identify the channel index from the `image` and the `channel_name`.
    ...
    """
```

You see, the arguments get rendered differently then (monospace).

image

If you decided to render the whole documentation as a HTML document at some later time, you could do this using sphinx, and sphinx would render the arguments in italic. If you wanted it to be monospace, you would rather use

```python
    """Identify the channel index from the ``image`` and the ``channel_name``.
    ...
    """
```

but I wouldn't suggest that (this should rather be used for short code examples or code citations). This would also be rendered monospace by VS code.

lldelisle: image

tools/idr_download/idr_download_by_ids.py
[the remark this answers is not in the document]
kostrykin: Much thanks for the explanation, now it is clear to me! Looks right then.

lldelisle:
Thank you so much @kostrykin for the careful review. I think we improved the python script a lot for further development.

kostrykin (approved):

PR #6376, opened 2024-09-26
Add bwa-mem2 data manager

natefoo:
FOR CONTRIBUTOR:


bernt-matthias (commented):
Can I persuade you to drop the python wrapper? If there is non much logic involved datamanagers can (and maybe should) be much simpler. 

An example might be https://github.com/galaxyproject/tools-iuc/blob/main/data_managers/data_manager_metaphlan_database_downloader/data_manager/data_manager_metaphlan_download.xml

natefoo:
Thanks, I didn't know there were already examples that dropped the wrapper, I'll convert it.

data_managers/data_manager_bwa_mem2_index_builder/tool_data_table_conf.xml.sample
```diff
@@ -0,0 +1,8 @@
+<!-- Use the file tool_data_table_conf.xml.oldlocstyle if you don't want to update your loc files as changed in revision 4550:535d276c92bc-->
+<tables>
+    <!-- Locations of indexes in the BWA-MEM2 mapper format-->
+    <table name="bwa_mem2_indexes" comment_char="#">
+        <columns>value, dbkey, name, path</columns>
```
bernt-matthias, line 5: Maybe we want to include the bwa-mem2 version here (that has been used for the generation).

data_managers/data_manager_bwa_mem2_index_builder/data_manager/bwa_mem2_index_builder.xml
```diff
@@ -0,0 +1,75 @@
...
+#set $value = $sequence_id or $all_fasta_source.fields.dbkey
+#set $name = $sequence_name or $all_fasta_source.fields.name
+{
+  "data_tables":{
+    "bwa_mem2_indexes":[
+      {
+        "value": "${value}",
+        "dbkey": "${all_fasta_source.fields.dbkey}",
+        "name": "${name}",
+        "path": "${fasta_file_name}"
```
bernt-matthias, line 30: should this be `${out_file.extra_files_path}/${fasta_file_name}`?

In the DMs that I created recently I just crated a directory (e.g. `DIR`) in the working dir and used this. I guess using `out_file.extra_files_path` is not really necessary.

data_managers/data_manager_bwa_mem2_index_builder/tool_data_table_conf.xml.sample
natefoo: The index format has been stable since 2020 and the Galaxy version of the tool was added after that index format change, so it should be compatible with all versions. For tools that have changed it in the past (e.g. STAR) we added a new versioned data table once that happens, but I guess if we want to be safe I could add a version column, even if it's unused for now?

data_managers/data_manager_bwa_mem2_index_builder/data_manager/bwa_mem2_index_builder.xml
natefoo: Doing so seems to cause the index to be referenced from the object store after the DM completes: the path in the loc file points to `/path/to/galaxy/database/objects/c/0/f/dataset_c0ffee_files/fooBar1.fa`. Whereas if it is just the relative name inside the `extra_files_path`, then the index is referenced from `tool_data_dir` as expected for a DM. Possibly just an artifact of `os.path.join([tool_data_dir, intermediate_dirs, '/path/to/galaxy/database/objects/c/0/f/dataset_c0ffee_files/fooBar1.fa'])` clobbering the earlier path elements, but either way that's not the desired outcome.

data_managers/data_manager_bwa_mem2_index_builder/tool_data_table_conf.xml.sample
bgruening: I think we should add a version column to all DM. Its just more future-proof and does not cost anything.

natefoo: Makes sense, but wouldn’t adding a column make it incompatible with the existing bwa-mem2 tool’s data table?

bgruening: This is true. I rephrase it, ideally, we should have a version column. Lets get it in and hope it will never change.

natefoo: In that case I think this is ready to go.

bgruening (approved):

bernt-matthias:
Great. Thanks a lot for the contribution.

natefoo:
Thanks @bernt-matthias @bgruening!

PR #6380, opened 2024-09-28
Sniffles: Update VCF info field table in help and add --detect-large-ins flag option

fubar2:
FOR CONTRIBUTOR:

- Add an option to set --detect-large-ins flag false. 
- enforce str() conversion for select values
- Replace the old help VCF information table - now mostly out of date.
  - Fastidiously formatted the info fields from an output VCF (!) - such fragility, both human and restructured text...


bgruening:
This might help next time: https://rsted.info.ucl.ac.be/

Can you please bump the version?

fubar2:
>This might help next time: https://rsted.info.ucl.ac.be/

Nice - thanks! 
Useful but doesn't make a table which is what this needed. They are painful in html too but not as bad as restructured text. Patience and precision that can probably be made easier - and it only works when it is right.

bgruening (approved):

PR #6381, opened 2024-09-30
bigwig_outlier_bed: Option to find regions with zero coverage rather than rely on a low quantile

fubar2:
FOR CONTRIBUTOR:

Add option to create a bed track with a feature for each contiguous region with zeros for coverage. A low quantile setting will not work for this purpose - it needs to be zero for a stretch of sequence.

Trying to understand some large (200k+) regions in some VGP haplotype assemblies where one or both haplotypes have absolutely no coverage from Pacbio hifi reads. Possibly from bionano optical or other sources during curation, but it may not really exist? Seeing zero - absolutely no reads - mapping over substantial regions. 

Image below shows two haplotypes with coverage and windowmasker repeat bigwigs, from [bColStr4](https://usegalaxy.eu/datasets/4838ba20a6d867657c43f8441d416326/preview) SUPER_5:22,191,743..23,883,213 - the two haplotypes have slightly staggered regions where no pacbio reads seem to be mapping, suggesting two different, potentially bogus inserts of sequence?

image

### DIFF
--- a/tools/bigwig_outlier_bed/bigwig_outlier_bed.py
+++ b/tools/bigwig_outlier_bed/bigwig_outlier_bed.py
@@ -300,8 +300,8 @@ class findOut:
                 t.write("\n")
         some = False
         if self.outbeds in ["outzero"]:
-                self.writeBed(bedzero, self.bedoutzeo)
-                some = True
+            self.writeBed(bedzero, self.bedoutzero)
+            some = True
         if self.qlo:
             if self.outbeds in ["outall", "outlo", "outlohi"]:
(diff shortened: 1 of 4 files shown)


bgruening (approved):
Nice idea!

fubar2:
@bgruening: odd that windowmasker/repeatmasker (when it ran!) show zero repeats in these regions - took a look and these are long stretches of N in one or both haplotype assemblies. We can easily identify them this way - need the curators to take a look but that's why there are zero PacBio reads mapping there - so at least I'm sure it's not one of my bugs.

PR #6393, opened 2024-10-01
Ensure correct output formats of seqtk_seq and seqtk_mergefa in all cases

RZ9082:
FOR CONTRIBUTOR:

Adjust output format when using `-A        force FASTA output (discard quality)` to FASTA


wm75:
You need to bump the wrapper version. Maybe just this single one and don't change the macros one.

In addition, I think the fasta output of the tool would likely have to be `fasta.gz` in case the input is compressed (see the `@CONDITIONAL_GZIP_OUT@` macro).

If you have the time, you could also pull in the changes from #6350 and fix the output format there. Very similar issue as yours: the output of that tool should always be fasta or fasta.gz, depending on input compression, but never fastq(.gz).

RZ9082:
> You need to bump the wrapper version. Maybe just this single one and don't change the macros one.
> 
> In addition, I think the fasta output of the tool would likely have to be `fasta.gz` in case the input is compressed (see the `@CONDITIONAL_GZIP_OUT@` macro).
> 
> If you have the time, you could also pull in the changes from #6350 and fix the output format there. Very similar issue as yours: the output of that tool should always be fasta or fasta.gz, depending on input compression, but never fastq(.gz).

seqtk seq actually produces also FASTQ as an output !

wm75:
yes, I realize. What I wanted to say is that the logic for handling output formats will be similar.

RZ9082 (commented):
Does it look good now? @wm75

wm75 (approved):

PR #6417, opened 2024-10-05
Linter fixes colibread

bernt-matthias:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/colibread/lordec.xml
+++ b/tools/colibread/lordec.xml
@@ -1,10 +1,11 @@
-<tool id="lordec" name="Lordec" version="0.9+galaxy2" profile="@PROFILE@">
+<tool id="lordec" name="Lordec" version="@LORDEC_VERSION@+galaxy3" profile="@PROFILE@">
     <description>is a set a programs for correcting sequencing errors in PacBio reads</description>
     <macros>
+        <token name="@LORDEC_VERSION@">0.9</token>
         <import>macros.xml</import>
     </macros>
     <requirements>
-        <requirement type="package" version="0.9">lordec</requirement>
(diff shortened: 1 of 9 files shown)


tools/colibread/lordec.xml
```diff
@@ -1,10 +1,11 @@
-<tool id="lordec" name="Lordec" version="0.9+galaxy2" profile="@PROFILE@">
+<tool id="lordec" name="Lordec" version="@LORDEC_VERSION@+galaxy3" profile="@PROFILE@">
     <description>is a set a programs for correcting sequencing errors in PacBio reads</description>
     <macros>
+        <token name="@LORDEC_VERSION@">0.9</token>
```
bgruening, line 4: I'm not sure the bot can bump those? Should we name all of those TOOL_VERSION?

bernt-matthias: All tools in colibread use different requirements with different versions. Still the macro is good IMO because it can be used in the tool and requirement version.

bernt-matthias: Now I see what you mean. One moment

tools/colibread/macros.xml
```diff
@@ -1,11 +1,11 @@
 <macros>
-    <token name="@DISCOSNP_VERSION@">2.5.4</token>
+    <token name="@TOOL_VERSION@">2.5.4</token>
```
bgruening, line 2: I would move this out to the specific tools, since I'm not sure which definition is overwriting the double defined var. I can work on this if you like.

bgruening (approved):

PR #6440, opened 2024-10-07
Use multithreaded pigz for compression in meryl

natefoo:
On very large meryldb archives, compressing the tar stream with a single-threaded zlib is the overwhelming amount of tool runtime. With 56 cores pigz decreased the compression runtime of a 24 GB meryldb from an estimated 30 hours to 9 minutes.

I also added `m` to the untar options so that mod times are not preserved. This can be useful when working with old datasets on HPC systems that have aggressive scratch cleanup.

Tested on the command line but not in the tools so I am relying on tool tests to catch my mistakes. =)

FOR CONTRIBUTOR:


natefoo:
Hrm, I'm seeing file size differences as large as 3k while trying to generate updated output size values for the tests. I can increase the delta but I wonder if there's a better idea.

mvdbeek:
>  I can increase the delta but I wonder if there's a better idea.

That seems fine to me

bgruening (approved):

PR #6443, opened 2024-10-09
Update fastp.xml

nekrut:
Setting "report_json" to "True"

fastp should really replace fastqc for good. To do this we need to explain users that json report can be readily visualized with multiqc, but for that it needs to be enabled. 

Enabling it by default will eliminate the needs to tell users to rerun their fastp runs to generate the reports.

FOR CONTRIBUTOR:


bgruening:
folded into https://github.com/galaxyproject/tools-iuc/pull/6446

PR #6452, opened 2024-10-12
hicexplorer buildmatrix

fubar2:
Make bed file optional. It always was optional but if you had a bed in the history, it could not be removed.


fubar2:
Documentation is wrong. It's needed.

PR #6465, opened 2024-10-17
Fix and update kraken2 data manager

wm75:
FOR CONTRIBUTOR:

Bug fixes:
- ensures custom databases always have a valid filename as their path

Enhancements:
- more standardized naming of custom db records by asking for a name and for source info
- for all build modes: adjust default minimizer-spaces setting to match the upstream change in https://github.com/DerrickWood/kraken2/commit/f7268d1dacb6e5f1a2e5a775f49f5150b9e7cad2
- add bio.tools xref and update citation

Requirement update:
- use Python 3.13

### DIFF
--- a/data_managers/data_manager_build_kraken2_database/data_manager/kraken2_build_database.py
+++ b/data_managers/data_manager_build_kraken2_database/data_manager/kraken2_build_database.py
@@ -1,12 +1,11 @@
 #!/usr/bin/env python
 
-from __future__ import print_function
-
 import argparse
 import datetime
 import errno
 import json
 import os
+import re
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


data_managers/data_manager_build_kraken2_database/data_manager/kraken2_build_database.py
```diff
@@ -367,9 +387,9 @@ def kraken2_build_custom(kraken2_args, custom_database_name, target_directory, d
         'data_tables': {
             data_table_name: [
                 {
-                    "value": custom_database_name,
-                    "name": custom_database_name,
-                    "path": custom_database_name
+                    "value": database_value,
```
bgruening, line 390: The value is the path, correct? Why?

wm75: Well, maybe just because I'm not creative enough, but I think for both path and value you want to ensure something uniquely identifying the build and timestamp + name + algorithmic settings seemed to do that job and because kraken2_build_special() does things very similarly.

PR #6372, opened 2024-09-26
Updating tools/varvamp from version 1.2.0 to 1.2.1

gxydevbot:
Hello! This is an automated update of the following tool: **tools/varvamp**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

I have updated tools/varvamp from version 1.2.0 to 1.2.1.

**Project home page:** https://github.com/jonas-fuchs/varVAMP/releases

For any comments, queries or criticism about the bot, not related to the tool being updated in this PR, please create an issue [here](https://github.com/planemo-autoupdate/autoupdate/issues/new).

If you want to skip this change, close this PR without deleting the branch. It will be reopened if another change is detected.

Any commit from another author than 'planemo-autoupdate' will prevent more auto-updates.

To ignore manual changes and allow autoupdates, delete the branch.

PR #6486, opened 2024-10-24
Update varVAMP to 1.2.1

wm75:
FOR CONTRIBUTOR:

supersedes  #6372


wm75:
@jonas-fuchs
https://github.com/jonas-fuchs/varVAMP/blob/ac4b1edb0f51965bc1116673f5cd1dcd2f600722/varvamp/command.py#L508-L509

@bgruening is right, I think, in https://github.com/galaxyproject/tools-iuc/pull/6372#issuecomment-2395003051 that this seems weird.
The consensus outputs are a property of the msa, not of the primer scheme, and it
 doesn't look helpful that these two files now have identical titles.
Is there an advantage to doing this that I'm not seeing?

wm75:
Tests are now passing with the consensus output name adjusted, but lets wait for what Jonas says.

jonas-fuchs:
Hi Wolfgang,

so from my side this was indeed intended as the type of consensus is still referenced in the name of the file. 'varVAMP_consensus' is also just the default name which you can adjust with the '--name' argument to something like 'HEV_consensus', which I think would make it more easy if you would like to differentiate between primer schemes. Might have not been that smart. So open to suggestions. We could also adjust it to something like 'varVAMP_consensus | ambig'.

Let me know ;)
Jonas

wm75:
Ok, lets get this merged and think things over again before the next release :-)

PR #6489, opened 2024-10-25
Normalize paths for genome fetch and some of the genome indexer data managers, plus additional moderinzation

natefoo:
Update the genome fetch and most commonly used indexer DMs to normalize the on-disk layout as proposed in https://github.com/galaxyproject/galaxy/discussions/19013.

In addition:

- For those that had Python wrappers, I dropped the wrappers. In some cases this avoids building a special mulled container just for the DM
- Updated some underlying tool versions
- Added some tests of non-default options
- For the STAR DM, automatically calculate the `--genomeSAindexNbases` and `--genomeChrBinNbits` options as recommended by the manual, to drastically reduce the index size for small genomes.

FOR CONTRIBUTOR:


data_managers/data_manager_star_index_builder/test-data/rnastar_index2x_versioned.loc
bgruening, line 1: Does this file need to be here?

data_managers/data_manager_star_index_builder/data_manager/rna_star_index_builder.xml
```diff
@@ -1,15 +1,15 @@
-<tool id="rna_star_index_builder_data_manager" name="rnastar index versioned" tool_type="manage_data" version="@IDX_VERSION@+galaxy@IDX_VERSION_SUFFIX@" profile="19.05">
+<tool id="rna_star_index_builder_data_manager" name="rnastar index versioned" tool_type="manage_data" version="@IDX_VERSION@+galaxy@IDX_VERSION_SUFFIX@" profile="23.0">
```
bgruening, line 1:
```suggestion
<tool id="rna_star_index_builder_data_manager" name="RNAStar index versioned" tool_type="manage_data" version="@IDX_VERSION@+galaxy@IDX_VERSION_SUFFIX@" profile="23.0">
```

data_managers/data_manager_bwa_mem_index_builder/data_manager_conf.xml
```diff
@@ -10,13 +9,12 @@
                 <column name="path" output_ref="out_file" >
                     <move type="directory" relativize_symlinks="True">
                         <!-- <source>${path}</source>--> <!-- out_file.extra_files_path is used as base by default --> <!-- if no source, eg for type=directory, then refers to base -->
-                        <target base="${GALAXY_DATA_MANAGER_DATA_PATH}">${dbkey}/bwa_mem_index/${value}</target>
+                        <target base="${GALAXY_DATA_MANAGER_DATA_PATH}">genomes/${dbkey}/bwa_mem_index/v1/${value}</target>
```
bgruening, line 12: Where does the "v1" comes from?

data_managers/data_manager_bwa_mem2_index_builder/data_manager_conf.xml
```diff
@@ -9,9 +9,9 @@
                 <column name="path" output_ref="out_file" >
                     <move type="directory" relativize_symlinks="True">
                         <!-- <source>${path}</source>--> <!-- out_file.extra_files_path is used as base by default --> <!-- if no source, eg for type=directory, then refers to base -->
-                        <target base="${GALAXY_DATA_MANAGER_DATA_PATH}">${dbkey}/bwa_mem2_index/${value}</target>
+                        <target base="${GALAXY_DATA_MANAGER_DATA_PATH}">genomes/${dbkey}/bwa_mem_index/v2/${value}</target>
                     </move>
-                    <value_translation>${GALAXY_DATA_MANAGER_DATA_PATH}/${dbkey}/bwa_mem2_index/${value}/${path}</value_translation>
+                    <value_translation>${GALAXY_DATA_MANAGER_DATA_PATH}/genomes/${dbkey}/bwa_mem_index/v2/${value}/${path}</value_translation>
```
bgruening, line 14: Ah, now I see ... Not sure if the v1/v2 should be under bwa_mem_index. I would assume those are separate tools - separate indices.

bgruening: Or in oder words, as an admin, I would search for `bwa_mem2_index`

bgruening (commented):
Great work Nate!

data_managers/data_manager_star_index_builder/test-data/rnastar_index2x_versioned.loc
natefoo: Yes, this is just a rename from `rnastar_index2_versioned.loc`, which was inconsistent with its name in all other tables. An empty `.loc` as referenced in `tool_data_table_conf.xml.test` must exist in order to be written to by the test.

data_managers/data_manager_bwa_mem_index_builder/data_manager_conf.xml
natefoo: See the [proposal](https://github.com/galaxyproject/galaxy/discussions/19013), this is because it's version 1 of bwa-mem.

data_managers/data_manager_bwa_mem2_index_builder/data_manager_conf.xml
natefoo: I did consider that, I ended up with this because under the scheme in the proposal every DM will now contain a version directory. So this will result in:

- `genomes/${dbkey}/bwa_mem_index/v1/${value}`
- `genomes/${dbkey}/bwa_mem2_index/v2/${value}`

Which is kind of redundant and negates the purpose of the version directory in this case. That said I can understand how you would expect to have the directory name match the indexer name, although we already violate that for the other DMs, since the table name (`bowtie_indexes`) is plural and the tool ID/directory (`bowtie_index`) is not (to say nothing of the bowtie DMs using both "indexes" in the table/directory and "indices" in the loc file name).

natefoo:
Converted to draft:

1. I need to add the sam_fasta_index DM as well.
2. I propose we remove the symlink to the reference genome [as proposed here](https://github.com/galaxyproject/galaxy/discussions/19013#discussioncomment-11079023). @mvdbeek already tested this for one tool (bowtie2?).

data_managers/data_manager_bowtie2_index_builder/data_manager/bowtie2_index_builder.xml
```diff
@@ -1,34 +1,72 @@
...
+]]></configfile>
+    </configfiles>
     <inputs>
         <param name="all_fasta_source" type="select" label="Source FASTA Sequence">
             <options from_data_table="all_fasta"/>
         </param>
         <param name="sequence_name" type="text" value="" label="Name of sequence" />
         <param name="sequence_id" type="text" value="" label="ID for sequence" />
-        <param name="tophat2" type="boolean" truevalue="--data_table_name tophat2_indexes" falsevalue="" checked="True" label="Also make available for TopHat" help="Adds values to tophat2_indexes tool data table" />
+        <param name="tophat2" type="boolean" checked="True" label="Also make available for TopHat" help="Adds values to tophat2_indexes tool data table" />
```
bernt-matthias, line 54: Cool to cover this, but it seems SNAFU anyway, since the [tophat2 data table refers to the bowtie2 loc file]( https://github.com/galaxyproject/tools-devteam/blob/9f0b8eb9fd7cf9e2513e7c822767153dbd4bc3b2/tools/tophat2/tool_data_table_conf.xml.sample#L6). 

Maybe we just deprecate the tophat2 datatable (and update the tool to use the bowtie2 one)?

data_managers/data_manager_bowtie2_index_builder/data_manager_conf.xml
```diff
@@ -10,9 +9,9 @@
                 <column name="path" output_ref="out_file" >
                     <move type="directory" relativize_symlinks="True">
                         <!-- <source>${path}</source>--> <!-- out_file.extra_files_path is used as base by default --> <!-- if no source, eg for type=directory, then refers to base -->
-                        <target base="${GALAXY_DATA_MANAGER_DATA_PATH}">${dbkey}/bowtie2_index/${value}</target>
+                        <target base="${GALAXY_DATA_MANAGER_DATA_PATH}">genomes/${dbkey}/bowtie_index/v2/${value}</target>
                     </move>
-                    <value_translation>${GALAXY_DATA_MANAGER_DATA_PATH}/${dbkey}/bowtie2_index/${value}/${path}</value_translation>
+                    <value_translation>${GALAXY_DATA_MANAGER_DATA_PATH}/genomes/${dbkey}/bowtie_index/v2/${value}/${path}</value_translation>
```
bernt-matthias, line 14: I do not understand why this is done? It will require all admins to restucture reference data, or?
But certainly this would be nicer if we would start from scratch.

data_managers/data_manager_bowtie_index_builder/data_manager/bowtie_color_space_index_builder.xml
```diff
@@ -1,34 +0,0 @@
-<tool id="bowtie_color_space_index_builder_data_manager" name="Bowtie Color index" tool_type="manage_data" version="1.2.1" profile="23.0">
```
bernt-matthias, line 1: We should move these to the `deprecated/data_managers/` folder of this repo.

bernt-matthias (approved):
Really appreciate that we have more and more DMs that do not require the extra python stuff. 

I'm a bit worried about the side effects for admins.

data_managers/data_manager_bowtie2_index_builder/data_manager_conf.xml
natefoo: They don't have to - if you update to the new DMs and run them, they will just place data in a new folder, but the old loc with the old data at the old paths will continue to be loaded. That said, in the proposal I suggested that we recommend admins to specify a new `tool_data_path` just for organizational purposes. I also said I'd write a script to restructure the data for anyone who preferred to unify it.

data_managers/data_manager_bowtie2_index_builder/data_manager/bowtie2_index_builder.xml
natefoo: Yes, this is probably just legacy. Good idea to update the tophat2 tool although I suppose in the unlikely case an admin has indexes in the `tophat2_indexes` table that are not in `bowtie2_indexes` then they would disappear. And tophat is of course deprecated itself.

data_managers/data_manager_bwa_mem2_index_builder/data_manager_conf.xml
natefoo: I don't want to push this if others feel it isn't a good design choice (since the design change is ultimately the motivation for this PR), so if you think it makes sense to change this I'll do so. I am not sure what the version numbers ought to be in this case - probably disconnected from the version of the underlying tool and always `v1` until another version is needed?

data_managers/data_manager_bowtie_index_builder/data_manager/bowtie_color_space_index_builder.xml
natefoo: Done

bgruening (approved):
Very cool!

PR #6493, opened 2024-10-28
Fix 2 bugs in SRST2

### LINKED ISSUE #6459: SRST2 has incorrect format on discover datasets outputs.
```
$ ag discover_datasets.*format.*,
tools/srst2/srst2.xml
200:            <discover_datasets pattern="(?P&lt;designation&gt;.+)" directory="geneTypingOutput" format="txt,tabular"/>
```

srst2 is the only tool to use a format with a `,` which seems incorrect, as the file gets marked 'binary or unknown' in galaxy even though it's a plain old tabular file. 

For me the results are tabular but I'm not very familiar with this tool so I'm guessing that `txt` was there for a reason.

@bernt-matthias perhaps this is something that should be linted against? The XSD does not have a problem with a comma there (or maybe the case where a format has a comma in the name is permitted? though that would be confusing.)

### LINKED ISSUE #6487: SRST2 MLST regex incorrect
Looks like `-` is intended to be included (also due to the use in the help text, and the workflow I imported), but due to its position in the regex its being interpreted as a range operator. Moving it to the end of the character class would fix this easily.

srst2

cc #6459

hexylena:
FOR CONTRIBUTOR:

Fixes #6487 
Fixes #6459 

The collections are also separated because it made them very difficult to use in an automated way, especially with hamronize. 

Either they're a separate collection or you have to
- extract element identifiers (`txt,tabular` formatted output due to #6459 prevents this tool from selecting this collection weirdly)
- select matching element identifiers
- filter collection (because it won't let you just .. grep. you have to specify the precise list of identifiers. I get it, but the UX isn't great.)

### DIFF
--- a/tools/srst2/macros.xml
+++ b/tools/srst2/macros.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <macros>
     <token name="@TOOL_VERSION@">0.2.0</token>
-    <token name="@VERSION_SUFFIX@">1</token>
+    <token name="@VERSION_SUFFIX@">2</token>
     <token name="@PROFILE@">20.01</token>
     <xml name="requirements">
         <requirements>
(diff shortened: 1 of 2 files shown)


bgruening (approved):
Great, thanks @hexylena!

hexylena:
there's no rush on this, I've switched to phac-nml's version and will probably use that instead since the outputs are organised a bit better for me.

PR #6494, opened 2024-10-28
Add galaxy tool wrapper for IRMA

aaronKol:
Adding a wrapper for the tool IRMA (Iterative Refinement Meta-Assembler).

Problem: The IRMA variable `SINGLE_LOCAL_PROC` doesnt set the thread usage. IRMA will start a subprocess for every (8) virus segment. The variable `SINGLE_LOCAL_PROC` determines on how many threads each of these subprocess should run. E.g. `SINGLE_LOCAL_PROC=4` would yield 4 times 8 = 32 subprocesses. 
I did not find a way to restrict cores used by irma.  

FOR CONTRIBUTOR:


bgruening (commented):
You need to add a .shed.yml file.

tools/irma/IRMA.xml
```diff
@@ -0,0 +1,303 @@
...
+    </macros>
+    <xrefs>
+        <xref type="bio.tools">irma-virus</xref> 
+    </xrefs>
+    <requirements>
+        <requirement type="package" version="@TOOL_VERSION@">IRMA</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        IRMA 
+        "${module}"
```
bgruening, line 15: single-quotes please, single-quotes are not evaluated in bash and hence are more safe

tools/irma/IRMA.xml
```diff
@@ -0,0 +1,303 @@
...
+        #set $input_type = $input_type_conditional.input_type
+        #if $input_type == "single"
+        "${input_type_conditional.fastq_input1}"
+        #elif $input_type == "paired_collection"
+        "${input_type_conditional.fastq_input.forward}" "${input_type_conditional.fastq_input.reverse}"
```
bgruening, line 20:
```suggestion
            '${input_type_conditional.fastq_input1}'
        #elif $input_type == "paired_collection"
            '${input_type_conditional.fastq_input.forward}'
            '${input_type_conditional.fastq_input.reverse}'
```

tools/irma/IRMA.xml
```diff
@@ -0,0 +1,303 @@
...
+        #end if
+        resultDir
+
+        &&
+        
+        python3 $__tool_directory__/createMissingFiles.py
```
bgruening, line 26:
```suggestion
        python3 '$__tool_directory__/createMissingFiles.py'
```

Just imaging a Galaxy server is deployed on a path with spaces, this command would break without quotes.

tools/irma/IRMA.xml
```diff
@@ -0,0 +1,303 @@
...
+
+    ]]></command>
+    <configfiles>
+        <configfile filename="irma_config.sh"><![CDATA[#if $advanced_config.customize == "yes":
```
bgruening, line 30:
```suggestion
        <configfile filename="irma_config.sh"><![CDATA[
#if $advanced_config.customize == "yes":
```

tools/irma/IRMA.xml
```diff
@@ -0,0 +1,303 @@
...
+            </param>
+            <when value="no" />
+            <when value="yes">
+                <section name="general" title="General settings" expanded="false">
+                    <param name="SINGLE_LOCAL_PROC" type="integer" min="4" max="32" value="8" label="Single local processes"/>
+                    <param name="DOUBLE_LOCAL_PROC" type="integer" min="2" max="16" value="4" label="Double local processes"/>
+                    <param name="TMP" type="text" value="/tmp" label="Define temp directory" />
+                </section>
+                <section name="reference" title="General settings" expanded="false">
+                    <param name="SKIP_E" type="select" label="Skip elongation?">
```
bgruening, line 132: Isn't that a boolean?

bgruening: why its own section?

tools/irma/IRMA.xml
```diff
@@ -0,0 +1,303 @@
...
+                        <option value="0">Average</option>
+                        <option value="1" selected="true">Median</option>
+                    </param>
+                    <param name="QUAL_THRESHOLD" type="integer" min="0" value="30" label="Minimum read statistic to keep (Phred)"/>
+                    <param name="MIN_LEN" type="integer" min="0" max="1000" value="125" label="Minimum read length to keep (bp)" />
+                    <param name="INCL_CHIM" type="select" label="Include chimera? Transposase adapter - clips 5' of the adapter on the forward strand and 3' on the reverse strand. applicable to NextTera pair-end reads">
+                        <option value="1">Yes</option>
+                        <option value="0" selected="true">No</option>
+                    </param>
+                    <param name="ADAPTER" type="text" value="AGATGTGTATAAGAGACAG" label="Adapter sequence" />
```
bgruening, line 149: Please add a validator here.

tools/irma/IRMA.xml
```diff
@@ -0,0 +1,303 @@
...
+    <tests>
+        <test expect_num_outputs="27">
+            <param name="fastq_input1" value="test_reads.fq" />
+            <param name="input_type" value="single" />
+            <conditional name="advanced_config">
+                <param name="customize" value="yes" />
+            </conditional>
+            <param name="module" value="FLU-utr" />
+            <param name="optional_outputs" value="vcf,bam"/>
+            <output_collection name="consensus_collection" type="list">
```
bgruening, line 277: how many elements are in this collection, I guess you can use `count` to test for it

tools/irma/IRMA.xml
```diff
@@ -0,0 +1,303 @@
...
+                <element name="HA">
+                  <assert_contents>
+                    <has_n_lines n="2"/>
+                  </assert_contents>
+                </element>
+            </output_collection>
+        </test>
+    </tests>
+    <help><![CDATA[
```
bgruening, line 286: It would be better if you explain what the tool is doing, when a user should use it ... 

The license etc is not relevant here.

tools/irma/createMissingFiles.py
```diff
@@ -0,0 +1,73 @@
...
+expectedSegments = ["A_MP", "A_NP", "A_HA", "A_PB1",
+                    "A_PB2", "A_NA", "A_PA", "A_NS"]
+
+
+def renameSubtypeFiles(identifier):
+    files = glob.glob(dirPrefix + "A_" + identifier + "_*.*")
+    for file in files:
+        ext = file.split('.')[-1]
+        os.rename(file, dirPrefix + "A_" + identifier + "." + ext)
+        print("Rename:")
```
bgruening, line 15: is this debug leftover?

bgruening (commented):
Please add a few more tests that are testing the conditionals. 
Are you sure all file in `irma_testSinge_out` are actually needed?

tools/irma/IRMA.xml
```diff
@@ -0,0 +1,303 @@
...
+    <description>Construct robust assembly of highly variable RNA viruses</description>
+    <macros>
+        <token name="@TOOL_VERSION@">1.2.0</token>
+        <token name="@VERSION_SUFFIX@">0</token>
+    </macros>
+    <xrefs>
+        <xref type="bio.tools">irma-virus</xref> 
+    </xrefs>
+    <requirements>
+        <requirement type="package" version="@TOOL_VERSION@">IRMA</requirement>
```
bgruening, line 11:
```suggestion
        <requirement type="package" version="@TOOL_VERSION@">irma</requirement>
```

tools/irma/IRMA.xml
```diff
@@ -0,0 +1,303 @@
+<tool id="IRMA" name="IRMA" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="23.1">
```
bgruening, line 1:
```suggestion
<tool id="irma" name="IRMA" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="23.1">
```

tools/irma/IRMA.xml
```diff
@@ -0,0 +1,303 @@
...
+            <param name="customize" type="select" label="Customize Advanced Settings?" help="Choosing Yes here, lets you overwrite advanced settings defined in IRMA's default configuration file.">
+                <option value="no">No, use defaults for advanced settings</option>
+                <option value="yes">Yes</option>
+            </param>
+            <when value="no" />
+            <when value="yes">
+                <section name="general" title="General settings" expanded="false">
+                    <param name="SINGLE_LOCAL_PROC" type="integer" min="4" max="32" value="8" label="Single local processes"/>
+                    <param name="DOUBLE_LOCAL_PROC" type="integer" min="2" max="16" value="4" label="Double local processes"/>
+                    <param name="TMP" type="text" value="/tmp" label="Define temp directory" />
```
bgruening, line 129:
```suggestion
```

tools/irma/IRMA.xml
```diff
@@ -0,0 +1,303 @@
...
+        <token name="@VERSION_SUFFIX@">0</token>
+    </macros>
+    <xrefs>
+        <xref type="bio.tools">irma-virus</xref> 
+    </xrefs>
+    <requirements>
+        <requirement type="package" version="@TOOL_VERSION@">IRMA</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        IRMA
```
bgruening, line 14:
```suggestion
        mkdir ./irma_temp &&
        IRMA_GALAXY_TEMP=$PWD/irma_temp
        IRMA 
```

tools/irma/IRMA.xml
```diff
@@ -0,0 +1,303 @@
...
+        python3 $__tool_directory__/createMissingFiles.py
+
+    ]]></command>
+    <configfiles>
+        <configfile filename="irma_config.sh"><![CDATA[#if $advanced_config.customize == "yes":
+
+### general
+SINGLE_LOCAL_PROC=$advanced_config.general.SINGLE_LOCAL_PROC
+DOUBLE_LOCAL_PROC=$advanced_config.general.DOUBLE_LOCAL_PROC	
+TMP=$advanced_config.general.TMP
```
bgruening, line 35:
```suggestion
TMP=$IRMA_GALAXY_TEMP
```

tools/irma/IRMA.xml
```diff
@@ -0,0 +1,303 @@
...
+        <conditional name="advanced_config">
+            <param name="customize" type="select" label="Customize Advanced Settings?" help="Choosing Yes here, lets you overwrite advanced settings defined in IRMA's default configuration file.">
+                <option value="no">No, use defaults for advanced settings</option>
+                <option value="yes">Yes</option>
+            </param>
+            <when value="no" />
+            <when value="yes">
+                <section name="general" title="General settings" expanded="false">
+                    <param name="SINGLE_LOCAL_PROC" type="integer" min="4" max="32" value="8" label="Single local processes"/>
+                    <param name="DOUBLE_LOCAL_PROC" type="integer" min="2" max="16" value="4" label="Double local processes"/>
```
bgruening, line 128:
```suggestion
```

bgruening: \${GALAXY_SLOTS:-4}`

tools/irma/IRMA.xml
wm75: @aaronKol as @bgruening points out there is no need to implement this as a select param. `type="boolean"` is clearer here.
Same everywhere else where you just have a Yes/No choice. Only exception is where the choice represents a conditional param and determines what's shown in the rest of the tool interface. Booleans are not allowed to cause such a change, i.e. then and only then would a `type="select"` be appropriate.

tools/irma/IRMA.xml
wm75: @aaronKol see https://docs.galaxyproject.org/en/latest/dev/schema.html#id151

tools/irma/IRMA.xml
wm75: @aaronKol does IRMA's temp dir have to be part of the job working directory?

If not, then `\$TMPDIR` is the best location (see, e.g., https://github.com/galaxyproject/tools-iuc/blob/90275e51aa271edacdd9d277c441e38e99502c53/tools/bbtools/bbnorm.xml#L50).

wm75: plus, as @bgruening suggested, you don't need to `export` the the temp dir, but it's sufficient to pass it only to the `IRMA` process. This is both shorter and safer because it comes with less side-effects.

tools/irma/createMissingFiles.py
```diff
@@ -0,0 +1,67 @@
...
+    cmd = "samtools view -H -b headerSamFile.sam > " \
+          + dirPrefix + identifier + ".bam"  # convert to bam
+    if subprocess.run(cmd, shell=True, text=True).returncode == 0:
+        os.remove("headerSamFile.sam")  # delete temporary sam file
+    else:
+        raise ValueError("Could not create bam file!")
+
+
+def writeEmptyFile(identifier, ext):
+    open(dirPrefix + identifier + "." + ext, 'a').close()
```
wm75, line 45: I think `x` is the open mode you'd want here because you want to fail here if the rest of the logic is wrong and for some reason the file exists.

tools/irma/createMissingFiles.py
```diff
@@ -0,0 +1,67 @@
...
+def samtoolsSortAllBam():
+    for segment in expectedSegments:
+        os.rename(dirPrefix + segment + ".bam",
+                  dirPrefix + segment + "_unsorted.bam")
+        cmd = "samtools sort " + dirPrefix + segment + "_unsorted.bam > " \
+              + dirPrefix + segment + ".bam"
+        if subprocess.run(cmd, shell=True, text=True).returncode == 0:
+            os.remove(dirPrefix + segment + "_unsorted.bam")
+        else:
+            raise ValueError("Could not sort bam file!")
```
wm75, line 57: seems you're re-implementing `subprocess.check_call`.

tools/irma/createMissingFiles.py
```diff
@@ -0,0 +1,67 @@
...
+def getMissingSegments():
+    presentSegments = [filename.split('.')[0].split('/')[-1]
+                       for filename in glob.glob(dirPrefix + "*.fasta")]
+    return [segment for segment in expectedSegments
+            if segment not in presentSegments]
+
+
+def getBamHeaderFromAnyFile():
+    anyBamFile = glob.glob(dirPrefix + "*.bam")[0]
+    samtoolsCmd = "samtools view -H " + anyBamFile
```
wm75, line 26: Dangerous!
Use of `shell=True` for subprocess calls is generally a bad habit, but here, in particular, you're relying on IRMA not to put weirdly named stuff into its output directory that you'd pick up as anyBamFile. While this may work for the current version, you don't want to leave surprises for the next person who updates the dependency.
=> drop the `shell="True"` and pass a list of command components instead.

tools/irma/createMissingFiles.py
```diff
@@ -0,0 +1,67 @@
...
+            os.remove(dirPrefix + segment + "_unsorted.bam")
+        else:
+            raise ValueError("Could not sort bam file!")
+
+
+if __name__ == "__main__":
+    renameSubtypeFiles("HA")
+    renameSubtypeFiles("NA")
+    for segment in getMissingSegments():
+        writeEmptyBam(segment, getBamHeaderFromAnyFile())
```
wm75, line 64: you're calling `getBamHeaderFromAnyFile` several times here, which is unnecessary.

tools/irma/createMissingFiles.py
```diff
@@ -0,0 +1,67 @@
...
+    return [segment for segment in expectedSegments
+            if segment not in presentSegments]
+
+
+def getBamHeaderFromAnyFile():
+    anyBamFile = glob.glob(dirPrefix + "*.bam")[0]
+    samtoolsCmd = "samtools view -H " + anyBamFile
+    result = subprocess.run(samtoolsCmd, shell=True,
+                            stdout=subprocess.PIPE, text=True)
+    header = result.stdout.split('\n')[0]
```
wm75, line 29: Why are you keeping only the first line? Is all the rest segment-specific?

tools/irma/createMissingFiles.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def renameSubtypeFiles(identifier):
+    files = glob.glob(dirPrefix + "A_" + identifier + "_*.*")
+    for file in files:
+        ext = file.split('.')[-1]
+        os.rename(file, dirPrefix + "A_" + identifier + "." + ext)
+
+
+def getMissingSegments():
```
wm75, line 17: `os.listdir` seems a more efficient way to get the segments here. Simplifies extraction of the segment names.

tools/irma/createMissingFiles.py
```diff
@@ -0,0 +1,67 @@
...
+            raise ValueError("Could not sort bam file!")
+
+
+if __name__ == "__main__":
+    renameSubtypeFiles("HA")
+    renameSubtypeFiles("NA")
+    for segment in getMissingSegments():
+        writeEmptyBam(segment, getBamHeaderFromAnyFile())
+        writeEmptyFile(segment, "fasta")
+        writeEmptyFile(segment, "vcf")
```
wm75, line 66: for the vcf files it would be better to extract a header from an exisitng file (similar, but simpler than for bams) because, strictl speaking an empty file is not a valid VCF.

tools/irma/IRMA.xml
```diff
@@ -0,0 +1,418 @@
...
+                    <param name="MIN_C" type="integer" min="1" value="2" label="The minimum count for variants"/>
+                    <param name="MIN_AQ" type="integer" min="1" value="24" label="The minimum average variant quality (does not apply to deletions)"/>
+                    <param name="MIN_TCC" type="integer" min="1" value="100" label="The minimum non-ambiguous column coverage"/>
+                    <param name="MIN_CONF" type="float" min="0" max="1" value="0.80" label="The minimum confidence"/>
+                    <param name="SIG_LEVEL" type="select" label="The significance test level for variant calling">
+                        <option value="0.999" selected="true">99.9%</option>
+                        <option value="0.99" >99%</option>
+                        <option value="0.95" >95%</option>
+                        <option value="0.90" >90%</option>
+                    </param>
```
wm75, line 227: Are these fixed levels in IRMA or could you use a simple float param?

tools/irma/IRMA.xml
```diff
@@ -0,0 +1,418 @@
...
+        <collection name="consensus_collection" type="list" label="Fasta Consensus Sequences">
+            <data name="HA" format="fasta" from_work_dir="resultDir/A_HA.fasta" label="HA"/>
+            <data name="MP" format="fasta" from_work_dir="resultDir/A_MP.fasta" label="MP"/>
+            <data name="NA" format="fasta" from_work_dir="resultDir/A_NA.fasta" label="NA"/>
+            <data name="NP" format="fasta" from_work_dir="resultDir/A_NP.fasta" label="NP"/>
+            <data name="NS" format="fasta" from_work_dir="resultDir/A_NS.fasta" label="NS"/>
+            <data name="PA" format="fasta" from_work_dir="resultDir/A_PA.fasta" label="PA"/>
+            <data name="PB1" format="fasta" from_work_dir="resultDir/A_PB1.fasta" label="PB1"/>
+            <data name="PB2" format="fasta" from_work_dir="resultDir/A_PB2.fasta" label="PB2"/>
+        </collection>
```
wm75, line 242: I think the order of the data elements determines their order in the collection and it would be better then to order things by segment number (which goes from largest to smallest sample) because this is what a virologist would expect to find.

tools/irma/createMissingFiles.py
aaronKol: If the file exists it is not altered in mode a.
I don't want this statement to fail. I would have to catch the FileExistsError...
So I think a would be the right choice

tools/irma/createMissingFiles.py
aaronKol: Yes:
@HD	VN:1.4	SO:queryname
@SQ	SN:A_HA_H1	LN:1125

First line is sufficient (and not segment specific)

tools/irma/IRMA.xml
aaronKol: One could pass any float. But these are the thresholds IRMA uses for calculation. See IRMA_RES/scripts/call.pl line 96f:
`if ( $sigLevel >= 1 ) {
		$sigLevel /= 100;
	}

	if ( $sigLevel >= .999 ) {
		$kappa = 3.090232;
	} elsif ( $sigLevel >= .99 ) {
		$kappa = 2.326348;
	} elsif ( $sigLevel >= .95 ) {
		$kappa = 1.644854;
	} elsif ( $sigLevel >= .90 ) {
		$kappa = 1.281552;
	} else {
		$kappa = 3.090232;
	}

	### second order correction ###
	$kappa2 = $kappa ** 2;
	$eta = $kappa2/3 + 1/6;
	$gamma1 = $kappa2*(13/18) + 17/18;
	$gamma2 = $kappa2*(1/18)+7/36;
	###############################`
	
I feel like its fair to not give the user the (mathematically correct) intuition that a significance level of  0.93 differs from 0.90, when it doesn't (in the context of IRMA :) )

tools/irma/createMissingFiles.py
```diff
@@ -15,53 +15,67 @@ def renameSubtypeFiles(identifier):
 
 
 def getMissingSegments():
-    presentSegments = [filename.split('.')[0].split('/')[-1]
-                       for filename in glob.glob(dirPrefix + "*.fasta")]
+    presentSegments = []
+    for file in os.listdir(dirPrefix):
+        if file.endswith(".fasta"):
+            presentSegments.append(file.split('.')[0])
+    print(presentSegments)
```
bgruening, line 22: remove?

tools/irma/IRMA.xml
```diff
@@ -0,0 +1,552 @@
...
+        #if $input_type_conditional.fastq_input1.is_of_type('fastq.gz')
+        ln -sf '${input_type_conditional.fastq_input1}' file1.fastq.gz &&
+        #end if
+
+        #elif $input_type == 'paired_collection'
+
+        #if $input_type_conditional.fastq_input.forward.is_of_type('fastq.gz')
+        ln -sf '${input_type_conditional.fastq_input.forward}' file1.fastq.gz &&
+        ln -sf '${input_type_conditional.fastq_input.reverse}'  file2.fastq.gz &&
+        #end if
```
bgruening, line 29:
```suggestion
            #if $input_type_conditional.fastq_input.forward.is_of_type('fastq.gz')
                ln -sf '${input_type_conditional.fastq_input.forward}' file1.fastq.gz &&
                ln -sf '${input_type_conditional.fastq_input.reverse}'  file2.fastq.gz &&
            #end if
```

tools/irma/IRMA.xml
```diff
@@ -0,0 +1,552 @@
...
+        ### irma command
+        IRMA 
+        '${module}'
+        #if $input_type_conditional.input_type == 'single'
+
+        #if $input_type_conditional.fastq_input1.is_of_type('fastq.gz')
+        file1.fastq.gz
+        #else 
+        '${input_type_conditional.fastq_input1}'
+        #end if
```
bgruening, line 43:
```suggestion
        #if $input_type_conditional.fastq_input1.is_of_type('fastq.gz')
            file1.fastq.gz
        #else 
            '${input_type_conditional.fastq_input1}'
        #end if
```

tools/irma/IRMA.xml
```diff
@@ -0,0 +1,552 @@
...
+
+        #elif $input_type_conditional.input_type == 'paired_collection'
+        
+        #if $input_type_conditional.fastq_input.forward.is_of_type('fastq.gz')
+        file1.fastq.gz
+        file2.fastq.gz
+        #else 
+        '${input_type_conditional.fastq_input.forward}'
+        '${input_type_conditional.fastq_input.reverse}'
+        #end if
```
bgruening, line 53:
```suggestion
        #if $input_type_conditional.fastq_input.forward.is_of_type('fastq.gz')
            file1.fastq.gz
            file2.fastq.gz
        #else 
            '${input_type_conditional.fastq_input.forward}'
            '${input_type_conditional.fastq_input.reverse}'
        #end if
```

wm75 (commented):
a few more suggestions, and I'm currently investigating the multiprocessing issue, but no promises yet

wm75 (approved):
Nice work @aaronKol!

mvdbeek:
Attention: deployment cancelled!

https://github.com/galaxyproject/tools-iuc/actions/runs/11755235436

PR #6499, opened 2024-10-30
Fastp - Format_source to ensure WF usability and remove non-sanger input formats

RZ9082:
FOR CONTRIBUTOR:

- Removed non-sanger input formats as wished.
- Back to using `format_source` to ensure WF usability. However 'paired' and 'paired_collection' use `format_source="in1"` or `format_source="in2"` which is not affecting functionality, but it is still wrong!('paired_collection' should use `format_source="paired_input['forward']"` or `format_source="paired_input['reverse']"`). I still didn't figure out how to solve that rather than creating distinct outputs for each 'paired' and 'paired_collection, Any Ideas, before I do this?

@mvdbeek, @wm75  Could you review this please?


tools/fastp/fastp.xml
```diff
@@ -664,7 +625,7 @@ afford high performance.
 
 **Inputs**
 
-Single-end or Paired-end (compressed) fastqsagnger or fastqillumina files
+Single-end or Paired-end (compressed) fastqsagnger files
```
mvdbeek, line 628:
```suggestion
Single-end or Paired-end (compressed) fastqsanger files
```

tools/fastp/fastp.xml
```diff
@@ -379,10 +335,10 @@ $read_mod_options.base_correction_options.correction
     </inputs>
 
     <outputs>
-        <data name="out1" format="auto" label="${tool.name} on ${on_string}: Read 1 output">
+        <data name="out1" format_source="in1" label="${tool.name} on ${on_string}: Read 1 output">
```
mvdbeek, line 338: Does fastp compress the output if the input is compressed ?
I would assume all output is uncompressed ?
```suggestion
        <data name="out1" format="fastqsanger" label="${tool.name} on ${on_string}: Read 1 output">
```

RZ9082: The tool itself produces a compressed output if the output name has the ext .gz regardless whether the input is compressed or not.
Actually i had it this way at the beginning, but changed to format_source to be able to connect the output to fastqsanger.gz inputs..

RZ9082:
If there any other suggestions, please let me know.

mvdbeek:
Fixes #6560

mvdbeek (approved):

PR #6524, opened 2024-11-05
Adding Single Cell category to tools

nomadscientist:
I've added the Single Cell category to a bunch of tools. A number of these are 'Exploratory analyses', which are not specific to Single Cell (but by extension are not specific to Transcriptomics either), as they look into gene lists to determine relationships.


mvdbeek:
Attention: deployment cancelled!

https://github.com/galaxyproject/tools-iuc/actions/runs/11685037129

PR #6542, opened 2024-11-08
add checkm2

astrovsky01:
FOR CONTRIBUTOR:


tools/checkm2/checkm2.xml
```diff
@@ -0,0 +1,126 @@
...
+        </param>
+        <conditional name="ttable_manual">
+            <param name="set_ttable" type="select" help="If not chosen, tool will automatically determine either 11 or 4" label="Manually set specific  progidal translation table?">
+                <option value='no'>No</option>
+                <option value="yes">Yes</option>
+            </param>
+            <when value="no"/>
+            <when value="yes">
+                <!-- It's not all numbers and there's a check internally if it's in a specific list, so it had to be spelled out -->
+                <param argument="ttable" type="select" label="Prodigal table">
```
bgruening, line 47: It would be useful for the user to tell what those numbers mean.

tools/checkm2/checkm2.xml
```diff
@@ -0,0 +1,126 @@
...
+    checkm2 predict
+    --input input_dir
+    $model
+    $genes
+    #if $ttable_manual.set_ttable == "yes":
+        --ttable $ttable_manual.ttable
+    #end if
+    -x .dat
+    --threads "\${GALAXY_SLOTS:-1}"
+    --database_path "\${CHECKM2_DB_PATH:-$__tool_directory__/tool-data/CheckM2_database/uniref100.KO.1.dmnd}"
```
bgruening, line 27: Is this database stable forever and will not change? If those databases update over time, we need a location file.

tools/checkm2/checkm2.xml
```diff
@@ -0,0 +1,126 @@
...
+    <xrefs>
+        <xref type="bio.tools">dada2</xref>
+    </xrefs>
+    <requirements>
+        <requirement type="package" version="@TOOL_VERSION@">checkm2</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+    mkdir input_dir &&
+    #for $i, $file in enumerate($input):
+        cp $file input_dir/${file.element_identifier}.dat &&
```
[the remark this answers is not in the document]
bernt-matthias: - can we symlink?
- `element_identifier` needs cleaning (eg using `re.sub`).

tools/checkm2/checkm2.xml
```diff
@@ -0,0 +1,135 @@
...
+        </param>
+
+        <param argument="genes" type="boolean" truevalue="--genes" falsevalue="" label="Treat input files as protein files"/>
+        <param name="model" type="select" label="Model options">
+            <option value="">None</option>
+            <option value="--general">Force the use of the general quality prediction model (gradient boost)</option>
+            <option value="--specific">Force the use of the specific quality prediction model (neural network)</option>
+            <option value="--allmodels">Output quality prediction for both models for each genome.</option>
+        </param>
+        <conditional name="ttable_manual">
```
bernt-matthias, line 48: Do not use a conditional. Instead an optional select be used.

tools/checkm2/checkm2.xml
bernt-matthias: Maybe use the code table names for the text https://www.ncbi.nlm.nih.gov/Taxonomy/Utils/wprintgc.cgi

tools/checkm2/checkm2.xml
```diff
@@ -0,0 +1,135 @@
...
+    checkm2 predict
+    --input input_dir
+    $model
+    $genes
+    #if $ttable_manual.set_ttable == "yes":
+        --ttable $ttable_manual.ttable
+    #end if
+    -x .dat
+    --threads "\${GALAXY_SLOTS:-1}"
+    --database_path $database.fields.path
```
bernt-matthias, line 28: also should be single quoted.

tools/checkm2/checkm2.xml
```diff
@@ -0,0 +1,135 @@
...
+    --output-directory output
+    ]]></command>
+    <inputs>
+        <param name="input" type="data" format="fasta" label="Input MAG/SAG datasets" multiple="true"/>
+
+        <param name="database" type="select" label="Select reference genome" help="Checkm2 Diamond database">
```
bernt-matthias, line 34: Would it be of interest if users upload their own dmnd databases (https://github.com/galaxyproject/tools-iuc/blob/main/tools/diamond/diamond_makedb.xml)?

tools/checkm2/checkm2.xml
```diff
@@ -0,0 +1,135 @@
...
+                    <option value="31">31</option>
+                    <option value="33">33</option>
+                </param>
+            </when>
+        </conditional>
+    </inputs>
+    <outputs>
+        <data name="quality" label="${tool.name} on ${on_string}: Quality report" format="tabular" from_work_dir="output/quality_report.tsv"/>
+        <collection name="protein_files" label="${tool.name} on ${on_string}: protein files" type="list">
+            <discover_datasets pattern="__name__" format="fasta" directory="output/protein_files"/>
```
bernt-matthias, line 89: - The extension of the files will be part  of the element identfiers. Should we remove them?

bernt-matthias (commented):
Excellent timing: One of my users just asked for the tool :) 

Could contribute a data manager.

tools/checkm2/checkm2.xml
astrovsky01: Based on the wording on their repo, I was under the impression you needed to use their specific diamond db?

tools/checkm2/checkm2.xml
```diff
@@ -0,0 +1,124 @@
...
+    </xrefs>
+    <requirements>
+        <requirement type="package" version="@TOOL_VERSION@">checkm2</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+    #import re
+    mkdir input_dir &&
+    #for $i, $file in enumerate($input):
+        #set $cleaned =  re.sub('[^\s\w\-\\.]', '_', str($file.element_identifier))
+        ln -s $file input_dir/${cleaned}.dat &&
```
bgruening, line 19:
```suggestion
        ln -s '$file' input_dir/${cleaned}.dat &&
```

tools/checkm2/tool-data/checkm2.loc.sample
```diff
@@ -0,0 +1,18 @@
...
+
+#The checkm2_db_versioned.loc file has this format (longer white space
+#characters are TAB characters):
+#
+#<unique_build_id>	<display_name>	<file_base_path>	<version>
+
+#The <version> column indicates the checkm2 version that generated the database
+
+#
+#diamond_db_1.0.2	Diamond database	/mnt/galaxyIndices/Checkm2_database/uniref100.KO.1.dmnd	1.0.2
```
bgruening, line 18: maybe the version before the path? I guess this is what we do in other location files.

tools/checkm2/checkm2.xml
```diff
@@ -0,0 +1,124 @@
...
+            <option value="31">31. Blastocrithidia Nuclear Code</option>
+            <option value="33">33. Cephalodiscidae Mitochondrial UAA-Tyr Code</option>
+        </param>
+    </inputs>
+    <outputs>
+        <data name="quality" label="${tool.name} on ${on_string}: Quality report" format="tabular" from_work_dir="output/quality_report.tsv"/>
+        <collection name="protein_files" label="${tool.name} on ${on_string}: protein files" type="list">
+            <discover_datasets pattern="__name__" format="fasta" directory="output/protein_files"/>
+        </collection>
+        <collection name="diamond_files" label="${tool.name} on ${on_string}: Diamond files" type="list">
```
bgruening, line 83: Should we add here that this is of type `tabular`

tools/checkm2/checkm2.xml
bgruening: is there a difference between `ext` and `format` (see below)

bernt-matthias: No, they are synonym. Changed it anyway.

bernt-matthias (approved):
Good from my side.

data_managers/data_manager_checkm2/tool-data/checkm2.loc.sample
```diff
@@ -0,0 +1,18 @@
...
+
+#The checkm2_db_versioned.loc file has this format (longer white space
+#characters are TAB characters):
+#
+#<unique_build_id>	<display_name>	<version>	<file_base_path>
+
+#The <version> column indicates the checkm2 version that generated the database
+
+#
+#diamond_db_1.0.2	Diamond database	1.0.2	/mnt/galaxyIndices/Checkm2_database/uniref100.KO.1.dmnd
```
bgruening, line 18: Is this really a diamond DB?
If so, this is interesting ... should we have a general Diamond location file and DM? with some `tag` for different tools?

bernt-matthias: I think so. And I agree that it would be interesting. 

But it would be good to know and store the diamond version that has been used to generate it, or? Seems difficult to find out from the sources. The tool just downloads the latest version from zenodo (and I could not even find the link).  Let me check if `diamond dbinfo` could help.

bgruening: Do we need diamond_build? But yes, we should do that :)

Thanks!

astrovsky01: Yes. I had to get something up and running somewhere I could test it, but I just got it going

astrovsky01: After some digging, I found that Checkm2 doesn't actually work with all diamond databases. It has an internal checksum to make sure it's the specific one from the database download command:

https://github.com/chklovski/CheckM2/blob/319dae65f1c7f2fc1c0bb160d90ac3ba64ed9457/checkm2/versionControl.py#L74

as such, I think that while it would be good to have the general Diamond db data manager, having a specific one for checkm2 is also a good idea

bernt-matthias: Yes, let's go ahead .. I would say. 

My feeling is that a general data manager would be too complex and multiple data managers writing to the same data table also seems confusing. Maybe it's better to have tools load multiple data tables?

astrovsky01: I was playing around with the writing basically extra labels, but that requires someone on the other end parsing the table. Also, depending on the tool, you start to get additional bloat in the requirements. At the very least, checkm2's tool would require its conda package, and that's just extra dependencies for the other tools, even when you don't need it. I think it's a good idea conceptually, but maybe not for this case, specifically

bernt-matthias: Can you elaborate? I do not understand what you want to say. 

> writing basically extra labels

What labels? 

>  Also, depending on the tool, you start to get additional bloat in the requirements. 

How? We would just load another datatable - we need to requirements for this.

astrovsky01: Oh I just mean that if we start to have tools that share the table but can't use all of them. Just referring to the `tool` label you mentioned at the top of the thread.
And I'd meant requirements for the data manager itself. In this case, you'd need the checkm2 conda package on top of the diamond package, as opposed to just the diamond package in the diamond_build_db data manager that already exists. If other tools operate similarly to checkm2 in the future, they'd need their conda package added to the data manager's xml

astrovsky01:
@bgruening Is there anything else we need to do for this?

bgruening:
Thanks @astrovsky01

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/12775674940

bernt-matthias:
There seems to be a Problem with the datatable

```
Unexpected HTTP status code: 500: {"err_msg": "Error attempting to parse file tool_data_table_conf.xml.sample: Merging tabular data tables with non matching columns is not allowed: checkm2:{'value': 0, 'dbkey': 1, 'name': 2, 'version': 3, 'path': 4} != checkm2:{'value': 0, 'name': 1, 'version': 2, 'path': 3}"}
```

@mvdbeek Do you know why the deployment to TTS always works? I have always thought that IT is a safeguard, bot seems not to be.

PR #6545, opened 2024-11-09
Fix wrapper name

wm75:
Fixes naming issue introduced last minute into #6494 :(


wm75:
Sorry, tests were not run before merging the previous PR because of the broken file name. Things look good now again.

PR #6553, opened 2024-11-12
patch krakentools beta_diversity.xml

mthang:
FOR CONTRIBUTOR:

The header section of the krakentools beta_diversity.xml output file embedded the galaxy directory and It affects the file parsing in the downstream. Therefore, I made the simple patch by updating this wrapper. My colleague in Australia is submitting a workflow containing this tool to the workflowhub soon. I just patched the wrapper by adding the element_identifier to each input file.

Here's the screenshot of the beta_diversity output
Screenshot 2024-10-02 at 1 53 19 PM

Many thanks,


tools/krakentools/beta_diversity.xml
```diff
@@ -7,10 +7,14 @@
...
     <command detect_errors="exit_code"><![CDATA[
+#for $input in $inputs
+	ln -s '$input' '${input.element_identifier}' &&
+#end for
+
 beta_diversity.py 
     --inputs
     #for $input in $inputs 
-        '$input' 
+	'${input.element_identifier}'
```
bernt-matthias, line 17: Needs some cleaning for unsafe characters, i.e. `re.sub` ...

bernt-matthias (commented):
We need a bump of the version.

Wondering:

- should free output be tabular
- so we want to have more output assertions?

mthang:
I believe you refer to the wrapper version not the tool version. I will bump the beta_diversity wrapper version up. 

Re:
- should the output be tabular
  The output format is defined as txt in the wrapper not tabular
- so we want to have more output assertions?
  any suggestion? can you clarify what other outputs should be included in the test unit.

bernt-matthias:
> I believe you refer to the wrapper version not the tool version.

Yes.

> The output format is defined as txt in the wrapper not tabular

The question is if this is correct :) The output looks pretty tabular to me :) 

> so we want to have more output assertions?

We could have an assertion that a line having the new identifiers in the output is included.

mthang:
I bumped the galaxy from galaxy1 to galaxy2 and added the assertion to the test unit.

Re:
The question is if this is correct :) The output looks pretty tabular to me :)

It looked like a tabular to me in the beginning as well. Then, I looked closely and realised it's not tabular. They are just the header lines. see the attached.

beta_diversity

Thanks

bernt-matthias:
Its tab separated so it's tabular :) (# commend lines should be ignored).

I removed one extra requirement and included it directly in bioconda https://github.com/bioconda/bioconda-recipes/pull/52125 as it should be. Saves us one container / environment (at least in new installations).

mthang:
@bernt-matthias do you want me to replace "txt" with "tabular" in beta_diversity.xml wrapper? I can do that as well.

bernt-matthias:
Already did so in https://github.com/galaxyproject/tools-iuc/pull/6553/commits/ad1217d87c477ccbfc949a7edd7761313684ae4e :)

tools/krakentools/beta_diversity.xml
```diff
@@ -7,10 +7,19 @@
     <expand macro="requirements"/>
     <expand macro="version"/>
     <command detect_errors="exit_code"><![CDATA[
+#import re
+#set input_names=[]
+
+#for $input in $inputs
+	#set identifier = re.sub('[^\s\w\-\\.]', '_', str($input.element_identifier))
```
bernt-matthias, line 14: I guess .. 

```suggestion
        #set identifier = re.sub('[^\s\w\-\.]', '_', str($input.element_identifier))
```
or just

```suggestion
        #set identifier = re.sub('[^\s\w-.]', '_', str($input.element_identifier))
```

bernt-matthias: Seems to be the former one. Wanted to make some experiments yesterday which version is correct (we have all of them in the repo).. but did not finish yet.

tools/krakentools/beta_diversity.xml
```diff
@@ -7,10 +7,19 @@
     <expand macro="requirements"/>
     <expand macro="version"/>
     <command detect_errors="exit_code"><![CDATA[
+#import re
+#set input_names=[]
+
+#for $input in $inputs
+	#set identifier = re.sub('[^\s\w-.]', '_', str($input.element_identifier))
```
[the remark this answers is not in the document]
mthang: Thanks @bernt-matthias . I just want to ensure I'll make the change correctly. Basically,  adding a backward slash before the hyphen and it will solve the issue? i.e  from [^\s\w-.] to  [^\s\w\-.] , right? or to this [^\s\w\-\.] to cover the dot as well.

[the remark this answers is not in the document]
mthang: @bernt-matthias Yup. Please merge it ! I saw your fix this commit [0faff71](https://github.com/galaxyproject/tools-iuc/pull/6553/commits/0faff71a1399b0d4b04d0eed17f6adbefcb36b0e), so it's all good. Thank you for your time and help!

bernt-matthias (approved):

PR #6557, opened 2024-11-12
update sniffles to new 2.5.2 release

fubar2:

Update sniffles to 2.5.2 release and update test outputs.


fubar2:
@bernt-matthias 

> Should new options like `--dev-emit-sv-lengths` be included (the only one according to the release notes)

The current VGP use cases are the only ones I know anything about and developer options are not part of their requirements, but if other users will need routine access to some or all of those developer options then of course, please add them to suit your requirements?

bernt-matthias (approved):

bernt-matthias:
Thanks @fubar2

PR #6558, opened 2024-11-13
Add bwa-mem2-idx

duartetorreserick:
FOR CONTRIBUTOR:

Often we use bwa in workflows for multiple samples, where the same index is created repeatedly—-hundreds or even thousands of times. This tool will allow the creation of an index before running bwa-mem2, so multiple bwa-mem2 instances can use the same index without creating it over and over again. This will require a modification to bwa-mem2 as well to be able to take advantage of the index.

### DIFF
--- a/tools/bwa_mem2/bwa-mem2-idx.xml
+++ b/tools/bwa_mem2/bwa-mem2-idx.xml
@@ -1,0 +1,72 @@
+<tool id="bwa_mem2_idx" name="BWA-MEM2-INDEX"  version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
+    <description>- creates indexes</description>
+    <macros>
+        <import>macros-index.xml</import>
+    </macros>
+    <expand macro="requirements"/>
+    <command><![CDATA[
+
+## Begin BWA-MEM command line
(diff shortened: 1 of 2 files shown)


tools/bwa_mem2/bwa-mem2-idx.xml
```diff
@@ -0,0 +1,72 @@
...
+cd '$output.files_path' &&
+bwa-mem2 index -p 'index' '${input_fasta}'
+    ]]></command>
+
+    <inputs>
+        <param name="input_fasta" type="data" label="Select a genome to index" help="Build an index for this FASTA sequence." format="fasta,fasta.gz"/>
+    </inputs>
+
+    <outputs>
+        <data name="output" format="text" label="Test Testov"/>
```
mvdbeek, line 21: Can you create a subclass of the directory datatype for this in https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/sample/datatypes_conf.xml.sample?

Something like
```
<datatype extension="bwa_mem_2_index" type="galaxy.datatypes.data:Directory" display_in_upload="true" subclass="true"/>
```

Then you can move the index directory `$output.files_path`, and in the same way the bwa_mem2 tool can consume the index directory from ``$output.extra_files_path`

PR #6574, opened 2024-11-19
 Add MMseqs2 clustering and taxonomy

hugolefeuvre:
FOR CONTRIBUTOR:


tools/mmseqs2/mmseqs2_taxonomy_assignment.xml
```diff
@@ -0,0 +1,711 @@
...
+ln -s -f '${createdb.input_fasta}' 'input' &&
+mmseqs createdb
+    'input'
+    'sequenceDB'
+    --dbtype '$createdb.dbtype'
+    --shuffle $createdb.shuffle &&
+ls -lah &&
+ls -lah '$createtaxdb.database_type.mmseqs2_db_select.fields.path'* &&
+mmseqs createtaxdb
+    '$createtaxdb.database_type.mmseqs2_db_select.fields.path'
```
bernt-matthias, line 21: Can you try: 

```suggestion
ls -lah '$createtaxdb.database_type.mmseqs2_db_select.fields.path'* &&
ln -s  '$createtaxdb.database_type.mmseqs2_db_select.fields.path' taxdb &&
mmseqs createtaxdb
    'taxdb'
```

If needed add an extension to `taxdb`.

It seems that in the end [this script](
https://github.com/soedinglab/MMseqs2/blob/93d74798d7fc584e28ca35ef9522dee7d9747c82/data/workflow/createtaxdb.sh#L119)
is executed and I guess that the trick should work. 

Also wondering if we should download all the databases with every job or if we should create (or reuse existing reference data)? At least the ncbi taxonomy dump files should already be handled by a another data manager. Not sure about the mappings to uniprot.

Of course the option to allow users to provide their own mapping needs to be preserved. 

Wondering if `createdb` and `createtaxdb` should be separate tools / data managers.

tools/mmseqs2/mmseqs2_taxonomy_assignment.xml
```diff
@@ -0,0 +1,711 @@
...
+        'taxo_result.html'
+        --report-mode 1
+        --threads "\${GALAXY_SLOTS:-1}"
+#end if
+    ##-v INT              Verbosity level: 0: quiet, 1: +errors, 2: +warnings, 3: +info [3]
+
+    ]]></command>
+    <inputs>
+        <section name="createdb" title="Convert FASTA/Q file(s) to MMseqs sequence DB format"  expanded="true">
+            <param name="input_fasta" type="data" format="fasta,fasta.gz" label="Input fasta file" help="" />
```
bernt-matthias, line 217: Section title suggests that FASTQ is allowed, but it's not listed in the `format`.

tools/mmseqs2/mmseqs2_taxonomy_assignment.xml
```diff
@@ -0,0 +1,711 @@
...
+                </when>
+            </conditional>
+            <param name="tax_mapping_file" type="data" format="tabular,tsv,txt" label="File to map sequence identifier to taxonomical identifier" help="(--tax-mapping-file)"  optional="true"/>
+            <param argument="--tax-mapping-mode" type="select" label="Map taxonomy based on sequence database" help="" >
+                <option value="0" selected="true">0: .lookup file</option>
+                <option value="1">1: .source file</option>
+            </param>
+        </section>
+        <section name="filtertaxseqdb" title="Filter taxonomy sequence database">
+            <conditional name="filtertaxseqdb_bool">
```
bernt-matthias, line 265: I do not think that you need this conditional. You can just check if the `taxon_list` valiable is `None`. Also, what would happen if you say "yes" and do not provide a taxon list?

tools/mmseqs2/mmseqs2_taxonomy_assignment.xml
```diff
@@ -0,0 +1,711 @@
...
+                <when value="no"/>
+                <when value="yes">
+                    <param argument="--taxon-list" type="text" optional="true" value="" label="Taxonomy ID" help="Possibly multiple values separated by ','"/>
+                </when>
+            </conditional>
+        </section>
+        <section name="taxonomy" title="Taxonomy assignment by computing the lowest common ancestor of homologs">
+            <conditional name="alph_type">
+                <param name="type" type="select" label="Alphabet type" help="" >
+                    <option value="amino_acid" selected="true">Amino acid</option>
```
bernt-matthias, line 279: There seem to be several places where you select AA / NT. Are these choices related?

tools/mmseqs2/mmseqs2_easy_linclust_clustering.xml
```diff
@@ -0,0 +1,300 @@
...
+    ]]></command>
+    <inputs>
+        <param name="input_fasta" type="data" format="fasta,fasta.gz" label="Input fasta file" help="" />
+        <conditional name="alph_type">
+            <param name="type" type="select" label="Alphabet type" help="" >
+                <option value="amino_acid" selected="true">Amino acid</option>
+                <option value="nucleotide">Nucleotide</option>
+            </param>
+            <when value="amino_acid">
+                <param name="alph_size_amino_acid" type="integer" min="2" max="5" value="5" label="Alphabet size" help=""/>
```
bernt-matthias, line 112: always use `argument` instead of `name` (or in addition) where possible.

bernt-matthias: Are alphabet sizes really of interest to users?

tools/mmseqs2/mmseqs2_easy_linclust_clustering.xml
```diff
@@ -0,0 +1,300 @@
...
+                <param argument="--comp-bias-corr-scale" type="float" min="0" max="1" value="1" label="Scale composition bias correction" help=""/>
+                <param name="kmer_per_seq_scale" type="float" min="0" value="0.000" label="Scale k-mer per sequence based on sequence length" help=""/>
+            </when>
+            <when value="nucleotide">
+                <param name="alph_size_nucleotide" type="integer" min="2" max="21" value="21" label="Alphabet size" help=""/>
```
bernt-matthias, line 117: Are the defaults for the alphabet size mixed up. 21 should be for AA and 5 (or 4?) for NT?

tools/mmseqs2/mmseqs2_easy_linclust_clustering.xml
```diff
@@ -0,0 +1,300 @@
...
+            </param>
+            <param argument="--wrapped-scoring" type="boolean" checked="false" truevalue="1" falsevalue="0" label="Double the (nucleotide) query sequence during the scoring process" help="Allow wrapped diagonal scoring around end and start"/>
+            <param name="evalue" type="float" min="0" value="1.000E-03" label="E-value threshold" help="List matches below this E-value (-e)"/>
+            <param argument="--min-aln-len" type="integer" min="0" value="0" label="Minimum alignment length" help=""/>
+            <param argument="--seq-id-mode" type="select" label="Sequence identity mode" help="" >
+                <option value="0" selected="true">Alignment length</option>
+                <option value="1">Shorter</option>
+                <option value="2">Longer sequence</option>
+            </param>
+            <param argument="--alt-ali" type="integer" min="0" value="0" label="Show up to this many alternative alignments" help=""/>
```
bernt-matthias, line 175: Maybe better label: `Number of alternative alignments to show` ?

tools/mmseqs2/mmseqs2_easy_linclust_clustering.xml
```diff
@@ -0,0 +1,300 @@
...
+                <option value="3">Global alignment</option>
+                <option value="4">Longest alignment fulfilling window quality criterion</option>
+            </param>
+            <param argument="--dbtype" type="select" label="Database type" help="" >
+                <option value="0" selected="true">Automatic</option>
+                <option value="1">Amino acid</option>
+                <option value="2">Nucleotides</option>
+            </param>
+            <param argument="--shuffle" type="boolean" checked="true" truevalue="1" falsevalue="0" label="Shuffle input database" help=""/>
+            <param argument="--createdb-mode" type="select" label="Createdb mode" help="" >
```
bernt-matthias, line 217: My intuition tells me that the user should not have control over this parameter.

tools/mmseqs2/mmseqs2_easy_linclust_clustering.xml
```diff
@@ -0,0 +1,300 @@
...
+        </section>
+        <section name="misc" title="Misc">
+            <param argument="--rescore-mode" type="select" label="Rescore diagonals with" help="" >
+                <option value="0" selected="true">Hamming distance</option>
+                <option value="1">Local alignment (score only)</option>
+                <option value="2">Local alignment</option>
+                <option value="3">Global alignment</option>
+                <option value="4">Longest alignment fulfilling window quality criterion</option>
+            </param>
+            <param argument="--dbtype" type="select" label="Database type" help="" >
```
bernt-matthias, line 211: This seems kind of redundant with `alph_type`?

tools/mmseqs2/mmseqs2_easy_linclust_clustering.xml
```diff
@@ -0,0 +1,300 @@
...
+            </param>
+            <param argument="--shuffle" type="boolean" checked="true" truevalue="1" falsevalue="0" label="Shuffle input database" help=""/>
+            <param argument="--createdb-mode" type="select" label="Createdb mode" help="" >
+                <option value="0">Copy data</option>
+                <option value="1" selected="true">Soft link data and write new index (works only with single line fasta/q)</option>
+            </param>
+            <param argument="--id-offset" type="integer" min="0" value="0" label="Numeric ids in index file are offset by this value" help=""/>
+        </section>
+        <section name="common" title="Common">
+            <param argument="--max-seq-len" type="integer" min="0" value="65535" label="Maximum sequence length" help=""/>
```
bernt-matthias, line 224: Also maybe optional?

tools/mmseqs2/mmseqs2_easy_linclust_clustering.xml
```diff
@@ -0,0 +1,300 @@
...
+        </section>
+        <section name="expert" title="Expert">
+            <param argument="--filter-hits" type="boolean" checked="false" truevalue="1" falsevalue="0" label="Filter hits by seq.id. and coverage" help=""/>
+            <param argument="--sort-results" type="select" label="Sort results" help="">
+                <option value="0" selected="true">No sorting</option>
+                <option value="1">Sort by E-value (Alignment) or seq.id. (Hamming)</option>
+            </param>
+        </section>
+        <section name="output_files" title="Selection of the output files">
+          <param name="output_selection" type="select" display="checkboxes" multiple="true"  label="Output files selection">
```
bernt-matthias, line 234: I remember that you can set `min="1"` to ensure that the tool has at least one output.

tools/mmseqs2/mmseqs2_easy_linclust_clustering.xml
```diff
@@ -0,0 +1,300 @@
...
+                <assert_contents>
+                    <has_text text="GAATAGCGGGACGCCAAGGGGCGGCCTTGCGTCCGCCCACGTGTGTGCTTGGCACGCGGGGCGTCCGCAAACCTTTGATCGGAACTTGCGATGGAGAAGCT"/>
+                    <has_size value="627000" delta="50000"/>
+                </assert_contents>
+            </output>
+            <output name="output_cluster" ftype="tabular">
+                <assert_contents>
+                    <has_line line="MYSTERY.13&#009;MYSTERY.13"/>
+                    <has_n_columns n="2"/>
+                    <has_size value="113000" delta="50000"/>
```
bernt-matthias, line 271: Quite a big delta? Maybe also/instead use `has_n_lines`?

tools/mmseqs2/mmseqs2_taxonomy_assignment.xml
```diff
@@ -0,0 +1,711 @@
...
+    --threads "\${GALAXY_SLOTS:-1}" &&
+    #if str($filtertaxseqdb.filtertaxseqdb_bool.use_filter) == "yes"
+        mmseqs filtertaxseqdb
+            'sequenceDB'
+            'sequenceDB_filtered' &&
+            #if str($filtertaxseqdb.filtertaxseqdb_bool.use_filter.taxon_list) != ""
+                --taxon-list '$filtertaxseqdb.filtertaxseqdb_bool.use_filter.taxon_list' &&
+            #end if
+    #end if
+chmod -Rv 766 '$createtaxdb.database_type.mmseqs2_db_select.fields.path'_taxonomy &&
```
bernt-matthias, line 36: You must not (and should not be able to) write to anything outside of the job working dir (+anything that is referred to by variables).

tools/mmseqs2/mmseqs2_taxonomy_assignment.xml
```diff
@@ -0,0 +1,711 @@
...
+                    <option value="yes">Yes</option>
+                </param>
+                <when value="no"/>
+                <when value="yes">
+                    <param argument="--taxon-list" type="text" optional="true" value="" label="Taxonomy ID" help="Possibly multiple values separated by ','"/>
+                </when>
+            </conditional>
+        </section>
+        <section name="taxonomy" title="Taxonomy assignment by computing the lowest common ancestor of homologs">
+            <conditional name="alph_type">
```
bernt-matthias, line 277: If multiple tools share parameters then often macros are a good idea.

tools/mmseqs2/mmseqs2_taxonomy_assignment.xml
```diff
@@ -0,0 +1,711 @@
...
+                    <option value="amino_acid" selected="true">Amino acid</option>
+                    <option value="nucleotide">Nucleotide</option>
+                </param>
+                <when value="amino_acid">
+                    <param name="alph_size_amino_acid" type="integer" min="2" max="21" value="21" label="Alphabet size" help=""/>
+                    <param argument="--comp-bias-corr-scale" type="float" min="0" max="1" value="1" label="Scale composition bias correction" help=""/>
+                </when>
+                <when value="nucleotide">
+                    <param name="alph_size_nucleotide" type="integer" min="2" max="5" value="5" label="Alphabet size" help=""/>
+                    <param name="zdrop" type="integer" min="0" value="40" label="Maximal allowed difference between score values before alignment is truncated" help=""/>
```
bernt-matthias, line 288: From the label I'm wondring if this is applicable to nt sequences only?

tools/mmseqs2/mmseqs2_taxonomy_assignment.xml
```diff
@@ -0,0 +1,711 @@
...
+                <param argument="--orf-start-mode" type="select" label="ORF fragment can be" help="" >
+                    <option value="0">From start to stop</option>
+                    <option value="1" selected="true">From any to stop</option>
+                    <option value="2">From last encountered start to stop (no start in the middle)</option>
+                </param>
+                <param argument="--forward-frames" type="text" value="1,2,3" label="Comma-separated list of frames on the forward strand to be extracted" help=""/>
+                <param argument="--reverse-frames" type="text" value="1,2,3" label="Comma-separated list of frames on the reverse strand to be extracted" help=""/>
+                <param argument="--translation-table" type="select" label="Translation table" help="">
+                    <option value="1" selected="true">Canonical</option>
+                    <option value="2">Vert_mitochondrial</option>
```
bernt-matthias, line 443: Maybe spell out: vertebrate mitochondrial .. also some other options. Maybe use these names: https://www.ncbi.nlm.nih.gov/Taxonomy/Utils/wprintgc.cgi

bernt-matthias (commented):
Did a first quick review of the tools. Thanks already for the massive amount of work.

tools/mmseqs2/mmseqs2_easy_linclust_clustering.xml
```diff
@@ -0,0 +1,300 @@
...
+
+
+By Martin Steinegger <martin.steinegger@snu.ac.kr> & Milot Mirdita <milot@mirdita.de> & Florian Breitwieser <florian.bw@gmail.com> & Eli Levy Karin <eli.levy.karin@gmail.com>
+
+-----
+
+**References**
+
+- Steinegger M, Soding J: MMseqs2 enables sensitive protein sequence searching for the analysis of massive data sets. Nature Biotechnology, 35(11), 1026-1028 (2017)
+- Mirdita M, Steinegger M, Breitwieser F, Soding J, Levy Karin E: Fast and sensitive taxonomic assignment to metagenomic contigs. Bioinformatics, btab184 (2021)
```
bernt-matthias, line 297: I think its better to have references only in the `citations` list.

tools/mmseqs2/mmseqs2_easy_linclust_clustering.xml
```diff
@@ -0,0 +1,300 @@
...
+The software is designed to run on multiple cores and servers and exhibits very good scalability. 
+MMseqs2 can run 10000 times faster than BLAST. At 100 times its speed it achieves almost the same sensitivity. 
+It can perform profile searches with the same sensitivity as PSI-BLAST at over 400 times its speed.
+
+**Usage** 
+MMseqs easy-linclust is useful to clusters entries from a FASTA/FASTQ file using the cascaded clustering algorithm.
+It offers an efficient clustering workflow, scaling linearly with input size. Similar to easy-cluster, but more suitable for handling very large datasets efficiently.
+
+
+By Martin Steinegger <martin.steinegger@snu.ac.kr> & Milot Mirdita <milot@mirdita.de> & Florian Breitwieser <florian.bw@gmail.com> & Eli Levy Karin <eli.levy.karin@gmail.com>
```
bernt-matthias, line 290: I think the citations are sufficient and this can be removed. Maybe link the source code repo instead. Opening issues is usually better than writing mails.

tools/mmseqs2/mmseqs2_taxonomy_assignment.xml
hugolefeuvre: > Also wondering if we should download all the databases with every job or if we should create (or reuse existing reference data)? At least the ncbi taxonomy dump files should already be handled by a another data manager. Not sure about the mappings to uniprot.

I don't know enough about this to be able to answer your question.

> Wondering if createdb and createtaxdb should be separate tools / data managers.

We've already thought about it, but we thought it would be more interesting to be able to perform an end-to-end taxonomy analysis (fasta file to contig taxonomy) with a single galaxy module. But it can be discussed if you think it might be useful for other uses.

bernt-matthias: I guess the main question is if the output of `createdb` and `createtaxdb` can be used by multiple tools / reused later. 

There are also disadvantages wrt reproducibility when the current DB is used. But of course it also has disadvantages. 

My intuition tells me to split it. Also because in jobs that mainly download data need to be handled differently on compute systems. This is not possible if compute and download tasks are mixed in a tool. 

But this is only my intuition and I leave this to you.

hugolefeuvre: In my opinion, one of the major problems with separating createdb and createtaxdb is the output from these 2 modules. There are several reasons for this: 
- The output format is not supported by Galaxy 
- createdb creates several files which must be used together (in a collection at the very most) 
- createtaxdb creates a file which must be integrated with the other files in order to be able to use the DB as a taxDB, it can take as input either the files generated by createdb, or the files from a DB which doesn't have a taxo.

In other words, createtaxdb might not be in the actual xml because the databases I'm giving it as input at the moment are only the taxonomy DBs proposed by MMseqs, but there are other DM mmseqs2 data_tables that don't have a taxonomy and for which createtaxdb could be useful (https://github.com/soedinglab/mmseqs2/wiki#downloading-databases).

> This is not possible if compute and download tasks are mixed in a tool 

Could this be why I'm having a problem in my test?
Finally, I don't really know what to do. On the one hand, I think it's a good idea to split it up, as this allows more flexibility for the user, but on the other hand, it seems more complicated to set up and less intuitive for the user.

tools/mmseqs2/mmseqs2_taxonomy_assignment.xml
hugolefeuvre: Indeed, I think fastQ files are allowed only with --createdb-mode 1
` # Create a seqDB by indexing existing FASTA/Q (for single line fasta entries only)
 mmseqs createdb seq.fasta sequenceDB --createdb-mode 1`
Do you think I have to filter depending on format ? (if fastq --> --createdb-mode 1)

tools/mmseqs2/mmseqs2_taxonomy_assignment.xml
hugolefeuvre: I'll try later by putting the database in a directory and making a symlink of that directory. Because the way the symlink was created only took into account the Swiss-Prot file in the database, but there are others (Swiss-Prot_h, Swiss-Prot_index...)

bernt-matthias: The directory datatype might be a solution for the datatype questions.

> Finally, I don't really know what to do. 

Follow your intuition. Your arguments are perfectly reasonable. And we can still improve this later.

tools/mmseqs2/mmseqs2_taxonomy_assignment.xml
hugolefeuvre: You mean in createtaxdb and in taxonomy for example ? If that's the case no they are not related

tools/mmseqs2/mmseqs2_easy_linclust_clustering.xml
hugolefeuvre: I think not, like all the parameters used in the alph_type condition, most of them are of type TWIN and I'm not sure they're really necessary, I have the impression that their value is automatically changed on the type of alphabet detected by the tool.

tools/mmseqs2/mmseqs2_easy_linclust_clustering.xml
hugolefeuvre: Yes I mixed up, changed it until we decide if it is necessary or not

tools/mmseqs2/mmseqs2_easy_linclust_clustering.xml
hugolefeuvre: I think it might be useful if the input file is a single line fasta or fastq (--createdb-mode 1). What are you afraid of ?

tools/mmseqs2/mmseqs2_easy_linclust_clustering.xml
hugolefeuvre: You are right, maybe I can use it as conditionnal in the part command instead of `alph_type` ?

tools/mmseqs2/mmseqs2_easy_linclust_clustering.xml
bernt-matthias: Its a technical detail that most users will not understand. 

I have to ask: What will be copied / symlinked? 

FASTA/Q metadata should contain the info on the number of sequences. So you might actually set this automatically.

tools/mmseqs2/mmseqs2_taxonomy_assignment.xml
hugolefeuvre: The whole label in the tool help is : `--zdrop INT                      Maximal allowed difference between score values before alignment is truncated  (nucleotide alignment only) [40]`

tools/mmseqs2/mmseqs2_taxonomy_assignment.xml
hugolefeuvre: You are right, working on it

hugolefeuvre:
It seems that the tests are not currently running because of problems with MMseqs2, in particular when downloading the taxonomy part for Uniprot (as the SILVA test is running correctly). I'll wait until the problem is solved, if it takes too long then I'll use other test databases

hugolefeuvre:
Hi @bernt-matthias, does the PR require further, more in-depth reviews (I guess so), perhaps by other reviewers ?

tools/mmseqs2/mmseqs2_easy_linclust_clustering.xml
```diff
@@ -0,0 +1,242 @@
+<tool id="mmseqs2_easy_linclust_clustering" name="MMseqs2 Sequence Clustering" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
+    <description>
+        Efficient clustering workflow suitable for handling very large datasets efficiently
```
bernt-matthias, line 3: Can this be as short as possible? For instance just of `of very large datasets`. Then the tool menu will show: 

`MMseqs2 Sequence Clustering of very large datasets`

bernt-matthias (commented):
Had another look, mainly on the data manager. 

Is there any relation between the nt and the tax-nt databases? For instance are there corresponding entries?

data_managers/data_manager_mmseqs2_database/data_manager/data_manager_mmseqs2_download.xml
```diff
@@ -0,0 +1,164 @@
...
+    <stdio>
+        <exit_code range=":-1" level="fatal" description="Error: Cannot open file"/>
+        <exit_code range="1:" level="fatal" description="Error"/>
+    </stdio>
+    <command><![CDATA[
+#set $database_name = str($database).split('/')[-1] if '/' in str($database) else str($database)
+mkdir -p '$database_name' &&
+mkdir -p '$out_file.extra_files_path' &&
+mmseqs databases 
+        '$database' '$database_name'/database
```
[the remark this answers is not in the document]
hugolefeuvre: Don't you lose the information about the database name in the path if you do this? 
And above all I wanted to have the name of the directory at the end of the path (the name of the database) which contains the various files named with the prefix database which make up the database. I'm currently testing with database at the end of the path and the tests are failing.

data_managers/data_manager_mmseqs2_database/data_manager/data_manager_mmseqs2_download.xml
```diff
@@ -0,0 +1,164 @@
...
+                    <option value="UniRef100" selected="true">UniRef100</option>
+                    <option value="UniRef90">UniRef90</option>
+                    <option value="UniRef50">UniRef50</option>
+                    <option value="UniProtKB">UniProtKB</option>
+                    <option value="UniProtKB/TrEMBL">TrEMBL (UniProtKB)</option>
+                    <option value="UniProtKB/Swiss-Prot">Swiss-Prot (UniProtKB)</option>
+                    <option value="NR">NR (Non-redundant protein sequences from GenPept, Swissprot, PIR, PDF, PDB, and NCBI RefSeq)</option>
+                    <option value="GTDB">GTDB (Genome Taxonomy Database)</option>
+                </param>
+                <param name="data_table" type="select" label="Select datatable">
```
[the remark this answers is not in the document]
hugolefeuvre: Yes you are right, I delete this param

data_managers/data_manager_mmseqs2_database/data_manager/data_manager_mmseqs2_download.xml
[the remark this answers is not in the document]
bernt-matthias: Got it. Should be fine.

hugolefeuvre:
> Is there any relation between the nt and the tax-nt databases? For instance are there corresponding entries?
What do you mean with relation ? If the databases in mmseqs2_nucleotide_databases and mmseqs2_nucleotide_taxonomy_databases can be the same ? For example SILVA which is in both datatable ?
If yes, there is no difference between them. Databases that are in nt and not in tax-nt databases don't have taxonomic information and need createtaxdb if you want to use them as taxonomic database.

data_managers/data_manager_mmseqs2_database/data_manager/data_manager_mmseqs2_download.xml
```diff
@@ -0,0 +1,164 @@
...
+    </requirements>
+    <stdio>
+        <exit_code range=":-1" level="fatal" description="Error: Cannot open file"/>
+        <exit_code range="1:" level="fatal" description="Error"/>
+    </stdio>
+    <command><![CDATA[
+#set $database_name = str($database).split('/')[-1] if '/' in str($database) else str($database)
+mkdir -p '$database_name' &&
+mkdir -p '$out_file.extra_files_path' &&
+mmseqs databases
```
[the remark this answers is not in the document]
bernt-matthias: > do I need to create an issue on MMseqs repo to ask them ?

This would be perfect.

> idea would be to take the fasta files already existing in Galaxy from the databases

This would be my idea.

> I wonder how easy it is to find out which fasta files are used to build mmseqs databases

This would be good anyway, otherwise we can not answer this question to users.

data_managers/data_manager_mmseqs2_database/data_manager/data_manager_mmseqs2_download.xml
```diff
@@ -0,0 +1,164 @@
...
+            <when value="aminoacid_taxonomy">
+                <param name="database" type="select" label="MMseqs2 aminoacid databases that can be used for taxonomy">
+                    <option value="UniRef100" selected="true">UniRef100</option>
+                    <option value="UniRef90">UniRef90</option>
+                    <option value="UniRef50">UniRef50</option>
+                    <option value="UniProtKB">UniProtKB</option>
+                    <option value="UniProtKB/TrEMBL">TrEMBL (UniProtKB)</option>
+                    <option value="UniProtKB/Swiss-Prot">Swiss-Prot (UniProtKB)</option>
+                    <option value="NR">NR (Non-redundant protein sequences from GenPept, Swissprot, PIR, PDF, PDB, and NCBI RefSeq)</option>
+                    <option value="GTDB">GTDB (Genome Taxonomy Database)</option>
```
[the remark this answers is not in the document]
bernt-matthias: OK. So one thing that we should probably take case of (if we use fasta from other data tables) is that the mmseqs files are installed to a separate folder.

bernt-matthias:
One suggestion: If we can agree on a set of data tables and their columns we can also split the PR in two parts: one for the tool and one for the data manager. To me it seems that the datamanager will need a while and will slow down the progreee of the tool.

data_managers/data_manager_mmseqs2_database/data_manager/data_manager_mmseqs2_download.xml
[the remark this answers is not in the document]
hugolefeuvre: It doesn't have to be in a separate folder, but I prefer to work that way as several files make up the output. 
After checking, the GTDB database using mmseqs is finally 170GB, with the file containing the sequences being 100GB.

hugolefeuvre:
> One suggestion: If we can agree on a set of data tables and their columns we can also split the PR in two parts: one for the tool and one for the data manager. To me it seems that the datamanager will need a while and will slow down the progree of the tool.

No problem if you think it's the best way to do it. I'm not an expert and I don't know much concerning all data tables already present in galaxy and what we can use and how to do it, but I'm curious to learn. 
Questions : 
-  Do I still need to create an issue on MMseqs repo to ask them about db versions ?
- How can we agree on a set of data tables ?
- The tool part needs a datamanager to function, so won't we have to wait until the DM is finished for the tool part to work? (just wonder why split the PR)

bernt-matthias (commented):
My suggestion regarding the datatables would be to join all into a single one. This joined data table should have two additional columns indicating

- sequence type (amino acid / nucleotide)
- whether it can be used for taxonomy

The information should be available in the table given at https://github.com/soedinglab/mmseqs2/wiki#downloading-databases

When using this data table in the tools you need to add filters that will keep only the appropriate entries. 

It seems that the download provides the information on the database version:

> The databases workflow will further more create a file with the .version suffix that indicates version number or - if not available - the date of the download.

I would suggest to have an additional column database version that should contain the content of this file (assuming that it contains just the value). 

It's still a bit unfortunate that we already have some of the data in other data tables (in particular gtdb which is huge). Can you ask in the upstream repo if this could be reused (but we can not change the data in those datatables). 

Sorrry for the delay.

bernt-matthias:
> Can you be more explicit about upstream repo ? The GTDB one, the mmseqs one or the main/IUC one ?

Sorry. I mean the mmseqs2 repo.

> Unfortunately, the value into this file is not just the DB version and the content is not the same depending on the database.

Do you have an example - to get the idea?

hugolefeuvre:
> Sorry. I mean the mmseqs2 repo.

Ok I will

> Do you have an example - to get the idea?

Yes, for example the Swiss-Prot one is like this : 
```
UniProt Knowledgebase Release 2024_05 consists of:
UniProtKB/Swiss-Prot Release 2024_05 of 02-Oct-2024
UniProtKB/TrEMBL Release 2024_05 of 02-Oct-2024
```
And the SILVA one start like that : 
```
README for SILVA 138 export files

RAST FILES:
===========

Specific export files for the MG-RAST server (Argonne National Lab)

TAXONOMY FILES:
===============

tax_slv_[ls]su_VERSION.txt
-------------------------
These files contain taxonomic rank designations for all taxonomic paths
...
```
And there is more then 100 lines into it.

bebatut:
@bernt-matthias do you think it is ready to be merged?

bernt-matthias (approved):

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/14108783543

PR #6595, opened 2024-11-29
Hicexplorer 3.7.6

joachimwolff:
FOR CONTRIBUTOR:

Update of HiCExplorer, new tool to build micro-C matrices.


tools/hicexplorer/hicBuildMatrixMicroC.xml
```diff
@@ -0,0 +1,225 @@
+<tool id="hicexplorer_hicbuildmatrixmicroc" name="@BINARY@" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@">
```
bgruening, line 1:
```suggestion
<tool id="hicexplorer_hicbuildmatrixmicroc" name="@BINARY@" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="23.0">
```

Can you please add this to all tools please?

tools/hicexplorer/hicBuildMatrixMicroC.xml
```diff
@@ -0,0 +1,225 @@
...
+    <description>create a contact matrix</description>
+    <macros>
+        <token name="@BINARY@">hicBuildMatrixMicroC</token>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements">
+        <requirement type="package" version="1.19">samtools</requirement>
```
bgruening:
```suggestion
        <requirement type="package" version="1.21">samtools</requirement>
```

bgruening (commented):
Hi @joachimwolff nice to see you here again.

I hope bumping the samtools requirement will fix the error.

tools/hicexplorer/hicBuildMatrixMicroC.xml
joachimwolff: Was already there for the most tools via macros.xml. Added it where it was missing.

joachimwolff:
I don't know why it fails here, on the local machine of mine everything is passing. Is there a way to get the tool_test_output.html? I searched in the logs for a hint what is wrong, but don't see something obvious.

bgruening:
You will find the html on this site https://github.com/galaxyproject/tools-iuc/actions/runs/12134110993?pr=6595

grafik


Unfortunately, HiCE is not providing any useful error message as it seems. Can it be that it uses too much memory?
Please also make sure you are running planmoe with `--biocontainers`, we do this here on CI as well.

joachimwolff:
Is it possible to increase the memory?

bgruening:
In GitHub Actions, I don't think so. But it should have 16GB? How much do you need?

joachimwolff:
16 GB is way more than it should need, this is the usual toy data we use since at least 2016. The data was not changed, only the organisation of HiCExplorer in the background and the micro-C part drops a few parameters in comparison to hicBuildMatrix. This should all run through as always.

bgruening:
Maybe you understand the error better than me. Memory was just an assumption. I will also clone it and run it locally later this day.

joachimwolff:
Using --biocontainers leads to a fail of all tools on my local machine, but only if a subset is run. A run of all leads not to a crash, but at some point it simply hangs and does not continue to test. Unclear to me what is happening here.

joachimwolff:
I got it locally with biocontainers running, and the hicBuildMatrix / hicBuildMatrixMicroC test cases were failing. The only difference I saw between the passing tools and the hicBuildMatrix / hicBuildMatrixMicroC was the additional requirement to samtools. After removal, the tests are passing.

Given that HiCExplorer installs via conda dependency resolving a samtools version, I am not sure why we need to declare this dependency explicit in the wrapper. Samtools was not listed in the first versions of the hicBuildMatrix wrapper, only since HiCExplorer version 1.5, with a commit in December 2016: https://github.com/deeptools/HiCExplorer/commit/d3b1279ef1819a7efa2ccc892e891a8a52f83c1a

bgruening:
Nice find! Thanks a lot!

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/12184251422

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/12184251422

joachimwolff:
```
Could not update suite_hicexplorer
Unexpected HTTP status code: 500: {"err_msg": "The repository_dependencies.xml file contains an invalid <repository> tag.  Unable to locate repository with name hicexplorer_hicbuildmatrixmicroc and owner bgruening.  "}
Failed to update repository contents for repository 'suite_hicexplorer' on the main Tool Shed.
Repository metadata updated successfully for repository 'suite_hicexplorer' on the main Tool Shed.
```

Did I forget to add the new tool hicBuildMatrixMicroC anywhere for the suite hicexplorer?

bgruening:
I did it manually, we have the problem that those tools are under `bgruening` user for historical reasons and new tools ... ah complicated. I think I fixed that yesterday evening. It should be there now.

PR #6611, opened 2024-12-05
update proteinortho=6.3.4

pmjklemm:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/proteinortho/proteinortho.xml
+++ b/tools/proteinortho/proteinortho.xml
@@ -99,13 +99,13 @@
             2> >(sed -E "s/.\[([0-9]{1,2}(;[0-9]{1,2})?)?[mGK]//g" 1>&2)
         #if $more_options.selfblast:
             &&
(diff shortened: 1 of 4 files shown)


bernt-matthias:
Thanks for keeping this up to date. I guess there is a problem to resolve the environment. Can the other requirements of the too be updated as well? Alternatively, would it be possible to include all requirements in the bioconda recipe?

pmjklemm:
@bernt-matthias, thanks for the tip, I was reading the error logs but did not find the cause of the failing CI.
I will update the requirements. 
For the bioconda recipe: I wanted to keep the bioconda as minimal as possible (so only diamond is included there) but include them here in the galaxy package

bernt-matthias:
The problem might be our container building / mamba .. I think this will be solved today. 

Thanks for updating the requirements. 

The idea to keep the bioconda recipe minimal sounds perfectly reasonable to me.

bgruening (approved):

PR #6618, opened 2024-12-11
Sourmash wrapper

renu-pal:
error in sourmash DNA sketch


renu-pal:
@bgruening Thanks for pointing it out. I have corrected it. I do not understand how  I create this mess every time even though I push my changes in a new branch and a new PR. Any idea why this is happening ?

bgruening:
You need to create your new branch from a clean branch. The best way is maybe to always go back to the `main` branch with `git checkout main` and then create a new branch. This will make sure you base your new branch on top of the main branch.

SaimMomin12:
Continued here: https://github.com/galaxyproject/tools-iuc/pull/7491

PR #6646, opened 2024-12-19
Add deeparg

hugolefeuvre:
FOR CONTRIBUTOR:


hugolefeuvre:
Problem when using deepARG in a container with the Theano package: `IOError: [Errno 13] Permission denied: '/usr/local/lib/python2.7/site-packages/Theano-0.8.2-py2.7.egg-info/PKG-INFO'` 
This problem has already been reported and closed in this issue: https://github.com/nf-core/funcscan/issues/23 but there's no clear explanation of how they solved the problem, so I have restarted the discussion to try to get more information.

jfy133:
Just so it doesn't get lost in the closed funcscan thread: 

We ended up mounting a dummy file (the bash binary of the execution node) but with the name PKG_INFO:

https://github.com/nf-core/funcscan/blob/master/modules%2Fnf-core%2Fdeeparg%2Fpredict%2Fmain.nf#L10-L18

Also I see lower down in the script execution that we also make a directory for theanos which you may also have to do. 

I hope that helps/works for you!

hugolefeuvre:
Thank you for your answer @jfy133 ! I'll see if it works for me !

hugolefeuvre:
> I did something here, but forgot what I wanted to do ;)
> 
> maybe useful [hugolefeuvre#1](https://github.com/hugolefeuvre/tools-iuc/pull/1) ?

It was maybe about the message I send on galaxy-iuc on Element ?

> Hello, is it possible to add a parameter for the docker run launched in the git tests ? Like --docker_run_extra_arguments which I can use for a planemo test --biocontainers
This would be useful for the deepARG wrapper, as the tool fails in a container and requires this parameter to be added in order to work: -v $(which bash):/usr/local/lib/python2.7/site-packages/Theano-0.8.2-py2.7.egg-info/PKG-INFO
Thank you !

And you answered

> I think this is possible, but we should fix the container that is a hack at its best.

But beyond that I think it can be useful in any case.

bgruening:
@bernt-matthias can you maybe look at the DM bits here?

data_managers/data_manager_deeparg/data_manager/data_manager_deeparg.xml
```diff
@@ -0,0 +1,65 @@
...
+        <token name="@VERSION_SUFFIX@">0</token>
+    </macros>
+    <requirements>
+        <requirement type="package" version="@TOOL_VERSION@">deeparg</requirement>
+    </requirements>
+    <stdio>
+        <exit_code range=":-1" level="fatal" description="Error: Cannot open file"/>
+        <exit_code range="1:" level="fatal" description="Error"/>
+    </stdio>
+    <command><![CDATA[
```
bernt-matthias, line 14: Use `detect_errors="exit_code"` in the command .. instead of the stdio block

data_managers/data_manager_deeparg/data_manager/data_manager_deeparg.xml
```diff
@@ -0,0 +1,65 @@
...
+        mkdir -p '$out_file.extra_files_path' &&
+        deeparg download_data -o 'deeparg_$version' &&
+        mv 'deeparg_$version' '$out_file.extra_files_path' &&
```
bernt-matthias, line 17: Just: 

```suggestion
        deeparg download_data -o '$out_file.extra_files_path/deeparg_$version/' &&
```

data_managers/data_manager_deeparg/data_manager/data_manager_deeparg.xml
```diff
@@ -0,0 +1,65 @@
...
+        "db_version": "$version"
+      }
+    ]
+  }
+}]]>
+        </configfile>
+    </configfiles>
+    <inputs>
+        <param name="version" type="select" label="DB version">
+            <option value="1.0.4" selected="true">Data needed for running DeepARG v1.0.4</option>
```
bernt-matthias, line 39: Can we use `@TOOL_VERSION@` here instead of `1.0.4`?

Do you know if the tool can also download other versions. Do you know if future versions can use older downloads?

data_managers/data_manager_deeparg/data_manager_conf.xml
```diff
@@ -0,0 +1,21 @@
...
+<data_managers>
+    <data_manager tool_file="data_manager/data_manager_deeparg.xml" id="data_manager_deeparg">
+        <data_table name="deeparg_database_versioned">  <!-- Defines a Data Table to be modified. -->
+            <output> <!-- Handle the output of the Data Manager Tool -->
+                <column name="value"/>  <!-- columns that are going to be specified by the Data Manager Tool -->
+                <column name="name"/>  <!-- columns that are going to be specified by the Data Manager Tool -->
+                <column name="path" output_ref="out_file">
+                    <move type="directory">
+                        <source>${path}</source>
+                        <target base="${GALAXY_DATA_MANAGER_DATA_PATH}">deeparg_db/${path}</target>
```
bernt-matthias, line 11: This should contain `value`. Otherwise it would be overwritten if one downloads multiple times for the same version.

data_managers/data_manager_deeparg/data_manager_conf.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
+<data_managers>
+    <data_manager tool_file="data_manager/data_manager_deeparg.xml" id="data_manager_deeparg">
+        <data_table name="deeparg_database_versioned">  <!-- Defines a Data Table to be modified. -->
```
bernt-matthias, line 4: We could just go with `deeparg` here.

data_managers/data_manager_deeparg/data_manager_conf.xml
bernt-matthias: I would use the datatable name instead of `deeparg_db`

tools/deeparg/deeparg_predict.xml
```diff
@@ -0,0 +1,107 @@
...
+    && rm -r deeparg_1.0.4
+#end if
+##
+]]></command>
+    <inputs>
+        <param name="hide_db_build" type="hidden" value=""/>
+        <param name="input" type="data" format="fasta" label="Input file"/>
+        <param name="deeparg_db" type="select" label="DeepARG database">
+            <options from_data_table="deeparg_database_versioned">
+                <validator message="No deeparg database is available" type="no_options"/>
```
bernt-matthias, line 37: Should we filter for the version here?

data_managers/data_manager_deeparg/data_manager/data_manager_deeparg.xml
hugolefeuvre: Yes I think we can. 
The tool download all the versions (v1 and v2), and then you can choose version with a deeparg predict parameter` --model-version`
There is no information if future versions can use older downloads, actually all the versions are in the zenodo that is download by deeparg as database.

data_managers/data_manager_deeparg/data_manager_conf.xml
hugolefeuvre: I've made a few changes, but I'm not sure it's what you thought it would be. Is it possible to do a double check ?

tools/deeparg/deeparg_predict.xml
hugolefeuvre: What type of filter did you have in mind?

data_managers/data_manager_deeparg/data_manager/data_manager_deeparg.xml
bernt-matthias: Can you share a link to the zenodo record?

hugolefeuvre: https://zenodo.org/records/8280582

data_managers/data_manager_deeparg/data_manager_conf.xml
bernt-matthias: Looks good to me

tools/deeparg/deeparg_predict.xml
bernt-matthias: I was thinking to filter for the version, such that the corresponding tool version can only process the values generated with exactly this version .. not sure if it makes sense.

hugolefeuvre: To make sure that the only DB that can be used is the one that corresponds to the version of the tool?
Someting like that ? `<filter column="version" value="@TOOL_VERSION@"/>`

hugolefeuvre:
Hi @bernt-matthias and @bgruening, thank you for the review ! I wanted to know if there were any other modifications you'd like to make to this wrapper ?

data_managers/data_manager_deeparg/data_manager/data_manager_deeparg.xml
```diff
@@ -0,0 +1,60 @@
...
+    <macros>
+        <token name="@TOOL_VERSION@">1.0.4</token>
+        <token name="@VERSION_SUFFIX@">0</token>
+    </macros>
+    <requirements>
+        <requirement type="package" version="@TOOL_VERSION@">deeparg</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        mkdir -p '$out_file.extra_files_path' &&
+        deeparg download_data -o '$out_file.extra_files_path/deeparg_$version/' &&
```
bernt-matthias, line 12: Wondering if this should be 

```suggestion
        deeparg download_data -o '$out_file.extra_files_path/' &&
```

data_managers/data_manager_deeparg/data_manager_conf.xml
```diff
@@ -0,0 +1,21 @@
...
+        <data_table name="deeparg">  <!-- Defines a Data Table to be modified. -->
+            <output> <!-- Handle the output of the Data Manager Tool -->
+                <column name="value"/>  <!-- columns that are going to be specified by the Data Manager Tool -->
+                <column name="name"/>  <!-- columns that are going to be specified by the Data Manager Tool -->
+                <column name="path" output_ref="out_file">
+                    <move type="directory">
+                        <source>${path}</source>
+                        <target base="${GALAXY_DATA_MANAGER_DATA_PATH}">deeparg/${value}/${path}</target>
+                    </move>
+                    <value_translation>${GALAXY_DATA_MANAGER_DATA_PATH}/deeparg/${value}/${path}</value_translation>
```
bernt-matthias, line 13: Because this will copy the contents of `'$out_file.extra_files_path` to `${GALAXY_DATA_MANAGER_DATA_PATH}/deeparg/${value}/${path}` .. i.e. the extra `deeparg_$version` which is appended to the `'$out_file.extra_files_path` is not needed.

tools/deeparg/deeparg_predict.xml
```diff
@@ -0,0 +1,129 @@
...
+#if str($hide_db_build) == 'true':
+    deeparg download_data -o deeparg_1.0.4 &&
+#end if
+##
+    mkdir -p deeparg_predict_output &&
+    deeparg predict
+        --model '$model'
+        -i '$input'
+        -o 'deeparg_predict_output/deeparg_predict'
+        -d '$deeparg_db.fields.path'
```
bernt-matthias, line 19: But I think the main problem is then here, because the data is in a subfolder of `$deeparg_db.fields.path` called `deeparg_$version` (you could get the version from `$deeparg_db.fields.db_version`).

tools/deeparg/deeparg_predict.xml
```diff
@@ -0,0 +1,129 @@
...
+    <description>Antibiotic Resistance Genes (ARGs) from metagenomes</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="xrefs"/>
+    <expand macro="requirements"/>
+    <command detect_errors="exit_code"><![CDATA[
+##Used only for test 
+#if str($hide_db_build) == 'true':
+    deeparg download_data -o deeparg_1.0.4 &&
```
bernt-matthias, line 11: I think this should be
```suggestion
    deeparg download_data -o '$deeparg_db.fields.path' &&
```

or 

```suggestion
    deeparg download_data -o '$deeparg_db.fields.path'/deeparg_$deeparg_db.fields.db_version &&
```

bernt-matthias: Wondering how this could pass tests .. since so far the reference data was downloaded to the job working dir. 

Btw. how lagrge is the test data?

hugolefeuvre: The test data or the database ? 
Test data are 3,1M large and database that is not included into test data is 4,8G large (try to reduce it at maximum but I can't reduce it to less than 2G).

tools/deeparg/deeparg_predict.xml
bernt-matthias: This might be nice. But I do not now id needed. I leave it to you.

bernt-matthias (approved):

bebatut:
Thanks you so much @hugolefeuvre and @bernt-matthias!

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/13174139851

Issue #6547, opened 2024-11-11
is there a plan to upgrade r941_prom_hac_g360+g422 model to any of latest model soon?


mthang:
The Galaxy clair3 tool currently uses r941_prom_hac_g360+g422 model and r941_prom_sup_g5014 model. 
https://github.com/galaxyproject/tools-iuc/blob/0ee665c3393af083833fdb9becbe6965d009e16c/tools/clair3/tool-data/model.loc.sample#L10

I would like to find out if the model will be upgraded to any of the model listed here https://github.com/nanoporetech/rerio/tree/master/clair3_models . 

Thanks

pvanheus:
Yes! My plan is to do two things:
1. Propose on the bioconda side that the 4 [latest models](https://github.com/nanoporetech/rerio?tab=readme-ov-file#clair3-models) should be included in the bioconda recipe (i.e. downloaded at build time and added to the existing models folder)
2. Update the Galaxy tool

If (1) is not accepted on the bioconda side we will need to build a tool to download models and / or a data manager. I'd prefer not to go that route.

Does this sound reasonable to you?

mthang:
Totally. (1) is probably the better solultion. I would also avoid to deal with data manager if possible. Do you know when will the model available ?

bernt-matthias:
How big are the models?

bernt-matthias:
>  latest are 310 MB in size in total.

Then (to me) a data manager seems better than inclusion in bioconda.

> I know that some Galaxy servers have negotiated for the use of "Free for Academic Use" software on their servers. 

Yep. Sometimes a statement in the help / description was sufficient, sometime a boolean parameter where users need to agree to the license.

pvanheus:
@mthang the clair3 data manager and updated clair3 tool have now been added, with support for the latest models from the Rerio repository. Hopefully you will get to test them on a Galaxy server soon, then we can close this issue.

PR #6659, opened 2025-01-07
Add data manager for clair3 models

pvanheus:
As noted in #6547 the clair3 tool does not come with models for newer Oxford Nanopore flow cells (the R10 flow cells). These are provided by Oxford Nanopore on their Rerio page. This data manager fetches Clair3 models from that page.

FOR CONTRIBUTOR:

### DIFF
--- a/data_managers/data_manager_clair3_models/data_manager/install_clair3_models.xml
+++ b/data_managers/data_manager_clair3_models/data_manager/install_clair3_models.xml
@@ -1,6 +1,6 @@
-<tool id="data_manager_clair3_models" name="Clair3 model downloader" version="0.0.1" tool_type="manage_data" profile="23.2" />
+<tool id="data_manager_clair3_models" name="Clair3 model downloader" version="0.0.1" tool_type="manage_data" profile="23.2">
     <requirements>
-        <requirement type="package">python</requirement>
+        <requirement type="package" version="3.12">python</requirement>
     </requirements>
     <command detect_errors="exit_code"><![CDATA[
         python '$tool_directory/model_fetcher.py'
@@ -17,12 +17,14 @@
                 <option value="latest">Latest models from rerio page</option>
                 <option value="chosen">User provided list of models</option>
             </param>
-        <when value="latest">
-        </when>
-        <when value="chosen">
-            <param name="model_list" type="text" label="List of models to download" help="A space separated list of model to download, e.g. 'r1041_e82_400bps_sup_v430,r1041_e82_400bps_hac_v430'">
-                <validator type="regex" message="Invalid model list. Format is a space separated list of model names (e.g. 'r1041_e82_400bps_sup_v430,r1041_e82_400bps_hac_v430')" value="^[a-z_0-9,]+$"/>
-            </param>
+            <when value="latest">
+            </when>
+            <when value="chosen">
+                <param name="model_list" type="text" label="List of models to download" help="A space separated list of model to download, e.g. 'r1041_e82_400bps_sup_v430,r1041_e82_400bps_hac_v430'">
+                    <validator type="regex" message="Invalid model list. Format is a space separated list of model names (e.g. 'r1041_e82_400bps_sup_v430,r1041_e82_400bps_hac_v430')">^[a-z_0-9,]+$</validator>
+                </param>
+            </when>
+        </conditional>
     </inputs>
     <outputs>
         <data name="output_file" format="data_manager_clair3_models" label="Downloaded models" />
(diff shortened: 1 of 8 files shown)


data_managers/data_manager_clair3_models/data_manager/install_clair3_models.xml
```diff
@@ -0,0 +1,99 @@
+<tool id="data_manager_clair3_models" name="Clair3 model downloader" version="0.0.1" tool_type="manage_data" profile="23.2">
+    <requirements>
+        <requirement type="package" version="3.12">python</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+            ## this code looks up the existing table and uses it to build a list of known models
```
bernt-matthias, line 6: What is the `known_models` code doing?

pvanheus: It ensures that the same model is not downloaded twice (duplicates are not allowed in the data table).

data_managers/data_manager_clair3_models/tool_data_table_conf.xml.sample
```diff
@@ -0,0 +1,6 @@
+<tables>
+    <table name="clair3_models" comment_char="#" allow_duplicate_entries="False">
```
bernt-matthias, line 2: I would have assumed that Galaxy automatically does not allow duplicates if `allow_duplicate_entries="False"`.

bernt-matthias: Alternatively the `value_in_data_table` might be easier?

pvanheus: Nope, it lets you provide duplicate entries and then gives an error.

pvanheus: `value_in_data_table` works for manually entered values, but doesn't work for the `latest` mode  - i.e. where the DM downloads the latest models.

pvanheus:
In a discussion with Jack Jacobs of ONT this morning he confirmed that ONT would prefer some kind of wording associated with a checkbox that users can check to acknowledge that they comply with the terms of ONT's license for these models. He will revert with specific wording. The main concern from ONT is not a commercial user simply using the models but rather a commercial user using the models to compete with ONT - that said, I will wait on the wording.

He also recommended a direct model download tool that can download the .tar.gz version of the models to the user history to ensure that they have the latest version of the models. This will be a separate project.

Finally he confirmed that the model name uniquely identifies a model: i.e. there will not be "silent updates" of models.

In conclusion, from this discussion I think:
1. The wording is needed from ONT
2. This tool will (like BRAKER3) likely need to find a home outside of tools-iuc.

bernt-matthias:
> This tool will (like BRAKER3) likely need to find a home outside of tools-iuc.

Not sure. Why?

pvanheus:
I'm noting this response that I got via Jack on the Oxford Nanopore support channel (email to me on my UWC address, Thursday 30 January 2025, 11:47am).

"I can confirm that Rerio is only meant to be used for internal research only and specifically excludes uses directed towards commercial advantages or monetary compensation (e.g. including it as part of a service/product that is billed for).
We can allow Peter Van Heusden from the South African National Bioinformatics Institute, to use it for the Galaxy Project as long as our license terms are included and users have to click and agree these terms before they can use it (much like the example provided in your email below).

The example I gave in the reference is : “I agree to the terms of this Licence agreement”"

To make things simpler for tool developers, I've added a `source` column to the data table, to identify where the models came from. Perhaps in the future there will be other sources besides the Rerio ones, and they can be handled differently depending on licensing.

pvanheus:
> > The main concern from ONT is not a commercial user simply using the models but rather a commercial user using the models to compete with ONT - that said, I will wait on the wording.
> 
> Well, yes, lets wait for that, but as long as the model is supplied via a data manager any user, commercial or not, will not have direct access to it. That would be different of course with the model download tool he suggested.

Understood. I have thus ticked the "License permits unrestricted use" box.

data_managers/data_manager_clair3_models/data_manager/model_fetcher.py
```diff
@@ -0,0 +1,147 @@
...
+                break1 = line.find(' ')
+                read_lines = True
+                continue
+            if init_line_seen and line.startswith('Config'):
+                config_line_seen = True
+                continue
+            if line.startswith('Clair3 models for the following configurations are available:'):
+                init_line_seen = True
+                continue
+        return models
```
wm75, line 46: puh, that looks very fragile.
A suggestion to make things a tiny bit more robust. Can you:
- rely on the `Clair3 Models` line to detect the correct section (instead of on `Clair3 models for the following configurations are available:`, which looks like it could be changed at any time)?
- add `Latest` to the required tokens before looking for `Config`?

pvanheus: Changed as suggested. I also look for just `=====` now because the point is the line starting with multiple `=`, not exactly how many there are.

data_managers/data_manager_clair3_models/data_manager/model_fetcher.py
```diff
@@ -0,0 +1,171 @@
...
+        # ========================== =================== =======================
+        # r1041_e82_400bps_sup_v500  R10.4.1 E8.2 (5kHz) v5.0.0 SUP
+        # r1041_e82_400bps_hac_v500  R10.4.1 E8.2 (5kHz) v5.0.0 HAC
+        # r1041_e82_400bps_sup_v410  R10.4.1 E8.2 (4kHz) v4.1.0 SUP
+        # r1041_e82_400bps_hac_v410  R10.4.1 E8.2 (4kHz) v4.1.0 HAC
+        # ========================== =================== =======================
+        #
+        # and the aim is to extract the list of model names from the table by successfully looking for
+        # "Clair3 Models", then "Latest:", then "Config" and then "=====" and then reading the lines until
+        # the next "=====" is encountered
```
wm75, line 49: nice, thanks!

data_managers/data_manager_clair3_models/data_manager/model_fetcher.py
```diff
@@ -0,0 +1,171 @@
...
+        for line in StringIO(data):
+            if read_lines:
+                if line.startswith('====='):
+                    read_lines = False
+                    break
+                model = line[:break1 - 1]
+                models.append(model)
+            if config_line_seen and line.startswith('====='):
+                break1 = line.find(' ')  # the first space is the break between the model name column and the rest of the line
```
wm75, line 58: wasn't paying close attention to this before, sorry.
While this happens to work with the current latest models, it would break for several of the deprecated ones where names either have the same length as the cell frame or are shorter than what you are expecting here.
Why not simply do: `line.split()[0]` to extract the model id?

pvanheus: I don't recall why I didn't use this simpler option, and have now implemented it. BTW note that this code is only extracting a list of the latest models.

PR #6674, opened 2025-01-13
drep: require at least two input genomes

bernt-matthias:
xref: https://github.com/MrOlm/drep/issues/249

FOR CONTRIBUTOR:

### DIFF
--- a/tools/drep/macros.xml
+++ b/tools/drep/macros.xml
@@ -22,9 +22,7 @@
     </xml>
 
     <xml name="genomes">
-        <param argument="--genomes" type="data" format="fasta" multiple="true" label="Genomes" help="One genome per input fasta">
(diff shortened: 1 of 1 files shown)


tools/drep/macros.xml
```diff
@@ -22,7 +22,9 @@
     </xml>
 
     <xml name="genomes">
-        <param argument="--genomes" type="data" format="fasta" multiple="true" label="Genomes to filer"/>
+        <param argument="--genomes" type="data" format="fasta" multiple="true" label="Genomes" help="One genome per input fasta">
+            <validator type="dataset_metadata_equal" metadata_name="sequences" value="1" message="Input FASTA needs to contain 1 genome per file"/>
```
bebatut, line 26: How does that work when you have a genome with multiple contigs?

bernt-matthias: Good question. Then I guess that my problem was that I just had one input dataset. 

In this case a better constraint might be that we require at least two input datasets?

PR #6679, opened 2025-01-16
Compleasm: use symlink instead of copying busco data

abretaud:
FOR CONTRIBUTOR:

I looked at the compleasm code, and found this way to avoid copying the busco dataset (which can take time and make compleasm jobs longer than busco jobs)


tools/compleasm/compleasm.xml
```diff
@@ -11,17 +11,17 @@
     </requirements>
     <version_command>compleasm --version</version_command>
     <command><![CDATA[
-    
-    mkdir -p galaxy_db &&
-    cp -r '${busco_database.fields.path}/lineages/${lineage_dataset}/' 'galaxy_db/' &&
+    mkdir -p 'galaxy_db/' &&
+    ln -s '${busco_database.fields.path}/lineages/${lineage_dataset}/' 'galaxy_db/${lineage_dataset}' &&
+    touch 'galaxy_db/${lineage_dataset}.done' &&
```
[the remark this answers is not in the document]
abretaud: It's just an empty file specific to compleasm, to make it understand that it should not redownload it, I'm gonna add a comment

tools/compleasm/test-data/busco_database.loc
```diff
@@ -2,5 +2,5 @@
 # - value
 # - name
 # - version
-# - /path/to/data 
-eukaryota_odb10	eukaryota	5.4.6	${__HERE__}/test-db/busco_downloads
```
bernt-matthias, line 6: May I ask why you updated the test (data)?

abretaud: to keep the test-data dir as small as possible as entomoplasmatales_odb10 is a much smaller lineage than eukaryota_odb10

bernt-matthias: True, but changing it will increase the size of the repo. The "problem" with `git` repos is that everything that is in in will be there forever.

bernt-matthias: But changing it might still be a good idea if the runtime is reduced significantly.

bgruening (approved):

tools/compleasm/test-data/busco_database.loc
abretaud: the previous dataset was non-functional and not really used, so I see no other way than adding this new minimal one

PR #6683, opened 2025-01-20
Add Cnvkit export subcomand tools to convert the CNVkit output into suitable formats

khaled196:
FOR CONTRIBUTOR:


tools/cnvkit/cdt.xml
```diff
@@ -0,0 +1,32 @@
+<tool id="cnvkit_cdt" name="CNVkit CDT" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="21.05">
+    <description>Convert log2 ratios to CDT format. Compatible with Java TreeView</description>
```
bgruening, line 2:
```suggestion
    <description>Convert log2 ratios to CDT</description>
```

Maybe add for what CDT stands for?

tools/cnvkit/jtv.xml
```diff
@@ -0,0 +1,32 @@
+<tool id="cnvkit_jtv" name="CNVkit JTV" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="21.05">
+    <description>Convert log2 ratios to Java TreeView's native format</description>
```
bgruening, line 2: Out of interest, do we have Jave TreeView already in Galaxy? Should we add it?

tools/cnvkit/jtv.xml
```diff
@@ -0,0 +1,32 @@
...
+    <expand macro="creators"/>
+    <expand macro="requirements"/>
+    <command detect_errors="exit_code"><![CDATA[  
+        ln -s '$input_segmented_file' ./sample.cnr &&
+        cnvkit.py export jtv
+            ./sample.cnr
+            --output sample.cnv.jtv
+    ]]></command>
+     <inputs>
+        <param name="input_segmented_file" type="data" format="tabular" label="Log2 Copy Ratio Data File" help="The output of the CNVkit 'fix' sub-command" />
```
bgruening, line 16: if this is the only valid input format, maybe we should add this functionality to the CNVkit fix tool?

tools/cnvkit/seg.xml
```diff
@@ -0,0 +1,35 @@
+<tool id="cnvkit_seg" name="CNVkit SEG" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="21.05">
+    <description>Convert segments to SEG format. Compatible with IGV and GenePattern</description>
```
bgruening, line 2:
```suggestion
    <description>Convert segments to SEG format</description>
```

tools/cnvkit/seg.xml
```diff
@@ -0,0 +1,35 @@
+<tool id="cnvkit_seg" name="CNVkit SEG" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="21.05">
```
bgruening, line 1: should the tool ideas include export? e.g. `cnvkit_export_seg`?

tools/cnvkit/jtv.xml
khaled196: this file "Log2 Copy Ratio Data File" can be generated from the batch (a toolkit pipeline that has the fix in it) and some other tools but the fix is the last step in preparing the data.

tools/cnvkit/jtv.xml
bgruening: Just keep in mind, that we added Java applications inside a virtual desktop already to Galaxy for other communities. e.g. https://github.com/usegalaxy-eu/docker-qupath

Whenever this is a need, we can add it.

tools/cnvkit/test-data/sample.cnv.vcf
```diff
@@ -1,5 +1,5 @@
 ##fileformat=VCFv4.2
-##fileDate=20250116
+##fileDate=20250120
```
bgruening, line 2: if this is only working today this is not good :)

Please add a lines_diff to the test :)

bgruening: Please use text comparison. Comparing by size is the worth test we can use.
Text comparisons can have differences. Please look at `lines_diff=2` for example.

khaled196:
@bgruening  I have a question not related to this pull request.

The next step is to use an existing beacon to check the variants if they are unique or not.

We are planning to use Progentix through API (https://docs.progenetix.org/services/#cytoband-mapping-cytomapper)

Is there an example of how to wrap this as a tool on galaxy

The tool will compare the variants generated with the variants listed there and highlight the unique variants.

PR #6684, opened 2025-01-21
Add drug2cell tool 

Mennayousef:
Wrapping drug2cell tool into Galaxy


tools/drug2cell/drug2cell.xml
```diff
@@ -0,0 +1,107 @@
+<tool id="drug2cell" name="drug2cell" version="0.1.2+galaxy0" python_template_version="3.5" profile="21.05">
```
bgruening, line 1:
```suggestion
<tool id="drug2cell" name="drug2cell" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="21.05">
```

tools/drug2cell/drug2cell.xml
```diff
@@ -0,0 +1,107 @@
...
+    <requirements>
+        <requirement type="package" version="1.0">drug2cell</requirement>
+        <requirement type="package" version="1.0">scanpy</requirement>
+        <requirement type="package" version="1.0">blitzgsea</requirement>
+    </requirements>
+
+    <command>
+        <![CDATA[
+        # Python script to execute drug2cell scoring and differential expression
+        python3 <<CODE
```
bgruening, line 19: you can use a `configfile` in Galaxy to store your code and reuse this.

https://github.com/galaxyproject/tools-iuc/blob/51995527835d18a4a3a498b140382ce433f19a71/tools/scanpy/inspect.xml#L55

tools/drug2cell/drug2cell.xml
```diff
@@ -0,0 +1,107 @@
...
+
+    <help>
+        <![CDATA[
+        This tool uses the drug2cell package to score gene groups based on drug-target data from the ChEMBL database. 
+        It performs differential expression analysis to identify gene groups up-regulated in particular clusters and visualizes the results in UMAP and dotplot formats.
+        ]]>
+    </help>
+
+    <citations>
+        <citation type="bibtex">@article{kanemaru2023spatially,
```
bgruening, line 96: You can use here `type="doi"` and use a doi instead.

bgruening (commented):
Thanks @Mennayousef! A few comments inside, looks very good!

Mennayousef:
@bgruening Could you please help? Why the checks are failing?

tools/drug2cell/.shed.yml
```diff
@@ -0,0 +1,9 @@
+name: drug2cell
+owner: iuc
+description: Collection of utility functions for gene group activity evaluation in scanpy.
+homepage_url: https://github.com/Teichlab/drug2cell
+long_description: |
+  This tool uses the drug2cell package to score gene groups based on drug-target data from the ChEMBL database. It performs differential expression analysis to identify gene groups up-regulated in particular clusters and visualizes the results in UMAP and dotplot formats.
+remote_repository_url: https://github.com/Teichlab/drug2cell
+categories:
+- Single cell
```
bgruening, line 9:
```suggestion
- Single Cell
```

tools/drug2cell/drug2cell.xml
```diff
@@ -0,0 +1,103 @@
...
+
+    <command detect_errors="exit_code"><![CDATA[
+      @CMD@
+      ]]></command>
+    
+    <configfiles>
+        <configfile name="script_file"><![CDATA[
+            @CMD_IMPORTS@
+            @CMD_READ_INPUTS@
+
```
bgruening, line 24:
```suggestion
```

This are macros, you are not using them, isn't it?

tools/drug2cell/drug2cell.xml
```diff
@@ -0,0 +1,103 @@
...
+    </macros>
+
+    <requirements>
+        <requirement type="package" version="1.0">drug2cell</requirement>
+        <requirement type="package" version="1.0">scanpy</requirement>
+        <requirement type="package" version="1.0">blitzgsea</requirement>
+    </requirements>
+
+    <command detect_errors="exit_code"><![CDATA[
+      @CMD@
```
bgruening, line 17:
```suggestion
     python '$script_file'
```

You need to run now your python file that you created with the configfile.

tools/drug2cell/drug2cell.xml
```diff
@@ -0,0 +1,100 @@
...
+        sc.pl.umap(adata.uns['drug2cell'], color="${drug}", color_map="OrRd", save="umap_plot.png")
+
+        # Perform differential expression analysis
+        sc.tl.rank_genes_groups(adata.uns['drug2cell'], method="wilcoxon", groupby="louvain")
+        sc.pl.rank_genes_groups_dotplot(adata.uns['drug2cell'], swap_axes=True, dendrogram=False, n_genes=${n_genes}, save="dotplot1.png")
+
+        # Prepare arguments for dotplot visualization
+        plot_args = d2c.util.prepare_plot_args(adata.uns['drug2cell'], categories="${categories}".split(","))
+        sc.pl.dotplot(adata.uns['drug2cell'], groupby="louvain", swap_axes=True, **plot_args, save="dotplot2.png")
+        @CMD_ANNDATA_WRITE_OUTPUTS@
```
bgruening, line 43: This is another macro ... if you are not defining them, please remove them. It was just an example I provide.

tools/drug2cell/drug2cell.xml
bgruening: Oh it seems you have reverted the configfile change.

Mennayousef: yeah, it was just a test to see if without configfile will pass all checks, unfortunately it is still failing the test. with this error "Test output file (expected-output/dotplot2.png) is missing. If you are using planemo, try adding --update_test_data to generate it." I tried to manually create the expected-output folder, but no luck. Although the test passes when I run local galaxy instance.

bgruening (commented):
remove the folder expected-output and put the dotplot1 and the umap_plot.png into the test-data folder

tools/drug2cell/drug2cell.xml
```diff
@@ -0,0 +1,98 @@
...
+    <!-- Test case 1 -->
+        <test>
+            <param name="adata" value="../test-data/pbmc3k_processed.h5ad" />
+            <param name="use_raw" value="true" />
+            <param name="drug" value="CHEMBL1743048|OBINUTUZUMAB" />
+            <param name="n_genes" value="10" />
+            <param name="categories" value="B01,B02,B03" />
+            <output name="umap_plot" file="test-data/umap_plot.png" />
+            <output name="dotplot1" file="test-data/dotplot1.png" />
+            <output name="dotplot2" file="test-data/dotplot2.png" />
```
bgruening, line 84:
```suggestion
            <param name="adata" value="pbmc3k_processed.h5ad" />
            <param name="use_raw" value="true" />
            <param name="drug" value="CHEMBL1743048|OBINUTUZUMAB" />
            <param name="n_genes" value="10" />
            <param name="categories" value="B01,B02,B03" />
            <output name="umap_plot" file="umap_plot.png" />
            <output name="dotplot1" file="dotplot1.png" />
            <output name="dotplot2" file="dotplot2.png" />
```

bgruening:
@Mennayousef if you get stuck, please leave me a message and I will try to generate those pngs.

Mennayousef:
> @Mennayousef if you get stuck, please leave me a message and I will try to generate those pngs.

Would be greatly appreciated!

bgruening:
I have found many small mistakes, that I tried to fix. If you now run planemo locally, I get

> KeyError: 'Could not find key CHEMBL1743048XOBINUTUZUMAB in .var_names or .obs.columns.'

I run planemo with:

> planemo t --biocontainers --update_test_data .

tools/drug2cell/drug2cell.xml
```diff
@@ -0,0 +1,97 @@
+<tool id="drug2cell" name="drug2cell" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="21.05">
+    <description>Collection of utility functions for gene group activity evaluation in scanpy</description>
+    <macros>
+        <token name="@TOOL_VERSION@">0.1.2</token>
+        <token name="@VERSION_SUFFIX@">0</token>
+    </macros>
+    <requirements>
+        <requirement type="package" version="0.1.2">drug2cell</requirement>
```
bgruening, line 8: Those version number needs to match the conda versions on anaconda.org. I changed them, please double check.

tools/drug2cell/drug2cell.xml
```diff
@@ -0,0 +1,97 @@
...
+        <data name="dotplot2" format="png" label="Dotplot with Custom Categories" from_work_dir="dotplot2.png"/>
+    </outputs>
+    <tests>
+        <!-- Test case 1 -->
+        <test>
+            <param name="adata" value="pbmc3k_processed.h5ad"/>
+            <param name="use_raw" value="true"/>
+            <param name="drug" value="CHEMBL1743048|OBINUTUZUMAB"/>
+            <param name="n_genes" value="10"/>
+            <!-- todo please add cetegories, how do users know which categories are valid? -->
```
bgruening, line 67: see this todo here please

tools/drug2cell/drug2cell.xml
```diff
@@ -0,0 +1,97 @@
...
+            <param name="categories" value="B01,B02,B03"/>
+            <output name="umap_plot" file="umap_plot.png"/>
+            <output name="dotplot1" file="dotplot1.png"/>
+            <output name="dotplot2" file="dotplot2.png"/>
+        </test>
+        <!-- is it possible to use the tiny dataset? -->
+        <test>
+            <param name="adata" value="pbmc3k_tiny.h5ad"/>
```
bgruening, line 75: I also added a tiny test file that I found. I'm not sure if you can use that instead. Can you try?

tools/drug2cell/drug2cell.xml
```diff
@@ -0,0 +1,97 @@
...
+        <!-- Test case 1 -->
+        <test>
+            <param name="adata" value="pbmc3k_processed.h5ad"/>
+            <param name="use_raw" value="true"/>
+            <param name="drug" value="CHEMBL1743048|OBINUTUZUMAB"/>
+            <param name="n_genes" value="10"/>
+            <!-- todo please add cetegories, how do users know which categories are valid? -->
+            <param name="categories" value="B01,B02,B03"/>
+            <output name="umap_plot" file="umap_plot.png"/>
+            <output name="dotplot1" file="dotplot1.png"/>
```
bernt-matthias, line 70: The file `dotplot1.png` is needed in the `test-data/` folder.

tools/drug2cell/drug2cell.xml
```diff
@@ -0,0 +1,97 @@
...
+sc.pl.dotplot(adata.uns['drug2cell'], groupby="louvain", swap_axes=True, **plot_args, save="dotplot2.png")
+        </configfile>
+    </configfiles>
+    <inputs>
+        <!-- Input: AnnData object (processed scRNA-seq data) -->
+        <param name="adata" type="data" format="h5ad" label="AnnData Object" help="Select the processed AnnData object file containing your single-cell RNA-seq data"/>
+        <!-- Option to select number of genes to visualize in the dotplot -->
+        <param name="n_genes" type="integer" min="1" max="20" label="Number of Genes for Dotplot" help="Specify the number of genes to display in the dotplot (range: 1 to 20)."/>
+        <!-- Option for visualizing a specific drug from ChEMBL -->
+        <param name="drug" type="text" label="Drug (CHEMBL ID)" help="Enter the ChEMBL ID of the drug to visualize."/>
```
bernt-matthias, line 46: If this is mandatory you should add a validator as shown in the 2nd example here: https://docs.galaxyproject.org/en/master/dev/schema.html#id42. Text parameters are special, since an empty string is also a string.

tools/drug2cell/drug2cell.xml
```diff
@@ -0,0 +1,97 @@
...
+        <!-- Option to perform differential expression analysis -->
+        <param name="use_raw" type="boolean"  truevalue="True" falsevalue="False" label="Use Raw Data for Drug Scoring" help="Choose whether to use raw data or normalized data for drug scoring."/>
+        <!-- Option to categorize data for dotplot -->
+        <param name="categories" type="text" label="Categories for Dotplot" help="Enter a comma-separated list of categories to display in the dotplot."/>
```
bernt-matthias, line 50: Are the categories stored in the h5ad file .. in particular any of the h5 metadata that Galaxy stores: https://github.com/galaxyproject/galaxy/blob/bdea7bcbede782cedcc9dd365d45adb42fdd1312/lib/galaxy/datatypes/binary.py#L1496

Can this be empty? If no: also a validator would be good. If yes: should dotplot2 also be produced?

Mennayousef: Categories is an argument in drug2cell tool, to subset drugs based on their ATC categorisation (level 1 and level 2) https://en.wikipedia.org/wiki/Anatomical_Therapeutic_Chemical_Classification_System I think I should add that to the help section of the categories.

bernt-matthias: Seems that this is a fixed set of possibilities. Then a select might be better than a text. 

The possible values might also to be in 
https://github.com/Teichlab/drug2cell/blob/a786c632fc521b94f597e452e2f09c45fd451640/drug2cell/data.py#L12

tools/drug2cell/drug2cell.xml
Mennayousef: Thanks @bgruening and @bernt-matthias for your help! I made some changes to the categories parameter, so that users can choose level 1 and level 2, then the python code should combine the entries to create the category entry. I think this is more user-friendly. However, still no luck passing the test.

tools/drug2cell/drug2cell.xml
Mennayousef: @bernt-matthias Do you mean we should create the files in the test-data folder?

bernt-matthias: Yes.

tools/drug2cell/drug2cell.xml
```diff
@@ -0,0 +1,127 @@
...
+    <configfiles>
+        <configfile name="script">
+import scanpy as sc
+import drug2cell as d2c
+
+# Load AnnData object
+adata = sc.read_h5ad("${adata}")
+
+# Combine ATC Level 1 and Level 2 into a single string
+categories = f"{level1}{level2}"
```
bernt-matthias, line 26: In order to define / modify cheetah variabkls you need to 
```suggestion
#set categories = f"{level1}{level2}"
```

The error is hidden in the logs: https://github.com/galaxyproject/tools-iuc/actions/runs/12985253324/job/36209809943#step:9:818

tools/drug2cell/drug2cell.xml
Mennayousef: I didn't find conda package for drug2cell tool, so I added one in bioconda and it was successfully merged last month. https://github.com/bioconda/bioconda-recipes/blob/master/recipes/drug2cell/meta.yaml

tools/drug2cell/drug2cell.xml
Mennayousef: I changed this file with fake_small_data.h5ad, it is the same concept of using a small file, but I manually added the desired drug for visualisation.

tools/drug2cell/drug2cell.xml
Mennayousef: Added the 3 required outputs as empty .png files in test-data folder

tools/drug2cell/drug2cell.xml
```diff
@@ -0,0 +1,113 @@
...
+    <configfiles>
+        <configfile name="script">
+import scanpy as sc
+import drug2cell as d2c
+
+# Load AnnData object
+adata = sc.read_h5ad("${adata}")
+
+# Combine ATC Level 1 and Level 2 into a single string
+set categories = f"{level1}{level2}"
```
bernt-matthias, line 26:
```suggestion
```

tools/drug2cell/drug2cell.xml
```diff
@@ -0,0 +1,113 @@
...
+
+# Visualize the specified drug (UMAP)
+sc.pl.umap(adata.uns['drug2cell'], color="${drug}", color_map="OrRd", save="umap_plot.png")
+
+# Perform differential expression analysis
+sc.tl.rank_genes_groups(adata.uns['drug2cell'], method="wilcoxon", groupby="louvain")
+sc.pl.rank_genes_groups_dotplot(adata.uns['drug2cell'], swap_axes=True, dendrogram=False, n_genes=${n_genes}, save="dotplot1.png")
+
+# Prepare arguments for dotplot visualization
+plot_args = d2c.util.prepare_plot_args(adata.uns['drug2cell'], categories="${categories}")
```
bernt-matthias, line 39:
```suggestion
plot_args = d2c.util.prepare_plot_args(adata.uns['drug2cell'], categories="${level1}${level2}")
```

bernt-matthias:
Cool. Now the actual tool is running and we get an error from it.

tools/drug2cell/drug2cell.xml
```diff
@@ -0,0 +1,67 @@
...
+]]>
+    </configfile>
+ </configfiles>
+
+ <inputs>
+     <param name="adata" type="data" format="h5ad" label="AnnData Object" help="Select the processed AnnData object file containing your single-cell RNA-seq data"/>
+ </inputs>
+
+ <outputs>
+    <data name="d2c_output" format="h5ad" label="drug2cell Results" from_work_dir="d2c_output.h5ad"/>
```
bgruening, line 43:
```suggestion
    <data name="d2c_output" format="h5ad" label="drug2cell Results"/>
```

bgruening: You are writing to the expected output directly in `d2cell.write("${d2c_output}")` and not to the file called `d2c_output.h5ad`

tools/drug2cell/drug2cell.xml
```diff
@@ -0,0 +1,67 @@
...
+ </outputs>
+
+ <tests>
+    <test>
+        <param name="adata" value="tiny_adata.h5ad"/>
+        <output name="d2c_output" file="d2c_output.h5ad"/>
```
bgruening, line 49: You could add here some asserts tests and check for specific keys etc in your h5ad file:

https://docs.galaxyproject.org/en/latest/dev/schema.html#has-h5-attribute

Mennayousef: <tests>
    <test>
        <param name="adata" value="tiny_adata.h5ad"/>
        <output name="d2c_output" file="d2c_output.h5ad">
            <has_h5_attribute key="obs"/>
            <has_h5_attribute key="var"/>
        </output>
    </test>
</tests>

bernt-matthias: This should be something like: 

```xml
        <output name="d2c_output" file="d2c_output.h5ad">
         <has_contents>
          <has_h5_attribute key="obs"/>
          <has_h5_attribute key="var"/>
         </has_contents>
       </output>
```

`has_contents` was missing and evertthingt needs to be in `output` (before you closed it immediately).

Mennayousef: ERROR (XSD): Invalid XML: Element 'has_h5_attribute': This element is not expected. Expected is one of ( element, discovered_dataset, assert_contents, extra_files, metadata ). 

Unfortunately this doesn't work if I added them to tests section, or outputs, or both. 

Do you have a working example on h5ad file ? Otherwise I think I will leave it as it is before adding has_h5_attribute

bernt-matthias: See here: https://github.com/galaxyproject/galaxy/blob/223f058efae1a0eca4e4b4632da0805b8758c7c1/test/functional/tools/validation_hdf5.xml#L15

bgruening:
> assert_has_h5_attribute() missing 1 required positional argument: 'value'

You are missing the value.

Mennayousef:
Hi @bgruening,
In the documentation, it should also work without providing a value. I also tried with value=True but it didn't work. If I gave any other value it will be so tailored to my test dataset and the tool will lack generalizability. I also do not see the point of checking obs and var presence, as those are essentials in any anndata object.

Mennayousef:
Finally, it passed! I had to change has_h5_attribute to has_h5_keys as obs and var are datasets in the file and not attributes. Many thanks @bgruening and @bernt-matthias for your great support!!!

bgruening (approved):

PR #6693, opened 2025-01-24
Update artic to v.1.6.0

ammaraziz:
FOR CONTRIBUTOR:

This pull request updates the `artic` wrapper to the latest version (1.6.0), incorporates the clair3 models data manager and also allows the user to manually specify the model from their history. There is one test that is purposefully commented out, it is to test specifying the model from history. The clair3 models are too big and are restricted, therefore they can not be included in the test-data. 

Pinging @pvanheus as he is/was working on the clair3 data manager.

There was a minor change to the `gupplyplex` wrapper to set the min/max as optional parameters. This makes writing workflows easier with optional inputs with a default value.

 I have ticked unrestricted checkbox even though it relies on the clair3 tools and the ONT models. The tool itself is MIT licensed. 

As always open to any changes!


pvanheus:
The [linting failure](https://github.com/galaxyproject/tools-iuc/actions/runs/12945279093/job/36107750021?pr=6693#step:5:425) is on a URL which is valid. Is there a way to re-run the tests?

ammaraziz:
Linting fails due to 404:

```
.. ERROR: Error '404 Client Error: Not Found for url: https://labs.primalscheme.com/' accessing https://labs.primalscheme.com/
```
But that URL works? The URL is mentioned in the HELP.

tools/artic/artic_minion.xml
```diff
@@ -83,67 +124,214 @@
...
+
+ONT provides models for some latest or specific chemistries and basecallers through `Rerio`_. 
+
+.. class:: warningmark
+
+**Please note:** By using these models you agree to ONT licensing which you `can read here`_.
+
+**Where to find amplicon schemes?**
+
+The central location for publically available amplicon schemes is the primalscheme website: https://labs.primalscheme.com/.
```
pvanheus, line 295: Since this is the line containing the URL that is causing a linting error, perhaps try and replace the URL with a reference (e.g. `primalscheme website_` with a `.. _primalscheme website: https://labs.primalscheme.com` further down and see if that fixes things?

ammaraziz:
The issue was the trailing `/`  in `https://labs.primalscheme.com/`. Ready for review!

tools/artic/artic_guppyplex.xml
```diff
@@ -67,10 +68,23 @@ ln -s '$elem' inputs/${i}.${elem.ext} &&
                 help="Multiple datasets selected here will get combined into a single output for a single assumed sample. Select a nested list to have its inner lists interpreted as data from one sample each and to obtain one output per inner list." />
             </when>
         </conditional>
-        <param name="max_length" type="integer" label="Remove reads longer than" value="700" help="remove reads greater than this number of base pairs" />
-        <param name="min_length" type="integer" label="Remove reads shorter than" value="400" help="remove reads less than this number of base pairs" />
-        <param name="min_quality" type="integer" min="0" value="7"
-        label="Eliminate reads with a mean base quality score of less than"
+        <param name="max_length"
+        optional="true"
```
bernt-matthias, line 72: Can you revert the formatting here. 

Why is this now optional? This would also require a change in the command block.

But it might be cool to add min/max where possible.

tools/artic/artic_guppyplex.xml
```diff
@@ -80,18 +94,24 @@ ln -s '$elem' inputs/${i}.${elem.ext} &&
         <test>
             <conditional name="input">
                 <param name="structure" value="one_to_one" />
-                <param name="reads" value="test.fastq" />
+                <param name="reads" value="test.fastq.gz" />
             </conditional>
-            <output name="output" file="gupplyplex_output.fastq"/>
+            <output name="output">
+                <assert_contents>
+                    <has_size value="72300" delta="1000" />
```
bernt-matthias, line 101: What has changed here? Is this expected?

tools/artic/artic_minion.xml
```diff
@@ -6,67 +6,108 @@
...
-            <param name="primer_scheme_source_selector" type="select" label="Select a primer scheme from your history or use one from a tool data table?"
-                   help="Screening files must be stored in the 'primer_scheme_bedfiles' tool data table">
+            <param name="primer_scheme_source_selector" type="select"
+            label="Select primer file"
+            help="See Help section below for how to find primer scheme data files.">
                 <option value="tool_data_table">From tool data table</option>
                 <option value="history">From history</option>
             </param>
             <when value="tool_data_table">
-                <param name="primer_scheme_bedfile" type="select" format="tabular" label="Primer Scheme">
```
bernt-matthias, line 58: Why rename the parameter?

tools/artic/artic_minion.xml
```diff
@@ -6,67 +6,108 @@
...
 	            <options from_data_table="primer_scheme_bedfiles">
-	                <validator type="no_options" message="No primer scheme .bed files are available" />
-                    </options>
+	                <validator type="no_options" message="No primer.bed files are available" />
+                </options>
 	        </param>
             </when>
             <when value="history">
-                <param name="primer_scheme_bedfile" type="data" format="tabular" label="Primer Scheme" />
+                <param name="bed" type="data" format="tabular" label="Primer Scheme"
```
bernt-matthias, line 102: The format is tabular - why does the help state to upload a `tar.gz`?

tools/artic/artic_minion.xml
```diff
@@ -83,67 +124,215 @@
...
+            <when value="built-in">
+                <param type="select" name="select_built_in" label="model" >
+                    <option value="r941_prom_sup_g5014">r941_prom_sup_g5014</option>
+                    <option value="r941_prom_hac_g360+g422">r941_prom_hac_g360+g422</option>
+                </param>
+            </when>
+            <when value="datatable">
+                <param  name="model" type="select" label="Cached model from server"
+                help="See Help section below for instructions on how to select the appropriate model for you dataset.">
+                    <options from_data_table="clair3_models">
```
bernt-matthias, line 145: The datatable needs to be defined in https://github.com/galaxyproject/tools-iuc/blob/main/tools/artic/tool_data_table_conf.xml.sample (and the corresponding test file).

tools/artic/artic_minion.xml
```diff
@@ -83,67 +124,215 @@
                        help="You can upload a FASTA sequence to the history and use it as reference" />
             </when>
         </conditional>
-        <param type="text" name="medaka_model" label="Medaka model" help="Model string to pass to medaka (see https://github.com/nanoporetech/medaka#models)">
-            <validator type="expression" message="Please specify a valid medaka model string (see https://github.com/nanoporetech/medaka#models)">(len(value.strip().split('_')) == 3 or len(value.strip().split('_')) == 4) and value.strip().startswith('r')</validator>
-        </param>
+
+        <!-- Model selection -->
+        <conditional name="model_source">
+            <param name="source" type="select" label="Clair3 model"
```
bernt-matthias, line 130: Having tests for all choices would be good.

tools/artic/artic_minion.xml
```diff
@@ -83,67 +124,215 @@
...
+                    <has_size value="3200" delta="200" />
+                </assert_contents>
+            </output>
         </test>
-        <test>
-            <param name="reference_source_selector" value="tool_data_table" />
-            <param name="read_file" value="SRR11410539_seqtk_sample_500_1.fastq" />
-            <param name="reference" value="test_entry" />
+        <!-- Test: select clair3 model from history -->
+        <!-- Note: this test can only run locally due to the size of the model -->
```
bernt-matthias, line 254: How big is it?

tools/artic/artic_guppyplex.xml
ammaraziz: I've changed the formatting to be similar to what it was before. 

> Why is this now optional? This would also require a change in the command block.

It's optional with a default value, which is near identical to mandatory. There's no need to change the command block because if not set by the user, the default is chosen. Unless I am wrong in how I think this works, please correct me. 

It's mainly for for writing workflows with a default optional inputs, instead of mandatory, which to my knowledge doesn't allow defaults. Here's what Im trying to achieve but with `guppyplex` instead of `fastp`:

image

I can't do this in a straight forward manner if the min/max is mandatory.

tools/artic/artic_guppyplex.xml
ammaraziz: Something about how guppyplex is combining the fastq files results in slight changes. Im not really sure why it does this.

tools/artic/artic_minion.xml
ammaraziz: Oh good catch, this is the help transposed from another.

tools/artic/artic_minion.xml
ammaraziz: Personally preference, happy to revert back if you think its better to be more explicit here.

tools/artic/artic_minion.xml
ammaraziz: I do have tests for built-in, data-table and history. The last one is commented out though.

tools/artic/macros.xml
```diff
@@ -1,12 +1,19 @@
 <macros>
-    <token name="@PACKAGE_VERSION@">1.2.3</token>
+    <token name="@PACKAGE_VERSION@">1.6.0</token>
```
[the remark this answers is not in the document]
ammaraziz: Thanks for spotting this. Done.

tools/artic/artic_minion.xml
ammaraziz: ~83MB.

tools/artic/artic_minion.xml
ammaraziz: Thank you for pointing this out, Im still getting my head around data tables. I've made what I think the changes should be. Hopefully it's correct.

tools/artic/artic_guppyplex.xml
bernt-matthias: > I've changed the formatting to be similar to what it was before.

You introduced linebreaks after each attribute which makes the diff hard to read.

> It's optional with a default value, which is near identical to mandatory.

But not identical :) 

> There's no need to change the command block because if not set by the user, the default is chosen. Unless I am wrong in how I think this works, please correct me.

Wrong, the user can empty the input field and you would need to check for this. 

In the end there are two ways to represent default values: 

- make the parameter optional and do not set it on the command line if empty. then the parameter help should document the default value that will be used in the underlying tool.
- make it non optional, but set the default value that will be used in the underlying tool

So far we have used the second option that is also my preference. In the end it does not make a difference .. so I would prefer to keep it as is (alternatively I would check if this is done consistenly for all parameters).

The same should be true for workflows. Note that in 24.2 workflow inputs will allow to cover this IMO:

Screenshot from 2025-01-26 11-10-57

tools/artic/artic_guppyplex.xml
bernt-matthias: Right, it's not really surprising given the huge update.

tools/artic/artic_minion.xml
bernt-matthias: Potentially makes workflow updates slightly more annoying.

tools/artic/macros.xml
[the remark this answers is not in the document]
bernt-matthias: Btw they just released  1.6.1 https://github.com/artic-network/fieldbioinformatics/releases/tag/v1.6.1

tools/artic/tool_data_table_conf.xml.test
```diff
@@ -9,4 +9,8 @@
         <columns>value, dbkey, name, path</columns>
         <file path="${__HERE__}/test-data/all_fasta.loc" />
     </table>
+    <table name="clair_models" comment_char="#">
```
bernt-matthias, line 12: You will also need a `tool_data_table_conf.xml.sample` file. 

Since clair already defines a data table (unfortunatelly with the generic name `model`) I would suggest to reuse this: 
https://github.com/galaxyproject/tools-iuc/blob/37e664b4211a39ac3dba2adb871c0b0ec80a4215/tools/clair3/tool_data_table_conf.xml.sample#L4. That is just copy paste from there.

tools/artic/artic_guppyplex.xml
ammaraziz: > You introduced linebreaks after each attribute which makes the diff hard to read.
Sorry about that, I will see if I can revert this change by diffing it locally. 

> Wrong, the user can empty the input field and you would need to check for this.
That is a very important distinction! Thank you for informing me. 

I will revert this change to make these params mandatory. 

>alternatively I would check if this is done consistenly for all parameters
Yes this is really important, I will test if removing any optional parameters has this effect.

ammaraziz: >The same should be true for workflows. Note that in 24.2 workflow inputs will allow to cover this IMO:

Awesome! I'll have to keep an eye out for this change.

ammaraziz:
All checks are passing! @bernt-matthias  I have incorporated all your comments. Im unsure about the data table integration, if I have gotten it wrong (sorry!) please let me know.

tools/artic/artic_guppyplex.xml
```diff
@@ -67,10 +68,11 @@ ln -s '$elem' inputs/${i}.${elem.ext} &&
                 help="Multiple datasets selected here will get combined into a single output for a single assumed sample. Select a nested list to have its inner lists interpreted as data from one sample each and to obtain one output per inner list." />
             </when>
         </conditional>
-        <param name="max_length" type="integer" label="Remove reads longer than" value="700" help="remove reads greater than this number of base pairs" />
-        <param name="min_length" type="integer" label="Remove reads shorter than" value="400" help="remove reads less than this number of base pairs" />
-        <param name="min_quality" type="integer" min="0" value="7"
-        label="Eliminate reads with a mean base quality score of less than"
+        <param name="max_length" type="integer" value="700" label="Remove reads longer than" help="remove reads greater than this number of base pairs" />
```
bernt-matthias, line 71: Lab `Maximum length` would be better IMO.

tools/artic/artic_guppyplex.xml
```diff
@@ -67,10 +68,11 @@ ln -s '$elem' inputs/${i}.${elem.ext} &&
...
+        <param name="min_length" type="integer" value="400" label="Remove reads shorter than" help="remove reads less than this number of base pairs" />
```
bernt-matthias, line 72: Same

tools/artic/artic_guppyplex.xml
```diff
@@ -67,10 +68,11 @@ ln -s '$elem' inputs/${i}.${elem.ext} &&
...
+        <param name="min_quality"
```
bernt-matthias, line 73: Analogous

tools/artic/tool_data_table_conf.xml.test
```diff
@@ -9,4 +9,8 @@
         <columns>value, dbkey, name, path</columns>
         <file path="${__HERE__}/test-data/all_fasta.loc" />
     </table>
+    <table name="model" comment_char="#">
+        <columns>value, name, platform, path</columns>
```
bernt-matthias, line 13: Would a version column be useful here? Or often has in other cases.

tools/artic/artic_guppyplex.xml
pvanheus: Addressed along with next two comments

tools/artic/tool_data_table_conf.xml.test
pvanheus: Updated to align with #6659. Each model is, according to the folks at Oxford Nanopore, released with a unique name and not versioned (i.e. new versions have new names). I presume that this is true for the models distributed with the clair3 tool too.

tools/artic/artic_guppyplex.xml
```diff
@@ -67,11 +68,12 @@ ln -s '$elem' inputs/${i}.${elem.ext} &&
                 help="Multiple datasets selected here will get combined into a single output for a single assumed sample. Select a nested list to have its inner lists interpreted as data from one sample each and to obtain one output per inner list." />
             </when>
         </conditional>
-        <param name="max_length" type="integer" label="Remove reads longer than" value="700" help="remove reads greater than this number of base pairs" />
-        <param name="min_length" type="integer" label="Remove reads shorter than" value="400" help="remove reads less than this number of base pairs" />
-        <param name="min_quality" type="integer" min="0" value="7"
-        label="Eliminate reads with a mean base quality score of less than"
-        help="Set to 0 to skip the quality check." />
+        <param name="max_length" type="integer" value="700" label="Maximum read length" help="Remove reads greater than this number of base pairs" />
```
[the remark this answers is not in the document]
pvanheus: Done. Set `min="1"` because it needs to be a positive integer.

bgruening (approved):

PR #6715, opened 2025-02-03
update ggplot scatterplot tool

### LINKED ISSUE #6692: Scatterplot: select the columns based on colnames rather than position in the table
Hi! Would it be possible to change in [scatterplot](https://github.com/galaxyproject/tools-iuc/tree/main/tools/ggplot2) the numerical columns choice based on the colnames rather than the position in the table? Especially in the larger tables it gets impractical to find the correct columns position :)

Image

Thanks for considering!

hechth:
FOR CONTRIBUTOR:

Closes #6692

### DIFF
--- a/tools/ggplot2/ggplot_point.xml
+++ b/tools/ggplot2/ggplot_point.xml
@@ -228,6 +228,9 @@ plot_out <- ggplot(input, aes(xcol, ycol)) + gg_point + gg_line + gg_facet +
         <!-- Test text options-->
         <test expect_num_outputs="2">
             <param name="input1" value="mtcars.txt" ftype="tabular" />
+            <conditional name="cols">
+                <param name="header" value="no"/>
+            </conditional>
             <param name="xplot" value="8"/>
             <param name="yplot" value="9"/>   
             <section name="adv">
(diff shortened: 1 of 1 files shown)


hechth:
@bgruening ready for review

SaimMomin12 (approved):

bgruening:
@hechth can you please test in planemo what happens when you provide a file without a header? How does the UI react?

hechth:
Well it just will always think there is a header ... so it treats the first line as a header.

hechth:
@bgruening I added a select to choose whether there is a header or not

bernt-matthias (commented):
In hindsight I wish I would not have implemented https://github.com/galaxyproject/galaxy/commit/2395e6579b28a8f7130b1980e3000768502ff522.

An.extra tool that just adds metadata to a tabular file would have been better. Maybe such a tool would even be possible without duplicating the dataset. 

Then you could just rely on the metadata here (and no conditional would be needed) and the user could use the additional tool to set the metadata of needed.

bgruening:
@bernt-matthias I'm not sure I have followed all the tabular discussions. Do you think we can merge this PR or not?

bernt-matthias:
I would prefer to improve the `data_column` parameter or any of these https://github.com/galaxyproject/galaxy/discussions/19586

hechth:
Please merge anyone.

bernt-matthias:
Not happy with the state of the art, but seems that it is a step forward anyway. Could you please add  a test for the case with columns?

hechth:
@bernt-matthias seems like the CI has some hickups?

PR #6716, opened 2025-02-03
Update Binette

SantaMcCloud:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/binette/binette.xml
+++ b/tools/binette/binette.xml
@@ -12,8 +12,8 @@
         <![CDATA[
             mkdir -p 'input' 'output' &&
 
-            #for $i, $file in enumerate($contig2bin):
-                ln -s '$file.contig2bin_tables' 'input/bin_table_${i}.tsv' &&
+            #for $i, $file in enumerate($contig2bin_tables):
+                ln -s '$file' 'input/bin_table_${i}.tsv' &&
             #end for
 
             ln -s '$contigs' 'input_contigs.fasta' &&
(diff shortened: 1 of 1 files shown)


SantaMcCloud:
Binette doesnt work in a workflow with a single data input becuase of this i change it to a repeat. Binette is similiar with the functionality like [DAS Tool](https://github.com/galaxyproject/tools-iuc/blob/main/tools/das_tool/das_tool.xml) so make it it similiar aswell as wrapper is better and make it suitable to build in into a worklow

bgruening:
@SantaMcCloud can you explain why it does not work in a workflow? Is there something we need to fix in Galaxy?

SantaMcCloud:
no there is nothing wrong in Galaxy the porblem is that using multiple list doesnt work with this tool. Sinlge files would be fine but  the result of most tools are list with mulitple entries which make it not suitable since only 1 list can be used as input when muliple should be used

tools/binette/binette.xml
```diff
@@ -44,8 +44,10 @@
         ]]>
     </command>
     <inputs>
-        <param argument="--contig2bin_tables" type="data" multiple="true" min="2" format="tabular" label="Input contig table"
```
bernt-matthias, line 47: Please revert this. The solution for the user in workflows are nested lists. Then Galaxy will loop over the outer lists an run binette for each inner list.

tools/binette/binette.xml
```diff
@@ -63,7 +65,7 @@
                 help="When a CheckM2 diamond database should be used download and input it here."/>
             </when>
             <when value="cached">
-                <param name="datamanager" type="select" label="Select reference genome" help="Checkm2 Diamond database">
```
bernt-matthias, line 66: This is the only change I would like to keep in this PR.

tools/binette/binette.xml
SantaMcCloud: Ah okay last time i used this this didnt help so i wast thinking of this to use it here. Thanks i tried it out and it work. I revert all except of the change in the other review

tools/binette/binette.xml
SantaMcCloud: Did not change it back!

bgruening (approved):

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/13215976881

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/13215976881

bernt-matthias:
@SantaMcCloud can you fix `icu` to `iuc` over here: 

https://github.com/SantaMcCloud/tools-iuc/blob/d6888da7aba38b97f6cb827355f2de436565684a/tools/binette/.shed.yml#L2

PR #6771, opened 2025-02-17
including ggplot boxplot and barplot; update to 3.5.1

hechth:
FOR CONTRIBUTOR:


hechth:
Locally all tests pass, no idea what is wrong with the CI now.

tools/ggplot2/ggplot_point.xml
```diff
@@ -135,8 +135,22 @@ plot_out <- ggplot(input, aes(xcol, ycol)) + gg_point + gg_line + gg_facet +
     </configfiles>
     <inputs>
         <param name="input1" type="data" format="tabular" label="Input in tabular format" />
-        <param name="xplot" type="integer" value="8" label="Column to plot on x-axis" />
-        <param name="yplot" type="integer" value="9" label="Column to plot on y-axis" />
+
```
bernt-matthias, line 138: This is now inconsistent between the different tools and also not yet consistent with the R code where a header is assumed anyway: `input <- read.csv('$input1', sep='\t', header=TRUE)`. Wrt this I also do not understand why columns are renamed (`names(input)[$xplot] <- "xcol"`)? 

Maybe in this sense a simple solution to the column name problem would be to just assume that the file has headers? I would count this as a bug fix.

tools/ggplot2/help.xml
```diff
@@ -0,0 +1,97 @@
+<macros>
+ 
+<token name="@BOXPLOT_HELP@">
+recetox-boxplot help
```
bernt-matthias, line 4: Do we need a macro that is only used once?

bernt-matthias: recetox-boxplot help -> boxplot help

tools/ggplot2/macros.xml
```diff
@@ -343,4 +342,57 @@
...
+    <xml name="data_input">
+        <param name="input_data" type="data" format="csv,tsv,txt,tabular,parquet" label="Input table" help= "Input file in a tabular/tsv/csv/parquet format"/>
+    </xml>
+
+    <xml name="axes_labels">
+        <param name="xlab" type="text" label="Label for the x axis" optional="true"/>
+        <param name="ylab" type="text" label="Label for the y axis"  optional="true"/>
+    </xml>
+
+    <xml name="boxplot_param">
```
bernt-matthias, line 354: ALso this could just be used without adding a macro?

tools/ggplot2/ggplot_point.xml
hechth: But then again, if the table does not have a header, we are losing one row for plotting or force someone to add a header. I'd rather be able to handle both, I assume also that R is smart enough to infer a header, so we could remove the ```header=TRUE``` section, I assume that the tool would work in the same way.

I can look into the renaming of columns so that this isn't done. Also, this is a legacy tool and not mine, so no idea why things were done like that.

hechth:
I'm using the fancy assertions now which the linter doesn't like. Sorry but this is getting a bit absurd.

bernt-matthias:
> I'm using the fancy assertions now which the linter doesn't like. Sorry but this is getting a bit absurd.

Guess this is https://github.com/galaxyproject/galaxy/issues/19705 .. probably you do not have the latest `galaxy.tool-util`  package locally .. but you can't do anything anyway :) the linter is wrong here

mvdbeek (approved):
Thanks a lot!

PR #6784, opened 2025-02-20
Adding new tool: MerquryFK CNPlot

SaimMomin12:
FOR CONTRIBUTOR:

---
This PR adds CNPlot, a new tool to the MerquryFK toolsuite.


tools/merquryfk/cnplot.xml
```diff
@@ -0,0 +1,78 @@
...
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements"/>
+    <command detect_errors="exit_code"><![CDATA[
+        ln -s $reads input.ktab &&
+        ln -s $assembly input.fastq &&
+        ln -s $fastk_ktab_files ktab_files.tar.gz &&
+        tar -xf ktab_files.tar.gz -C '.' &&
+        mv "\$(find ktabfiles -type f -name '*.ktab.1')" ".input.ktab.1" &&
```
bgruening, line 12: Are those files in nested folders? A simple glob `**/*` does not work?

SaimMomin12: I didn't try glob here. I assumed finding the hidden ktab file and renaming a straightforward way? Or can we do it in a more cleaner way using glob?

bgruening: I just want to avoid `find` if possible. It has different behaviour under OSX and Linux, ... so I was wondering if this is really needed

SaimMomin12: I replaced find with a simpler implementation

Issue #6211, opened 2024-08-06
clair3 breaks galaxy due to discrepancy in all_fasta columns


cat-bro:
Galaxy Australia's staging server failed to restart this morning. This entry in the shed_tool_data_table_conf.xml

```
    <table name="all_fasta" comment_char="#">
        <columns>value, name, path</columns>
        <file path="/mnt/galaxy/tool-data/shed_tool_data/toolshed.g2.bx.psu.edu/repos/iuc/clair3/c8ba8286e7c0/all_fasta.loc"/>
        <tool_shed_repository>
            <tool_shed>toolshed.g2.bx.psu.edu</tool_shed>
            <repository_name>clair3</repository_name>
            <repository_owner>iuc</repository_owner>
            <installed_changeset_revision>c8ba8286e7c0</installed_changeset_revision>
        </tool_shed_repository>
    </table>
```

has the wrong columns in `columns`. Each other entry for all_fasta has `<columns>value, dbkey, name, path</columns>` and galaxy would not start until this new entry was commented out.

bernt-matthias:
I would consider this a bug in galaxy.

bernt-matthias:
> We can't really do anything there in Galaxy, failing to boot is probably the correct thing to do.

We could just fail to load the tool, like we do for xml parsing errors, etc.

mvdbeek:
This is a server misconfiguration that might have a huge impact, this needs to be fixed, failing is the right thing to do.

wm75:
is all_fasta even the right table to use? wouldn't fasta_indexes be the correct one?

bernt-matthias:
Still don't get why this break on au and not eu?

mvdbeek:
`all_fasta` is for fasta files, `fasta_indexes` would be for `.fai` files as produced by `samtools faidx`

wm75:
> `all_fasta` is for fasta files, `fasta_indexes` would be for `.fai` files as produced by `samtools faidx`

https://github.com/galaxyproject/tools-iuc/blob/6da96d865a3a557cfa3ad09e1cfa830519e73748/tools/clair3/clair3.xml#L20

yes, that's why

mvdbeek:
I guess the idea was that all_fasta is more widely populated ? I have no particular problem with this, of course it would be a little faster to go with fasta indexes.

wm75:
> Still don't get why this break on au and not eu?

[...]

PR #6790, opened 2025-02-22
Minor fixes to tool_data_table_conf and related files for clair3 and artic

pvanheus:
FOR CONTRIBUTOR:


tools/clair3/tool_data_table_conf.xml.sample
```diff
@@ -2,8 +2,8 @@
 <tables>
     <!-- Locations of clair3 model folders -->
     <table name="model" comment_char="#">
-        <columns>value, name, platform, path, source</columns>
-        <file path="tool-data/model.loc" />
+        <columns>value, platform, sha256, path, source</columns>
```
wm75, line 5: That's very unfortunate, but you cannot change the columns of an existing data table.
You need also a new table name now.

pvanheus: Ok. I'll close the PR then. The tool is not, in fact, broken, despite this mistake.

pvanheus: Just to clarify - the table is not like the one described here - the canonical version of the table is the one described here: https://github.com/galaxyproject/tools-iuc/pull/6659 (there was something to create the table before this DM was created, even though it was referenced in the clair3 tool)

wm75: Ah ok, I see (was wondering how I had missed this when reviewing the PR for the DM). Then this might work.
Not exactly sure how all the different versions of this file are treated by Galaxy though.

bernt-matthias: > Not exactly sure how all the different versions of this file are treated by Galaxy though.

This will create problems if someone already installed the version without the fix.

pvanheus: @bernt-matthias will the problem be there if people installed the tool with the older version of the data table or only for those installing the DM (if, for example, a data table changed after the DM was installled)?

bernt-matthias: The problem will occur for anyone that have 1.6.0+galaxy0 as well as the DM/newer tool versions installed. 

> That's very unfortunate, but you cannot change the columns of an existing data table.

But in the end it's a bug fix. Still there might be a problem if instances already deployed the tool. The faster we act the smaller the problem will be. 

If eu does not have the [tool installed](https://github.com/usegalaxy-eu/usegalaxy-eu-tools/blob/1cd22e99d7fa68b9d98856fe7ede4aa1127dab07/tools_iuc.yaml.lock#L7873) I would be in favor of merging.

bernt-matthias: EU already has it ..

wm75: Yeah, but I guess/hope we can manage to clean things up.

wm75: but @bernt-matthias you'd merge without a version bump then, correct?

bernt-matthias: Yes.

wm75: ok, then let's do that. @pvanheus can you undo the wrapper version bump and I'll try to sort out the EU data tables?

wm75: It also seems that you haven't captured all the changes from #6796 (which unfortunately has been merged already) in this PR here (which means we cannot overwrite the +galaxy0 version of artic_minion here anymore). Can you please just rebase?

tools/clair3/tool_data_table_conf.xml.sample
```diff
@@ -2,8 +2,8 @@
 <tables>
     <!-- Locations of clair3 model folders -->
     <table name="model" comment_char="#">
```
wm75, line 4: the name of the table is `clair3_models`. Can you update that please?

wm75: same for the `artic_minion` tool!

wm75:
@pvanheus another question while we're at it: have the model files ever changed their format?
Newly installed clair3_models are now also user-selectable for old versions of clair3 (like 0.1.11+galaxy0 installed on EU). Would they be compatible?

pvanheus:
> @pvanheus another question while we're at it: have the model files ever changed their format? Newly installed clair3_models are now also user-selectable for old versions of clair3 (like 0.1.11+galaxy0 installed on EU). Would they be compatible?

The model files that are built by the clair3 group ([available here](http://www.bio8.cs.hku.hk/clair3/clair3_models/)) have not changed since October 2021, which is before the release date of any of the versions of clair3 used in Galaxy, so it seems that the model format is very stable and definitely stable across all Galaxy tool releases.

tools/clair3/tool_data_table_conf.xml.sample
pvanheus: @wm75 some guidance on how to rebase correctly would be very appreciated!

wm75:
Hmm, that was a semi-successful integration. I think things are actually correct now, just a lot of  annoying in-between commits from main.

tools/artic/tool-data/models.loc.sample
```diff
@@ -1,13 +1,4 @@
...
-#This file lists the locations and dbkeys of all the model files
-#under the "models" directory (a directory that contains a directory
-#for each build). This file has the format (white space characters are
-#TAB characters):
-#
-#<id>    <name>    <platform>    <path>
-#
-#So, all_model.loc could look something like this:
-test_model	"the_model_name"	r941_prom_hac_g360+g422	$(dirname $(which run_clair3.sh))/models/r941_prom_hac_g360+g422#This file lists the names and locations of all the clair3 model files
+#This file lists the names and locations of all the clair3 model files
```
wm75, line 1: this entire file still needs to be renamed to `clair3_models.loc.sample` I guess.
Then the tool_data_table.conf needs to be adjusted too.

tools/clair3/clair3.xml
```diff
@@ -2,7 +2,7 @@
     <description>germline small variant caller for long-reads</description>
     <macros>
         <token name="@TOOL_VERSION@">1.0.10</token>
-        <token name="@VERSION_SUFFIX@">1</token>
+        <token name="@VERSION_SUFFIX@">2</token>
```
wm75, line 5: don't bump this version so that we can overwrite the faulty previous attempt

martenson:
xref https://github.com/galaxyproject/tools-iuc/issues/6211

mvdbeek:
You cannot ever change data table column definitions, no matter how minor the change is.

martenson:
Can we detect this and refuse repo update?

PR #6815, opened 2025-03-05
Lowered required galaxy tool version

RealYHD:
Simply lowers the minimum Galaxy version that should be required to run this tool.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/autobigs/autobigs-cli.xml
+++ b/tools/autobigs/autobigs-cli.xml
@@ -1,9 +1,9 @@
-<tool id="autobigs-cli" name="autoBIGS.cli" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="21.05">
+<tool id="autobigs-cli" name="autoBIGS.cli" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="19.05">
     <description>Automated MLST typing</description>
     <macros>
         <import>macros.xml</import>
         <token name="@TOOL_VERSION@">0.6.2</token>
-        <token name="@VERSION_SUFFIX@">0</token>
+        <token name="@VERSION_SUFFIX@">1</token>
         <expand macro="bio_tools"/>
     </macros>
 


SaimMomin12 (approved):

SaimMomin12:
Failed linting
.. WARNING (TestsCaseValidation): Test 1: failed to validate test parameters against inputs - tests won't run on a modern Galaxy tool profile version. Validation errors are [1 validation error for PydanticModelFor[autobigs-cli]
mode.__absent__.fasta
  Field required [type=missing, input_value={}, input_type=dict]]
.. WARNING (TestsCaseValidation): Test 2: failed to validate test parameters against inputs - tests won't run on a modern Galaxy tool profile version. Validation errors are [1 validation error for PydanticModelFor[autobigs-cli]
mode.__absent__.fasta
  Field required [type=missing, input_value={}, input_type=dict]]
.. WARNING (TestsCaseValidation): Test 3: failed to validate test parameters against inputs - tests won't run on a modern Galaxy tool profile version. Validation errors are [1 validation error for PydanticModelFor[autobigs-cli]
mode.__absent__.fasta
  Field required [type=missing, input_value={}, input_type=dict]]
.. WARNING (TestsCaseValidation): Test 4: failed to validate test parameters against inputs - tests won't run on a modern Galaxy tool profile version. Validation errors are [1 validation error for PydanticModelFor[autobigs-cli]
mode.__absent__.fasta
  Field required [type=missing, input_value={}, input_type=dict]]
.. WARNING (TestsCaseValidation): Test 5: failed to validate test parameters against inputs - tests won't run on a modern Galaxy tool profile version. Validation errors are [1 validation error for PydanticModelFor[autobigs-cli]
mode.__absent__.fasta

SaimMomin12:
@RealYHD Please add `.lint_skip` file with TestsCaseValidation as its contents (For Eg. https://github.com/galaxyproject/tools-iuc/blob/4fde0bb4e8badcb215d1e8ee0bc787c3b98b6958/tools/mcl/.lint_skip)

bernt-matthias (changes requested):
The `from_url` feature has been introduced only in 24.0 https://github.com/galaxyproject/galaxy/pull/17087

So I think this should be the minimal version, or is the tool still usable in older galaxy versions?

nilchia:
@SaimMomin12 in the test specify the conditional
```
            <param name="operation" value="st" />
            <param name="fasta" value="tohama_I_bpertussis_minimized_features.fasta" />
            <param name="scheme" value="MLST"
```
to
```
            <conditional name="mode">
                <param name="operation" value="st" />
                <param name="fasta" value="tohama_I_bpertussis_minimized_features.fasta" />
                <param name="scheme" value="MLST" />
            </conditional>
```

RealYHD:
> The `from_url` feature has been introduced only in 24.0 [galaxyproject/galaxy#17087](https://github.com/galaxyproject/galaxy/pull/17087)
> 
> So I think this should be the minimal version, or is the tool still usable in older galaxy versions?

Hi @bernt-matthias, thanks for that catch. Unfortunately, no, it would no longer be usable. I will more than likely need to change a few of the inputs. I did not see this listed on the [tool XML changelogs](https://docs.galaxyproject.org/en/latest/dev/schema.html#id11), hence my lapse in judgement. I will be more careful going forward.

Thank you everyone for your very helpful input. I've converted this PR into a draft as this turned out to be a much more complicated change than I initially thought. I will work on these items and mark it as ready for review after more thorough testing.

bernt-matthias:
> I will need to change a few of the inputs.

How? What would be your idea?

> As a side note, this may be a good addition to the list of [behaviour changes](https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-profile)?

If we would include every new feature in this list, it would cover most of the Galaxy's [changelog](https://docs.galaxyproject.org/en/master/releases/index.html). This list is about changed behavior only (so called profile versions where we change Galaxy's behavior depending on tool version).

RealYHD:
> > I will need to change a few of the inputs.
> 
> How? What would be your idea?

The CLI this tool is wrapping (autobigs-cli) can output the list of databases to a CSV from the same web API. This was mostly covered in 938c471f59dfe6a7b4fce75bca007a55983fed9d.

> 
> > As a side note, this may be a good addition to the list of [behaviour changes](https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-profile)?
> 
> If we would include every new feature in this list, it would cover most of the Galaxy's [changelog](https://docs.galaxyproject.org/en/master/releases/index.html). This list is about changed behavior only (so called profile versions where we change Galaxy's behavior depending on tool version).

No problem. I misunderstood that list then.

RealYHD:
> @RealYHD in the test specify the conditional
> 
> ```
>             <param name="operation" value="st" />
>             <param name="fasta" value="tohama_I_bpertussis_minimized_features.fasta" />
>             <param name="scheme" value="MLST"
> ```
> 
> to
> 
> ```
>             <conditional name="mode">
>                 <param name="operation" value="st" />
>                 <param name="fasta" value="tohama_I_bpertussis_minimized_features.fasta" />
>                 <param name="scheme" value="MLST" />
>             </conditional>
> ```

Thanks for letting me know. I've updated all test cases to reflect this syntax in e7cb818d7570674a63619981cd787d6a55947db3.

RealYHD:
I believe this PR is now ready for review again. I installed it on a 20.01 test Galaxy server and tried it out and it seems to work. All current tests pass on the local 20.01 instance as well.

RealYHD:
I've just specified the latest (tested in practice) version as client in 3ee9e12.

RealYHD:
@bernt-matthias I believe I solved the issues you've noted described in [this comment](https://github.com/galaxyproject/tools-iuc/pull/6815#issuecomment-2707065973). Please let me know if there's anything else you'd like for me to change or explain.

tools/autobigs/autobigs-cli.xml
```diff
@@ -13,128 +13,153 @@
...
-                                    }
-                                }
-                            }
-                            return options
-                        }]]></postprocess_expression>
-                    </options>
-                </param>
-            </when>
-            <when value="institutpasteur">
-                <param name="bigsdb" label="PubMLST Database" type="select" help="Select from a list of sequence definition databases hosted by Institut Pasteur." optional="false">
```
bernt-matthias, line 45: I'm really a bit sad that we will loose this cool feature (at least I think it's cool -- correct me if I'm wrong).

bernt-matthias (commented):
Dear @RealYHD . Really appreciate your work, but I need to understand the motivation to lower the profile (to a really old one) and to remove all the `institutpasteur` parts of the tool.

Could you give me some insights?

tools/autobigs/.lint_skip
```diff
@@ -0,0 +1 @@
+TestsCaseValidation
```
bernt-matthias, line 1: We must not add this here, but fix the tests. Let me know if you need help.

RealYHD:
> Dear @RealYHD . Really appreciate your work, but I need to understand the motivation to lower the profile (to a really old one) and to remove all the `institutpasteur` parts of the tool.
> 
> Could you give me some insights?

I'm working on this project on behalf of a lab and unfortunately, the network Galaxy instance we're running is on 20.01. I'm trying to satisfy that requirement.

This being said, I'd be happy to push another version for the latest version of Galaxy that supports the latest version of autoBIGS after setting this one up.

I appreciate you getting back to me!

tools/autobigs/autobigs-cli.xml
RealYHD: I like that feature too. I'm willing to also make a PR for the latest version of Galaxy running the latest version of AutoBIGS that does support that feature after completing this one if that works for the IUC team?

bernt-matthias: This would be 24.0 https://github.com/galaxyproject/galaxy/pull/17087

RealYHD: I just noticed, you mentioned removing the `institutpasteur` parts of the tool, however, all Institut Pasteur databases still work. I think it is just that it is no longer explicitly mentioned in the configuration.

Please see https://github.com/galaxyproject/tools-iuc/pull/6815#issuecomment-2783552410 for my answer for why I'm trying to release a lower version number. Thanks!

PR #6823, opened 2025-03-09
Add BWA-MEM2 indexing tool

mvdbeek:
Builds on https://github.com/galaxyproject/tools-iuc/pull/6558

FOR CONTRIBUTOR:


bgruening:
Can we somehow reuse the data managers? It seems redundant to create a new index tool for all our reference consuming tools.

mvdbeek:
I don't see how. You'd have to look at every data manager and then allowlist user execution. Picking up a directory from extra_files_path is way simpler.

bgruening:
Somehow, I thought the plan was to enforce `data-bundles` for non-admin users and consume those data-bundles by tools. Did I mixed that up?

mvdbeek:
That was the plan, but I don't think it's working out. It starts with data managers being pretty complicated to write, then you got a single extension `.data_manager_json` for all of the different table types, data managers can create more than one entry, the dbkey displayname etc values don't really make sense in the context of history items.

I think it would make more sense to go the other way around and have a data manager that knows how to consume datatypes like `bwa_mem2_index` and turns it into a loc file entry by adding the required metadata.

mvdbeek:
How can we advance here ? Do you want me to create indexers for the data managers we have ?

bgruening:
I don't know, sorry. I also would like to make DM easier or redundant. If we have a plan, I'm all in. But at the moment, I see only more work and no clear way to consolidate all that.

> I think it would make more sense to go the other way around and have a data manager that knows how to consume datatypes like bwa_mem2_index and turns it into a loc file entry by adding the required metadata.

Is that possible? And what are we doing with the DM that just download stuff? Do they stay?

mvdbeek:
There doesn't need to be consolidation right away ? In this tool it's just another way to provide a reference, and I would assume some / most of these references you wouldn't ever want to put into reference data, like when doing self alignments in PAF mode -- that's for one of the VGP workflows. 

In any case a prerequisite would be datatypes corresponding to the type of artefact that the data manager would produce.

> Is that possible?

Sure, I guess the bigger question is does it make sense ? From the admin perspective you turned a one-step process into 2 steps. That said, you could for instance add another option to https://github.com/galaxyproject/tools-iuc/blob/main/data_managers/data_manager_manual/data_manager/data_manager_manual.xml that would pick up the directory from extra_files_path.

pavanvidem:
@mvdbeek the indexing tool does not require samtools but the mapping tool still [requires samtools](https://github.com/galaxyproject/tools-iuc/blob/main/tools/bwa_mem2/bwa-mem2.xml#L117-L121). Was that deliberately removed in bwa_mem2?

PR #6450, opened 2024-10-11
Updating tools/augustus to 3.5.0

SaimMomin12:
This PR updates the Augustus tool with addition of new species and updates to the test data

Fixes: https://github.com/galaxyproject/tools-iuc/pull/6433

---
FOR CONTRIBUTOR:
* [x] I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [x] License permits unrestricted use (educational + commercial)
* [ ] This PR adds a new tool or tool collection
* [x] This PR updates an existing tool or tool collection
* [ ] This PR does something else (explain below)

PR #6842, opened 2025-03-14
Bump augustus

bernt-matthias:
While updating BUSCO a comment asked me to update species list at Augustus as well... just updated the whole thing.

FOR CONTRIBUTOR:


tools/augustus/macros.xml
```diff
@@ -8,7 +8,7 @@
     </xml>
 
     <token name="@TOOL_VERSION@">3.4.0</token>
```
bgruening, line 10:
```suggestion
    <token name="@TOOL_VERSION@">3.5.0</token>
```

?

bernt-matthias:
xref https://github.com/galaxyproject/tools-iuc/pull/6450

The training tool will fail due to: 

```
LibMambaUnsatisfiableError: Encountered problems while solving:
  - package augustus-3.5.0-pl5321h700735d_3 requires perl >=5.32.1,<5.33.0a0 *_perl5, but none of the providers can be installed

Could not solve for environment specs
The following packages are incompatible
├─ augustus 3.5.0**  is installable and it requires
│  └─ perl >=5.32.1,<5.33.0a0 *_perl5, which can be installed;
└─ maker 3.01.03**  is not installable because it requires
   └─ perl >=5.26.2,<5.26.3.0a0 , which conflicts with any installable versions previously reported.
```

any idea?

PR #6844, opened 2025-03-15
Add a new option to SemiBin

SantaMcCloud:
FOR CONTRIBUTOR:

Add the option to use the self-supervised learning function from SemiBin.

### DIFF
--- a/tools/semibin/macros.xml
+++ b/tools/semibin/macros.xml
@@ -142,7 +142,8 @@ ln -s '$e' '${identifier}.bam' &&
     ]]></token>
     <xml name="ref_select">
         <param name="select" type="select" label="Reference database">
-            <option value="cached" selected="true">Cached database</option>
+            <option value="ml" selected="true">Use SemiBin ML function</option>
+            <option value="cached">Cached database</option>
             <option value="taxonomy">Pre-computed taxonomy</option>
         </param>
     </xml>
@@ -162,6 +163,7 @@ ln -s '$e' '${identifier}.bam' &&
             <when value="taxonomy">
                 <param argument="--taxonomy-annotation-table" type="data" format="tabular" label="Pre-computed mmseqs2 format taxonomy TSV file"/>
             </when>
+            <when value="ml"/>
         </conditional>
     </xml>
     <xml name="ref-multi">
--- a/tools/semibin/semibin.xml
+++ b/tools/semibin/semibin.xml
@@ -20,7 +20,7 @@ SemiBin2
     #end if
     #if $mode.ref.select == "cached"
     --reference-db-data-dir '$mode.ref.cached_db.fields.path'
-    #else
+    #else if $mode.ref.select == "taxonomy":
     --taxonomy-annotation-table '$mode.ref.taxonomy_annotation_table'
     #end if
 #else
@@ -28,7 +28,7 @@ SemiBin2
     --separator '$separator'
     #if $mode.ref.select == "cached"
     --reference-db-data-dir '$mode.ref.cached_db.fields.path'
-    #else
+    #else if $mode.ref.select == "taxonomy":
     --taxonomy-annotation-table 
         #for $e in $mode.ref.taxonomy_annotation_table
             '$e' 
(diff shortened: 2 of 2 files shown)


bernt-matthias:
Would you have the time to fix the linter errors https://toolshed.g2.bx.psu.edu/repos/artbio/manta (just asking 🤔)

bernt-matthias:
Wrong link https://github.com/galaxyproject/tools-iuc/blob/72f8438806394993d86286f4b3967d533f6ddb0d/tools/semibin/.lint_skip

SantaMcCloud:
> Would you have the time to fix the linter errors https://toolshed.g2.bx.psu.edu/repos/artbio/manta (just asking 🤔)

I can have a look into it :)

SantaMcCloud:
The liting should be fix now but there is a problem with a test which i want to have a look at it when this is done so pls no merging when evreything is good.

If there is still someting which has to be done please let me know and i will have a look into it!

bernt-matthias:
OK. Please also remove the `.lint_skip` file.

SantaMcCloud:
Okay i found some problems why the linting is failing. I will change them such that the linintig is not failing anymore.

bernt-matthias:
> value should equal an empty string which is in this case

You can use `json_value="null"`. It has to be this way, because one could also have an option with value `""`.. which would be indistinguishable from nothing selected...

Would be good to have this documented.

SantaMcCloud:
> > value should equal an empty string which is in this case
> 
> You can use `json_value="null"`. It has to be this way, because one could also have an option with value `""`.. which would be indistinguishable from nothing selected...
> 
> Would be good to have this documented.

You mean `value` in this case instead of `json_value` or am i wrong?

bernt-matthias:
I mean that you need to specify `json_value="null"` if you want to have a test where an optional select has no values selected. I think it would be good to have this covered. In particular because I think this is a bug:

on line 56 

```diff
- #if ($mode.select == 'single' or $mode.select == 'co') and "pre_reclustering_bins" in $extra_output
+ #if ($mode.select == 'single' or $mode.select == 'co') and $extra_output and "pre_reclustering_bins" in $extra_output
```

SantaMcCloud:
Okay this with the `json_value` is new to me this should be mentinon somewherer in the documentation i guess.

This on line 56 seems fine but i will change it just to be sure. This was added 3 years ago possible it was fine back then i guess.

bernt-matthias (approved):

bernt-matthias:
> This on line 56 seems fine but i will change it just to be sure. This was added 3 years ago possible it was fine back then i guess.

If you like you could add a test case with `extra_output`. This should make the old version fail.

SantaMcCloud:
> > This on line 56 seems fine but i will change it just to be sure. This was added 3 years ago possible it was fine back then i guess.
> 
> If you like you could add a test case with `extra_output`. This should make the old version fail.

Will do it the next few days because im courise about it.  But i dont think this is not needed to make the merge here or isnt it? I did change it because i thing this will not work aswell but i wasnt sure since a lot of linting errors did happen because if old stuff so i just thought this might be one thing aswell but now i thing it was just more about a typo and not more.

bernt-matthias:
> But i dont think this is not needed to make the merge here or isnt it? 

So you prefer to merge now?

SantaMcCloud:
> > But i dont think this is not needed to make the merge here or isnt it?
> 
> So you prefer to merge now?

yes there is nothing to chnage and the recommanded option from the devs is usable now

tools/semibin/generate_cannot_links.xml
```diff
@@ -18,11 +18,12 @@ SemiBin2
     --cannot-name 'cannot'
 #if $mode.ref.select == "cached"
     --reference-db-data-dir '$mode.ref.cached_db.fields.path'
-#else
+#end if
+#if $mode.ref.select == "taxonomy"
```
bernt-matthias, line 22: @SantaMcCloud : This will not work for the multi mode, or? We will need to loop over: '$mode.ref.taxonomy_annotation_table'

SantaMcCloud: I just checked the macros and it seems that the person who wrote it wanted to give the user the choice to input muliple `tsv `files. Im not 100% sure if thos option is viable via galaxy since evreything is nested in macros but if you want i can have a look into it and can make a PR to change it? :)

PR #6857, opened 2025-03-17
Adapted MMUPHin for workflow use

SantaMcCloud:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/mmuphin/macros.xml
+++ b/tools/mmuphin/macros.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <macros>
     <token name="@TOOL_VERSION@">1.16.0</token>
-    <token name="@VERSION_SUFFIX@">0</token>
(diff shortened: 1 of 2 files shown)


SantaMcCloud:
I was ask to change this wrapper such that this tool can work in a workflow since the `data_column" type is not good to in a workflow. In this case all param which used this type and multiple column can be choosed was set into a conditional which either use the `data_column` type or a text input is used where a list in comma seperated column number can be inputed.

paulzierep:
Same issue as here: https://github.com/galaxyproject/tools-iuc/pull/6858

SantaMcCloud:
> Same issue as here: #6858

all issues should be done herer and the validator is now there for only digits input in a comma seperated list

paulzierep:
if we could fix https://github.com/galaxyproject/galaxy/issues/19832 this tool fix would not be needed

PR #6858, opened 2025-03-17
Adapted MaAsLin2 for workflow use

SantaMcCloud:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/maaslin2/maaslin2.xml
+++ b/tools/maaslin2/maaslin2.xml
@@ -1,4 +1,4 @@
-<tool id="maaslin2" name="MaAsLin 2" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
+<tool id="maaslin2" name="MaAsLin2" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
     <description>Microbiome Multivariable Association with Linear Models</description>
     <macros>
         <import>macros.xml</import>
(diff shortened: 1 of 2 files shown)


SantaMcCloud:
I was ask to change this wrapper such that this tool can work in a workflow since the data_column type is not good to in a workflow. In this case all param which used this type and multiple column can be choosed was set into a conditional which either use the data_column type or a text input is used where a list in comma seperated column number can be inputed.

bernt-matthias:
@mvdbeek

tools/maaslin2/maaslin2.xml
```diff
@@ -8,93 +8,118 @@
...
-echo 'Assigned fixed effects as:' \$fixed_effects &&
-#end if
+        #if $fixed_effects
+            #set idx = []
+            #if $columns.is_select == "data":    
+                #for $i in $fixed_effects:
+                    #silent idx.append(f'${i}')
+                #end for
+            #else:
+                #for $i in str($fixed_effects).split( "," ):
```
paulzierep, line 22: you are using split here and then join below, can you for this case not use $fixed_effects for idx_for_awk ?

tools/maaslin2/maaslin2.xml
```diff
@@ -8,93 +8,118 @@
...
-echo 'Assigned random effects as:' \$random_effects &&
-#end if
+        #if $random_effects
+            #set idx = []
+            #if $columns.is_select == "data":
+                #for $i in $random_effects:
+                    #silent idx.append(f'${i}')
+                #end for
+            #else:
+                #for $i in str($random_effects).split( "," ):
```
paulzierep, line 41: same as before

tools/maaslin2/maaslin2.xml
```diff
@@ -8,93 +8,118 @@
...
+                <option value="data" selected="true">Via data column</option>
+                <option value="manually">Integer list</option>
+            </param>
+            <when value="data">
+                <param argument="--fixed_effects" type="data_column" data_ref="input_metadata" use_header_names="true" multiple="true" optional="true" label="Interactions: Fixed effects" help="The fixed effects for the model, comma-delimited for multiple effects, Default value: All "/>
+                <param argument="--random_effects" type="data_column" data_ref="input_metadata" use_header_names="true" multiple="true" optional="true" label="Random effects" help="The random effects for the model, comma-delimited for multiple effects, Default: None"/>
+            </when>
+            <when value="manually">
+                <param argument="--fixed_effects" type="text" value="" optional="true" label="Interactions: Fixed effects" help="The fixed effects for the model, comma-delimited for multiple effects. Enter the selected column(s) in a comma separated list which should be selected. NOTE: the first column is 1 not 0! Default value: All "/>
```
paulzierep, line 118: can you add a validator, that only allows comma and numbers

tools/maaslin2/maaslin2.xml
```diff
@@ -340,29 +506,30 @@ cp *.pdf figures
...
+        <test expect_num_outputs="6">
+            <param name="input_data" value="HMP2_taxonomy.tsv"/>
             <param name="input_metadata" value="HMP2_metadata.tsv"/>
-            <param name="fixed_effects" value="7,9"/>
-            <param name="random_effects" value="3" />
-          
+            <conditional name="columns">
+                <param name="is_select" value="data"/>
+                <param name="fixed_effects" value="7,9"/>
+                <param name="random_effects" value="3"/>
```
paulzierep, line 531: can we have one test with random effects ?

paulzierep:
> I was ask to change this wrapper such that this tool can work in a workflow since the data_column type is not good to in a workflow. In this case all param which used this type and multiple column can be choosed was set into a conditional which either use the data_column type or a text input is used where a list in comma seperated column number can be inputed.

We want to use this tool in a workflow, where the user specifies on workflow level what columns based on id should be used for differential analysis. But we can only connect an integer input to this parameter, which does not allow choosing multiple columns. This is a workaround, until the UI allows connecting text with comma or another option.

tools/maaslin2/maaslin2.xml
SantaMcCloud: No becuase the problem is that you need the doller symbol befor any number for awk to function. So to split them up transform them into a list to which then they will added into a different list but with a doller symbol befor the number

tools/maaslin2/maaslin2.xml
SantaMcCloud: same answer as before

tools/maaslin2/maaslin2.xml
SantaMcCloud: last test has both with a manually input (line 585 -587)

tools/maaslin2/maaslin2.xml
SantaMcCloud: https://github.com/galaxyproject/tools-iuc/pull/6858/commits/b345d58c37c04213768589a9b17fad25b100c2f6

tools/maaslin2/maaslin2.xml
SantaMcCloud: i try to make a test but the problem is that i get this error always: 

```
Error in Maaslin2(positional_args[1], positional_args[2], positional_args[3],  : 
  Please provide the reference for the variable 'site' which includes more than 2 levels: Cedars-Sinai, Cincinnati, Emory, MGH.
Execution halted
```

and this was the std from the tool:

```
Assigned random effects as: age
[1] "Creating output folder"
[1] "Creating output feature tables folder"
[1] "Creating output fits folder"
[1] "Creating output figures folder"
2025-03-18 12:10:04.614039 INFO::Writing function arguments to log file
2025-03-18 12:10:04.638288 INFO::Verifying options selected are valid
2025-03-18 12:10:04.639894 INFO::Determining format of input files
2025-03-18 12:10:04.642318 INFO::Input format is data samples as rows and metadata samples as rows
2025-03-18 12:10:04.650447 WARNING::Feature name included as fixed and random effect, check that this is intended: age
2025-03-18 12:10:04.653872 INFO::Formula for random effects: expr ~ (1 | age)
2025-03-18 12:10:04.656785 INFO::Formula for fixed effects: expr ~  site + age + diagnosis + subject + antibiotics + dysbiosis_binary + dysbiosis + dysbiosisnonIBD + dysbiosisUC + dysbiosisCD + collection
```
You know why the error happen?

paulzierep: yes one of the parameters (fixed effects or random effects) corresponds to the column `site`. This variable has multiple levels and then maaslin2 needs to know which level to take as reference, this you need to specify in the reference param e.g. with `site,Cedars-Sinai`

SantaMcCloud: Thanks this did help me! 

Here is the random_effects only test https://github.com/galaxyproject/tools-iuc/pull/6858/commits/89ed6091e7fc77977596b183aeaf2959a1fcec73

paulzierep:
> @mvdbeek

if we could fix this: https://github.com/galaxyproject/galaxy/issues/19832 this tool fix would not be needed

PR #6864, opened 2025-03-19
Add a DB filter to prevent using the newest DB from CheckM2

SantaMcCloud:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/checkm2/checkm2.xml
+++ b/tools/checkm2/checkm2.xml
@@ -3,7 +3,7 @@
     <macros>
         <!-- enable tests and output assertions on update -->
         <token name="@TOOL_VERSION@">1.0.2</token>
-        <token name="@VERSION_SUFFIX@">0</token>
+        <token name="@VERSION_SUFFIX@">1</token>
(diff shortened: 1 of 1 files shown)


SantaMcCloud:
@bernt-matthias This should work to only use the DB version 1.0.2 for this wrapper. 

We have to wait till this version is on shed till we can merge the 1.1.0 version and we can remove the filter in this wrapper version then too since the older DB works there?

tools/checkm2/checkm2.xml
```diff
@@ -36,6 +36,7 @@
         <param name="database" type="select" label="Select reference genome" help="Checkm2 Diamond database">
             <options from_data_table="@IDX_DATA_TABLE@">
                 <filter type="sort_by" column="2"/>
+                <filter type="regexp" column="1" value="1.0.2"/>
```
bernt-matthias, line 39: Is this covered by a test?

SantaMcCloud: no but i tried it manually with planemo serve which work fine. The problem with this wrapper is that using a test DB can not work which stated as a comment in the test section

PR #6896, opened 2025-03-31
Update mmseqs2 to latest version in date (17-b804f)

hugolefeuvre:
FOR CONTRIBUTOR:

### DIFF
--- a/data_managers/data_manager_mmseqs2_database/data_manager/data_manager_mmseqs2_download.xml
+++ b/data_managers/data_manager_mmseqs2_database/data_manager/data_manager_mmseqs2_download.xml
@@ -1,7 +1,7 @@
 <tool id="data_manager_mmseqs2_download" name="Download MMseqs2 databases" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" tool_type="manage_data" profile="22.05">
     <description></description>
     <macros>
-        <token name="@TOOL_VERSION@">15.6f452</token>
+        <token name="@TOOL_VERSION@">17-b804f</token>
         <token name="@VERSION_SUFFIX@">0</token>
     </macros>
     <requirements>
@@ -122,8 +122,8 @@ cp '$dmjson' '$out_file'
             <output name="out_file">
                 <assert_contents>
                     <has_text text='"mmseqs2_databases":'/>
-                    <has_text text='"version": "15.6f452"'/>
-                    <has_text_matching expression='"value": "SILVA-15.6f452-[0-9]{8}"'/>
+                    <has_text text='"version": "17-b804f"'/>
+                    <has_text_matching expression='"value": "SILVA-17-b804f-[0-9]{8}"'/>
                     <has_text_matching expression='"name": "SILVA [0-9]{8}"'/>
                     <has_text text='"type": "nucleotide"'/>
                     <has_text text='"taxonomy": "yes"'/>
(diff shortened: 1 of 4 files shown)


hugolefeuvre:
Update because of this automated update fail : https://github.com/galaxyproject/tools-iuc/pull/6893

tools/mmseqs2/macro.xml
```diff
@@ -38,6 +38,7 @@
                 <option value="0" selected="true">Include region</option>
                 <option value="1">Exclude region</option>
             </param>
+            <param argument="--mask-n-repeat" type="integer" min="0" value="0" label="Repeat letters that occure > threshold in a rwo" help=""/>
```
bebatut, line 41: Do you have a test testing this parameter?

hugolefeuvre: No I haven't, I'm making one.

SaimMomin12, line 41:
```suggestion
            <param argument="--mask-n-repeat" type="integer" min="0" value="0" label="Repeat letters that occur > threshold in a row" help=""/>
```

tools/mmseqs2/macro.xml
```diff
@@ -38,6 +38,7 @@
...
     </xml>
     <xml name="search_type_aa">
                     <param argument="--search-type" type="select" label="Search type" help="" >
```
SaimMomin12, line 44: Missing help text?

hugolefeuvre: The help part will be the '--mask-n-repeat' argument, I haven't found any information about this parameter other than the Label part.

data_managers/data_manager_mmseqs2_database/data_manager/data_manager_mmseqs2_download.xml
```diff
@@ -122,8 +122,8 @@ cp '$dmjson' '$out_file'
             <output name="out_file">
                 <assert_contents>
                     <has_text text='"mmseqs2_databases":'/>
-                    <has_text text='"version": "15.6f452"'/>
-                    <has_text_matching expression='"value": "SILVA-15.6f452-[0-9]{8}"'/>
+                    <has_text text='"version": "17-b804f"'/>
```
bernt-matthias, line 125: One could use `@TOOL_VERSION@` here.

hugolefeuvre:
I put the PR in draft because the way the tool is wrapped requires the database to be copied to the working directory at each run. This is the solution we found because otherwise the createtaxdb module modifies the real database. It works but it's not optimal. I'd like to try adding this createtaxdb module to the DM part so that I don't have to copy the DB at each run.

hugolefeuvre:
- The DB is no longer copied for each job (except for the test where this is necessary because createtaxdb is applied). 
 - The taxonomic part initially created with createtaxdb is obtained directly when downloading with the DM of the DBs that have already been taxonomically enhanced. 
 - Only DBs with taxonomy are accepted for the mmseqs2_taxonomy_assignment tool.

data_managers/data_manager_mmseqs2_database/data_manager/data_manager_mmseqs2_download.xml
```diff
@@ -40,24 +40,16 @@ cp '$dmjson' '$out_file'
...
                 <option value="aminoacid_taxonomy">Aminoacid databases that can be used for taxonomy</option>
-                <option value="nucleotide">Nucleotide databases</option>
+                <option value="nucleotide">Nucleotide databases without taxonomy</option>
                 <option value="nucleotide_taxonomy">Nucleotide databases that can be used for taxonomy</option>
                 <option value="profile">Profile databases</option>
             </param>
             <when value="aminoacid">
                 <param name="db_type" value="aminoacid" type="hidden"/>
                 <param name="db_taxonomy" value="no" type="hidden"/>
                 <param name="database" type="select" label="MMseqs2 aminoacid databases">
```
bgruening, line 52: this has now only one option, correct? do we need a select here?

hugolefeuvre: Indeed we didn't need a select param, what is the type if we only have one value and we don't want the param to be hidden ?

bernt-matthias: The laternative would be to hardcode it, e.g. by another hidden parameter or just in the command section. 

IMO I think you can leave it as it is. It might get more options in the future and also its transparent to the user.

bgruening:
Thanks!

PR #6922, opened 2025-04-17
Added seaborn plotting suite

hechth:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/seaborn/seaborn.xml
+++ b/tools/seaborn/seaborn.xml
@@ -149,7 +149,7 @@ plt.savefig(f"{output_file}", format=output_format, dpi=300)
                     <has_image_channels channels="4"/>
                     <has_image_height height="1440"/>
                     <has_image_width width="1920" />
-                    <has_image_center_of_mass center_of_mass="971.42, 692.08" eps="0.1"/>
+                    <has_image_center_of_mass center_of_mass="971.48, 692.18" eps="0.2"/>
                 </assert_contents>
             </output>
         </test>


tools/seaborn/.shed.yml
```diff
@@ -0,0 +1,16 @@
...
+categories:
+  - Visualization
+description: "Seaborn is a Python data visualization library based on Matplotlib."
+long_description: "This tool integrates Seaborn, a Python data visualization library, into Galaxy. It supports various plot types such as scatterplots, barplots, boxplots, and pairwise plots. Input data can be provided in tabular, CSV, or Parquet formats."
+type: unrestricted
+auto_tool_repositories:
+  name_template: "{{ tool_id }}"
+  description_template: "{{ tool_name }} tool for data visualization using Seaborn."
+suite:
+  name: suite_seaborn_tools
```
bgruening, line 14:
```suggestion
  name: suite_seaborn
```

tools/seaborn/macros.xml
```diff
@@ -0,0 +1,87 @@
...
+
+    <xml name="requirements">
+        <requirements>
+            <requirement type="package" version="@TOOL_VERSION@">seaborn</requirement>
+            <yield/>
+        </requirements>
+    </xml>
+
+    <xml name="inputs">
+        <param argument="--input_data" type="data" format="tsv,txt,csv,parquet" label="Input data table" help="Provide the input data file in one of the supported formats: TSV, TXT, CSV, or Parquet. This file will be used to generate the plot." />
```
bgruening, line 22: Should we really add txt here? This will include a lot of unrelated datasets.

tools/seaborn/macros.xml
```diff
@@ -0,0 +1,87 @@
...
+    </xml>
+
+    <xml name="transformation">
+        <param name="transformation" type="select" label="Transformation" help="Choose a transformation function to apply to the numerical data in the input file. This can be useful for scaling or normalizing the data before plotting.">
+            <option value="lambda x: x" selected="true">None</option>
+            <option value="np.log10">np.log10</option>
```
bgruening, line 28: Maybe just `log 10`, `log 2` and `no transformation`

tools/seaborn/seaborn.xml
```diff
@@ -0,0 +1,246 @@
+<tool id="seaborn" name="seaborn" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="20.01" license="MIT">
```
bgruening, line 1: higher profile?

tools/seaborn/seaborn.xml
```diff
@@ -0,0 +1,246 @@
...
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+
+    <expand macro="edam"/>
+    <expand macro="requirements"/>
+    <expand macro="creator" />
+
+    <command detect_errors="exit_code"><![CDATA[
+        python3 "${run_script}"
```
bgruening, line 13: single quote please
```suggestion
        python3 '${run_script}'
```

tools/seaborn/seaborn.xml
```diff
@@ -0,0 +1,246 @@
...
+            <option value="sns.displot">displot</option>
+            <option value="sns.jointplot">jointplot</option>
+            <option value="sns.barplot">barplot</option>
+            <option value="sns.boxplot">boxplot</option>
+            <option value="sns.violinplot">violinplot</option>
+            <option value="sns.lmplot">lmplot</option>
+        </param>
+
+        <conditional name="header">
+            <param name="has_header" type="select" display="radio" label="Does your table have a header?" truevalue="yes" falsevalue="no" checked="true" help="Specify whether the input data contains a header row. If yes, column names will be used for selection." refresh_on_change="true">
```
bgruening, line 71:
```suggestion
            <param name="has_header" type="select" label="Does your table have a header?" help="Specify whether the input data contains a header row. If yes, column names will be used for selection." refresh_on_change="true">
```

refresh_on_change is not needed I think, isn't it deprecated already?

tools/seaborn/seaborn.xml
```diff
@@ -0,0 +1,246 @@
...
+            </param>
+            <param name="legend" type="select" label="Legend" help="Specify the type of legend to display on the plot. Options include 'auto', 'brief', 'full', or None to disable the legend." optional="true">
+                <option value="'auto'">auto</option>
+                <option value="'brief'">brief</option>
+                <option value="'full'">full</option>
+                <option value="False">none</option>
+                <sanitizer invalid_char="">
+                    <valid initial="string.ascii_letters">
+                        <add value="'" />
+                    </valid>
```
bgruening, line 101: Not sure what this is doing here. you have the allowed values under control.

tools/seaborn/seaborn.xml
```diff
@@ -0,0 +1,246 @@
...
+        <!-- Test 1: Generate a scatterplot -->
+        <test>
+            <param name="input_data" value="mtcars.txt"/>
+            <param name="plot" value="sns.scatterplot" />
+            <param name="x" value="2" /> <!-- Column 2: mpg -->
+            <param name="y" value="3" /> <!-- Column 3: cyl -->
+            <param name="has_header" value="yes" />
+            <output name="output_file">
+                <assert_contents>
+                    <has_size size="45125" delta="200"/>
```
bgruening, line 127: we do have better tests explicit for images https://docs.galaxyproject.org/en/latest/dev/schema.html#id117 e.g. image_diff

tools/seaborn/seaborn_pairgrid.xml
```diff
@@ -0,0 +1,114 @@
+<tool id="seaborn_pairgrid" name="seaborn pair-wise scatterplot" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="20.01" license="MIT">
+    <description>scatterplot matrix of all-vs-all column combinations</description>
```
bgruening, line 2:
```suggestion
    <description>of all-vs-all column combinations</description>
```

tools/seaborn/seaborn_pairgrid.xml
```diff
@@ -0,0 +1,114 @@
...
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+
+    <expand macro="edam"/>
+    <expand macro="requirements"/>
+    <expand macro="creator" />
+
+    <command detect_errors="exit_code"><![CDATA[
+        python3 "${run_script}"
```
bgruening, line 12:
```suggestion
        python3 '${run_script}'
```

tools/seaborn/seaborn_pairgrid.xml
```diff
@@ -0,0 +1,114 @@
...
+    <outputs>
+        <data name="output_file" format="png" label="${tool.name} on ${on_string}" />
+    </outputs>
+    <tests>
+        <!-- Test 1: Generate a pairgrid plot with default settings -->
+        <test>
+            <param name="input_data" value="mtcars.txt" />
+            <output name="output_file">
+                <assert_contents>
+                    <has_size size="4643277" delta="200"/>
```
bgruening, line 54: please use the better imaging asserts

bgruening (commented):
Thanks @hechth

tools/seaborn/seaborn.xml
hechth: This is needed since something needs to be passed as string and none needs to be passed without - its a bit hacky but this way it works.

tools/seaborn/seaborn.xml
hechth: Its needed, otherwise the UI is buggy - this is the case on some other tools.

hechth:
@bgruening @bernt-matthias seems like the CI is still somewhat broken? Complains that there is no repository, which is true since this is a new tool.

bernt-matthias:
> @bgruening @bernt-matthias seems like the CI is still somewhat broken? Complains that there is no repository, which is true since this is a new tool.

This should be fixed in (https://github.com/galaxyproject/planemo/pull/1494) https://github.com/galaxyproject/planemo/releases/tag/0.75.28

hechth:
bioblend ERROR: GET: error 404: b'{"status": "error", "message": "No repository named seaborn_pairgrid found with owner iuc"}', 0 attempts left

tools/seaborn/macros.xml
hechth: Done

bernt-matthias:
> bioblend ERROR: GET: error 404: b'{"status": "error", "message": "No repository named seaborn_pairgrid found with owner iuc"}', 0 attempts left

I agree that this is confusing, but it's just a logging message from bioblend and not a linter error. xref https://github.com/galaxyproject/bioblend/issues/502

tools/seaborn/seaborn.xml
```diff
@@ -0,0 +1,246 @@
...
+            </output>
+        </test>
+
+        <!-- Test 3: Generate a boxplot with hue -->
+        <test>
+            <param name="input_data" value="mtcars.txt"/>
+            <param name="plot" value="sns.boxplot" />
+            <param name="x" value="3" /> <!-- Column 3: cyl -->
+            <param name="y" value="2" /> <!-- Column 2: mpg -->
+            <param name="hue" value="4" /> <!-- Column 4: disp -->
```
bernt-matthias, line 152: You need to add to add the section to the tests as well to make the new strict linting pass.

tools/seaborn/seaborn_pairgrid.xml
hechth: I didn't want to pollute the repo with big images and keep the tests rather small. Also I think we would figure if the tool stopped creating images.

hechth:
> > bioblend ERROR: GET: error 404: b'{"status": "error", "message": "No repository named seaborn_pairgrid found with owner iuc"}', 0 attempts left
> 
> I agree that this is confusing, but it's just a logging message from bioblend and not a linter error. xref [galaxyproject/bioblend#502](https://github.com/galaxyproject/bioblend/issues/502)

I see - yeah it is a bit confusing that it shows up in the linter section.

hechth:
Maybe @bebatut or @nsoranzo could have a look as well?

tools/seaborn/seaborn_pairgrid.xml
bgruening: You don't need to add images. But @kostrykin has added some nice asserts for images. For example an `image_diff`. https://docs.galaxyproject.org/en/latest/dev/schema.html#id117 or `<has_image_depth depth="512" delta="2" />` or similar. This way we don't need to store images but have more meaningful asserts than the file size.

bgruening: [has_image_height:](https://docs.galaxyproject.org/en/latest/dev/schema.html#has-image-height)

kostrykin: Yes, testing against full image data is probably the best option in most cases, but if it's not preferable, then assertions for image metadata like [`has_image_width`](https://docs.galaxyproject.org/en/latest/dev/schema.html#has-image-width), [`has_image_height`](https://docs.galaxyproject.org/en/latest/dev/schema.html#has-image-height), [`has_image_channels`](https://docs.galaxyproject.org/en/latest/dev/schema.html#has-image-channels) and/or content like [`has_image_center_of_mass`](https://docs.galaxyproject.org/en/latest/dev/schema.html#has-image-center-of-mass), [`has_image_mean_intensity`](https://docs.galaxyproject.org/en/latest/dev/schema.html#has-image-mean-intensity) certainly also is a good option for regression testing.

tools/seaborn/seaborn.xml
```diff
@@ -143,6 +147,10 @@ plt.savefig(f"{output_file}", format=output_format, dpi=300)
             <output name="output_file">
                 <assert_contents>
                     <has_size size="33364" delta="200"/>
```
kostrykin, line 149: I think it's better to remove the `has_size` assertions, since, even though they pass now, they might cause false-positive error detections in regression.

bernt-matthias (approved):
Good from my side. I agree that the size assertions can be removed. The other assertions should be sufficient to make sure that the produced files are not empty.

kostrykin (approved):
Looks good from my side. The linting errors is what we also have been experiencing in other repos since a few weeks, I have no clue what causes them.

hechth:
@kostrykin would it be fine to merge it?

kostrykin:
> @kostrykin would it be fine to merge it?

I think so, but I have no merge rights here

bernt-matthias:
Thanks

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/15005071745

bernt-matthias:
Ah, we need a `.lint_skip` file containing `TestsAssertionValidation` until the failing linter is fixed

hechth:
@mvdbeek how do we now actually get them to be deployed?

bgruening:
@hechth Matthias wrote this in the comment above.

But I guess we get new packages with the real fix upstream really soon, so I will push manually.

bgruening:
Done.

PR #6946, opened 2025-04-30
add bellavista prepare

nilchia:
Large spatial transcriptomic images take a long time to prepare for [bellavista interactive tool](https://usegalaxy.eu/?tool_id=interactive_tool_bellavista&version=latest). 
This tool will prepare inputs in advance to reduce computation time. 

FOR CONTRIBUTOR:


tools/bellavista/bellavista.bash
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/bash
+
+cd "$BELLAVISTA_DIR" &&
+touch './bellavista.log' &&
+/opt/bellavista/bellavista/bin/bellavista "./config.json" | tee './bellavista.log' &&
+TOOL_PID=$!
+
+while sleep 1; do
+    if grep -q "OME-Zarr image saved successfully" './bellavista.log'; then
+        echo "Bella Vista input files created! Stopping the tool..."
```
bgruening, line 10: I think you need an additional safe-guards here.

What happens if this string never appears? Should we kill the tool after X hours? The X hours can be taken optionally from an ENV var, so the admin can overwrite it?

Also would it make sense to actually check if the result file exist?

tools/bellavista/bellavista.bash
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/bash
+
+cd "$BELLAVISTA_DIR" &&
```
bgruening, line 3: I would not change any directory inside this script. Do it outside as part of the cheatha command.

And maybe also consider using `pushd` and `popd` instead of cd

tools/bellavista/bellavista_prepare.xml
```diff
@@ -0,0 +1,180 @@
...
+        #end if
+        #if $technology_conditional.plot_nuclear_segmentation.plot_nuclear_seg == "true":
+            ln -s '$technology_conditional.plot_nuclear_segmentation.nuclear_segmentation' './input/${technology_conditional.plot_nuclear_segmentation.nuclear_segmentation.element_identifier}.${technology_conditional.plot_nuclear_segmentation.nuclear_segmentation.ext}' &&
+        #end if
+        #if $technology_conditional.technology == "MERSCOPE":
+            ln -s '$technology_conditional.um_to_px_transform' './input/micron_to_mosaic_pixel_transform.csv' &&
+        #end if
+        cat ./input/config.json &&
+
+        bash '$__tool_directory__/bellavista.bash'
```
bgruening, line 35:
```suggestion
        pushd '$BELLAVISTA_DIR' &&
        bash '$__tool_directory__/bellavista.bash' &&
        popd &&
```

bgruening (approved):
Ok, lets try this.

PR #6947, opened 2025-05-02
Update beagle tool to handle compressed VCF files and large datasets. 

elmedjadjirayane:
**Added** Xmx option in order to **set maximum RAM usage** for the Java Virtual Machine. 
**Added** option to take **compressed VCF file** as input.
**Added** option to return **compressed VCF file** as output.

FOR CONTRIBUTOR:


tools/beagle/beagle.xml
```diff
@@ -10,9 +10,19 @@
         #if $optional_inputs.ref
             ln -s '${optional_inputs.ref}' ref.$optional_inputs.ref.ext &&
         #end if
-
+        #if $gt.ext=="vcf_bgzip"
+            echo "The input is a compressed VCF file, beagle recocnizes it by adding the .gz extention" &&
+            cp '$gt' tmp.gz &&
```
bgruening, line 15: can you use `ln -s` here?

tools/beagle/beagle.xml
```diff
@@ -49,13 +59,27 @@
...
+            && echo "Moving decompressed beagle output to galaxy output"
+            && mv 'out.vcf' '$vcf_file'
+        #end if
     ]]>    </command>
     <inputs>
         <param argument="gt" type="data" format="vcf" label="VCF file" 
             help="It specifies a VCF file containing genotypes for the study samples. 
                 Each VCF record must contain a GT (genotype) format field"/>
         <section name="optional_inputs" title="Optional input files" expanded="true">
+            <param argument="ram" type="integer" optional="true" label="Specify maximum RAM usage (g/1024m) (optional)"
```
bgruening, line 81: We don't let users specify the memory. Users normally do not know how much memory is needed or how much memory is available in the server.

Instead the Admin can set the amount of memory depending on the destination where the job is running or depending on the input file size.

For example here https://github.com/galaxyproject/tpv-shared-database/blob/main/tools.yml#L46 we set the Java Xmx flag.

bgruening (commented):
Cool, the changes to support gz files look good to me.

The memory parameter we should avoid, this is handled on the Admin level.

tools/beagle/beagle.xml
elmedjadjirayane: Thank you so much for your review. I will remove this option.

tools/beagle/beagle.xml
bgruening:
```suggestion
            ln -s '$gt' tmp.gz &&
```

bgruening: @elmedjadjirayane this should do the trick. The gz extension is still there.

bgruening:
The tests are failing, because the output data has changed slightly. Can you please update those?

elmedjadjirayane:
> The tests are failing, because the output data has changed slightly. Can you please update those?

@bgruening I changed Beagle version since the tool wrapper version is determined by the Beagle version and that the wrapper version must be newer, this is what led to its failure I believe. Do you prefer that I keep a version that is compatible with the current tests or would it be better to update the Beagle version to the latest one and change the outputs?

elmedjadjirayane:
I updated the tests to be compatible with the new Beagle version. They should pass this time.

tools/beagle/beagle.xml
```diff
@@ -49,7 +56,19 @@
...
+            && mv 'out.vcf.gz' '$vcf_file'
+        #else 
+            && echo "decompressing beagle output"
+            && gunzip 'out.vcf.gz'
+            && echo "Moving decompressed beagle output to galaxy output"
+            && mv 'out.vcf' '$vcf_file'
+        #end if
     ]]>    </command>
     <inputs>
         <param argument="gt" type="data" format="vcf" label="VCF file"
```
bgruening, line 74: You want to allow vcf_bgzip here as well do?

Do you mind adding a test for this? You could change one of the existing tests, but provide a gz file.

elmedjadjirayane: Yes, since Beagle gives a compressed VCF file as output by default. The user chooses the output and depending on the user's choice, the tool returns either type. I will add the test for this case.

elmedjadjirayane: Note: by default, the tool returns a VCF.

tools/beagle/beagle.xml
```diff
@@ -49,7 +56,19 @@
...
+        #end if 
+        #if $vcf_file.ext=="vcf_bgzip"
+            && echo "Moving compressed beagle output to galaxy output"
+            && mv 'out.vcf.gz' '$vcf_file'
+        #else 
+            && echo "decompressing beagle output"
+            && gunzip 'out.vcf.gz'
+            && echo "Moving decompressed beagle output to galaxy output"
+            && mv 'out.vcf' '$vcf_file'
+        #end if
```
bgruening, line 71:
```suggestion
        #if $out_format == "vcf_bgzip"
            && mv out.vcf.gz '$vcf_file'
        #else 
            && echo "decompressing beagle output"
            && gunzip 'out.vcf.gz'
            && mv out.vcf '$vcf_file'
        #end if
```

Lets use the param name for the if condition. We also don't need to delete files, this is handled ba Galaxy.

elmedjadjirayane: Are there any modifications that need to be made?

bgruening:
Thanks a lot @elmedjadjirayane! Great work, thanks for contributing to Galaxy tools!

PR #6950, opened 2025-05-04
fix for --variant-input in freebayes

bgruening:
Our user confirmed that this fix is working.

* [ ] bump is needed
* [ ] maybe new freebayes version?
* [ ] profile needed
* [ ] add a test for this case


wm75:
Does that mean https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-inputs-param-conversion got broken at some point?
Seems freebayes is the only iuc tool that used it, but we should still at least document it as broken.

bgruening:
Either this, or it never worked :(
We never had a test for this function.

wm75:
> * [ ]  maybe new freebayes version?

I think it's better to fix the current version before updating to the next one (which we should also do, but in a separate PR).

Going to introduce a profile once tests have passed.

wm75 (approved):

bgruening:
Its unfortunately not a suite based tool. So both will be shipped in one package. I guess is safer if we bump the leftalign as well?

PR #6960, opened 2025-05-09
Update beagle to accept vcf_bgzip in input

elmedjadjirayane:
FOR CONTRIBUTOR:
* [ ] This PR does something else (explain below).

I forgot to add vcf_bgzip in the format for the input in my last pull request.

### DIFF
--- a/tools/beagle/beagle.xml
+++ b/tools/beagle/beagle.xml
@@ -64,7 +64,7 @@
         #end if
     ]]>    </command>
     <inputs>
-        <param argument="gt" type="data" format="vcf" label="VCF file" 
+        <param argument="gt" type="data" format="vcf,vcf_bgzip" label="VCF file" 
             help="It specifies a VCF file containing genotypes for the study samples. 
                 Each VCF record must contain a GT (genotype) format field"/>
         <section name="optional_inputs" title="Optional input files" expanded="true">


tools/gemma/.shed.yml
bgruening, line 1: please extend this file a bit :)

tools/gemma/.shed.yml
```diff
@@ -0,0 +1,3 @@
+description: gemma
+name: gemma
+owner: rayane_elm
```
bgruening, line 3: iuc

tools/beagle/beagle.xml
```diff
@@ -64,7 +64,7 @@
         #end if
     ]]>    </command>
     <inputs>
-        <param argument="gt" type="data" format="vcf" label="VCF file" 
+        <param argument="gt" type="data" format="vcf,vcf_bgzip" label="VCF file"
```
bgruening, line 67: can you please put this into a separate PR?

tools/gemma/gemma.xml
```diff
@@ -0,0 +1,98 @@
+<tool id="gemma-tool" name="GEMMA Analysis" version="1.0.0">
```
bgruening, line 1:
```suggestion
<tool id="gemma" name="GEMMA Analysis" version="1.0.0" profile="23.2">
```

tools/gemma/.shed.yml
elmedjadjirayane: I didn't understand this comment, do you mean I should refer to iuc tools to extend this file?

tools/beagle/beagle.xml
elmedjadjirayane: Yes, I just added it.

tools/gemma/.shed.yml
SaimMomin12: @elmedjadjirayane The owner needs to be changed to IUC

tools/gemma/gemma.xml
```diff
@@ -0,0 +1,98 @@
...
+      <when value="lmm">
+        <param name="kinship" type="data" format="tsv" label="Kinship Matrix File" />
+      </when>
+      <when value="lm"></when>
+      <when value="gk"></when>
+    </conditional>
+  </inputs>
+
+  <outputs>
+    <data name="gemma_output" format="tsv" label="GEMMA output file">
```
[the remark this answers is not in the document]
elmedjadjirayane: > does this file has a column header? If so you can specify here the metadata as names.

It has a column header if and only if the analysis is not kinship calculation (gk). Is there a way to differentiate the metadata following a condition?

tools/gemma/.shed.yml
elmedjadjirayane: The tool is already in the galaxy toolshed and deployed in an the FR instance. How would that affect the deployment.

bgruening:
@elmedjadjirayane do you need any help here?

elmedjadjirayane:
> @elmedjadjirayane do you need any help here?

Hello, I apologize for letting this open for this long but I couldn't make myself available until now. I think there is several issues with the XML file : the owner must be changed to iuc but I don't which impact that would have on instances that already installed it and copying to galaxy object store with cp because of Galaxy's ambiguous prefix with composite datatype "pbed" which has "Composite Dataset" prefix. For the letter i suggest creating symbolic links like this : `ln -s ./*.bed input.bed` for all three files to avoid `cp` and then give gemma the prefix of the symbolic link.
@bgruening @SaimMomin12

elmedjadjirayane:
Hello,

I had made changes to this pull request and the tests has passed, are there other changes that need to be made? @bgruening

elmedjadjirayane:
It seems like the tool version is causing lint to fail. I am not sure how we would fix this since I haven’t followed IUC standards for version before.

bgruening:
You can read more here: https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tool-versions

elmedjadjirayane:
> You can read more here: https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tool-versions

Yes, I am aware of those standards but the issue here is that I think 0.98+suffix will always be less than 1.

tools/gemma/gemma.xml
elmedjadjirayane, line 1: I think we either change the version to something like 1.0.2+galaxy0 and pass the test while being not fully compatible with iuc standards, or simply keep it like this and ignore the lint error.

tools/gemma/.shed.yml
bgruening: Ah, now the liniting error also makes sense.

I was wondering why you get this linting error if the tool is "new".

If we change the owner, the tool would appear twice in the France Galaxy instance. @abretaud and Co would need to hide the older one.

@abretaud what is your suggestion?

abretaud: Hi!

> If we change the owner, the tool would appear twice in the France Galaxy instance. @abretaud and Co would need to hide the older one.

Sure we can hide the older one if you change the owner to iuc, just ping us when it's merge :)

bgruening:
Thanks @elmedjadjirayane! Very good work!

elmedjadjirayane:
> Thanks @elmedjadjirayane! Very good work!

Thank you so much for the help. How will the deployment in Galaxy France be affected @abretaud .

PR #6963, opened 2025-05-09
Ensure text parameter sanitizers aren't bypassed

mvdbeek:
Hoping to get away with not bumping versions ...

FOR CONTRIBUTOR:

### DIFF
--- a/tool_collections/snpsift/snpsift/snpSift_extractFields.xml
+++ b/tool_collections/snpsift/snpsift/snpSift_extractFields.xml
@@ -21,7 +21,7 @@ cat '${input}'
     -e '${empty_text}'
 #end if
 -
-#echo ' '.join(['"%s"' % x for x in $extract.split()])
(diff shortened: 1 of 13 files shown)


bernt-matthias:
What is the behavior before the change and what was wrong with that?

nsoranzo (approved):

mvdbeek:
https://github.com/galaxyproject/galaxy/blob/025da67ede648933d40caf198b067363e0c2b937/lib/galaxy/security/object_wrapper.py#L227 only applies sanitization if you cast to string, which is what happens if you reference `$text_param` in the template, but if you're inside a python expression there's no automatic sanitization if you don't cast to string.

bernt-matthias:
Thanks for the explanation. Can we force sanitization or document this behavior?

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/14965742028

PR #6980, opened 2025-05-14
Add bracken fix and kraken note

paulzierep:
FOR CONTRIBUTOR:

While following a problem from GTA I realized, the bracken option `prebuilt == "no"` does not work, first due to:
`ln -s '${kraken_db.fields.path}/*' '$db_dir/' &&` 
This is fixed now, the DB builds. But this never worked still, since the DB is only built in the working dir.
I am not really sure how to fix it properly, since: (a) Bracken cannot write in the kraken DB folder. So to really fix it we probably have to make a copy of the kraken2 DB, and add the Bracken part. But this will also add a lot of extra code, since 
in the data_manager_conf.xml, we can either move the path or not, but here it would be conditional.
Any suggestions @bernt-matthias @bebatut

### DIFF
--- a/data_managers/data_manager_build_bracken_database/data_manager/bracken_build_database.xml
+++ b/data_managers/data_manager_build_bracken_database/data_manager/bracken_build_database.xml
@@ -38,15 +38,6 @@
                 <validator type="no_options" message="No Kraken2 databases are available" />
             </options>
         </param>
-        <param name="read_len" type="select" label="Read length" help="The read length of the bracken DB should be the same as the length of the reads to be analysis (one read in case of paired reads). The prebuilt DBs where built using the same K-mer length as the kraken2 DB.">
-            <option value="50">50</option>
-            <option value="75">75</option>
-            <option value="100" selected="true">100</option>
-            <option value="150">150</option>
-            <option value="200">200</option>
-            <option value="250">250</option>
-            <option value="300">300</option>
-        </param>
         <conditional name="check_prebuilt">
             <param name="prebuilt" type="select" label="Use Pre-built DB" help="Use existing pre-built DB. This only works for kraken DBs, that already contain additional bracken DBs. This is the case for DBs downloaded from https://benlangmead.github.io/aws-indexes/k2.">
                 <option value="no">No</option>
--- a/data_managers/data_manager_build_bracken_database/data_manager/bracken_build_database.xml
+++ b/data_managers/data_manager_build_bracken_database/data_manager/bracken_build_database.xml
@@ -3,7 +3,7 @@
     <description>bracken database builder</description>
     <macros>
         <token name="@TOOL_VERSION@">2.8</token>
-        <token name="@VERSION_SUFFIX@">0</token>
+        <token name="@VERSION_SUFFIX@">1</token>
         <token name="@PROFILE@">22.01</token>
     </macros>
     <requirements>
@@ -15,7 +15,7 @@
         #set db_dir = os.path.basename($kraken_db.fields.path)
 
         mkdir '$db_dir' &&
-        ln -s '${kraken_db.fields.path}/*' '$db_dir/' &&
+        ln -s '${kraken_db.fields.path}'/* '$db_dir/' &&
         python '$__tool_directory__/bracken_build_database.py'
           '${out_file}'
           
@@ -37,6 +37,15 @@
             <options from_data_table="kraken2_databases">
                 <validator type="no_options" message="No Kraken2 databases are available" />
             </options>
+        </param>
+        <param name="read_len" type="select" label="Read length" help="The read length of the bracken DB should be the same as the length of the reads to be analysis (one read in case of paired reads). The prebuilt DBs where built using the same K-mer length as the kraken2 DB.">
+            <option value="50">50</option>
+            <option value="75">75</option>
+            <option value="100" selected="true">100</option>
+            <option value="150">150</option>
+            <option value="200">200</option>
+            <option value="250">250</option>
+            <option value="300">300</option>
         </param>
         <conditional name="check_prebuilt">
             <param name="prebuilt" type="select" label="Use Pre-built DB" help="Use existing pre-built DB. This only works for kraken DBs, that already contain additional bracken DBs. This is the case for DBs downloaded from https://benlangmead.github.io/aws-indexes/k2.">
--- a/data_managers/data_manager_build_kraken2_database/data_manager/kraken2_build_database.xml
+++ b/data_managers/data_manager_build_kraken2_database/data_manager/kraken2_build_database.xml
@@ -2,14 +2,14 @@
     <description>database builder</description>
     <macros>
         <token name="@TOOL_VERSION@">2.1.3</token>
-        <token name="@VERSION_SUFFIX@">6</token>
+        <token name="@VERSION_SUFFIX@">7</token>
         <token name="@PROFILE@">22.01</token>
         <xml name="common_params">
             <param name="kmer_len" type="integer" value="35" label="K-mer length in BP" />
             <param name="minimizer_len" type="integer" value="31" label="Minimizer length" />
             <param name="minimizer_spaces" type="integer" value="7" label="Minimizer spaces" />
             <param name="load_factor" type="float" value="0.7" min="0" max="1" label="Load factor" help="Proportion of the hash table to be populated" />
-            <param name="clean" type="boolean" truevalue="--clean" falsevalue="" checked="true" label="Clean up extra files" />
+            <param name="clean" type="boolean" truevalue="--clean" falsevalue="" checked="false" label="Clean up extra files. Note: If the extra files are removed this DB cannot be used to build a bracken DB!" />
         </xml>
         <xml name="viral">
             <option value="viral">Viral (viral; ~0.5 GB)</option>


bwlang (approved):
looks good to me

SaimMomin12:
@paulzierep is it ready for review and merge?

paulzierep:
Note, this only fixes part of the issue. The actual problem is still not solved as 
```
I am not really sure how to fix it properly, since: (a) Bracken cannot write in the kraken DB folder. So to really fix it we probably have to make a copy of the kraken2 DB, and add the Bracken part. But this will also add a lot of extra code, since
in the data_manager_conf.xml, we can either move the path or not, but here it would be conditional.
Any suggestions @bernt-matthias @bebatut
```

data_managers/data_manager_build_bracken_database/data_manager/bracken_build_database.xml
```diff
@@ -15,7 +15,7 @@
         #set db_dir = os.path.basename($kraken_db.fields.path)
 
         mkdir '$db_dir' &&
-        ln -s '${kraken_db.fields.path}/*' '$db_dir/' &&
+        ln -s '${kraken_db.fields.path}'/* '$db_dir/' &&
```
bernt-matthias, line 28: Does this need a bash for loop, like 

```
for i in '${kraken_db.fields.path}'/*
do
    ln -s "\$i" '$db_dir/'
done
```

bernt-matthias (commented):
One question is also how big these data bases are? Maybe it's not worth all the effort and we just should accept that we have some duplicate reference data.

paulzierep:
> One question is also how big these data bases are? Maybe it's not worth all the effort and we just should accept that we have some duplicate reference data.

It really is tricky. Often there are a multiple bracken DBs required, that are all build on the same Kraken2 DB. E.g. the prebuild DBs are build like: `All packages contain a Kraken 2 database along with Bracken databases built for 50, 75, 100, 150, 200, 250 and 300-mers. `
If we copy the full kraken2 db for each of the additional bracken2 dbs, thats a lot of duplication.

But OK lets say we copy the folder. The problem for me to write this is that we have 2 different things that need to happen in https://github.com/galaxyproject/tools-iuc/blob/main/data_managers/data_manager_build_bracken_database/data_manager_conf.xml
A) Do nothing for the prebuild option
B) Move the folder for the new build option, like:
```
  <column name="path" output_ref="out_file">
      <move type="directory">
          <source>${path}</source>
          <target base="${GALAXY_DATA_MANAGER_DATA_PATH}">kraken2_databases/${path}</target>
      </move>
      <value_translation>${GALAXY_DATA_MANAGER_DATA_PATH}/kraken2_databases/${path}</value_translation>
      <value_translation type="function">abspath</value_translation>
  </column>
```

Is it currently possible to have conditional behavior for this ? Have not seen it yet.

paulzierep:
If we cannot have both data_manager_conf.xml behaviors in one wrapper, I would propose to split this one to two wrappers:
A) Does only create the entries for the prebuild DBs. And should probably create all entries e.g. Bracken databases built for 50, 75, 100, 150, 200, 250 and 300-mers (since they are there anyway). This DM will not create new data.
B) Copies the kraken2 DB to the backen folder and adds the bracken DB part. Since this option did not work yet, I assume it will not be used too much anyway-

paulzierep:
I actually found an elegant way to support both methods using prebuild and newly build bracken DBs.
For both cases, only the symlinks of kraken2 are transferred to the dedicated bracken DB. 
@bernt-matthias wdyt ?

data_managers/data_manager_build_bracken_database/data_manager/bracken_build_database.xml
paulzierep: works as is

data_managers/data_manager_build_bracken_database/data_manager/bracken_build_database.xml
```diff
@@ -3,29 +3,29 @@
...
     <command>
     <![CDATA[
-        #import os
-        #set db_dir = os.path.basename($kraken_db.fields.path)
+        #set $db_dir = $out_file.extra_files_path
 
         mkdir '$db_dir' &&
-        ln -s '${kraken_db.fields.path}/*' '$db_dir/' &&
+        ln -s '${kraken_db.fields.path}'/* '$db_dir/' &&
         python '$__tool_directory__/bracken_build_database.py'
```
bernt-matthias: Can we simplify this like so (untested) and remove the python script:

```suggestion
        #if $check_prebuilt.prebuilt == "no"
            bracken-build 
                -t "\${GALAXY_SLOTS:-1}"
                -k $check_prebuilt.kmer_len
                -l $check_prebuilt.read_len
                -d '$db_dir/database${check_prebuilt.read_len}mers.kmer_distrib' &&
         #end if
         cp '$dmjson' '$out_file'
```

```xml
<configfiles>
        <configfile name="dmjson"><![CDATA[#slurp
#import uuid
#set $database_value = str(uuid.uuid4())
#set $db_dir = $out_file.extra_files_path
{
  "data_tables":{
    "bracken_databases":[
        {
          "value": database_value,
          "name": $database_name,
          "path": '$db_dir/database${check_prebuilt.read_len}mers.kmer_distrib',
        }
    ]
  }
}
]]></configfile>
    </configfiles>
```

bernt-matthias:
Alternative suggestion: 

We should extend bracken to make use of either the bracken data table or the kraken datatable + the user needs to set a kmer size (which we unfortunately do not know from the datatable, so we need to add a bit of docs).

bernt-matthias:
Hey @paulzierep are the changes I made fine for you? Then we could try to find someone to review this.

bernt-matthias:
Ping @paulzierep

data_managers/data_manager_build_bracken_database/data_manager/bracken_build_database.xml
```diff
@@ -1,37 +1,77 @@
...
+
+        ## Create symlinks to the contents of the kraken database, since otherwise
+        ## the DM tool (ie bracken-build) would write to the kraken DB folder. 
+        ## 
+        ## These symlinks will be part of the final db_dir which will be moved
+        ## to the tool-data (except for some of the pre-exsting *mers.kraken
+        ## and *mers.kmer_distrib which may be deleted -- depending on the
+        ## choice of prebuilt).
+        ## Hence the final db_dir will have
+        ## - one mers.kmer_distrib file / symlink and
```
paulzierep:
```suggestion
        ## - either a created mers.kmer_distrib file or a symlink pointing to the kraken2 db source  
```

data_managers/data_manager_build_bracken_database/data_manager/bracken_build_database.xml
```diff
@@ -1,37 +1,77 @@
...
+        ## Create symlinks to the contents of the kraken database, since otherwise
+        ## the DM tool (ie bracken-build) would write to the kraken DB folder. 
+        ## 
+        ## These symlinks will be part of the final db_dir which will be moved
+        ## to the tool-data (except for some of the pre-exsting *mers.kraken
+        ## and *mers.kmer_distrib which may be deleted -- depending on the
+        ## choice of prebuilt).
+        ## Hence the final db_dir will have
+        ## - one mers.kmer_distrib file / symlink and 
+        ## - one mers.kraken file / symlink
```
paulzierep:
```suggestion
         ## - either a created mers.kraken file or a symlink pointing to the kraken2 db source  
```

data_managers/data_manager_build_bracken_database/data_manager/bracken_build_database.xml
```diff
@@ -1,37 +1,77 @@
...
+        ## - and symlinks to the krakenDB
+        ## Note that only the mers.kmer_distrib file is needed
```
paulzierep: not sure what that means, we do need all files ?

paulzierep:
I like the purely bash based refactor and the symlink logic works !
Minor changes to the comments - mainly to make it more clear to me :) 
There are also prebuild Kraken 2 / Bracken 16s RNA indexes now: https://benlangmead.github.io/aws-indexes/k2 We could add those as well.

paulzierep:
> I like the purely bash based refactor and the symlink logic works ! Minor changes to the comments - mainly to make it more clear to me :) There are also prebuild Kraken 2 / Bracken 16s RNA indexes now: https://benlangmead.github.io/aws-indexes/k2 We could add those as well. But the link logic is different - that sucks

paulzierep:
ping @bernt-matthias added the prebuilt amplicon dbs

data_managers/data_manager_build_kraken2_database/data_manager/kraken2_build_database.xml
```diff
@@ -44,57 +44,147 @@
...
+    #if $database_type.clean
+        #silent command = ["kraken2-build", "--threads", '"${GALAXY_SLOTS:-1}"',
+            "--clean",
+            "--db", "'" + $out_file.extra_files_path + "'/'" + database_value + "'"]
+        #silent commands.append(" ".join(command))
+    #end if
 #end if
+
+#for command in commands
+    #if $run_test_command == "false"
```
paulzierep, line 184: whats the point of this echo ?

data_managers/data_manager_build_bracken_database/data_manager/bracken_build_database.xml
bernt-matthias: This means that in the end only the path to the `xyz.kmer_distrib` file/symlink ends up in the data table, see e.g. the [test data table entry of the tool](https://github.com/galaxyproject/tools-iuc/blob/7ba11a2aa8901b555a95e284e4fb83c8dca2833e/tools/bracken/test-data/bracken_databases.loc#L6)

How about:

```suggestion
        ## Note that in the data table entry only the path to mers.kmer_distrib file is used 
```

data_managers/data_manager_build_kraken2_database/data_manager/kraken2_build_database.xml
bernt-matthias: Good point lets add a comment

```suggestion
    ## In test mode the tool executes `echo COMMAND` instead of `COMMAND`
    #if $run_test_command == "false"
```

bernt-matthias:
I do not understand the need for https://github.com/galaxyproject/tools-iuc/pull/6980/commits/638edc0e39af19bbf50697d6fb4aea3ad6f4c9ec ? In the final loop we execute all commands separate by `&&` anyway.

bernt-matthias (approved):
This seems ready. @paulzierep are you fine with my latest changes?

paulzierep:
yes lets do it !

PR #7034, opened 2025-06-08
new tool: galaxy wrappers for the pcdl physicell data lader command line commands

elmbeech:
FOR CONTRIBUTOR:
* [ yes] I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [ yes] License permits unrestricted use (educational + commercial)
* [ yes] This PR adds a new tool or tool collection
* [ no] This PR updates an existing tool or tool collection
* [ no] This PR does something else (explain below)


tools/pcdl/pcdl_get_anndata.xml
```diff
@@ -0,0 +1,206 @@
+<tool id="pcdl_get_anndata" name="pcdl_get_anndata" version="3.0.1+galaxy0" python_template_version="3.5" profile="21.05">
```
bgruening:
```suggestion
<tool id="pcdl_get_anndata" name="pcdl_get_anndata" version="3.0.1+galaxy0" profile="21.05">
```

tools/pcdl/pcdl_get_anndata.xml
```diff
@@ -0,0 +1,206 @@
...
+    <requirements>
+        <requirement type="package" version="3.3.7">pcdl</requirement>
+    </requirements>
+
+    <stdio> <exit_code range=":" level="log" description="hello world!"/> </stdio>
```
bgruening: hello world is not a good description :)
But I guess this entire line can be removed

tools/pcdl/pcdl_get_anndata.xml
```diff
@@ -0,0 +1,206 @@
...
+            name="physiboss" label="physiboss"
+            type="boolean" truevalue="true" falsevalue="false" checked="true"
+            help="if found, should physiboss state data be extracted and loaded into the anndata object? default is True."
+        />
+        <param
+            name="settingxml" label="settingxml"
+            type="text" value="PhysiCell_settings.xml" optional="true"
+            help="the settings.xml that is loaded, from which the cell type ID label mapping, is extracted, if this information is not found in the output xml file. set to None or False if the xml file is missing! default is PhysiCell_settings.xml."
+        />
+        <param
```
bgruening: The verbose parameter is a bit useless in the Galaxy context. You can set it to true without exposing this parameter to the user.

tools/pcdl/pcdl_get_anndata.xml
```diff
@@ -0,0 +1,206 @@
...
+    <stdio> <exit_code range=":" level="log" description="hello world!"/> </stdio>
+    <command><![CDATA[
+        #import re
+        mkdir output_pc &&
+        #for $file in $path:
+            #set $filename = re.sub('[^\w\-\.\s]', '_', str($file.element_identifier))
+            ln -s $file output_pc/$filename &&
+        #end for
+
+        pcdl_get_anndata 'output_pc' $focus --custom_data_type $custom_data_type --microenv $microenv --graph $graph --physiboss $physiboss --settingxml '$settingxml' --verbose $verbose --drop $drop --keep $keep --scale maxabs --collapse $collapse
```
bgruening: can you break this line into multiple ones, look at other tools in this repo ... this makes it easier to review and to comment.

bgruening:
```suggestion
        pcdl_get_anndata 'output_pc' $focus --custom_data_type $custom_data_type $microenv --graph $graph --physiboss $physiboss --settingxml '$settingxml' --verbose $verbose --drop $drop --keep $keep --scale maxabs --collapse $collapse
```

You could simplify your command here it you use truevalue / falsevalue like below.

tools/pcdl/pcdl_get_anndata.xml
```diff
@@ -0,0 +1,206 @@
...
+
+        <section name="options" title="options:" expanded="true">
+        <param
+            name="custom_data_type" label="custom_data_type"
+            type="text" value="" optional="true"
+            help="parameter to specify custom_data variable types other than float (namely: int, bool, str) like this var:dtype myint:int mybool:bool mystr:str. downstream float and int will be handled as numeric, bool as Boolean, and str as categorical data. default is an empty string."
+        />
+        <param
+            name="microenv" label="microenv"
+            type="boolean" truevalue="true" falsevalue="false" checked="true"
```
bgruening:
```suggestion
            type="boolean" truevalue="--microenv true" falsevalue="" checked="true"
```

tools/pcdl/.shed.yml
```diff
@@ -0,0 +1,11 @@
+categories: [Systems Biology, Data Export, Graphics]
+description: Galaxy wrapper for the PhysiCell data output loader command line commands.
+homepage_url: https://github.com/elmbeech/physicelldataloader
+long_description: Galaxy wrapper for the pcdl PhysiCell Data Loader command line commands.
+  Pcdl is paramount for downstream analysis from PhysiCell output. As such, the pcdl
+  Galaxy tools are useful if you work with the interactive PhysiCell Studio Galaxy
+  tool.
+name: pcdl
+owner: bb-8
```
bgruening:
```suggestion
owner: iuc
```

tools/pcdl/.shed.yml
```diff
@@ -0,0 +1,11 @@
...
+remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/main/tools/pcdl/.shed.yml
```
bgruening:
```suggestion
remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/main/tools/pcdl/
```

tools/pcdl/.shed.yml
bgruening, line 1: this entire file needs to be a suite-like file, similar to https://github.com/galaxyproject/tools-iuc/blob/main/tools/ampvis2/.shed.yml

tools/pcdl/pcdl_get_anndata.xml
```diff
@@ -0,0 +1,206 @@
...
+                        <element name="PhysiCell_settings.xml" value="PhysiCell_settings.xml" />
+                    </collection>
+                </param>
+            </section>
+            <section name="options">
+                <param name="verbose" value="true" />
+                <param name="collapse" value="true" />
+            </section>
+            <output_collection name="anndata_h5ad" count="1">
+                 <element name="timeseries_cell_maxabs" file="timeseries_cell_maxabs.h5ad" ftype="h5ad" />
```
bgruening: we do have some specific asserts that you can use for hdf5 files, I guess this makes your testing more reliable

https://docs.galaxyproject.org/en/latest/dev/schema.html#has-h5-attribute

tools/pcdl/pcdl_get_cell_attribute.xml
```diff
@@ -0,0 +1,145 @@
+<tool id="pcdl_get_cell_attribute" name="pcdl_get_cell_attribute" version="3.0.0+galaxy0" python_template_version="3.5" profile="21.05">
+    <requirements>
```
bgruening, line 5: You could make use of macros and share the "requirements", the "citations" and probably a lot of parameters with multiple tools. This will make maintenance and reuse much easier.

tools/pcdl/pcdl_make_gif.xml
```diff
@@ -0,0 +1,79 @@
...
+        #for $file in $path:
+            #set $filename = re.sub('[^\w\-\.\s]', '_', str($file.element_identifier))
+            ln -s $file timeseries/$filename &&
+        #end for
+
+        pcdl_make_gif 'timeseries' $interface
+    ]]></command>
+
+    <inputs>
+        <section name="positional_arguments" title="positional arguments:" expanded="true">
```
bgruening: This section does not make much sense to a user, does it?

tools/pcdl/pcdl_make_gif.xml
```diff
@@ -0,0 +1,79 @@
...
+        <param
+            name="path" label="path"
```
bgruening: the label "path" is also not very meaningful for users as they do not provide any path

tools/pcdl/pcdl_make_gif.xml
```diff
@@ -0,0 +1,79 @@
+<tool id="pcdl_make_gif" name="pcdl_make_gif" version="3.0.0+galaxy0" python_template_version="3.5" profile="21.05">
```
bgruening: Is this a very generic tool that takes images and creates a gif? Or are there some PCDL specific bits in it?

tools/pcdl/pcdl_make_movie.xml
```diff
@@ -0,0 +1,87 @@
...
+        #for $file in $path:
+            #set $filename = re.sub('[^\w\-\.\s]', '_', str($file.element_identifier))
+            ln -s $file timeseries/$filename &&
+        #end for
+
+        pcdl_make_movie 'timeseries' $interface --framerate $framerate
+    ]]></command>
+
+    <inputs>
+        <section name="positional_arguments" title="positional arguments:" expanded="true">
```
bgruening: please remove those sections, if they don't enhance UX for users

tools/pcdl/pcdl_make_movie.xml
```diff
@@ -0,0 +1,87 @@
...
+        <requirement type="package" version="3.3.7">pcdl</requirement>
+    </requirements>
+
+    <stdio> <exit_code range=":" level="log" description="hello world!"/> </stdio>
+    <command><![CDATA[
+        #import re
+        mkdir timeseries &&
+        #for $file in $path:
+            #set $filename = re.sub('[^\w\-\.\s]', '_', str($file.element_identifier))
+            ln -s $file timeseries/$filename &&
```
bgruening:
```suggestion
            ln -s '$file' timeseries/$filename &&
```

Single-quote, for security reasons, all "path" and "text" parameters

tools/pcdl/pcdl_get_anndata.xml
elmbeech: hi @bgruening , 
thank you for reviewing!

unfortunately, this line cannot be removed.
if you write a python package with command line hocks, the python output is piped to sterr.
if i remove this line, galaxy thinks it always caught an error when the script runs successfully. 
this is why i have to catch all exit_codes as log.
i agree, `hello world!` is not a good description.
i will change this to `catch python script output.`

would this be acceptable?

tools/pcdl/pcdl_get_anndata.xml
elmbeech: ok! will do that!

tools/pcdl/pcdl_get_anndata.xml
bgruening: I think what you are looking for is https://docs.galaxyproject.org/en/latest/dev/schema.html#error-detection e.g. `detect_errors="exit_code"` on the command tag. See other Python based tools here in IUC.

tools/pcdl/pcdl_get_anndata.xml
elmbeech: if possible, i would like to have it there for the user.
here are my arguments:
1. my aim is, that the galaxy wrapper not differ dor the regular pcdl command line command.
2. if you in the history, click the "info" (dataset details) you can see the difference between verbose and non-verbose in "Tool Standard Output". this might very well be interesting information for the power user.

tools/pcdl/pcdl_get_anndata.xml
elmbeech: ok! will do.

elmbeech: I don't think so. `microenv` is not a positional argument.
here is the command line help:

```
usage: pcdl_get_anndata [-h] [--custom_data_type [CUSTOM_DATA_TYPE ...]] [--microenv MICROENV] [--graph GRAPH] [--physiboss PHYSIBOSS] [--settingxml SETTINGXML] [-v VERBOSE]
                        [--drop [DROP ...]] [--keep [KEEP ...]] [--scale SCALE] [--collapse COLLAPSE]
                        [path] [values]

function to transform mcds time steps into one or many anndata objects for downstream analysis.

positional arguments:
  path                  path to the PhysiCell output directory or a outputnnnnnnnn.xml file. default is . .
  values                minimal number of values a variable has to have in any of the mcds time steps to be outputted. variables that have only 1 state carry no information. None is a
                        state too. default is 1.

options:
  -h, --help            show this help message and exit
  --custom_data_type [CUSTOM_DATA_TYPE ...]
                        parameter to specify custom_data variable types other than float (namely: int, bool, str) like this var:dtype myint:int mybool:bool mystr:str . downstream float
                        and int will be handled as numeric, bool as Boolean, and str as categorical data. default is an empty string.
  --microenv MICROENV   should the microenvironment be extracted and loaded into the anndata object? setting microenv to False will use less memory and speed up processing, similar to
                        the original pyMCDS_cells.py script. default is True.
  --graph GRAPH         should neighbor graph, attach graph, and attached spring graph be extracted and loaded into the anndata object? default is True.
  --physiboss PHYSIBOSS
                        if found, should physiboss state data be extracted and loaded into the anndata object? default is True.
  --settingxml SETTINGXML
                        the settings.xml that is loaded, from which the cell type ID label mapping, is extracted, if this information is not found in the output xml file. set to None or
                        False if the xml file is missing! default is PhysiCell_settings.xml.
  -v VERBOSE, --verbose VERBOSE
                        setting verbose to False for less text output, while processing. default is True.
  --drop [DROP ...]     set of column labels to be dropped for the dataframe. don't worry: essential columns like ID, coordinates and time will never be dropped. Attention: when the keep
                        parameter is given, then the drop parameter has to be an empty string! default is an empty string.
  --keep [KEEP ...]     set of column labels to be kept in the dataframe. set values=1 to be sure that all variables are kept. don't worry: essential columns like ID, coordinates and
                        time will always be kept. default is an empty string.
  --scale SCALE         specify how the data should be scaled. possible values are None, maxabs, minmax, std. None: no scaling. set scale to None if you would like to have raw data or
                        entirely scale, transform, and normalize the data later. maxabs: maximum absolute value distance scaler will linearly map all values into a [-1, 1] interval. if
                        the original data has no negative values, the result will be the same as with the minmax scaler (except with attributes with only one value). if the attribute has
                        only zeros, the value will be set to 0. minmax: minimum maximum distance scaler will map all values linearly into a [0, 1] interval. if the attribute has only one
                        value, the value will be set to 0. std: standard deviation scaler will result in sigmas. each attribute will be mean centered around 0. ddof delta degree of
                        freedom is set to 1 because it is assumed that the values are samples out of the population and not the entire population. it is incomprehensible to me that the
                        equivalent sklearn method has ddof set to 0. if the attribute has only one value, the value will be set to 0. default is maxabs
  --collapse COLLAPSE   should all mcds time steps from the time series be collapsed into one big anndata h5ad file, or a many h5ad, one h5ad for each time step?, default is True.

homepage: https://github.com/elmbeech/physicelldataloader
```

tools/pcdl/pcdl_make_gif.xml
elmbeech: yes, it does, if you take a look at the command line help above.
in my opinion, the galaxy wrapper should really just be a wrapper for the regular command line command.

tools/pcdl/pcdl_make_gif.xml
elmbeech: same argument as above.
in my opinion, it is very useful for a user, if the tools have the same parameter labels in python, at the command line, and in galaxy. galaxy should not differ.

tools/pcdl/pcdl_make_gif.xml
elmbeech: it is generic, but is one of the python and command line command that pcdl offers. so, i wanna offer it as well in galaxy.

tools/pcdl/pcdl_make_movie.xml
elmbeech: same argument as above.

tools/pcdl/pcdl_make_movie.xml
elmbeech: ok! will do.

tools/pcdl/pcdl_get_anndata.xml
bgruening: Yes, we do capture the verbose output, that why I think just enable it by default if it contains useful information. This way all user have all informations. And then the need for this param in the UI is not needed anymore.

tools/pcdl/pcdl_get_anndata.xml
bgruening: This change goes hand in hand with the below one to set truevlaue="--microenv True".
It is not important, I just wanted to point you to this possibility. This gets more interesting if you have the more traditional option of "--microenv" as a boolean.

tools/pcdl/pcdl_make_gif.xml
bgruening: The IUC is really trying hard to make the UI even more useable than a CLI. Improving help text, arranging params in sections that cluster nicely together, avoiding params that we can deduce elsewhere etc ...
I'm not sure which user-group you are targeting, but I'm petty sure many Galaxy users have no idea what "positional arguments" mean.

tools/pcdl/pcdl_make_gif.xml
bgruening: Sorry, but "path" will be very confusing, you are not putting in here a string. Please feel encouraged to design a graphical UI that goes beyond the UX of the CLI. You can still add a `help="(path)"` in the argument to indicate the connection to the CLI tool. Or for non-positional arguments use `<param argument="--foo"` which will do this automatically for you.

tools/pcdl/pcdl_get_anndata.xml
elmbeech: No, exactly not.
As you can see, if I do so an error is detected, even if there is none. 
image

elmbeech:
I think this is what you mean with "langauage server". 
Looks like i have to check my implementation with vscode . 
ok, i can do that. 
+ https://github.com/galaxyproject/galaxy-language-server

Could you please point me to what you mean with "macro"?
Thank you!

tools/pcdl/pcdl_get_anndata.xml
bgruening: Ok, that mean that your tool is reporting an error code of 1, is that the case?

For any other tool, text on stderr is not a problem, you can instruct Galaxy to only react on error codes, not text on stderr.

bgruening:
> I think this is what you mean with "langauage server". Looks like i have to check my implementation with vscode . ok, i can do that.
> 
>     * https://github.com/galaxyproject/galaxy-language-server
> 
> 
> Could you please point me to what you mean with "macro"? Thank you!

Sure. https://planemo.readthedocs.io/en/latest/writing_advanced.html#macros-reusable-elements
e.g. here: https://github.com/galaxyproject/tools-iuc/blob/main/tools/bedtools/macros.xml

tools/pcdl/pcdl_get_anndata.xml
elmbeech: no it is not reporting an error code of 1.
as you can see, in this particular command (which in the pcdl command line version just output to the screen) , the bash line is catching the standard error with `2>` and is writing this to `version.txt`.
and galaxy interpret this as error code 1.

elmbeech: i followed this instruction in the galaxy documentation:
+ https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-stdio-exit-code

elmbeech:
thank you for the macro link.
will look into this.

tools/pcdl/pcdl_make_gif.xml
elmbeech: ok. will do so. will change.

tools/pcdl/pcdl_make_gif.xml
elmbeech: ok. will do so. will change.

tools/pcdl/pcdl_get_anndata.xml
elmbeech: yes, but this might be a lot of output and uses up space that is not needed.
additionally, no-verbose processing is faster.
I would prefer to have it as an option with the default set to false.

tools/pcdl/pcdl_get_anndata.xml
elmbeech: I understand now, what you are trying to point out!
Frankly, I prefer my code style.

tools/pcdl/pcdl_get_anndata.xml
elmbeech: I understand now, what you are trying to point out!
Frankly, I prefer my code style.

tools/pcdl/.shed.yml
elmbeech: ok, will do so!

tools/pcdl/.shed.yml
elmbeech: ok! will do so.

tools/pcdl/.shed.yml
elmbeech: thank you for the link! i will look into it.

tools/pcdl/pcdl_make_movie.xml
elmbeech: ok! will do so.

tools/pcdl/pcdl_get_cell_attribute.xml
elmbeech: thank you for the macro link. 
i will look into this.

tools/pcdl/pcdl_get_anndata.xml
elmbeech: thank you for the link! i will look into it.

tools/pcdl/pcdl_make_gif.xml
elmbeech: Look, numerous parameters are there to reduce RAM consumption and offer processing speed up, if possible.
I think it would be bad to kick out all these options from the GUI because we assume the users are not clever enough to use them.
I can rearrange the parameters and make an "essential" and a collapsed "advanced" section, if this is ok, but I really don't wanna kick out these parameters!

bgruening: Good example. RAM consumption, or memory allocation, is done by the admin or the server and should not be handled by the Galaxy tool logic. A user, even an experienced user, does not know at which server the tool will run, or how much memory the node will have. Nor does the user know how much memory is available. So Galaxy has an entire subsystem to do this for each tool, configurable by the server admin, the compute provider etc  ...

Lets maybe consider an extreme case like number of CPUs. We don't want user to specify the number of CPUs for the given arguments above - at least not on public Galaxy servers. If the user provides a number that is too high, its might be inefficient, or the job might not start at all. If the number is too low the tool might crash (there are tools like that ...). Since Galaxy knows about available "destinations" we adjust the CPU count to fit optimally into the infrastructure where it is supposed to be running.

I hope that makes sense.

elmbeech: Yes, this makes sense. What I am trying to say is something totally different.

When we do agent-based model simulation, we record agents, the substrate concentrations at each location in the domain, and possible internal state of the cells. Now when you do a sudden plot or downstream analysis where you e.g. would like to analyze the agent, but you will for that not need the information you have about the substrate concentration at every place in the domain, and the internal cell states you should be able to tell that you not like to load this data because it will cause unnecessary use of resources and time. And only you, who is analyzing the data, knows whether you have to load this data or not. Your subsystem will not know or be able to help you there.
When you do a one-off plot or analysis, this does not matter. But when you analyze a parameter sweep with hundreds or even thousands of runs in a workflow, then it does, and an advanced user should be able to finetune this.

tools/pcdl/pcdl_get_anndata.xml
elmbeech: hmm .... 
there migh be more than one h5ad file.
how can i assert_contents form a output collection?
something like this on an `element` of an `output_collection`:

````
            <assert_contents>                                                   
                <has_h5_keys keys="X" />                                        
                <has_h5_keys keys="layers" />                                   
                <has_h5_keys keys="obs/ID, obs/cell_type, obs/chemotaxis_index, obs/current_death_model, obs/current_phase, obs/cycle_model, obs/id_time, obs/time, obs/z_layer" />
                <has_h5_keys keys="obsm/spatial" />                             
                <has_h5_keys keys="uns" />                                      
                <has_h5_keys keys="var/_index" />                               
                <has_h5_keys keys="varm" />                                     
                <has_h5_keys keys="varp" />                                     
            </assert_contents>                                                  
````

**Comment by elmbeech** [comment]
hi @bgruening !

I think I addresse all issues you were asking for.
Could you please take another look?
Locally, `planemo lint` and `planemo test` is passing.

thank you, Elmar

image

**Comment by elmbeech** [comment]
hmmm, in the combined chunked test result is written: 
`Unsuccessful tests found, inspect the 'All tool test results' artifact for details.`
where can i find the "All tool test results artifact"?
and the check workflow success said i have an exit status 1.
should in this case locally `planemo test` not fail too?

**Comment by bgruening** [comment]
You can find the xunit/error reports here: https://github.com/galaxyproject/tools-iuc/actions/runs/16511690042?pr=7034

When one test files with error code 1, planemo should not fail, you probably want to know what the others tests do, if they also fail or if only test 1 fails.

**Comment by bgruening** [comment]
Looking at your error report it seems some test files are missing:

* output00000000_oxygen.jpeg
* timeseries_cell_maxabs.h5ad
* output00000000_cell.csv

etc

File: tools/pcdl/pcdl_macros.xml [hunk 15]
```diff
@@ -0,0 +1,257 @@
+<macros>
+    <!-- requirements -->
+    <xml name="requirement">
+        <requirement type="package" version="3.3.8">pcdl</requirement>
+    </xml>
+    <!-- input -->
+    <xml name="aggregate_num">
+        <param name="aggregate_num" label="aggregate_num" type="select" display="radio" help="aggregation function {max, mean, median, min, std, var} for focus_num data. default is mean.">
```
bgruening: Can the label maybe more human readable. Something like `Choose the aggregrate function for the focus number data`

tools/pcdl/pcdl_macros.xml
```diff
@@ -0,0 +1,257 @@
...
+            <option value="max">max</option>
+            <option value="mean" selected="true">mean</option>
+            <option value="median">median</option>
+            <option value="min">min</option>
+            <option value="std">std</option>
+            <option value="var">var</option>
+        </param>
+    </xml>
+    <xml name="allvalues">
+        <param name="allvalues" label="allvalues" type="boolean" truevalue="true" falsevalue="false" checked="false" help="for numeric data, should only the min and max values or all values be returned? default is false."/>
```
bgruening:
```suggestion
        <param argument="--allvalues" label="Which values should be returned" type="boolean" truevalue="--allvalues" falsevalue="" checked="false" help="For numeric data, should only the min and max values or all values be returned? default is false."/>
```

tools/pcdl/pcdl_get_cell_attribute.xml
```diff
@@ -0,0 +1,70 @@
...
+
+        pcdl_get_cell_attribute 'output_pc' $entropy
+        --custom_data_type $custom_data_type
+        --microenv $microenv
+        --physiboss $physiboss
+        --settingxml '$settingxml'
+        --verbose $verbose
+        --drop $drop
+        --keep $keep
+        --allvalues $allvalues
```
bgruening, line 24:
```suggestion
        $allvalues
```

The the other comment about allvalues.
If you have booleans you can use this trick here. Because the truevalue will be `--allvalues true` for example.

tools/pcdl/pcdl_macros.xml
bgruening: The `argument="--allvalues"` is a nice extension, to indicate this is the real argument for your tool.

It will also add `(--allvalues)` to the parameter help text, so people used to the CLI or want to learn the CLI can reuse them.

tools/pcdl/pcdl_macros.xml
```diff
@@ -0,0 +1,257 @@
...
+    </xml>
+    <xml name="alpha">
+        <param name="alpha" label="alpha" type="float" value="1.0" optional="false" help="alpha channel transparency value between 1 (not transparent at all) and 0 (totally transparent). default is 1.0."/>
```
bgruening:
```suggestion
        <param argument="--alpha" label="alpha" type="float" value="1.0" optional="false" help="alpha channel transparency value between 1 (not transparent at all) and 0 (totally transparent). default is 1.0."/>
```

Should work as it is. If you like please do that for all arguments.

tools/pcdl/pcdl_macros.xml
```diff
@@ -0,0 +1,257 @@
...
+    </xml>
+    <xml name="attribute">
+        <param name="attribute" label="attribute" type="text" value="cell_type" optional="false" help="listing of mcds.get_cell_df dataframe column names, used for cell attributes. default is a single term: cell_type."/>
+    </xml>
+    <xml name="cmap">
+        <param name="cmap" label="cmap" type="text" value="viridis" optional="false" help="matplotlib colormap string from https://matplotlib.org/stable/tutorials/colors/colormaps.html . default is viridis."/>
```
bgruening:
```suggestion
        <param argument="--cmap" label="cmap" type="text" value="viridis" optional="false" help="Matplotlib colormap string from https://matplotlib.org/stable/tutorials/colors/colormaps.html . default is viridis."/>
```

But I think you should steal this https://github.com/galaxyproject/tools-iuc/blob/6693d8dba096b5905523f4369cb66e6507290811/tools/schicexplorer/macros.xml#L38

tools/pcdl/pcdl_macros.xml
```diff
@@ -0,0 +1,257 @@
...
+    </xml>
+    <xml name="collapse">
+        <param name="collapse" label="collapse" type="boolean" truevalue="true" falsevalue="false" checked="true" help="should all mcds time steps from the time series be collapsed into one big anndata h5ad file, or a many h5ad, one h5ad for each time step?, default is True."/>
```
bgruening:
```suggestion
        <param argument="--collapse" label="collapse" type="boolean" truevalue="--collapse" falsevalue="" checked="true" help="should all mcds time steps from the time series be collapsed into one big anndata h5ad file, or a many h5ad, one h5ad for each time step?, default is True."/>
```

Some trick, you need to change the commandline section as well if you take this over.

tools/pcdl/pcdl_macros.xml
```diff
@@ -0,0 +1,257 @@
...
+        <param name="ext" label="ext" type="select" display="radio" help="output image format. possible formats are jpeg, png, and tiff. default is jpeg.">
+            <option value="jpeg" selected="true">jpeg</option>
+            <!--
+        <option value="png">png</option>
+        <option value="tiff">tiff</option>
+        -->
+        </param>
+    </xml>
+    <xml name="extrema">
+        <param name="extrema" label="extrema" type="text" value="none" optional="false" help="two floats separated by a space. None takes min and max from data. default is None."/>
```
bgruening:
```suggestion
        <param name="extrema" label="extrema" type="text" value="none" optional="false" help="Two floats separated by a space. None takes min and max from data. default is None."/>
```

The help is usually a full sentence, please start with capital letters.

tools/pcdl/pcdl_macros.xml
```diff
@@ -0,0 +1,257 @@
...
+        <param name="cmap" label="cmap" type="text" value="viridis" optional="false" help="matplotlib colormap string from https://matplotlib.org/stable/tutorials/colors/colormaps.html . default is viridis."/>
+    </xml>
+    <xml name="collapse">
+        <param name="collapse" label="collapse" type="boolean" truevalue="true" falsevalue="false" checked="true" help="should all mcds time steps from the time series be collapsed into one big anndata h5ad file, or a many h5ad, one h5ad for each time step?, default is True."/>
+    </xml>
+    <xml name="color">
+        <param name="color" label="color" type="text" value="none" optional="false" help="listing of color strings referred to by name, RGB or RGBA code. default is None."/>
+    </xml>
+    <xml name="custom_data_type">
+        <param name="custom_data_type" label="custom_data_type" type="text" value="" optional="true" help="parameter to specify custom_data variable types other than float (namely: int, bool, str) like this var:dtype myint:int mybool:bool mystr:str. downstream float and int will be handled as numeric, bool as Boolean, and str as categorical data. default is an empty string."/>
```
bgruening: maybe a select?

tools/pcdl/pcdl_macros.xml
```diff
@@ -0,0 +1,257 @@
...
+        </param>
+    </xml>
+    <xml name="extrema">
+        <param name="extrema" label="extrema" type="text" value="none" optional="false" help="two floats separated by a space. None takes min and max from data. default is None."/>
+    </xml>
+    <xml name="figbgcolor">
+        <param name="figbgcolor" label="figbgcolor" type="text" value="none" optional="false" help="figure background color. None is transparent (png) or white (jpeg, tiff). default is None."/>
+    </xml>
+    <xml name="figsizepx">
+        <param name="figsizepx" label="figsizepx" type="text" value="640 480" optional="false" help="size of the figure in pixels (integer), x y. the given x and y will be rounded to the nearest even number, to be able to generate movies from the images. default is 640 480."/>
```
bgruening: Two integer fields maybe?

Its like type-hits, the API will be much nicer if we model it with types.

tools/pcdl/pcdl_macros.xml
bgruening: Please consider using this for all booleans.

tools/pcdl/pcdl_macros.xml
```diff
@@ -0,0 +1,257 @@
...
+        <param name="interface" label="interface" type="select" display="radio" help="specify the image format from which the gif will be generated. these images can be generated with the pcdl_plot_scatter or pcdl_plot_contour function. default is jpeg.">
+            <option value="jpeg" selected="true">jpeg</option>
+            <!--
+        <option value="png">png</option>
+        <option value="tiff">tiff</option>
+        -->
+        </param>
+    </xml>
+    <xml name="keep">
+        <param name="keep" label="keep" type="text" value="" optional="true" help="set of column labels to be kept in the dataframe. set values=1 to be sure that all variables are kept. don't worry: essential columns like ID, coordinates and time will always be kept. default is an empty string."/>
```
bgruening: I don't undestand this parameter. Is only values=1 possible? Or also values="foo,bar" etc ...

tools/pcdl/pcdl_macros.xml
```diff
@@ -0,0 +1,257 @@
...
+        <param name="attribute" label="attribute" type="text" value="cell_type" optional="false" help="listing of mcds.get_cell_df dataframe column names, used for cell attributes. default is a single term: cell_type."/>
+    </xml>
+    <xml name="cmap">
+        <param name="cmap" label="cmap" type="text" value="viridis" optional="false" help="matplotlib colormap string from https://matplotlib.org/stable/tutorials/colors/colormaps.html . default is viridis."/>
+    </xml>
+    <xml name="collapse">
+        <param name="collapse" label="collapse" type="boolean" truevalue="true" falsevalue="false" checked="true" help="should all mcds time steps from the time series be collapsed into one big anndata h5ad file, or a many h5ad, one h5ad for each time step?, default is True."/>
+    </xml>
+    <xml name="color">
+        <param name="color" label="color" type="text" value="none" optional="false" help="listing of color strings referred to by name, RGB or RGBA code. default is None."/>
```
bgruening: select?

tools/pcdl/pcdl_macros.xml
```diff
@@ -0,0 +1,257 @@
...
+    <xml name="linestyle">
+        <param name="linestyle" label="linestyle" type="select" display="radio" help="matplotlib line style {-, --, .-, :} string. default is - .">
+            <option value="-" selected="true">-</option>
+            <option value="--">--</option>
+            <option value=".-">.-</option>
+            <option value=":">:</option>
+        </param>
+    </xml>
+    <xml name="linewidth">
+        <param name="linewidth" label="linewidth" type="text" value="none" optional="false" help="line width in points, integer. default is None."/>
```
bgruening: integer?

tools/pcdl/pcdl_macros.xml
```diff
@@ -0,0 +1,257 @@
...
+        <param name="node_attribute" label="node_attribute" type="text" value="" optional="true" help="listing of pcdl_get_cell_attribute attributes, used for node attributes. default is and empty list."/>
+    </xml>
+    <xml name="path">
+        <param name="path" label="data_collection" type="data_collection" collection_type="list" help="PhysiCell output data collection."/>
+    </xml>
+    <xml name="physiboss">
+        <param name="physiboss" label="physiboss" type="boolean" truevalue="true" falsevalue="false" checked="true" help="if found, should physiboss state data be extracted and loaded into the anndata object? default is True."/>
+    </xml>
+    <xml name="s">
+        <param name="s" label="s" type="float" value="1.0" optional="false" help="scatter plot dot size scale factor. with figsizepx extracted from initial.svg, scale factor 1.0 should be ok. adjust if necessary. default 1.0."/>
```
bgruening: min=0.01 or something like that? Maybe also add a "max"?

tools/pcdl/pcdl_macros.xml
```diff
@@ -0,0 +1,257 @@
...
+    <xml name="scale">
+        <param name="scale" label="scale" type="select" display="radio" help="specify how the data should be scaled. possible values are None, maxabs, minmax, std. None: no scaling. set scale to None if you would like to have raw data or entirely scale, transform, and normalize the data later. maxabs: maximum absolute value distance scaler will linearly map all values into a [-1, 1] interval. if the original data has no negative values, the result will be the same as with the minmax scaler (except with attributes with only one value). if the attribute has only zeros, the value will be set to 0. minmax: minimum maximum distance scaler will map all values linearly into a [0, 1] interval. if the attribute has only one value, the value will be set to 0. std: standard deviation scaler will result in sigmas. each attribute will be mean centered around 0. ddof delta degree of freedom is set to 1 because it is assumed that the values are samples out of the population and not the entire population. it is incomprehensible to me that the equivalent sklearn method has ddof set to 0. if the attribute has only one value, the value will be set to 0. default is maxabs.">
```
bgruening: That is a long help text, can it be shorted and the long version be moved into the tool-help section?

tools/pcdl/pcdl_macros.xml
```diff
@@ -0,0 +1,257 @@
...
+            <option value="maxabs" selected="true">maxabs</option>
+            <option value="minmax">minmax</option>
+            <option value="std">std</option>
+        </param>
+    </xml>
+    <xml name="secondary_y">
+        <param name="secondary_y" label="secondary_y" type="boolean" truevalue="true" falsevalue="false" checked="false" help="whether to plot on the secondary y-axis. if a listing of string, which columns to plot on the secondary y-axis. default is False."/>
+    </xml>
+    <xml name="settingxml">
+        <param name="settingxml" label="settingxml" type="text" value="PhysiCell_settings.xml" optional="true" help="the settings.xml that is loaded, from which the cell type ID label mapping, is extracted, if this information is not found in the output xml file. set to None or False if the xml file is missing! default is PhysiCell_settings.xml."/>
```
bgruening: this does not make sense in the Galaxy context? It should be a data parameter I think.

tools/pcdl/pcdl_macros.xml
```diff
@@ -0,0 +1,257 @@
...
+    </xml>
+    <xml name="sharex">
+        <param name="sharex" label="sharex" type="boolean" truevalue="true" falsevalue="false" checked="false" help="in case subplots is True, share x-axis by setting some x-axis labels to invisible. default is False."/>
+    </xml>
+    <xml name="sharey">
+        <param name="sharey" label="sharey" type="boolean" truevalue="true" falsevalue="false" checked="false" help="in case subplots is True, share y-axis range and possibly setting some y-axis labels to invisible. default is False."/>
+    </xml>
+    <xml name="subplots">
+        <param name="subplots" label="subplots" type="boolean" truevalue="true" falsevalue="false" checked="false" help="whether to split the plot into subplots, one per column. default is False."/>
```
bgruening: The suffix `default is False.` is also not need in Galaxy. Since Galaxy is displaying the default.

tools/pcdl/pcdl_macros.xml
```diff
@@ -0,0 +1,257 @@
...
+    </xml>
+    <xml name="title">
+        <param name="title" label="title" type="text" value="none" optional="false" help="title to use for the plot. None will print no title. default is None."/>
```
bgruening: You probably want to restrict / "validator" the input here to only characters and numbers? Alphanumeric characters?

Same for below.

tools/pcdl/pcdl_macros.xml
```diff
@@ -0,0 +1,257 @@
...
+        <param name="verbose" label="verbose" type="boolean" truevalue="true" falsevalue="false" checked="false" help="setting verbose to True for more text output, while processing. default is False."/>
+    </xml>
+    <xml name="xlim">
+        <param name="xlim" label="xlim" type="text" value="none" optional="false" help="two floats separated by a space. x axis min and max value. None takes min and max from mesh x axis range. default is None."/>
+    </xml>
+    <xml name="xyequal">
+        <param name="xyequal" label="xyequal" type="boolean" truevalue="true" falsevalue="false" checked="true" help="to specify equal axis spacing for x and y axis. default is true."/>
+    </xml>
+    <xml name="ylim">
+        <param name="ylim" label="ylim" type="text" value="none" optional="false" help="two floats separated by a space. y axis min and max value. None takes min and max from mesh y axis range. default is None."/>
```
bgruening: model as two floats :) and then each with min/max?

tools/pcdl/pcdl_macros.xml
```diff
@@ -0,0 +1,257 @@
...
+    </xml>
+    <xml name="yunit">
+        <param name="yunit" label="yunit" type="text" value="none" optional="false" help="string to specify y-axis unit. None will not print a unit on the y-axis. default is None."/>
+    </xml>
+    <xml name="z_axis">
+        <param name="z_axis" label="z_axis" type="text" value="none" optional="false" help="for a categorical focus: list of labels; for a numeric focus: list of two floats; None, depending on the focus column variable dtype, extracts labels or min and max values from data. default is None."/>
+    </xml>
+    <xml name="z_slice">
+        <param name="z_slice" label="z_slice" type="float" value="0.0" optional="false" help="z-axis position to slice a 2D xy-plain out of the 3D mesh. if z_slice position numeric but not an exact mesh center coordinate, then z_slice will be adjusted to the nearest mesh center value, the smaller one, if the coordinate lies on a saddle point. default is 0.0."/>
```
bgruening: min=0.0? max?

bgruening (commented):
This looks really nice @elmbeech!

I added a bunch of comments that should improve future updates but also UX for the user. Please let me know if you have any questions.

tools/pcdl/pcdl_macros.xml
elmbeech: will look into this.

tools/pcdl/pcdl_macros.xml
elmbeech: i will look into this.
does this mean, when i use the `param` parameter, i do not need a `name` parameter?

tools/pcdl/pcdl_macros.xml
elmbeech: yep! look into this.

tools/pcdl/pcdl_macros.xml
elmbeech: will look into it.

tools/pcdl/pcdl_macros.xml
elmbeech: yes, i get your point!
i might  do this for all. i like that in this way the cli argument is displayed too.

tools/pcdl/pcdl_macros.xml
elmbeech: : ) . i will look into it.

tools/pcdl/pcdl_macros.xml
elmbeech: no, this will not work. this is model specific.
it is based on custom varaiables user add.

tools/pcdl/pcdl_macros.xml
elmbeech: i can try. 
the challenge will be to transform the two files into the proper cli argument.

tools/pcdl/pcdl_macros.xml
elmbeech: oh! this docstring is wrong.
this is a mix up from two docstings. that was happening when i was writing the macros.
thank you for pointing this out!

tools/pcdl/pcdl_macros.xml
elmbeech: i don't think this will work with a select command.
e.g. if you have 3 cell types, you can specify 3 different colors. if you have 4 celltypes you can give 4 specific colors and so on. if you don't give colors ("none"), then the colormap will be used to extract the colors from there. this is the default setting.

tools/pcdl/pcdl_macros.xml
elmbeech: yes and no.
if we give this as a cli argument, i think it will have to be a string.
i will look into it.

tools/pcdl/pcdl_macros.xml
elmbeech: ? 
i think this is the wrong line in the source code.

tools/pcdl/pcdl_macros.xml
elmbeech: will look into this.

tools/pcdl/pcdl_macros.xml
elmbeech: you mean `settingxml`?
well, this is a bit related to how PhysiCell out put is.
PhysiCell output is a folder full of files. There is a `PhysiCell_setting.xml` inside, which is the model specification.
`PhysiCell_setting.xml` is the default name, but if you like you can save it in the pcstudio under an other name.
Some researchers do.
How could I handle such a thing with a data parameter?
I don't know which name they will give.

tools/pcdl/pcdl_macros.xml
elmbeech: yes, can take this out.
this comes from the origainl cli docsting.

tools/pcdl/pcdl_macros.xml
elmbeech: i do this with Cheetan?

tools/pcdl/pcdl_macros.xml
elmbeech: will look into this.

tools/pcdl/pcdl_macros.xml
elmbeech: for the z_slice?
no, no min max.
it is just a z stack cut through an xyz domain to plot it in 2D.
maybe slice is not the best word for that?
i will look into this.

elmbeech:
could this `warning`, printed via `standard error`, cause troubles that the github workflow is passing? 
(pcdl_plot_contour and pcdl_plot_scatter have the same issue as this pcdl_make_gif issue.)

image

elmbeech:
this `pcdl_make_move` command which makes mp4 movies (with the use of the ffmpg library) seems to hang ...
not sure why.
image

elmbeech:
Note to my self: 
+ pcdl_plot_contour focus default has to be empty and not cell_type !!! 
+  pcdl_plot_scatter focus default cell_type is perfect !!!

elmbeech:
these are the only issues i could manually detect that could cause the two tests to fail.
still trouble to understand what i could lern from this workflow output.
+ https://github.com/galaxyproject/tools-iuc/actions/runs/16556891445/job/46819899269?pr=7034

here all tests are simply marked as failed.
i don't beleave that.

image

elmbeech:
i wonder if i simply should exclude `pcdl_make_gif` and `pcdl_make_movie`, since they built on linux libraries  (image magick and ffmpeg).

tools/pcdl/pcdl_macros.xml
bgruening: yes, the argument parameter, will implicitly create a `name` by stripping the leading `-` and converting middle `-` to `_`

tools/pcdl/pcdl_macros.xml
bgruening: Two hints, not sure if applicable. You can have multiple=true in a select param. And you can have a `text` param with `options`. So essentially provide options but let the user still overwrite with text.

In case of text you might want to restrict the user input to a few allowed characters etc ... no numbers...

tools/pcdl/pcdl_macros.xml
bgruening: `--figsize $x $y` should work no?

tools/pcdl/pcdl_macros.xml
bgruening: you can have a select with multiple=true

tools/pcdl/pcdl_macros.xml
bgruening: Not sure what you mean here. If you declare it as type="integer" Galaxy will make your users can not enter something else. On the CLI its of course a string like `--linewidth $linewidth`.

tools/pcdl/pcdl_macros.xml
bgruening: I just wanted to say that you can also add `min="0.01"` and max="10.0" to this parameter, whatever makes sense. You can also go negative, if your scalefactor allows that.

tools/pcdl/pcdl_macros.xml
bgruening: Oh ok, so a very advanced parameter, I see. thanks for explaining.

tools/pcdl/pcdl_macros.xml
bgruening: No, you can do that with the schema: https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-inputs-param-validator

Cheetah is executed during job process time. The XML stuff can be done in the UI before the job is handed off for processing.

bgruening:
You need to go to the CI summary page, this one: https://github.com/galaxyproject/tools-iuc/actions/runs/16556891445?pr=7034

Than you see:

image


you still have a lot of files missing:

> Collection element output00000000_cell_maxabs of collection anndata_h5ad: Test output file (output00000000_cell_maxabs.h5ad) is missing. If you are using planemo, try adding --update_test_data to generate it.

tools/pcdl/pcdl_get_anndata.xml
```diff
@@ -0,0 +1,123 @@
...
+                <param name="path">
+                    <expand macro="output"/>
+                </param>
+                <param name="collapse" value="false"/>
+            </section>
+            <section name="advanced">
+                <param name="verbose" value="false"/>
+            </section>
+            <output_collection name="anndata_h5ad" count="2">
+                <element name="output00000000_cell_maxabs" file="output00000000_cell_maxabs.h5ad" ftype="h5ad">
```
bgruening: @elmbeech important note, if you use asserts, you don't need the `file`. 

`file` is used to compare an expected file with a generated one. If you use asserts you don't have an expected file, you assert against the generated one. I guess this is while most of your tests are failing.

bgruening:
Please see my last comment, I think this explains your failing tests.

tools/pcdl/pcdl_get_anndata.xml
elmbeech: this might it be!
here i found it in the docu.
+ https://docs.galaxyproject.org/en/master/dev/schema.html#tool-tests-test

I somehow misunderstood the idea of the file parameter when i was writing the tests.

elmbeech:
ogge! now the test are passing :").
the rest is cosmetic. i will do it later this week or over the weekend or so.

tools/pcdl/pcdl_macros.xml
elmbeech: My professor , Paul Macklin,  said  z_slice is correct.
"Since 2D PhysiCell is actually a thin slice of 3D (just like an actual pathology sample, FWIW), there is a notion of z limits (typically z +/- 1/2 dz)"
In other words: It is an actual slice with wide dz  but we only specify the z coordinate for the the cut of the slice.

tools/pcdl/pcdl_macros.xml
elmbeech: ah i see. to set a range of possible values.
will look into it. 
this is just a plotting value which will influence the marker size. it is called "s" in matplotlib. 
if you give bad values, the plot will just look ugly, but i don't think anything will crash.
e.g., if you you give a negative value and matplotlib cannot handle it, matplotlib will through an error.

tools/pcdl/pcdl_macros.xml
elmbeech: but the order of the color is important too.
first color goes to the cell_type with ID 0 and so on.
in python you can alternatively give a dictionary. from the cli, you can only give a list.
i mean, the default options work well.
this is only if you wanna tweak your plot for a presentation or such.

tools/pcdl/pcdl_macros.xml
elmbeech: i think so!

tools/pcdl/pcdl_macros.xml
elmbeech: this are user-specified variables (called custom_data in PhysiCell).
they can have any name.
internally, in C++, they are handled as doubles, but in Python they can be anything (boolean, integer,  float, string) by default they are floats. this is kind of a typecast option for downstream analysis.
this e.g. matters if you want to pack your model output into an anndata object and have custom_data specified that is not really numerical but categorical. the very beginner modeler will not use this. this is advanced modeling.

tools/pcdl/pcdl_macros.xml
elmbeech: cool!

tools/pcdl/pcdl_get_cell_attribute.xml
elmbeech: i am sorry, I don't like this style.
i like the explicit true false style more.

tools/pcdl/pcdl_macros.xml
elmbeech: i looked into this.
the `argument="--` thing is perfect and i did it for all arguments!
the way you handle Booleans i don't like. i prefer my  expresive style. 
so i ended up not doing these changes for the Booleans  and i hope this is ok.

tools/pcdl/pcdl_macros.xml
elmbeech: so the `secondary_y`, i really have to look into this if this makes sense in Galaxy.

that `setting.xml` parameter is needed is a relic from old PhysiCell version. 
nowadays the default setting should be `None`, because we can extract this information from other output.
if it is none it will as well not break if user specifies and other name for the settings.xml file than the default PhysiCell_settings.xml. 
frankly, i do not expect someone to upload old PhysiCell output to galaxy to analyze it. 
i expect users to analyse the output they generate with the pcstudio galaxy tool, which of course works with a recent physicell version.
because of that, i will totally hide this parameter in galaxy and set if fix to None.

tools/pcdl/pcdl_macros.xml
elmbeech: but then `None` cannot be a valid input.
and `None` is the default, the most simple to use. 
giving two specific numbers is really for tweaking the plot output.

tools/pcdl/pcdl_macros.xml
elmbeech: i adjusted the help text slightly (mentioned the cut).

bgruening:
Please let me know when you think you are ready.

tools/pcdl/pcdl_macros.xml
elmbeech: that is more or less the help text  ...
ok

tools/pcdl/pcdl_macros.xml
elmbeech: this color maps are totally outdated.
state of the rate seqencial maps like viridis are not available.
even the link they provide is out dated.

tools/pcdl/pcdl_macros.xml
elmbeech: default in matplotlib 2.0 is actually 1.5. so float.

tools/pcdl/pcdl_macros.xml
elmbeech: agree and did so.

tools/pcdl/pcdl_macros.xml
elmbeech: maybe not really there. users should be able to type the title they want the whole UTF8 set with emojis and what ever. but validator are useful for when I only want none, space, and float values allowed.

tools/pcdl/pcdl_macros.xml
elmbeech: will use regex validator for this fields.

elmbeech:
@bgruening i think it is ready.

tools/pcdl/pcdl_macros.xml
bgruening: The website is not working for me. The warning at the top does not get me to another side. 
But that aside, my point was more like can you do something similar. Let pick users a colormap. Or do you assume people know them of the top of their head?

maybe https://github.com/galaxyproject/tools-iuc/blob/46fdfdaee17ad325380feb4bb54d4fd143a9ba4f/tools/scanpy/macros.xml#L293

or the seaborn one ontop?

bgruening:
Great, I will try to have a look asap again. This is a really nice set of tools, thanks a lot!

tools/pcdl/pcdl_macros.xml
elmbeech: I ended up updating the color map list and organized it in the way it is organized in the current matplotlib documentation.
+ https://matplotlib.org/stable/users/explain/colors/colormaps.html

i have two questions:
1. is there a way to collapse the radio button list, because it is very long and dominant now?
1. is there a way to have separate block in the radio button list that are currently simply comments in the macro (namely: perceptually uniform sequential, sequential, sequential 2, diverging, cyclic, qualitative, miscellaneous)?

elmbeech:
thank you for the appreciation! 
honestly, i did not expect that it would be that much work and take that much time to write a wrapper for a command line script.
to be fair, i have to say i am not too familiar with all this web framework stuff and had only little knowledge about galaxy dateing some years back. 
in any case, happy when all this work ends up in a final and hopefully useful product. 
that Randy got the pcstudio in galaxy running (as i understood with help from you too) was very impressive!

elmbeech:
thank you for all your help with the PR!

tools/pcdl/pcdl_macros.xml
bgruening: With radio we can not do that, I recommend to use `select`. People can than browse, serach etc ...

bgruening:
>  honestly, i did not expect that it would be that much work and take that much time to write a wrapper for a command line script.

You picked the IUC battle, rigorous reviews and annoying dudes that care about UX :)
Your integration was working in the first days, you could deploy that. On IUC we would like to go a few steps ahead and make those integrations readable, usable and maintainable.
I hope it was worth your time.

If you like use `select` for the colormap and then we are ready to go I think.

elmbeech:
Thank you, Bjoern!
As I undertook this task, I wanted to do something proper so it was worth the time. And I got help and advice. 
I changed to select and re-formated with vscode for a last time. 
Now all should be ready.

bgruening (approved):

bgruening:
Step 3 of https://usegalaxy-eu.github.io/posts/2020/08/22/three-steps-to-galaxify-your-tool/ if you like :)

bgruening:
Merci!!!

elmbeech:
hi @bgruening ,

I hope you still see this message.

Somehow, naively, I thought when I merge my tools into the tools-iuc repo, they will become avaliable in at https://galaxy.eu and https://usegalaxy.org/ . But when I am searching for them in the "search tools" bar or in the panel below, I cannot find them. 
I definitely can find them in the main toolshed (https://toolshed.g2.bx.psu.edu/), but that's it.

Could you please point me to the documentation where I can learn how to use the pcdl tools online at galaxy eu  or usa? Or are these tools only for use on a local installation?

thank you, Elmar

bgruening:
See my last message, step 3 ;-)

Your tool is now in our App store and can be installed by all admins. However, admins are able to decide what goes in.

elmbeech:
Thank you Bjoern, for that swift reply! 
Totally missed this. I will give it a try in Europe and the US.
Have a good rest of the weekend. Best, Elmar

elmbeech:
i think first just Europe. 
let's see what happens.

PR #7046, opened 2025-06-12
New tool Addition: Chopper a tool for fast processing (filtering, trimming, ..) of long reads

lsterck:
FOR CONTRIBUTOR:

Hi,
this is a pull request to add the Chopper tool to the iuc toolshed.
I'm happy to amend any review comments.

best regards,
lieven


tools/chopper/chopper.xml
```diff
@@ -0,0 +1,150 @@
+<tool id="chopper" name="Chopper" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" python_template_version="3.5" profile="23.2">
```
SaimMomin12:
```suggestion
<tool id="chopper" name="Chopper" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="23.2">
```

tools/chopper/chopper.xml
```diff
@@ -0,0 +1,150 @@
...
+	<description>Filtering and trimming of fastq files (long reads).</description>
```
SaimMomin12:
```suggestion
	<description>Filtering and trimming of long reads</description>
```

tools/chopper/chopper.xml
```diff
@@ -0,0 +1,150 @@
...
+ #if $output_params.gzip == "no"
+	> $fq_filt 
+ #else
+	| gzip > $fq_filt_gz
+ #end if
+
+ ]]></command>
+	<inputs>
+		<param argument="--input" type="data" label="FASTQ file to check" format="fastq,fastq.gz,fastqsanger.gz,fastqsanger" />
+		<param argument="--contam" type="data" format="fasta" optional="True"  size="30" label="Reference FASTA" help="FASTA file with reference to check potential contaminants against."/>
```
SaimMomin12:
```suggestion
		<param argument="--contam" type="data" format="fasta" optional="True" label="Reference FASTA" help="FASTA file with reference to check potential contaminants against."/>	
```
afaik `size` is deprecated.

tools/chopper/chopper.xml
```diff
@@ -0,0 +1,150 @@
...
+	| gzip > $fq_filt_gz
+ #end if
+
+ ]]></command>
+	<inputs>
+		<param argument="--input" type="data" label="FASTQ file to check" format="fastq,fastq.gz,fastqsanger.gz,fastqsanger" />
+		<param argument="--contam" type="data" format="fasta" optional="True"  size="30" label="Reference FASTA" help="FASTA file with reference to check potential contaminants against."/>	
+
+		<section name="option_params" title="Optional Parameters" expanded="True">
+			<param argument="--quality" type="integer" label="Minimal quality score" value="0" min="0" max="60" help="Sets a minimum Phred average quality score [default: 0]" />
```
SaimMomin12:
```suggestion
			<param argument="--quality" type="integer" label="Minimal quality score" value="0" min="0" max="60" help="Sets a minimum Phred average quality score"/>
```

tools/chopper/chopper.xml
```diff
@@ -0,0 +1,150 @@
...
+			<param argument="--maxqual" type="integer" label="Maximal quality score" value="60" min="0" max="60" help="Sets a maximum Phred average quality score [default: 60]" />
```
SaimMomin12:
```suggestion
			<param argument="--maxqual" type="integer" label="Maximal quality score" value="60" min="0" max="60" help="Sets a maximum Phred average quality score/>
```

tools/chopper/chopper.xml
```diff
@@ -0,0 +1,150 @@
...
+			<param argument="--minlength" type="integer" label="Sets a minimum read length" value="1" min="1"  help="Minimal length of read to keep [default: 1]"  optional="True"/>
```
SaimMomin12:
```suggestion
			<param argument="--minlength" type="integer" label="Sets a minimum read length" value="1" min="1"  help="Minimal length of read to keep"  optional="True"/>
```

tools/chopper/chopper.xml
```diff
@@ -0,0 +1,150 @@
...
+			<param argument="--maxlength" type="integer" label="Sets a maximum read length" help="Maximal length of read to keep" optional="True"/>
+			<param argument="--headcrop" type="integer" label="Headcrop" value="0" min="0" help="Trim N nucleotides from the start of a read." optional="True"/>
+			<param argument="--tailcrop" type="integer" label="Tailcrop" value="0" min="0" help="Trim N nucleotides from the end of a read." optional="True"/>
+			<param argument="--mingc" type="float" label="Minimum GC content" value="0.0" min="0.0" max="1.0" help="Sets a minimum GC content for reads to keep [default: 0.0]" optional="True"/>
```
SaimMomin12:
```suggestion
			<param argument="--mingc" type="float" label="Minimum GC content" value="0.0" min="0.0" max="1.0" help="Sets a minimum GC content for reads to keep" optional="True"/>
```

tools/chopper/chopper.xml
```diff
@@ -0,0 +1,150 @@
...
+			<param argument="--maxgc" type="float" label="Maximum GC content" value="1.0" min="0.0" max="1.0" help="Sets a maximum GC content for reads to keep [default: 1.0]" optional="True"/>
```
SaimMomin12:
```suggestion
			<param argument="--maxgc" type="float" label="Maximum GC content" value="1.0" min="0.0" max="1.0" help="Sets a maximum GC content for reads to keep" optional="True"/>
```

tools/chopper/chopper.xml
```diff
@@ -0,0 +1,150 @@
...
+		<test expect_num_outputs="1">
+			<param name="input" value="other-test.fastq"/>
+			<section name="output_params">
+				<param name="gzip" value="false"/>
+			</section>
+			<output name="fq_filt">
+				<assert_contents>
+					<has_text_matching expression="^@"/>
+					<has_text_matching expression="\n\+\n"/>
+				</assert_contents>
```
SaimMomin12: Can we have asserts based on filtered reads contents?

SaimMomin12 (commented):
Thanks @lsterck for the addition, some review comments inline.

tools/chopper/chopper.xml
lsterck: Sorry, don't fully get this remark ...
Do you want to somehow test the removed reads (as well?). My initial thought was to simply check if it is valid fastq in the output file, or at least something 'easy' and as close as possible

SaimMomin12: Yes, it would be nice to have some asserts based on specific contents of the file. Currently the asserts are quite general.

lsterck: like checking if a certain ID is still present in the output file?
or add the expected output file to the test-folder?

SaimMomin12: Both are okay

lsterck:
right, I think I know what the problem is ... on it.

lsterck:
OK,
I'm stuck now ....
I thought it was (and it still is apparently) the comparison of the gzipped output files, but that still seems to fail once in a while. The only thing I can imagine is that the stream that gzip receives is not deterministic and thus the gzip version is different (though content is the same but likely different order of reads in the output)

I've ran the local planemo test a couple of times and after the edit I pushed before it seems to consistently pass (which was indeed not the case before) but not here apparently. 

What am I missing? Any ideas how to resolve this?

lsterck:
I could remove the gzip test but I would prefer to keep it if in any way possible (just to also check the gzip option)

SaimMomin12:
>I could remove the gzip test but I would prefer to keep it if in any way possible (just to also check the gzip option)

Maybe you can have try using asserts as per here: https://github.com/galaxyproject/tools-iuc/blob/eca92f6ec1b61f1a772b308cfcb1c7b3b7b22ea5/tools/pbtk/bam2fastx.xml#L46

tools/chopper/chopper.xml
```diff
@@ -0,0 +1,152 @@
...
+					<has_text text="@35febf09-dcbc-424c-987e-9f3f80fe73a5"/>
+					<has_text text="@3fda06e9-62ef-4448-9993-b90124a793d5"/>
+					<has_text text="@19d9337f-4fb6-46e5-b484-14d05f562506"/>
+				</assert_contents>
+			</output>
+		</test>
+		<test expect_num_outputs="1">
+			<param name="input" value="other-test.fastq"/>
+			<param name="contam" value="random_contam.fa"/>
+			<output name="fq_filt" ftype="fastq.gz" decompress="true" value="out2.fq.gz"/>
```
SaimMomin12:
```suggestion
			<output name="fq_filt" ftype="fastq.gz" decompress="true" file="out2.fq.gz"/>
```

lsterck: ah, well spotted ... 
would that make a difference for the testing?

SaimMomin12: Using asserts will specifically check for your assert condition rather than entire file. 

Other thing you can also try using `--update_test_data` if your gzip test passes locally.

lsterck: if I use the assert approach, it makes no sense to include the zip file in the test-data dir, correct?

bgruening: Correct :-)

lsterck: I'll go the assert route as it turns out the order of reads in the output is not deterministic, thus the gzip will differ randomly.

SaimMomin12:
Thanks @lsterck!

lsterck:
🙏 thanks @SaimMomin12

Issue #7043, opened 2025-06-09
Different scRNA workflow failing due to running out of memory


scottcain:
**Describe the bug**
When testing BRC-Analytics.org functionality and trying to run the "Single-Cell RNA-seq Preprocessing: 10X Genomics CellPlex Multiplexed Samples" workflow (https://iwc.galaxyproject.org/workflow/fastq-to-matrix-10x-scrna-seq-fastq-to-matrix-10x-cellplex/) it failed on 7th step of the linked history, "RNA STARSolo on data 5, data 4, and others" with an error that indicated that it ran out of memory and suggested providing a flag to increase the amount of memory used to sort the bam file:

```
EXITING because of fatal ERROR: not enough memory for BAM sorting: 
SOLUTION: re-run STAR with at least --limitBAMsortRAM 29816090916
Jun 06 17:15:02 ...... FATAL ERROR, exiting
```

However, there isn't an option in the UI to provide this flag, so there is no way to run this step in Galaxy, regardless of the workflow.

My suggestion (which echos @mvdbeek 's) is to allow this step to not sort the BAM but rather sort it in a subsequent step with samtools.

**Galaxy Version and/or server at which you observed the bug**
Galaxy Version: `{"version_major":"25.0","version_minor":"rc1"}`

**Browser and Operating System**
Operating System: macOS 15.5
Browser: Safari

**To Reproduce**
In this history: https://usegalaxy.org/u/scottcain/h/t-brucie-cell-cycle
See step 7.

bernt-matthias:
We need to improve stdio handling and match the OOM message https://github.com/galaxyproject/tools-iuc/blob/a9d3251b40bc6ae1147b2e013d623f957a981421/tools/rgrnastar/macros.xml#L191

mvdbeek:
it's not OOM, the message is misleading

mvdbeek:
https://github.com/alexdobin/STAR/issues/870 is relevant there.

scottcain:
Would it be possible (and quick to implement) exposing the `--outSAMtype` option (which I don't think is available yet) so that the option `BAM Unsorted` to be provided and then the BAM file could be sorted by samtools in a subsequent step? I think that would make the workflow that initiated this bug report more resilient.

pavanvidem:
There are also 2 more questions on GalaxyHelp. Can someone please review the above PR?

scottcain:
[...]

PR #7051, opened 2025-06-13
STARSolo add --outBAMsortingBinsN an add a param to force samtools sort

pavanvidem:
Quick fix to https://github.com/galaxyproject/tools-iuc/issues/7043

- This PR does NOT export `--outSAMtype`. This patch should still keep the existing functionality of the tool. It adds an extra boolean param `use_samtools_sort` which forces `--outSAMtype BAM Unsorted` and uses `samtools sort` to sort the BAM file.
- It also exposes `--outBAMsortingBinsN` and `--winAnchorMultimapNmax` params which already exist in STAR.

### DIFF
--- a/tools/rgrnastar/macros.xml
+++ b/tools/rgrnastar/macros.xml
@@ -5,7 +5,7 @@
     the index versions in sync, but you should manually update @IDX_VERSION_SUFFIX@ -->
     <!-- STAR version to be used -->
     <token name="@TOOL_VERSION@">2.7.11b</token>
-    <token name="@VERSION_SUFFIX@">0</token>
+    <token name="@VERSION_SUFFIX@">1</token>
     <token name="@PROFILE@">21.01</token>
     <!-- STAR index version compatible with this version of STAR
     This is the STAR version that introduced the index structure expected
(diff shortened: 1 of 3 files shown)


mvdbeek:
> * It also exposes `--outBAMsortingBinsN` and `--winAnchorMultimapNmax` params which already exist in STAR.

I don't think `--outBAMsortingBinsN` should be a user option, can you calculate reasonable value for that ?

tools/rgrnastar/macros.xml
```diff
@@ -584,4 +584,30 @@
             #end if
     ]]></token>
 
+    <xml name="perf_params">
```
mvdbeek, line 587: There's no way user will set reasonable values here. I would suggest calculating reasonable values for these.

pavanvidem: I don't think we have enough data to estimate them. Those params were already in use in STAR tool. Most users do not use/set them.

bernt-matthias: Lets remove the parameters? Maybe also from STAR?

pavanvidem: Then how should we solve the original [issue](https://github.com/galaxyproject/tools-iuc/issues/7043)? 

The only option I see is to set `--outSAMtype BAM Unsorted` by default and use `samtools sort` if the user requests a sorted output. 

But there is one issue with a STARSolo param: If the user requires `UB` as one of the output tags, then we are forced to use `--outSAMtype BAM SortedByCoordinate`. See [here](https://github.com/alexdobin/STAR/blob/master/docs/STARsolo.md#bam-tags). In that case, the user might need to set `--outBAMsortingBinsN` param.

PR #7078, opened 2025-06-26
Update snapatac2 to version 2.8.0

nilchia:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/snapatac2/dimension_reduction_clustering.xml
+++ b/tools/snapatac2/dimension_reduction_clustering.xml
@@ -324,8 +324,7 @@ sa.tl.aggregate_cells(
                 <param name="adata" location="https://zenodo.org/records/11260316/files/tl.umap.pbmc_500_chr21.h5ad"/>
                 <param name="n_neighbors" value="50"/>
                 <param name="use_rep" value="X_spectral"/>
-                <param name="method_" value="kdtree"/>
-                <param name="inplace" value="True"/>
+                <param name="algorithm" value="kdtree"/>
                 <param name="random_state" value="0"/>
             </conditional>
             <section name="advanced_common">
--- a/tools/snapatac2/macros.xml
+++ b/tools/snapatac2/macros.xml
@@ -1,7 +1,7 @@
 <macros>
-    <token name="@TOOL_VERSION@">2.6.4</token>
-    <token name="@VERSION_SUFFIX@">1</token>
-    <token name="@PROFILE@">23.0</token>
+    <token name="@TOOL_VERSION@">2.8.0</token>
+    <token name="@VERSION_SUFFIX@">0</token>
+    <token name="@PROFILE@">24.0</token>
     <xml name="xrefs">
         <xrefs>
             <xref type="bio.tools">snapatac</xref>
--- a/tools/snapatac2/.lint_skip
+++ b/tools/snapatac2/.lint_skip
@@ -1,2 +1,0 @@
-TestsCaseValidation
-TestsParamInInputs


bgruening:
> https://kzhang.org/SnapATAC2/api/_autosummary/snapatac2.pp.import_data.html

does not exist anymore

SaimMomin12:
@nilchia All the older links of https://kzhang.org/SnapATAC2/api/ have been migrated to https://scverse.org/SnapATAC2/api/, we need to change them accordingly. This will solve most of the linting errors.

nilchia:
>RuntimeError: Some worker process has died unexpectedly.

Is it a timeout error, or was the memory not sufficient?

tools/snapatac2/peaks_and_motif_analysis.xml
```diff
@@ -5,159 +5,190 @@
...
-        blacklist = '$method.blacklist',
+        #if $method_condi.blacklist
+        blacklist = '$method_condi.blacklist',
         #end if
-        key_added = '$method.key_added',
+        key_added = '$method_condi.key_added',
         inplace = True,
         tempdir = '.',
-        n_jobs = 1
+        n_jobs = int(os.getenv("GALAXY_SLOTS", 4))
```
bernt-matthias: The error: 

```
  File "h5py/_objects.pyx", line 56, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 57, in h5py._objects.with_phil.wrapper
  File "h5py/h5f.pyx", line 122, in h5py.h5f.create
BlockingIOError: [Errno 11] Unable to synchronously create file (unable to lock file, errno = 11, error message = 'Resource temporarily unavailable')
```

indicates that the tool has a problem with parallel processing. Maybe leave at 1 and open an upstream issue?

nilchia: The default is 8, and I could run it locally :thinking:

bernt-matthias: Have you tried to test with `--biocontainers --no_dependency_resolution --no_conda_auto_init`?

nilchia: This is my script, which I use:
```
#!/bin/sh
time planemo test $1 \
		--galaxy_root /home/nilchia/Documents/github/galaxyproject/galaxy/ \
		--biocontainers --no_dependency_resolution --no_conda_auto_init \
		--no_cleanup \
		--update_test_data
```

nilchia: @pavanvidem is this the reason you set it to 1 initially?

pavanvidem: Probably, it is default. Otherwise, it might  be the reason :)

nilchia: True, there is an issue with this function on HPC. Look at this [issue](https://github.com/scverse/SnapATAC2/issues/298).
We already used `if __name__ == '__main__':` but apparently it is not working.

tools/snapatac2/dimension_reduction_clustering.xml
```diff
@@ -5,161 +5,212 @@
     </macros>
     <expand macro="xrefs"/>
     <requirements>
-        <expand macro="requirements"/>
+        <expand macro="requirements">
```
bernt-matthias: My personal preference is to have one set of requirements for a tool suite, i.e. just use the union of all requirements of all tools in all tools.

Advantage is that this is then one container. Otherwise each set of requirements gets its own container (/conda env)

tools/snapatac2/dimension_reduction_clustering.xml
```diff
@@ -5,161 +5,212 @@
...
+            <requirement type="package" version="0.8.37">hdbscan</requirement>
+            <requirement type="package" version="0.10.2">leidenalg</requirement>
+            <requirement type="package" version="0.5.7">umap-learn</requirement>
```
bernt-matthias: Wondering if (some of) these should be covered by the bioconda recipe?

tools/snapatac2/dimension_reduction_clustering.xml
```diff
@@ -5,161 +5,212 @@
...
 )
 #end if
 
-@CMD_anndata_write_outputs@
+@CONF_ANNDATA_WRITE_OUTPUTS@
     ]]></configfile>
     </configfiles>
     <inputs>
-        <conditional name="method">
+        <conditional name="method_condi">
```
bernt-matthias: Why rename?

tools/snapatac2/dimension_reduction_clustering.xml
```diff
@@ -5,161 +5,212 @@
...
+                <option value="tl.multi_spectral">similar to 'tl.multi_spectral' but it can work on multiple modalities.</option>
                 <option value="tl.umap">Compute Umap, using 'tl.umap'</option>
                 <option value="pp.knn">Compute a neighborhood graph of observations, using 'pp.knn'</option>
                 <option value="tl.leiden">Cluster cells into subgroups, using 'tl.leiden'</option>
                 <option value="tl.kmeans">Cluster cells into subgroups using the K-means algorithm, using 'tl.kmeans'</option>
                 <option value="tl.dbscan">Cluster cells into subgroups using the DBSCAN algorithm, using 'tl.dbscan'</option>
                 <option value="tl.hdbscan">Cluster cells into subgroups using the HDBSCAN algorithm, using 'tl.hdbscan'</option>
-                <option value="tl.aggregate_X">Aggregate values in adata.X in a row-wise fashion, using 'tl.aggregate_X'</option>
+                <!-- It is implemented in select_feature function, and it is problematic if the user don't select a groupby  (will return an array). i think this can be skipped unless needed -->
```
bernt-matthias: I do not understand what this means?

tools/snapatac2/dimension_reduction_clustering.xml
```diff
@@ -5,161 +5,212 @@
...
+                <!-- <option value="tl.aggregate_X">Aggregate values in adata.X in a row-wise fashion, using 'tl.aggregate_X'</option> -->
                 <option value="tl.aggregate_cells">Aggregate cells into pseudo-cells, using 'tl.aggregate_cells'</option>
             </param>
             <when value="tl.spectral">
-                <expand macro="inputs_anndata"/>
-                <expand macro="param_n_comps"/>
+                <expand macro="param_inputs_anndata"/>
+                <param argument="n_comps" type="integer" value="30" label="Number of dimensions to keep" help="The result is insensitive to this parameter when `weighted_by_sd` is set, as long as it is large enough, e.g. 30."/>
```
bernt-matthias: Why not keep the macro?

tools/snapatac2/dimension_reduction_clustering.xml
```diff
@@ -5,161 +5,212 @@
...
                 <option value="tl.hdbscan">Cluster cells into subgroups using the HDBSCAN algorithm, using 'tl.hdbscan'</option>
-                <option value="tl.aggregate_X">Aggregate values in adata.X in a row-wise fashion, using 'tl.aggregate_X'</option>
+                <!-- It is implemented in select_feature function, and it is problematic if the user don't select a groupby  (will return an array). i think this can be skipped unless needed -->
+                <!-- <option value="tl.aggregate_X">Aggregate values in adata.X in a row-wise fashion, using 'tl.aggregate_X'</option> -->
                 <option value="tl.aggregate_cells">Aggregate cells into pseudo-cells, using 'tl.aggregate_cells'</option>
             </param>
             <when value="tl.spectral">
-                <expand macro="inputs_anndata"/>
-                <expand macro="param_n_comps"/>
+                <expand macro="param_inputs_anndata"/>
```
bernt-matthias, line 214: Why rename the macro?

tools/snapatac2/dimension_reduction_clustering.xml
```diff
@@ -170,8 +221,18 @@ sa.tl.aggregate_cells(
                 </param>
                 <param argument="weighted_by_sd" type="boolean" truevalue="True" falsevalue="False" checked="true" label="Whether to weight the result eigenvectors by the square root of eigenvalues"/>
             </when>
+            <when value="tl.multi_spectral">
+                <param name="adata_rna" type="data" format="h5ad" label="Annotated data matrix containing RNA data"/>
+                <param name="adata_atac" type="data" format="h5ad" label="Annotated data matrix containing ATAC data"/>
```
bernt-matthias: Could these parameters reuse the `inputs_anndata` macro?

tools/snapatac2/dimension_reduction_clustering.xml
```diff
@@ -479,31 +595,39 @@ sa.tl.aggregate_cells(
...
+            <output name="anndata_out" ftype="h5ad">
+                <assert_contents>
+                    <has_h5_keys keys="obs/hdbscan"/>
+                </assert_contents>
+            </output>
         </test>
-        <test expect_num_outputs="2">
-            <!-- tl.aggregate_X -->
-            <conditional name="method">
+        <!-- <test expect_num_outputs="2">
```
bernt-matthias, line 604: Why is this commented?

tools/snapatac2/macros.xml
```diff
@@ -1,113 +1,111 @@
...
-        <param argument="shift_right" type="integer" value="-5" label="Insertion site correction for the right end" help="Note this has no effect on single-end reads"/>
+    <xml name="param_shift" tokens="varname" token_value="0" token_label="Insertion site correction for the left end">
+        <param argument="@VARNAME@" type="integer" value="@VALUE@" label="@LABEL@" help="Note this has no effect on single-end reads"/>
     </xml>
     <xml name="param_chunk_size" tokens="size">
         <param argument="chunk_size" type="integer" value="@SIZE@" label="chunk size"/>
     </xml>
-    <xml name="min_max_frag_size">
+    <xml name="param_min_max_frag_size">
         <param argument="min_frag_size" type="integer" optional="true" value="" label="Minimum fragment size to include"/>
```
bernt-matthias, line 156: For numerical parameters min/max would be good to have if possible.

bernt-matthias (commented):
Some first comments. Would it be possible to keep names of conditionals, parameters and macros?

- Changing names of conditionals and parameters breaks workflows
- Changing names of macros makes reviewing harder.

tools/snapatac2/dimension_reduction_clustering.xml
nilchia: True, I separated them to have the minimum requirement for each tool.
I can also make them one. No problem :)

tools/snapatac2/dimension_reduction_clustering.xml
nilchia: No, they are not in the [recipe](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/snapatac2/meta.yaml)

tools/snapatac2/dimension_reduction_clustering.xml
nilchia: The filtering in the output could not recognize this conditional from the method param.

tools/snapatac2/dimension_reduction_clustering.xml
nilchia: I meant the `select_feature` in preprocessing.xml already uses this function.
[Here is the upstream code](https://github.com/scverse/SnapATAC2/blob/98180f3f9ce3e625da5ed942f57e41ef01c61b3a/snapatac2-python/python/snapatac2/preprocessing/_basic.py#L1071).

tools/snapatac2/dimension_reduction_clustering.xml
nilchia: Just to look nicer :)

tools/snapatac2/dimension_reduction_clustering.xml
nilchia: I don't recall the reason. Probably no reason :)
I'll put it back

tools/snapatac2/dimension_reduction_clustering.xml
nilchia: They should be separate, as they are different anndatas.

tools/snapatac2/dimension_reduction_clustering.xml
nilchia: The function is also commented out.
I just wanted to keep the template there.

tools/snapatac2/macros.xml
nilchia: In the tool document, it is not indicated. Not sure what to put as min/max.
@pavanvidem, any idea?

tools/snapatac2/dimension_reduction_clustering.xml
bernt-matthias: Then there is likely another problem. I would suggest to revert.

tools/snapatac2/dimension_reduction_clustering.xml
bernt-matthias: Could a user still want to use this? Otherwise refer also to the preprocessing tool.

nilchia:
The metrics tool is failing because of the metadata:
```
galaxy.jobs.runners ERROR 2025-10-09 10:24:16,166 [pN:main.1,p:432374,tN:LocalRunner.work_thread-1] (9/434145) Job wrapper finish method failed
  File "/home/nilchia/Documents/github/galaxyproject/galaxy/lib/galaxy/jobs/runners/__init__.py", line 680, in _finish_or_resubmit_job
    job_wrapper.finish(
  File "/home/nilchia/Documents/github/galaxyproject/galaxy/lib/galaxy/jobs/__init__.py", line 2189, in finish
    self._finish_dataset(output_name, dataset, job, context, final_job_state, remote_metadata_directory)
  File "/home/nilchia/Documents/github/galaxyproject/galaxy/lib/galaxy/jobs/__init__.py", line 1999, in _finish_dataset
    dataset.datatype.set_meta(dataset, overwrite=False)
  File "/home/nilchia/Documents/github/galaxyproject/galaxy/lib/galaxy/datatypes/binary.py", line 1674, in set_meta
    dataset.metadata.shape = tuple(anndata_file["X"].shape)
TypeError: 'NoneType' object is not iterable
```

I don't think it is related to the change in the requirements. or is it?

nilchia:
The rest is passing locally :thinking:

bgruening:
Please try to run with `planemo test --biocontainers .`

bernt-matthias:
> The metrics tool is failing because of the metadata:
> 
> ```
> galaxy.jobs.runners ERROR 2025-10-09 10:24:16,166 [pN:main.1,p:432374,tN:LocalRunner.work_thread-1] (9/434145) Job wrapper finish method failed
> Traceback (most recent call last):
>   File "/home/nilchia/Documents/github/galaxyproject/galaxy/lib/galaxy/jobs/runners/__init__.py", line 680, in _finish_or_resubmit_job
>     job_wrapper.finish(
>   File "/home/nilchia/Documents/github/galaxyproject/galaxy/lib/galaxy/jobs/__init__.py", line 2189, in finish
>     self._finish_dataset(output_name, dataset, job, context, final_job_state, remote_metadata_directory)
>   File "/home/nilchia/Documents/github/galaxyproject/galaxy/lib/galaxy/jobs/__init__.py", line 1999, in _finish_dataset
>     dataset.datatype.set_meta(dataset, overwrite=False)
>   File "/home/nilchia/Documents/github/galaxyproject/galaxy/lib/galaxy/datatypes/binary.py", line 1674, in set_meta
>     dataset.metadata.shape = tuple(anndata_file["X"].shape)
> TypeError: 'NoneType' object is not iterable
> ```
> 
> I don't think it is related to the change in the requirements. or is it?

No. 

This is probably fixed in 25.1: https://github.com/galaxyproject/galaxy/pull/20778 

If you want this right away I would suggest to backport the fix to an  older release. If you want to wait for 25.1 we also should update the profile of the tool.

nilchia:
@bgruening 
I always run the tests with this:
```
#!/bin/sh
time planemo test $1 \
		--galaxy_root /home/nilchia/Documents/github/galaxyproject/galaxy/ \
		--biocontainers \
		--no_cleanup
```

I removed all the containers and reran it. They passed (for peaks_and_motif_analysis.xml):
```
galaxy.jobs DEBUG 2025-10-11 11:45:17,782 [pN:main.1,p:707056,tN:LocalRunner.work_thread-3] job_wrapper.finish for job 17 executed (33.204 ms)
Shut down
supervisord has terminated
All 7 test(s) executed passed.
snapatac2_peaks_and_motif (Test #1): passed
snapatac2_peaks_and_motif (Test #2): passed
snapatac2_peaks_and_motif (Test #3): passed
snapatac2_peaks_and_motif (Test #4): passed
snapatac2_peaks_and_motif (Test #5): passed
snapatac2_peaks_and_motif (Test #6): passed
snapatac2_peaks_and_motif (Test #7): passed
20.98user 2.02system 17:25.17elapsed 2%CPU (0avgtext+0avgdata 277392maxresident)k
71528inputs+2392outputs (190major+297882minor)pagefaults 0swaps
```

bernt-matthias:
On which branch is `/home/nilchia/Documents/github/galaxyproject/galaxy/`?

nilchia:
@bernt-matthias 
 release_25.0

nilchia:
I can't understand why it passed now :shaking_face: 
@bgruening @bernt-matthias

bernt-matthias:
Can you `git revert 924a4ff`? I would like to see it again and maybe test locally myself.

nilchia:
now only the metrics_xml is failing for the metadata error

bernt-matthias:
In order to understand: can you change https://github.com/galaxyproject/tools-iuc/blob/1775d01920f27931a1935e233766f615d753502b/.github/workflows/pr.yaml#L18 to `release_25.1`

nilchia:
Cool, it passes with 25.1

nilchia:
Can I keep 25.1 change or should I revert that?

bernt-matthias:
Please revert. We just need to backport https://github.com/galaxyproject/galaxy/pull/20778 to the release that you want to use as profile for these tools.

tools/snapatac2/dimension_reduction_clustering.xml
nilchia: reverted here: 47c59166e

tools/snapatac2/dimension_reduction_clustering.xml
nilchia: No I don't think so. It is also not included in their tutorials.

nilchia: >Otherwise refer also to the preprocessing tool.

sorry, I didn't understand what to do in preprocessing.

tools/snapatac2/dimension_reduction_clustering.xml
nilchia: macro reverted here: [93c4ef9](https://github.com/galaxyproject/tools-iuc/pull/7078/commits/93c4ef986a97344f141b49dec35850718ba6c3b6)

tools/snapatac2/dimension_reduction_clustering.xml
bernt-matthias: You should mention in the comment that this is already done in the preprocessing tool (if this is the case) ... since I could not find any reference to `select_feature`  in the tools .. but maybe I overlooked it.

tools/snapatac2/dimension_reduction_clustering.xml
```diff
@@ -39,8 +61,31 @@ sa.tl.spectral(
...
+    features = features_mask,
+    #else
+    features = None,
+    #end if
+    weights = None,  # Will enable if requested by users
+    random_state = $method.random_state,
+    weighted_by_sd = $method.weighted_by_sd,
+)
+
+adata = atac.copy()
```
pavanvidem: if in the end `atac` is copied to `adata`, you don't need a separate `atac` anndata object. You can use the existing macros to read atac object to `adata`. You need only one additional anndata object called `adata_rna` when the `multi_spectral` function is used.

tools/snapatac2/dimension_reduction_clustering.xml
```diff
@@ -39,8 +61,31 @@ sa.tl.spectral(
...
+adata.uns['spectral_eigenvalue_joint'] = embedding[0]
```
pavanvidem, line 90: does `embedding[0]` contain rna embeddings? why do you need to store it?

tools/snapatac2/dimension_reduction_clustering.xml
nilchia: Sorry, you were right.
I'll fix that

pavanvidem: instead, use `inputs_anndata` to read atac anndata as standard and in this case, read an additional rna anndata.

tools/snapatac2/dimension_reduction_clustering.xml
```diff
@@ -170,9 +217,19 @@ sa.tl.aggregate_cells(
                 </param>
                 <param argument="weighted_by_sd" type="boolean" truevalue="True" falsevalue="False" checked="true" label="Whether to weight the result eigenvectors by the square root of eigenvalues"/>
             </when>
+            <when value="tl.multi_spectral">
+                <param name="adata_rna" type="data" format="h5ad" label="Annotated data matrix containing RNA data"/>
+                <param name="adata_atac" type="data" format="h5ad" label="Annotated data matrix containing ATAC data"/>
+                <!-- Will enable if requested by users -->
+                <!-- <param name="weights" type="data" format="tabular" optional="true" label="Weights" help="Text file indicating weights for each modality. Each line contains only floats.If not provided, all modalities are weighted equally" /> -->
+                <expand macro="param_n_comps"/>
+                <param argument="features" type="data" format="txt" optional="true" label="Text file indicating features to keep. Each line contains only word (True/False)." help="True means that the feature is kept. False means the feature is removed"/>
```
pavanvidem: Make the format `txt,tabular`. Usually, this file is generated by annotating one of the columns in `anndata.var`, which is a tabular file.

tools/snapatac2/dimension_reduction_clustering.xml
```diff
@@ -238,40 +296,37 @@ sa.tl.aggregate_cells(
...
                 <expand macro="param_random_state"/>
                 <expand macro="param_key_added" key_added="pseudo_cell"/>
             </when>
         </conditional>
-        <expand macro="inputs_common_advanced"/>
+        <expand macro="param_common_advanced"/>
     </inputs>
     <outputs>
-        <data name="anndata_out" format="h5ad" from_work_dir="anndata.h5ad" label="${tool.name} (${method.method}) on ${on_string}: Annotated data matrix"/>
+        <data name="anndata_out" format="h5ad" from_work_dir="anndata.h5ad.gz" label="${tool.name} (${method.method}) on ${on_string}: Annotated data matrix"/>
```
pavanvidem, line 324: good!

tools/snapatac2/macros.xml
pavanvidem: min can be `1` or a biologically relevant min is maybe around 50. But I would still leave it empty to strictly follow the tool defaults.

tools/snapatac2/metrics.xml
```diff
@@ -0,0 +1,285 @@
...
+                <expand macro="param_inputs_anndata"/>
+                <expand macro="param_chrom_sizes" varname="gene_anno" label="GTF/GFF file containing the gene annotation"/>
+                <param argument="exclude_chroms" type="text" value="chrM, M" optional="true" label="A list of chromosomes to exclude">
+                    <expand macro="sanitize_query"/>
+                </param>
+            </when>
+            <when value="metrics.frip">
+                <expand macro="param_inputs_anndata"/>
+                <repeat name="regions_repeat" title="Regions" help="Creates a dictionary of regions to compute FRiP for.">
+                    <param argument="region_bed" type="data" format="bed" label="Regions to compute FRiP for"/>
```
pavanvidem: switch the order of `region_bed` and `region_name`

tools/snapatac2/metrics.xml
```diff
@@ -0,0 +1,285 @@
...
+                    <param argument="region_name" type="text" value="" label="Name of the region">
```
pavanvidem: expand with help a region can be a promoter or enhancer and so on.

tools/snapatac2/metrics.xml
```diff
@@ -0,0 +1,285 @@
...
+            </when>
+        </conditional>
+        <expand macro="param_common_advanced"/>
+    </inputs>
+    <outputs>
+        <data name="anndata_out" format="h5ad" from_work_dir="anndata.h5ad.gz" label="${tool.name} (${method.method}) on ${on_string}: Annotated data matrix"/>
+        <data name="hidden_output" format="txt" label="Log file">
+            <filter>advanced_common['show_log']</filter>
+        </data>
+        <data name="out_summary" format="tabular" from_work_dir="summary_by_chrom.tabular" label="${tool.name} (${method.method}) on ${on_string}: Summary by chromosome">
```
pavanvidem: switch the order of `hidden_output` and `out_summary`. Maybe rename `out_summary` to `summary_by_chrom`.

tools/snapatac2/macros.xml
```diff
@@ -117,79 +125,57 @@ with open('anndata_info.txt','w', encoding='utf-8') as ainfo:
...
+            <option value="fragment">"fragment": based on the number of fragments that overlap with a region of interest</option>
+            <option value="insertion">"insertion": based on the number of insertions that overlap with a region of interest</option>
+            <option value="paired-insertion" selected="true">"paired-insertion": similar to "insertion", but it only counts the insertions once if the pair of insertions of a fragment are both within the same region of interest</option>
         </param>
     </xml>
-    <xml name="mat">
-        <param argument="peak_mat" type="data" format="h5ad" optional="true" label="AnnData or AnnDataSet object storing the cell by peak count matrix"/>
-        <param argument="gene_mat" type="data" format="h5ad" optional="true" label="AnnData or AnnDataSet object storing the cell by gene count matrix"/>
+    <xml name="param_chrom_sizes" tokens="varname" token_label="Select genome to retrieve chromosome size">
+        <param name="@VARNAME@" type="select" label="@LABEL@">
```
pavanvidem: Do we get similar results if we use the Galaxy cached genomes? I thinks It is faster.

tools/snapatac2/network.xml
```diff
@@ -0,0 +1,276 @@
...
+        alpha = $alpha,
+        l1_ratio = $l1_ratio,
+        overwrite=False
+        )
+
+    # add tf binding
+    snap.tl.add_tf_binding(
+        network,
+        motifs = $motifs,
+        genome_fasta = $anno_file,
```
pavanvidem: use Galaxy cached genomes?

tools/snapatac2/network.xml
```diff
@@ -0,0 +1,276 @@
...
+        </param>
+        <param argument="pvalue_tf" type="float" value="0.01" label="P-value threshold to detect TF binding"/>
+        <expand macro="param_render_plot"/>
+        <expand macro="param_common_advanced"/>
+    </inputs>
+    <outputs>
+        <data name="hidden_output" format="txt" label="Log file" >
+            <filter>advanced_common['show_log']</filter>
+        </data>
+        <collection name="out_network" type="list" label="${tool.name} (${method}) on ${on_string} - Network">
```
pavanvidem, line 154: what is the actual tool output here? `pl.network_edge_stat` generated a png per cluster. 

Here, you select non-svgs and svgs. Does the tool create svg and html by default in addition to the png?

tools/snapatac2/peaks_and_motif_analysis.xml
```diff
@@ -182,30 +211,55 @@ sa.pl.regions(
...
                 <param argument="half_width" type="integer" value="250" label="Half width of the merged peaks"/>
             </when>
             <when value="pp.make_peak_matrix">
-                <expand macro="inputs_anndata"/>
-                <param argument="use_rep" type="text" optional="true" value="" label="Used to read peak information from .uns[use_rep]">
-                    <expand macro="sanitize_query"/>
-                </param>
-                <param argument="merged_peaks" type="data" format="tabular" label="Merged peaks file"/>
+                <expand macro="param_inputs_anndata"/>
+                <param argument="use_rep" type="text" value="" optional="true" label="Used to read peak information from .uns[use_rep]"/>
```
pavanvidem: what is an example input here?

tools/snapatac2/peaks_and_motif_analysis.xml
```diff
@@ -46,53 +52,72 @@ if __name__ == '__main__': ## a temporary fix https://github.com/kaizhang/SnapAT
...
 
 #else if $method.method == 'pp.make_peak_matrix'
-import polars
-peaks = polars.read_csv('$method.merged_peaks', separator='\t')
-adata = sa.pp.make_peak_matrix(
+snap.pp.make_peak_matrix(
     adata,
-    use_rep = peaks['Peaks'],
+    #if $method.use_rep != ''
+    use_rep = '$method.use_rep',
```
pavanvidem, line 85: did you test this?

tools/snapatac2/peaks_and_motif_analysis.xml
```diff
@@ -46,53 +52,72 @@ if __name__ == '__main__': ## a temporary fix https://github.com/kaizhang/SnapAT
...
+    #end if
+    inplace = True,
+    #if str($method.peak_file) != 'None'
+    peak_file = '$method.peak_file',
```
pavanvidem, line 88: If I understood it correctly, this method should use one of `use_rep` or `peak_file`. If that's true, it should be a conditional. 

In the current implementation, I don't know what happens if both are given :)

tools/snapatac2/peaks_and_motif_analysis.xml
```diff
@@ -182,30 +211,55 @@ sa.pl.regions(
...
+                    <param name="motif_enrichment" type="select" label="Perform motif enrichment analysis?">
+                        <option value="yes">Yes</option>
+                        <option value="no">No</option>
+                    </param>
+                    <when value="yes">
+                        <param name="motifs" type="select" label="A list of transcription factor motifs">
+                            <option value="snap.datasets.cis_bp(unique=True)">A list of transcription factor motifs curated by the CIS-BP database</option>
+                            <option value="snap.datasets.Meuleman_2020">A list of transcription factor motifs from the Meuleman et al. 2020 dataset</option>
+                        </param>
+                        <expand macro="param_chrom_sizes" varname="genome_fasta"/>
```
pavanvidem: try to use cached genomes if possbile

tools/snapatac2/preprocessing.xml
```diff
@@ -174,62 +219,103 @@ sa.pp.filter_doublets(
...
-    inplace = True,
+    groupby = '$method.groupby',
+    selections = None, # will add if requested by users
+    bin_size = $method.bin_size,
+    #if $method.blacklist:
+    blacklist = $method.blacklist,
+    #end if
+    normalization = '$method.normalization',
+    #if $method.include_for_norm:
+    include_for_norm = $method.include_for_norm,
```
pavanvidem:
```suggestion
    include_for_norm = '$method.include_for_norm',
```

tools/snapatac2/preprocessing.xml
```diff
@@ -174,62 +219,103 @@ sa.pp.filter_doublets(
...
+    #end if
+    #if $method.exclude_for_norm:
+    exclude_for_norm = $method.exclude_for_norm,
```
pavanvidem:
```suggestion
    exclude_for_norm = '$method.exclude_for_norm',
```

tools/snapatac2/preprocessing.xml
```diff
@@ -667,35 +843,164 @@ sa.metrics.tsse(
...
                 <assert_contents>
-                    <has_text_matching expression="sa.metrics.frag_size_distr"/>
-                    <has_text_matching expression="add_key = 'frag_size_distr'"/>
+                    <has_text_matching expression="snap.ex.export_fragments"/>
+                    <has_text_matching expression="groupby = &apos;leiden&apos;"/>
                 </assert_contents>
             </output>
-            <output name="anndata_out" location="https://zenodo.org/records/11260316/files/metrics.frag_size_distr.pbmc_500_chr21.h5ad" ftype="h5ad" compare="sim_size" delta_frac="0.1" />
+            <output_collection name="export_fragment" type="list" count="8">
+                <element name="-1.bed">
```
pavanvidem, line 965: why is it named like this?

nilchia: Somehow the Leiden clusters start  from -1 to 6
Files are named after the `groupby`

tools/snapatac2/peaks_and_motif_analysis.xml
nilchia: This not a FASTA file, it is a `Genome object` from [snapatac2](https://scverse.org/SnapATAC2/version/2.8/api/_autosummary/snapatac2.genome.Genome.html)

tools/snapatac2/peaks_and_motif_analysis.xml
nilchia: Good catch!

tools/snapatac2/network.xml
nilchia: The output is a `rx.PyDiGraph`, which is a directed network graph. I don't know of a way to save it on disc except as a pickle. That's why I ignored it

nilchia: Galaxy can not auto-detect SVG and PDF here correctly. That's why I separated them.
But also here, I'm not assigning the type :rofl:

tools/snapatac2/network.xml
nilchia: This not a FASTA file, it is a `Genome object` from [snapatac2](https://scverse.org/SnapATAC2/version/2.8/api/_autosummary/snapatac2.genome.Genome.html)

tools/snapatac2/macros.xml
nilchia: This is not a FASTA file, it is a `Genome object` from [snapatac2](https://scverse.org/SnapATAC2/version/2.8/api/_autosummary/snapatac2.genome.Genome.html)
Can we cache those objects on Galaxy?

nilchia:
Thanks a lot @bernt-matthias @pavanvidem for the review. I'll resolve the issues you mentioned and will ping you again when I'm done ✌️

tools/snapatac2/peaks_and_motif_analysis.xml
pavanvidem: Take cached FASTA file, GTF file from history and a chr lengths file as inputs and you can create such an object.

tools/snapatac2/dimension_reduction_clustering.xml
nilchia: it stores the embedding based on all the anndatas.
It is needed for running umap and visualization.
This is the actual output

tools/snapatac2/dimension_reduction_clustering.xml
nilchia: Sorry, just for me to make sure I understood correctly, did you mean I should add those to the recipe? Or you were just asking if they are included in the recipe?

tools/snapatac2/dimension_reduction_clustering.xml
nilchia: mention added here: [9d5bd55](https://github.com/galaxyproject/tools-iuc/pull/7078/commits/9d5bd557975a58800e69efa95266a77865d0fdfe)

tools/snapatac2/dimension_reduction_clustering.xml
nilchia: updated here: [8476dec](https://github.com/galaxyproject/tools-iuc/pull/7078/commits/8476dec2e4872280719c367966a9b9acbedb2fb7)

tools/snapatac2/dimension_reduction_clustering.xml
nilchia: fixed here: [8476dec](https://github.com/galaxyproject/tools-iuc/pull/7078/commits/8476dec2e4872280719c367966a9b9acbedb2fb7)

tools/snapatac2/dimension_reduction_clustering.xml
nilchia: added here: [1621910](https://github.com/galaxyproject/tools-iuc/pull/7078/commits/16219107f27a59b888ec44f0b3e914449d4c7ab6)

tools/snapatac2/metrics.xml
nilchia: fixed here: [0291e2e](https://github.com/galaxyproject/tools-iuc/pull/7078/commits/0291e2ed8e6df5928f291d6a160d98ef067733ba)

tools/snapatac2/metrics.xml
nilchia: fixed here: [de43308](https://github.com/galaxyproject/tools-iuc/pull/7078/commits/de4330835941c4b2010d2eba3a3f8b27df5f7bab)

tools/snapatac2/metrics.xml
nilchia: fixed here: [3ebcb83](https://github.com/galaxyproject/tools-iuc/pull/7078/commits/3ebcb835993b887d9da1b86985bc1f34f0cdab57)

tools/snapatac2/dimension_reduction_clustering.xml
bernt-matthias: I was wondering if these requirements should be part of the recipe. If the software does not run without them we should add it to the recipe.

nilchia: https://github.com/bioconda/bioconda-recipes/pull/60444

I created a PR

nilchia:
```
AssertionError: Test input file (https://zenodo.org/records/17512085/files/chr21.gff3.gz) cannot be found.
AssertionError: Test input file (https://zenodo.org/records/17512085/files/chr21.fasta.gz) cannot be found.
AssertionError: Test input file (https://zenodo.org/records/17512085/files/chrom_size.tabular) cannot be found.
```
Why can't it access these files?

nilchia:
What does it mean?
```
Parameter 'adata': the previously selected dataset has entered an unusable state
```

nilchia:
Thanks a lot @bernt-matthias @pavanvidem !! :tulip: 

It is again ready for review :pray:

tools/snapatac2/macros.xml
nilchia: It is using cached now

tools/snapatac2/network.xml
nilchia: It is using cached now

tools/snapatac2/peaks_and_motif_analysis.xml
nilchia: the info is stored in uns now when calling `merge_peaks` function.
test added

tools/snapatac2/peaks_and_motif_analysis.xml
nilchia: test added

tools/snapatac2/peaks_and_motif_analysis.xml
nilchia: It is using cached now

nilchia:
I commented out the test for the cached genome because it needs the full genome file.

nilchia:
The test in peak.xml that uses an indexed fasta fails because it requires a full fasta file, not a subsample.
So I changed that to be expected to fail.

tools/snapatac2/macros.xml
```diff
@@ -36,6 +36,20 @@
     <token name="@CMD_PRETTIFY_STDOUT@"><![CDATA[
         | sed -r '1 s|AnnData object with (.+) = (.*)\s*|\1: \2|g' | sed "s|'||g"  | sed -r 's|^\s*(.*):\s(.*)|[\1]\n-    \2|g' | sed 's|, |\n-    |g'
     ]]></token>
+    <token name="@CMD_GET_GFF@"><![CDATA[
+        #if $method.gff_file_condi.gffSource == 'cached':
+            ln -s '$method.gff_file_condi.gff_pre_installed.fields.path' gff &&
```
pavanvidem, line 41: usually there are no cached annotations available. This should always be taken from the history.

tools/snapatac2/macros.xml
```diff
@@ -36,6 +36,20 @@
...
+        #else:
+            ln -s '$method.gff_file_condi.gff_history' gff &&
+        #end if
+    ]]></token>
+    <token name="@CMD_GET_FASTA@"><![CDATA[
+        #if $method.fasta_file_condi.fastaSource == 'indexed':
+            zcat '$method.fasta_file_condi.fasta_pre_installed.fields.path' > fasta.fa &&
```
pavanvidem, line 48: do you have to unzip? what happens without `zcat`?

tools/snapatac2/macros.xml
```diff
@@ -145,17 +159,44 @@ key_added = '$method.key_added',
...
+            </param>
+            <when value="indexed">
+                <param name="fasta_pre_installed" type="select" label="Select a FASTA file" help="Select the FASTA file from a list of pre-installed genomes">
+                    <options from_data_table="all_fasta">
+                        <filter type="sort_by" column="2" />
+                    </options>
+                </param>
+            </when>
+            <when value="history">
+                <param name="fasta_history" type="data" format="fasta.gz" label="FASTA file" />
```
pavanvidem: can be also `fasta`format?

tools/snapatac2/macros.xml
```diff
@@ -145,17 +159,44 @@ key_added = '$method.key_added',
...
+            </when>
+        </conditional>
+    </xml>
+    <xml name="param_gene_anno">
+        <conditional name="gff_file_condi">
+            <param name="gffSource" type="select" label="Select a built-in GFF file or one from your history"  help="Choose history if you don't see the correct GFF" >
```
pavanvidem, line 212: As indicated above, always select the GFF from the history. Remove this conditional and make a plain data param that takes a GFF file from history.

tools/snapatac2/network.xml
```diff
@@ -151,12 +162,119 @@ print("All clusters processed!")
...
+            <param name="adata_atac" location="https://zenodo.org/records/17512085/files/snap_datasets_pbmc10k_multiome_ATAC.h5ad.gz"/>
+            <param name="groupby" value="cell_type"/>
+            <param name="pvalue_marker" value="0.1"/>
+            <conditional name="gff_file_condi">
+                <param name="gffSource" value="history"/>
+                <param name="gff_history" location="https://zenodo.org/records/17512085/files/chr21.gff3.gz"/>
+            </conditional>
+            <conditional name="fasta_file_condi">
+                <param name="fastaSource" value="history"/>
+                <param name="fasta_history" location="https://ftp.ebi.ac.uk/pub/databases/gencode/Gencode_human/release_41/GRCh38.primary_assembly.genome.fa.gz"/>
```
pavanvidem: That's a lot of data download for testing. Can you please subset only chr21 and store it on Zenodo. chr21 is tiny which should be less than 2% in size compared to the whole genome.

tools/snapatac2/peaks_and_motif_analysis.xml
```diff
@@ -460,7 +523,10 @@ snap.pl.regions(
                 <conditional name="enrichment_condi">
                     <param name="motif_enrichment" value="yes"/>
                     <param name="motifs" value="snap.datasets.cis_bp(unique=True)"/>
-                    <param name="genome_fasta" value="snap.genome.hg38"/>
+                    <conditional name="fasta_file_condi">
+                        <param name="fastaSource" value="history"/>
+                        <param name="fasta_history" location="https://ftp.ebi.ac.uk/pub/databases/gencode/Gencode_human/release_41/GRCh38.primary_assembly.genome.fa.gz"/>
```
pavanvidem: again, please use only chr21.

pavanvidem: btw, you can use the same file as cached file. Please check some mapping tools for examples.

tools/snapatac2/peaks_and_motif_analysis.xml
```diff
@@ -492,12 +558,37 @@ snap.pl.regions(
...
+                <param name="groupby" value="leiden"/>
+                <param name="pvalue" value="0.1"/>
+                <param name="out_file" value="png"/>
+                <expand macro="test_param_render_plot"/>
+                <conditional name="enrichment_condi">
+                    <param name="motif_enrichment" value="yes"/>
+                    <param name="motifs" value="snap.datasets.cis_bp(unique=True)"/>
+                    <conditional name="fasta_file_condi">
+                        <param name="fastaSource" value="indexed"/>
+                        <param name="fasta_pre_installed" value="hg38"/>
```
pavanvidem, line 580: I think  you can create a loc file and point the file location to the fasta file then use the value of the entry here.

tools/snapatac2/preprocessing.xml
```diff
@@ -125,7 +144,11 @@ snap.pp.add_tile_matrix(
 #else if $method.method == 'pp.make_gene_matrix'
 adata = snap.pp.make_gene_matrix(
     adata,
-    gene_anno = '$method.gene_anno',
+    #if $method.gff_file_condi.gffSource == 'cached':
+    gene_anno = '$method.gff_file_condi.gff_pre_installed.fields.path',
```
pavanvidem, line 148: again, can be removed.

tools/snapatac2/macros.xml
bernt-matthias: Why?

bernt-matthias: I find it incredibly useful to provide reference annotations to my users. Forcing users to find the reference annotations online seems wrong to me (at least for model species).

nilchia: I also noticed that on EU server at least there are no annotartions stored.

However I also agree to have them. From EBI at least.

Or at least mention that the reference is from which database and what annotation is compatible with that.

We got many error report that are just caused by different chromosome naming in annotation file

tools/snapatac2/macros.xml
nilchia: The tool fails. it excepts a decompressed fasta

tools/snapatac2/network.xml
nilchia: the tool fails with subsetted fasta file.
I think to use chr21.fasta file we need to do the whole preprocessing again so the anndata matches the fasta

tools/snapatac2/peaks_and_motif_analysis.xml
nilchia: the tool fails with that fasta

tools/snapatac2/peaks_and_motif_analysis.xml
nilchia: I did the same here, no?

tools/snapatac2/macros.xml
nilchia: [8e404b2](https://github.com/galaxyproject/tools-iuc/pull/7078/commits/8e404b282777d125b52af39c809be20643b0d6a2)
added here

tools/snapatac2/macros.xml
bgruening: GFFs are updated frequently, its hard to keep track which one is the best version for a given reference geonome with some special encoding. I find that very hard without domain expertise.

I agree that it would be nice to have this, but imho we never had the bandwidth as a community to do this :(

nilchia:
for reprodicibility, @pavanvidem and I thought we should have the motifs locally. Here I did put the motifs from snapatac tool in the tool directory.
However in next version maybe we can add a data table for this

nilchia:
As discussed today at the tool dev SIG meeting, I brought back the GFF cached, as some instances provide this for the users.

nilchia:
Also we discussed that MEME files should be shipped as data table

SaimMomin12:
Thanks @nilchia and @pavanvidem

mvdbeek:
Deployment status: **skipped**

https://github.com/galaxyproject/tools-iuc/actions/runs/19670677363

nilchia:
@mvdbeek all tests passed before merge. Do you know why the deployment skipped?

somehow one test failed?

SaimMomin12:
> @mvdbeek all tests passed before merge. Do you know why the deployment skipped?
> 
> somehow one test failed?

I have retriggered the CI, lets see if it goes through

mvdbeek:
Deployment status: **skipped**

https://github.com/galaxyproject/tools-iuc/actions/runs/19670677363

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/19670677363

nilchia:
@SaimMomin12 thanks, it passed

PR #7084, opened 2025-06-27
Add genome_uploader

SantaMcCloud:
FOR CONTRIBUTOR:


SantaMcCloud:
Have to wait a bit till i can get to work but i just did a init commit to save a little bit of work. Right now i have to wait for answer in this RP: https://github.com/bioconda/bioconda-recipes/pull/47588 till either i get one or i just do the recipe this wrapper has to be in this state

bedroesb:
I wonder how we could distinguish this better from the Webin CLI, since this is tailored for Metagenomics if I am correct, and not all genome types? (the tool name reads quite generic)

bedroesb:
I also am not yet convinced that ENA Webin CLI will create the traffic we need :)

bgruening:
What kind of traffic are we talking here about :)

bedroesb:
haha sounds weird indeed,I mean, now the name is the tool that gets wrapped, but if one wants to upload a genome to ENA, that person will propably not look for "webin cli" (ENA nomenclature), but something with Ena genome upload or something. This is me reflecting in on my other pull request :)

bedroesb:
Sorry for bothering with this,  I got notifications from your commits because of my previous comment. I don't thing providing credentials like this is a good way, but it is suggested to use `user_preferences_extra_conf`. It also looks like you are adding credentials in your tests to let the GitHub Actions temporarily succeed? @bgruening this is also kinda related to my other PR where I am trying to avoid adding credentials as ENV variables

SantaMcCloud:
@bedroesb thank you for pointing it out. I wanted to changr this of course but first i wanted to see if there is any other way for not providing any credentials  in the test. If there is no other way it is fine to use this since i will change the password and hope this will be fine.

paulzierep:
We should check with Alireza Heidari how to use the server credentials securely: EGD: `Simple and Secure Credential Handling for Tools in Galaxy`.

paulzierep:
This is the way how credentials should be set up these days: https://github.com/galaxyproject/galaxy/pull/19084

tools/genome_uploader/genome_uploader.xml
```diff
@@ -0,0 +1,145 @@
+<tool id="genome_uploader" name="Genome-uploader" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
+    <description>Python script to upload bins and MAGs in fasta format to ENA</description>
```
bernt-matthias:
```suggestion
    <description>upload bins and MAGs to ENA</description>
```

tools/genome_uploader/genome_uploader.xml
```diff
@@ -0,0 +1,145 @@
...
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements"/>
+    <command detect_errors="exit_code">
+        <![CDATA[
+
+            mkdir outputs genome_contigs &&
+
+            echo -e 'ENA_WEBIN=\$ENA_WEBIN_ACCOUNT_NAME}\nENA_WEBIN_PASSWORD=\$ENA_WEBIN_PASSWORD' > '.env' &&
```
bernt-matthias: This will expose the password and username on the command line. Use a `configfile` to create `.env`.

tools/genome_uploader/genome_uploader.xml
```diff
@@ -0,0 +1,145 @@
...
+
+            #for $file in $contigs:
+                ln -s '$file' 'genome_contigs/${file.element_identifier}' &&
```
bernt-matthias: `element_identifier` needs to be sanitized (the `re.sub` "trick").

tools/genome_uploader/genome_uploader.xml
```diff
@@ -0,0 +1,145 @@
...
+
+            genome_upload
+            -u '$study'
+            --genome_info '$genome_info'
+            --out outputs
+            #if $type == 'bin':
+                --bins
+            #else
+                --mags
+            #end if
```
bernt-matthias:
```suggestion
            $type
```

tools/genome_uploader/genome_uploader.xml
```diff
@@ -0,0 +1,145 @@
...
+            <validator type="regex">^(?:ERP\d{6}|PRJEB\d{6})$</validator>
+        </param>
+        <param argument="--genome_info" type="data" format="tabular" label="Input genomes metadata file (tsv file)"
+            help="To see an example how the tsv file should look like have a look into the help section an the bottom of the tool!"/>
+        <param name="contigs" type="data" format="fasta,fasta.gz" multiple="true" label="Input genome contigs"/>
+        <param name="type" type="select" label="What will be uploaded"
+            help="Refer to the ENA definition in the help section at the bottom of the tool if you have doubt!">
+            <option value="bin">Bins</option>
+            <option value="mag">MAGs</option>
+        </param>
```
bernt-matthias:
```suggestion
        <param name="type" type="select" label="What will be uploaded" help="Refer to the ENA definition in the help section at the bottom of the tool if you have doubt!">
            <option value="--bins">Bins</option>
            <option value="--mags">MAGs</option>
        </param>
```

also adapt the test values.

tools/genome_uploader/genome_uploader.xml
```diff
@@ -0,0 +1,145 @@
...
+        <!-- PLEASE READ THE COMMENT DOWN BELOW THIS PARAMETER -->
+        <param argument="--live" type="boolean" truevalue="--live" falsevalue="" label="Upload to ENA live server"
+            help="WARNING: only use this option if you want to upload the data to the live ENA server. If you want to TEST your uplaod dint use this option then!"/>
```
bernt-matthias: fix `uplaod dint`

bernt-matthias:
What is the relation to https://github.com/galaxyproject/tools-iuc/pull/7605

tools/genome_uploader/genome_uploader.xml
SantaMcCloud: Is there any tool which use this? So i can check it otherwise i will habe a look and try it out

SantaMcCloud:
> What is the relation to #7605

This tool is also from them so it have the same issue as the mention tool there.

tools/genome_uploader/genome_uploader.xml
SantaMcCloud: I think i will take the same idea from the ena_web_cli too and use it here aswell since it should work the same as it does there

SantaMcCloud:
@bernt-matthias i think this should be fine or not?

tools/genome_uploader/genome_uploader.xml
```diff
@@ -0,0 +1,167 @@
...
+            #end if
+            ${tpa}
+            ${private}
+
+
+        ]]>
+    </command>
+    <configfiles>
+        <configfile name="credentials"><![CDATA[
+            #set $webin_id = $__user__.extra_preferences.get('ena_webin_account|webin_id', "").strip()
```
bernt-matthias: The new credential mechanism should be used instead of `__user__.extra_preferences`

tools/genome_uploader/genome_uploader.xml
```diff
@@ -0,0 +1,167 @@
...
+            if [ -z "\$webin_id" ]; then
+                ## Fail early with guidance if no username is present.
+                echo "No ENA credentials defined. Set your credentials via: User -> Preferences -> Manage Information" >&2;
+                exit 1;
+            else
+            ## Pull the password similarly.
+            password=`grep 'password' $credentials | cut -d':' -f2`;
+            fi;
+
+            echo -e 'ENA_WEBIN=\$webin_id\nENA_WEBIN_PASSWORD=\$password' > '.env' &&
```
bernt-matthias: directly create the `.env` file as a named configfile.

`echo ...` exposes the password in the process list

SantaMcCloud: okay i did remove it and change the name of the file

tools/genome_uploader/genome_uploader.xml
SantaMcCloud: I did change it but i think i did it wrong since the test are not working anymore, Can you have a look into it and let me know what i did wrong? Thank you!

tools/genome_uploader/genome_uploader.xml
```diff
@@ -0,0 +1,153 @@
...
+        <configfile name=".evn"><![CDATA[
+            #set $webin_id = $ENA_WEBIN_ACCOUNT_NAME
+            #set $password = $ENA_WEBIN_PASSWORD
+            #if  $webin_id == "" and $live == "false":
+            #set $webin_id = "test-webin"
+            #set $password = "test-password"
+            #end if
+            #if $webin_id != "" :
+            ENA_WEBIN=$webin_id
+            nENA_WEBIN_PASSWORD=$password
```
bernt-matthias, line 44: the `n` need to be removed

tools/genome_uploader/genome_uploader.xml
```diff
@@ -0,0 +1,153 @@
...
+                --centre_name '$centre_name'
+            #end if
+            ${tpa}
+            ${private}
+
+
+        ]]>
+    </command>
+    <configfiles>
+        <configfile name=".evn"><![CDATA[
```
bernt-matthias, line 35: This part is missing: https://docs.galaxyproject.org/en/master/dev/schema.html#tool-requirements-credentials

Note that you can not use environment variables in config file , but only cheetah variables.

SantaMcCloud:
@bernt-matthias now it should be fine. Sorry for the mistakes and such. I had my exam phase and it was my first time working with both of this. But thanks to you i did learn some new stuff about wrapping :)

SantaMcCloud:
okay the file can not be moved....can i use `cat` since it si read only?

tools/genome_uploader/genome_uploader.xml
```diff
@@ -0,0 +1,154 @@
...
+            #if '$centre_name':
+                --centre_name '$centre_name'
+            #end if
+            ${tpa}
+            ${private}
+
+        ]]>
+    </command>
+    <configfiles>
+        <configfile name="env"><![CDATA[
```
bernt-matthias: You can use the filename attribute of configfile. Then mv is not needed

SantaMcCloud: The problem is that this file has to be in the working dir when the tool is called. Therefor i tried to use the mv and my test work locally. I now did try it with linking and hope this works with the galaxy side now too since locally it did work

SantaMcCloud:
i did change the test now. The linking of the file is working and now is the probelm that the `Credentials` are not set so the tool can not run. I now do test on this error. 

I could add a test parameter to force `Credentials` to run this tool but im not sure if this is wanted or not.

paulzierep:
I think this tool should create some files: https://github.com/EBI-Metagenomics/genome_uploader#produced-files
Which is then used by webin_cli_handler to upload. We still need to meet with EBI and get a demo on how this all actually works.
But can you add the output and a test for it ?

SantaMcCloud:
> I think this tool should create some files: https://github.com/EBI-Metagenomics/genome_uploader#produced-files Which is then used by webin_cli_handler to upload. We still need to meet with EBI and get a demo on how this all actually works. But can you add the output and a test for it ?

This is not possible because it does require an account to upload it and either there is a way to set them for the credentials or we need a Dummy account which then can be used to run it because expose this on command line is not a problem.

I did run it manually which is used for the output section.

paulzierep:
> > I think this tool should create some files: https://github.com/EBI-Metagenomics/genome_uploader#produced-files Which is then used by webin_cli_handler to upload. We still need to meet with EBI and get a demo on how this all actually works. But can you add the output and a test for it ?
> 
> This is not possible because it does require an account to upload it and either there is a way to set them for the credentials or we need a Dummy account which then can be used to run it because expose this on command line is not a problem.
> 
> I did run it manually which is used for the output section.

But even if you run in test mode it should create the manifest files, no ?

SantaMcCloud:
> > > I think this tool should create some files: https://github.com/EBI-Metagenomics/genome_uploader#produced-files Which is then used by webin_cli_handler to upload. We still need to meet with EBI and get a demo on how this all actually works. But can you add the output and a test for it ?
> > 
> > 
> > This is not possible because it does require an account to upload it and either there is a way to set them for the credentials or we need a Dummy account which then can be used to run it because expose this on command line is not a problem.
> > I did run it manually which is used for the output section.
> 
> But even if you run in test mode it should create the manifest files, no ?

In test mode you still upload your files to a test server so no upload there means no output of the tool

tools/genome_uploader/genome_uploader.xml
bernt-matthias: It is in the working dir. Just use the filename attribute

SantaMcCloud: Then im lost sorry, how do you mean it?

tools/genome_uploader/genome_uploader.xml
```diff
@@ -0,0 +1,172 @@
...
```
bernt-matthias:
```suggestion
        <configfile filename=".env"><![CDATA[
```

Should be all you need.
Is the config file docs unclear?

SantaMcCloud: I did it like this but it didnt work testing it locally. I did change it now and see if it works here in GitHub,

SantaMcCloud: Sorry for the confusion and all other mistakes etc.

Issue #5783, opened 2024-02-20
Improve/Fix slices in Table Compute 


paulzierep:
Table Compute slices like this: 2:-2 specifies 2,1,0,-1,-2; 
this makes no sense at all, why would one want to slice the first 3 columns and then the last two, the usual application of slice would be to select all from the second to the second last column.
The correct logic is implemented in `toolshed.g2.bx.psu.edu/repos/iuc/filter_tabular/filter_tabular/3.3.0`
```
Python offset indexes or slices. Examples:
Column offset indexes: 0,3,1 (selects the first,fourth, and second columns)
Negative column numbers: -1,-2 (selects the last, and second last columns)
python slices ( slice(start, stop[, step]) select a range of columns):
0:3 or :3 (selects the first 3 columns)
3:5 (selects the fourth and fifth columns)
2: (selects all columns after the second)
-2: (selects the last 2 columns)
2::-1 (selects the first 3 columns n reverse order: third,second,first)
```

To fix it one needs to look into 
https://github.com/galaxyproject/tools-iuc/blob/2af4dcd9b3cea71fdae6b3aafd31079f8135a205/tools/table_compute/scripts/table_compute.py#L135-L156

Maybe the `rangemaker` can just be removed, and the slices passed to pandas `iloc` as it, will investigate ...

PR #7105, opened 2025-06-30
Update table compute - Change rangemaker logic to allow pandas like indexing 

paulzierep:
FOR CONTRIBUTOR:

Fix https://github.com/galaxyproject/tools-iuc/issues/5783


wm75:
I like this! Before I start reviewing it more carefully: is there any functionality that you'll loose with this change? Doesn't look like from a first glance, but just to make sure :-)

Other question: when we're moving that close to actual pandas syntax, is it worthwhile keeping the 1-based positive indexing, or would it be better to switch to pandas-style completely?

paulzierep:
> I like this! Before I start reviewing it more carefully: is there any functionality that you'll loose with this change? Doesn't look like from a first glance, but just to make sure :-)
> 
> Other question: when we're moving that close to actual pandas syntax, is it worthwhile keeping the 1-based positive indexing, or would it be better to switch to pandas-style completely?

I thought the same, usually we use 1-based indexing for most column-related tools in Galaxy, that was the reason, otherwise we can keep pandas / python based indexing. We could make it optionally, but that might confuse more than help. I really just want to get rid of the current indexing nonsense.

wm75 (commented):
https://github.com/galaxyproject/tools-iuc/blob/9e338cf5937ab230ed2e194411eb30a25a5b4ed0/tools/table_compute/table_compute.xml#L96

Looks like this should reuse `@VERSION@`

tools/table_compute/scripts/table_compute.py
```diff
@@ -50,62 +50,79 @@ def readcsv(filedict, narm):
...
+            - Positive indices are 1-based and converted to 0-based (1 → 0, 2 → 1, etc.).
+            - Negative indices are left unchanged (e.g., -1, -2), compatible with Pandas.
+            - 0 is not allowed and will raise an error.
+            - Ranges (e.g., "3:1", "1:-3") are inclusive and can be ascending or descending.
+            - Repeated indices are allowed and preserved.
+
+        Examples:
+            "1:3"       → [0, 1, 2]
+            "3:1"       → [2, 1, 0]
+            "2:-2"      → [1, 0, -1, -2]
```
wm75, line 87: isn't that the *old* behavior instead of the one you're aiming for?

tools/table_compute/scripts/table_compute.py
```diff
@@ -50,62 +50,79 @@ def readcsv(filedict, narm):
...
+            "-3:-1"     → [-3, -2, -1]
+            "2"         → [1]
+            "1,1,3:1"   → [0, 0, 2, 1, 0]
+            "1:-3"      → [0, -1, -2, -3]
```
wm75, line 91: same here?

wm75:
I think the change also deserves highlighting somewhere. Otherwise this can be a very unpleasant surprise for people updating, say, a workflow to use the new tool version and suddenly performing calculations on the wrong columns.
Probably a warning box at the very top of the tool help?

@bgruening do you think that would be enough? or a banner? but then that's only visible on .eu, or is it a bad idea after all to have such behavior change in the tool?

bgruening:
I had hoped to stay out of this :)

What about changing the param name as well ... this way people are forced to look more closely at the param. We can then also add a warning to the param and the help section as well. Which we remove in a few versions.

wm75:
That's a good idea! It will cause many false alerts for people who aren't using slices with negative indices, but better this way than introducing silent bugs I think.

paulzierep:
> I had hoped to stay out of this :)
> 
> What about changing the param name as well ... this way people are forced to look more closely at the param. We can then also add a warning to the param and the help section as well. Which we remove in a few versions.

You mean Name or Label ? So changing what the user sees, or actually changing the name ?

paulzierep:
Will make an update with the correct logic today, still up to discussion is if we should use 0-based indexing, then people can fully rely on pandas based indexing ? Maybe this can be part of the hackathon ? I do not mind either way as long as we inform the user.

wm75:
changing the actual parameter name so that tool updates inside a workflow will break connections.

paulzierep:
> changing the actual parameter name so that tool updates inside a workflow will break connections.

ok now I got the logic

wm75:
wait, another idea, maybe silly:

what if you'd disallow slices with negative stops in the current version, but introduce alternative rust-like syntax for slices using `..` as the start..stop separator, and in that new alternative syntax, and only there, allow negative stops?

That way users that don't use expressions affected by the change could update to the new version without hassle, but if you are using them you have to conciously change to the new syntax.

wm75:
and one more note: your changes introduce a new inconsistency.
While `1:-2` now behaves like pandas df.iloc, `-2:2` still wraps around in forward direction creating a new asymmetry.
If you want to be consistent with iloc, you need to disallow cases where `left < 0`, but `right >=0`.

wm75:
> wait, another idea, maybe silly:
> 
> what if you'd disallow slices with negative stops in the current version, but introduce alternative rust-like syntax for slices using `..` as the start..stop separator, and in that new alternative syntax, and only there, allow negative stops?
> 
> That way users that don't use expressions affected by the change could update to the new version without hassle, but if you are using them you have to conciously change to the new syntax.

Thinking about this more, a good way forward might be:
1. leave the current `:` slice syntax as it is entirely and mention it as legacy in the parameter help
2. introduce new, alternative `..` syntax and implement the new behavior just for this one

This way, no exisiting usage breaks, and there are no surprises either. The new implementation can then also use 0-based indexing.

paulzierep:
> > wait, another idea, maybe silly:
> > what if you'd disallow slices with negative stops in the current version, but introduce alternative rust-like syntax for slices using `..` as the start..stop separator, and in that new alternative syntax, and only there, allow negative stops?
> > That way users that don't use expressions affected by the change could update to the new version without hassle, but if you are using them you have to conciously change to the new syntax.
> 
> Thinking about this more, a good way forward might be:
> 
> 1. leave the current `:` slice syntax as it is entirely and mention it as legacy in the parameter help
> 2. introduce new, alternative `..` syntax and implement the new behavior just for this one
> 
> This way, no exisiting usage breaks, and there are no surprises either. The new implementation can then also use 0-based indexing.

I agree, the best solution without breaking stuff but also allowing new logic would be the introduction of a new parameter with pandas based slicing. Since I am out of time before parental leave, I would propose this as a nice project for the tools' hackathon. In my opinion, I would completely revert my changes to `rangemaker` and just use the 0-based pandas iloc logic directly as the other parameter. This way users can apply pandas slicing, for which there are plenty of docs.

PR #7106, opened 2025-06-30
tool wrappers for LexicMap

Smeds:
tool wrapper for LexicMap commands: 
 - search
 - index

dependent on the following pull-request [20586](https://github.com/galaxyproject/galaxy/pull/20586) to galaxy which will add datatype  lexicmap_index

FOR CONTRIBUTOR:


bgruening:
lexicmap_query.fasta.gz seems to be missing and one tool test is failing.

Thanks @Smeds

tools/lexicmap/macros.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0"?>
+<macros>
+    <token name="@TOOL_VERSION@">0.7.0</token>
+    <token name="@VERSION_SUFFIX@">0</token>
+    <token name="@PROFILE_VERSION@">24.2</token>
```
bernt-matthias: Needs to be a version which includes the datatype.

Smeds: Planed to be incorporated with 25.1

https://github.com/galaxyproject/galaxy/pull/20586

Smeds:
@bernt-matthias , @nsoranzo could any one of you take a look at this pull-request?

tools/lexicmap/.shed.yml
```diff
@@ -0,0 +1,10 @@
+categories:
+- Next Gen Mappers
+description: LexicMap is a nucleotide sequence alignment tool for efficiently querying genomes. 
+long_description: |
+  LexicMap is a nucleotide sequence alignment tool for efficiently querying gene, plasmid, virus, or long-read sequences (>100 bp) against up to millions of prokaryotic genomes.
+name: lexicmap
+owner: iuc
+remote_repository_url: https://github.com/shenwei356/LexicMap
```
SaimMomin12:
```suggestion
remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/main/tools/lexicmap
```

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,148 @@
+<tool id="lexicmap_index" name="LexicMap Index" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" python_template_version="3.5" profile="@PROFILE_VERSION@">
```
SaimMomin12:
```suggestion
<tool id="lexicmap_index" name="LexicMap Index" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE_VERSION@">
```

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,148 @@
...
+    <description>
+        This tool builds LexicMap index.
```
SaimMomin12:
```suggestion
        Build LexicMap index.
```

tools/lexicmap/lexicmap.xml
```diff
@@ -0,0 +1,150 @@
+<tool id="lexicmap_search" name="LexicMap Search" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" python_template_version="3.5" profile="@PROFILE_VERSION@">
```
SaimMomin12:
```suggestion
<tool id="lexicmap_search" name="LexicMap Search" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE_VERSION@">
```

tools/lexicmap/lexicmap.xml
```diff
@@ -0,0 +1,150 @@
...
+    <description>
+        LexicMap is a nucleotide sequence alignment tool for efficiently querying genomes.
```
SaimMomin12:
```suggestion
        A nucleotide sequence alignment tool for efficiently querying genomes.
```

tools/lexicmap/macros.xml
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0"?>
```
SaimMomin12:
```suggestion
```

SaimMomin12 (commented):
Some preliminary comments inline

tools/lexicmap/.shed.yml
```diff
@@ -0,0 +1,10 @@
...
+homepage_url: https://bioinf.shenwei.me/LexicMap/
```
bernt-matthias: I would refer to the github repo here: https://github.com/shenwei356/LexicMap .. 

You may crosslink to  https://bioinf.shenwei.me/LexicMap/ in the help.

tools/lexicmap/lexicmap-index.xml
bernt-matthias: As @SaimMomin12 wrote, but without trailing `.`

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,148 @@
...
+    </description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="bio_tools"/>
+    <expand macro="requirements"/>
+
+    <command detect_errors="exit_code"><![CDATA[
+
+echo $input_genomes | sed 's/,/\n/g' > fasta_list.tsv &&
```
bernt-matthias: You will need a loop here: 

```suggestion
#for genome in $input_genomes
   echo '$genome' >> fasta_list.tsv &&
#end for
```

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,148 @@
...
+mkdir '$index.extra_files_path' &&
+lexicmap index 
+
+    --threads \${GALAXY_SLOTS:-1}
```
bernt-matthias: Use double quotes for shell variables

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,148 @@
...
+
+    -X \$(pwd)/fasta_list.tsv -O '$index.extra_files_path'
```
bernt-matthias: Just `./fasta_list.tsv` should work

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,148 @@
...
+        --seq-name-filter ${seq_name_filter}
+    #end if
+
+    ]]></command>
+    <inputs>
+        <param name="input_genomes" format="@FASTA_TYPES@" type="data" multiple="true" label="FASTA files" help="Should be of datatype &quot;fasta.gz&quot; or &quot;fasta&quot;"  />
+        <section name="advanced_settings" title="Advanced settings" expanded="false">
+            <param name="batch_size" type="integer" optional="true" label="Batch Size" help="Maximum number of genomes in each batch (maximum value: 131072)" />
+            <param name="big_genomes" type="text" optional="true" label="Big Genomes skiped output file" help="Out file of skipped files with $total_bases + ($num_contigs - 1) * $contig_interval &lt;= -g/--max-genome. The second column is one of theskip types: no_valid_seqs, too_large_genome, too_many_seqs." />
+            <param name="chunks" type="integer" optional="true" label="Number of chunks" help="Number of chunks for storing seeds (k-mer-value data) files. Max: 128. Default: the value of -j/--threads." />
```
bernt-matthias: Does it make sense for the user to control that. Does it influence parallelisation or also the results

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,148 @@
...
+
+    #if $advanced_settings.seq_name_filter
+        --seq-name-filter ${seq_name_filter}
+    #end if
+
+    ]]></command>
+    <inputs>
+        <param name="input_genomes" format="@FASTA_TYPES@" type="data" multiple="true" label="FASTA files" help="Should be of datatype &quot;fasta.gz&quot; or &quot;fasta&quot;"  />
+        <section name="advanced_settings" title="Advanced settings" expanded="false">
+            <param name="batch_size" type="integer" optional="true" label="Batch Size" help="Maximum number of genomes in each batch (maximum value: 131072)" />
```
bernt-matthias: Instead of optional parameters my preference would be to fill the default value as `value`. This is way more informative to the user and simplifies the command section, since you do not need the if statemenst.

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,148 @@
...
+            <param name="big_genomes" type="text" optional="true" label="Big Genomes skiped output file" help="Out file of skipped files with $total_bases + ($num_contigs - 1) * $contig_interval &lt;= -g/--max-genome. The second column is one of theskip types: no_valid_seqs, too_large_genome, too_many_seqs." />
```
bernt-matthias: This is a output file path? Should not be in the inputs.

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,148 @@
...
+            <param name="chunks" type="integer" optional="true" label="Number of chunks" help="Number of chunks for storing seeds (k-mer-value data) files. Max: 128. Default: the value of -j/--threads." />
+            <param name="contig_interval" type="integer" optional="true" label="Contig interval" help="Length of interval (N's) between contigs in a genome. It can't betoo small (&gt;1000) or some alignments might be fragmented" />
+            <param name="file_regexp" type="text" optional="true" label="Sequence file regex" help="Regular expression for matching sequence files in -I/--in-dir, case ignored. Attention: use double quotation marks for patterns containing commas, e.g., -p '&quot;A{2,}&quot;'. (default &quot;\\.(f&#91;aq&#93;(st&#91;aq&#93;)?|fna)(\\.gz|\\.xz|\\.zst|\\.bz2)?$&quot;)" />
```
bernt-matthias: `optional="true"` has no meaning for text parameters (since also an empty string is a string). You need to be aware that many of the characters that can be used in a regex are sanitized,  i.e. will be replaced by `X` in the command section. Check sanitizers if this is not intended. 

The variable in the command block needs to be single quoted and we need to forbid single quotes and probably also backslashes to make it safe.

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,148 @@
...
+    #if $advanced_settings.chunks
+        --chunks ${chunks}
+    #end if
+
+    #if $advanced_settings.contig_interval
+        --contig-interval ${contig_interval}
+    #end if
+
+    #if $advanced_settings.file_regexp
+        --file-regexp ${file_regexp}
```
bernt-matthias: Use single quotes for all text and data parameters.

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,148 @@
...
+            <param name="chunks" type="integer" optional="true" label="Number of chunks" help="Number of chunks for storing seeds (k-mer-value data) files. Max: 128. Default: the value of -j/--threads." />
+            <param name="contig_interval" type="integer" optional="true" label="Contig interval" help="Length of interval (N's) between contigs in a genome. It can't betoo small (&gt;1000) or some alignments might be fragmented" />
+            <param name="file_regexp" type="text" optional="true" label="Sequence file regex" help="Regular expression for matching sequence files in -I/--in-dir, case ignored. Attention: use double quotation marks for patterns containing commas, e.g., -p '&quot;A{2,}&quot;'. (default &quot;\\.(f&#91;aq&#93;(st&#91;aq&#93;)?|fna)(\\.gz|\\.xz|\\.zst|\\.bz2)?$&quot;)" />
+            <param name="kmer" type="integer" optional="true" label="Max k-mer size" help="Maximum k-mer size. K needs to be &gt;= 32." />
```
bernt-matthias: If possible add min and max for numeric parameters.

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,148 @@
...
+            <param name="mask_file" type="text" optional="true" label="Mask file" help="File of custom masks. This flag oversides -k/--kmer, -m/--masks, -s/--rand-seed etc." />
```
bernt-matthias: This is a file parameter. Should probably be `data`?

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,148 @@
...
+            <param name="masks" type="integer" optional="true" label="LexicHash masks" help="Number of LexicHash masks." />
+            <param name="max_genome" type="integer" optional="true" label="Max genome size" help="Maximum genome size. Genomes with any single contig larger than the threshold will be skipped, while fragmented (with many contigs) genomes larger than the threshold will be split into chunks and alignments from these chunks will be merged in &quot;lexicmap search&quot;. The value needs to be smaller than the maximum supported genome size: 268435456." />
+            <param name="max_open_files" type="integer" optional="true" label="Max open files" help="Maximum opened files, used in merging indexes. If there are &lt;100 batches, please increase this value and set a bigger &quot;ulimit -n&quot; in shell." />
```
bernt-matthias: I also would not make this a user choice?

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,148 @@
...
+            <param name="min_seq_len" type="integer" optional="true" label="Max sequence length" help="Maximum sequence length to index. The value would be k for values &gt;= 0." />
+            <param name="partitions" type="integer" optional="true" label="Number of partitions" help="Number of partitions for indexing seeds (k-mer-value data) files. The value needs to be the power of 4." />
```
bernt-matthias: Also remove?

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,148 @@
...
+            <param name="rand_seed" type="integer" optional="true" label="Rand seed" help="Rand seed for generating random masks." />
+            <param name="ref_name_regexp" type="text" optional="true" label="Reference name regular exptression" help="Regular expression (must contains &quot;(&quot; and &quot;)&quot;) for extracting the reference name from the filename. Attention: use double quotation marks for patterns containing commas, e.g., -p '&quot;A{2,}&quot;'. (default &quot;(?i)(.+)\\.(f&#91;aq&#93;(st&#91;aq&#93;)?|fna)(\\.gz|\\.xz|\\.zst|\\.bz2)?$\&quot;" />
```
bernt-matthias: Same as for the other regex parameter

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,148 @@
...
+            <param name="seed_data_threads" type="integer" optional="true" label="Seed data threads" help="Number of threads for writing seed data and merging seed chunks from all batches, the value should be in range of &#91;1, -c/--chunks&#93;. If there are &lt;100 batches, please also increase the value of --max-open-files and set a bigger &quot;ulimit -n&quot; in shell." />
```
bernt-matthias: SHould not be a user parameter

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,148 @@
...
+            <param name="seed_in_desert_dist" type="integer" optional="true" label="Seed in desert dist" help="Distance of k-mers to fill deserts." />
+            <param name="seed_max_desert" type="integer" optional="true" label="Seed max desert" help="Maximum length of sketching deserts, or maximum seed distance. Deserts with seed distance larger than this value will be filled by choosing k-mers roughly every --seed-in-desert-dist bases." />
+            <param name="seq_name_filter" type="text" optional="true" label="Sequence name filter" help="List of regular expressions for filtering out sequences by contents in FASTA/Q header/name, case ignored." />
```
bernt-matthias: Same as for the other regexp

tools/lexicmap/tool_data_table_conf.xml.test
bernt-matthias, line 1: You will also need to have a `tool_data_table_conf.xml.sample` file which refers to a loc.sample file in `tool-data/`

tools/lexicmap/lexicmap.xml
```diff
@@ -0,0 +1,150 @@
...
+    #if $advanced_settings.seed_min_single_prefix
+        --seed-min-single-prefix ${advanced_settings.seed_min_single_prefix}
+    #end if
+
+    ]]></command>
+    <inputs>
+        <param name="query" type="data" format="fasta.gz" label="LexicMap query file" multiple="true"  help=""/>
+        <conditional name="db_opts">
+            <param name="db_opts_selector" type="select" label="LexiMap index source">
+              <option value="histdb" selected="true">From your history</option>
```
bernt-matthias, line 53: Is there a chance to add a test for the histdb option?

tools/lexicmap/lexicmap.xml
```diff
@@ -0,0 +1,150 @@
...
+
+    ${advanced_settings.load_whole_seeds}
+
+    #if $db_opts.db_opts_selector == "histdb"
+        --index '${db_opts.histdb.extra_files_path}'
+    #else:
+        --index '${db_opts.lexicmap_index.fields.path}'
+    #end if
+
+    $query --out-file $out_file
```
bernt-matthias: single quotes for all I/O files

tools/lexicmap/lexicmap.xml
```diff
@@ -0,0 +1,150 @@
...
+            <when value="histdb">
+                <param name="histdb" type="data" format="lexicmap_index" optional="false" label="LexicMap index" />
+            </when>
+            <when value="db">
+                <param name="lexicmap_index" type="select" optional="false" label="LexicMap index file">
+                    <options from_data_table="lexicmap_index"/>
+                </param>
+            </when>
+        </conditional>
+            <param name="top_n_genomes" type="integer" value="0" label="Keep top N genome matches for a query (0 for all)" multiple="false" />
```
bernt-matthias: multiple is not valid for integer parameters

tools/lexicmap/lexicmap.xml
```diff
@@ -0,0 +1,150 @@
...
+            <param name="align_max_gap" type="integer" optional="true" label="Align max gap" help="Maximum gap in a HSP segment." />
+            <param name="align_min_match_len" type="integer" optional="true" label="Align min match length" help="Minimum aligned length in a HSP segment." />
+            <param name="align_min_match_pident" type="float" optional="true" label="Align min match pident" help="Minimum base identity (percentage) in a HSP segment." />
+            <param name="all" type="boolean" truevalue="--all" falsevalue="" checked="false" label="All all columns" help="Output more columns, e.g., matched sequences. Use this if you want to output blast-style format with 'lexicmap utils 2blast'." />
+            <param name="load_whole_seeds" type="boolean" truevalue="--load-whole-seeds" falsevalue="" checked="false" label="Load whole seeds" help="Load the whole seed data into memory for faster search" />
```
bernt-matthias: Use `argument` for all parameters?

tools/lexicmap/lexicmap.xml
```diff
@@ -0,0 +1,150 @@
...
+            <param name="max_evalue" type="float" optional="true" label="Max evalue" help="Maximum evalue of a HSP segment." />
+            <param name="max_open_files" type="integer" optional="true" label="Max open files" help="Maximum number of open files" multiple="false" />
```
bernt-matthias: Should maybe not be user controllable

tools/lexicmap/lexicmap.xml
```diff
@@ -0,0 +1,150 @@
...
+                <param name="load_whole_seeds" value="true" />
+            </section>
+            <output name="out_file" value="lexicmap_query_result.tsv" />
+        </test>
+    </tests>
+    <help><![CDATA[
+    
+    Search sequences against an LexicMap index Database. For more information about settings
+    please visit: https://bioinf.shenwei.me/LexicMap/usage/search
+
```
bernt-matthias, line 127: Can you add info on what is in the output file?

tools/lexicmap/lexicmap-index.xml
bernt-matthias: Cool this actually seems to work. My mistake

tools/lexicmap/lexicmap-index.xml
Smeds: I have added default value where the documentation has default values

tools/lexicmap/lexicmap-index.xml
Smeds: that's true! I have changed it to an output

tools/lexicmap/lexicmap-index.xml
Smeds: done

tools/lexicmap/lexicmap-index.xml
Smeds: true! Changed now

tools/lexicmap/lexicmap-index.xml
Smeds: @bernt-matthias I have set a max values to 32 since the documentation say "Maximum k-mer size. K needs to be <= 32". But for min length they have no min value, but I guess it should be at leaste  1 or bigger, probable at least much higher.

tools/lexicmap/lexicmap-index.xml
Smeds: @bernt-matthias Chunk will affect number of generated output files and from I remember it will not affect the exact result except for number of files created. I have set it to 2 to make the testing easier. But I could update the test and make sure that I generate none empty files, maybe only looking at the first few files generated.

tools/lexicmap/lexicmap-index.xml
Smeds: @bernt-matthias Could there be a point of moving this to an admin configurable variable (if that is possible)? I don't have a strong opinion about keeping this.?

tools/lexicmap/lexicmap-index.xml
Smeds: @bernt-matthias  So to summarize 

1. I will remove the optional field for text.

2. I must add a sanitizer like this
```yaml
<param argument="--file-regexp" type="text" label="Sequence file regex" help="Regular expression for matching sequence files in -I/--in-dir, case ignored. Attention: use double quotation marks for patterns containing commas, e.g., -p '&quot;A{2,}&quot;'. (default &quot;\\.(f&#91;aq&#93;(st&#91;aq&#93;)?|fna)(\\.gz|\\.xz|\\.zst|\\.bz2)?$&quot;)" >
    <sanitizer>
            <!-- Remove single quote -->
            <remove value="&apos;"/>
            <remove value="\"/>
        </valid>
    </sanitizer>
</param>
```

tools/lexicmap/lexicmap-index.xml
Smeds: @bernt-matthias What would be the best option:
1. use galaxy GALAXY_SLOTS
2. somehow make it configurable be the administrator
3. remove completely  and relay on the default value

tools/lexicmap/lexicmap.xml
Smeds: I haven’t found any examples of how to implement such a test. In the MagicBlast wrapper, they have a similar option where the user can select to use a file, a database, or a history — but tests only exist for the file and database cases. There are a few more wrappers with similar options, but none tests the history option.

tools/lexicmap/lexicmap.xml
Smeds: All parameters should now be using argument

tools/lexicmap/lexicmap.xml
Smeds: @bernt-matthias  would it be a point of making it configurable by the admin?

Smeds:
@bernt-matthias  @SaimMomin12 Thanks for the reviews! I have made some of the changes locally and will push them when I get response on my questions.

tools/lexicmap/lexicmap-index.xml
bernt-matthias: In hindsight: since you do not use `-I/--in-dir` the parameter can be removed completely.

tools/lexicmap/lexicmap-index.xml
bernt-matthias: As for default values I often have a look in the source code :) Otherwise add min/max if obvious. here for instance values smaller 2 just make no sense.

tools/lexicmap/lexicmap-index.xml
bernt-matthias: Does `batches` mean `chunks`? Maybe that is another reason why I would try to avoid leaving `--chunks` user controllable. 

I would maybe just ignore the parameter for now.

tools/lexicmap/lexicmap-index.xml
bernt-matthias: I know to little about how parallelisation is done here. Do you?

tools/lexicmap/lexicmap-index.xml
bernt-matthias: It's super hard to make regular expressions shell safe. Without `\` regexes loose needed functionality and with them they are unsafe. 

I do not see much use in this parameter. Can we remove it (maybe leaving a comment why it has been removed). Users can use other tools to filter fastq data.

tools/lexicmap/lexicmap.xml
bernt-matthias: The MagicBlast example is a bad example :) The history option use a dataset of type `blastdbn` which is a composite datatype (i.e. not a directory datatype) and the `file` option takes fasta dataset from which a database is build. In both cases a dataset is consumed (in the 1st case of a composite datatype and in the 2nd case a simple datatype). I remember that composite datatypes were often difficult/impossible to use in tool tests. 

Note that from Galaxy point of view the magicblast tool just takes a dataset as input in the `history` and `file` case and an options from a datatable in the other case. IOn particular it does not take a history as input.

In your case you have a directory datatype. unfortunately I also have not seen a tool tests that uses a directory as input. I recall that I heard that you can simply consider it as a nomal datatype and provide a zip file of the directory contents in the test. Can you try this.

tools/lexicmap/lexicmap-index.xml
bernt-matthias: > Chunk will affect number of generated output files and from I remember

Why should a user control that?

tools/lexicmap/lexicmap.xml
Smeds: I will remove this for now and if we need to be able to modify this parameter value we can add it later.

tools/lexicmap/lexicmap-index.xml
Smeds: I will remove this for now and if we need to be able to modify this parameter value we can add it later.

tools/lexicmap/lexicmap-index.xml
Smeds: I will remove this for now and if we need to be able to modify this parameter value we can add it later.

tools/lexicmap/lexicmap-index.xml
Smeds: I will remove this for now and if we need to be able to modify this parameter value we can add it later.

Smeds:
> @bernt-matthias @SaimMomin12 Thanks for the reviews! I have made some of the changes locally and will push them when I get response on my questions.

I have uploaded my local changes

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,123 @@
+<tool id="lexicmap_index" name="LexicMap Index" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE_VERSION@">
+    <description>
+        This tool builds LexicMap index
```
SaimMomin12:
```suggestion
        Build LexicMap index
```

tools/lexicmap/lexicmap.xml
```diff
@@ -0,0 +1,141 @@
...
+            <param argument="--seed-max-gap" value="1000" type="integer" label="Seed max gap" help="Minimum gap in seed chaining." />
+            <param argument="--seed-min-prefix" value="15" type="integer" label="Seed min prefix" help="Minimum (prefix/suffix) length of matched seeds (anchors)." />
+            <param argument="--seed-min-single-prefix" value="17" type="integer" label="Seed min single prefix" help="Minimum (prefix/suffix) length of matched seeds (anchors) if there's only one pair of seeds matched." />
+        </section>
+    </inputs>
+    <outputs>
+        <data name="out_file" format="tsv" />
+    </outputs>
+    <tests>
+        <test>
```
SaimMomin12:
```suggestion
        <test expect_num_outputs="1">
```

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,123 @@
...
+            <param argument="--seed-max-desert" value="100" type="integer" label="Seed max desert" help="Maximum length of sketching deserts, or maximum seed distance. Deserts with seed distance larger than this value will be filled by choosing k-mers roughly every --seed-in-desert-dist bases." />
+            <param argument="--seq-name-filter" type="text" optional="true" label="Sequence name filter" help="List of regular expressions for filtering out sequences by contents in FASTA/Q header/name, case ignored." />
+        </section>
+    </inputs>
+    <outputs>
+        <data name="lexicmap_index" format="lexicmap_index" label="index data" />
+        <data name="lexicmap_big_genomes" format="tsv" label="out file with metrics about skipped genomes" />
+    </outputs>
+    <tests>
+        <test>
```
SaimMomin12:
```suggestion
        <test expect_num_outputs="1">
```

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,123 @@
...
+                </sanitizer>
+            </param>
+            <param argument="--seed-in-desert-dist" value="50" type="integer" label="Seed in desert dist" help="Distance of k-mers to fill deserts." />
+            <param argument="--seed-max-desert" value="100" type="integer" label="Seed max desert" help="Maximum length of sketching deserts, or maximum seed distance. Deserts with seed distance larger than this value will be filled by choosing k-mers roughly every --seed-in-desert-dist bases." />
+            <param argument="--seq-name-filter" type="text" optional="true" label="Sequence name filter" help="List of regular expressions for filtering out sequences by contents in FASTA/Q header/name, case ignored." />
+        </section>
+    </inputs>
+    <outputs>
+        <data name="lexicmap_index" format="lexicmap_index" label="index data" />
+        <data name="lexicmap_big_genomes" format="tsv" label="out file with metrics about skipped genomes" />
```
SaimMomin12, line 50: Is it possible to add a corresponding test to this output?

tools/lexicmap/lexicmap-index.xml
Smeds: change

tools/lexicmap/lexicmap.xml
Smeds: added

tools/lexicmap/lexicmap-index.xml
Smeds: added

tools/lexicmap/lexicmap-index.xml
Smeds: @SaimMomin12  I have updated the tests to also look at the lexicmap_big_genomes file

Smeds:
@SaimMomin12 @bernt-matthias  could one of you take one more look at my pull-request please?

tools/lexicmap/lexicmap-index.xml
SaimMomin12: Still missing?

tools/lexicmap/lexicmap.xml
Smeds: @bernt-matthias  Do you know if it is possible to test the history option? I haven't really found any information about it.

bernt-matthias: Just create a zip / tar file with the contents of the directory and use it as value of the test parameter. This will implicitly call the built in converter tool that crates a directory from the contents of the tar/zip file.

Smeds:
@SaimMomin12 @bernt-matthias would you have time to take a new look at the pull-request?

tools/lexicmap/lexicmap-index.xml
Smeds: Looks like it! I'm adding it now.

tools/lexicmap/lexicmap.xml
Smeds: I will try this! I didn't see this information=/

bernt-matthias: My bad. I forgot to submit the review :)

Smeds: @bernt-matthias  I have tried to setup this test without much success. The current implementation works if I spin up a local server, I can select data from the table or from the history and use it to run a searches. 

My current problems are:

1.  If don't change `<param name="histdb" type="data" format="lexicmap_index" optional="false" label="LexicMap index" />` to `<param name="histdb" type="data" format="directory" optional="false" label="LexicMap index" />`, i.e `lexicmap_index` to `directory`, the data will not be decompressed.
2. If I use directory as type the data is decompressed but will end up in a extra subfolder, ex `/tmp/tmpdd99snhi/files/7/1/b/dataset_71b67705-560c-43f2-9c04-b4484fccac5f_files/dataset_46c8a89c-d98e-4785-b50f-71a1890af206`. I would like to have it in `dataset_71b67705-560c-43f2-9c04-b4484fccac5f_files/`

Smeds:
@bernt-matthias I got one more request from Anton. For the query he would like to be able to input one ore multiple fasta entries in a text field (not having to upload data the history). Ex

```
>query1
ACGTGTGCA
>query2
ACGTCCGCA
```

is this possible and in that case what is the best approach?  

I have tried with this setup without success
```xml
<param label="Sequences" name="query_list" type="text" format="fasta" area="true" help="List of sequences in FASTA format">
   <sanitizer>
    <valid initial="default">
        <add value="&gt;"/>      <!-- allow '>' -->
        <add value="&#10;"/>    <!-- allow newline -->
    </valid>
</sanitizer>
```
I have been able to keep `>` but galaxy will still replace the new line with spaces
```xml
<tests>
<test>
...
<param name="query_list" value=">query1
GTTTGTATACAATAGATTAAGTCACCCCAACAATGTTGCTTCACACGTCTATTTTAATGTATATATTTCAGCAGTGAATT
TGGAACTCTTTGGACCTGTATACCATGTCATGGATGCGGCTGCTCAGGCCGGTGATGAGGAGTCATTTAGTTCAACCCCA
GAGGCAATCCAAAATCAACCCGATCCTTGTGGTGGTATTACAAGCCCTAAAGAACTGAAAGGCAAGGCAAATCAGGGTAA" />
</test>
</tests>
```

results in the following data:

```
>query1 GTTTGTATACAATAGATTAAGTCACCCCAACAATGTTGCTTCACACGTCTATTTTAATGTATATATTTCAGCAGTGAATT TGGAACTCTTTGGACCTGTATACCATGTCATGGATGCGGCTGCTCAGGCCGGTGATGAGGAGTCATTTAGTTCAACCCCA GAGGCAATCCAAAATCAACCCGATCCTTGTGGTGGTATTACAAGCCCTAAAGAACTGAAAGGCAAGGCAAATCAGGGTAA
```

tools/lexicmap/lexicmap.xml
Smeds: done

tools/lexicmap/lexicmap.xml
Smeds: true, removed it

tools/lexicmap/lexicmap.xml
Smeds: updated

tools/lexicmap/tool_data_table_conf.xml.test
Smeds: added

tools/lexicmap/lexicmap-index.xml
Smeds: I will remove it! If someone requests it we could look into adding it.

tools/lexicmap/lexicmap-index.xml
Smeds: I will remove this for now and if someone requests the feature we could look into adding it

tools/lexicmap/lexicmap-index.xml
Smeds: I will set it to 2, to have a default value

tools/lexicmap/lexicmap-index.xml
Smeds: true, removed this variable

tools/lexicmap/lexicmap-index.xml
bernt-matthias: Why? Please just take the default value defined in the tool. In this case it even seems to be reported in the command line help: https://github.com/shenwei356/LexicMap/blob/cc03c57ede9b5e27d31f999cd1df10ad805227ba/lexicmap/cmd/index.go#L107

Make sure that all default values used in the Galaxy tools are identical to the values used in the program (unless there is good reason to change it).

tools/lexicmap/lexicmap-index.xml
Smeds: updated

tools/lexicmap/lexicmap-index.xml
Smeds: done

tools/lexicmap/lexicmap-index.xml
Smeds: done

tools/lexicmap/.shed.yml
Smeds: done

tools/lexicmap/macros.xml
Smeds: done

tools/lexicmap/lexicmap.xml
Smeds: done

tools/lexicmap/lexicmap.xml
Smeds: done

tools/lexicmap/lexicmap-index.xml
Smeds: done

tools/lexicmap/lexicmap-index.xml
Smeds: done

tools/lexicmap/.shed.yml
Smeds: done

bernt-matthias:
> I have been able to keep `>` but galaxy will still replace the new line with spaces

Guess you need to use `&#10;` in the tests as well?

tools/lexicmap/lexicmap-index.xml
bernt-matthias: Might also be nice to add the superb comments to the parameter help, i.e. `Bigger values improve the search specificity and do not increase the index size.`

bernt-matthias:
> I got one more request from Anton. 

In general I do not like the idea (to implement such a copy paste machanism for one tool) to much. He could just copy paste the data in the upload form. 

If this is to much effort (indeed, its a few more clicks and a bit of waiting time for the upload) I would rather suggest to open a feature request for Galaxy in general.

Smeds:
> > I have been able to keep `>` but galaxy will still replace the new line with spaces
> 
> Guess you need to use `&#10;` in the tests as well?

I will play around with this feature in a new pull-request. Would be great to have this done first. 

**Saw you comment after posting this**
*But you don't see any problems with inputting fasta sequences in a text field, with newlines? Somewhere I read that galaxy will always clean away all new lines and that it's nothing you can turn off (don't remember where I read it though).*

tools/lexicmap/lexicmap.xml
bernt-matthias: We probably need to register the converters for the lexicmap_index datatypes in Galaxy's datatypes conf. I prepared this, but accidentally pushed directly. Currently trying to get this undone: https://github.com/galaxyproject/galaxy/pull/20854/

Will submit a new PR as soon as possible

tools/lexicmap/lexicmap-index.xml
Smeds: I added this to make the testing easier! I have now removed the option of setting chunks and updated the test to handle it.

tools/lexicmap/lexicmap-index.xml
Smeds: sorry! I mistyped, default is 31 and min is 2

tools/lexicmap/lexicmap.xml
Smeds: great!

Smeds: > We probably need to register the converters for the lexicmap_index datatypes in Galaxy's datatypes conf. I prepared this, but accidentally pushed directly. Currently trying to get this undone: [galaxyproject/galaxy#20854](https://github.com/galaxyproject/galaxy/pull/20854)
> 
> Will submit a new PR as soon as possible

@bernt-matthias 
How come you reverted the pull-request? 

 It looks like it works when I run it locally. When I add the changes to my local galaxy repo galaxy will try to extract the data:
 ```bash
 mkdir '/tmp/tmpaboebf0f/job_working_directory/000/3/outputs/dataset_ff3aafca-eb4e-479f-bac0-c8da53b5a7ee_files' && /
  cd '/tmp/tmpaboebf0f/job_working_directory/000/3/outputs/dataset_ff3aafca-eb4e-479f-bac0-c8da53b5a7ee_files' &&  /
  python -c "from galaxy.util.compression_utils import CompressedFile; CompressedFile('/tmp/tmpaboebf0f/files/7/4/2/dataset_74240678-a520-48e1-a811-6a11172fbf0c.dat').extract('.')"
 ```
 
 However with the current implementation I will end up with an extra folder, it will use the name of the archive as an output folder: 
 
https://github.com/galaxyproject/galaxy/blob/af2d24082c52cbf8709561044f2082758aec6ff3/lib/galaxy/util/compression_utils.py#L238-L283

```python
extraction_path = os.path.join(path, self.file_name)
```

Creating the following structure:
```bash
#Path: /tmp/tmpaboebf0f/job_working_directory/000/3/outputs/dataset_ff3aafca-eb4e-479f-bac0-c8da53b5a7ee_files/dataset_74240678-a520-48e1-a811-6a11172fbf0c/
genomes
genomes.chunks.bin
genomes.map.bin
info.toml
masks.bin
seeds
```

I would expect the files to end up in `/tmp/tmpaboebf0f/job_working_directory/000/3/outputs/dataset_ff3aafca-eb4e-479f-bac0-c8da53b5a7ee_files` and not in `/tmp/tmpaboebf0f/job_working_directory/000/3/outputs/dataset_ff3aafca-eb4e-479f-bac0-c8da53b5a7ee_files/dataset_74240678-a520-48e1-a811-6a11172fbf0c`

Smeds: Modifying https://github.com/galaxyproject/galaxy/blob/af2d24082c52cbf8709561044f2082758aec6ff3/lib/galaxy/util/compression_utils.py#L238-L283 to not add the filename (history id) to the extraction path will give me the result I expect `extraction_path = path # os.path.join(path, self.file_name)`. But I guess that may break some other part of galaxy, if this is the expected behavior.

bernt-matthias: > How come you reverted the pull-request?

I accidentally directly pushed to the release branch which needed to be undone properly. 

Now there is https://github.com/galaxyproject/galaxy/pull/20857

How was the zip file constructed?

Smeds: I tried compressing the data using both tar and zip. For tar, I navigated into the folder containing the content I wanted to compress and ran:
`tar -cf db.lmi.tar *`

For zip, I used the UI. 

When I manually extract either archive, all files are placed directly in the current directory rather than inside an enclosing folder.

Smeds: @bernt-matthias If I’m reading the code correctly, we’ll always end up with an extra folder inside the XXX_files directory, at least with this command and code combination. The code outputs the data into a folder named after the compressed archive.

Smeds: @bernt-matthias  
Because I compressed the folder contents rather than the folder itself, the code will find no common prefix ([#L256-L259](https://github.com/galaxyproject/galaxy/blob/af2d24082c52cbf8709561044f2082758aec6ff3/lib/galaxy/util/compression_utils.py#L256-L259))
```python
if not common_prefix_dir:
       extraction_path = os.path.join(path, self.file_name)
       if not os.path.exists(extraction_path):
            os.makedirs(extraction_path)
```

Compressing the folder itself (e.g., `tar -cf db.lmi.tar db.lmi`) would create a common prefix, and the file_name wouldn’t be added to extraction_path. However, I don’t want my content inside a db.lmi folder. What I want is to extract just the archive’s contents directly into my folder: `/tmp/tmpaboebf0f/job_working_directory/000/3/outputs/dataset_ff3aafca-eb4e-479f-bac0-c8da53b5a7ee_files`

bernt-matthias: > Modifying https://github.com/galaxyproject/galaxy/blob/af2d24082c52cbf8709561044f2082758aec6ff3/lib/galaxy/util/compression_utils.py#L238-L283 to not add the filename (history id) to the extraction path will give me the result I expect `extraction_path = path # os.path.join(path, self.file_name)`. But I guess that may break some other part of galaxy, if this is the expected behavior.

I think the `extraction_path = os.path.join(path, self.file_name)` is only executed if [`not common_prefix_dir:`](https://github.com/galaxyproject/galaxy/blob/af2d24082c52cbf8709561044f2082758aec6ff3/lib/galaxy/util/compression_utils.py#L257). This leads me to the working hypothesis that the directory data type silently assumes a directory (i.e. that all contents are contained in one directory -- which would be the common prefix).

Could you test this, i.e. if the whole roundtrip works? In particular how the zip file looks like if you download directory? And what the converter does with it in the tool test.

If this still adds an extra directory level I could fix the converter, i.e. if the converter creates the extra dir (dataset name) then move its contents one level up.

Smeds: @bernt-matthias  If I create my archive like this `cd db.lmi && tar -cf db.lmi.v2.tar *`. I will end up with a index path for lexicmap search that looks like this
```bash
ls -l /tmp/tmpqrgrae_f/files/b/9/1/dataset_b91c656b-886d-48cf-aad3-6fbacf7bf1d8_files
drwxr-xr-x 4 pks5769 domain users 4096 Sep  4 11:48 dataset_22b2c788-f3a6-428f-a3dd-42a83f9980b3

ls -l /tmp/tmpqrgrae_f/files/b/9/1/dataset_b91c656b-886d-48cf-aad3-6fbacf7bf1d8_files/dataset_22b2c788-f3a6-428f-a3dd-42a83f9980b3
drwxr-xr-x 3 pks5769 domain users   4096 Sep  4 11:48 genomes
-rw-r--r-- 1 pks5769 domain users      0 Sep  4 11:48 genomes.chunks.bin
-rw-r--r-- 1 pks5769 domain users     54 Sep  4 11:48 genomes.map.bin
-rw-r--r-- 1 pks5769 domain users    592 Sep  4 11:48 info.toml
-rw-r--r-- 1 pks5769 domain users 160032 Sep  4 11:48 masks.bin
drwxr-xr-x 2 pks5769 domain users   4096 Sep  4 11:48 seeds
```

If I instead use an archive where I zipped the db.lmi like this  `tar -cf db.lmi.v2.tar db.lmi`
I end up with this input index
```bash
ls -l /tmp/tmpmgs0fc8x/files/a/e/d/dataset_aedd72ed-0f75-41a4-8f89-8e52a64ee360_files
drwxr-xr-x 4 pks5769 domain users 4096 Sep  4 14:22 db.lmi

ls -l /tmp/tmpmgs0fc8x/files/a/e/d/dataset_aedd72ed-0f75-41a4-8f89-8e52a64ee360_files/db.lmi
drwxr-xr-x 3 pks5769 domain users   4096 Sep  4 14:22 genomes
-rw-r--r-- 1 pks5769 domain users      0 Sep  4 14:22 genomes.chunks.bin
-rw-r--r-- 1 pks5769 domain users     54 Sep  4 14:22 genomes.map.bin
-rw-r--r-- 1 pks5769 domain users    592 Sep  4 14:22 info.toml
-rw-r--r-- 1 pks5769 domain users 160032 Sep  4 14:22 masks.bin
drwxr-xr-x 2 pks5769 domain users   4096 Sep  4 14:22 seeds
```

What I want is
```bash
ls -l /tmp/tmpmgs0fc8x/files/a/e/d/dataset_aedd72ed-0f75-41a4-8f89-8e52a64ee360_files
drwxr-xr-x 3 pks5769 domain users   4096 Sep  4 14:22 genomes
-rw-r--r-- 1 pks5769 domain users      0 Sep  4 14:22 genomes.chunks.bin
-rw-r--r-- 1 pks5769 domain users     54 Sep  4 14:22 genomes.map.bin
-rw-r--r-- 1 pks5769 domain users    592 Sep  4 14:22 info.toml
-rw-r--r-- 1 pks5769 domain users 160032 Sep  4 14:22 masks.bin
drwxr-xr-x 2 pks5769 domain users   4096 Sep  4 14:22 seeds
```

Which I can get if I modify the `common_prefix_dir(self)` function to return "./" if the archive contains more then 1 file/folder but no common prefix.
```python
    def common_prefix_dir(self) -> str:
        """
        Get the common prefix directory for all the files in the archive, if any.

        Returns '' if the archive contains multiple files and/or directories at
        the root of the archive.
        """
        contents = self.getmembers()
        common_prefix = ""
        if len(contents) > 1:
            print('Multiple files')
            common_prefix = os.path.commonprefix([self.getname(item) for item in contents])
            # If the common_prefix does not end with a slash, check that is a
            # directory and all other files are contained in it
            common_prefix_member = self.getmember(common_prefix)
            print(common_prefix_member)
            if (
                len(common_prefix) >= 1
                and not common_prefix.endswith(os.sep)
                and common_prefix_member
                and self.isdir(common_prefix_member)
                and all(self.getname(item).startswith(common_prefix + os.sep) for item in contents if self.isfile(item))
            ):
                common_prefix += os.sep
            if not common_prefix.endswith(os.sep):
                common_prefix = "./"
        return common_prefix

```

bernt-matthias: I suggested an alternative fix: https://github.com/galaxyproject/galaxy/pull/20857/commits/46792317d2eb8ffece37852b1199c87a0e2dd4c2

Smeds: Nice=)

tools/lexicmap/lexicmap.xml
```diff
@@ -0,0 +1,150 @@
...
+            <param name="query" value="lexicmap_query.fasta.gz" />
+            <section name="advanced_settings">
+                <param name="load_whole_seeds" value="true" />
+            </section>
+            <output name="out_file" value="lexicmap_query_result.tsv" />
+        </test>
+        <test expect_num_outputs="1">
+            <conditional name="db_opts">
+                <param name="db_opts_selector" value="histdb"/>
+                <param name="histdb" value="db.lmi.zip" />
```
bernt-matthias:
```suggestion
                <param name="histdb" ftype="lexicmap_index" class="Directory" value="FOLDER" />
```

where `FOLDER` is a folder in `test-data`.

Smeds: @bernt-matthias  looks like that did it=) The test is now working and import the folder to the history.

Smeds: @bernt-matthias guess we don't need to wait for [#20857](https://github.com/galaxyproject/galaxy/pull/20857) since changing to `class="Directory"` worked for the test.

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,88 @@
...
+    </description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="bio_tools"/>
+    <expand macro="requirements"/>
+
+    <command detect_errors="exit_code"><![CDATA[
+
+echo $input_genomes | sed 's/,/\n/g' > fasta_list.tsv &&
```
bgruening:
```suggestion
echo '$input_genomes' | sed 's/,/\n/g' > fasta_list.tsv &&
```

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,88 @@
...
+    #if $mask_file
+        --mask-file '$mask_file'
+    #end if
+
+    ]]></command>
+    <inputs>
+        <param argument="--input-genomes" format="@FASTA_TYPES@" type="data" optional="false" multiple="true" label="FASTA files" help="Should be of datatype &quot;fasta.gz&quot; or &quot;fasta&quot;"  />
+        <section name="advanced_settings" title="Advanced settings" expanded="false">
+            <param argument="--batch-size" value="5000" max="131072" type="integer" label="Batch Size" help="Maximum number of genomes in each batch (maximum value: 131072)" />
+            <param argument="--contig-interval" value="1000" type="integer" label="Contig interval" help="Length of interval (N's) between contigs in a genome. It can't betoo small (&gt;1000) or some alignments might be fragmented" />
```
bgruening: you can add min / max to those arguments if you like, min="1000" might make sense?

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,88 @@
...
+            <param argument="--kmer" value="31" type="integer" min="2" max="32" label="Max k-mer size" help="Maximum k-mer size. K needs to be &gt;= 32." />
+            <param argument="--mask-file" format="txt" type="data" optional="true" label="Mask file" help="File of custom masks. This flag oversides -k/--kmer, -m/--masks, -s/--rand-seed etc." />
+            <param argument="--masks" value="20000" type="integer" label="LexicHash masks" help="Number of LexicHash masks." />
```
bgruening: min=1?

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,88 @@
...
+            <param argument="--max-genome" value="15000000" max="268435456" type="integer" label="Max genome size" help="Maximum genome size. Genomes with any single contig larger than the threshold will be skipped, while fragmented (with many contigs) genomes larger than the threshold will be split into chunks and alignments from these chunks will be merged in &quot;lexicmap search&quot;. The value needs to be smaller than the maximum supported genome size: 268435456." />
+            <param argument="--min-seq-len" value="-1" type="integer" label="Max sequence length" help="Maximum sequence length to index. The value would be k for values &gt;= 0." />
```
bgruening: what is k here?

tools/lexicmap/lexicmap.xml
```diff
@@ -0,0 +1,150 @@
...
+            <when value="db">
+                <param name="lexicmap_index" type="select" optional="false" label="LexicMap index file">
+                    <options from_data_table="lexicmap_index"/>
+                </param>
+            </when>
+        </conditional>
+        <param argument="--top-n-genomes" type="integer" value="0" label="Keep top N genome matches for a query (0 for all)" />
+        <section name="advanced_settings" title="Advanced settings" expanded="false">
+            <param argument="--align-band" value="100" type="integer" label="Align band" help="Band size in backtracking the score matrix (pseudo alignment" />
+            <param argument="-align-ext-len" value="1000" type="integer" label="Align extend length" help="Extend length of upstream and downstream of seed regions, for extracting query and target sequences for alignment. It should be &lt;= contig interval length in database." />
```
bgruening: min=0

tools/lexicmap/lexicmap.xml
```diff
@@ -0,0 +1,150 @@
+<tool id="lexicmap_search" name="LexicMap Search" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE_VERSION@">
+    <description>
+        A nucleotide sequence alignment tool for efficiently querying genomes.
+    </description>
```
bgruening:
```suggestion
    <description>nucleotide sequence tool for querying genomes</description>
```

That is a bit shorter, do you think still correct?

tools/lexicmap/lexicmap.xml
```diff
@@ -0,0 +1,150 @@
...
+            <param argument="--min-qcov-per-genome" type="float" optional="true" help="Minimum query coverage (percentage) per genome." />
+            <param argument="--min-qcov-per-hsp" type="float" optional="true" help="Minimum query coverage (percentage) per HSP." />
+            <param argument="--seed-max-dist" value="1000" type="integer" label="Seed max dist" help="Minimum distance between seeds in seed chaining. It should be &lt;= contig interval length in database." />
+            <param argument="--seed-max-gap" value="1000" type="integer" label="Seed max gap" help="Minimum gap in seed chaining." />
+            <param argument="--seed-min-prefix" value="15" type="integer" label="Seed min prefix" help="Minimum (prefix/suffix) length of matched seeds (anchors)." />
+            <param argument="--seed-min-single-prefix" value="17" type="integer" label="Seed min single prefix" help="Minimum (prefix/suffix) length of matched seeds (anchors) if there's only one pair of seeds matched." />
+        </section>
+    </inputs>
+    <outputs>
+        <data name="out_file" format="tsv" />
```
bgruening:
```suggestion
        <data name="out_file" format="tabular" />
```

And you can set the column names with https://github.com/galaxyproject/tools-iuc/blob/15f3ce6780216fc864f70a1759a9a98a5f6b11d5/tools/revoluzer/crex.xml#L48

tools/lexicmap/lexicmap.xml
Smeds: should be okej

tools/lexicmap/lexicmap-index.xml
Smeds: @bgruening 
Looking at the source code it looks like k is the same as the  kmer size parameter:

I could change it to say 'The value would be equal to the kmer parameter value for values <= 0'

tools/lexicmap/lexicmap-index.xml
Smeds: 1000 as a min for contig-interval is probably good . Don't know about a good max value, didn't find anything in the documentation or code.

tools/lexicmap/lexicmap.xml
Smeds: Is this needed if the output file includes a header row?  If I read the documentation galaxy will parse this and use that header as metadata?

Smeds:
> Very cool @Smeds. Just some last smaller comments.

thanks

tools/lexicmap/lexicmap.xml
Smeds: change to your suggestion

tools/lexicmap/lexicmap.xml
bernt-matthias: Only for tsv not for tabular.  @bgruening  .. 

I think more and more that we should use `tsv` more? But unfortunately tsv is not tabular.

Smeds: @bernt-matthias @bgruening I have tried with setting the output to a tabular format and also doing conditional settings depending on the all parameter

```xml
<outputs>
      <data name="out_file" format="tabular">
          <actions>
              <conditional name="advanced_settings.all">
                  <when value="true">
                      <action name="column_names" type="metadata" default="Qquery,qlen,hits,sgenome,sseqid,qcovGnm,cls,hsp,qcovHSP,alenHSP,pident,gaps,qstart,qend,sstart,send,sstr,slen,evalue,bitscore,cigar,qseq,sseq,align" />
                  </when>
                  <when value="false">
                      <action name="column_names" type="metadata" default="Qquery,qlen,hits,sgenome,sseqid,qcovGnm,cls,hsp,qcovHSP,alenHSP,pident,gaps,qstart,qend,sstart,send,sstr,slen,evalue,bitscore" />
                  </when>
              </conditional>
          </actions>
      </data>
  </outputs> 
```
  this will create the following output
  
  
Screenshot from 2025-09-16 08-40-48

Is there anything I can do about the first line that will be displayed? I could remove it using using `tail`, but I don't know if that would be bad practice, I usually like to keep headers in the file.

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,88 @@
+<tool id="lexicmap_index" name="LexicMap Index" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE_VERSION@">
+    <description>
+        Builds LexicMap index
+    </description>
```
SaimMomin12:
```suggestion
    <description>Builds LexicMap index</description>
```

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,88 @@
...
+    <command detect_errors="exit_code"><![CDATA[
+
+echo '$input_genomes' | sed 's/,/\n/g' > fasta_list.tsv &&
+mkdir '$lexicmap_index.extra_files_path' &&
+lexicmap index 
+
+    --threads "\${GALAXY_SLOTS:-1}"
+
+    -X ./fasta_list.tsv -O '$lexicmap_index.extra_files_path'
+
```
SaimMomin12:
```suggestion
```

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,88 @@
...
+    --big-genomes '$lexicmap_big_genomes'
+
```
SaimMomin12:
```suggestion
```

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,88 @@
...
+    --batch-size '$batch_size'
+    --contig-interval '$contig_interval'
+    --kmer '$kmer'
+    --masks '$masks'
+    --max-genome '$max_genome'
+    --min-seq-len '$min_seq_len'
+
```
SaimMomin12:
```suggestion
```

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,88 @@
...
+    --rand-seed '$rand_seed'
+
```
SaimMomin12:
```suggestion
```

tools/lexicmap/lexicmap-index.xml
```diff
@@ -0,0 +1,88 @@
...
+    --seed-in-desert-dist '$seed_in_desert_dist'
+    --seed-max-desert '$seed_max_desert'
+
```
SaimMomin12:
```suggestion
```

tools/lexicmap/lexicmap.xml
bernt-matthias: Guess you can leave it.

SaimMomin12 (approved):

tools/lexicmap/lexicmap.xml
Smeds: @bernt-matthias I have pushed the updates for the  output format and will leave the line

bgruening:
Thanks @Smeds!

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/17767873676

bernt-matthias:
We need `lexicmap_index.loc.sample` in `tool-data/`. Can you add this in a new PR?

PR #7129, opened 2025-07-10
New tool addition: Dos2unix

dadrasarmin:
FOR CONTRIBUTOR:

There is a tool wrapped with the same name [here](https://toolshed.g2.bx.psu.edu/repository/view_repository?f-free-text-search=dos2unix&sort=name&operation=view_or_manage_repository&id=a4213b20212375be) but it is not the identical tool. That one is a Perl script. There was no contact info to contact the developer of that tool.

The tool is not available on `biocontainer` but rather on `conda-forge` channel [here](https://anaconda.org/conda-forge/dos2unix).


tools/dos2unix/.shed.yml
```diff
@@ -0,0 +1,15 @@
...
+   The dos2unix tool is a command-line tool used to convert plain text files from the DOS/Microsoft 
+   Windows format (which uses carriage return and line feed CRLF line endings) to the Unix/Linux format 
+   (which uses only line feed LF line endings). This conversion is essential when transferring text files 
+   between different operating systems to ensure compatibility with Unix-based tools and scripts that may 
+   misinterpret DOS-style endings. In addition to basic line ending conversion, dos2unix can handle 
+   different encodings (including UTF-8 and BOMs).
+categories:
+  - Text Manipulation
+homepage_url: https://waterlan.home.xs4all.nl/dos2unix.html
+remote_repository_url: https://sourceforge.net/projects/dos2unix/
```
SaimMomin12:
```suggestion
remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/main/tools/dos2unix/
```

tools/dos2unix/dos2unix.xml
```diff
@@ -0,0 +1,76 @@
+<tool id="dos2unix" name="Text file format converters (dos2unix)" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE_VERSION@">
+  <description>
+      Convert text files with DOS or Mac line breaks to Unix line breaks and other text cleaning options.
+    </description>
```
SaimMomin12, line 4:
```suggestion
   </description>
```
Please check indentation here

tools/dos2unix/.shed.yml
```diff
@@ -0,0 +1,15 @@
+# .shed.yml supporting automatic pushes.
```
SaimMomin12:
```suggestion
```

SaimMomin12 (commented):
Nicely done @dadrasarmin! Some minor comments inline.

tools/dos2unix/test-data/censored.txt
bernt-matthias, line 1: Can we use a text with a less militaristic  background. Is copyright a problem? Maybe a single Chinese character is sufficient for the example?

bernt-matthias (commented):
What is the usercase of this?

Can't remember that I needed the tool in the last decade...

tools/dos2unix/test-data/censored.txt
dadrasarmin: Hi,

We noticed that the following tutorial is behaving differently for different users during a workshop:
https://training.galaxyproject.org/training-material/topics/digital-humanities/tutorials/text_mining_chinese/tutorial.html

After some investigation, we noticed that if a text file is created/edited with some Windows versions or MS office software, then they have a BOM. This affects down stream analysis for text cleaning as some tools behave differently for file with/without BOM. According to Wolfgang:

> I guess the time window when Windows programs defaulted to utf-8 and wrote the BOM was just too short for many users to be affected.
> For ages Windows used cp-1252, then gradually switched to utf-8, but only initially had "with BOM" as the default. Notepad e.g. defaults to standard utf-8 since Windows 10 19H1, Powershell since v6.

The issue is that in some fields people tend to use Windows machines and old software. With this tool the tutorials/workflows that are targeting that audience will behave as expected as well.

I cannot speak or read Chinese. This text was taken from the tutorial mentioned above (from a user who reported the issue). I can reduce the text file to 2 characters and address the concern about the content of test file. I think the text is generated by ChatGPT for the tutorial. I am not sure.

bernt-matthias: > I think the text is generated by ChatGPT for the tutorial. I am not sure.

According to google translate, it seems to be from a hongkong newspaper from the time when the Japanese invaded mainland China.

SaimMomin12: @dadrasarmin In this case, the best way ahead would be to replace the `censored.txt` to some neutral content.

dadrasarmin: I changed the content to include "this" and "paper". Thanks for your feedback.

SaimMomin12:
Thanks, everyone! Also, @dadrasarmin could you add this tool [here](https://github.com/usegalaxy-eu/usegalaxy-eu-tools/blob/master/tools_iuc.yaml) so it becomes available on EU.

PR #7141, opened 2025-07-14
Add MetaSBT tool suite

cumbof:
Add [MetaSBT](https://github.com/cumbof/MetaSBT) tool suite to Galaxy.

FOR CONTRIBUTOR:


tools/metasbt/.shed.yml
```diff
@@ -0,0 +1,18 @@
+name: metasbt
+owner: fabio
+categories:
+  - Metagenomics
+description: Microbial genomes characterization with Sequence Bloom Trees
+long_description: |
+  A scalable framework for automatically indexing microbial genomes and accurately
+  characterizing metagenome-assembled genomes with Sequence Bloom Trees
+remote_repository_url: https://github.com/cumbof/galaxytools/tree/main/tools/metasbt
```
bernt-matthias: This should point to the IUC repo (ie the link will only work after merging)

tools/metasbt/index.xml
```diff
@@ -0,0 +1,329 @@
...
+
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="creator"/>
+    <expand macro="requirements"/>
+
+    <command detect_errors="exit_code"><![CDATA[
+source /home/fabio/mambaforge/etc/profile.d/conda.sh && conda activate metasbt &&
+
```
bernt-matthias:
```suggestion
```

tools/metasbt/index.xml
```diff
@@ -0,0 +1,329 @@
...
+#set input_dir = "./genomes"
+
+mkdir -p "${input_dir}" &&
+
+#for $genome in $genomes:
+    #if $genome.ext.endswith("gz"):
+        gzip -dc "$genome" > "${input_dir}/${genome.name}.fna" &&
```
bernt-matthias: `.element_identifier` is better than `.name`. Also this needs to be made command line safe because the names/identifiers could contain characters that are harmful on the command line. 

Like so: https://github.com/galaxyproject/tools-iuc/blob/89ee0af6e955ff964b7984d77ad536e0a9154278/tools/coverm/macros.xml#L55

Also dataset variables (eg `genome`) need to be single quoted for security. Same for everything that is user supplied (e.g. the data in the taxonomy file)

tools/metasbt/index.xml
```diff
@@ -0,0 +1,329 @@
...
+            #if $advanced.quality_control_selection.quality_control == "true":
+                  --completeness "$advanced.quality_control_selection.completeness"
+                  --contamination "$advanced.quality_control_selection.contamination"
+            #end if
+
+            &&
+#else:
+    find "${input_dir}" -type f -name "*.fna" -exec realpath {} \; > "${input_dir}/genomes.txt" &&
+
+    ln -s "${advanced.index_update_conditional.database_selection.db_tarball}" "./MetaSBT-Database.tar.gz" &&
```
bernt-matthias: You will need a if-else statemant here for the cvmfs case. 

For the cvmfs case you want to use `{advanced.index_update_conditional.database_selection.db_tarball.fields.path}`

A test for this case would be great.

tools/metasbt/index.xml
```diff
@@ -0,0 +1,329 @@
...
+        gzip -dc "$genome" > "${input_dir}/${genome.name}.fna" &&
+    #else
+        ln -s "$genome" "${input_dir}/${genome.name}.fna" &&
+    #end if
+#end for
+
+#if $advanced.index_update_conditional.index_update_option == "false":
+    while IFS=$'\t' read -r seq_name seq_taxonomy; do 
+        if [ -f "${input_dir}/${seq_name}.fna" ]; then
+            echo -e "${input_dir}/${seq_name}.fna\t${seq_taxonomy}" >> "${input_dir}/genomes.tsv";
```
bernt-matthias: Also here `seq_name`  needs to be made command line safe.

Also use single quotes.

tools/metasbt/index.xml
```diff
@@ -0,0 +1,329 @@
...
+            #end if
+
+            #if $advanced.quality_control_selection.quality_control == "true":
+                  --completeness "$advanced.quality_control_selection.completeness"
+                  --contamination "$advanced.quality_control_selection.contamination"
+            #end if
+
+            &&
+#else:
+    find "${input_dir}" -type f -name "*.fna" -exec realpath {} \; > "${input_dir}/genomes.txt" &&
```
bernt-matthias: This `find` call is not really necessary. You can just du this in the for loop, where the genomes are symlinked.

tools/metasbt/index.xml
```diff
@@ -0,0 +1,329 @@
...
+
+                        <when value="history">
+                            <param name="db_tarball" type="data" format="tar"
+                                    label="Select a MetaSBT database" />
+                        </when>
+                    </conditional>
+                </when>
+            </conditional>
+
+            <conditional name="dereplication_selection">
```
bernt-matthias, line 193: FYI: Conditionals have the disadvantage that they can not be alterered in workflows.

tools/metasbt/index.xml
```diff
@@ -0,0 +1,329 @@
...
+        <!-- Table with list of genomes and their assignments -->
+        <data format="tabular" name="genomes_table" label="${tool.name} on ${on_string}: genomes" from_work_dir="genomes.tsv">
+            <actions>
+                <action name="column_names" type="metadata" default="Genome,Type,Assigned Taxonomy,Internal Taxonomy" />
+                <action name="column_types" type="metadata" default="str,str,str,str" />
+            </actions>
+        </data>
+
+        <!-- Database as compressed tarball -->
+        <data format="tar" name="database" label="${tool.name} on ${on_string}: database" from_work_dir="MetaSBT-Database.tar.gz" />
```
bernt-matthias, line 256: format = tar.gz?

tools/metasbt/profile.xml
```diff
@@ -0,0 +1,147 @@
...
+
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="creator"/>
+    <expand macro="requirements"/>
+
+    <command detect_errors="exit_code"><![CDATA[
+source /home/fabio/mambaforge/etc/profile.d/conda.sh && conda activate metasbt &&
+
```
bernt-matthias:
```suggestion
```

tools/metasbt/profile.xml
```diff
@@ -0,0 +1,147 @@
...
+#for $genome in $genomes:
+    #if $genome.ext.endswith("gz"):
+        gzip -dc "$genome" > "${input_dir}/${genome.name}.fna" &&
+    #else
+        ln -s "$genome" "${input_dir}/${genome.name}.fna" &&
+    #end if
+#end for
+
+find "${input_dir}" -type f -name '*.fna' -exec realpath {} \; > "${input_dir}/genomes.txt" &&
```
bernt-matthias: Just fill the file in the for loop

tools/metasbt/profile.xml
```diff
@@ -0,0 +1,147 @@
...
+
+ln -s "$database.db_tarball" "./MetaSBT-Database.tar.gz" &&
```
bernt-matthias: This needs to consider the cached case.

tools/metasbt/profile.xml
```diff
@@ -0,0 +1,147 @@
...
+               --tarball "./MetaSBT-Database.tar.gz" &&
+
+metasbt profile --workdir "."
+                --database "Database"
+                --genomes "${input_dir}/genomes.txt"
+                --uncertainty "$advanced.uncertainty"
+                --pruning-threshold "$advanced.pruning_threshold"
+                --nproc "\${GALAXY_SLOTS:-4}" &&
+
+rm "./MetaSBT-Database.tar.gz" &&
```
bernt-matthias: Not needed. The job working dir will be cleaned anyway.

tools/metasbt/profile.xml
```diff
@@ -0,0 +1,147 @@
...
+
+mkdir output &&
+
+#for $i, $genome in enumerate($genomes):
+    sed 's/^# *//' "./tmp/profiles/${genome.name}.txt" > "./output/${genome.name}.tsv"
+
+    #if $i < len($genomes) - 1:
+        &&
+    #end if
```
bernt-matthias: You could just prepend the `&&` to the `sed`, `mkdir`, and  `rm` command. Alternatively, call `true` after the loop.

tools/metasbt/profile.xml
```diff
@@ -0,0 +1,147 @@
...
+            <discover_datasets pattern="(?P&lt;name&gt;.+)\.tsv" directory="output" ext="tabular" visible="false" />
+        </collection>
+    </outputs>
+
+    <tests>
+        <test>
+            <param name="genomes" value="genome_1.fna.gz,genome_2.fna.gz,genome_3.fna.gz,genome_4.fna.gz,genome_5.fna.gz,genome_6.fna.gz" ftype="fna.gz" />
+
+            <conditional name="database">
+                <param name="source" value="history" />
```
bernt-matthias: Test for the cvmfs case would be great.

tools/metasbt/profile.xml
```diff
@@ -0,0 +1,147 @@
...
+                        <validator message="No MetaSBT database is available" type="no_options"/>
+                    </options>
+                </param>
+            </when>
+
+            <when value="history">
+                <param name="db_tarball" type="data" format="tar"
+                       label="Select a MetaSBT database tarball" />
+            </when>
+        </conditional>
```
bernt-matthias: Use macros for code that appears in multiple tools.

tools/metasbt/tool_data_table_conf.xml.sample
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<tables>
+    <!-- Locations of MetaSBT databases in the required format -->
+    <table name="metasbt_databases" comment_char="#">
+        <columns>value, name, path</columns>
```
bernt-matthias: Wondering if some kind of version information would be useful here, i.e. the version of the tool used to generate the index and maybe versioning for the indices itself.

tools/metasbt/tool_data_table_conf.xml.sample
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<tables>
```
bernt-matthias, line 2: Do you plan to add a data manager for filling the data tables?

tools/metasbt/index.xml
cumbof: Ouch.. I forgot to remove this.. thanks for catching it

tools/metasbt/.shed.yml
cumbof: Done. I have also set `owner` to `iuc`

tools/metasbt/index.xml
cumbof: I replaced `.name` with `.element_identifier`, added single quotes, and sanitized user supplied things

tools/metasbt/index.xml
cumbof: I added the CVMFS case, thanks for catching that! I could definitely add a test for that as well

tools/metasbt/index.xml
cumbof: I’ve done something like this. It should be enough:
`genome_name='$(echo "$genome_name" | sed "s/[^[:alnum:]_.-]/_/g")' && `

tools/metasbt/index.xml
cumbof: I agree! I moved the `realpath` call into the for loop

tools/metasbt/index.xml
cumbof: I’ll take that in mind, thanks for pointing that out!

tools/metasbt/index.xml
cumbof: The problem with `tar.gz` as a type is that Galaxy tries to unpack it and keep the first file in the tarball only. That’s definitely something I do not want to happen. Using `tar` as a type here makes it work as expected

tools/metasbt/profile.xml
cumbof: Removed, thanks!

tools/metasbt/profile.xml
cumbof: Done!

tools/metasbt/profile.xml
cumbof: As I did in the `index.xml`, I added the CVMFS case here as well

tools/metasbt/profile.xml
cumbof: Yeah, removed

tools/metasbt/profile.xml
cumbof: I opted for prepending the `&&`, thanks for the advice!

tools/metasbt/profile.xml
cumbof: As I said earlier for the `index.xml` tool, yes, I could definitely add a test for the CVMFS case as well

tools/metasbt/profile.xml
cumbof: I moved the conditional to the `macros.xml` file

tools/metasbt/tool_data_table_conf.xml.sample
cumbof: The database version is encoded in the database name in the form `<db_name>-<yyyymmdd>` where `yyyymmdd` is the release date (year, month, day). I also keep track of the MetaSBT versions in a `metadata.json` file into the compressed tarballs.

tools/metasbt/tool_data_table_conf.xml.sample
cumbof: That’s a good point. It would be great to add a data manager, yes. I will hopefully start working on that soon

tools/metasbt/tool_data_table_conf.xml.sample
bernt-matthias: If the tool (maybe in the future) should be able to show the user only data from compatible versions then it would be good to store it in the data table. Note that we can not change a data table.

tools/metasbt/index.xml
bernt-matthias: Then you need sed installed in the container. The preferred solution is to do this in cheetah

tools/metasbt/index.xml
bernt-matthias: IMO: sanitation  would be better done in cheetah.

tools/metasbt/index.xml
```diff
@@ -0,0 +1,312 @@
...
+#for $genome in $genomes:
+    genome_name='$(echo "${genome.element_identifier}" | sed "s/[^[:alnum:]_.-]/_/g")' &&
+
+    #if $genome.ext.endswith("gz"):
+        gzip -dc '${genome}' > "${input_dir}/${genome_name}.fna" &&
+    #else
+        ln -s '${genome}' "${input_dir}/${genome_name}.fna" &&
+    #end if
+
+    realpath "${input_dir}/${genome_name}.fna" >> "${input_dir}/genomes.txt" &&
```
bernt-matthias: Why symlink if you then call realpath?

tools/metasbt/index.xml
bernt-matthias: Also if the user (or the tests) sets `tar.gz` as datatype?

tools/metasbt/index.xml
cumbof: I might be wrong but I cannot use cheetah here since it doesn’t have access to the content of `seq_name` and `seq_taxonomy` yet (now `genome_name` and `genome_taxonomy`). 
That’s the reason why I used `sed` and, since I have to do the same thing in the for loop a few lines above, I used `sed` there as well to make it consistent.

tools/metasbt/index.xml
cumbof: See my comment below about why I haven’t used cheetah here

tools/metasbt/index.xml
cumbof: So, let’s say that I’m a Galaxy user and I want to upload a MetaSBT database (a compressed tarball) into my history. If I upload a file and specify its datatype as `tar.gz`, that is what happens in Galaxy. It doesn’t actually keeps the file as it is. It extracts its content and keeps the first file in the tarball. 

I briefly discussed this with @jmchilton during the CoFest last month. Using `tar` definitely solves the problem

tools/metasbt/index.xml
cumbof: Replaced with an `echo -e`

tools/metasbt/index.xml
cumbof: Also, as far as I know, biocontainers use busybox by default and `sed` should be available in there

bgruening: It not included, and we should not assume that anything is installed :)

cumbof: Got it! I added `sed` as a requirement

tools/metasbt/tool_data_table_conf.xml.sample
cumbof: Makes sense to me. Does it mean that I should simply modify the `columns` entry with something like this in both `tool_data_table_conf.xml.sample` and `tool_data_table_conf.xml.test`?

```
<column>value, version, name, path</column>
```

And I guess I should also add a new column in `tool-data/metasbt_databases.loc` to reflect the change, correct?

cumbof:
@bernt-matthias @bgruening  I have added a test for the CVMFS case, both in `index.xml` and `profile.xml`.

The tests work but I got a problem with `planemo lint`. It works perfectly locally but it fails here because of the `--fail_level warn`. The warning that makes the linting failing is:

```
WARNING (InputsOptionsFiltersAllowedAttributes): Select parameter [db_tarball] 'sort_by' filter specifies unnecessary attribute 'reverse_sort_order'
```

This is because of the `<filter type="sort_by" column="0" reverse_sort_order="true"/>` in the `db_tarball` param’s list of options. It probably consider that unnecessary because there is only one entry in the test loc file, but it doesn’t apply for the real tool loc.

This is probably something to be fixed in the XSD?

bernt-matthias:
This is a problem with the linter that will be fixed here https://github.com/galaxyproject/galaxy/pull/20693

Meanwhile you can create a file `.lint_skip` in the tool directory

```
# linter bug https://github.com/galaxyproject/galaxy/pull/20693
InputsOptionsFiltersAllowedAttributes
```
then this linter will be ignored

cumbof:
Thanks @bernt-matthias , I’ve added a `.lint_skip` file as suggested

cumbof:
@bernt-matthias thanks for working on this PR! Do you think it is ready to be merged now?

cumbof:
@bgruening closed without merging?

bgruening:
Hä? How can this happen? I just merged the upstream Galaxy release branch in our EU Galaxy branch. 

This has nothing to do with IUC. Sorry, that was not intended ...

bgruening:
Really sorry.

cumbof:
@bgruening No problem! I see an error in the linting that is due to a connection issue with bioblend:
```
bioblend ERROR: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response')), 0 attempts left
```

Could you please restart the linting?

Also, is there any chance this could be merged soon? We have a paper ready to be pushed out 🙂

cumbof:
@bernt-matthias your PR has been merged https://github.com/galaxyproject/galaxy/pull/20693
I guess it is safe to remove `InputsOptionsFiltersAllowedAttributes` from the `.lint_skip` file now, isn’t it?

bernt-matthias:
Won't work yet, because we need a new point release of the Galaxy python packages. I hope that I can do this soon.

tools/metasbt/profile.xml
```diff
@@ -0,0 +1,139 @@
...
+mkdir -p output
+
+#for $genome in $genomes:
+    && genome_name="\$(echo '${genome.element_identifier}' | sed 's/[^[:alnum:]_.-]/_/g')"
+    && sed 's/^# *//' "./tmp/profiles/\${genome_name}.txt" > "./output/\${genome_name}.tsv"
+#end for
+]]></command>
+
+    <inputs>
+        <param name="genomes" format="fasta,fasta.gz" multiple="true" type="data"
```
bernt-matthias, line 55: One last question from my side: Is `multiple="true"` the right thing to do here? I had a quick look in the code and it seemed to me that MetaSBT processes the genomes individually. 

Even if it does this in parallel it would be more performant to leave the parallelization to Galaxy (and maybe the underlying HPC scheduler) -- currently the maximum number of cores is limited to the number of cores in a single node. Also the wrapper would become a bit simpler. 

If you make this `multiple="false"` users can still use a collection of genomes as input and will get collections of output files. The difference is Galaxy will create independent jobs, ie one per genome. 

Still `multiple="true"` might actually be the right thing to do here, if the processing time of a single genome is expected to be small (eg a multiple of the job scheduling overheads of Galaxy and maybe the HPC).

cumbof: MetaSBT can process genomes sequentially or in parallel (--nproc). The profiling of a single genome is generally quite fast, so I honestly wouldn’t change anything here.

blankenberg:
Thanks so much to @bernt-matthias for such a thorough review!

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/17163491320

bgruening:
> Unexpected HTTP status code: 500: {"err_msg": "Metadata may have been defined for some items in revision '47189346d00d'.  Correct the following problems if necessary and reset metadata.<br/><b>index.xml<\/b> - This file refers to a file named <b>metasbt_databases.loc<\/b>.  Upload a file named <b>metasbt_databases.loc.sample<\/b> to the repository to correct this error.<br/>"}

tools/metasbt/tool-data/metasbt_databases.loc
bernt-matthias, line 1: This file should be named `metasbt_databases.loc.sample` and it must be empty (we can not assume that people use cvmfs)

PR #7165, opened 2025-07-22
Update tools/gffcompare to 0.12.10

SaimMomin12:
FOR CONTRIBUTOR:

---
Fixes: https://github.com/galaxyproject/tools-iuc/pull/7157

### DIFF
--- a/tools/gffcompare/gffcompare.xml
+++ b/tools/gffcompare/gffcompare.xml
@@ -16,13 +16,13 @@
             ln -s '$input' '$escaped_element_identifier' &&
         #end for
         #if $conditional_annotation.selector == "yes":
-            #if $conditional_annotation.ref_source.ref_source_sel == "history":
+            #if $conditional_annotation.ref_source.selector == "history":
                 ln -s '$conditional_annotation.ref_source.reference_annotation' reference_annotation &&
             #else:
                 ln -s '${conditional_annotation.ref_source.index.fields.path}' reference_annotation &&
             #end if
         #end if
-        #if $seq_data.seq_data_selector == "Yes":
+        #if $seq_data.selector == "Yes":
             #if $seq_data.seq_source.index_source == "history":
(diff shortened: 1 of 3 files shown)


tools/gffcompare/gffcompare.xml
```diff
@@ -16,13 +16,13 @@
             ln -s '$input' '$escaped_element_identifier' &&
         #end for
         #if $conditional_annotation.selector == "yes":
-            #if $conditional_annotation.ref_source.selector == "history":
+            #if $conditional_annotation.ref_source.ref_source_sel == "history":
```
bgruening: those renames, will make automatic workflow updates impossible. Is it worth the pain?

SaimMomin12: @bgruening I am not sure, but the earlier version of this wrapper had several linting issues (such as deduplicated params,etc) which didn't make sense. At some point it needed an update, but if these updates create more problem downstream, I think we can live with .lint_skip file. What do you say?

bernt-matthias: True, but changing tests and (likely also) output labels does not influence the functionality of workflows. I would also vote for not changing parameter names.

SaimMomin12: Okay I will revert back the parameter names.

bernt-matthias: I do not understand why the linting errors can not be fixed without changing the parameter names?

SaimMomin12: If you see carefully in the conditionals, `selector` has been deduplicated multiple time.

SaimMomin12: Please ignore my last comment - i think i found why the lint fails

SaimMomin12: Fixed in https://github.com/galaxyproject/tools-iuc/pull/7165/commits/996d5843bdd6d8271ac84d16f211690b00f357be

bgruening (approved):

PR #7168, opened 2025-07-23
Add Galaxy wrapper for BiaPy tool

danifranco:
Hi all,

I'm submitting this initial version of a Galaxy wrapper for [BiaPy](https://biapy.readthedocs.io/), a deep learning framework for bioimage analysis.
This tool is designed to run a complete BiaPy workflow with two main usage modes:

1. Use a custom YAML configuration: Users can upload their own YAML config file and run BiaPy directly.
2. Create the YAML interactively: The tool generates a config file based on user inputs through a set of guided questions (conditional logic).

I’ve been working on this for a few days and, as this is my first Galaxy tool, I have some questions and doubts that I would really appreciate your feedback on:

1. **Parameter help text**. Is there a better way to provide detailed help for parameters, other than using long help="..." strings directly in the XML? The inputs have quite a bit of context, so I'm wondering if there's a cleaner way to manage this.
2. **Conda packaging**. I’m in the process of adding BiaPy to `conda-forge` ([PR here](https://github.com/conda-forge/staged-recipes/pull/30597)). Once it’s accepted, can I declare it in the requirements section like this?

    ```
    <requirement type="package" version="3.6.2">biapy</requirement>
    ```

    using the `conda-forge` channel?
3. **GPU support**. BiaPy supports specifying a GPU index as a parameter (e.g., `--gpu 0`). Training without GPU is very slow, but inference is fine on CPU. What's the best practice in Galaxy for handling GPU support? Should I remove this parameter or keep it optional?
4. **Clickable links in parameter help**. Can we include URLs in `help=""` so they are rendered as clickable links in the Galaxy interface? I did not test how they look like, because there are still things to do in this PR so it passes the checks, and to try it locally, so maybe the are already displayed like links. 
5. **Conditional logic depending on two parameters**. I’d like to show or hide certain inputs based on the combination of two user choices. From [this issue](https://github.com/galaxyproject/galaxy/issues/13931), it seems that’s not yet supported. Is that still the case? If so, I’ve repeated some inputs unnecessarily to work around this, but it would be good to confirm.
6. **GUI-side "check consistency" button**. In our current GUI version of BiaPy (outside Galaxy), we allow users to click a button to validate dataset consistency after selecting input folders. Is there a Galaxy-native way to do this (e.g., a button, a validation script or pre-run check)?
7. **EDAM terms**. I'm not 100% sure the EDAM operations I selected are correct. Some aren’t even defined in EDAM, but are common DL tasks (like classification, image-to-image translation, etc.). Would it be better to refer to EDAM-bioimaging terms? Or just leave out undefined terms entirely?
8. **Creating the YAML config programmatically**. In the second usage mode, I need to generate the YAML file first, then pass it to the BiaPy command. Currently I use a Python script to generate it. Is that acceptable for Galaxy tool execution flow (i.e., intermediate file creation within the command section)?
9. **Testing output folders**. Since Galaxy doesn’t support a `type="directory"` output, I was advised (by ChatGPT) to compress the output folder into a .tar.gz file:

    ```
    <data name="results_archive" format="tar.gz" ... />
    ```
    Is this the correct way to handle directory outputs? 

    I had two test scenarios in mind:

    1. Inference using a custom config (`test_data/input_test_data/example.yaml`) and pretrained model (just to check basic output).
    2. Running with full interactive config creation, to verify YAML creation and complete execution.

    I wanted to validate the test by asserting the presence of specific files in the tar archive (e.g. checkpoints, predictions). The full output structure is in `test_data/my_experiment_example.tar.gz`, and the test data is in` test-data/input_test_data/`, but I wasn't sure how best to structure this in the test block. In general, I have doubts on how to configure the output and the tests cases so maybe you can guide me a little bit.

Thanks a lot!

@bgruening @iuc-member

### DIFF
--- a/tools/biapy/biapy.xml
+++ b/tools/biapy/biapy.xml
@@ -173,7 +173,7 @@
             --result_dir './output'
             --name 'my_experiment'
             --run_id 1
-           \${GALAXY_BIAPY_GPU_STRING:-""}
+            --gpu \${GALAXY_BIAPY_GPU_STRING:-""}
 
         #set $outs = $selected_outputs or []
 


tools/biapy/biapy.xml
```diff
@@ -0,0 +1,316 @@
+<tool id="biapy" name="Build a workflow with BiaPy" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@" license="MIT">
```
SaimMomin12, line 1:
```suggestion
<tool id="biapy" name="Build a workflow with BiaPy" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@"  license="MIT">
```

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,316 @@
...
+                </section>
+            </when>
+        </conditional>
+        <section name="general_opts" title="General options" expanded="True">
+            <param name="result_dir" type="text" label="Select the output directory" help="A directory path where all the workflow files will be stored. Just a folder name, not a full path"/>
+            <param name="name" type="text" label="Select a name for the workflow" help="Just a name for the job. E.g. 'my_first_semantic_segmentation'"/>
+        </section>
+        <section name="advanced_opts" title="Advanced options" expanded="False">
+            <param name="run_id" type="integer" value="1" min="1" max="100" optional="true" label="ID" help="Number that should be increased when one need to run the same job multiple times (reproducibility)"/>
+            <param name="gpu" type="text" optional="true" label="Select a GPU to use" help="Number of the GPU to run the job in"/>
```
bernt-matthias: A Galaxy user (typically) has no control on CPU/GPU/memory parameters this should be admin controllable.

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,316 @@
...
+                                </conditional>
+                            </section> 
+                        </when>
+                    </conditional>
+
+                </section>
+            </when>
+        </conditional>
+        <section name="general_opts" title="General options" expanded="True">
+            <param name="result_dir" type="text" label="Select the output directory" help="A directory path where all the workflow files will be stored. Just a folder name, not a full path"/>
```
bernt-matthias: This must not be an arbitrary string. Users could enter malicious text. Maybe use a regex validator. 

Does the user really need to control this?

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,316 @@
...
+                            </section> 
+                            <section name="test_sec" title="Test data" expanded="True"> 
+                                <!-- Q11 -->
+                                <param name="test_raw_path" type="data" format="txt" label="Specify the location of the test raw image folder" help="In this step, you need to specify the folder where the images for testing the network are located. All these images must have the same number of channels. It's also important that each channel contains the same type of information to avoid confusing the model"/>
```
bernt-matthias: User should not have control over any folder names / paths. Instead use a data parameter where the images are given to the tool.

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,316 @@
...
...s for creating the coordinates are present in each file. Additionally, if the files contain a 'class' column, the number of classes will automatically be set to the highest value found in that column across all files. For Super-resolution, high-resolution images are expected, meaning these images will be 2x or 4x larger than the versions in the 'Train data (raw)' folder from the previous step. The checks include: 1) ensuring all images can be read correctly; 2) verifying that all images have the same number of input channels; 3) checking that all images are within the same value range. For example, if the images are uint8, their values should be between 0 and 255 across all images. For Image-to-Image, images are expected. The checks include: 1) ensuring all images can be read correctly; 2) verifying that all images have the same number of input channels; 3) checking that all images are within the same value range. For instance, if the images are uint8, the values should range between 0 and 255 across all images. You can find more information about each workflow and examples of dataset in BiaPy documentation: https://biapy.readthedocs.io/en/latest/get_started/select_workflow.html"/>
```
bernt-matthias: Do not provide paths, but images.

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,316 @@
...
+                            <option value="biapy" selected="true">No, I want to build a model from scratch</option>
+                            <option value="biapy_pretrained" selected="true">Yes, I have a model previously trained in BiaPy</option>
+                            <option value="bmz_torchvision" selected="true">Yes, I want to check if there is a pre-trained model I can use</option>
+                        </param>
+                        <when value="biapy_pretrained">
+                            <!-- Q4 -->
+                            <param name="biapy_model_path" type="data" format="data" label="Select the model trained with BiaPy before" help="This question is about setting the path to a pre-trained model trained before using BiaPy. The model should be stored in the folder you previously selected for saving results. Specifically, in the 'checkpoints' folder, there will be a file with a '.pth' extension, which represents the checkpoint of the pre-trained model"/>
```
bernt-matthias: do we have a more specific format than data in Galaxy

bernt-matthias (commented):
Few first comments.

bgruening:
Wuha! This is cool!

> 
> I'm submitting this initial version of a Galaxy wrapper for [BiaPy](https://biapy.readthedocs.io/), a deep learning framework for bioimage analysis. This tool is designed to run a complete BiaPy workflow with two main usage modes:
> 
>     1. Use a custom YAML configuration: Users can upload their own YAML config file and run BiaPy directly.

There are no paths in such files, aren't they?

>     2. Create the YAML interactively: The tool generates a config file based on user inputs through a set of guided questions (conditional logic).

You can use https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-configfiles-configfile for this, it might make the wrapper easier.

> 
> I’ve been working on this for a few days and, as this is my first Galaxy tool, I have some questions and doubts that I would really appreciate your feedback on:
> 
>     1. **Parameter help text**. Is there a better way to provide detailed help for parameters, other than using long help="..." strings directly in the XML? The inputs have quite a bit of context, so I'm wondering if there's a cleaner way to manage this.

The tool help is the best place. You can write this tool help in markdown or rst.

>     2. **Conda packaging**. I’m in the process of adding BiaPy to `conda-forge` ([PR here](https://github.com/conda-forge/staged-recipes/pull/30597)). Once it’s accepted, can I declare it in the requirements section like this?
>        ```
>        <requirement type="package" version="3.6.2">biapy</requirement>
>        ```
>        using the `conda-forge` channel?

Yes! :)

>     3. **GPU support**. BiaPy supports specifying a GPU index as a parameter (e.g., `--gpu 0`). Training without GPU is very slow, but inference is fine on CPU. What's the best practice in Galaxy for handling GPU support? Should I remove this parameter or keep it optional?

There is no automatic detection in place?

If so use an environment variable. Something like `\${GALAXY_GPU:-0}` but I guess we also need to disable it dynamically. So maybe  `\${BIAPY_GPU:-}` and then an admin could `BIAPY_GPU='--gpu=0'`.

>     4. **Clickable links in parameter help**. Can we include URLs in `help=""` so they are rendered as clickable links in the Galaxy interface? I did not test how they look like, because there are still things to do in this PR so it passes the checks, and to try it locally, so maybe the are already displayed like links.

I think you can, I would not do that but rather use the tool-help.

>     5. **Conditional logic depending on two parameters**. I’d like to show or hide certain inputs based on the combination of two user choices. From [this issue](https://github.com/galaxyproject/galaxy/issues/13931), it seems that’s not yet supported. Is that still the case? If so, I’ve repeated some inputs unnecessarily to work around this, but it would be good to confirm.

You can have two nested conditionals.

>     6. **GUI-side "check consistency" button**. In our current GUI version of BiaPy (outside Galaxy), we allow users to click a button to validate dataset consistency after selecting input folders. Is there a Galaxy-native way to do this (e.g., a button, a validation script or pre-run check)?

On which data should this operate? The input data? If so you can add a separate tool that validates the input and then you create a workflow that runs 1. validation and 2. your tool.

>     7. **EDAM terms**. I'm not 100% sure the EDAM operations I selected are correct. Some aren’t even defined in EDAM, but are common DL tasks (like classification, image-to-image translation, etc.). Would it be better to refer to EDAM-bioimaging terms? Or just leave out undefined terms entirely?

Leave them out for now. @matuskalas has promised to merge those BIII stuff into EDAM at some point :(

>     8. **Creating the YAML config programmatically**. In the second usage mode, I need to generate the YAML file first, then pass it to the BiaPy command. Currently I use a Python script to generate it. Is that acceptable for Galaxy tool execution flow (i.e., intermediate file creation within the command section)?

It is acceptable, but I think its rather cumbersome for you, I can imagine. Please look at https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-configfiles-configfile ... you can just "fill-out" a YAML file.

>     9. **Testing output folders**. Since Galaxy doesn’t support a `type="directory"` output, I was advised (by ChatGPT) to compress the output folder into a .tar.gz file:
>        ```
>        <data name="results_archive" format="tar.gz" ... />
>        ```
>        Is this the correct way to handle directory outputs?

What is your output? You can put `N` images for example in a collection. And have a second output with a single output txt file. Outputs can be very complex if you like. And what many people call directory, we call collection - and collections can be structured.

>        I had two test scenarios in mind:
>        
>        1. Inference using a custom config (`test_data/input_test_data/example.yaml`) and pretrained model (just to check basic output).
>        2. Running with full interactive config creation, to verify YAML creation and complete execution.
>        
>        I wanted to validate the test by asserting the presence of specific files in the tar archive (e.g. checkpoints, predictions). The full output structure is in `test_data/my_experiment_example.tar.gz`, and the test data is in` test-data/input_test_data/`, but I wasn't sure how best to structure this in the test block. In general, I have doubts on how to configure the output and the tests cases so maybe you can guide me a little bit.

Can you provide us with a listing of a typical output file and mark what is relevant for users?

Thanks for the tool, that looks very cool!
Bjoern

> Thanks a lot!
> 
> @bgruening @iuc-member

danifranco:
Here I go:

1. **Parameter help text**. 

> There are no paths in such files, aren't they?

Oh, true. The paths should be updated as they must match the ones locally. I will redo this part as a config file will always need to be stored. 

> You can use https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-configfiles-configfile for this, it might make the wrapper easier.

Thank for the hint. Will take a look!

> The tool help is the best place. You can write this tool help in markdown or rst.

Oh, perfect!

2. **Conda packaging**. 
> I’m in the process of adding BiaPy to `conda-forge` ([PR here](https://github.com/conda-forge/staged-recipes/pull/30597)). Once it’s accepted, can I declare it in the requirements section like this?
> >    ```
> >    <requirement type="package" version="3.6.2">biapy</requirement>
> >    ```
> >    using the `conda-forge` channel?
> > 
> 
> Yes! :)

It's being accepted in conda so everything is ready to go! ([https://anaconda.org/conda-forge/biapy](https://anaconda.org/conda-forge/biapy))

3. **GPU support**. 
> There is no automatic detection in place?
> 
> If so use an environment variable. Something like `\${GALAXY_GPU:-0}` but I guess we also need to disable it dynamically. So maybe `\${BIAPY_GPU:-}` and then an admin could `BIAPY_GPU='--gpu=0'`.
> 

It's not automatic in BiaPy, it is up to the user. I don't know if I'm understanding correctly how the GPU is selected here, but maybe we can add some check before calling BiaPy to detect if there is a GPU available and if so call BiaPy with the '--gpu 0' . Something like this (assuming the shell is bash):

```
if nvidia-smi > /dev/null 2>&1; then
    echo "GPU is available"
    gpu=1
else
    echo "No GPU found"
    gpu=0
fi
```
What do you think?

4. **Clickable links in parameter help**.  
> I think you can, I would not do that but rather use the tool-help.
 
I made it with markdown so it will be better now I think. 

5. **Conditional logic depending on two parameters**. 
> You can have two nested conditionals.

I think I can not use them (I just checked again in ChaGPT). I'm referring to having nested conditionals not inside the `<command>` section but in the rest of the file. ChaGPT pointed me to the same issue I found https://github.com/galaxyproject/galaxy/issues/13931 . Anyway, I have move those param definitions to the `macros.xml` so the main xml is more clean. 

6. **GUI-side "check consistency" button**.
> On which data should this operate? The input data? If so you can add a separate tool that validates the input and then you create a workflow that runs 1. validation and 2. your tool.

Yeah, the validation would be done in the inputs paths provided. I mean, that check is a thing that is done inside BiaPy always but just to let the user check it before running anything else. I'd like to have something after the path is given so if it is not possible don't know if creating a tool to check the data will add more complexity. If there is no option to do that but creating another tool maybe we can leave this and see if it will be a good thing to add in the future. 
 
7. **EDAM terms**.  
> Leave them out for now. @matuskalas has promised to merge those BIII stuff into EDAM at some point :(

ok, no problem!
 
8. **Creating the YAML config programmatically**. 
> It is acceptable, but I think its rather cumbersome for you, I can imagine. Please look at https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-configfiles-configfile ... you can just "fill-out" a YAML file.

Mmmm, I'm not understanding very well how the things are filled there. My approach should work too, no?

9. **Testing output folders**.
> What is your output? You can put `N` images for example in a collection. And have a second output with a single output txt file. Outputs can be very complex if you like. And what many people call directory, we call collection - and collections can be structured.

The output is a folder with different files depending on the workflow. For example, for a [semantic segmentation](https://biapy.readthedocs.io/en/latest/workflows/semantic_segmentation.html) workflow the output is something like this:

```
my_2d_semantic_segmentation/
├── config_files
│   └── my_2d_semantic_segmentation_1.yaml
├── checkpoints
│   └── my_2d_semantic_segmentation_1-checkpoint-best.pth
└── results
    └── my_2d_semantic_segmentation_1
        ├── aug
        │   └── .tif files
        ├── charts
        │   ├── my_2d_semantic_segmentation_1_*.png
        │   └── my_2d_semantic_segmentation_1_loss.png
        ├── per_image
        │   ├── .tif files
        │   └── .zarr files (or.h5)
        ├── per_image_binarized
        │   └── .tif files
        ├── full_image
        │   └── .tif files
        ├── full_image_binarized
        │   └── .tif files
        ├── full_post_processing
        │   └── .tif files
        ├── tensorboard
        └── train_logs
```

Some of the files are optional so I was thinking on create a test to just check the main files to be generated such as the predictions from the deep learning model that are placed in `my_2d_semantic_segmentation/results/my_2d_semantic_segmentation_1/per_image` . 

Thanks a lot for the support and the quick response!

tools/biapy/biapy.xml
danifranco: In BiaPy the user typically sets a path to a folder to store the results of the workflow. If there is an automatic output folder in Galaxy maybe we can use that automatically? What happens if the users want to run a workflow more than once? the output folder is the same? this will overwrite all the files no?

bernt-matthias: Every job runs in its own working dir and the tool can only write to this directory. So I guess `mkdir output` and hardcode `result_dir` as `output` is the solution.

tools/biapy/biapy.xml
danifranco: To train/test a workflow, e.g. [semantic segmentation](https://biapy.readthedocs.io/en/latest/workflows/semantic_segmentation.html), the users need to provide a dataset that must be structured as follows:

```
dataset/
├── train
│   ├── raw
│   │   ├── training-0001.tif
│   │   ├── training-0002.tif
│   │   ├── . . .
│   │   └── training-9999.tif
│   └── label
│       ├── training_groundtruth-0001.tif
│       ├── training_groundtruth-0002.tif
│       ├── . . .
│       └── training_groundtruth-9999.tif
└── test
    ├── raw
    │   ├── testing-0001.tif
    │   ├── testing-0002.tif
    │   ├── . . .
    │   └── testing-9999.tif
    └── label
        ├── testing_groundtruth-0001.tif
        ├── testing_groundtruth-0002.tif
        ├── . . .
        └── testing_groundtruth-9999.tif
```

I thought you could upload a zip or a folder with the data, as it is done in Google Colab for example, and then point to the path were the data is with that parameter. Following the example, a path to the test data (test_raw_path parameter) can be like: `/content/dataset/test/raw` . I assumed the root folder of Galaxy is `/content` in this example, as in Google Colab. I don't know how this things are handled here but that's the general idea.

tools/biapy/biapy.xml
danifranco: Answered in the above comment.

tools/biapy/biapy.xml
danifranco: Same here, I'm not sure how I need to handle this.

tools/biapy/biapy.xml
bernt-matthias: Essentially a user uploads data to Galaxy, the files are stored in a read-only dir using a random-hash and a `.dat` extension. 

In tools we typically we have a data parameter

```
<param name="raw_train" type="data" format="tiff" multiple="true"/>
```

then the user can select a bunch of the images he has uploaded previously. 

In the tool command section we can then:

```
mkdir -p dataset/train/raw &&
#for $i, $image in enumerate($raw_train)
    ln -s '$image' dataset/train/raw/training-${i}.tif &&
#end for
...
```

That is Galaxy will resolve `$image` to the internal file name.

danifranco: Perfect, thank you so much! I have modified how the train/test data is handled. Hope it is better now!

tools/biapy/biapy.xml
danifranco: The model checkpoint expected here is in `state_dict` format (`.pth` extension). Do you have something for that or I can leave it as `data` as it is? I suppose that I would need to do the `ln -s` as it is done with images, no?

bgruening:
@danifranco please let us know if you need any help.

The linter complains that you are missing a .shed.yml file. This will also trigger than additional tests.

If you have any questions for the config-file handling let me know.

danifranco:
Hi,

I still need help with the point 3 and 9. The third one maybe is quick but for the 9 I would need a bit more guidance on how to test the output. 
Thanks a lot!

bgruening:
For 3) just assume there is a variable (maybe call if BIAPY_GPU_ENABLED) and its set to "true". If this is there just enable GPU. The Galaxy admin can then set this for this tool if needed.

For 9) I would do the following:

Add a user-input option (type=select multiple=true) and let the user specify which output they want. e.g. charts, binarized images, post-processed images ....

And then you define the output collections like this:

https://github.com/galaxyproject/tools-iuc/blob/bb37578aa61bf4a47af262e02baf0a1c1d9d02c6/tools/hybpiper/hybpiper.xml#L220

Every collection can contain as many files as you want.

I hope that makes sense. Let me know if we can help.

tools/biapy/biapy.xml
danifranco: Can someone confirm this please? @bgruening

bernt-matthias: We do not have a datatype for these pytorch models. Seems that under the hood these are just pickled files. We try to avoid these files since pickle seems to unsafe: https://stackoverflow.com/questions/25353753/python-can-i-safely-unpickle-untrusted-data

Are there alternative formats that can be used?

bernt-matthias: Maybe `torch.onnx.export` .. seems that we already export `onnx`.

danifranco: As there is no specific format for this file extension I can just put a generic data format as the one I was using I asumme. My question was more related to how to handle the file. If I need to proceed as with the training/test data.

tools/biapy/create_yaml.py
```diff
@@ -0,0 +1,140 @@
+import argparse
+import yaml
+import requests
+
+def download_yaml_template(template_name="template_biax.yaml"):
+    url = f"https://raw.githubusercontent.com/BiaPyX/BiaPy/master/templates/{template_name}"
```
bernt-matthias: I would not download these files (github applies rate limiting more and more), but rather ship it with the tool. 

At least we need a versioned URL, e.g. https://raw.githubusercontent.com/BiaPyX/BiaPy/refs/tags/v3.6.2/templates/

tools/biapy/biapy.xml
bernt-matthias: > My question was more related to how to handle the file. If I need to proceed as with the training/test data.

Can you be more specific? 

> As there is no specific format for this file extension I can just put a generic data format as the one I was using I asumme. 

I need to stress that processing pth file (ie pickle files) is problematic. There is no problem if you work on private infrastructure and only have trustworthy users. But on public infrastructure it may become one because pickle files may allow arbitrary code execution. 

It would be really cool if biapy would support a safer format for saving the models.

bgruening: @danifranco if onnx is not possible, do you think you can support https://github.com/huggingface/safetensors ?

danifranco: > Can you be more specific?

In the same way you explained here: https://github.com/galaxyproject/tools-iuc/pull/7168#discussion_r2230773776 . Because this is the way Galaxy handles the input files, no?

Regarding the checkpoint we do not support any other extension but .pth. We load the weights as follows:
```
torch.load(checkpoint_path, weights_only=True, map_location="cpu")
```
This restricts deserialization to tensors only, blocking arbitrary Python object execution and greatly reducing pickle-related risks. We also load these weights into a predefined, trusted model architecture (never from user-supplied code).

tools/biapy/biapy.xml
danifranco: Great, I will do it like that. Thank you!

tools/biapy/create_yaml.py
danifranco: ok, will change it now. Thanks!

tools/biapy/biapy.xml
bernt-matthias: >In the same way you explained here: https://github.com/galaxyproject/tools-iuc/pull/7168#discussion_r2230773776 . Because this is the way Galaxy handles the input files, no?

If biapy requires a specific file extensions for any of the files you need to create a symlink before calling the main program, e.g. `ln -s $mode_selection.selected_mode.cfg_config_process.pretrained_model.biapy_model_path` biapy_model.pth` and then use the symlink.

tools/biapy/biapy.xml
```diff
@@ -60,47 +60,14 @@
...
         ########## Reuse provided yaml file and update paths ##########
         #if $mode_selection.selected_mode == 'custom_cfg':
             #if $mode_selection.get('test_sec') and $mode_selection['test_sec'].get('gt_test'):
                 #set $test_gt_avail = 'test_gt_yes'
             #end if
-            #if $mode_selection.get('biapy_model_path') is not None
-                ln -s '$mode_selection["biapy_model_path"]' ${checkpoint_file} 2>/dev/null || true &&
+            #set $mpath = $mode_selection.get('biapy_model_path')
+            #if $mpath and str($mpath) not in ['None', '']
+                ln -s '$mpath' ${checkpoint_file} 2>/dev/null || true &&
```
bernt-matthias: Why are you hiding the errors? Would the following be an option for all the symlink creations:

```suggestion
                ln -fs '$mpath' ${checkpoint_file} &&
```

danifranco: You are right! I have just added some changes.

danifranco:
I think I have finally corrected everything. What do you think?

tools/biapy/.shed.yml
```diff
@@ -0,0 +1,11 @@
+name: biapy
+owner: danifranco
```
bernt-matthias:
```suggestion
owner: iuc
```

tools/biapy/macros.xml
```diff
@@ -0,0 +1,63 @@
...
+    <token name="@PROFILE@">25.0</token>
+
+    <xml name="requirements">
+        <requirements>
+            <requirement type="package" version="@PYTHON_VERSION@">python</requirement>
+            <requirement type="package" version="@TOOL_VERSION@">biapy</requirement>
+            <!-- Make cv2 come from conda-forge; pin to a recent 4.10 series -->
+            <requirement type="package" version="4.10.*">opencv</requirement>
+            <!-- Ensure OpenEXR ABI matches OpenCV’s build -->
+            <requirement type="package" version="3.*">openexr</requirement>
```
bernt-matthias: If you specify this here you must be specific with the version, `*` is not valid.

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,639 @@
...
+        <!-- <edam_operation>operation_3442</edam_operation> Image classification -->
+        <!-- <edam_operation>Self-supervision learning (not in EDAM)</edam_operation>  -->
+        <edam_operation>operation_2944</edam_operation>  <!-- Image analysis -->
+    </edam_operations>
+    <expand macro="requirements" />
+    <expand macro="creators" />
+    <stdio>
+        <exit_code range=":-1" level="fatal" description="Error occurred. Please check Tool Standard Error"/>
+        <exit_code range="137" level="fatal_oom" description="Out of Memory"/>
+        <exit_code range="1:" level="fatal" description="Error occurred. Please check Tool Standard Error"/>
```
bernt-matthias: You can leave it if you add this for the error message, otherwise this is covered by `detect_errors="exit_code"`.

```suggestion
        <exit_code range=":-1" level="fatal" description="Error occurred. Please check Tool Standard Error"/>
        <exit_code range="137" level="fatal_oom" description="Out of Memory"/>
        <exit_code range="1:" level="fatal" description="Error occurred. Please check Tool Standard Error"/>
```

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,639 @@
...
+                                    <expand macro="test_gt_param"/>
+                                </when>
+                                <when value="test_gt_no"/>
+                            </conditional>
+                        </section> 
+                    </when>
+                </conditional>
+            </when>
+        </conditional>
+        <param name="selected_outputs" type="select" display="checkboxes" multiple="true" label="Select the outputs" help="Select which outputs to generate from running BiaPy (e.g. predictions, metrics, logs, or intermediate results).">
```
bernt-matthias, line 406: Should this be `optional="false"`?

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,639 @@
...
+                    </when>
+                    <when value="bmz_torchvision">
+                        <!-- Q5 -->
+                        <conditional name="bmz_torchvision_model">
+                            <param name="bmz_or_torchvision" type="select" label="Which is the source of the model?" help="Enter the source of the model, whether if it is available through the BioImage Model Zoo or TorchVision">
+                                <option value="bmz" selected="true">BioImage Model Zoo</option>
+                                <option value="torchvision">TorchVision</option>
+                            </param>
+                            <when value="bmz">
+                                <param name="bmz_model_name" type="text" format="data" value="sensible-cat" label="BioImage Model Zoo model name" help="Enter the name of a pre-trained model from the BioImage Model Zoo (https://bioimage.io/#/models); filter by the BiaPy icon and ensure the model matches your dimensionality (2D/3D) and task (e.g. semantic segmentation)."/>
```
bernt-matthias: type="data"?

Would be cool to have tests for more of the possibilities how this tool can be used.

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,639 @@
...
+                            </when>
+                            <when value="torchvision">
+                                <param name="torchvision_model_name" type="text" format="data" label="TorchVision model name" help="Enter the name of a pre-trained model from TorchVision (see https://docs.pytorch.org/vision/0.21/models.html#general-information-on-pre-trained-weights), e.g. 'alexnet' for classification."/>
```
bernt-matthias: `type="data"`

tools/biapy/biapy.xml
bernt-matthias: If you use `name="model_name"` here and in the torchvision case you can simplify the command section (because the parameter can be accessed in the same way in the parameter dict).

bernt-matthias: I guess you want to add a `no_value` validator in order to require users to enter something here.

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,639 @@
...
+                                <param name="test_gt_avail" type="select" format="txt" label="Do you have test ground truth (target) data?" help="Specify whether ground truth (target data) is available for the test set; if provided, BiaPy will compute performance metrics appropriate to the selected workflow.">
+                                    <option value="test_gt_no" selected="true">No</option>
+                                    <option value="test_gt_yes">Yes</option>
+                                </param>
+                                <when value="test_gt_yes">
+                                    <!-- Q13 -->
+                                    <expand macro="test_gt_param"/>
+                                </when>
+                                <when value="test_gt_no"/>
+                            </conditional>
```
bernt-matthias: Instead of this extra conditional you could make the `test_gt_param` optional. You can use the possibility to parametrize the macro: https://planemo.readthedocs.io/en/latest/writing_advanced.html#parameterizing-xml-macros-with-tokens

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,639 @@
...
+                                <param name="test_gt_avail" type="select" format="txt" label="Do you have test ground truth (target) data?" help="Indicate if ground truth (target data) is available for the test set; if so, BiaPy will compute workflow-specific metrics to evaluate model performance.">
+                                    <option value="test_gt_no" selected="true">No</option>
+                                    <option value="test_gt_yes">Yes</option>
+                                </param>
+                                <when value="test_gt_yes">
+                                    <!-- Q13  -->
+                                    <expand macro="test_gt_param"/>
+                                </when>
+                                <when value="test_gt_no"/>
+                            </conditional>
```
bernt-matthias: Same: make test_gt_param optional?

tools/biapy/biapy.xml
bernt-matthias: Do we know anything about the names, could we add a regexp validator to avoid users entering malicious text.

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,639 @@
...
+        #set $checkpoint_dir = './output/my_experiment/checkpoints'
+        #set $checkpoint_file = $checkpoint_dir + '/checkpoint.pth'
+
+        ## Decide phase and GT availability without touching missing names
+        #set $phase = 'train_test'
+        #set $test_gt_avail = 'test_gt_no'
+
+        #if $mode_selection['selected_mode'] == 'create_new_cfg'
+            #set $phase = $mode_selection['phase_decision']['phases']
+            #if $phase in ['train_test', 'test']
```
bernt-matthias: Looks weired to use `phase` as variable and loop variable?

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,639 @@
...
+                    $mode_selection['phase_decision'].get('test_sec') and
+                    $mode_selection['phase_decision']['test_sec'].get('test_data_avail') and
+                    $mode_selection['phase_decision']['test_sec']['test_data_avail'].get('test_gt_avail')
+                ) or 'test_gt_no'
+            #end if
+        #end if
+
+        ## Define output directory
+        mkdir -p output &&
+        chmod -R 777 output &&
```
bernt-matthias: Why chmod?

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,639 @@
...
+
+        ## Define checkpoint directory in case it is needed
+        mkdir -p '$checkpoint_dir' && 
+        chmod -R 777 '$checkpoint_dir' &&
+
+        ########## Reuse provided yaml file and update paths ##########
+        #if $mode_selection.selected_mode == 'custom_cfg':
```
bernt-matthias, line 62: Maybe you can move the common parameters of the if-else branches to before the if?

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,639 @@
...
+                    elif [ -d "output/my_experiment/results/my_experiment_1/full_image" ]; then
+                        cp output/my_experiment/results/my_experiment_1/full_image/* raw/;
+
+                    ## Detection
+                    elif [ -d "output/my_experiment/results/my_experiment_1/per_image_local_max_check" ]; then
+                        cp output/my_experiment/results/my_experiment_1/per_image_local_max_check/* raw/;
+
+                    ## Detection, Denoising, I2I, SSL, SR
+                    elif [ -d "output/my_experiment/results/my_experiment_1/per_image" ]; then
+                        cp output/my_experiment/results/my_experiment_1/per_image/* raw/;
```
bernt-matthias: Would it be an option to `discover_datasets` from the `output/my_experiment/results/my_experiment_1/` and use recursive? Then we can save all this copying. Otherwise, since you only use h5 and tiff files we might want to copy only those.

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,639 @@
...
+
+                    ## Classification
+                    elif [ -f "output/my_experiment/results/my_experiment_1/predictions.csv" ]; then
```
bernt-matthias, line 217: Since you only use h5 and tiff files this will be ignored.

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,639 @@
...
+                    ## Instance segmentation
+                    elif [ -d "output/my_experiment/results/my_experiment_1/full_image_post_processing" ]; then
+                        cp output/my_experiment/results/my_experiment_1/full_image_post_processing/* post_proc/;
+
+                    ## Detection
+                    elif [ -d "output/my_experiment/results/my_experiment_1/per_image_local_max_check_post_proc" ]; then
+                        cp output/my_experiment/results/my_experiment_1/per_image_local_max_check_post_proc/* post_proc/;
+                    fi;
+                }
+            #end if
```
bernt-matthias: Same question: discover form the longest common prefix using recursive?

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,639 @@
...
+            #if 'tlogs' in $outs
+                && mkdir -p train_logs && chmod -R 777 train_logs && 
+                cp output/my_experiment/train_logs/*.txt train_logs/ 2>/dev/null || true
+            #end if
+        #end if
+
+        #if 'checkpoint' in $outs
+            && mkdir -p checkpoints && chmod -R 777 checkpoints && 
+            cp output/my_experiment/checkpoints/*.pth checkpoints/ 2>/dev/null || true
+        #end if
```
bernt-matthias: Can these copies be saved? Just discover from these paths?

tools/biapy/biapy.xml
danifranco: I put that because I saw it somewhere. If that's covered by default I'm deleting it.

tools/biapy/biapy.xml
danifranco: Thanks for the comments, I thing I have addressed everything I can in the commits that I will push soon. 

Regarding the tests: I have added one more test, although it is quite similar to the second one. At the moment, I am somewhat limited, as I cannot train models on CPU, this would take too much time. Because of this, the tests are restricted to inference only.

In the first test, I demonstrated the branch that allows loading a custom configuration file. The results are not good, since no pretrained weights are being loaded.

In the second test, I showcased the branch for creating a custom configuration file and loading a pretrained model from the BioImage Model Zoo. In this case, the results are good, as the pretrained model I used was originally trained on data similar to my input.

Finally, I added a third test using a different workflow in BiaPy to denoise images. The results here are also not good, again because I am not loading a pretrained model.

The main challenge is that I cannot include a local .pth (model weights) file in the test-data folder, as these files exceed 5MB and would cause the Galaxy tool linting to fail. The only workaround is to configure the tests to load pretrained models directly from the BioImage Model Zoo (BMZ). However, I would prefer not to rely exclusively on this setup, since it depends on the BMZ website, which is not always consistently available and BiaPy has much more to offer than loading always pretrained models from the zoo.

tools/biapy/biapy.xml
danifranco: I saw it in some of the tools and thought it was necessary. I'm removing all of them.

tools/biapy/biapy.xml
danifranco: I have created 'common_yaml_args' variable to store the common two arguments that were repeated. I think nothing else is repeated...

tools/biapy/biapy.xml
danifranco: We cannot really do it that way. Each experiment generates additional tiff/h5 files inside subfolders, which is why I originally proposed using the entire results folder as the output. That approach avoids having to reproduce all this logic and also keeps the configuration much simpler. It is also important to note that, in this case with the current if-else commands, the order in which the data is copied must be preserved.

tools/biapy/biapy.xml
danifranco: same answer as before

tools/biapy/biapy.xml
danifranco: For these ones I think we can apply discover_datasets, yes. I'm pushing all the changes soon.

tools/biapy/biapy.xml
danifranco: I have just made a new release of BiaPy supporting [safetensors](https://github.com/huggingface/safetensors). In the code here I mader changes to always load and generate .safetensors checkpoints. I'm pushing the changes now.

tools/biapy/.shed.yml
```diff
@@ -0,0 +1,11 @@
...
+owner: iuc
+description: Accessible deep learning workflows on bioimages using BiaPy
+homepage_url: https://biapyx.github.io/
+long_description: |
+  This Galaxy wrapper runs BiaPy, a bioimage analysis pipeline designed to simplify training, prediction, and evaluation across tasks like image segmentation, classification, denoising, and more.
+type: unrestricted
+categories:
+  - Machine Learning
+  - Imaging
+remote_repository_url: https://github.com/BiaPyX/BiaPy
```
SaimMomin12:
```suggestion
remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/main/tools/biapy/
```

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,642 @@
...
+        <edam_operation>operation_2945</edam_operation>  <!-- Image segmentation -->
+        <edam_operation>operation_3925</edam_operation>  <!-- Object detection -->
+        <edam_operation>operation_3443</edam_operation>  <!-- Image denoising -->
+        <!-- <edam_operation>Single image super-resolution (not in EDAM)</edam_operation> -->
+        <edam_operation>operation_2946</edam_operation>  <!-- Image restoration -->
+        <!-- <edam_operation>Image-to-image translation (not in EDAM)</edam_operation>   -->
+        <!-- <edam_operation>operation_3442</edam_operation> Image classification -->
+        <!-- <edam_operation>Self-supervision learning (not in EDAM)</edam_operation>  -->
+        <edam_operation>operation_2944</edam_operation>  <!-- Image analysis -->
+    </edam_operations>
```
SaimMomin12, line 20: This can go in macros.xml

tools/biapy/biapy.xml
bernt-matthias: Why should the profile be removed? This is fine IMO.

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,642 @@
+<tool id="biapy" name="Build a workflow with BiaPy" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@" license="MIT">
+    <description>: Accessible deep learning on bioimages</description>
```
bernt-matthias:
```suggestion
    <description>Accessible deep learning on bioimages</description>
```

tools/biapy/biapy.xml
bernt-matthias: > This can go in macros.xml

Both fine. Up to @danifranco

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,642 @@
...
+            <param name="selected_mode" type="select" label="Do you have a configuration file?">
+                <option value="custom_cfg" selected="true">
+                    Yes, I already have one and I want to run BiaPy directly.
+                </option>
+                <option value="create_new_cfg">
+                    No, I want to create one from scratch.
+                </option>
+            </param>
+            <when value="custom_cfg">
+                <param name="config_path" type="data" format="yaml,txt" optional="false" label="Select a configuration file" help="Input configuration file"/>
```
bernt-matthias: Are both formats needed?

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,642 @@
...
+                <param name="biapy_model_path" type="data" format="data" optional="true" label="Select the model checkpoint (if needed)" help="Path to a pre-trained model checkpoint (.safetensors) generated by BiaPy. Use this only if MODEL.LOAD_CHECKPOINT is set to True in your configuration. Checkpoints are saved in the 'checkpoints' folder inside the results directory."/>
```
bernt-matthias: https://github.com/galaxyproject/galaxy/pull/20754/ will be merged soon. So we should:

```suggestion
                <param name="biapy_model_path" type="data" format="safetensors" optional="true" label="Select the model checkpoint (if needed)" help="Path to a pre-trained model checkpoint (.safetensors) generated by BiaPy. Use this only if MODEL.LOAD_CHECKPOINT is set to True in your configuration. Checkpoints are saved in the 'checkpoints' folder inside the results directory."/>
```

bernt-matthias: Can the parameter help be more tailored for Galaxy? 

- "path" and "directory" have no meaning for a Galaxy user
-  Can `if MODEL.LOAD_CHECKPOINT is set to True` be reformulated (in terms of the tool form) such that a Galaxy user knows when this is the case.

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,642 @@
...
+                        <option value="biapy" selected="true">No, I want to build a model from scratch</option>
+                        <option value="biapy_pretrained">Yes, I have a model previously trained in BiaPy</option>
+                        <option value="bmz_torchvision">Yes, I want to check if there is a pre-trained model I can use</option>
+                    </param>
+                    <when value="biapy_pretrained">
+                        <!-- Q4 -->
+                        <param name="biapy_model_path" type="data" format="data" optional="false" label="Select the model trained with BiaPy before" help="Specify the path to a pre-trained BiaPy model checkpoint (.safetensors), stored in the 'checkpoints' folder inside the results directory."/>
```
bernt-matthias: Please already use `safetensors` datatype. 

Analogous to the previous: some terms in the help make little sense to the Galaxy user.

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,642 @@
...
+                    </when>
+                    <when value="bmz_torchvision">
+                        <!-- Q5 -->
+                        <conditional name="bmz_torchvision_model">
+                            <param name="bmz_or_torchvision" type="select" label="Which is the source of the model?" help="Enter the source of the model, whether if it is available through the BioImage Model Zoo or TorchVision">
+                                <option value="bmz" selected="true">BioImage Model Zoo</option>
+                                <option value="torchvision">TorchVision</option>
+                            </param>
+                            <when value="bmz">
+                                <param name="bmz_model_name" type="text" optional="false" value="sensible-cat" label="BioImage Model Zoo model name" help="Enter the name of a pre-trained model from the BioImage Model Zoo (https://bioimage.io/#/models); filter by the BiaPy icon and ensure the model matches your dimensionality (2D/3D) and task (e.g. semantic segmentation).">
```
bernt-matthias, line 350: Just as a note, if https://bioimage.io/#/models would have an API to list model names we could create a select from that... maybe for a next version :)

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,642 @@
...
+                    (mode_selection['selected_mode'] == 'create_new_cfg' and
+                    mode_selection['phase_decision']['phases'] in ['test','train_test'])
+                    or
+                    (mode_selection['selected_mode'] == 'custom_cfg')
+                )
+            ]]></filter>
+        </collection>
+
+        <collection name="predictions_post_proc" type="list" label="Post-processed test predictions">
+            <discover_datasets directory="post_proc" pattern="(?P&lt;name&gt;.+)\.tif"  format="tif" recurse="false" />
```
bernt-matthias: If I remember correctly `designation` is preferred over `name` in the pattern.

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,642 @@
...
+            <option value="metrics">Evaluation metrics (if exist, on test data)</option>
+            <option value="tcharts">Training charts (if exist)</option>
+            <option value="tlogs">Training logs (if exist)</option>
+            <option value="checkpoint">Model checkpoint</option>
+        </param>
+
+    </inputs>
+
+    <outputs>
+        <collection name="predictions_raw" type="list" label="Test predictions">
```
bernt-matthias: By convention output labels should be prefixed with `${tool.name} on ${on_string}: `. Please add this for all data and collection outputs.

bernt-matthias (commented):
Hi @danifranco . Thanks for the effort to add safetensors. I gave the tool a last round of reviews and added some (relatively) minor comments. 

Really looking forward to have this in Galaxy in the near future :)

tools/biapy/biapy.xml
danifranco: I remove this suggestion then

tools/biapy/biapy.xml
danifranco: There is no need to of having .yaml or .txt extension. Any plain text will work as long as the content is in yaml format so I put `.yaml` and `.txt` to cover a more range of cases. What should I put?

tools/biapy/biapy.xml
danifranco: Oh, I see! I have changed the description for these parameters so it is more clear now. This param is defined two times, for the first one (the one you are pointing) makes sense to have this BiaPy key as the user is providing the configuration file. For the second one, and the one most users will follow if they are nort familiarized with BiaPy, I remove it completely and tried to add a more Galaxy tailored description.

tools/biapy/biapy.xml
danifranco: Answered in the above conversation. Hope it is better now.

tools/biapy/biapy.xml
danifranco: Yeah, you are right. We can elavorate this a bit further in the next versions

tools/biapy/biapy.xml
danifranco: Great, I've just changed in all the regex

danifranco:
I think I have addressed all the comments but the tool is not passing all checks because safetensor datatype is not being merged. So, I assume this will not be merged until that is merged too?

tools/biapy/biapy.xml
bernt-matthias: IMO: go for `yaml` only

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,642 @@
...
+        <edam_operation>operation_3925</edam_operation>  <!-- Object detection -->
+        <edam_operation>operation_3443</edam_operation>  <!-- Image denoising -->
+        <!-- <edam_operation>Single image super-resolution (not in EDAM)</edam_operation> -->
+        <edam_operation>operation_2946</edam_operation>  <!-- Image restoration -->
+        <!-- <edam_operation>Image-to-image translation (not in EDAM)</edam_operation>   -->
+        <!-- <edam_operation>operation_3442</edam_operation> Image classification -->
+        <!-- <edam_operation>Self-supervision learning (not in EDAM)</edam_operation>  -->
+        <edam_operation>operation_2944</edam_operation>  <!-- Image analysis -->
+    </edam_operations>
+    <expand macro="requirements" />
```
bgruening, line 21:
```suggestion
    <expand macro="requirements" />
    <required_files>
        <include path="create_yaml.py" />
    </required_files>
```

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,642 @@
...
+            #end if
+        #end if
+
+        ########## Run BiaPy ##########
+        biapy 
+            --config '$generated_cfg'
+            --result_dir './output'
+            --name 'my_experiment'
+            --run_id 1
+            #if '$BIAPY_GPU_ENABLED' == 'true':
```
bgruening: This variable is not used afaik. And if its true why is it set to 0?

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,642 @@
...
+                --gpu '0'
+            #end if
```
bgruening:
```suggestion
            --gpu \${GALAXY_GPU:-0}
```
Will this work. So by default its always 0 - disabled, but if we have this ENV var set we can set it to 1 or 2?

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,642 @@
...
+        #if $selected_phase in ['train_test', 'test']:
+
+            #if 'raw' in $outs
+                ######## 
+                ## RAW #
+                ########
+                && mkdir -p raw && {
+                    ## Instance segmentation
+                    if [ -d "output/my_experiment/results/my_experiment_1/per_image_instances" ]; then
+                        cp output/my_experiment/results/my_experiment_1/per_image_instances/* raw/;
```
bgruening: is it possible to switch to "mv" here? its on the same harddrive most of the time, so mv should be faster.

bgruening (commented):
Cool, sorry that it took a while.
We also integrated safetensors here and it now gets merged: 
https://github.com/galaxyproject/galaxy/pull/20754

So I think we are all set. I just had a few last remarks.

Thanks a lot @danifranco

tools/biapy/biapy.xml
danifranco: GPU number is expected there, so a `0` means that the GPU number 0 will be used. To not use any GPU it must be blank, e.g. `--gpu ""` or not add the `--gpu` option directly. Maybe we cab change the if condition to: `if '$BIAPY_GPU_ENABLED' != '':` . What do you think?

bgruening: The #if condition refers to the template language used by Galaxy to create the command line ... so this is evaluated during submission time, NOT runtime - its not a bash "if".

I will make a new suggestion.

tools/biapy/biapy.xml
```diff
@@ -0,0 +1,645 @@
...
+
+        ########## Run BiaPy ##########
+        biapy 
+            --config '$generated_cfg'
+            --result_dir './output'
+            --name 'my_experiment'
+            --run_id 1
+            #if '$BIAPY_GPU_ENABLED' == 'true':
+                --gpu '0'
+            #end if
```
bgruening:
```suggestion
           \${GALAXY_BIAPY_GPU_STRING:-""}
```

What about this. Now the Galaxy admin and Galaxy can create dynamically a bash env var that contains `--gpu 0`

bgruening: If this env does not exists the string is empty. So no --gpu param et all.

tools/biapy/biapy.xml
danifranco: yeah, of course. I've just done it!

tools/biapy/biapy.xml
danifranco: It was crashing so I've just the upload this line with: `--gpu \${GALAXY_BIAPY_GPU_STRING:-""}` so it is empty when nothing is set. Whenever you want to use a GPU you can then set `GALAXY_BIAPY_GPU_STRING` with an integer there (or multiple ints separated by comma e.g. `"0,1,2"`)

bgruening (approved):
Cool! Very nice first version! Lets get this deployed :)

If you feel like it: https://galaxyproject.eu/posts/2020/08/22/three-steps-to-galaxify-your-tool/

Step 3 :)

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/18356256831

bgruening:
This is now deployed on the ToolShed, our Appstore. Thanks again @danifranco - great work!

danifranco:
Thank you so much for the help!

PR #7173, opened 2025-07-25
Add Galaxy wrapper for LEMUR: Latent Embedding Multivariate Regression

dianichj:
FOR CONTRIBUTOR:

This PR adds a new Galaxy tool for the [LEMUR](https://bioconductor.org/packages/3.20/bioc/html/lemur.html) R package, designed for fitting a latent embedding multivariate regression model to multi-condition single-cell data.

### 🧬 Tool purpose

LEMUR provides a parametric framework to:
- Align multi-condition single-cell transcriptomics data
- Predict log fold changes between conditions at single-cell resolution
- Identify spatially coherent neighborhoods of cells with consistent transcriptional shifts
- Perform pseudobulk differential expression testing on these neighborhoods

It is especially suited for complex experimental designs such as treatment vs. control, time-course, or disease progression studies.

### 📂 Tool contents

- `lemur.xml`: Galaxy wrapper with required parameters and outputs
- `lemur.R`: R script that runs the LEMUR pipeline (fit, align, test)
- `test-data/`: Includes example RDS input and expected PDF/TSV outputs:
  - UMAPs and volcano plots
  - DE results and neighborhood summaries

### 🧪 Test data

The test data included are derived from a publicly available glioblastoma single-cell dataset. They demonstrate:
- Successful model fitting
- Harmony-based alignment
- Visualization of DE results per condition

### 🛠️ Requirements

Conda packages required:
- `bioconductor-lemur`
- `bioconductor-singlecellexperiment`
- `r-optparse`
- `r-tidyverse`
- `r-uwot`

These are defined in the wrapper's `<requirements>` section.

### 📌 Notes

Please help me improve this tool wrapper, it still needs work. Thanks so much for reviewing and for your help!

cc: @nilchia 🚀🖖


nilchia:
Cool :tada:

tools/lemur/.gitignore
```diff
@@ -0,0 +1,2 @@
+.RData
+.Rhistory
```
nilchia: I think it is not needed

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,106 @@
...
+        <edam_operation>operation_3659</edam_operation>
+        <edam_operation>operation_3560</edam_operation>
+    </edam_operations>
+
+    <requirements>
+        <requirement type="package" version="1.4.0">bioconductor-lemur</requirement>
+        <requirement type="package" version="1.7.5">r-optparse</requirement>
+        <requirement type="package" version="2.0.0">r-tidyverse</requirement>
+        <requirement type="package" version="1.28.0">bioconductor-singlecellexperiment</requirement>
+        <requirement type="package" version="0.1.16">r-uwot</requirement>
```
nilchia: Please also add ggplot here.

tools/lemur/lemur.R
```diff
@@ -0,0 +1,145 @@
...
+
+# ---- Load ----
+cat("Reading input:", opt$input_rds, "\n")
+sce <- readRDS(opt$input_rds)
+
+design_formula <- as.formula(opt$design)
+fit <- lemur(sce, design = design_formula,
+             n_embedding = opt$n_embedding,
+             test_fraction = opt$test_fraction)
+fit <- align_harmony(fit)
```
nilchia: Probably this should be optional

tools/lemur/lemur.R
```diff
@@ -0,0 +1,145 @@
...
+
+# ---- UMAP ----
+umap <- uwot::umap(t(fit$embedding))
+umap_df <- as_tibble(fit$colData) |>
+  mutate(UMAP1 = umap[,1], UMAP2 = umap[,2])
+
+p_umap <- ggplot(umap_df, aes(x = UMAP1, y = UMAP2)) +
+  geom_point(aes(color = patient_id, shape = condition), size = 0.5) +
+  coord_fixed() + theme_minimal()
+ggsave(opt$output_umap, p_umap)
```
nilchia: It would be nice to also have multiple formats included here. jpg, png,...

nilchia: also for other plots as well

tools/lemur/lemur.R
```diff
@@ -0,0 +1,145 @@
...
+      tum_nei <- find_de_neighborhoods(
+        tumor_fit,
+        group_by = vars(patient_id, condition),
+        verbose = FALSE
+      )
+      write.table(as.data.frame(tum_nei), opt$output_tumor_neigh, sep="\t", quote=F, row.names=F)
+    }
+
+  } else {
+    cat("rowData has no 'chromosome' column. Tumor plots skipped.\n")
```
nilchia: What if it has another annotation, and the user still wants this plot?

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,106 @@
...
+          --test_fraction '$test_fraction'
+          --output_umap 'umap.pdf'
+          --output_volcano 'volcano.pdf'
+          --output_de 'de_results.tsv'
+          --sel_gene '$sel_gene'
+          --output_gene_umap 'gene_umap.pdf'
+          --output_gene_hist 'gene_hist.pdf'
+          --output_chr_scatter 'chr_scatter.pdf'
+          --output_tumor_umap 'tumor_umap.pdf'
+          --output_tumor_neigh 'tumor_neigh.tsv'
```
nilchia: Do you want to make outputs optional? Maybe the user does not want all of them?

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,106 @@
...
+        ]]>
+    </command>
+
+    <inputs>
+        <param name="sce" type="data" format="rds" label="SingleCellExperiment object (RDS)" />
+        <param name="design_formula" type="text" value="~ patient_id + condition" label="Design formula" />
```
nilchia: IMO it is better to use a tabular file here and extract the values from the table, instead of a text param.
something like the design matrix in DESeq2: https://github.com/galaxyproject/tools-iuc/blob/078ade4ee3f8566373ff88bdab4c0c766e62a347/tools/deseq2/deseq2.xml#L131

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,106 @@
...
+        <param name="n_embedding" type="integer" value="15" label="Number of latent dimensions (embedding)" />
```
nilchia:
```suggestion
        <param name="n_embedding" type="integer" min="1" value="15" label="Number of latent dimensions (embedding)" />
```

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,106 @@
...
+        <param name="test_fraction" type="float" value="0.5" min="0.01" max="0.99" label="Fraction of genes used for training" />
+        <param name="contrast_condition" type="text" value="panobinostat" label="Condition for contrast (e.g. treatment)" />
```
nilchia: Same here, a data_column would be a better option here

nilchia: and for reference_condition

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,106 @@
...
+        <param name="reference_condition" type="text" value="ctrl" label="Reference condition (e.g. control)" />
+        <param name="sel_gene" type="text" value="ENSG00000210082" optional="true" label="Gene ID for gene-specific plots (optional)" />
+    </inputs>
+
+    <outputs>
+        <data name="umap_pdf" format="pdf" from_work_dir="umap.pdf" label="UMAP plot" />
```
nilchia:
```suggestion
        <data name="umap_pdf" format="pdf" from_work_dir="umap.pdf" label="${tool.name} on ${on_string}: UMAP plot" />
```

nilchia: So that the user knows which tool was used on which data

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,106 @@
...
+        <data name="gene_umap_pdf" format="pdf" from_work_dir="gene_umap.pdf" label="Gene UMAP plot" />
+        <data name="gene_hist_pdf" format="pdf" from_work_dir="gene_hist.pdf" label="Gene histogram plot" />
+        <data name="chr_scatter_pdf" format="pdf" from_work_dir="chr_scatter.pdf" label="Chromosome scatter plot" />
+        <data name="tumor_umap_pdf" format="pdf" from_work_dir="tumor_umap.pdf" label="Tumor UMAP plot" />
+        <data name="tumor_neigh" format="tabular" from_work_dir="tumor_neigh.tsv" label="Tumor neighborhood DE results" />
+    </outputs>
+
+    <tests>
+        <test>
+            <param name="sce" value="glioblastoma_example_data.rds"/>
```
nilchia: This is more than 1 MB, Please put it on zenodo

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,106 @@
...
+            <param name="design_formula" value="~ patient_id + condition"/>
+            <param name="n_embedding" value="15"/>
+            <param name="test_fraction" value="0.5"/>
+            <param name="contrast_condition" value="panobinostat"/>
+            <param name="reference_condition" value="ctrl"/>
+            <param name="sel_gene" value="ENSG00000210082"/>
+
+            <output name="umap_pdf" file="umap.pdf" compare="sim_size"/>
+            <output name="volcano_pdf" file="volcano.pdf" compare="sim_size"/>
+            <output name="de_results" file="de_results.tsv" compare="contains"/>
```
nilchia: For text outputs please use asserts

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,106 @@
...
+            <output name="tumor_umap_pdf" file="tumor_umap.pdf" compare="sim_size"/>
+            <output name="tumor_neigh" file="tumor_neigh.tsv" compare="contains"/>
+        </test>
+    </tests>
+
+    <help><![CDATA[
+LEMUR identifies transcriptional differences across conditions in latent space using single-cell RNA-seq data.
+
+**Inputs:**
+- A `SingleCellExperiment` object in RDS format.
```
nilchia, line 397: Can you please explain more here about the sce?
As I remember, it required some specific columns in the sce

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,106 @@
...
+- UMAP plot
+- Volcano plot
+- Table of differentially expressed neighborhoods (without neighborhood list-column)
+- Gene UMAP + histogram plots if a gene ID is given
+- Tumor plots if chromosome data and thresholds are provided
+    ]]></help>
+
+    <citations>
+        <citation type="doi">10.1038/s41588-024-01996-0</citation>
+    </citations>
```
nilchia, line 428:
```suggestion
    </citations>
    <creator>
        <organization name="European Galaxy Team" url="https://galaxyproject.org/eu/"/>
        <person givenName="Diana" familyName="Chiang Jurado" email="dianachj@informatik.uni-freiburg.de"/>
    </creator>
```

nilchia (changes requested):
Super nice!!
Thanks @dianichj

nilchia:
Please add a `.shed.yml` file.
https://github.com/dianichj/tools-iuc/blob/main/tools/seurat/.shed.yml

tools/lemur/.gitignore
dianichj: Thanks for your review comments! ❤️ - I thought I erased those files, I will remove them now.

tools/lemur/lemur.xml
dianichj: I did my best to address this comment lol! I hope I did well. I will update the PR in a bit

tools/lemur/.shed.yml
```diff
@@ -0,0 +1,35 @@
+name: lemur
+owner: iuc
+description: "LEMUR - Latent modeling and DE analysis for single-cell data"
+homepage_url: https://doi.org/10.1038/s41588-024-01996-0
```
nilchia: I think this should be a GitHub link

tools/lemur/lemur.R
```diff
@@ -9,10 +9,19 @@ suppressPackageStartupMessages({
   library(ggplot2)
 })
 
+#----- Function to save plots in different formats ----
+save_plot <- function(filename, plot, format = "pdf", width = 6, height = 5, dpi = 300) {
```
nilchia: Please make the dpi, width, and height configurable.

tools/lemur/lemur.xml
```diff
@@ -1,96 +1,161 @@
-<tool id="lemur" name="LEMUR" version="1.0.1+galaxy0" profile="25.0" license="MIT">
+<tool id="lemur" name="LEMUR" version="1.0.1+galaxy1" profile="25.0" license="MIT">
```
nilchia: No need to bump version now.
```suggestion
<tool id="lemur" name="LEMUR" version="1.0.1+galaxy0" profile="25.0" license="MIT">
```

tools/lemur/lemur.xml
```diff
@@ -1,96 +1,161 @@
...
-        <param name="reference_condition" type="text" value="ctrl" label="Reference condition (e.g. control)" />
-        <param name="sel_gene" type="text" value="ENSG00000210082" optional="true" label="Gene ID for gene-specific plots (optional)" />
-    </inputs>
+<inputs>
+    <param name="sce" type="data" format="rds" label="SingleCellExperiment object (RDS)" />
+    <param name="meta_table" type="data" format="tabular" label="Sample metadata table (TSV)" help="TSV file with one row per sample/cell. Must contain columns for condition and optionally for batch." />
+    <param name="condition_column" type="data_column" data_ref="meta_table" label="Condition column" help="Select the condition column (e.g., treatment vs control). Only appears after loading metadata." />
+    <param name="batch_column" type="data_column" data_ref="meta_table" optional="true" label="Batch column (optional)" help="Optional batch variable (e.g., patient ID). Only appears after loading metadata." />
+    <param name="contrast_condition" type="text" value="panobinostat" label="Condition for contrast (e.g. treatment)" />
```
nilchia: I'm not sure if it is OK here to have a default value.
Maybe at least keep it as "treatment" please.

nilchia: And please add validators for both text params.

tools/lemur/lemur.R
```diff
@@ -9,10 +9,19 @@ suppressPackageStartupMessages({
...
+  ext <- tolower(format)
```
nilchia: Your formats are already lowercase, no?

tools/lemur/lemur.xml
```diff
@@ -1,96 +1,161 @@
...
+                <when input="plot_format" value="png" format="png"/>
+                <when input="plot_format" value="jpg" format="jpg"/>
+            </change_format>
+        </data>
         <data name="de_results" format="tabular" from_work_dir="de_results.tsv" label="LEMUR DE results" />
-        <data name="gene_umap_pdf" format="pdf" from_work_dir="gene_umap.pdf" label="Gene UMAP plot" />
-        <data name="gene_hist_pdf" format="pdf" from_work_dir="gene_hist.pdf" label="Gene histogram plot" />
-        <data name="chr_scatter_pdf" format="pdf" from_work_dir="chr_scatter.pdf" label="Chromosome scatter plot" />
-        <data name="tumor_umap_pdf" format="pdf" from_work_dir="tumor_umap.pdf" label="Tumor UMAP plot" />
+        <data name="gene_umap" format="pdf" from_work_dir="gene_umap.pdf" label="Gene UMAP plot">
```
nilchia: Will this work? You are hardcoding it to look for gene_umap.pdf file.
Please add a test using other formats as output.

tools/lemur/lemur.xml
```diff
@@ -1,96 +1,161 @@
...
+                <when input="plot_format" value="jpg" format="jpg"/>
+            </change_format>
+        </data>
+        <data name="volcano_plot" format="pdf" from_work_dir="volcano.pdf" label="Volcano plot">
+            <change_format>
+                <when input="plot_format" value="png" format="png"/>
+                <when input="plot_format" value="jpg" format="jpg"/>
+            </change_format>
+        </data>
         <data name="de_results" format="tabular" from_work_dir="de_results.tsv" label="LEMUR DE results" />
```
nilchia: Please use this lable for all outputs:
${tool.name} on ${on_string}:

tools/lemur/lemur.xml
```diff
@@ -1,96 +1,161 @@
...
-            <output name="tumor_umap_pdf" file="tumor_umap.pdf" compare="sim_size"/>
-            <output name="tumor_neigh" file="tumor_neigh.tsv" compare="contains"/>
+            <param name="plot_format" value="pdf"/>
+            <param name="tumor_annotation_column" value="chromosome"/>
+    
+            <output name="de_results" file="https://zenodo.org/record/16575009/files/de_results.tsv?download=1">
+                <assert_contents>
+                    <has_line_matching expression="^name\tn_cells\t.*did_lfc$"/>
+                    <has_text text="ENSG00000210082"/>
+                </assert_contents>
```
nilchia, line 170: Please also add `has_n_lines` so we know the file has correct number of genes

tools/lemur/lemur.xml
```diff
@@ -1,96 +1,161 @@
...
 - Parameters for dimensionality and model training.
 - Optionally, a gene ID for extra gene-specific plots.
+- Apply Harmony alignment: If selected, the latent space is aligned using Harmony before differential expression analysis.
+- Selectable output format for all plots: PDF, PNG, or JPG.
+
+**Why use `SingleCellExperiment` (SCE)?**  
+LEMUR uses the `SingleCellExperiment` (SCE) format because it is the standard Bioconductor structure for single-cell data in R. Unlike Seurat or AnnData, SCE is lightweight, interoperable, and not tied to a specific framework. It cleanly separates assays (e.g., expression values), cell metadata (`colData`), feature metadata (`rowData`), and reduced dimensions (`reducedDims`).
+
+For LEMUR to work correctly, the SCE object **must include**:
+- **`logcounts()`**: matrix of log-normalized gene expression.
```
nilchia: It also needs `counts`

nilchia (changes requested):

nilchia:
Categories [['Statistical Analysis']] unknown.

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,171 @@
...
+                <when input="plot_format" value="png" format="png"/>
+                <when input="plot_format" value="jpg" format="jpg"/>
+            </change_format>
+        </data>
+        <data name="tumor_neigh" format="tabular" from_work_dir="tumor_neigh.tsv" label="Tumor neighborhood DE results" />
+    </outputs>
+
+    <tests>
+        <test>
+            <param name="sce" value="https://zenodo.org/record/16575009/files/glioblastoma_example_data.rds?download=1"/>
```
nilchia:
```suggestion
            <param name="sce" value="https://zenodo.org/record/16575009/files/glioblastoma_example_data.rds"/>
```

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,171 @@
...
+            <param name="batch_column" value="1"/>
+            <param name="contrast_condition" value="panobinostat"/>
+            <param name="reference_condition" value="ctrl"/>
+            <param name="n_embedding" value="15"/>
+            <param name="test_fraction" value="0.5"/>
+            <param name="sel_gene" value="ENSG00000210082"/>
+            <param name="plot_format" value="pdf"/>
+            <param name="tumor_annotation_column" value="chromosome"/>
+    
+            <output name="de_results" file="https://zenodo.org/record/16575009/files/de_results.tsv?download=1">
```
nilchia:
```suggestion
            <output name="de_results" file="https://zenodo.org/record/16575009/files/de_results.tsv">
```

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,171 @@
...
+                <assert_contents>
+                    <has_line_matching expression="^name\tn_cells\t.*did_lfc$"/>
+                    <has_text text="ENSG00000210082"/>
+                </assert_contents>
+            </output>
+    
+            <output name="tumor_neigh" file="https://zenodo.org/record/16575009/files/tumor_neigh.tsv?download=1">
```
nilchia:
```suggestion
            <output name="tumor_neigh" file="https://zenodo.org/record/16575009/files/tumor_neigh.tsv">
```

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,171 @@
...
+                    <has_text text="ENSG00000210082"/>
+                </assert_contents>
+            </output>
+    
+            <output name="umap_plot" file="https://zenodo.org/record/16575009/files/umap.pdf?download=1" compare="sim_size"/>
+            <output name="volcano_plot" file="https://zenodo.org/record/16575009/files/volcano.pdf?download=1" compare="sim_size"/>
+            <output name="gene_umap" file="https://zenodo.org/record/16575009/files/gene_umap.pdf?download=1" compare="sim_size"/>
+            <output name="gene_hist" file="https://zenodo.org/record/16575009/files/gene_hist.pdf?download=1" compare="sim_size"/>
+            <output name="chr_scatter" file="https://zenodo.org/record/16575009/files/chr_scatter.pdf?download=1" compare="sim_size"/>
+            <output name="tumor_umap" file="https://zenodo.org/record/16575009/files/tumor_umap.pdf?download=1" compare="sim_size"/>
```
nilchia:
```suggestion
            <output name="umap_plot" file="https://zenodo.org/record/16575009/files/umap.pdf" compare="sim_size"/>
            <output name="volcano_plot" file="https://zenodo.org/record/16575009/files/volcano.pdf" compare="sim_size"/>
            <output name="gene_umap" file="https://zenodo.org/record/16575009/files/gene_umap.pdf" compare="sim_size"/>
            <output name="gene_hist" file="https://zenodo.org/record/16575009/files/gene_hist.pdf" compare="sim_size"/>
            <output name="chr_scatter" file="https://zenodo.org/record/16575009/files/chr_scatter.pdf" compare="sim_size"/>
            <output name="tumor_umap" file="https://zenodo.org/record/16575009/files/tumor_umap.pdf" compare="sim_size"/>
```

nilchia (changes requested):

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,171 @@
...
+                <when input="plot_format" value="png" format="png"/>
+                <when input="plot_format" value="jpg" format="jpg"/>
+            </change_format>
+        </data>
+        <data name="tumor_neigh" format="tabular" from_work_dir="tumor_neigh.tsv" label="Tumor neighborhood DE results" />
+    </outputs>
+
+    <tests>
+        <test>
+            <param name="sce" value="https://zenodo.org/record/16575009/files/glioblastoma_example_data.rds"/>
```
nilchia:
```suggestion
            <param name="sce" location="https://zenodo.org/record/16575009/files/glioblastoma_example_data.rds"/>
```

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,171 @@
...
+            <param name="tumor_annotation_column" value="chromosome"/>
+    
+            <output name="de_results" file="https://zenodo.org/record/16575009/files/de_results.tsv">
+                <assert_contents>
+                    <has_line_matching expression="^name\tn_cells\t.*did_lfc$"/>
+                    <has_text text="ENSG00000210082"/>
+                </assert_contents>
+            </output>
+    
+            <output name="tumor_neigh" file="https://zenodo.org/record/16575009/files/tumor_neigh.tsv">
```
nilchia:
```suggestion
            <output name="tumor_neigh" location="https://zenodo.org/record/16575009/files/tumor_neigh.tsv">
```

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,171 @@
...
+                    <has_text text="ENSG00000210082"/>
+                </assert_contents>
+            </output>
+    
+            <output name="umap_plot" file="https://zenodo.org/record/16575009/files/umap.pdf" compare="sim_size"/>
+            <output name="volcano_plot" file="https://zenodo.org/record/16575009/files/volcano.pdf" compare="sim_size"/>
+            <output name="gene_umap" file="https://zenodo.org/record/16575009/files/gene_umap.pdf" compare="sim_size"/>
+            <output name="gene_hist" file="https://zenodo.org/record/16575009/files/gene_hist.pdf" compare="sim_size"/>
+            <output name="chr_scatter" file="https://zenodo.org/record/16575009/files/chr_scatter.pdf" compare="sim_size"/>
+            <output name="tumor_umap" file="https://zenodo.org/record/16575009/files/tumor_umap.pdf" compare="sim_size"/>
```
nilchia:
```suggestion
            <output name="umap_plot" location="https://zenodo.org/record/16575009/files/umap.pdf" compare="sim_size"/>
            <output name="volcano_plot" location="https://zenodo.org/record/16575009/files/volcano.pdf" compare="sim_size"/>
            <output name="gene_umap" location="https://zenodo.org/record/16575009/files/gene_umap.pdf" compare="sim_size"/>
            <output name="gene_hist" location="https://zenodo.org/record/16575009/files/gene_hist.pdf" compare="sim_size"/>
            <output name="chr_scatter" location="https://zenodo.org/record/16575009/files/chr_scatter.pdf" compare="sim_size"/>
            <output name="tumor_umap" location="https://zenodo.org/record/16575009/files/tumor_umap.pdf" compare="sim_size"/>
```

nilchia (changes requested):

dianichj:
Ping @nilchia :)!

tools/lemur/.shed.yml
```diff
@@ -0,0 +1,35 @@
...
+  LEMUR is an R package that identifies transcriptional differences across conditions 
+  in latent space using single-cell RNA-seq data. It enables users to train 
+  interpretable latent variable models and perform differential expression analysis
+  across conditions, patients, or neighborhoods in the latent space.
+remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/master/tools/lemur
+type: unrestricted
+categories:
+- Single Cell
+- Transcriptomics
+- Statistical Analysis
```
RZ9082:
```suggestion
- Statistics
```

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,271 @@
+<tool id="lemur" name="LEMUR" version="1.0.1+galaxy0" profile="25.0" license="MIT">
```
bgruening: The version should be 1.4.0+galaxy0 I think to indicate the version for the main dependency.

For that you can use a macro

tools/lemur/lemur.xml
bgruening: ping for the validator

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,271 @@
...
+<inputs>
+    <param name="sce" type="data" format="rds" label="SingleCellExperiment object (RDS)" />
+    <param name="meta_table" type="data" format="tabular" label="Sample metadata table (TSV)" help="TSV file with one row per sample/cell. Must contain columns for condition and optionally for batch." />
+
+    <param name="cell_id_column" type="data_column" data_ref="meta_table" label="Cell ID column" help="Column in metadata that contains the cell IDs matching colnames of SCE" />
+    <param name="condition_column" type="data_column" data_ref="meta_table" label="Condition column" help="Select the condition column (e.g., treatment vs control). Only appears after loading metadata." />
+    <param name="batch_column" type="data_column" data_ref="meta_table" optional="true" label="Batch column (optional)" help="Optional batch variable (e.g., patient ID). Only appears after loading metadata." />
+
+    <param name="contrast_condition" type="text" value="panobinostat" label="Condition for contrast (e.g. treatment)" />
+    <param name="reference_condition" type="text" value="ctrl" label="Reference condition (e.g. control)" />
```
bgruening: a validator would be nice here as well

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,271 @@
...
+        <option value="jpg">JPG</option>
+    </param>
+
+    <param name="plot_width" type="float" value="6" label="Plot width (in inches)" optional="true"/>
+    <param name="plot_height" type="float" value="5" label="Plot height (in inches)" optional="true"/>
+
+    <param name="tumor_annotation_column" type="text" value="chromosome"
+           label="Tumor annotation column in rowData"
+           help="Used for tumor classification. Default is 'chromosome'. Override if your SCE uses a different annotation column." />
+</inputs>
```
bgruening: indentation is a bit off here, please use the galaxy-language-server to reformat this tool

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,271 @@
...
+
+<outputs>
```
bgruening: these are a lot of outputs, are all of them always needed? Should the user be able to select only a few of them?

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,271 @@
...
+  --batch_column '$batch_column'
+  #end if
+  --contrast_condition '$contrast_condition'
+  --reference_condition '$reference_condition'
+  --n_embedding '$n_embedding'
+  --test_fraction '$test_fraction'
+  --output_umap 'umap.$plot_format'
+  --output_volcano 'volcano.$plot_format'
+  --output_de 'de_results.tsv'
+  #if str($sel_gene):
```
bgruening: "#end if" is missing for some reason.

dianichj:
Hola @nilchia! I made some good progress on the PR since the last round of comments (it's been a while haha).
Quick update in case it's useful. I fixed a bunch of bugs that were breaking the command line and R script based  (a missing #end if, a harmony version mismatch, and a few others), and addressed the comments from yours and bgruenings feeback. The version now uses a macro (1.4.0+galaxy0), added validators, standard output labels, and so on.

I also made the chr7/chr10 tumor classification optional (a new run_tumor_analysis param, off by default) instead of baking it into the tool, since it's specific to the glioblastoma example rather than the core LEMUR method (to address @RZ9082 concern on this tool being too specific to the tutorial dataset). The tool should now work generically for any tissue or cell type!

Also fixed the SCE help text, turns out it needs both counts() and logcounts(), not just logcounts() as you suggested. All 3 tests (PDF, PNG, JPG) pass now.

Would love  your eyes on it and discuss whenever you get a chance, no rush 🖖😊

tools/lemur/.shed.yml
```diff
@@ -0,0 +1,35 @@
+name: lemur
+owner: iuc
+description: "LEMUR - Latent modeling and DE analysis for single-cell data"
```
nilchia, line 3:
```suggestion
description: "LEMUR - Latent Embedding Multivariate Regression"
```

tools/lemur/.shed.yml
```diff
@@ -0,0 +1,35 @@
...
+categories:
+- Single Cell
+- Transcriptomics
+- Statistics
+auto_tool_repositories:
+  name_template: "{{ tool_id }}"
+  description_template: "Wrapper for the LEMUR tool: {{ tool_name }}"
+suite:
+  name: "suite_lemur"
+  description: "LEMUR - Latent modeling and DE analysis for single-cell data"
```
nilchia, line 21:
```suggestion
  description: "LEMUR - Latent Embedding Multivariate Regression"
```

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+        --chrom1_name '$tumor_analysis_cond.chrom1_name'
+        --chrom2_name '$tumor_analysis_cond.chrom2_name'
+        --chrom1_thresh '$tumor_analysis_cond.chrom1_thresh'
+        --chrom2_thresh '$tumor_analysis_cond.chrom2_thresh'
+    #end if
+    ]]></command>
+
+    <inputs>
+        <param name="sce" type="data" format="rds" label="SingleCellExperiment object (RDS)" />
+        <param name="meta_table" type="data" format="tabular" label="Sample metadata table (TSV)" help="TSV file with one row per sample/cell. Must contain columns for condition and optionally for batch." />
```
nilchia, line 50: Why does it need to have a separate tabular data for the sample annotation?
Isn't it included in the SCE object as `colData`?

nilchia: Haha, I see a wise man once suggested this to be a table :)
https://github.com/galaxyproject/tools-iuc/pull/7173#discussion_r2231500579

nilchia, line 50:
```suggestion
        <param name="meta_table" type="data" format="tabular" label="Sample metadata table (TSV)" help="Tabular file with one row per sample/cell. Must contain columns for condition and optionally for batch or covariates." />
```

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+
+        <param name="cell_id_column" type="data_column" data_ref="meta_table" label="Cell ID column" help="Column in metadata that contains the cell IDs matching colnames of SCE" />
+        <param name="condition_column" type="data_column" data_ref="meta_table" label="Condition column" help="Select the condition column (e.g., treatment vs control). Only appears after loading metadata." />
```
nilchia, line 53: What do you mean by: `Only appears after loading metadata.`

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+        <param name="batch_column" type="data_column" data_ref="meta_table" optional="true" label="Batch column (optional)" help="Optional batch variable (e.g., patient ID). Only appears after loading metadata." />
```
nilchia, line 54: This should be `multiple="true"` to take into account the covariates as well.

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+
+        <param name="contrast_condition" type="text" value="treatment" label="Condition for contrast (e.g. treatment)">
+            <validator type="empty_field" message="This field cannot be empty."/>
+        </param>
+        <param name="reference_condition" type="text" value="ctrl" label="Reference condition (e.g. control)">
+            <validator type="empty_field" message="This field cannot be empty."/>
+        </param>
+
+        <param name="n_embedding" type="integer" min="1" value="15" label="Number of latent dimensions (embedding)" />
```
nilchia, line 63: This is an argument from the lemur function itself. I would suggest using `argument="n_embedding"` instead of `name="n_embedding"` so it shows the `n_embedding` in the interface

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+        <param name="test_fraction" type="float" value="0.5" min="0.01" max="0.99" label="Fraction of genes used for training" />
```
nilchia, line 64:
```suggestion
        <param name="test_fraction" type="float" value="0.2" min="0.01" max="1.0" label="The fraction of cells that are split off before the model fit to keep an independent set of test observations" />
```

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+
+        <param name="use_harmony" type="boolean" truevalue="yes" falsevalue="no" checked="true" label="Apply Harmony alignment?" />
```
nilchia, line 66:
```suggestion
        <param name="use_harmony" type="boolean" truevalue="yes" falsevalue="no" checked="true" label="Apply Harmony alignment?" help="Enforces additional alignment of cell clusters beyond the direct differential embedding." />
```

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+        <param name="sel_gene" type="text" optional="true" label="Gene ID for gene-specific plots (optional)" />
```
nilchia, line 67: I think it should match row names of the SCE, right?
Please add a help for it.

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+
+        <param name="plot_format" type="select" label="Plot format">
+            <option value="pdf" selected="true">PDF</option>
+            <option value="png">PNG</option>
+            <option value="jpg">JPG</option>
+        </param>
+
+        <param name="plot_width" type="float" value="6" label="Plot width (in inches)" optional="true"/>
```
nilchia, line 75: Isn't it better to use a metric system? Maybe "cm"?

nilchia, line 75:
```suggestion
        <param name="plot_width" type="float" value="6" label="Plot width (in inches)"/>
```

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+        <param name="plot_height" type="float" value="5" label="Plot height (in inches)" optional="true"/>
```
nilchia, line 76:
```suggestion
        <param name="plot_height" type="float" value="5" label="Plot height (in inches)"/>
```

tools/lemur/lemur.xml
nilchia: Please also set `min` and `max` for height and width

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+
+        <conditional name="tumor_analysis_cond">
+            <param name="run_tumor_analysis" type="select"
+                label="Run tumor-specific classification (chromosome-based)"
+                help="Optional: classifies cells as tumor/non-tumor based on chromosome gain/loss (as in the glioblastoma example). Not applicable to all datasets — enable only if relevant to your data.">
```
nilchia, line 81: Please run `planemo format` in your directory to fix the formatting in your tool.

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+                <option value="no" selected="true">No</option>
+                <option value="yes">Yes</option>
+            </param>
+            <when value="yes">
+                <param name="tumor_annotation_column" type="text" value="chromosome"
+                    label="Tumor annotation column in rowData"
```
nilchia, line 87: This is gene annotation, no?

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+                    help="Used for tumor classification. Default is 'chromosome'. Override if your SCE uses a different annotation column." />
+                <param name="chrom1_name" type="text" value="7"
+                    label="Chromosome expected to be gained in tumor cells"
+                    help="Name/ID of the chromosome expected to show increased expression ('gain') in tumor cells, as it appears in the tumor annotation column above. This whole-chromosome gain/loss classification is specific to cancer types with known chromosome-arm-level copy number alterations (e.g. glioblastoma, default: chromosome 7 gain) — confirm this pattern applies to your cancer type/dataset before using it.">
```
nilchia, line 91: I think it is not just for cancer. It might be applicable to some other genetic diseases as well.

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+            <param name="run_tumor_analysis" type="select"
+                label="Run tumor-specific classification (chromosome-based)"
+                help="Optional: classifies cells as tumor/non-tumor based on chromosome gain/loss (as in the glioblastoma example). Not applicable to all datasets — enable only if relevant to your data.">
+                <option value="no" selected="true">No</option>
+                <option value="yes">Yes</option>
+            </param>
+            <when value="yes">
+                <param name="tumor_annotation_column" type="text" value="chromosome"
+                    label="Tumor annotation column in rowData"
+                    help="Used for tumor classification. Default is 'chromosome'. Override if your SCE uses a different annotation column." />
```
nilchia, line 88: Please use the same style as other sample annotations like "batch".
I mean either all should be a `data_column` or a `text` param

tools/lemur/lemur.xml
nilchia: For this and the next param you can use macros.

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+            <when value="no"/>
+        </conditional>
+    </inputs>
+
+    <outputs>
+        <data name="umap_plot" format="pdf" from_work_dir="umap.$plot_format" label="${tool.name} on ${on_string}: UMAP plot">
+            <change_format>
+                <when input="plot_format" value="png" format="png"/>
+                <when input="plot_format" value="jpg" format="jpg"/>
+            </change_format>
```
nilchia, line 115: this can be in macros

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+        </data>
+        <data name="tumor_neigh" format="tabular" from_work_dir="tumor_neigh.tsv" label="${tool.name} on ${on_string}: Tumor neighborhood DE results">
+            <filter>tumor_analysis_cond['run_tumor_analysis'] == 'yes'</filter>
+        </data>
+    </outputs>
+
+    <tests>
+        <!-- Test 1: PDF -->
+        <test expect_num_outputs="8">
+            <param name="sce" location="https://zenodo.org/records/16575009/files/glioblastoma_example_data.rds"/>
```
nilchia, line 160: Please add your Zenodo repo to the Galaxy tool dev community on Zenodo

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+                    <has_n_lines n="15878" delta="2000"/>
+                    <has_n_columns n="13"/>
+                    <has_line_matching expression="^name\tneighborhood\t.*did_lfc$"/>
+                    <has_text text="ENSG00000210082"/>
+                </assert_contents>
+            </output>
+
+            <output name="umap_plot" ftype="pdf">
+                <assert_contents>
+                    <has_size min="0"/>
```
nilchia, line 195: Maybe also use `value`?

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+        </test>
+    </tests>
+
+
+    <help><![CDATA[
+LEMUR identifies transcriptional differences across conditions in latent space using single-cell RNA-seq data. This tool wraps the core LEMUR workflow (latent embedding multivariate regression) and can be applied to single-cell data from any tissue or cell type — it is not limited to any specific disease or dataset.
+
+**Inputs:**
+- A `SingleCellExperiment` object in RDS format.
+- Design formula indicating batch and condition variables.
```
nilchia, line 366: you have changed this now.

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+
+**Why use `SingleCellExperiment` (SCE)?**
+LEMUR uses the `SingleCellExperiment` (SCE) format because it is the standard Bioconductor structure for single-cell data in R. Unlike Seurat or AnnData, SCE is lightweight, interoperable, and not tied to a specific framework. It cleanly separates assays (e.g., expression values), cell metadata (`colData`), feature metadata (`rowData`), and reduced dimensions (`reducedDims`).
+
+For LEMUR to work correctly, the SCE object **must include**:
+- **`counts()`**: matrix of raw (non-normalized) integer counts, required for the pseudobulk differential expression testing performed by `find_de_neighborhoods()`.
+- **`logcounts()`**: matrix of log-normalized gene expression, used for model fitting and visualization.
+- **`colData`**: must contain the variables referenced in the `design` formula (e.g., `patient_id`, `condition`).
+
+You can easily convert Seurat objects using `as.SingleCellExperiment()` in R.
```
nilchia, line 380: Please refer to other galaxy tools that we have for this conversion. The sceasy and anndata2ri

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+
+**Outputs:**
+- UMAP plot
+- Volcano plot
+- Table of differentially expressed neighborhoods (without neighborhood list-column)
```
nilchia, line 385: (without neighborhood list-column)
what does it mean?

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+- Gene UMAP + histogram plots if a gene ID is given
+
+**Optional tumor-specific classification:**
+Enabling "Run tumor-specific classification" performs a chromosome gain/loss-based tumor/non-tumor cell classification, using two configurable chromosome-arm markers (default: chromosome 7 gain / chromosome 10 loss, as used in the glioblastoma example below). This approach can be applied to other tumor types with known chromosome-level copy number alterations by adjusting the chromosome names and thresholds to match your data — it is disabled by default since it's only relevant if your dataset includes this kind of copy-number-based tumor signal. When enabled, it adds three outputs: a chromosome expression scatter plot, a tumor/non-tumor UMAP plot, and a table of differentially expressed neighborhoods restricted to the classified tumor cells.
+
+This analysis was inspired by the glioblastoma example in the official LEMUR repository, which also has more details on the core method: https://github.com/const-ae/lemur
```
nilchia, line 391: I would suggest removing this sentence.

tools/lemur/macros.xml
```diff
@@ -0,0 +1,15 @@
+<macros>
+    <token name="@TOOL_VERSION@">1.4.0</token>
+    <token name="@VERSION_SUFFIX@">0</token>
```
nilchia, line 3:
```suggestion
    <token name="@VERSION_SUFFIX@">0</token>
    <token name="@PROFILE@">25.0</token>
```

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
+<tool id="lemur" name="LEMUR" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="25.0" license="MIT">
```
nilchia, line 1:
```suggestion
<tool id="lemur" name="LEMUR" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@" license="MIT">
```

tools/lemur/lemur.xml
nilchia, line 75: the width and height are not really optional, you are setting a value for them when `null`.

I would suggest removing the optional.

tools/lemur/lemur.R
```diff
@@ -0,0 +1,213 @@
...
+set.seed(42)
+
+save_plot <- function(filename, plot, format = "pdf", width = NULL, height = NULL, dpi = 300) {
+    ext <- tolower(format)
+    out <- sub("\\.pdf$", paste0(".", ext), filename)
+    if (is.null(width)) width <- 6
+    if (is.null(height)) height <- 5
+
+    ggsave(out, plot, device = ext, width = width, height = height, dpi = dpi, units = "in", bg = "white")
+}
```
nilchia, line 21:
```suggestion
save_plot <- function(filename, plot, format = "pdf", width = 6, height = 5, dpi = 300) {
    ggsave(out, plot, device = ext, width = width, height = height, dpi = dpi, units = "in", bg = "white")
}
```

tools/lemur/lemur.R
```diff
@@ -0,0 +1,213 @@
...
+option_list <- list(
+    make_option(c("--input_rds"), type = "character"),
+    make_option(c("--meta_table"), type = "character"),
+    make_option(c("--cell_id_column"), type = "character"),
+    make_option(c("--condition_column"), type = "character"),
+    make_option(c("--batch_column"), type = "character"),
+    make_option(c("--contrast_condition"), type = "character"),
+    make_option(c("--reference_condition"), type = "character"),
+    make_option(c("--n_embedding"), type = "integer", default = 15),
+    make_option(c("--test_fraction"), type = "double", default = 0.5),
```
nilchia, line 33:
```suggestion
    make_option(c("--test_fraction"), type = "double", default = 0.2),
```

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,401 @@
...
+                <when input="plot_format" value="jpg" format="jpg"/>
+            </change_format>
+            <filter>tumor_analysis_cond['run_tumor_analysis'] == 'yes'</filter>
+        </data>
+        <data name="tumor_neigh" format="tabular" from_work_dir="tumor_neigh.tsv" label="${tool.name} on ${on_string}: Tumor neighborhood DE results">
+            <filter>tumor_analysis_cond['run_tumor_analysis'] == 'yes'</filter>
+        </data>
+    </outputs>
+
+    <tests>
```
nilchia, line 157: Please also add a test without the `batch_column`
and also test to exclude non-optional outputs. You should make sure the filters in the output section work, and you get different `expect_num_outputs`

tools/lemur/lemur.R
```diff
@@ -0,0 +1,213 @@
...
+sce <- readRDS(opt$input_rds)
+
+# ---- Load and prepare metadata ----
+meta <- read.delim(opt$meta_table, sep = "\t", check.names = FALSE)
+
+opt$cell_id_column <- as.integer(opt$cell_id_column)
+opt$condition_column <- as.integer(opt$condition_column)
+opt$batch_column <- if (nzchar(opt$batch_column)) as.integer(opt$batch_column) else NULL
+
+cell_id_colname <- colnames(meta)[opt$cell_id_column]
```
nilchia, line 66: I would suggest adding `cat(cell_id_colname` and then checking the stdout in the test for having correct colname
also for condition and batch

tools/lemur/lemur.R
```diff
@@ -0,0 +1,213 @@
...
+condition_name <- colnames(meta)[opt$condition_column]
+batch_name <- if (!is.null(opt$batch_column)) colnames(meta)[opt$batch_column] else NULL
+
+rownames(meta) <- meta[[cell_id_colname]]
+
+stopifnot(cell_id_colname %in% colnames(meta))
```
nilchia, line 72:
```suggestion
stopifnot(cell_id_colname %in% colnames(colData(sce)))
```

nilchia: There is no way that this check fails. Because the column id is coming from "meta" itself, so the colname should exist there. A proper check is to see if that column exists in the SCE object

tools/lemur/lemur.R
```diff
@@ -0,0 +1,213 @@
...
+stopifnot(condition_name %in% colnames(meta))
+if (!is.null(batch_name)) stopifnot(batch_name %in% colnames(meta))
+
+stopifnot(all(colnames(sce) %in% rownames(meta)))
```
nilchia, line 76:
```suggestion
stopifnot(identical(colnames(sce), rownames(meta)))
```

tools/lemur/lemur.R
```diff
@@ -0,0 +1,213 @@
...
+meta <- meta[!is.na(meta[[condition_name]]), , drop = FALSE]
+meta <- meta[colnames(sce), , drop = FALSE]
```
nilchia, line 78: This part is not needed I think

tools/lemur/lemur.R
```diff
@@ -0,0 +1,213 @@
...
+colData(sce) <- S4Vectors::DataFrame(meta)
+
+# ---- Build design formula ----
+design_formula <- if (!is.null(batch_name)) {
+    as.formula(sprintf("~ %s + %s", batch_name, condition_name))
+} else {
+    as.formula(sprintf("~ %s", condition_name))
+}
```
nilchia, line 86: If the column names are irregular like having space, the formula will fail. Please set check.names=TRUE when reading the meta

tools/lemur/lemur.R
```diff
@@ -0,0 +1,213 @@
...
+
+# ---- Run LEMUR ----
+fit <- lemur(
```
nilchia, line 89:
```
linear_coefficient_estimator
specify which estimator is used to center the conditions. "linear" runs simple
regression it works well in many circumstances but can produce poor results if
the composition of the cell types changes between conditions (e.g., one cell type
disappears). "mean", "cluster_median" and "zero" are alternative estimators,
which are each supposed to be more robust against compositional changes but
cannot account for genes that change for all cells between conditions. "linear"
is the default as it works best with subsequent alignment steps.
```
please include this param as well. It can be a select param.

nilchia: and this:
```
use_assay if data is a SummarizedExperiment / SingleCellExperiment object, which
assay should be used.
```

tools/lemur/lemur.R
```diff
@@ -0,0 +1,213 @@
...
+contrast_expr <- substitute(
+    cond(condition = A) - cond(condition = B),
+    list(
+        condition = as.name(condition_name),
+        A = opt$contrast_condition,
+        B = opt$reference_condition
+    )
+)
+
+print(contrast_expr)
```
nilchia, line 114: please assert this in the test

tools/lemur/lemur.R
nilchia: I see that you are doing it later in the script :)

Please then check it in the test

tools/lemur/lemur.R
```diff
@@ -0,0 +1,213 @@
...
+print(fit)
+
+cat("Condition column:", condition_name, "\n")
+cat("Unique values in the condition column:\n")
+print(unique(colData(fit)[[condition_name]]))
+cat("First values:\n")
+print(head(colData(fit)[[condition_name]]))
+
+fit <- test_de(fit, contrast = !!contrast_expr)
```
nilchia, line 123: please add this param:
```
consider specify which part of the model are considered for the differential expression
test
```

nilchia (commented):
Well done! @dianichj 

I added some minor comments :)

nilchia:
image
Is it possible to have overexpressed as red and low-expressed as blue?

nilchia:
You don't have the `cell_id` column in the SCE object. That's why the test is failing.

tools/lemur/lemur.xml
dianichj: This is a leftover from an earlier draft, it just meant the dropdown populates once a metadata file is selected. I will remove it, it's confusing.

tools/lemur/lemur.xml
dianichj: Yes, it has to match rownames(sce). Adding help text.

tools/lemur/lemur.xml
dianichj: Will switch to cm and convert in ggsave().

tools/lemur/lemur.xml
dianichj: it's gene-level annotation from rowData. I will rename the label.

tools/lemur/lemur.xml
dianichj: The column selectors point at meta_table, which is tabular, so Galaxy can read its header to build the dropdown. tumor_annotation_column points at rowData inside the SCE, that is binary. Galaxy can't enumerate its columns, so this one has to stay text.

tools/lemur/lemur.xml
dianichj: Good catch! Will rename to covariate_columns since it's no longer just batch.

tools/lemur/lemur.xml
dianichj: find_de_neighborhoods() returns a column holding each neighborhood's cell IDs as a nested list, which can't be written to TSV, so it's dropped before export. I will reword it.

tools/lemur/lemur.R
dianichj: Agreed on the issue  but `check.names=TRUE` would rename every column, including the ones the user picked in the selectors. I'll apply `make.names()` only to the formula terms instead.

dianichj:
Hi @nilchia ! It is me again after your review:

I pushed everything. All 4 tests pass locally. I did most of the changes as you suggested, here are some remarks that were done a bit diff. 

1. Your `has_size` comment was the most valuable one! Replacing `min="0"` with real measured sizes revealed that no plot had ever been collected by Galaxy: `from_work_dir` isn't Cheetah-templated, so `from_work_dir="umap.$plot_format"` was looked up literally and every plot came back as a 0-byte dataset. The old assertions hid it with `min="0"` passes on an empty file, and `gene_hist`'s `value="5216" delta="10000"` gives a range that includes 0. Fixed by writing plots to static filenames and letting `change_format` assign the datatype.

2. Your covariates comment turned up something too: `batch_column` is now `covariate_columns` with `multiple="true"`, but the no-covariate test can't pass: with a condition-only design, `find_de_neighborhoods` forms exactly 2 pseudobulk samples for 2 coefficients, so glmGamPoi has no residual degrees of freedom and the run dies inside `handle_design_parameter`. Covariate selection is now required, with an upfront error explaining why. Added a minimal-outputs test instead (no `sel_gene`, tumor analysis off, `expect_num_outputs="3"`) to exercise the output filters.

3. One more bug: `tumor_neigh` kept the neighborhood list-column, which `write.table` can't flatten: 300 gene rows were being serialised into ~16k lines of malformed TSV. That's what the old `has_n_lines n="15878"` was measuring. Now dropped like `de_results`, assertion corrected to 301.

Thanks very much!

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,399 @@
...
+        --chrom2_name '$tumor_analysis_cond.chrom2_name'
+        --chrom1_thresh '$tumor_analysis_cond.chrom1_thresh'
+        --chrom2_thresh '$tumor_analysis_cond.chrom2_thresh'
+    #end if
+    ]]></command>
+    <inputs>
+        <param name="sce" type="data" format="rds" label="SingleCellExperiment object (RDS)"/>
+        <param name="meta_table" type="data" format="tabular" label="Sample metadata table (TSV)" help="Tabular file with one row per sample/cell. Must contain columns for condition and optionally for batch or covariates."/>
+        <param name="cell_id_column" type="data_column" data_ref="meta_table" label="Cell ID column" help="Column in metadata that contains the cell IDs matching colnames of SCE"/>
+        <param name="condition_column" type="data_column" data_ref="meta_table" label="Condition column" help="Select the condition column (e.g., treatment vs control). Only appears after loading metadata."/>
```
nilchia:
```suggestion
        <param name="condition_column" type="data_column" data_ref="meta_table" label="Condition column" help="Select the condition column (e.g., treatment vs control)."/>
```

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,399 @@
...
+        <param name="covariate_columns" type="data_column" data_ref="meta_table" multiple="true" label="Covariate columns" help="One or more covariate variables to include in the design formula, in addition to the condition (e.g. batch, patient, sex, age). At least one is required: find_de_neighborhoods forms pseudobulk samples by grouping on the covariate(s) together with the condition, and a condition-only design cannot be fit (as many coefficients as pseudobulk samples)."/>
```
nilchia: If this is the case, then I think we need two params, one for covariates and one for selecting the `group_by` for doing the pseudobulk.

tools/lemur/lemur.xml
nilchia: Then please relabel. Tumor annotation is in ColData. This one is in RowData

tools/lemur/lemur.xml
nilchia: Can you please show me what is the structure? probably it can be rbind()ed

tools/lemur/lemur.R
nilchia: The user will pick the column ID (although they see the column name). That's why here in the tests you also put a number as the value

tools/lemur/lemur.R
```diff
@@ -0,0 +1,260 @@
...
+umap_df <- as_tibble(fit$colData) |> mutate(UMAP1 = umap[, 1], UMAP2 = umap[, 2])
+p_umap <- ggplot(umap_df, aes(x = UMAP1, y = UMAP2)) +
+    geom_point(aes(color = if (!is.null(batch_names)) .data[[batch_names[1]]] else NULL, shape = .data[[condition_name]]), size = 0.5, na.rm = TRUE) +
+    coord_fixed() +
+    theme_minimal()
+save_plot(opt$output_umap, p_umap, format = opt$plot_format, width = opt$plot_width, height = opt$plot_height)
+
+# ---- Volcano plot ----
+group_vars <- vars(!!!rlang::syms(c(batch_names, condition_name)))
+neighborhoods <- find_de_neighborhoods(fit, group_by = group_vars)
```
nilchia:
```
test_method choice of test for the pseudobulked differential expression. glmGamPoi and
edgeR work on the counts. limma works on the continuous assay
```

please add this as well

nilchia (changes requested):
Can you please check what this function does?
```
test_global(
fit,
contrast,
reduced_design = NULL,
consider = c("embedding+linear", "embedding", "linear"),
variance_est = c("analytical", "resampling", "none"),
verbose = TRUE,
...
)
```
Maybe we should include that as well.

dianichj:
> Can you please check what this function does?
> 
> ```
> test_global(
> fit,
> contrast,
> reduced_design = NULL,
> consider = c("embedding+linear", "embedding", "linear"),
> variance_est = c("analytical", "resampling", "none"),
> verbose = TRUE,
> ...
> )
> ```
> 
> Maybe we should include that as well.

I checked `test_global` and isn't exported in lemur 1.4.0. It's in the namespace but not part of the public API, so it isn't usable from the wrapper without `:::`. Beyond that, it answers a different question from the rest of the tool: it tests whether the *global embedding* differs between conditions, rather than which genes are differentially expressed in which neighborhoods. It returns a data.frame, so it would be a new output rather than a parameter on an existing one.

There's also a practical constraint: `variance_est = "analytical"` only works with `consider = "linear"`, and the other options need resampling (100 iterations by default), which would add significant runtime.

Can we leave it for a follow-up PR than add it here? I am happy to open an issue for it if you think it's worth tracking :D

nilchia:
@dianichj, thanks for checking.
Yes, in that case, please leave it.

nilchia (approved):
GREAT!

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,412 @@
+<tool id="lemur" name="LEMUR" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@" license="MIT">
+    <description>Multi-condition single-cell latent modeling analysis</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements"/>
```
Marie59, line 6: Hi there, just passing by ;) There seem to be missing the required_files tag around here I think

tools/lemur/lemur.xml
```diff
@@ -0,0 +1,412 @@
...
+    <command detect_errors="exit_code"><![CDATA[
```
nilchia, line 11:
```suggestion
    <required_files>
        <include path="lemur.R"/>
    </required_files>
    <command detect_errors="exit_code"><![CDATA[
```

nilchia (changes requested):

nilchia (approved):

nilchia:
@SaimMomin12 can you please review/merge? thanks!

SaimMomin12:
Thanks @dianichj for the persistence on this one, and @nilchia for the very thorough rounds so far. CI is green, `planemo lint` is clean, and `planemo format` reports both XML files already formatted, so the mechanics are in good shape. Below is what I found reviewing against IUC standards. A few are real correctness bugs that CI can't see because no test exercises those paths.

## 🔴 Blocking

**1. `test_method="none"` always crashes the tool**

`find_de_neighborhoods()` skips the test entirely for `none` ([upstream](https://github.com/const-ae/lemur/blob/master/R/find_de_neighborhoods.R#L118)):

```r
skip_independent_test <- is.null(test_data) || test_method == "none"
# ...
if (skip_independent_test) {
  colnames <- c("name", "indices", "n_cells", "sel_statistic")
}
```

So there is no `lfc`/`pval`/`adj_pval` column, and [the volcano plot](https://github.com/galaxyproject/tools-iuc/blob/b4ad82fb731833eeba005f550aaf821fc1d46ff3/tools/lemur/lemur.R#L196-L201) fails at `ggsave()` with `object 'lfc' not found`. Please either drop `none` from the select, or guard the volcano plot + document that the DE table carries no statistics in that mode.

**2. The metadata TSV silently overwrites the SCE's own `colData`**

[`lemur.R#L134`](https://github.com/galaxyproject/tools-iuc/blob/b4ad82fb731833eeba005f550aaf821fc1d46ff3/tools/lemur/lemur.R#L134) does a wholesale replacement:

```r
colData(sce) <- S4Vectors::DataFrame(meta)
```

Everything already in `colData` (cell type labels, cluster assignments, QC metrics) is discarded, and users are required to export and re-supply annotation that SCE objects normally already carry. This is @nilchia's still-open question ("Isn't it included in the SCE object as `colData`?") and it's also what caused the earlier `cell_id`-not-in-SCE test failure. Suggested fix: a conditional — *"Take variables from the SCE `colData`"* (default) vs *"Supply an external metadata table"*. At minimum, merge rather than replace, and state the overwrite in the help.

**3. Help renders as one run-on paragraph**

The help is written in Markdown but `<help>` has no `format` attribute, so Galaxy parses it as reStructuredText. Rendering it through docutils gives:

```html
<p><strong>Inputs:</strong>
- A <cite>SingleCellExperiment</cite> object in RDS format.
- A metadata table, with column selectors for ...</p>
```

The `- ` bullets never become a list (no blank line before them), single backticks become `<cite>` italics instead of code, and `**Why use `SingleCellExperiment` (SCE)?**` renders with literal backticks and broken bold. One-line fix — the XSD supports it and the rest of the block is already valid Markdown:

```xml
<help format="markdown"><![CDATA[
```

**4. Non-matching chromosome names fail with a cryptic error**

[`lemur.R#L234-L237`](https://github.com/galaxyproject/tools-iuc/blob/b4ad82fb731833eeba005f550aaf821fc1d46ff3/tools/lemur/lemur.R#L234-L237): if `rowData` uses `chr7` while the default is `7`, the logical subscript matches nothing, `colMeans()` on a zero-row matrix returns `NaN`, `is_tumor` becomes `NA`, and `fit[, tumor_label_df$is_tumor]` dies on "subscript contains NAs" — after the full model fit has already run. `NA`s in that column do the same. You already have an excellent explicit `stop()` for the missing *column*; please add the equivalent check for the chromosome *values* matching at least one gene.

**5. Tumor branch hardcodes `logcounts()` while `use_assay` is user-configurable**

Same lines as above use `logcounts(fit)` directly. If someone sets "Assay to use for model fitting" to anything else and the object has no `logcounts`, this fails late. Use `assay(fit, opt$use_assay)`.

## 📦 Dependencies

**6. `bioconductor-lemur` 1.8.0 is available on bioconda; the wrapper pins 1.4.0.** That's two Bioconductor cycles behind (upstream devel is 1.11.2). IUC expects new wrappers to target the latest packaged release. If you bump, the `bioconductor-edger` 4.4.0 / `bioconductor-limma` 3.62.1 pins need to move to the same Bioconductor release (they're currently the Bioc 3.20 set that matches lemur 1.4.0), and the size assertions will need regenerating. If there's a reason to stay on 1.4.0, worth noting it in the PR.

**7. `r-tidyverse` 2.0.0** pulls the whole tidyverse for what is effectively `dplyr` + `tibble` + `tidyr` + `rlang`, with `r-ggplot2` already pinned separately. Narrowing this will meaningfully shrink the environment.

## 📋 Missing IUC boilerplate

**8. No `<version_command>`:**

```xml
<version_command><![CDATA[Rscript -e "cat(as.character(packageVersion('lemur')))"]]></version_command>
```

**9. No `<xrefs>` — and there is no bio.tools entry for lemur at all** (`bio.tools/api/t/?q=lemur` returns 0 hits). IUC standards ask for the cross-reference, so this means registering it first, then adding `<xref type="bio.tools">lemur</xref>` plus `<edam_topics>` / `<edam_operations>`.

**10. Nothing bounds thread usage.** lemur, harmony and uwot will all happily grab every core on a shared node. The usual guard, following e.g. `tools/hyphy/hyphy_relax.xml`:

```
export OMP_NUM_THREADS="\${GALAXY_SLOTS:-1}" &&
export OPENBLAS_NUM_THREADS="\${GALAXY_SLOTS:-1}" &&
Rscript '$__tool_directory__/lemur.R'
```

and passing `n_threads` through to `uwot::umap()`.

## 🧪 Test suite

**11. Tests 1–3 are identical except for `plot_format`.** Each one re-downloads the RDS and re-runs the full fit plus two `find_de_neighborhoods()` calls — 3× the CI cost of the most expensive tool in the PR, to vary one attribute. Keep one PDF test with the full output set, keep one PNG test to prove `change_format` works, and drop or repurpose the JPG one.

**12. Parameter coverage gaps.** None of these is exercised anywhere: `use_harmony=no`, any `test_method` other than the default, `consider`, `linear_coefficient_estimator`, and a run with **no covariate columns** (every test sets `covariate_columns value="1"` — @nilchia asked for the no-covariate case explicitly). Repurposing test 3 into *"harmony off + limma + no covariates"* covers several at once for free, and would have caught finding #1 if it used `test_method=none`.

**13. Image size windows.** `has_size value="186480" delta="75000"` is ±40% — effectively "non-empty and plausible", but it will still churn on every ggplot2/R update. Now that the `from_work_dir` bug is fixed, `<has_size min="10000"/>` says the same thing more robustly. (Measuring real sizes was absolutely the right move for *finding* the 0-byte bug — nice catch on that.)

**14. Pipe syntax for conditionals** — IUC prefers explicit nesting:

```xml
<conditional name="tumor_analysis_cond">
    <param name="run_tumor_analysis" value="yes"/>
    <param name="tumor_annotation_column" value="chromosome"/>
</conditional>
```

**15.** `tumor_neigh` asserts `has_n_columns n="12"` but `de_results` doesn't — worth adding for symmetry, since both tables have the same shape.

## ✏️ Params & polish

**16.** `test_fraction max="1.0"` — 1.0 leaves no training data at all. Cap it lower (0.5, or 0.9 if you want the headroom).

**17.** `plot_width` / `plot_height` still have no `min`/`max` (@nilchia's 2026-08-07 request). Right now a negative value reaches `ggsave()` and fails *after* the whole model fit.

**18.** `argument=` still isn't used for the params that map 1:1 onto lemur function arguments — `n_embedding`, `test_fraction`, `linear_coefficient_estimator`, `use_assay`, `consider`, `test_method`, `group_by` (also @nilchia's earlier request). It surfaces the upstream name in the UI, which matters a lot for a tool whose docs are the R man pages.

**19.** Naming drift between XML and R: `covariate_columns` → `--batch_column`, `group_by_columns` → `--group_by_column`. Renaming the R options to match is worth doing — that exact mismatch is what produced the covariate bug you found last round.

**20.** Empty-string handling: `#if str($sel_gene):` is true for whitespace-only input, and the output filter `<filter>sel_gene</filter>` would then promise two datasets that the job never produces. Use `#if str($sel_gene).strip():` and a matching filter. Same idea for `#if str($covariate_columns) != 'None'` → `#if str($covariate_columns).strip() not in ('', 'None')`.

**21.** [`lemur.R#L186`](https://github.com/galaxyproject/tools-iuc/blob/b4ad82fb731833eeba005f550aaf821fc1d46ff3/tools/lemur/lemur.R#L186) colours the UMAP by `batch_names[1]` only — if a user selects three covariates, two are silently ignored. Either say so in the help or use `interaction()`.

**22.** Galaxy's default text sanitizer allows only `letters + digits + " -=_.()/+*^,:?!"`. A condition level containing `%`, `&`, `#`, `'`, `[`, `]` or `;` gets replaced with `X`, so the contrast silently won't match and the user gets a confusing lemur error. A `<sanitizer>` on `contrast_condition` / `reference_condition` would be prudent.

**23.** `.shed.yml`: `auto_tool_repositories` + `suite` are for multi-tool suites — a plain `.shed.yml` is simpler for one tool. Also a few trailing spaces in the `long_description` blocks.

**24.** Optional enhancement: several `find_de_neighborhoods()` arguments that materially change results are pinned at defaults — `selection_procedure`, `directions`, `min_neighborhood_size`, `add_diff_in_diff`. Worth considering for a follow-up if not now.

## ✅ What's already good

- CI green, `planemo lint` clean, `planemo format` reports both files already formatted.
- Excellent diagnosis of the `from_work_dir`-isn't-Cheetah-templated bug; static filenames + `change_format` is exactly the right fix, and it's a genuinely non-obvious trap.
- `license="MIT"` matches upstream `DESCRIPTION` (`License: MIT + file LICENSE`).
- Actionable `stop()` messages for the metadata↔SCE mismatches and the missing `rowData` column — including the reverse-direction check, which is easy to forget.
- `set.seed(42)`, `<creator>`, `<required_files>`, DOI citation, `${tool.name} on ${on_string}` labels, output filters, and `test-data/metadata.tsv` at 592 KB all conform.
- Making the chr7/chr10 classification opt-in was the right call for generality.

Findings **1–5** are what I'd want fixed before merge; the rest can be batched. Happy to re-review once they're in.

---

*🤖 Drafted with [Claude Code](https://claude.com/claude-code), reviewed and verified by @SaimMomin12 before posting.*

tools/lemur/lemur.xml
```diff
@@ -7,14 +7,17 @@
     <required_files>
         <include path="lemur.R"/>
     </required_files>
+    <version_command>Rscript -e 'cat(as.character(packageVersion("lemur")))'</version_command>
     <command detect_errors="exit_code"><![CDATA[
+export OMP_NUM_THREADS="\${GALAXY_SLOTS:-1}" &&
+export OPENBLAS_NUM_THREADS="\${GALAXY_SLOTS:-1}" &&
```
nilchia, line 13: These should be set with https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-environment-variables, no?

tools/lemur/lemur.xml
```diff
@@ -63,7 +67,7 @@ Rscript '$__tool_directory__/lemur.R'
             <validator type="empty_field" message="This field cannot be empty."/>
         </param>
         <param name="n_embedding" type="integer" min="1" value="15" label="Number of latent dimensions (embedding)"/>
-        <param name="test_fraction" type="float" value="0.2" min="0.01" max="1.0" label="The fraction of cells that are split off before the model fit to keep an independent set of test observations"/>
+        <param name="test_fraction" type="float" value="0.2" min="0.01" max="0.5" label="The fraction of cells that are split off before the model fit to keep an independent set of test observations"/>
```
nilchia: Wasn't it 0.2 before?
Please set it to the default 0.2

dianichj: I do not understand, the max you mean?

nilchia: oh sorry! My mistake.

I thought you changed the `value=`

Why changing it to 0.5?

What if I want to put it to 0.51?

dianichj: 1.0 leaves no training data at all. You're right that 0.5 is arbitrary though, perhaps it's better to change it to 0.9?

tools/lemur/lemur.xml
dianichj: turns out `<environment_variables>` gets rendered by Cheetah against the tool inputs, and `$GALAXY_SLOTS` isn't available there (it's exported by the job wrapper afterwards), so it would either error or set the literal string. Found that `tools/ncbi_fcs_gx/ncbi_fcs_gx.xml` uses both in the same file and keeps them separate for exactly this, the GALAXY_SLOTS one goes inline in `<command>`. So I'll leave it as is, unless you know another way?

dianichj: Just tried it to be sure:

job never starts:
```
NameMapper.NotFound: cannot find 'GALAXY_SLOTS'
ToolTemplatingException: Error occurred while building environment variables for tool 'lemur'
```

Escaped (`\$GALAXY_SLOTS`)  runs, but the value comes out as the literal string:
```
TEST_SLOTS_VALUE=[$GALAXY_SLOTS]
```

tools/lemur/lemur.xml
nilchia: OK :+1:

tools/lemur/lemur.xml
nilchia: Oh! OK, Please keep it as is!

tools/lemur/lemur.xml
dianichj: Done!!

nilchia (approved):

SaimMomin12 (approved):

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/32823499567

PR #7206, opened 2025-08-06
New tool addition: GAMMA , Gene Allele Mutation Microbial Assessment

lsterck:
FOR CONTRIBUTOR:

This will likely evolve to a suite of (two) tools. Any pointers how to best achieve this?


tools/gamma/.shed.yml
```diff
@@ -0,0 +1,17 @@
...
+    then translates and annotates the match by providing the type (i.e., mutant, truncation, etc.) 
+    and a translated description (i.e., Y190S mutant, truncation at residue 110, etc.). Because 
+    microbial gene families often have multiple alleles and existing databases are rarely exhaustive, 
+    GAMMA is helpful in both identifying and explaining how unique alleles differ from their closest 
+    known matches.    
+name: gamma
+owner: iuc
+remote_repository_url: "https://github.com/galaxyproject/tools-iuc/tree/main/tools/gamma/"
+homepage_url: "https://github.com/rastanton/GAMMA"
+type: unrestricted
```
bernt-matthias, line 17: Make this a suite of there are multiple tools. Have a look at the shed yml files of ampvis

tools/gamma/gamma.xml
```diff
@@ -0,0 +1,116 @@
...
+        <organization name="Sciensano-BioIT" url="https://github.com/BioinformaticsPlatformWIV-ISP" />
+    </creator>
+	<requirements>
+		<requirement type="package" version="@TOOL_VERSION@">GAMMA</requirement>
+	</requirements>
+	<version_command>echo @TOOL_VERSION@</version_command>
+	<command detect_errors="exit_code"><![CDATA[
+ GAMMA.py
+
+ $input_fasta
```
bernt-matthias: Variables corresponding to files or text parameters need single quotes

tools/gamma/gamma.xml
```diff
@@ -0,0 +1,116 @@
...
+        <param argument="--extended" type="boolean" truevalue="-e" falsevalue="" checked="false" label="Return all gene mutations" help="Returns all gene mutations, otherwise if there are more than 10 mutations present the count is given" />
+        <section name="out_options" title="Output options">
+            <param argument="--fasta" type="boolean" truevalue="-f" falsevalue="" checked="false" label="Write out a multifasta file of the gene matches" />
+            <param argument="--gff" type="boolean" truevalue="-g" falsevalue="" checked="false" label="Generate a general feature format (.GFF) file of the output gene matches" />
+            <param argument="--headless" type="boolean" truevalue="-l" falsevalue="" checked="false" label="Remove the column headers in the .gamma output file" />
+        </section>
+	</inputs>
+	<outputs>
+        <data name="gamma_out" format="tabular" from_work_dir="gamma_out.gamma" label="${tool.name} on ${input_fasta.name} ($on_string): GAMMA Output" />
```
bernt-matthias: Can you remove the`input_fasta.name`

tools/gamma/gamma.xml
lsterck: argh, indeed, my bad.

tools/gamma/gamma.xml
lsterck: I can (and will if requested)

Just want to point out that the users in our institute often request if this can not be included as it makes it easier for them to keep track of input files. 
Also when the results are downloaded it will indicate the input file name rather than for instance 'data X and Y'.

Is there any other way to achieve something similar that is generally accepted then?

bernt-matthias: Collections are the solution IMO. Just organize all datasets (maybe even if its just one) in a collection. In the collection you can use arbitrary names (eg samples). Then you run the tool on the collection and the output collection will have the same names. The additional advantage is that you do not need to start the tool for each sample manuall, but galaxy will start one for each sample automatically.

lsterck:
thanks for the pointer @bgruening  , I'll consider it :-)

PR #7220, opened 2025-08-14
BUSCO wrapper update to correct issue of empty table

scorreard:
FOR CONTRIBUTOR:

This PR is to correct the issue raised in https://github.com/galaxyproject/tools-iuc/issues/6664

Here is a breakdown of the changes :
**When the lineage is set to auto**
In that case, BUSCO runs a generic lineage analysis and then go to more specific ones to identify the best lineage for the species
- Extract the names of the specific and generic lineages
- Renames the files with "specific" or "generic" depending on which lineage was used to generate them
- Output the results for both the generic and specific lineages for the output type requested by the user

**When the lineage is selected by the user**
- Renames the files with "specific"
- Output the results for the specific lineages for the output type requested by the user

I tested it locally and it behaved as expected.

### DIFF
--- a/tools/busco/busco.xml
+++ b/tools/busco/busco.xml
@@ -486,6 +486,29 @@ busco
                 <has_text text="BUSCO analysis done"/>
             </assert_stdout>
         </test> -->
+		<test expect_num_outputs="2">
+            <param name="test" value="true"/>
+            <param name="input" value="bacilli_odb_test.fasta"/>
+            <conditional name="lineage">
+                <param name="lineage_mode" value="auto_detect"/>
+                <param name="auto_lineage" value="--auto-lineage-prok"/>
+            </conditional>
+            <conditional name="busco_mode">
+                <param name="mode" value="geno"/>
+            </conditional>
+            <param name="outputs" value="short_summary"/>
+            <output name="busco_sum">
+                <assert_contents>
+                    <has_text text="Gene predictor used: prodigal"/>
+                </assert_contents>
+            </output>
+            <output name="busco_table">
+                <assert_contents>
+                    <has_text text="# BUSCO version is: @TOOL_VERSION@"/>
+                    <has_text text="The lineage dataset is: bacilli_odb10"/>
+                </assert_contents>
+            </output>
+        </test>
         <test expect_num_outputs="10">
             <param name="test" value="true"/>
             <param name="input" value="genome.fa"/>


scorreard:
Set to draft as we may want to add the test (https://github.com/galaxyproject/tools-iuc/pull/7209) before solving the issue

scorreard:
Following feedback from @rlibouba, I'll make an update so that :
- By default, only results from the specific lineage are displayed
- Upon selection from the user, it will also show the results from the generic lineage in their history

lsterck (approved):
Nice one @scorreard ! :-) 
And excellent suggestion @rlibouba , thanks!

tools/busco/busco.xml
```diff
@@ -434,23 +503,50 @@ busco
             <param name="outputs" value="short_summary,missing,image,gff"/>
             <output name="busco_sum">
                 <assert_contents>
-                    <has_text text="Gene predictor used: metaeuk"/>
```
lsterck, line 437: can't we keep the gene predictor used in this assertion? Just to make it a bit more strict

lsterck (commented):
was the test with the bacilli input removed? (or is that file simply renamed?)

scorreard:
> was the test with the bacilli input removed? (or is that file simply renamed?)

Are you talking about the test developped [in PR 7209](https://github.com/galaxyproject/tools-iuc/pull/7209)?

tools/busco/busco.xml
scorreard: Yes, adding it again below too. Thanks for the review

lsterck:
> > was the test with the bacilli input removed? (or is that file simply renamed?)
> 
> Are you talking about the test developped [in PR 7209](https://github.com/galaxyproject/tools-iuc/pull/7209)?

ah, yes indeed.
or did we decide not to include it yet?
anyway, not a big point as the 'procedure' is tested with the current dataset as well.

scorreard:
> > > was the test with the bacilli input removed? (or is that file simply renamed?)
> > 
> > 
> > Are you talking about the test developped [in PR 7209](https://github.com/galaxyproject/tools-iuc/pull/7209)?
> 
> ah, yes indeed. or did we decide not to include it yet? anyway, not a big point as the 'procedure' is tested with the current dataset as well.

I'd like to include it but we did not get replies from bgruening or bernt-matthias. Should I just include your code here? I think it would be great to include it

bernt-matthias:
> I'd like to include it but we did not get replies from bgruening or bernt-matthias

I would say: yes :)

tools/busco/busco.xml
scorreard: It's because it changed, now it's 'prodigal'

scorreard:
Ok, it should be in now, wiaiting for the tests to pass and then it's ready for review.

scorreard:
This PR includes :
- An update to BUSCO outputs to deal with the multiple output with similar names
- A new test to identify empty files in BUSCo ouptuts.

It's **ready for reviews**

rlibouba (commented):
Great job, thank you!

bgruening (approved):
Let's ship this and then work on the 6.0 version :)

Thanks everyone!

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/18247366074

bernt-matthias:
Send deployed

PR #6998, opened 2025-05-19
add: metaMDBG wrapper

Kulik7:
Adding tool wrapper based on request https://github.com/galaxyproject/tools-iuc/issues/6578

FOR CONTRIBUTOR:
* [x] I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [x] License permits unrestricted use (educational + commercial)
* [x] This PR adds a new tool or tool collection
* [ ] This PR updates an existing tool or tool collection
* [ ] This PR does something else (explain below)

PR #7224, opened 2025-08-15
New tool addition: metaMDBG suite 

tcollins2011:
Add metaMDBG tool suite to galaxy. 

Different approach to #6998

FOR CONTRIBUTOR:

### DIFF
--- a/tools/metamdbg/macros.xml
+++ b/tools/metamdbg/macros.xml
@@ -1,7 +1,7 @@
 <macros>
     <token name="@TOOL_VERSION@">1.2</token>
     <token name="@VERSION_SUFFIX@">0</token>
-    <token name="@PROFILE@">21.01</token>
+    <token name="@PROFILE@">25.0</token>
     <xml name="requirements">
         <requirements>
             <requirement type="package" version="@TOOL_VERSION@">metamdbg</requirement>


tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,320 @@
...
+  <macros>
+    <import>macros.xml</import>
+  </macros>
+  <expand macro="edam"/>
+
+  <requirements>
+    <requirement type="package" version="@TOOL_VERSION@">metamdbg</requirement>
+  </requirements>
+
+  <stdio><exit_code range="1:" level="fatal"/></stdio>
```
bgruening: Is this still needed, if you use detect_errors="exit_code"?

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,320 @@
...
+  <version_command>metaMDBG --version</version_command>
+  <command detect_errors="exit_code"><![CDATA[
+    mkdir -p assemblyDir ;
```
bgruening:
```suggestion
    mkdir -p assemblyDir &&
```

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,320 @@
...
+
+    ## Collect reads from file(s) and/or list collection (note: inputs.* because they live in a <section name="inputs">)
+    #set reads_list = []
+    #if $inputs.reads
+        #for $r in $inputs.reads
+        #set $x = $reads_list.append($r.file_name)
```
bgruening:
```suggestion
            #set $x = $reads_list.append($r.file_name)
```

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,320 @@
...
+        #end for
+    #end if
+    #if $inputs.reads_collection
+        #for $c in $inputs.reads_collection
+        #set $y = $reads_list.append($c.file_name)
```
bgruening:
```suggestion
            #set $y = $reads_list.append($c.file_name)
```

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,320 @@
...
+    metaMDBG asm 
+        --out-dir assemblyDir
+        --threads \${GALAXY_SLOTS:-4} 
+    #if $read_type == "hifi"
+        --in-hifi 
+    #else
+        --in-ont 
+    #end if
+    #for $p in $reads_list
+        "$p"
```
bgruening:
```suggestion
        '$p'
```

Please use single-quotes for security reasons.

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,320 @@
...
+    #if $correction.min_read_identity is not None and str($correction.min_read_identity) != ""
+        --min-read-identity $correction.min_read_identity 
+    #end if
+    #if $correction.min_read_overlap is not None and str($correction.min_read_overlap) != ""
+        --min-read-overlap $correction.min_read_overlap 
+    #end if
+    #if str($correction.skip_correction) == "true"
+        --skip-correction 
+    #end if
+    ;
```
bgruening:
```suggestion
    &&
```

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,320 @@
...
+
+    ## Bundle for downstream graphing
+    tar -C assemblyDir -czf assemblyDir.tgz . ;
+  ]]></command>
+
+  <inputs>
+    <param name="read_type" type="select" label="Read type (maps to --in-*)">
```
bgruening:
```suggestion
    <param name="read_type" type="select" label="Read type" help="maps to --in-*">
```

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,320 @@
...
+    </param>
+
+    <section name="inputs" title="Input reads"  expanded="true">
+      <param name="reads" type="data" format="fastq,fastq.gz,fastqsanger,fastqsanger.gz"
+             multiple="true" optional="true"
+             label="FASTQ file(s)"
+             help="Multi-select one or more FASTQ files for (co-)assembly."/>
+      <param name="reads_collection" type="data_collection" collection_type="list"
+             optional="true"
+             label="OR: FASTQ collection (list)"
```
bgruening: what means OR here?

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,320 @@
...
+    <section name="correction" title="Correction options (advanced)" expanded="false">
+        <param name="min_read_quality" type="integer" optional="true"
+                label="Minimum read average quality (--min-read-quality)"/>
+        <param name="density_correction" type="float" optional="true"
+                label="Fraction of k-mers used for correction (--density-correction)"/>
+        <param name="min_read_identity" type="float" optional="true"
+                label="Minimum read identity (--min-read-identity)"/>
+        <param name="min_read_overlap" type="integer" optional="true"
+                label="Minimum read overlap length (--min-read-overlap)"/>
+        <param name="skip_correction" type="boolean" value="false"
```
bgruening: if you use


```suggestion
        <param argument="--skip-correction" type="boolean" value="false"
```

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,320 @@
...
+                label="Skip read correction (--skip-correction)"/>
```
bgruening:
```suggestion
                label="Skip read correction"/>
```

You will get this for free.

bgruening: Please consider using arguments wherever you can.

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,320 @@
...
+        <assert_contents>
+          <has_archive_member path="^(\.\/)?contigs\.fasta\.gz$"/>
+        </assert_contents>
+      </output>
+    </test>
+
+  </tests>
+
+
+
```
bgruening:
```suggestion
```

tools/metamdbg/metamdbg_asm.xml
tcollins2011: I think I'm probably not setting this up correctly. The tool can take in a single fastq file or a list of space separated fastq files for a co assembly, this can consist of any number o files. The attempt here is to allow users to pass in a collection of fastqs and treat them as a separated list for co assembly purposes. This can be done in combination with selecting individual files from a history, but I'm not sure if this is the most efficient way to set that up.

tools/metamdbg/metamdbg_asm.xml
tcollins2011: No it shouldn't be needed with detect errors.

tools/metamdbg/metamdbg_asm.xml
tcollins2011: Is it best practice to use argument rather than name for all of the passabel arguments in the input section?

tcollins2011:
@bgruening Does the tar ball approach also make sense? This should allow users to create multiple gfa plots off of different k values without needing to re-run the expensive assembly step. Additionally, we need to complete the assembly step to even know what k values are allowed. But I wanted to make sure that this approach is sound. 

Currently I'm requiring users to resubmit the same read files used, metaMDBG stores temporary local paths that can be referenced on the same galaxy server but break upon anything chaning. Is this an adequate solution or do you know of a better method?

tools/metamdbg/macros.xml
```diff
@@ -0,0 +1,20 @@
+<macros>
+    <token name="@TOOL_VERSION@">1.2</token>
+    <token name="@VERSION_SUFFIX@">0</token>
+    <token name="@PROFILE@">21.01</token>
+    <xml name="edam">
```
bernt-matthias, line 10: Add https://bio.tools/metamdbg

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,316 @@
...
+  ]]></command>
+
+  <inputs>
+    <param name="read_type" type="select" label="Read type" help="maps to --in-*">
+      <option value="hifi">PacBio HiFi (uses --in-hifi)</option>
+      <option value="ont">Nanopore R10.4+ (uses --in-ont)</option>
+    </param>
+
+    <section name="inputs" title="Input reads"  expanded="true">
+      <param name="reads" type="data" format="fastq,fastq.gz,fastqsanger,fastqsanger.gz"
```
bernt-matthias: Remove fastq and fastq.gz

tools/metamdbg/metamdbg_asm.xml
bernt-matthias: Just use data param with `multiple="True"`. Users can provide multiple data sets or a collection.

For a separate assembly of single fastq files users should provide a collection of collections, where the inner collection contains a single dataset)

I hope that this gets better documented soon https://github.com/galaxyproject/galaxy/issues/20657

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,316 @@
...
+             label="FASTQ file(s)"
+             help="Multi-select one or more FASTQ files for (co-)assembly."/>
+      <param name="reads_collection" type="data_collection" collection_type="list"
+             optional="true"
+             label="OR: FASTQ collection (list)"
+             help="Provide a list collection of FASTQs instead of selecting files above."/>
+    </section>
+
+    <section name="assembly" title="Assembly options (advanced)" expanded="false">
+        <param argument="--kmer_size" type="integer" optional="true"
```
bernt-matthias: What is the default?

Either document the value or just use it as (default)value.. Later would be my preference.

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,316 @@
...
+
+  </inputs>
+
+  <outputs>
+    <data name="contigs" format="fasta.gz"
+          from_work_dir="assemblyDir/contigs.fasta.gz"
+          label="metaMDBG contigs on ${on_string}"/>
+    <data name="assembly_bundle" format="tar.gz"
+          from_work_dir="assemblyDir.tgz"
+          label="metaMDBG assembly bundle (tar)"/>
```
bernt-matthias: I would prefer  the same style for the labels that we use everywhere else.

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,316 @@
...
+    #end if
+    #if $inputs.reads_collection
+        #for $c in $inputs.reads_collection
+            #set $y = $reads_list.append($c.file_name)
+        #end for
+    #end if
+
+    metaMDBG asm 
+        --out-dir assemblyDir
+        --threads \${GALAXY_SLOTS:-4}
```
bernt-matthias: Double quote shell variables

tools/metamdbg/metamdbg_asm.xml
bgruening: Yes :)

bgruening: When possible.

tcollins2011:
@bernt-matthias @bgruening I think I have everything cleaned up now.

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,279 @@
...
+  <description>Assemble long-read metagenomes (HiFi / ONT)</description>
+
+  <macros>
+    <import>macros.xml</import>
+  </macros>
+  <expand macro="edam"/>
+  <expand macro="biotools"/>
+  <requirements>
+    <requirement type="package" version="@TOOL_VERSION@">metamdbg</requirement>
+  </requirements>
```
bernt-matthias: This should be a macro as well.

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,279 @@
...
+
+  <version_command>metaMDBG --version</version_command>
```
bernt-matthias: Maybe also this?

tools/metamdbg/metamdbg_asm.xml
bernt-matthias:
```suggestion
        --threads "\${GALAXY_SLOTS:-4}"
```

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,279 @@
...
+
+
+    metaMDBG asm 
+        --out-dir "assemblyDir"
+        --threads \${GALAXY_SLOTS:-4} 
+    #if $read_type == "hifi"
+        --in-hifi 
+    #else
+        --in-ont 
+    #end if
```
bernt-matthias:
```suggestion
    --in-$read_type
```

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,279 @@
...
+    #for $p in $reads_list
+        '$p'
+    #end for
+    #if $assembly.kmer_size is not None and str($assembly.kmer_size) != ""
+        --kmer-size $assembly.kmer_size 
+    #end if
```
bernt-matthias: The parameter is non optional, so you we can simply: 

```suggestion
    --kmer-size $assembly.kmer_size
```

Please do this fo all non-optional parameters.

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,279 @@
...
+        <param argument="--min_read_quality" type="integer" value="0"
+                label="Minimum read average quality"/>
+        <param argument="--density_correction" type="float" value="0.025"
+                label="Fraction of k-mers used for correction"/>
+        <param argument="--min_read_identity" type="float" value="0.96"
+                label="Minimum read identity"/>
+        <param argument="--min_read_overlap" type="integer" value="1000"
+                label="Minimum read overlap length"/>
+        <param argument="--skip_correction" type="boolean" value="false"
+                label="Skip read correction"/>
```
bernt-matthias, line 83:
```suggestion
        <param argument="--skip_correction" type="boolean" checked="false" truevalue="--skip_correction" falsevalue="" label="Skip read correction"/>
```

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,279 @@
...
+    #end if
+    #if $correction.min_read_identity is not None and str($correction.min_read_identity) != ""
+        --min-read-identity $correction.min_read_identity 
+    #end if
+    #if $correction.min_read_overlap is not None and str($correction.min_read_overlap) != ""
+        --min-read-overlap $correction.min_read_overlap 
+    #end if
+    #if str($correction.skip_correction) == "true"
+        --skip-correction 
+    #end if
```
bernt-matthias: Can be simplified to (with my suggestion on the param's XML)

```suggestion
    $skip_correction
```

tools/metamdbg/metamdbg_gfa.xml
```diff
@@ -0,0 +1,148 @@
...
+    </test>
+
+    <!-- List mode -->
+    <test expect_num_outputs="1">
+      <param name="assembly_bundle" ftype="tar.gz" value="small_assembly_bundle.tar.gz"/>
+      <conditional name="mode">
+        <param name="action" value="list"/>
+      </conditional>
+      <output name="k_values">
+        <assert_contents><has_text text="k"/></assert_contents>
```
bernt-matthias: Can you add assertions on the number of lines (and columns?)? Should there ba a specific value for k that we can assert?

tools/metamdbg/metamdbg_gfa.xml
```diff
@@ -0,0 +1,148 @@
...
+
+    <conditional name="mode">
+      <param name="action" type="select" label="What do you want to do?">
+        <option value="list" selected="true">List available k-values only (no graph)</option>
+        <option value="graph">Generate GFA graph</option>
+      </param>
+
+      <when value="list"/>
+      <when value="graph">
+        <param argument="--k" type="integer" label="k-mer size"/>
```
bernt-matthias: Please add min (and max)?

tools/metamdbg/metamdbg_gfa.xml
```diff
@@ -0,0 +1,148 @@
...
+  <command detect_errors="exit_code"><![CDATA[
+    tar -xzf '$assembly_bundle' ;
+
+    ## Two modes: list k-values or build the graph
+    #if $mode.action == "list"
+      metaMDBG gfa --assembly-dir "./" --k 0 --threads "\${GALAXY_SLOTS:-4}" > k_values.txt 2>&1 ;
+    #else
+      ## (Re)write ./tmp/input.txt with the provided read files
+      #if $mode.reads
+        : > "/tmp/input.txt" ;
```
bernt-matthias:
```suggestion
```

or at least prefix with `./`

tools/metamdbg/metamdbg_gfa.xml
```diff
@@ -0,0 +1,148 @@
...
+        #for $r in $mode.reads
+          printf '%s\n' '${r.file_name}' >> "./tmp/input.txt" ;
+        #end for
+      #end if
+
+      metaMDBG gfa --assembly-dir "./" --k $mode.k --threads "\${GALAXY_SLOTS:-4}" ;
+
+      GFA_WITH="\$(ls -1 assemblyGraph_k*.gfa 2>/dev/null | grep -v 'noseq' | head -n 1)" ;
+      if [ -z "\$GFA_WITH" ]; then echo "No sequence-containing GFA produced." >&2 ; exit 1 ; fi ;
```
bernt-matthias, line 28: Is this always an error? I.e. might it be better to produce an empty result file?

tools/metamdbg/metamdbg_gfa.xml
tcollins2011: It would be hard to know what to size them to. The available k-mers for this have to be calculated by the command and can greatly vary. Also, the only certain k-mer values will work that are unique to each sample based on the options made during the assembly.

bernt-matthias: Certainly we can use at least `min="1"`?

tcollins2011: That we can definitely do

tools/metamdbg/metamdbg_gfa.xml
tcollins2011: I think realistically if there isn't a valid gfa file generated we should probably flag it and let the user know. In theory, if there was a terrible blank assembly generated before hand you might want to give an empty file, but I think that it would probably be better for the user to get an explicit error rather than just a blank file.

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,278 @@
...
+        <param argument="--kmer_size" type="integer" value="15"
+                label="k-mer size"/>
+        <param argument="--density_assembly" type="float" value="0.005"
+                label="Fraction of k-mers used for assembly"/>
+        <param argument="--max_k" type="integer" value="0"
+                label="Stop assembly after k iterations"/>
+        <param argument="--min_abundance" type="integer" value="0"
+                label="Minimum k-min-mer abundance"/>
+        <param argument="--all_assembly_graph" type="boolean" value="false"
+                label="Emit assembly graph at each multi-k iteration"/>
```
bernt-matthias:
```suggestion
        <param argument="--all-assembly-graph" type="boolean" checked="false" truevalue="--all-assembly-graph" falsevalue=""
                label="Emit assembly graph at each multi-k iteration"/>
```

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,278 @@
...
+    #end if
+    #if $assembly.max_k is not None and str($assembly.max_k) != ""
+        --max-k $assembly.max_k 
+    #end if
+    #if $assembly.min_abundance is not None and str($assembly.min_abundance) != ""
+        --min-abundance $assembly.min_abundance 
+    #end if
+    #if str($assembly.all_assembly_graph) == "true"
+        --all-assembly-graph 
+    #end if
```
bernt-matthias:
```suggestion
        $all_assembly_graph
```

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,278 @@
...
+    --kmer-size $assembly.kmer_size
+    #if $assembly.density_assembly is not None and str($assembly.density_assembly) != ""
+        --density-assembly $assembly.density_assembly 
+    #end if
+    #if $assembly.max_k is not None and str($assembly.max_k) != ""
+        --max-k $assembly.max_k 
+    #end if
+    #if $assembly.min_abundance is not None and str($assembly.min_abundance) != ""
+        --min-abundance $assembly.min_abundance 
+    #end if
```
bernt-matthias: The checks are not needed for non-optional parameters

```suggestion
        --density-assembly $assembly.density_assembly
        --max-k $assembly.max_k
        --min-abundance $assembly.min_abundance
```

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,278 @@
...
+
+
+    metaMDBG asm 
+        --out-dir "assemblyDir"
+        --threads "\${GALAXY_SLOTS:-4}"
+    --in-$read_type
+    #for $p in $reads_list
+        '$p'
+    #end for
+    --kmer-size $assembly.kmer_size
```
bernt-matthias: Please fix indentation.

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,278 @@
...
+    #end if
+    #if $correction.min_read_identity is not None and str($correction.min_read_identity) != ""
+        --min-read-identity $correction.min_read_identity 
+    #end if
+    #if $correction.min_read_overlap is not None and str($correction.min_read_overlap) != ""
+        --min-read-overlap $correction.min_read_overlap 
+    #end if
+    #if str($correction.skip_correction) == "true"
+      --skip-correction
+    #end if
```
bernt-matthias: Please change analogously...

tcollins2011:
@bernt-matthias @bgruening Just coming back to this tool. I just pushed up the changes that were requested last time. I was hoping for a quick re-review to see if we can it merged.

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,260 @@
...
+        --kmer-size $assembly.kmer_size
+        --density-assembly $assembly.density_assembly 
+        --max-k $assembly.max_k 
+        --min-abundance $assembly.min_abundance 
+        $all_assembly_graph
+        --min-read-quality $correction.min_read_quality 
+        --density-correction $correction.density_correction 
+        --min-read-identity $correction.min_read_identity 
+        --min-read-overlap $correction.min_read_overlap 
+        --skip-correction
```
bernt-matthias, line 40: This should not be hardcoded. Or you should remove the parameter

tools/metamdbg/metamdbg_asm.xml
bernt-matthias: ?

tools/metamdbg/.shed.yml
```diff
@@ -0,0 +1,23 @@
+name: metamdbg
+owner: iuc
+type: unrestricted
+categories:
+- Assembly
+- Metagenomics
+homepage_url: https://github.com/GaetanBenoitDev/metaMDBG
+remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/master/tools/metamdbg
```
SaimMomin12:
```suggestion
remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/main/tools/metamdbg
```

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,260 @@
...
+  <expand macro="version"/>
+
+  <command detect_errors="exit_code"><![CDATA[
+    mkdir -p assemblyDir &&
+
+    ## Collect reads from file(s) and/or list collection
+    #set reads_list = []
+    #if $inputs.reads
+        #for $r in $inputs.reads
+            #set $x = $reads_list.append($r.file_name)
```
bernt-matthias, line 19:
```suggestion
            #silent $reads_list.append($r.file_name)
```

tools/metamdbg/macros.xml
```diff
@@ -0,0 +1,33 @@
+<macros>
+    <token name="@TOOL_VERSION@">1.2</token>
+    <token name="@VERSION_SUFFIX@">0</token>
+    <token name="@PROFILE@">21.01</token>
```
SaimMomin12:
```suggestion
    <token name="@PROFILE@">25.0</token>
```

Perhaps a recent profile?

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,260 @@
...
+    #end if
+
+
+    metaMDBG asm 
+        --out-dir "assemblyDir"
+        --threads "\${GALAXY_SLOTS:-4}"
+        --in-$read_type
+        #for $p in $reads_list
+        '$p'
+        #end for
```
bernt-matthias, line 30: You could consider

```suggestion
        #echo "'"+"' '".join($reads_list)+"'"
``` 

but maybe yours is more readable

tools/metamdbg/macros.xml
bernt-matthias: Use a recent profile?

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,260 @@
...
+        --kmer-size $assembly.kmer_size
+        --density-assembly $assembly.density_assembly 
+        --max-k $assembly.max_k 
+        --min-abundance $assembly.min_abundance 
+        $all_assembly_graph
```
bernt-matthias, line 35: `$assembly.all_assembly_graph`

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,260 @@
...
+    tar -C assemblyDir -czf assemblyDir.tgz . ;
+  ]]></command>
+
+  <inputs>
+    <param name="read_type" type="select" label="Read type" help="maps to --in-*">
+      <option value="hifi">PacBio HiFi (uses --in-hifi)</option>
+      <option value="ont">Nanopore R10.4+ (uses --in-ont)</option>
+    </param>
+
+    <section name="inputs" title="Input reads" expanded="true">
```
bernt-matthias, line 53: I would remove the section (for a single parameter) or move the read_type parameter into the sectin.

SaimMomin12:
Thanks @tcollins2011 for your contribution.

tools/metamdbg/metamdbg_gfa.xml
```diff
@@ -0,0 +1,145 @@
...
+    #if $mode.action == "list"
+      metaMDBG gfa --assembly-dir "./" --k 0 --threads "\${GALAXY_SLOTS:-4}" > k_values.txt 2>&1 ;
+    #else
+      ## (Re)write ./tmp/input.txt with the provided read files
+      #if $mode.reads
+        : > "./tmp/input.txt" ;
+        #for $r in $mode.reads
+          printf '%s\n' '${r.file_name}' >> "./tmp/input.txt" ;
+        #end for
+      #end if
```
bernt-matthias, line 23: A config file would be more elegant to create the file? https://docs.galaxyproject.org/en/master/dev/schema.html#tool-configfiles-configfile

tools/metamdbg/metamdbg_gfa.xml
```diff
@@ -0,0 +1,145 @@
...
+
+      metaMDBG gfa --assembly-dir "./" --k $mode.k --threads "\${GALAXY_SLOTS:-4}" ;
+
+      GFA_WITH="\$(ls -1 assemblyGraph_k*.gfa 2>/dev/null | grep -v 'noseq' | head -n 1)" ;
+      if [ -z "\$GFA_WITH" ]; then echo "No sequence-containing GFA produced." >&2 ; exit 1 ; fi ;
+      cp "\$GFA_WITH" metamdbg.gfa ;
+
+      GFA_NOSEQ="\$(ls -1 assemblyGraph_k*.noseq.gfa 2>/dev/null | head -n 1)" ;
+      if [ -z "\$GFA_NOSEQ" ]; then echo "No no-seq GFA produced." >&2 ; exit 1 ; fi ;
```
bernt-matthias, line 32: You could set a more custom exit code which we could then detect elegantly in the tool https://docs.galaxyproject.org/en/master/dev/schema.html#tool-stdio

tools/metamdbg/metamdbg_asm.xml
```diff
@@ -0,0 +1,260 @@
...
+                label="Skip read correction"/>
+    </section>
+
+  </inputs>
+
+  <outputs>
+    <data name="contigs" format="fasta.gz"
+          from_work_dir="assemblyDir/contigs.fasta.gz"
+          label="${tool.name} on ${on_string}: contigs"/>
+    <data name="assembly_bundle" format="tar.gz"
```
bernt-matthias, line 92: Wondering if we should have a separate data type for this?

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/21216503655

bernt-matthias:
The following needs to be fixed: 

```
Correct the following problems if necessary and reset metadata.<br/><b>metamdbg_gfa.xml<\/b> - parameter 'k': the attribute 'value' must be set for non optional parameters<br/>"}
```

tools/metamdbg/macros.xml
tcollins2011: Updated

tools/metamdbg/metamdbg_asm.xml
tcollins2011: I think the for loop is definitely less elegant, but would probably be easier for someone to quickly glance at and see what is happening. Is there a best practice for this? Maybe adopt the change and leave a comment explaining it?

tcollins2011:
@bernt-matthias The k-kmer value is hard to set by default. It can be a variety of different valid values depending on the assembly. In the _gfa tool section the first mode identifies the valid k-values that the assembly produced. As these can constantly differ what should the default value be?

tools/metamdbg/metamdbg_asm.xml
tcollins2011: Possibly? I'm not sure if it would ever be used outside of this specific tool suite though.

bernt-matthias:
> @bernt-matthias The k-kmer value is hard to set by default. It can be a variety of different valid values depending on the assembly. In the _gfa tool section the first mode identifies the valid k-values that the assembly produced. As these can constantly differ what should the default value be?

You just need to add `value=""`. Ist a bit stupid. Hope this will be solved with toolshed 2.0.

tools/metamdbg/metamdbg_asm.xml
bernt-matthias: Leave the loop. Readability counts

bernt-matthias:
FYI you will need to open a new PR

PR #7234, opened 2025-08-19
New tool: GAMMA-S , part of the GAMMA package

lsterck:
FOR CONTRIBUTOR:

GAMMA-S is the other tool from the GAMMA package.


bgruening:
Thanks @lsterck

lsterck:
thanks @bgruening 

small question while we're at it:
by adding this tool the lint check fails because the other tool in the package is the same version number. (I think the person who added seqkit-grep reported the same). How should this be handled? Is there a way to avoid the lint-fail in such cases?

PR #7257, opened 2025-09-04
Update kraken2 and add support for custom databases

pvanheus:
FOR CONTRIBUTOR:


pvanheus:
This both updates the kraken2 tool to version 2.1.6 and adds support for custom databases. The custom databases are provided by directory types and because zip files without a common prefix get unpacked to a directory prefixed by the name of the original dataset (see [the galaxy code](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/util/compression_utils.py#L256)), it uses some shell code to find the complete path to the custom database.

tool_collections/kraken2/macros.xml
```diff
@@ -1,15 +1,26 @@
...
+            </param>
+            <when value="builtin">
+                <param label="Select a Kraken2 database" name="builtin_database" type="select">
+                    <options from_data_table="kraken2_databases">
+                        <validator message="No Kraken2 database is available" type="no_options" />
+                    </options>
+                </param>
+            </when>
+            <when value="history">
+                <param type="data" name="custom_database" label="Kraken2 database" format="directory" help="A kraken2 is a directory containing the files hash.k2d, opts.k2d and taxo.k2d"/>
```
bernt-matthias, line 21: - How big are those typically? 
- Should we subclass directory for that? 
- We should also add a tool that crates the K2 databases, or?

tool_collections/kraken2/kraken2/kraken2.xml
```diff
@@ -14,10 +14,17 @@
...
+            DB_DIR=\$(ls -d '${kraken2_database.custom_database.extra_files_path}/dataset_'*/) &&
+        #end if 
         kraken2
             --threads \${GALAXY_SLOTS:-1}
-            --db '${kraken2_database.fields.path}'
-
+            #if $kraken2_database.database_source == "builtin"
+                --db '${kraken2_database.builtin_database.fields.path}'
+            #elif $kraken2_database.database_source == "history"
+                --db "\$DB_DIR"
```
bernt-matthias: We should wait for a proper fix https://github.com/galaxyproject/galaxy/pull/20857 and use 

```suggestion
                --db '$kraken2_database.custom_database.extra_files_path'
```

tool_collections/kraken2/macros.xml
pvanheus: Gigabyte scale. E.g. the Mycobacterium genus databases from Hall is 7.6 GB (https://zenodo.org/records/8343322). And yes, ideally we need a datatype and tool for creating Kraken2 databases, but there are already several Kraken2 databases available online that I am using for e.g. my M. tuberculosis analyses. I'm happy to work on PRs for the Kraken2 database datatype and the tool that creates them, but I don't think that that work should block this PR.

tool_collections/kraken2/kraken2/kraken2.xml
pvanheus: Yes, we need a fix for that bug, but for now this works and a fix for the Galaxy bug will likely take some times (months) to be available in a Galaxy release.

tool_collections/kraken2/macros.xml
bernt-matthias: Would this be of use for many users? 

How about just adding this to the data manager?

We are just updating the dm anyway https://github.com/galaxyproject/tools-iuc/pull/6980

tool_collections/kraken2/kraken2/kraken2.xml
bernt-matthias: I hope that this will be considered a bug fix and goes in the current release.

tool_collections/kraken2/macros.xml
pvanheus: The DM is very useful... the use case here is in Galaxy's Workflow Landings API, which is oriented around inputs, not DMs. And also, once there is a kraken2-build tool, custom databases.

tool_collections/kraken2/kraken2/kraken2.xml
```diff
@@ -199,9 +219,28 @@
...
+            <conditional name="single_paired">
+                <param name="single_paired_selector" value="no"/>
+                <param name="input_sequences" value="kraken_test1.fa" ftype="fasta"/>
+            </conditional>
+            <param name="split_reads" value="false"/>
+            <param name="quick" value="no"/>
+            <param name="confidence" value=".2"/>
+            <conditional name="kraken2_database">
+                <param name="database_source" value="history" />
+                <param name="custom_database" ftype="zip" value="test_db.zip" />
```
mvdbeek, line 241: Can you unzip this, then use
```suggestion
                <param name="custom_database" class="Directory" ftype="kraken2_database" value="path_to_directory" />
```
?

mvdbeek: (kraken2_database or similar would subclass the directory datatype)

tool_collections/kraken2/macros.xml
bernt-matthias: @mvdbeek when are data manager bundles expected. Would this help here?

mvdbeek: Directory datasets are a much better idea

tool_collections/kraken2/kraken2/kraken2.xml
pvanheus: I have added a test for the Directory provided directly *and* provided via a zip. This revealed a bug in my previous version of the code so that is now updated.

tool_collections/kraken2/kraken2/kraken2.xml
pvanheus: I have reference your bugfix in a comment in the latest version of the code

pvanheus:
Some notes on the future kraken2 custom database builder tool:

The kraken2 db building process involves three steps:
1. Download the NCBI taxonomy (`kraken2-build --download-taxonomy` or `k2 download-taxonomy`) to the database directory. This downloads some 7GB of data from ftp.ncbi.nlm.nih.gov/pub/taxonomy, uncompressing to 54GB
2. Either or both of:
  1. Download library sequences files (`kraken2-build --standard` / `kraken2-build --special` / `kraken2-build --download-library` or `k2 download-library`) to the DB directory
  2. Add your own FASTA sequence data (`kraken2-build --add-to-library` or `k2 add-to-library`) 
4. Build the database (`kraken2-build --build` or `k2 build`)
5. Clean up (remove temporary files)

This gives flexibility in how databases are built, and therefore it makes sense to have tools for (1), and some combination of (2), (3) and (4), balancing the need to not continually re-download the taxonomy files with the need to not pollute the history with intermediate products.

tool_collections/kraken2/macros.xml
bernt-matthias: > the use case here is in Galaxy's Workflow Landings API

Would this then mean that the user uploads multiple GB "just" for a single workflow run? I have not yet used Galaxy's Workflow Landings API - how is this supposed to be used?

tool_collections/kraken2/kraken2/kraken2.xml
bernt-matthias: Fix should be out now.

bernt-matthias:
> Download the NCBI taxonomy (kraken2-build --download-taxonomy or k2 download-taxonomy) to the database directory. This downloads some 7GB of data from ftp.ncbi.nlm.nih.gov/pub/taxonomy, uncompressing to 54GB

The data is available from the `ncbi_taxonomy` data table. 

The procedure that you outlined is exactly what should happen for building custom entries with the DM https://github.com/galaxyproject/tools-iuc/blob/main/data_managers/data_manager_build_kraken2_database/data_manager/kraken2_build_database.xml

pvanheus:
Just to clarify this most recent work: I have postponed working on the kraken2 database builder and, as per @bernt-matthias' suggestion, will draw on the similar code that is in the data manager. This PR simply updates to the latest kraken2 and adds support for custom databases (via a Directory input).

PR #7264, opened 2025-09-08
Add required_files to addName

natefoo:
Although the quoting is valid, I think the way it's done prevents Pulsar from implicitly copying the wrapper script? On .org we're getting:

```
python: can't open file '/jetstream2/scratch/main/jobs-vgp/70483536/tool_files/add_input_name_as_column.py': [Errno 2] No such file or directory
```

This adjusts the quoting _and_ adds required_files for good measure. No tool version change because there's no change in functionality, just a fix for running with Pulsar.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/add_input_name_as_column/add_input_name_as_column.xml
+++ b/tools/add_input_name_as_column/add_input_name_as_column.xml
@@ -3,8 +3,11 @@
     <requirements>
         <requirement type="package" version="3.7">python</requirement>
     </requirements>
+    <required_files>
+        <include path="add_input_name_as_column.py"/>
+    </required_files>
     <command><![CDATA[
-python '$__tool_directory__/'add_input_name_as_column.py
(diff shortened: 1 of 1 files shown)


natefoo:
Heh, lint failed because I didn't update the version.

I do get why we require this but it's a bit annoying to have two identical versions installed except one works half the time and the other doesn't. Obviously I can deal with this on my end in multiple ways, however.

bernt-matthias:
We can merge. The linter is skipped when merging.

Now that the older version can not be used anymore.

bernt-matthias:
Do all tools that use `__tool_directory__` should have `required_files`?

natefoo:
Someone (@jmchilton) correct me if I'm wrong, but yes I believe that is the desired state. Pulsar will implicitly find, transfer, and rewrite files used in the `<command>`, but going forward I believe the goal is to have them explicitly referenced.

PR #7283, opened 2025-09-17
Pairtools - fix error when outputing stats 

Delphine-L:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/pairtools/parse.xml
+++ b/tools/pairtools/parse.xml
@@ -19,7 +19,7 @@
             $drop_seq
             $drop_sam
             #if $output_stats:
-                 --output-stats  output.stat
+                 --output-stats '$parsed_pairs_stats'
             #end if
             #if $select_add_columns.add_columns_selection == "yes":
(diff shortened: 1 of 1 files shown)


tools/pairtools/parse.xml
```diff
@@ -37,7 +36,9 @@
             #else
                  -o '$output'
             #end if
-
+            #if $output_stats:
+               &&  mv output.stat '$parsed_pairs_stats'
+            #end if
```
bernt-matthias:
```suggestion
```

Maybe just `from_workdir="output.stat"`?

bernt-matthias (commented):
The problem is that the program needs a `.stat` extension?

Delphine-L:
> The problem is that the program needs a `.stat` extension?

I am not sure but the tool was returning an error when I was using the file path directly

bernt-matthias:
Do you have the error message?

Delphine-L:
Well, I can't repeat the error, it works now. So I simplified :)

PR #7294, opened 2025-09-21
Update lexicmap

Smeds:
These pull-request will make>
- update lexicmap version to 0.8.0
- make it possible to select multiple indices to run search against

When searching against multiple indices, we end up with multiple result files that need to be merged using the LexicMap utilities. If multiple queries are used, this merge must be performed for each query. Note that if only a single index is used, the -q flag is ignored, and in that case we do not want to run the merge event for each query, since this would produce duplicate results.

FOR CONTRIBUTOR:

Closes: https://github.com/galaxyproject/tools-iuc/pull/7288

### DIFF
--- a/tools/lexicmap/lexicmap.xml
+++ b/tools/lexicmap/lexicmap.xml
@@ -79,9 +79,9 @@ extract_query_ids() {
             -q "\${query_array[\$i]}" lexicmap_search_result__index*.tsv
             -j "\${GALAXY_SLOTS:-1}";
     done &&
-    cat combined_result.*.tsv | awk 'NR==1 || $0 !~ /^query\tqlen\thits/' > "$out_file"
+    cat combined_result.*.tsv | awk 'NR==1 || $0 !~ /^query\tqlen\thits/' > '$out_file'
 #else
-    mv lexicmap_search_result__index0.tsv "$out_file"
+    mv lexicmap_search_result__index0.tsv '$out_file'
 #end if
     ]]></command>
     <inputs>


Smeds:
We need the multiple indices option because we have creating  indices for different datasets that we want to provide access to using galaxy, and each dataset consists of multiple indices. Running the tool separately for each index would be highly inconvenient.

```bash
ls completed_indices
Genomic_bacteria.index
Genomic_mouse.index
Genomic_plant_fungi.index
Metagenomic_human.index
Metagenomic_other.index
Metatranscriptomic_human.index
Metatranscriptomic_other.index
Transcriptomic_human.index
Genomic_human.index
Genomic_other.index
Metagenomic_bacteria.index
Metagenomic_mouse.index
Metatranscriptomic_bacteria.index
Metatranscriptomic_mouse.index
Transcriptomic_bacteria.index

ls completed_indices/Genomic_bacteria.index/
group.Genomic_bacteria.n34-00.lmi
group.Genomic_bacteria.n34-01.lmi
group.Genomic_bacteria.n34-02.lmi
...
group.Genomic_bacteria.n34-32.lmi
group.Genomic_bacteria.n34-33.lmi
Viral.index
```

tools/lexicmap/lexicmap.xml
```diff
@@ -7,43 +7,74 @@
     <expand macro="requirements"/>
 
     <command detect_errors="exit_code"><![CDATA[
-    
-lexicmap search 
+#if $db_opts.db_opts_selector == "histdb"
+    #set INDICES = str($db_opts.histdb).replace('.dat', '_files').split(",")
```
Smeds: Before this changes I run  `        --index '${db_opts.histdb.extra_files_path}'` to get the path but I don't seem to be able to do it while using `#set INDICES = ...`. Should I be able to call `extra_files_path` instead of having to do the replace statement?

Smeds: this solution works but I don't know if it's a good solution.

bernt-matthias:
>  Running the tool separately for each index would be highly inconvenient.

Can you elaborate? The obvious downside of this is that all indices are processed in one job in a serial fashion, if you just map over a collection the indices may be processed in parallel (given free resources).

Smeds:
> > Running the tool separately for each index would be highly inconvenient.
> 
> Can you elaborate? The obvious downside of this is that all indices are processed in one job in a serial fashion, if you just map over a collection the indices may be processed in parallel (given free resources).

It’s true that we would lose some parallelism, but it would make things easier for new users who aren’t familiar with Galaxy. Even if we decide not to set up Galaxy to run against multiple indices in the same job, having the option would still be very useful. For the LexicMap pre-built indices, we’d probably want to run one job per collection of indices. Still, nothing prevents us from also offering a configuration where all indices are listed—though that dropdown could become quite large. I also asked why they chose to build multiple groups of indices for a given data type rather than a single large one.

Smeds:
@bernt-matthias do you see any downside with having the option of in one job running against multiple indices?

tools/lexicmap/lexicmap.xml
bernt-matthias: > this solution works but I don't know if it's a good solution.

definitely not :). Probably something like the following should work:

```
#set INDICES = [db.extra_files_path for db in db_opts.histdb]
```

tools/lexicmap/lexicmap.xml
```diff
@@ -7,47 +7,74 @@
...
+echo '${INDICES}' &&
+#import gzip
+#for $q in $query
+    #set q_path = $q
+    #if $q.extension.endswith('gz')
+        #set queries = queries + [line.strip()[1:] for line in gzip.open(str($q_path)).read().decode('utf-8').splitlines() if line.startswith('>')]
+    #else:
+        #set queries = queries + [line.strip()[1:] for line in open(str($q_path)).read().decode('utf-8').splitlines() if line.startswith('>')]
+    #end if
+#end for
```
bernt-matthias: What are you doing here? Reading the fasta in cheetah is likely a very bad idea since this happens in one of the Galaxy processes which might be struggling in case of large fasta files...

tools/lexicmap/lexicmap.xml
```diff
@@ -7,47 +7,74 @@
...
+#set counter = 0
+#for $index in $INDICES:
+    #set counter = $counter + 1
```
bernt-matthias:
```suggestion
#for $counter, $index in enumerate($INDICES):
```

tools/lexicmap/lexicmap.xml
```diff
@@ -7,47 +7,74 @@
...
+        &&
+#end for
+#set counter = 0
+#if len($INDICES) > 1
+    #for query in $queries
+        #set counter = $counter + 1
+        lexicmap utils merge-search-results --out-file 'combined_result.${counter}.tsv'  -q ${query} lexicmap_search_result__index*.tsv -j "\${GALAXY_SLOTS:-1}" &&
+    #end for
+#else:
+    lexicmap utils merge-search-results --out-file 'combined_result.${counter}.tsv'  -q ${query} lexicmap_search_result__index*.tsv -j "\${GALAXY_SLOTS:-1}" &&
```
bernt-matthias: Why is this step needed for one file? Isn't 'combined_result.${counter}.tsv'  equal to 'lexicmap_search_result__index${counter}.tsv' for one input file?

tools/lexicmap/lexicmap.xml
```diff
@@ -7,47 +7,74 @@
...
-        --min-qcov-per-hsp '$min_qcov_per_hsp'
-    #end if
+        #if $min_qcov_per_hsp
+            --min-qcov-per-hsp '$min_qcov_per_hsp'
+        #end if
+        &&
+#end for
+#set counter = 0
+#if len($INDICES) > 1
+    #for query in $queries
```
bernt-matthias: What is lexicmap utils merge-search-results doing? Should we mention it in the help?

bernt-matthias:
> do you see any downside with having the option of in one job running against multiple indices?

might be fine. Since power users can still apply the files separately. Wondering if the merge step could be / needs to be a separate Galaxy tool.

tools/lexicmap/lexicmap.xml
Smeds: I need to extract the fasta header for each sequence to later use them to extract search result from the different indices results. https://bioinf.shenwei.me/LexicMap/usage/utils/merge-search-results/#examples

bernt-matthias: You will need to do this in a bash loop.

tools/lexicmap/lexicmap.xml
Smeds: Your suggestion worked with a small modification!

tools/lexicmap/lexicmap.xml
Smeds: Your right, it's not needed. I have change it to a move command instead!

tools/lexicmap/lexicmap.xml
Smeds: I have added a section to the help informing that merge-search-results will be used to combine the search results when multiple indices are used, I have also added a link to the tool

tools/lexicmap/lexicmap.xml
bernt-matthias: Alternatively: could you run `lexicmap search ` separately for each query file?

Smeds: I re-wrote part of the wrapper extracting the query ids using a function and using more bash loops

Smeds:
> > do you see any downside with having the option of in one job running against multiple indices?
> 
> might be fine. Since power users can still apply the files separately. Wondering if the merge step could be / needs to be a separate Galaxy tool.

@bernt-matthias I would prefer to keep it in the lexicmap search wrapper, however I could add a wrapper for the merge-search-results for user who want parallelize and then merge.

bernt-matthias:
> owever I could add a wrapper for the merge-search-results for user who want parallelize and then merge.

This is what I meant.

Smeds:
@bernt-matthias Would you be fine with me adding it in a separate pull-request so that this one could be completed?

tools/lexicmap/lexicmap.xml
```diff
@@ -7,48 +7,80 @@
...
+#end if
+
+extract_query_ids() {
+    local input_files="\$1";
+    local query_ids="";
+    IFS=',' read -ra files <<< "\$input_files";
+    query_ids="";
+    for query_file in "\${files[@]}"; do
+        if file --mime-type "\$query_file" | grep -q "gzip"; then
+            query_ids+=\$(zcat "\$query_file" | awk '{if(/^>/) {sub(/^>/, "", \$0); print \$0" "}}');
```
bernt-matthias: Can you check if this can be done wo awk? Otherwise we need to add awk/gawk as requirement. 

https://github.com/dylanaraps/pure-bash-bible

tools/lexicmap/lexicmap.xml
```diff
@@ -7,48 +7,80 @@
...
+        else
+            query_ids+=\$(cat "\$query_file" | awk '{if(/^>/) {sub(/^>/, "", \$0); print \$0" "}}');
+        fi
+    done;
+    echo "\$query_ids";
```
bernt-matthias: If the IDs have spaces we are in trouble (in the for loop). Guess you need to return a list?

tools/lexicmap/lexicmap.xml
```diff
@@ -124,7 +156,56 @@ lexicmap search
...
+                <param name="db_opts_selector" value="db"/>
+                <param name="lexicmap_index" value="LexicMapIndexCombined" />
+            </conditional>
+            <param name="query" value="lexicmap_query.fasta.gz" />
+            <section name="advanced_settings">
+                <param name="load_whole_seeds" value="true" />
+            </section>
+            <output name="out_file" value="lexicmap_query_result.tsv" />
+        </test>
+        <!-- Test 4 - query a two local index with multiple query files -->
```
bernt-matthias: Test seems to use only one index?

Smeds: Maybe I'm a little unclear! It's true that we're not making two selections here, but LexicMapIndexCombined points to two local indexes.

tools/lexicmap/lexicmap.xml
Smeds: I have removed the dependency of awk

tools/lexicmap/lexicmap.xml
Smeds: with the new implementation we can handle spaces. But I discovered that `lexicmap search` remove anything after a space in a query id.

tools/lexicmap/lexicmap.xml
Smeds: @bernt-matthias Talked to the lexicmap developer. If a query id contains spaces the part after the space will be removed (the part after the space is considered as a description).

tools/lexicmap/lexicmap.xml
Smeds: I have updated the code to remove anything after the first space in the query id, to match how lexicmap handle the ids. Also added a note about it in the help section.

tools/lexicmap/lexicmap.xml
```diff
@@ -7,48 +7,82 @@
...
+    for ((i=0; i<\${#query_array[@]}; i++)); do
+        counter=\$((counter + 1));
+        lexicmap utils merge-search-results
+            --out-file "combined_result.\${counter}.tsv"
+            -q "\${query_array[\$i]}" lexicmap_search_result__index*.tsv
+            -j "\${GALAXY_SLOTS:-1}";
+    done &&
+    cat combined_result.*.tsv | awk 'NR==1 || $0 !~ /^query\tqlen\thits/' > "$out_file"
+#else
+    mv lexicmap_search_result__index0.tsv "$out_file"
```
bgruening:
```suggestion
    cat combined_result.*.tsv | awk 'NR==1 || $0 !~ /^query\tqlen\thits/' > '$out_file'
#else
    mv lexicmap_search_result__index0.tsv '$out_file'
```

Smeds:
@bgruening could you also merge the pull-request? I seem to be missing that button.

PR #7304, opened 2025-09-25
New tool addition: Samestr2 

xens25:
FOR CONTRIBUTOR:


tools/samestr/samestr_compare.xml
xens25, line 1: Hi @SaimMomin12 , is the requirement section correct? All tests pass locally in my terminal, but in the PR, they fail with the following error: Job in error state.. tool_id: samestr_compare, exit_code: 127, stderr: /tmp/tmp0j3m59cf/job_working_directory/000/3/tool_script.sh: line 10: samestr: command not found
. I verified that the specified version of samestr exists on Conda. Could this be caused by the requirements section or is there another file that could cause such and error?

SaimMomin12: Let me restart the CI

xens25:
Hi @SaimMomin12 could I get an initial review of the tools? I was able to solve the failing tests in the PR. I still get a lint error that I do not get in the terminal so I'm unsure on how to fix it.

tools/samestr/samestr_compare.xml
```diff
@@ -0,0 +1,203 @@
+<tool id="samestr_compare" name="SameStr Compare" version="0.1.0+galaxy0" python_template_version="3.5" profile="21.05">
```
SaimMomin12:
```suggestion
<tool id="samestr_compare" name="SameStr Compare" version="TOOL_VERSION+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
```

Add the above tokens to macros.xml file

tools/samestr/samestr_compare.xml
```diff
@@ -0,0 +1,203 @@
...
+    <description>Convert metagenomic alignments to nucleotide variant profiles for SameStr analysis</description>
+    <requirements>
+        <requirement type="package" version="1.2025.111">samestr</requirement>
+    </requirements>
```
SaimMomin12: This can go in macros.xml too

tools/samestr/samestr_compare.xml
```diff
@@ -0,0 +1,203 @@
+<tool id="samestr_compare" name="SameStr Compare" version="0.1.0+galaxy0" python_template_version="3.5" profile="21.05">
+    <description>Convert metagenomic alignments to nucleotide variant profiles for SameStr analysis</description>
```
SaimMomin12:
```suggestion
    <description>Converts metagenomic alignments to nucleotide variant profiles for SameStr analysis</description>
```

tools/samestr/samestr_compare.xml
```diff
@@ -0,0 +1,203 @@
...
+    <citations>
+        <citation type="bibtex">
+        @software{samestr,
+        title = {SameStr},
+        author = {Podlesny, Daniel and Fricke, W. Florian},
+        year = {2022},
+        url = {https://github.com/danielpodlesny/samestr},
+        license = {GNU Affero General Public License v3.0},
+        }</citation>
+    </citations>
```
SaimMomin12: This can also go to macros.xml

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,229 @@
+<tool id="samestr_convert" name="SameStr Convert" version="0.1.0+galaxy0" python_template_version="3.5" profile="21.05">
```
SaimMomin12:
```suggestion
<tool id="samestr_convert" name="SameStr Convert" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
```

tools/samestr/samestr_compare.xml
SaimMomin12: I think the description here is incorrect, pls check

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,229 @@
...
+    <description>Convert MetaPhlAn marker alignments to nucleotide variant profiles</description>
+    <requirements>
+        <requirement type="package" version="1.2025.111">samestr</requirement>
+        <requirement type="package" version="0.1.19">samtools</requirement>
+    </requirements>
```
SaimMomin12: This can also go to macros.xml

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,229 @@
...
+    --min-base-qual '$alignment.min_base_qual'
+    --min-aln-qual '$alignment.min_aln_qual'
+
+    #for $profile in $tax_profiles:
+        cp '$profile' tax_profiles_dir/'$profile.element_identifier' &&
+    #end for
+    #if $tax_profiles_extension
+        --tax-profiles-extension '$tax_profiles_extension'
+    #end if
+    $db_force
```
SaimMomin12:
```suggestion
```

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,229 @@
...
+            <param name="min_aln_len" type="integer" value="40" label="Minimum alignment length" help="Minimum alignment length." />
+            <param name="min_base_qual" type="integer" value="20" label="Minimum base quality" help="Minimum base call quality." />
+            <param name="min_aln_qual" type="integer" value="0" label="Minimum alignment quality" help="Increasing this threshold can drastically reduce the number of considered variants." />
+            <param name="min_vcov" type="integer" value="3" label="Minimum vertical coverage" help="Minimum vertical coverage." />
+        </section>
+
+        <param name="tax_profiles" type="data_collection" collection_type="list" optional="true" format="txt" label="Taxonomic profiles" help="Collection of MetaPhlAn .profile.txt files"/>
+        <param name="tax_profiles_extension" type="text" value=".profile.txt" label="Taxonomic profile extension" help="File extension of taxonomic profiles (default: .profile.txt)"/>
+
+        <param argument="--db-force" type="boolean" truevalue="--db-force" falsevalue="" label="Force database compatibility" help="Force execution, even when database version is not an exact match." />
```
SaimMomin12:
```suggestion
```

tools/samestr/samestr_db.xml
```diff
@@ -0,0 +1,115 @@
+<tool id="samestr_db" name="SameStr Database Builder" version="0.1.0+galaxy0" python_template_version="3.5" profile="21.05">
```
SaimMomin12: This can be a data manager. You can find similar here: https://github.com/galaxyproject/tools-iuc/tree/62d845456b6048e7f22ecdc2b323217e00253908/data_managers

We have a documentation (still in progress) about how to create a data manager: https://github.com/wm75/training-material/blob/edc3e900e785c66c30275807ff24aa5b2e7aaabf/topics/dev/tutorials/data-manager-architecture/tutorial.md

Maybe you can refer to it and let us also know if there are any error in the docs

tools/samestr/samestr_extract.xml
```diff
@@ -0,0 +1,183 @@
+<tool id="samestr_extract" name="SameStr Extract" version="0.1.0+galaxy0" python_template_version="3.5" profile="21.05">
```
SaimMomin12: Replace with TOKENS

tools/samestr/samestr_extract.xml
```diff
@@ -0,0 +1,183 @@
...
+    <description>Extract MetaPhlAn marker sequences from reference genomes using BLASTN</description>
+    <requirements>
+        <requirement type="package" version="1.2025.111">samestr</requirement>
+        <requirement type="package" version="1.5.5">zstd</requirement>
+    </requirements>
```
SaimMomin12: macros.xml

tools/samestr/samestr_extract.xml
```diff
@@ -0,0 +1,183 @@
...
+    --marker-dir '$database.fields.path'
+    --output-dir out_extract/
+    --aln-program '$alignment.alignment_program'
+    --marker-trunc-len '$alignment.truncate_marker'
+    $save_marker_aln
+    ]]></command>
+    <inputs>
+        <param name="reference_genomes" type="data" format="fasta,fasta.gz" multiple="true" label="Reference genomes" help="Reference genome sequences in FASTA format."/>
+
+        <param name="clade" type="text" label="Clade to process from input files." help="Names must correspond to the taxonomy of the respective database [e.g. t__SGB10068 for MetaPhlAn]">
```
SaimMomin12: Is this optional param?

tools/samestr/samestr_filter.xml
```diff
@@ -0,0 +1,231 @@
+<tool id="samestr_filter" name="SameStr Filter" version="0.1.0+galaxy0" python_template_version="3.5" profile="21.05">
```
SaimMomin12: TOKENS

tools/samestr/samestr_filter.xml
```diff
@@ -0,0 +1,231 @@
...
+    <description>SameStr Filter filters nucleotide variant profiles using coverage, sample, marker, and position-based criteria</description>
+    <requirements>
+        <requirement type="package" version="1.2025.111">samestr</requirement>
+    </requirements>
```
SaimMomin12: macros

tools/samestr/samestr_filter.xml
```diff
@@ -0,0 +1,231 @@
...
+        </section>
+
+        <section name="clade_settings" title="Clade settings">
+            <param name="clade" type="text" label="Clades to process." optional="true" help="Processing all if not specified. Names must correspond to the taxonomy of the respective database [e.g. t__SGB10068 for MetaPhlAn]"/>
+            <param name="clade_min_samples" type="integer" value="2" min="0" label="Minimum samples per clade" help="Skip clades with fewer than this number of samples (default: 2)" />
+            <param name="marker_remove" type="text" label="Markers to exclude." optional="true" help="List of Markers to remove for selected clade. Requires clade to be specified."/>
+            <param name="marker_keep" type="text" label="Markers to include." optional="true" help="List of Markers to keep for selected clade. Requires clade to be specified. Overrides Markers to remove from clade."/>
+            <param name="marker_trunc_len" type="integer" value="50" min="0" label="Marker truncation length" help="Number of nucleotides to trim from each end of a marker sequence. (default: 50)" />
```
SaimMomin12:
```suggestion
            <param argument="--clade" type="text" label="Clades to process." optional="true" help="Processing all if not specified. Names must correspond to the taxonomy of the respective database [e.g. t__SGB10068 for MetaPhlAn]"/>
            <param argument="--clade-min-samples" type="integer" value="2" min="0" label="Minimum samples per clade" help="Skip clades with fewer than this number of samples (default: 2)" />
            <param name="marker-remove" type="text" label="Markers to exclude." optional="true" help="List of Markers to remove for selected clade. Requires clade to be specified."/>
            <param argument="--marker-keep" type="text" label="Markers to include." optional="true" help="List of Markers to keep for selected clade. Requires clade to be specified. Overrides Markers to remove from clade."/>
            <param argument="--marker-trunc-len" type="integer" value="50" min="0" label="Marker truncation length" help="Number of nucleotides to trim from each end of a marker sequence. (default: 50)" />
```

tools/samestr/samestr_filter.xml
```diff
@@ -0,0 +1,231 @@
...
+        </section>
+
+        <section name="sample_variant_filtering" title="Sample Variant Filtering">
+            <param name="sample_var_min_n_vcov" type="integer" value="2" min="0" label="Minimum variant coverage (nucleotides)" help="Remove variants with nucleotide coverage below this threshold (default: 2)." />
+            <param name="sample_var_min_f_vcov" type="float" value="0.05" min="0" max="1" label="Minimum variant coverage (fraction)" help="Remove variants with coverage below this fraction (default: 0.05)." />
+        </section>
+
+        <section name="sample_position_filtering" title="Sample Position Filtering">
+            <param name="sample_pos_min_n_vcov" type="integer" value="1" min="0" label="Minimum position coverage (nucleotides)" help="Remove positions with nucleotide coverage below this threshold (default: 1)." />
+            <param name="sample_pos_min_sd_vcov" type="float" value="3.0" min="0" label="Position coverage standard deviation cutoff" help="Remove positions with coverage outside mean ± this standard deviation (default: 3.0)." />
```
SaimMomin12: Please change name to arguments overall

tools/samestr/samestr_merge.xml
```diff
@@ -0,0 +1,112 @@
+<tool id="samestr_merge" name="SameStr Merge" version="0.1.0+galaxy0" python_template_version="3.5" profile="21.05">
```
SaimMomin12: TOKENS

tools/samestr/samestr_merge.xml
```diff
@@ -0,0 +1,112 @@
...
+    <description>Merge nucleotide variant profiles obtained from metagenomic samples for filtering and pairwise comparisons</description>
+    <requirements>
+        <requirement type="package" version="1.2025.111">samestr</requirement>
+    </requirements>
```
SaimMomin12: macros

tools/samestr/samestr_stats.xml
```diff
@@ -0,0 +1,111 @@
+<tool id="samestr_stats" name="SameStr Stats" version="0.1.0+galaxy0" python_template_version="3.5" profile="21.05">
```
SaimMomin12: TOKENS

tools/samestr/samestr_stats.xml
```diff
@@ -0,0 +1,111 @@
...
+    <description>Generate statistics for SameStr variant profiles including coverage and diversity metrics</description>
+    <requirements>
+        <requirement type="package" version="1.2025.111">samestr</requirement>
+    </requirements>
```
SaimMomin12: macros

tools/samestr/samestr_summarize.xml
```diff
@@ -0,0 +1,146 @@
+<tool id="samestr_summarize" name="SameStr Summarize" version="0.1.0+galaxy0" python_template_version="3.5" profile="21.05">
```
SaimMomin12: TOKENS

tools/samestr/samestr_summarize.xml
```diff
@@ -0,0 +1,146 @@
...
+    <description>Identify and summarize shared bacterial strains across samples using variant profile similarity</description>
+    <requirements>
+        <requirement type="package" version="1.2025.111">samestr</requirement>
+    </requirements>
```
SaimMomin12: macros.xml

tools/samestr/samestr_summarize.xml
```diff
@@ -0,0 +1,146 @@
...
+    --aln-pair-min-similarity '$aln_pair_min_similarity'
+    ]]></command>
+    <inputs>
+        <param name="database" type="select" label="Marker database" help="MetaPhlAn clade marker database.">
+            <options from_data_table="metaphlan_database_versioned">
+                <filter type="sort_by" column="2" />
+            </options>
+        </param>
+
+        <param name="input_dir" type="data_collection" collection_type="list" format="tabular" label="Samestr compare outputs" help="Add the output directory from samestr compare. Must contain pairwise comparison of alignment files (.fraction.txt, .overlap.txt)"/>
```
SaimMomin12: Ideally, you could tar.gz the pairwise comparison of alignment files (.fraction.txt, .overlap.txt) in samestr compare and use it here. I dont think so, data_collection would work here.

tools/samestr/samestr_summarize.xml
```diff
@@ -0,0 +1,146 @@
...
+        <param name="tax_profiles" type="data_collection" collection_type="list" format="txt" label="Taxonomic profiles" help="Collection of MetaPhlAn .profile.txt files"/>
+        <param name="tax_profiles_extension" type="text" value=".profile.txt" label="Taxonomic profile extension" optional="true" help="File extension of taxonomic profiles (default: .profile.txt)"/>
```
SaimMomin12: arguments

tools/samestr/samestr_summarize.xml
```diff
@@ -0,0 +1,146 @@
...
+
+        <section name="arguments" title="Summary threshold arguments">
+            <param name="aln_pair_min_overlap" type="integer" value="5000" min="0" label="Minimum overlap for comparison" help="Minimum number of overlapping positions required in both alignments to evaluate similarity (default: 5000)." />
+            <param name="aln_pair_min_similarity" type="float" value="0.999" min="0" max="1" label="Minimum similarity for shared strains" help="Minimum pairwise alignment similarity to call shared strains (default: 0.999)." />
```
SaimMomin12: arguments

SaimMomin12 (commented):
Hi @xens25

Following up. I’ve added a few preliminary inline comments. We’ll need to build a data manager for `samestr_db`, which can then be reused downstream across the other tools.

tools/samestr/samestr_extract.xml
xens25: I think they are obligatory since in the samestr extract help section they are both marked without [ ]

xens25:
Hi @SaimMomin12 , I have now added the samestr_db as a data manager. Let me know in case there's further things to modify. Thanks in advanced.

SaimMomin12:
> Hi @SaimMomin12 , I have now added the samestr_db as a data manager. Let me know in case there's further things to modify. Thanks in advanced.

@xens25 Thanks, could you create a seperate PR for the DM?

xens25:
Hi @SaimMomin12 . The DM has been implemented back. The clade argument in the DM has been removed. I added the script [customizemetadata.py](https://github.com/galaxyproject/tools-iuc/pull/7304/changes#diff-cd08602c40d936633ab871b2a3db705a495f4aa9d69f0eddc78ddd4daf82110f) taken from https://github.com/galaxyproject/tools-iuc/tree/main/tools/metaphlan  to convert Galaxy's MetaPhlAn database format (.json) to .pkl which is what samestr db requires.  For the motus DM, it seems that it takes the existing mOTUS database files (.tsv and .fasta) which is also what samestr db needs. I haven't changed yet the samestr_db in tools/samestr in case the DM is not yet correct.

data_managers/data_manager_samestr_db/data_manager/macros.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0"?>
```
SaimMomin12:
```suggestion
```

data_managers/data_manager_samestr_db/data_manager/macros.xml
```diff
@@ -0,0 +1,27 @@
...
+<macros>
+    <token name="@TOOL_VERSION@">1.2025.111</token>
+    <token name="@VERSION_SUFFIX@">0</token>
+    <token name="@PROFILE@">21.05</token>
```
SaimMomin12:
```suggestion
    <token name="@PROFILE@">25.0</token>
```

data_managers/data_manager_samestr_db/.shed.yml
```diff
@@ -0,0 +1,13 @@
+name: samestr
+owner: iuc
+type: unrestricted
+description: Strain-level analysis and comparison of metagenomic samples
+long_description: |
+  SameStr is a tool for identifying shared microbial strains across metagenomic
+  samples using clade-specific markers from MetaPhlAn or mOTUs. Unlike methods
+  that only consider majority variants, SameStr analyzes all nucleotide variants
+  in alignments to detect strain sharing with higher sensitivity.
+remote_repository_url: https://github.com/danielpodlesny/samestr
```
SaimMomin12:
```suggestion
remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/main/data_managers/data_manager_samestr_db
```

data_managers/data_manager_samestr_db/data_manager_conf.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0"?>
```
SaimMomin12:
```suggestion
```

data_managers/data_manager_samestr_db/tool_data_table_conf.xml.sample
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0"?>
```
SaimMomin12:
```suggestion
```

data_managers/data_manager_samestr_db/tool_data_table_conf.xml.sample
```diff
@@ -0,0 +1,16 @@
...
+<tables>
+    <table name="samestr_db" comment_char="#">
+        <columns>value, dbkey, name, path, clade, version_file</columns>
```
SaimMomin12: do we still need clade here?

tools/samestr/.shed.yml
```diff
@@ -0,0 +1,14 @@
+name: samestr
+owner: iuc
+type: unrestricted
+description: Strain-level analysis and comparison of metagenomic samples
+long_description: |
+  SameStr is a tool for identifying shared microbial strains across metagenomic
+  samples using clade-specific markers from MetaPhlAn or mOTUs. Unlike methods
+  that only consider majority variants, SameStr analyzes all nucleotide variants
+  in alignments to detect strain sharing with higher sensitivity.
+remote_repository_url: https://github.com/danielpodlesny/samestr
```
SaimMomin12:
```suggestion
remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/main/tools/samestr
```

tools/samestr/macros.xml
```diff
@@ -0,0 +1,27 @@
+<macros>
+        <token name="@TOOL_VERSION@">1.2025.111</token>
+        <token name="@VERSION_SUFFIX@">0</token>
+        <token name="@PROFILE@">21.05</token>
```
SaimMomin12:
```suggestion
        <token name="@PROFILE@">25.1</token>
```

tools/samestr/macros.xml
```diff
@@ -0,0 +1,27 @@
...
+        <xml name="citations">
+                <citations>
+                        <citation type="bibtex">
+                @software{samestr,
+                title = {SameStr},
+                author = {Podlesny, Daniel and Fricke, W. Florian},
+                year = {2022},
+                url = {https://github.com/danielpodlesny/samestr},
+                license = {GNU Affero General Public License v3.0},
+                }
```
SaimMomin12: You can use DOI here:  https://doi.org/10.1186/s40168-022-01251-w

tools/samestr/samestr_compare.xml
```diff
@@ -0,0 +1,230 @@
...
+            <discover_datasets pattern="(?P&lt;designation&gt;.+\.fraction\.txt)$" directory="out_compare" format="tabular"/>
+        </collection>
+
+        <collection name="msa_files" type="list" label="${tool.name} on ${on_string}: Dominant variant alignments">
+            <discover_datasets pattern="(?P&lt;designation&gt;.+\.msa\.fa)$" directory="out_compare" format="fasta"/>
+        </collection>
+
+    </outputs>
+    <tests>
+        <test>
```
SaimMomin12: Add expected_num_outputs for each test

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,150 @@
...
+        </repeat>
+
+        <param name="database" type="select" label="SameStr database" help="SameStr database built from MetaPhlAn markers">
+            <options from_data_table="samestr_db">
+                <filter type="sort_by" column="2" />
+            </options>
+        </param>
+        
+        <section name="alignment" title="Alignment Parameters">
+            <param name="min_aln_identity" type="float" value="0.9" min="0" max="1" label="Percent identity" help="Minimum percent identity in alignment." />
```
SaimMomin12:
```suggestion
            <param argument="--min-aln-identity" type="float" value="0.9" min="0" max="1" label="Percent identity" help="Minimum percent identity in alignment." />
```

Please use arguments wherever required in all tools.

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,150 @@
...
+                </assert_contents>
+            </output>
+
+            <output name="stats_file">
+                <assert_contents>
+                    <has_text text="alignment"/>
+                    <has_text text="genome"/>
+                    <has_text text="mean_cov"/>
+                    <has_text text="n_sites"/>
+                    <has_text text="f_covered"/>
```
SaimMomin12: Some important content of the stats file would be rather beneficial

SaimMomin12 (commented):
Cool work @xens25! Some preliminary comments inline

tools/samestr/samestr_compare.xml
```diff
@@ -0,0 +1,230 @@
...
+    #end for
+    --output-dir out_compare/
+    $variant_comparison_settings.dominant_variants
+    $variant_comparison_settings.dominant_variants_added
+    $variant_comparison_settings.dominant_variants_msa
+    &&
+
+    
+    cd out_compare &&
+    tar -czf compare_output.tgz *.overlap.txt *.closest.txt *.fraction.txt
```
wm75: @xens25 I know this goes back to advice from @SaimMomin12, but I don't think this is a very Galactic way of approaching this.
All three output types are gathered in output collections and these should be three inputs to samestr_summarize. I guess you were on the right track with your initial "data_collection" approach.
You'd use three of these params in samestr_summarize, then link all their elements under appropriate names into `out_compare`. If this becomes convoluted, it may be best to use a configfile that you call as a script from the command section, similar to but probably more complicated than e.g.:

https://github.com/galaxyproject/tools-iuc/blob/8c30cf098c8d30e2f8a7916de329911c4b9ed08c/tools/cojac/cooc_mutbamscan.xml#L57-L69

It sure makes life a bit more complicated for you and for the user, but it's also not good to have arbitrary tar.gz archives hanging around in histories that you cannot interact with in Galaxy.

I just started going through all the code though so it's also possible I just don't realize that this cannot work so let me know if you think this comment doesn't make sense.

tools/samestr/tool-data/metaphlan_database.loc.sample
```diff
@@ -0,0 +1,4 @@
+#This is a sample file distributed with Galaxy that enables tools
+#to use a directory of metagenomics files.  
+#file has this format (white space characters are TAB characters)
+#02_16_2014	 MetaPhlAn2 clade-specific marker genes	db_v20	/path/to/data	SGB_or_legacy
```
wm75: This looks wrong?

The same_str tools aren't using the metaphlan data table, are they?

What you should have instead is a .loc.sample file for the samestr_db data table.

data_managers/data_manager_samestr_db/tool-data/samestr_db.loc.sample
```diff
@@ -0,0 +1 @@
+#value	dbkey	name	path	version_file
```
wm75: The DM uses *three* data tables and all three should have .loc.sample files here.

data_managers/data_manager_samestr_db/motus_db_versioned.loc
```diff
@@ -0,0 +1,2 @@
+#value	version	name	path
+motus_test	3.1.0	"mOTUs Test Database"	${__HERE__}/
```
wm75: This file should not be here. Guess just a duplicate from the one in test-data?

wm75 (changes requested):

tools/samestr/.shed.yml
```diff
@@ -0,0 +1,14 @@
...
+long_description: |
+  SameStr is a tool for identifying shared microbial strains across metagenomic
+  samples using clade-specific markers from MetaPhlAn or mOTUs. Unlike methods
+  that only consider majority variants, SameStr analyzes all nucleotide variants
+  in alignments to detect strain sharing with higher sensitivity.
+remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/main/tools/samestr
+homepage_url: https://github.com/danielpodlesny/samestr
+categories:
+  - Metagenomics
+  - Strain typing
```
wm75: This one is the reason for the failed linting.
You can only use categories that exist in the toolshed.

tools/samestr/test-data/test_database_versioned.loc.test
```diff
@@ -0,0 +1,11 @@
...
+# - value (Galaxy records this in the Galaxy DB)
+# - name (Galaxy shows this in the UI)
+# - dbkey
+# - path (folder name containing the Kraken DB)
+# - db version (whether it is SGB of Metaphlan v4 or not)
+test-db-20210409	"Test Database"	test-db	${__HERE__}/test-db	SGB
+# the TOY DB should be called in tests with the hidden test parameter enables this will make Metaphlan
+# download the data to the JWD
+mpa_vJan21_TOY_CHOCOPhlAnSGB	mpa_vJan21_TOY_CHOCOPhlAnSGB	mpa_vJan21_TOY_CHOCOPhlAnSGB_202103	${__HERE__}/samestr-db	SGB
+# mpa_vJun23_CHOCOPhlAnSGB_202403	mpa_vJun23_CHOCOPhlAnSGB_202403	mpa_vJun23_CHOCOPhlAnSGB_202403	${__HERE__}/mpa_vJun23_CHOCOPhlAnSGB_202403/	SGB
```
wm75: This file seems unnecessary.

tools/samestr/test-data/tool_data_table_conf.xml.test
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0"?>
+<tables>
+    <table name="samestr_db" comment_char="#">
+        <columns>value, dbkey, name, path, clade, version_file</columns>
+        <file path="${__HERE__}/samestr_db.loc" />
+    </table>
+</tables>
```
wm75: This file looks unnecessary, too. Only the outer one with the same name should be relevant.

tools/samestr/test-data/samestr_db.loc
```diff
@@ -0,0 +1,2 @@
+#value	dbkey	name	path	version_file
+mpa_test-db-without-one-marker	mpa_test-db-without-one-marker	"Test Database from without-one-marker"	${__HERE__}/samestr_db/	""	${__HERE__}/mpa_latest
```
wm75: The record looks weird.
What's with the empty string in it? and where is that mpa_latest file that's mentioned?

data_managers/data_manager_samestr_db/data_manager_conf.xml
```diff
@@ -0,0 +1,27 @@
...
+                        <target base="${GALAXY_DATA_MANAGER_DATA_PATH}">samestr_db/</target>
+                    </move>
+                    <value_translation>${GALAXY_DATA_MANAGER_DATA_PATH}/samestr_db/</value_translation>
+                    <value_translation type="function">abspath</value_translation>
+                </column>
+                <column name="version_file" output_ref="out_file">
+                    <move type="file" relativize_symlinks="True">
+                        <src>${out_file.extra_files_path}/samestr_db/${version_file}</src>
+                        <target base="${GALAXY_DATA_MANAGER_DATA_PATH}/samestr_db/">${version_file}</target>
+                    </move>
```
wm75: I'm not 100% sure what this is good for. What's the final file tree in samestr_db that you're trying to generate?
Can you explain?

data_managers/data_manager_samestr_db/data_manager/macros.xml
```diff
@@ -0,0 +1,33 @@
...
+
+
+    <xml name="requirements">
+        <requirements>
+            <requirement type="package" version="@TOOL_VERSION@">samestr</requirement>
+        </requirements>
+    </xml>
+        
+    <xml name="citations">
+        <citations>
```
wm75, line 14: Also include the doi-based citation from the tool macro.

wm75:
Ok, enough comments for a first round. Try to address them and I'll continue reviewing later again.

data_managers/data_manager_samestr_db/data_manager_conf.xml
xens25: The structure should be as follows:

(base) xenia@LAPTOP-MN3GGLUF:~/tools-iuc/tools/samestr/test-data/samestr_db$ tree
.
├── command_log.json
├── db_clades.json
├── db_manifest.json
├── db_markers
│   └── t__Propio
│       └── nib
│           └── act
│               ├── t__Propionibacterium_phage_PAS50.contig_map.txt.gz
│               ├── t__Propionibacterium_phage_PAS50.gene_file.txt.gz
│               ├── t__Propionibacterium_phage_PAS50.markers.fa.gz
│               └── t__Propionibacterium_phage_PAS50.positions.txt.gz
├── db_taxonomy.tsv

The files inside db_markers depends on the input files used.  The version file would be the mpa_latest (metaphlan) or db_mOTU_versions. It contains the version of the database and it's an obligatory argument for the samestr db command.

xens25:
@wm75 Thank you for your initial comments, I have uploaded a new version with your suggestions. Let me know if there´s further modifications to be done.

data_managers/data_manager_samestr_db/test-data/samestr_db.loc
```diff
@@ -1,3 +1,5 @@
 #value	dbkey	name	path	version_file
 mpa_test-db-without-one-marker	mpa_test-db-without-one-marker	"Test Database with Multiple Markers"	${__HERE__}/	${__HERE__}/mpa_latest
 motus_test	motus_test	"mOTUs Test Database"	${__HERE__}/motus_test_DB_non_functional	${__HERE__}/motus_test_DB_non_functional/db_mOTU_versions
+motus_test	motus_test	mOTUs Test Database	/tmp/tmplep3xyy7/galaxy-dev/tool-data/samestr_db	/tmp/tmplep3xyy7/galaxy-dev/tool-data/samestr_db/db_mOTU_versions
+mpa_test-db-without-one-marker	mpa_test-db-without-one-marker	Test Database with Multiple Markers	/tmp/tmplep3xyy7/galaxy-dev/tool-data/samestr_db	/tmp/tmplep3xyy7/galaxy-dev/tool-data/samestr_db/mpa_latest
```
wm75: before we merge this, remember to remove lines like these, which get added by your local planemo test runs.

data_managers/data_manager_samestr_db/data_manager/samestr_db.xml
```diff
@@ -0,0 +1,150 @@
+<tool id="samestr_db" name="SameStr Database Builder" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" tool_type="manage_data" profile="@PROFILE@">
+    <description>Build SameStr database from MetaPhlAn or mOTUs markers</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements"/>
+    <command detect_errors="exit_code"><![CDATA[
+        mkdir -p '$out_file.extra_files_path'/samestr_db/ &&
```
wm75: you're using this path three times verbatim in the command section -> should become a cheetah variable, too.

data_managers/data_manager_samestr_db/data_manager_conf.xml
wm75: The moving over of the `version_file` doesn't seem to be working though. It's a required input for samestr db, but not produced as an output again?
The end result is that the version_file column of data table records points to a file that doesn't exist.
I see why it may be interesting to record the file, you should copy it over from the original db to the output folder of samestr db as part of the data manager xml.

wm75: Another issue is that currently all recorded paths (path and version_file) differ from the stored tool-data by one level of nesting. This is what you're installing into:

```
samestr_db
└── samestr_db
    ├── command_log.json
    ├── db_clades.json
    ├── db_manifest.json
    ├── db_markers
...
```

but in your paths, you're only recording down to the first samestr_db folder.
What you *should* do: under the outer samestr_db create subfolders named like the `value` in the data table. This is necessary or your DM will always write to the *same* subfolder and overwrite previously installed data.

data_managers/data_manager_samestr_db/tool_data_table_conf.xml.sample
```diff
@@ -0,0 +1,15 @@
+<tables>
+    <table name="samestr_db" comment_char="#">
+        <columns>value, dbkey, name, path, version_file</columns>
```
wm75: I don't see any added value in having the `dbkey` column which only repeats the content of the `value` column.
What you should add instead is a `samestr_version` column, in which for now you can record just a `1`.
Background there is that tools like this one are very likely to change their DB structure at some point and, unfortunately, it is not possible to add a column to an existing data table.

wm75 (changes requested):
Another round of comments/suggestions all around the data tables again.
Sorry that they are just so complicated.

xens25:
Hi @wm75 , thanks for the comments. I've updated your suggestions regarding the data tables. Let me know if there's further modifications for them or other parts of the code. Thank you in advance.

wm75:
@xens25 the DM seems to work now as intended. It's funny that one of the example tools in https://github.com/galaxyproject/training-material/pull/6642 uses `<src>` instead of the correct `<source>` tag. The wrong tag simply gets ignored it seems. I've corrected the example accordingly over in the other PR now.

I'm going to review the actual tool next.

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,155 @@
...
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements"/>
+    <command detect_errors="exit_code"><![CDATA[
+    mkdir -p out_convert/ tax_profiles_dir/ &&
+
+    echo "Database path: '$database.fields.path'" >&2 &&
+    echo "Database files:" >&2 &&
+    ls -la '$database.fields.path' >&2 &&
```
wm75: Should be removed eventually

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,155 @@
...
+
+    #for $i, $input in enumerate($input_file): 
+        #if str($input.alignment.ext) == 'sam':
+            bzip2 -c '$input.alignment' > 'file_${i}.sam.bz2' &&
```
wm75: why this explicit step? I thought the tool handles SAM format itself?

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,155 @@
...
+    --min-aln-len '$alignment.min_aln_len'
+    --min-base-qual '$alignment.min_base_qual'
+    --min-aln-qual '$alignment.min_aln_qual'
+
+    #for $profile in $tax_profiles:
+        cp '$profile' tax_profiles_dir/'$profile.element_identifier' &&
+    #end for
+    #if $tax_profiles_extension
+        --tax-profiles-extension '$tax_profiles_extension'
+    #end if
```
wm75: you don't need this argument because you're controlling the file extension in this wrapper.

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,155 @@
...
+    --output-dir out_convert/
+    --min-vcov '$alignment.min_vcov'
+    --min-aln-identity '$alignment.min_aln_identity'
+    --min-aln-len '$alignment.min_aln_len'
+    --min-base-qual '$alignment.min_base_qual'
+    --min-aln-qual '$alignment.min_aln_qual'
+
+    #for $profile in $tax_profiles:
+        cp '$profile' tax_profiles_dir/'$profile.element_identifier' &&
+    #end for
```
wm75: I guess there's sort of a duplication going on here: you're already collecting `input_file.profile`.

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,155 @@
...
+    #if $tax_profiles_extension
+        --tax-profiles-extension '$tax_profiles_extension'
+    #end if
+    ]]></command>
+    <inputs>
+
+        <repeat name="input_file" title="Samples" min="1">
+            <param name="alignment" type="data" format="sam,bam" label="Alignment file" help="MetaPhlAn or mOTUS marker alignment."/>
+            <param name="profile" type="data" format="txt" label="Taxonomic profile" help="MetaPhlAn or mOTUS taxonomic profile."/>
+        </repeat>
```
wm75: I think that's the wrong design for this wrapper.

Unless there is some information borrowing across samples going on in `convert` code, this wrapper should consume exactly *one* alignment and *one* profile file, and let Galaxy handle the case of multiple files or a collection for efficient parallelization.

But please check the logic of the convert command to see if independent sample processing is indeed what's going on inside!

tools/samestr/samestr_convert.xml
wm75: from a quick glance over the source code my suspicion is that it converts sam to bam internally  first.
If that's so, just disallow sam and let Galaxy do the conversion.

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,155 @@
...
+
+        <param name="database" type="select" label="SameStr database" help="SameStr database built from MetaPhlAn or mOTUS markers">
+            <options from_data_table="samestr_db">
+                <filter type="sort_by" column="2" />
```
wm75, line 43: with the new samestr_version column, you can now filter on records that specify "1" there.

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,155 @@
...
+        <param name="tax_profiles" type="data_collection" collection_type="list" optional="true" format="txt" label="Taxonomic profiles" help="Collection of MetaPhlAn or mOTUS .profile.txt files"/>
+        <param argument="--tax-profiles-extension" type="text" value=".profile.txt" label="Taxonomic profile extension" help="File extension of taxonomic profiles (default: .profile.txt)"/>
+    </inputs>
+    <outputs>
+        <data name="command_log" format="json" from_work_dir="out_convert/command_log.json" label="${tool.name} on ${on_string}: Command log"/>
+        <data name="stats_file" format="tabular" from_work_dir="out_convert/*/*.aln_stats.txt" label="${tool.name} on ${on_string}:Alignment statistics"/>
+        
+        <collection name="npz_files" type="list" label="${tool.name} on ${on_string}: Clade NPZ files">
+            <discover_datasets pattern="(?P&lt;designation&gt;.+)\.npz$" directory="out_convert" recurse="true" format="npz"/>
+        </collection>
```
wm75: If, indeed, you feed in only single files, this will become a simple dataset output.

tools/samestr/samestr_extract.xml
```diff
@@ -0,0 +1,173 @@
...
+    #end for
+    --clade '$clade'
+    --marker-dir '$database.fields.path'
+    --output-dir out_extract/
+    --aln-program '$alignment.alignment_program'
+    --marker-trunc-len '$alignment.truncate_marker'
+    $save_marker_aln
+    ]]></command>
+    <inputs>
+        <param name="reference_genomes" type="data" format="fasta,fasta.gz" multiple="true" label="Reference genomes" help="Reference genome sequences in FASTA format."/>
```
wm75: same design issue as with the convert wrapper. Do not consume multiple files unnecessarily, but leave parallelization over multiple inputs up to the Galaxy scheduling framework.

Only if a single run of a tool needs to see all inputs to do its job is the current way justified.
In every other case you want single jobs to work on single inputs.

tools/samestr/samestr_extract.xml
```diff
@@ -0,0 +1,173 @@
...
+
+        <param argument="--clade" type="text" label="Clade to process from input files." help="Names must correspond to the taxonomy of the respective database [e.g. t__SGB10068 for MetaPhlAn]">
+            <validator type="empty_field" message="Please enter a clade name."/>
+        </param>
+        
+        <param name="database" type="select" label="SameStr database" help="SameStr database built from MetaPhlAn or mOTUS markers">
+            <options from_data_table="samestr_db">
+                <filter type="sort_by" column="2" />
```
wm75, line 35: additional filtering

tools/samestr/samestr_extract.xml
```diff
@@ -0,0 +1,173 @@
...
+
+        <collection name="marker_alignments" type="list" label="${tool.name} on ${on_string}: Marker alignments">
+            <filter>save_marker_aln</filter>
+            <discover_datasets pattern="(?P&lt;designation&gt;.+\.marker_aligned\.fasta)$" directory="out_extract" format="fasta"/>
+        </collection>
+        
+        <collection name="msa_files" type="list" label="${tool.name} on ${on_string}: MSA files">
+            <filter>save_marker_aln</filter>
+            <discover_datasets pattern="(?P&lt;designation&gt;.+\.msa\.fa)$" directory="out_extract" format="fasta"/>
+        </collection>
```
wm75: all of this will be simplified down to simple dataset outputs

tools/samestr/samestr_merge.xml
```diff
@@ -0,0 +1,108 @@
...
+    <expand macro="requirements"/>
+    <command detect_errors="exit_code"><![CDATA[
+    mkdir -p out_merge/ &&
+
+    #for $i, $file in enumerate($input_files):
+        #if str($file.ext) in ['npz','npy']:
+            cp '$file' '$file.element_identifier' &&
+        #else if str($file.ext) == 'npy.gz':
+            cp '$file' '$file.element_identifier' &&
+        #end if
```
wm75: in this form the if block is pointless.
Are you sure you don't need to preserve the file extensions?

Also, again, do you have a specific reason why you're copying the data instead of symlinking?

tools/samestr/samestr_merge.xml
```diff
@@ -0,0 +1,108 @@
...
+    #if $clade:
+        --clade '$clade'
+    #end if
+    ]]></command>
+    <inputs>
+        <param name="input_files" type="data" format="npz,npy" multiple="true" label="Input SNV profiles" help="NPZ/NPY files from samestr convert or extract. Select multiple files to merge."/>
+
+        <param name="database" type="select" label="SameStr database" help="SameStr database built from MetaPhlAn or mOTUS markers">
+            <options from_data_table="samestr_db">
+                <filter type="sort_by" column="2" />
```
wm75, line 27: additional filtering

tools/samestr/samestr_merge.xml
```diff
@@ -0,0 +1,108 @@
...
+    <outputs>
+        <data name="command_log" format="json" from_work_dir="out_merge/command_log.json" label="${tool.name} on ${on_string}: Command log"/>
+
+        <collection name="npz_files" type="list" label="${tool.name} on ${on_string}: Merged NPZ files">
+            <discover_datasets pattern="(?P&lt;designation&gt;.+\.npz)$" directory="out_merge" format="npz"/>
+        </collection>
+
+        <collection name="sample_name" type="list" label="${tool.name} on ${on_string}: Sample identifiers">
+            <discover_datasets pattern="(?P&lt;designation&gt;.+\.names\.txt)$" directory="out_merge" format="txt"/>
+        </collection>
```
wm75: does the tool really produce multiple outputs of each kind?
The name `merge` suggests otherwise?

tools/samestr/samestr_merge.xml
```diff
@@ -0,0 +1,108 @@
...
+    --input-files
+    #for $file in $input_files:
+        '$file.element_identifier'
+    #end for
+    #if $clade:
+        --clade '$clade'
+    #end if
+    ]]></command>
+    <inputs>
+        <param name="input_files" type="data" format="npz,npy" multiple="true" label="Input SNV profiles" help="NPZ/NPY files from samestr convert or extract. Select multiple files to merge."/>
```
wm75: yes, this time that's justified usage of multiple inputs.

tools/samestr/samestr_filter.xml
```diff
@@ -0,0 +1,230 @@
...
+    #if $samples_min_m_vcov:
+        --samples-min-m-vcov '$samples_min_m_vcov'
+    #end if
+    --global-pos-min-n-vcov '$global_pos_min_n_vcov'
+    #if $global_pos_min_f_vcov:
+        --global-pos-min-f-vcov '$global_pos_min_f_vcov'
+    #end if
+    ]]></command>
+    <inputs>
+        <param name="input_files" type="data" format="npz,npy" multiple="true" label="Input SNV profiles" help="NPZ/NPY files from samestr merge."/>
```
wm75: Investigate how this tool really works. It's acting downstream of merge so if that one only produces one merged output, I guess this wrapper should only consume one such merged npz/npy dataset.

tools/samestr/samestr_filter.xml
```diff
@@ -0,0 +1,230 @@
...
+        <param name="input_names" type="data" format="txt" multiple="true" label="Sample identifiers" help="Text files listing the sample/gene identifiers."/>
+
+        <param name="database" type="select" label="SameStr database" help="SameStr database built from MetaPhlAn or mOTUS markers">
+            <options from_data_table="samestr_db">
+                <filter type="sort_by" column="2" />
```
wm75, line 52: additional filtering

tools/samestr/samestr_filter.xml
```diff
@@ -0,0 +1,230 @@
...
+    <outputs>
+        <data name="command_log" format="json" from_work_dir="out_filter/command_log.json" label="${tool.name} on ${on_string}: Command log"/>
+
+        <collection name="npz_files" type="list" label="${tool.name} on ${on_string}: Filtered NPZ files">
+            <discover_datasets pattern="(?P&lt;designation&gt;.+\.npz)$" directory="out_filter" format="npz"/>
+        </collection>
+
+        <collection name="sample_name" type="list" label="${tool.name} on ${on_string}: Sample identifiers">
+            <discover_datasets pattern="(?P&lt;designation&gt;.+\.names\.txt)$" directory="out_filter" format="txt"/>
+        </collection>
```
wm75: If only one input above, then only simple outputs here.

wm75 (changes requested):

wm75:
like for `filter`, you should also investigate whether `stats`, and `compare` are really meant to work with several inputs or just with one merged input.

Similarly, for `summarize` you need to find out whether `compare` produces more than one output of each kind and that determines whether `summarize` needs multi-inputs.

wm75:
Oh and before implementing any of my suggestions please pull my commits first to avoid merge conflicts.

tools/samestr/samestr_convert.xml
xens25: Yes, that's correct it handles the conversion internally. I initially thought of implementing sam, sam.bz2 and bam, since that's also what the original samestr convert allows as input. But at the end the file that matters is the bam file. I will modify it to only allow bam for now.

tools/samestr/samestr_convert.xml
xens25: Yes, that's correct. Samestr convert can process either one file at a time or multiple ones but it doesn't need to see all files at the same time. I will modify to accept only one alignment and profile at a time.

xens25:
Hi, I needed to get the original samestr_db folder to test some things for the samestr tools, so I tried planemo test with  --no_cleanup as I usually did.  I noticed that now I don't see anything inside samestr_db , whereas before I was able to see the markers and the json files inside. This is what I see now:

(base) xenia@LAPTOP-MN3GGLUF:/tmp/tmpwam0677d/job_working_directory/000/2/outputs/dataset_df2246ff-67c6-4c41-a31c-cd127ec666aa_files/samestr_db$ ls -la
total 8
drwxr-xr-x 2 xenia xenia 4096 Mar 22 16:35 .
drwxr-xr-x 3 xenia xenia 4096 Mar 22 16:35 ..

Is this normal or should the files be in there? Just wanted to make sure everything is still correct with the data manager.

tools/samestr/samestr_extract.xml
xens25: In this case, looking at the samestr extract original code we have: 

https://github.com/danielpodlesny/samestr/blob/main/samestr/extract/ref2freq.py#L260

I think this collects all the fasta input files and then makes a single BLAST database from all of them which is then used to search against all markers at once. In this case, would it be correct to leave it with multiple="true" ?

tools/samestr/samestr_merge.xml
xens25: I think based on this file it looks like it's merging by clade: https://github.com/danielpodlesny/samestr/blob/main/samestr/merge/freq2freqs.py
so if there's different files (from extract and or convert) containing the same clade then it merges those files into a single npz output, so then we would end with multiple npz files depending on how many clades we have. Do we use collection in that case or something else?

wm75: Thanks for investigating. Yes, collections are the correct output types here :+1:

tools/samestr/samestr_extract.xml
wm75: Well spotted, thanks! Yes, that's a very good reason for using multiple="true".

wm75:
> Hi, I needed to get the original samestr_db folder to test some things for the samestr tools, so I tried planemo test with --no_cleanup as I usually did. I noticed that now I don't see anything inside samestr_db , whereas before I was able to see the markers and the json files inside. This is what I see now:
> 
> (base) xenia@LAPTOP-MN3GGLUF:/tmp/tmpwam0677d/job_working_directory/000/2/outputs/dataset_df2246ff-67c6-4c41-a31c-cd127ec666aa_files/samestr_db$ ls -la total 8 drwxr-xr-x 2 xenia xenia 4096 Mar 22 16:35 . drwxr-xr-x 3 xenia xenia 4096 Mar 22 16:35 ..
> 
> Is this normal or should the files be in there? Just wanted to make sure everything is still correct with the data manager.

`job_working_directory/000/2/outputs/dataset_df2246ff-67c6-4c41-a31c-cd127ec666aa_files/` above is the `extra_files_path` that the DM initially deposits its data into.
The empty folder inside of it is the expected result I would say because Galaxy, at the end of the DM run will `<move>` the contents of that directory to its final location. You can check the generated data table record to learn about this location and get the DB contents from there.
If you previously had data left there, then something wasn't fully working at that point. Maybe as a side-effect of using `<src>` instead of `<source>` in the data_manager_conf.xml.

xens25:
@wm75 I updated the comments, let me know if there's still things to modify. Also, here is a brief summary of the tools and the structure I took:

- **convert**: takes single BAM file and outputs a collection of NPZ files (one NPZ for each clade found).
- **extract**: takes collection of FASTA files and outputs collection of NPZ files (one for each genome found), a collection of marker alignments (one file for each marker found)
- **merge**: takes collection of NPZ files from multiple samples (convert and/or extract) and outputs collection of NPZ files (one for each clade)
- **filter**: takes single NPZ and name file and outputs a single filtered NPZ and name file. This would be repeated for each clade in Galaxy.
- **stats**: takes single NPZ and name file and outputs single stats file. This would be repeated for each clade in Galaxy.
- **compare**: takes single NPZ and name file and outputs several individual files. This would be repeated for each clade in Galaxy.
- **summarize**: takes a collection of all the overlap, close, and fraction files of the out_compare folder (meaning of all clades, after all the galaxy runs of compare) and a collection of the original taxonomic profiles from all samples (the ones also used as input in convert). And it outputs single  summary files.

wm75:
@xens25 all this sounds very reasonable! Thanks a lot for this work.
Now give me two days for testing everything carefully.

tools/samestr/samestr_extract.xml
```diff
@@ -0,0 +1,171 @@
...
+        <collection name="npz_files" type="list" label="${tool.name} on ${on_string}: Extracted NPZ files">
+            <discover_datasets pattern="(?P&lt;designation&gt;.+\.npz)$" directory="out_extract" format="npz"/>
+        </collection>
+
+        <collection name="marker_alignments" type="list" label="${tool.name} on ${on_string}: Marker alignments">
+            <filter>save_marker_aln</filter>
+            <discover_datasets pattern="(?P&lt;designation&gt;.+\.marker_aligned\.fasta)$" directory="out_extract" format="fasta"/>
+        </collection>
+        
+        <data name="msa_file" format="fasta" from_work_dir="out_extract/*.msa.fa" label="${tool.name} on ${on_string}: Multiple sequence alignment">
```
wm75: seems this output is in fasta.gz format, not fasta

tools/samestr/samestr_extract.xml
```diff
@@ -0,0 +1,171 @@
...
+                </element>
+            </output_collection>
+
+            <output_collection name="marker_alignments" type="list" count="1"/>
+
+            <output name="msa_file" ftype="fasta">
+                <assert_contents>
+                    <has_size min="100" />
+                </assert_contents>
+            </output>
```
wm75: This test needs to be improved.

It could be that Galaxy actually decompresses the fasta.gz when you declare fasta here (forgot the details; could also be you need to add decompress="true").

In any case you shouldn't just check the size, but for some actual content.

wm75 (commented):
In general, all those command_log outputs are entirely uninformative and should be removed.
This means that you will have to adjust all tests. Where possible don't just remove the command_log test, but add sth meaningful for other outputs instead. The current tests are in part not very stringent.

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,112 @@
+<tool id="samestr_convert" name="SameStr Convert" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
+    <description>Convert MetaPhlAn or mOTUS marker alignments to nucleotide variant profiles</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements"/>
+    <command detect_errors="exit_code"><![CDATA[
+    mkdir -p out_convert/ &&
```
wm75: mkdir is unnecessary here and in all other wrappers since samestrcan do the job itself

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,112 @@
...
+
+    cp '$alignment' 'file_0.bam' &&
+    cp '$profile' 'file_0.profile.txt' &&
```
wm75: 1. why are you copying these inputs instead of symlinking them?
    Is there a specific reason?
2. the hardcoded `file_0` will get baked into the npz output so you should really use the `element_identifier` in the names instead

tools/samestr/samestr_extract.xml
```diff
@@ -0,0 +1,157 @@
...
+    <command detect_errors="exit_code"><![CDATA[
+    mkdir -p out_extract/ &&
+
+    #for $i, $genome in enumerate($reference_genomes):
+        #if str($genome.ext) in ['fasta', 'fa', 'fna']:
+            cp '$genome' '$genome.element_identifier' &&
+        #else if str($genome.ext) == 'fasta.gz':
+            cp '$genome' '$genome.element_identifier' &&
+        #end if
+    #end for
```
wm75: 1. again, is copying necessary?
2. inside the if/else block you need to append correct file extensions or the exercise is pointless.
3. enumerate is not used

tools/samestr/samestr_extract.xml
```diff
@@ -0,0 +1,157 @@
...
+        </collection>
+
+        <collection name="marker_alignments" type="list" label="${tool.name} on ${on_string}: Marker alignments">
+            <filter>save_marker_aln</filter>
+            <discover_datasets pattern="(?P&lt;designation&gt;.+\.marker_aligned\.fasta)$" directory="out_extract" format="fasta"/>
+        </collection>
+        
+        <data name="msa_file" format="fasta.gz" from_work_dir="out_extract/*.msa.fa" label="${tool.name} on ${on_string}: Multiple sequence alignment">
+            <filter>save_marker_aln</filter>
+        </data>
```
wm75: the msa_file is not conditional on --save-marker-aln so shouldn't be here either.

tools/samestr/samestr_extract.xml
```diff
@@ -0,0 +1,157 @@
...
+    SameStr Extract obtains MetaPhlAn or mOTUS marker sequences from reference genomes using BLASTN. It extracts clade-specific marker 
+    sequences for strain-level analysis. The resulting variant profiles can be used in downstream SameStr workflow steps.
+
+    **Input Requirements**
+
+    Samestr Extract can process multiple reference genomes simultaneously. For each analysis, the following inputs must be provided:
+    - One or more files in FASTA format (.fasta, .fa, .fna or gzipped versions .fasta.gz, .fa.gz, .fna.gz) containing complete bacterial genomes.
+    - The exact clade identifier corresponding to the taxonomic group of interest in the selected database.
+
+    It is not possible to enter multiple bacterial genome files if the file extensions are different.
```
wm75: This is too hidden.
Instead say this directly in the param help: "Note: You cannot mix compressed and uncompressed fasta here."

tools/samestr/samestr_filter.xml
```diff
@@ -0,0 +1,203 @@
...
+    </macros>
+    <expand macro="requirements"/>
+    <command detect_errors="exit_code"><![CDATA[
+    mkdir -p out_filter/ &&
+
+    #set $input_filename = $input_files.name
+    #set $names_filename = $input_names.name
+
+    cp '$input_files' '$input_filename' &&
+    cp '$input_names' '$names_filename' &&
```
wm75: It seems that `filter` is able to work across files with potentially some information borrowing.
I'm afraid you need to switch this back to accepting multiple inputs and returning collection outputs.

tools/samestr/samestr_merge.xml
```diff
@@ -0,0 +1,98 @@
...
+    <description>Merge nucleotide variant profiles obtained from metagenomic samples for filtering and pairwise comparisons</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements"/>
+    <command detect_errors="exit_code"><![CDATA[
+    mkdir -p out_merge/ &&
+
+    #for $i, $file in enumerate($input_files):
+        ln -s '$file' '$file.element_identifier' &&
```
wm75: you need to append the suffix according to the ext attribute on those symlink names.
(the tests are only working because the right suffix is in the element_identifier.)

tools/samestr/samestr_stats.xml
```diff
@@ -0,0 +1,81 @@
...
+    <description>Generate statistics for SameStr variant profiles including coverage and diversity metrics</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements"/>
+    <command detect_errors="exit_code"><![CDATA[
+    mkdir -p out_stats/ &&
+
+    #set $input_filename = $input_files.name
+    #set $names_filename = $input_names.name
```
wm75: The .name is completely irrelevant for this tool, but you need to put the right file extensions on the inputs (again tests work only because the name has the suffix).

tools/samestr/samestr_stats.xml
```diff
@@ -0,0 +1,81 @@
...
+
+    cp '$input_files' '$input_filename' &&
+    cp '$input_names' '$names_filename' &&
```
wm75: would symlinking work?

tools/samestr/samestr_summarize.xml
```diff
@@ -0,0 +1,155 @@
...
+    <description>Identify and summarize shared bacterial strains across samples using variant profile similarity</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements"/>
+    <command detect_errors="exit_code"><![CDATA[
+    mkdir -p out_summarize/ tax_profiles/ out_compare/ &&
+
+    #for $file in $tax_profiles_dir:
+        cp '$file' tax_profiles/'$file.element_identifier' &&
```
wm75: why cp this input specifically, when you're symlinking the rest?

tools/samestr/samestr_summarize.xml
```diff
@@ -0,0 +1,155 @@
...
+    #end for
+
+    #for $file in $overlap_files:
+        ln -s '$file' out_compare/'$file.element_identifier' &&
+    #end for
+
+    #for $file in $close_files:
```
wm75: https://github.com/danielpodlesny/samestr/blob/ada3ca5a30ce3e0d869fb304182916e302f7ee56/samestr/summarize/read_samestr_data.py#L90 seems to not read these files at all, but only overlap and fraction files.

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,112 @@
...
+    --marker-dir '$database.fields.path'
+    --output-dir out_convert/
+    --min-vcov '$alignment_param.min_vcov'
+    --min-aln-identity '$alignment_param.min_aln_identity'
+    --min-aln-len '$alignment_param.min_aln_len'
+    --min-base-qual '$alignment_param.min_base_qual'
+    --min-aln-qual '$alignment_param.min_aln_qual'
+    &&
+
+    rm -f out_convert/command_log.json
```
wm75: This is unnecessary (here and in all other wrappers). Just not picking up the file as an output is sufficient.

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,112 @@
...
+                <element name="t__Propionibacterium_phage_PAS50.file_0" ftype="npz">
+                    <assert_contents>
+                        <has_size min="100" />
+                    </assert_contents>
+                </element>
+            </output_collection>
+        </test>
+    </tests>
+
+    <help><![CDATA[
```
wm75, line 91: It would be good to start the help with an introduction to the suite of tools that's shared (via a TOKEN) among all wrappers. This section should explain what samestr is, in general, and perhaps provide an overview of how the different wrappers interact.

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,112 @@
...
+    **SameStr Convert**
+    
+    SameStr Convert processes marker gene alignments from taxonomic profiling tools (MetaPhlAn or mOTUS) to nucleotide variant profiles 
+    for strain-level analysis. It outputs per-clade variant profiles that can be used in downstream SameStr workflow steps. 
+
+    **Input Requirements**
+
+    The following files must be provided:
+    - A file in BAM (.bam) format containing marker alignments generated by MetaPhlAn or mOTUS.
+    - A taxonomic profile file (.profile.txt) corresponding to the alignment.
```
wm75: You have several cases of such mal-formatted lists spread throughout the help sections. Please fix them (an empty line is required between the first list item and the line before.

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,112 @@
...
+    
+    **Alignment Parameters**
+
+    - Marker database. A MetaPhlAn or mOTUS clade marker database which matches the version used to generate the input alignments must be selected.
```
wm75: like here

tools/samestr/samestr_convert.xml
```diff
@@ -0,0 +1,112 @@
...
+    **Alignment Parameters**
+
+    - Marker database. A MetaPhlAn or mOTUS clade marker database which matches the version used to generate the input alignments must be selected. 
+    - Percent identity. Minimum percent identity between read and marker sequence. Alignments with lower identity are filtered out. 
+    - Minimum alignment length. Minimum number of aligned bases required. Shorter alignments are discarded.
+    
+    **Results and Outputs**
+
+    The tool generates three main outputs:
+    - Alignment statistics (".aln_stats.txt"). Contains a summary table with statistics for each processed alignment.
```
wm75: but not like here

wm75 (changes requested):

tools/samestr/samestr_convert.xml
xens25: I used to get errors from the symlinks so I switched to cp. But with the element_identifier symlinking is now working, I will update it in the PR.

wm75:
Ok, half-way done with what's hopefully a final pass over all wrappers.

xens25:
Hi @wm75 , let me know in case more changes need to be done. Thanks in advance!

wm75:
@xens25 I fixed up everything including another refactoring of the filter wrapper which is now back at consuming single datasets.

Remaining tasks for you would be to:
- test everything *carefully* by hand (with planemo serve)
- write some additional tests (like for the samples-select, marker-remove and marker-keep filter options)
- fix some more wrapper help sections and param help text (in compare, stats and summarize) to avoid command line jargon. Look at convert, extract, merge, etc. and see how I did it there.

xens25:
Hi @wm75 , I added the samples-select, marker-remove and marker-keep tests in Samestr Filter. I wasn't exactly sure which things needed to be improved in the help sections and help tags, so I just tried to change some to use similar wording that you used in the other tools. Let me know in case some still need to be improved. As for planemo serve, in Samestr convert I get for (SameStr Convert on : per-clade SNV profiles) the following: a list with 0 datasets . I'm not sure why that is since the test is the same that passed in planemo test. I think it might be related to the fact that npz files are stored not directly in out_convert but inside a subdirectory with the name of the file, in this case out_convert/phage_padded, but I'm not sure if that's the reason. The rest of the tools worked as intended.

wm75:
> SameStr Convert on : per-clade SNV profiles) the following: a list with 0 datasets . I'm not sure why that is since the test is the same that passed in planemo test.

Are you sure you configured things exactly like in that test? In particular, have you set "min_vcov" to 1? When I tried this last it was working.

xens25:
> > SameStr Convert on : per-clade SNV profiles) the following: a list with 0 datasets . I'm not sure why that is since the test is the same that passed in planemo test.
> 
> Are you sure you configured things exactly like in that test? In particular, have you set "min_vcov" to 1? When I tried this last it was working.

You are right! I completely missed that, I tested it again, Samestr Convert works fine!

wm75:
Cool! Then I don't think I have anything more to add here.

@SaimMomin12 ready from my side!

wm75 (approved):

SaimMomin12:
> Cool! Then I don't think I have anything more to add here.
> 
> @SaimMomin12 ready from my side!

Many thanks @wm75 for the thorough and detailed review :pray:

SaimMomin12 (approved):

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/24440679374

SaimMomin12:
>Unexpected HTTP status code: 500: {"err_msg":"Metadata may have been defined for some items in revision '5e4329e09b16'.  Correct the following problems if necessary and reset metadata.<br/><b>samestr_compare.xml</b> - This file requires an entry in the tool_data_table_conf.xml file.  Upload a file named tool_data_table_conf.xml.sample to the repository that includes the required entry to correct this error.<br/><b>samestr_convert.xml</b> - This file requires an entry in the tool_data_table_conf.xml file.  Upload a file named tool_data_table_conf.xml.sample to the repository that includes the required entry to correct this error.<br/><b>samestr_extract.xml</b> - This file requires an entry in the tool_data_table_conf.xml file.  Upload a file named tool_data_table_conf.xml.sample to the repository that includes the required entry to correct this error.<br/><b>samestr_filter.xml</b> - This file requires an entry in the tool_data_table_conf.xml file.  Upload a file named tool_data_table_conf.xml.sample to the repository that includes the required entry to correct this error.<br/><b>samestr_merge.xml</b> - This file requires an entry in the tool_data_table_conf.xml file.  Upload a file named tool_data_table_conf.xml.sample to the repository that includes the required entry to correct this error.<br/><b>samestr_stats.xml</b> - This file requires an entry in the tool_data_table_conf.xml file.  Upload a file named tool_data_table_conf.xml.sample to the repository that includes the required entry to correct this error.<br/><b>samestr_summarize.xml</b> - This file requires an entry in the tool_data_table_conf.xml file.  Upload a file named tool_data_table_conf.xml.sample to the repository that includes the required entry to correct this error.<br/>","err_code":500001}

@xens25 could you check on this and create a corresponding PR?

wm75:
@xens25 looks like we simply forgot to put the `tool_data_table_conf.xml.sample` file from the data manager also into the folder of the actual tools. Just copying over the file in a separate PR should be all that's needed.

PR #7314, opened 2025-09-29
Add regex switch expression tool

npinter:
As discussed here https://github.com/galaxyproject/galaxy/pull/20966 I added the regex switch module as an expression tool.

Workflow example:
grafik

Workflow route 1:
grafik

Workflow route 2:
grafik

@mvdbeek mentioned there is a Galaxy bug regarding the workflow invocation. Tool works but "Attempting to schedule workflow invocation" takes 5 min to pipe the boolean to the target tools.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/regex_switch/regex_switch.xml
+++ b/tools/regex_switch/regex_switch.xml
@@ -55,6 +55,7 @@
 
 Matches a regular expression against the name of the input file. It emits two booleans:
 `matched` (the match result) and `unmatched` (its negation). Connect either to a step's **Conditionally skip step?** input.
+
 **Regex syntax reference:** https://262.ecma-international.org/16.0/index.html#sec-patterns
 ]]></help>
 </tool>


npinter:
Other regex tools and other expression tools are located in the category "Text Manipulation", so I changed "Expression tools" to that.

npinter:
I don't see why it fails..

npinter:
Is it because the tool doesn't generate a real output (only booleans?)? @mvdbeek

tools/regex_switch/regex_switch.xml
```diff
@@ -0,0 +1,77 @@
...
+                <assert_contents>
+                    <has_line line="true"/>
+                </assert_contents>
+            </output>
+        </test>
+    </tests>
+    <help><![CDATA[
+**Regex switch**
+
+Matches a regular expression against an element name. It emits two booleans:
```
bernt-matthias: Guess "element name" needs better explanation. I guess you assume that the tool is mapped over a collection.

npinter: It's the name of the dict element (job["input"]["name"]) of cwl (?), which looks like that:

```
    "engineConfig": [],
    "job": {
        "input": {
            "file_ext": "tabular",
            "file_size": 14,
            "name": "table.tsv",
            "metadata": {
                "dbkey": "?",
                "data_lines": 2,
                "comment_lines": 0,
                "columns": 2,
                "column_types": [
                    "int",
                    "int"
                ],
                "column_names": [],
                "delimiter": "__tc__"
            },
            "src": {
                "src": "hda",
                "id": 4
            }
        },
        "pattern": "\.csv$"
    },
    "context": null,
    "outdir": null,
    "tmpdir": null,

```

Not sure if this name comes either of the element identifier or the element name (I guessed the latter).

npinter:
@bernt-matthias any idea why the testing fails?

npinter:
@bernt-matthias looks like tests are green now :)

tools/regex_switch/.shed.yml
```diff
@@ -0,0 +1,10 @@
+categories:
+- Text Manipulation
+description: Regex switch
+long_description: |
+  Regex switch outputs a boolean from regex on element identifier.
+name: regex_switch
+owner: iuc
+remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/master/tools/regex_switch
```
bernt-matthias: Master is now main

tools/regex_switch/regex_switch.xml
```diff
@@ -0,0 +1,60 @@
...
+        <token name="@TOOL_VERSION@">1.0.0</token>
+        <token name="@VERSION_SUFFIX@">0</token>
+    </macros>
+    <expression type="ecma5.1"><![CDATA[{
+        var name = $job.input.name;
+        var pat = $job.pattern;
+        var matched = false;
+
+        try {
+            if (pat !== "") {
```
bernt-matthias: Should we add a `empty_field` validator instead to the param?

tools/regex_switch/regex_switch.xml
```diff
@@ -0,0 +1,60 @@
+<tool name="Regex switch" id="regex_switch" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" tool_type="expression">
+    <description>boolean from regex on element identifier</description>
+    <macros>
+        <token name="@TOOL_VERSION@">1.0.0</token>
+        <token name="@VERSION_SUFFIX@">0</token>
+    </macros>
+    <expression type="ecma5.1"><![CDATA[{
+        var name = $job.input.name;
```
bernt-matthias, line 8: Should this be `element_identifier` instead of `name`?

tools/regex_switch/regex_switch.xml
```diff
@@ -0,0 +1,60 @@
...
+            <param name="input" value="table.tsv" ftype="txt"/>
+            <param name="pattern" value="\.csv$"/>
+            <output name="matched"   value_json="false"/>
+            <output name="unmatched" value_json="true"/>
+        </test>
+    </tests>
+    <help><![CDATA[
+**Regex switch**
+
+Matches a regular expression against the name of the input file. It emits two booleans:
```
bernt-matthias, line 56: Would be good to link a website on the syntax of ecma5.1 regex.

tools/regex_switch/regex_switch.xml
```diff
@@ -0,0 +1,60 @@
...
+    </macros>
+    <expression type="ecma5.1"><![CDATA[{
+        var name = $job.input.name;
+        var pat = $job.pattern;
+        var matched = false;
+
+        try {
+            if (pat !== "") {
+                var re = new RegExp(pat);
+                matched = re.test(name);
```
bernt-matthias: Is this checking if the regex is contained in name or if it matches the complete name (analogous to python match Vs search).

tools/regex_switch/regex_switch.xml
npinter: `element_identifier` is not available in the json return from cwtool:

```
    "engineConfig": [],
    "job": {
        "input": {
            "file_ext": "tabular",
            "file_size": 14,
            "name": "table.tsv",
            "metadata": {
                "dbkey": "?",
                "data_lines": 2,
                "comment_lines": 0,
                "columns": 2,
                "column_types": [
                    "int",
                    "int"
                ],
                "column_names": [],
                "delimiter": "__tc__"
            },
            "src": {
                "src": "hda",
                "id": 4
            }
        },
        "pattern": "\.csv$"
    },
    "context": null,
    "outdir": null,
    "tmpdir": null,

```

tools/regex_switch/regex_switch.xml
npinter: Matched is true, if the pattern (`pat` f.e. `\.csv$`) is contained within the name (f.e. `sample.csv`). So it's comparable to the search approach in Python. The match approach is a special case of the search approach. So I would suggest to keep `re.test(pat)` since it has more more degrees of freedom and a full match can be expressed by anchors (`^$`).

tools/regex_switch/regex_switch.xml
npinter: Yes! Will link https://262.ecma-international.org/16.0/index.html#sec-patterns

tools/regex_switch/regex_switch.xml
npinter: Thanks! Added!

npinter:
Testing fails again with the same cryptic error as before (..which turned green at some point :) )

bernt-matthias:
Let me use the opportunity to debug the failure. I guess the problem is due to https://github.com/galaxyproject/tools-iuc/pull/7293

.github/workflows/pr.yaml
```diff
@@ -354,6 +354,7 @@ jobs:
         mode: check
     - name: Check if all test chunks succeeded
       run: |
+        ls artifacts/
         NFILES=$(ls artifacts/ | grep "Tool test output" | wc -l)
```
bernt-matthias:
```suggestion
        ls artifacts/
        NFILES=$(find artifacts/ -name "tool_test_output.json" | wc -l)
```

bernt-matthias:
Fix is coming https://github.com/galaxyproject/tools-iuc/pull/7348 (I hope)

bernt-matthias (approved):

npinter:
sorry for bump, but looks like it worked :) thank you @bernt-matthias

PR #7318, opened 2025-09-30
Adding a new tool: read2tree

nromashchenko:
FOR CONTRIBUTOR:

As the title says, this request adds [read2tree](https://github.com/DessimozLab/read2tree).

### DIFF
--- a/tools/read2tree/read2tree.xml
+++ b/tools/read2tree/read2tree.xml
@@ -1,4 +1,4 @@
-<tool id="read2tree" name="Read2Tree" version="@TOOL_VERSION@" python_template_version="3.5" profile="24.0">
+<tool id="read2tree" name="Read2Tree" version="@TOOL_VERSION@" profile="24.0">
     <description>Infer a species tree from sequencing reads</description>
     <macros>
         <token name="@TOOL_VERSION@">2.0.1</token>


tools/read2tree/.shed.yml
```diff
@@ -0,0 +1,12 @@
...
+description: Infer a species tree from sequencing reads.
+homepage_url: https://github.com/DessimozLab/read2tree
+long_description: |
+    read2tree is a software tool for generating alignment matrices and performing tree inference directly from sequencing reads. It leverages the OMA database and a set of input reads, bypassing many of the standard steps typically required in phylogenomic analysis. In particular, it avoids for read filtering, assembly, gene prediction, gene annotation, all-vs-all comparison, orthology prediction, alignment, and concatenation.
+
+    See https://www.nature.com/articles/s41587-023-01753-4.
+
+remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/main/tools/read2tree
+categories:
+    - "Genomics Analysis"
```
bernt-matthias: You need to use one of the categories of the toolshed.

tools/read2tree/read2tree.xml
```diff
@@ -0,0 +1,107 @@
+<tool id="read2tree" name="Read2Tree" version="2.0.1" python_template_version="3.5">
```
bernt-matthias: Can you use a `TOOL_VERSION` macro for the version? Like here: https://github.com/galaxyproject/tools-iuc/blob/main/tools/berokka/berokka.xml

tools/read2tree/read2tree.xml
```diff
@@ -0,0 +1,107 @@
...
+       <xref type="bio.tools">read2tree</xref>
+    </xrefs>
+    
+    <requirements>
+        <requirement type="package" version="2.0.1">read2tree</requirement>
+    </requirements>
+
+    <command detect_errors="exit_code"><![CDATA[ 
+        rm -rf ./genes_dir; 
+        rm -rf ./output;
```
bernt-matthias: This should not be needed. Every jobs gets a fresh working dir.

tools/read2tree/read2tree.xml
```diff
@@ -0,0 +1,107 @@
...
+        
+        mkdir -p ./genes_dir;
+
+        #set $i = 0
+        #for $f in $marker_genes
+            #set $i = $i + 1
```
bernt-matthias:
```suggestion
        #for $i, $f in enumerate($marker_genes)
```

tools/read2tree/read2tree.xml
```diff
@@ -0,0 +1,107 @@
...
+            ln -s -- `readlink -f $f` "./genes_dir/$i".fa;
```
bernt-matthias:
```suggestion
            ln -s '$f' "./genes_dir/$i".fa;
```

tools/read2tree/read2tree.xml
```diff
@@ -0,0 +1,107 @@
...
+        #end for
+
+        #for $f in $reads
+            ln -s -- `readlink -f $f` "./$f.name";
```
bernt-matthias: same

tools/read2tree/read2tree.xml
```diff
@@ -0,0 +1,107 @@
...
+        #end for
+
+        read2tree --tree --standalone_path `realpath genes_dir` --reads
```
bernt-matthias: Also here `realpath` should not be needed?

tools/read2tree/read2tree.xml
bernt-matthias: Instead of `$f.name` use `f.element_identifier`. Plus: it needs to be single quoted and cleaned for characters that are malicious on the bash: eg https://github.com/galaxyproject/tools-iuc/blob/9fdf4ad9c2a9c026720e2b01460b7ac85056b068/tools/coverm/macros.xml#L55

tools/read2tree/read2tree.xml
```diff
@@ -0,0 +1,107 @@
...
+            ln -s -- `readlink -f $f` "./$f.name";
+        #end for
+
+        read2tree --tree --standalone_path `realpath genes_dir` --reads 
+        #for $read in $reads
+            $read.name
+        #end for
+        --output_path ./output --dna_reference $dna_ref > ./r2t.out 2> ./r2t.err;
+
+        ln -s -- `find output -type f -name "tree_*.nwk"` "./tree.nwk";
```
bernt-matthias: Is there only one nwk file? Then you could also directly use `from_workdir="tree_*.nwk"`

tools/read2tree/read2tree.xml
```diff
@@ -0,0 +1,107 @@
...
+
+        { cat ./r2t.out; cat ./r2t.err >&2; }
```
bernt-matthias: This line will not be executed if one of the previous calls fails (Galaxy uses `set -e` for each tool run if you set a recent profile). I would suggest not to use output redirection for the call to `read2tree`

tools/read2tree/read2tree.xml
```diff
@@ -0,0 +1,107 @@
...
+    ]]></command>
+
+    <inputs>
+        <param name="reads" type="data" multiple="true" format="fasta,fastq" label="Input reads in FASTA or FASTQ"
```
bernt-matthias: Use `fastqsanger` and/or `fastqillumina` instead of fastq. 

Does the software accept gzipped data?

tools/read2tree/read2tree.xml
```diff
@@ -0,0 +1,107 @@
...
+            help="Set of input reads for the species of interest."/>
+
+        <!--param name="marker_genes" type="data" multiple="true" format="fasta" label="Marker genes"/-->
+
+        <param name="marker_genes" type="data_collection" collection_type="list"
```
bernt-matthias: You could also use a data parameter with `multiple="true"`. Users can still use collections. But there might be good reason for using a acollection?

tools/read2tree/read2tree.xml
```diff
@@ -0,0 +1,107 @@
...
+        <!--data name="output_tree" format="newick" label="Output tree" from_work_dir="$output_tree"/-->
+
+        <data name="output_tree" format="newick" label="Output tree" from_work_dir="tree.nwk"/>
+    </outputs>
+    
+
+
+    <tests>
+        <test>
+            <param name="reads" value="sample_1.fastq,sample_2.fastq"/>
```
bernt-matthias: Guess the fastq files could be downsampled to reduce size?

tools/read2tree/read2tree.xml
```diff
@@ -0,0 +1,107 @@
...
+    <inputs>
+        <param name="reads" type="data" multiple="true" format="fasta,fastq" label="Input reads in FASTA or FASTQ"
+            help="Set of input reads for the species of interest."/>
+
+        <!--param name="marker_genes" type="data" multiple="true" format="fasta" label="Marker genes"/-->
+
+        <param name="marker_genes" type="data_collection" collection_type="list"
+            label="Marker gene files" help="A set of reference orthologous groups, i.e. marker genes. See https://github.com/DessimozLab/read2tree for detail."/>
+
+        <param name="dna_ref" type="data" format="fasta" label="DNA reference"
```
bernt-matthias: This seems to be optional? Could you add a few words to the help when this is needed and what it is?

tools/read2tree/read2tree.xml
bernt-matthias: Add `profile="24.0"`

bernt-matthias (commented):
Thanks for the submission. I added a few comments.

tools/read2tree/read2tree.xml
bernt-matthias: Can you use the `argument` attribute instead of (or in addition to) `name`

tools/read2tree/read2tree.xml
nromashchenko: marker genes could be dozens or more files. I figured it's easier to upload them and build into a collection first than do `multiple="true"`. Otherwise on the tool page the user has to select all them one by one again

bernt-matthias: If you look at the tool form there are a few buttons left of the data selector (for data inputs with `multiple="true"`) 

1. `Multiple datasets`: the user can select multiple datasets
2. `Dataset collection` the user can select a collection

That is collection input also works in this case. The only downside is that the `element_identifiers` are not really useful for users choosing option 1. 

I'm completely fine with collection input. Just wanted to make you aware.

tools/read2tree/read2tree.xml
```diff
@@ -0,0 +1,107 @@
...
+                    <element name="gene17" value="marker_genes/OMAGroup_1133018.fa"/>
+                    <element name="gene18" value="marker_genes/OMAGroup_1151179.fa"/>
+                    <element name="gene19" value="marker_genes/OMAGroup_1163384.fa"/>
+                    <element name="gene20" value="marker_genes/OMAGroup_1171372.fa"/>
+                    <element name="gene21" value="marker_genes/OMAGroup_1188079.fa"/>
+                </collection>
+                </param>
+            <param name="dna_ref" value="dna_ref.fa"/>
+            <output name="output_tree">
+                 <assert_contents>
```
bgruening: maybe making this test a bit more stricter? For example by asserting the line count?

tools/read2tree/.shed.yml
nromashchenko: fixed

tools/read2tree/read2tree.xml
nromashchenko: fixed

tools/read2tree/read2tree.xml
nromashchenko: fixed

tools/read2tree/read2tree.xml
nromashchenko: fixed

tools/read2tree/read2tree.xml
nromashchenko: fixed

tools/read2tree/read2tree.xml
nromashchenko: It does. The test uses gzipped fastq. I updated the tag format

tools/read2tree/read2tree.xml
nromashchenko: I downsampled and gzipped it, so now test files fit 1Mb

tools/read2tree/read2tree.xml
nromashchenko: It used to be, not anymore. The tool's help section is outdated and still says it's optional. In reality it crashes if you don't provide this argument. I'll report to the devs.

I expanded the description of what it is for the wrapper page.

tools/read2tree/read2tree.xml
nromashchenko: fixed

tools/read2tree/read2tree.xml
nromashchenko: fixed

tools/read2tree/read2tree.xml
nromashchenko: I find it's strict enough. If the output tree contains the sample name of the reads, it means read2tree successfully made reads place in the tree. Line count is usually always 1 in newick files anyway

tools/read2tree/read2tree.xml
```diff
@@ -0,0 +1,98 @@
+<tool id="read2tree" name="Read2Tree" version="@TOOL_VERSION@" python_template_version="3.5" profile="24.0">
```
bernt-matthias: <tool id="read2tree" name="Read2Tree" version="@TOOL_VERSION@" profile="24.0">

tools/read2tree/read2tree.xml
```diff
@@ -0,0 +1,98 @@
...
+            "$readln"
+        #end for
+        --output_path ./output --dna_reference $dna_ref
+    ]]></command>
+
+    <inputs>
+        <param argument="reads" type="data" multiple="true" format="fasta,fastqsanger,fastqillumina" label="Input reads in FASTA or FASTQ"
+            help="Set of input reads for the species of interest."/>
+
+        <param argument="marker_genes" type="data_collection" collection_type="list"
```
bernt-matthias: Just was wondering if there are typical sets of marker genes that are useful for many users. Then we could cache them in Galaxy. Would require the setup of a data table (and maybe data manager).

Definitely can leave this for the future. Just an idea.

tools/read2tree/read2tree.xml
```diff
@@ -0,0 +1,98 @@
...
+        read2tree --tree --standalone_path genes_dir --reads 
+        #for $read in $reads
+            #set readln = re.sub(r'[^\w\-\\.]', '_', str($f.element_identifier)) 
+            "$readln"
+        #end for
+        --output_path ./output --dna_reference $dna_ref
+    ]]></command>
+
+    <inputs>
+        <param argument="reads" type="data" multiple="true" format="fasta,fastqsanger,fastqillumina" label="Input reads in FASTA or FASTQ"
```
bernt-matthias:
```suggestion
<param argument="reads" type="data" multiple="true" format="fasta,fasta.gz,fastqsanger,fastqsanger.gz,fastqillumina,fastqillumina.gz" label="Input reads in FASTA or FASTQ"
```

The test works, but what is actually happening is that Galaxy automatically unpacks it.

tools/read2tree/read2tree.xml
```diff
@@ -0,0 +1,98 @@
...
+    </inputs>
+    
+    <outputs>
+        <data name="output_tree" format="newick" label="Output tree" from_work_dir="output/tree_*.nwk"/>
+    </outputs>
+    
+
+    <tests>
+        <test>
+            <param name="reads" value="sample_1.fastq.gz,sample_2.fastq.gz"/>
```
bernt-matthias: Add `ftype="fastqsanger.gz"`

tools/read2tree/read2tree.xml
bernt-matthias: Add `format`. I guess `fasta,fasta.gz`.

tools/read2tree/read2tree.xml
```diff
@@ -0,0 +1,98 @@
...
+    ]]></command>
+
+    <inputs>
+        <param argument="reads" type="data" multiple="true" format="fasta,fastqsanger,fastqillumina" label="Input reads in FASTA or FASTQ"
+            help="Set of input reads for the species of interest."/>
+
+        <param argument="marker_genes" type="data_collection" collection_type="list"
+            label="Marker gene files" help="A set of reference orthologous groups, i.e. protein sequences of the marker genes for the reference species. See https://github.com/DessimozLab/read2tree for an example and information on how to obtain a list of marker genes."/>
+
+        <param argument="dna_ref" type="data" format="fasta" label="DNA reference"
```
bernt-matthias: `fasta.gz`?

tools/read2tree/read2tree.xml
```diff
@@ -0,0 +1,98 @@
...
+
+        <param argument="marker_genes" type="data_collection" collection_type="list"
+            label="Marker gene files" help="A set of reference orthologous groups, i.e. protein sequences of the marker genes for the reference species. See https://github.com/DessimozLab/read2tree for an example and information on how to obtain a list of marker genes."/>
+
+        <param argument="dna_ref" type="data" format="fasta" label="DNA reference" 
+            help="Reference file containing nucleotide sequences for the set of marker genes presented. See https://github.com/DessimozLab/read2tree for detail."/>
+    </inputs>
+    
+    <outputs>
+        <data name="output_tree" format="newick" label="Output tree" from_work_dir="output/tree_*.nwk"/>
```
bernt-matthias: Please remove the `label`. How do the output filenames look like, i.e. what is `output/tree_*.nwk` matching?

tools/read2tree/read2tree.xml
nromashchenko: I tried doing this, but somehow for the data collection it doesn't allow me to choose the collection in the browser if I do this. "No list dataset collections with fasta or fasta.gz elements available" although I upload a bunch of .fa files and build them. Shouldn't be accepted format something else in this case?

nromashchenko: There is no such problem if I don't mention format for marker_genes

bernt-matthias: What is the data type of the elements of you collection?

tools/read2tree/read2tree.xml
nromashchenko: done

tools/read2tree/read2tree.xml
nromashchenko: Okay thanks! It turns out I forgot to point out the format when making the collection. I added `format`

tools/read2tree/read2tree.xml
nromashchenko: it seems to have been addressed as currently your suggestion matches the code

tools/read2tree/read2tree.xml
nromashchenko: done

tools/read2tree/read2tree.xml
nromashchenko: I removed `label`. As for the pattern, it matches `tree_${one of the paired reads filename}.nwk`, so depends on what's user input files are named.

tools/read2tree/read2tree.xml
nromashchenko: I am afraid it's difficult to select such sets. I'd leave this for future

tools/read2tree/read2tree.xml
nromashchenko: Thanks. Incorporated in [035a](https://github.com/galaxyproject/tools-iuc/pull/7318/commits/035a70bd7eafa1429d7d429fe97dcab28639ae9d)

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/19610718720

bernt-matthias:
Deployment will be fixed here: https://github.com/galaxyproject/tools-iuc/pull/7482

Issue #7327, opened 2025-10-03
Remarks for Jbrowse2


lldelisle:
With @hrhotz we found that:
- "Track Group" maybe transformed to "Track Categories"
- "Annotation Track" maybe transformed to simply "Track" as it can contains both annotations and bigwig/bam

I am wondering if the "Track Visibility" should not be "On when browser opens" by default.

I also suggested that you put a sanitizer for the "default region to display", you may use something like:
https://github.com/galaxyproject/tools-iuc/blob/5efe64a2c3cb7495d708f539d341fa53fba07fe4/tools/pygenometracks/pyGenomeTracks.xml#L833

abretaud:
Thanks @lldelisle, I've started https://github.com/galaxyproject/tools-iuc/pull/7330 :)

PR #7330, opened 2025-10-04
JBrowse2 polishing

abretaud:
FOR CONTRIBUTOR:

Addressing comments from https://github.com/galaxyproject/tools-iuc/issues/7327

"I am wondering if the "Track Visibility" should not be "On when browser opens" by default." -> that's already the default

I'll finish adding gtf support next week

### DIFF
--- a/tools/jbrowse2/jbrowse2.xml
+++ b/tools/jbrowse2/jbrowse2.xml
@@ -310,7 +310,7 @@ cp '$output.files_path/index.html' '$output';
             </when>
         </conditional>
 
-        <repeat name="assemblies" min="1" title="Genome Assemblies">
+        <repeat name="assemblies" min="1" title="Genome Assembly">
         <conditional name="reference_genome">
             <param label="Reference genome to display" name="genome_type_select" type="select">
                 <option selected="True" value="indexed">Use a built-in genome</option>
@@ -348,19 +348,21 @@ cp '$output.files_path/index.html' '$output';
              https://github.com/GMOD/jbrowse-components/issues/1765
         -->
 
-        <param label="Default region to display" type="text" name="defaultLocation" value="" help="e.g. 'ctgA:1234..5678'"/>
+        <param label="Default region to display" type="text" name="defaultLocation" value="" help="e.g. 'ctgA:1234..5678'">
+            <validator type="expression" message="Region should be like ctgA:1234..5678">'^[a-zA-Z0-9_]:\d+\.\.\d+$'</validator>
+        </param>
 
         <param name="ref_name_aliases" type="data" format="tsv,tabular" label="Reference sequence name aliases" help="Use when chromosomes have alternate names (TSV file: first column is the names from your genome sequence, the rest of the columns are aliases." optional="true" />
 
         <param name="cytobands" type="data" format="bed" label="cytobands data" help="BED file containing cytobands locations" optional="true" />
 
-        <repeat name="track_groups" title="Track Group">
+        <repeat name="track_groups" title="Track Category">
             <param label="Track Category"
                 name="category"
                 type="text"
                 value="Default"
                 help="Organise your tracks into Categories for a nicer end-user experience. You can use #date# and it will be replaced with the current date in 'yyyy-mm-dd' format, which is very useful for repeatedly updating a JBrowse instance when member databases / underlying tool versions are updated." optional="False"/>
-            <repeat name="data_tracks" title="Annotation Track">
+            <repeat name="data_tracks" title="Track">
                 <conditional name="data_format" label="Track Options">
                     <param type="select" label="Track Type" name="data_format_select">
                         <option value="gene_calls">GFF/GFF3/BED Features</option>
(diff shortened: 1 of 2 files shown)


abretaud:
Green and ready for review :)

GTF can be loaded know, although GFF is probably better for performances and test indexing

I also fixed a bug on synteny data that I hadn't seen in the previous PR

@lldelisle any other things to change?

tools/jbrowse2/jbrowse2.xml
```diff
@@ -348,19 +348,21 @@ cp '$output.files_path/index.html' '$output';
...
+        <param label="Default region to display" type="text" name="defaultLocation" value="" help="e.g. 'ctgA:1234..5678'">
+            <validator type="expression" message="Region should be like ctgA:1234..5678">'^[a-zA-Z0-9_]:\d+\.\.\d+$'</validator>
+        </param>
 
         <param name="ref_name_aliases" type="data" format="tsv,tabular" label="Reference sequence name aliases" help="Use when chromosomes have alternate names (TSV file: first column is the names from your genome sequence, the rest of the columns are aliases." optional="true" />
 
         <param name="cytobands" type="data" format="bed" label="cytobands data" help="BED file containing cytobands locations" optional="true" />
 
-        <repeat name="track_groups" title="Track Group">
+        <repeat name="track_groups" title="Track Category">
```
lldelisle: I think I would put here the help: "Organise your tracks into different Categories for a nicer end-user experience."

And for the param below (Track Category) I would put "Track Category Label" because if I understand well the parameter defines the label of the track category. And then put as help "You can use #date#..."

lldelisle: :heart:

abretaud:
Thanks! I've made the help/label changes, it's better indeed :)

For the tests, I used sim_size because there are many history/dataset ids and paths in both xml and json files, that keep on changing on each planemo run, I tried initially to use lines_diff, but it's painful to count the lines, and it will be painful too to update it. While using sim_size with delta=0, I'm quite sure that the json and xml will be exactly like expected, except the fixed-length-random-chunk-of-strings. There was also the idea of using asserts, but I think we'd miss a lot of small potential differences in the output. We'll see in the future if it's convenient or not like that

bgruening:
You can use the asserts __in addition__ if you like. And for XML we have those fancy asserts: https://docs.galaxyproject.org/en/latest/dev/schema.html#attribute-is

might be also a good assert to have: https://docs.galaxyproject.org/en/latest/dev/schema.html#is-valid-xml

For json as well: https://docs.galaxyproject.org/en/latest/dev/schema.html#has-json-property-with-text

mvdbeek:
Could you also replace the `.get_metadata()` function call and use `.metadata` (a property) instead ?

abretaud:
Ok, I added the is-valid-xml test, and replaced the get_metadata() call. For detailed json/xml content asserts I'd vote for doing it later as I don't really have the time to do it now

abretaud:
Anyone willing to review/merge this one?

tools/jbrowse2/jbrowse2.xml
```diff
@@ -144,8 +144,8 @@ cp '$output.files_path/index.html' '$output';
                             display_name="Unnamed History"/>
                             #end if
                         <metadata
-                            #for (key, value) in $dataset.get_metadata().items():
-                            #if "_types" not in $key and $value is not None and len(str($value)) < 5000:
+                            #for (key, value) in $dataset.metadata.items():
+                            #if "_types" not in $key and not $key.endswith("_index") and $key != "species_chromosomes" and $key != "metadata_incomplete" and $value is not None and len(str($value)) < 5000:
```
bgruening, line 152: That could really need a comment :)

tools/jbrowse2/jbrowse2.xml
```diff
@@ -703,8 +718,12 @@ cp '$output.files_path/index.html' '$output';
...
             </repeat>
             <output name="output">
-                <extra_files name="galaxy.xml" type="file" value="out/02_remote/galaxy.xml" compare="sim_size" />
-                <extra_files name="config.json" type="file" value="out/02_remote/config.json" compare="sim_size" />
+                <extra_files name="galaxy.xml" type="file" value="out/02_remote/galaxy.xml" compare="sim_size" delta="0">
+                    <assert_contents>
+                        <is_valid_xml />
+                    </assert_contents>
+                </extra_files>
+                <extra_files name="config.json" type="file" value="out/02_remote/config.json" compare="sim_size" delta="0" />
```
bgruening: has_json_property_with_text and has_json_property_with_value exist.

SaimMomin12:
[‎.github/workflows/pr.yaml](https://github.com/galaxyproject/tools-iuc/pull/7330/files#diff-1eb4e5fd5611777d4e597ef299a1cb5ba8050c28a2dabbd4fbc56205d69e5dda) I assume this was included in the PR by mistake??

tools/jbrowse2/jbrowse2.xml
abretaud: I've added 2 examples of has_json_property_with_value, but it's a bit painful to write, we can add more later

PR #7342, opened 2025-10-07
Update kma DM

bgruening:
FOR CONTRIBUTOR:

### DIFF
--- a/data_managers/data_manager_build_kma_index/data_manager/kma_build_index.xml
+++ b/data_managers/data_manager_build_kma_index/data_manager/kma_build_index.xml
@@ -1,9 +1,9 @@
 <?xml version="1.0"?>
-<tool id="kma_build_index" name="KMA Index Builder" tool_type="manage_data" version="1.2.21+galaxy0" profile="19.01">
+<tool id="kma_build_index" name="KMA Index Builder" tool_type="manage_data" version="1.6.6+galaxy0" profile="24.1">
     <description>kma index builder</description>
     <requirements>
-        <requirement type="package" version="1.2.21">kma</requirement>
-        <requirement type="package" version="3.7">python</requirement>
+        <requirement type="package" version="1.6.6">kma</requirement>
+        <requirement type="package" version="3.12">python</requirement>
     </requirements>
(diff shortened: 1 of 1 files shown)


SaimMomin12 (approved):

wm75:
> Let's remove the python wrapper.

Fair suggestion. I can add the necessary changes later. Currently trying to find out why the kma aligner tool itself produces empty outputs for the index I built while testing the current state :-(

data_managers/data_manager_build_kma_index/data_manager_conf.xml
```diff
@@ -6,10 +6,10 @@
 		<column name="name"/>
 		<column name="path" output_ref="out_file">
 		    <move type="directory">
-		        <source>#import os#${os.path.dirname($path)}</source>
-			<target base="${GALAXY_DATA_MANAGER_DATA_PATH}">kma_index/#import os#${os.path.dirname($path)}</target>
+		        <source>${path}</source>
+				<target base="${GALAXY_DATA_MANAGER_DATA_PATH}">kma_index/${value}</target>
 		    </move>
-                    <value_translation>${GALAXY_DATA_MANAGER_DATA_PATH}/kma_index/${path}</value_translation>
+			<value_translation>${GALAXY_DATA_MANAGER_DATA_PATH}/kma_index/${value}/index</value_translation>
```
bernt-matthias, line 12: Compared to the old version there is now and additional `/index` in the path, or am I wrong? Is this intended? 

Maybe we should document in the `tool-data/loc` file what the path points to?

wm75: It's the same as before just that in the old version put the `/index` onto `$path`, then had to do the `#import os#${os.path.dirname($path)}` trick to discover the directory, while the new version uses the directory as `$path` and adds the `/index` only to the final record. The new way feels a bit less hacky because it avoids the Python in the data_manager_conf.xml

wm75: documenting it is a good idea!

Issue #7344, opened 2025-10-08
Semibin2: working files cannot be found if they start with ‘.’ (filenames inherited from input files)


cat-bro:
We’ve had some jobs fail on Galaxy Australia because the name of the bam file in the working directory is “.fastqsanger.bam” which is inherited from the input file name “.fastqsanger” and the job fails with "Error: Expected file '*.bam' does not exist” and I think it just can’t find it because it starts with a “.” and is therefore hidden.

the working bam file name is set here: https://github.com/galaxyproject/tools-iuc/blob/main/tools/semibin/macros.xml#L161-L162

This is similar but not the same as issue https://github.com/galaxyproject/tools-iuc/issues/7326

SantaMcCloud:
I will change this in the new version of the wrapper since therer is an update to 2.2.0 or should it be first getting fix in the current version and then have the update comming?

SantaMcCloud:
https://github.com/galaxyproject/tools-iuc/pull/7366

Issue #7326, opened 2025-10-03
Semibin2 semibin_concatenate_fasta fails due to file names


paulzierep:
The tool fails with `Error: Make sure that every contig file have different names.`if files have a `.` due to this code upstream
https://github.com/BigDataBiology/SemiBin/blob/886913b48814cc0cd426ee46ec20d22cee875bf5/SemiBin/utils.py#L498
Need to fix the sanitizer in the wrapper.

SantaMcCloud:
Since i update the wrapper for Semibin i can change this. I will replace the current way with just a counting to have file_x which should be fine

SantaMcCloud:
Just befor that should i fix it first on the current version and then let the update be merged or is fixing this in the new version fine?

bernt-matthias:
> I will replace the current way with just a counting to have file_x which should be fine

Sample names are not needed? 

Alternatively just fix the sanitation in the wrapper

paulzierep:
I would also think its better to just fix the senitizer

SantaMcCloud:
I will fix it then do you have the filename which caused the error on hand?

bernt-matthias:
I guess it will be sufficient to remove `\\.` from the regex https://github.com/galaxyproject/tools-iuc/blob/848922d15b3e17a864b3e40fdfed97e35f5646e5/tools/semibin/macros.xml#L127 .. then all `.` will be replaced by `_`

SantaMcCloud:
@bernt-matthias i did change it here: https://github.com/galaxyproject/tools-iuc/pull/7347/commits/b100b632e71d027d277ddd84efd6b35232a845f5

bernt-matthias:
> [@bernt-matthias](https://github.com/bernt-matthias) i did change it here: [b100b63](https://github.com/galaxyproject/tools-iuc/commit/b100b632e71d027d277ddd84efd6b35232a845f5)

Excellent.  Can you make this a separate PR? I guess https://github.com/galaxyproject/tools-iuc/pull/7347 might need a bit, because of the pt problem.

SantaMcCloud:
> > [@bernt-matthias](https://github.com/bernt-matthias) i did change it here: [b100b63](https://github.com/galaxyproject/tools-iuc/commit/b100b632e71d027d277ddd84efd6b35232a845f5)
> 
[...]

PR #7347, opened 2025-10-08
Update Semibin 2.0.2 to 2.3.0

SantaMcCloud:
FOR CONTRIBUTOR:


SantaMcCloud:
Test will fail because a datatype is still missing: https://github.com/galaxyproject/galaxy/pull/21024
Also i did comment out a test since i think the tool can not detect anything to make this test work with the amount of data got with the test but im not sure in this case.

SantaMcCloud:
The file `model.pt` now the question could this be slip such that even it is bigger then 1 MB it will be approved or should i remove it and also remove all the test using this as input to test certain functionality?

SaimMomin12:
> The file `model.pt` now the question could this be slip such that even it is bigger then 1 MB it will be approved or should i remove it and also remove all the test using this as input to test certain functionality?

For file sizes >1 MB, you can consider uploading it to Zenodo and using the link in the test

SantaMcCloud:
> > The file `model.pt` now the question could this be slip such that even it is bigger then 1 MB it will be approved or should i remove it and also remove all the test using this as input to test certain functionality?
> 
> For file sizes >1 MB, you can consider uploading it to Zenodo and using the link in the test

Ah okay thank you i didnt know i will do it this way then!

SantaMcCloud:
https://github.com/galaxyproject/tools-iuc/issues/7344

SantaMcCloud:
https://github.com/galaxyproject/tools-iuc/issues/7326

tools/semibin/bin.xml
```diff
@@ -44,7 +44,7 @@ SemiBin2 bin
                 <expand macro="environment"/>
             </when>
             <when value="history">
-                <param argument="--model" type="data" format="h5" label="Trained semi-supervised deep learning model"/>
+                <param argument="--model" type="data" format="pt" label="Trained semi-supervised deep learning model"/>
```
bgruening: @SantaMcCloud do you know if we can use the safetensor datatype here? Or that this tool can consume safetensor datatypes?

SantaMcCloud: Im not sure i didnt see anything about this in the documentation. This might work with the old file too but i have to look at it since certain packages has to be downgraded in this case. 

I will test it first if this works with safetensor or not and will let you know if it works or not!

SantaMcCloud: @bgruening i try to run sembin2 with `safetensor `but it looks like it can not be used. The only option here is to run a convert befor and after when a `.pt `file is used/outputed somewhere. My question now is it is possible to write a short python script in the commnad line of the wrapper to start torch and tranform it to `safetensor`?

With this it is possible to ouput the model created in the `sembin_train` will be safetensor. For `semibin_bin` the convert to `safetensor` to `.pt`  should be also doable so semibin can run without problem and the `.pt` file can be deleted via commnad line.

bgruening: From the Galaxy side that is possible, yes. You can include a small auxiliary Python script with the tool and then do the conversion.

If is possible from the semibin side, e.g. that no data gets lost during the transformation, I do not know :(

SantaMcCloud: I can not tell it either i can try to have some runs with the test data and see if there is any problem with it. and will give you feedback then.

For the script is there any source or exmaple for it so i can have a look into it?

SantaMcCloud: @bgruening so on the first look the little script i wrote for the specific usecase for semibin to convert the .pt file to safetensor and back works. both result with the testdata are the same with the original model and the converted model. So this should work as workaround! :)

bgruening: Nice! This maybe helps https://github.com/galaxyproject/tools-iuc/blob/848922d15b3e17a864b3e40fdfed97e35f5646e5/tools/macs2/macs2_callpeak.xml#L127

its just an additional script that you call around the original command.
You can put many commands in the command section, just separate them by &&

SantaMcCloud: I did add the script and i hope now that it works with it, i also open an issue to see if the upstream devs might change it in the futher

SantaMcCloud:
```
  File "/home/runner/work/tools-iuc/tools-iuc/tools/semibin/convert.py", line 4, in <module>
    from safetensors.torch import save_file, load_file
ModuleNotFoundError: No module named 'safetensors'
```

based on this i should rewrote the recipe and include the safetensors package? @bgruening

paulzierep: If this is the correct package: https://anaconda.org/anaconda/safetensors, you could include it in the requirements. Galaxy will take care of a mulled package / container that has both dependencies.

paulzierep:
Independent of the safetensor issue, there is also a bug in the upstream tool:
https://github.com/BigDataBiology/SemiBin/issues/204
Can we sneak a fallback into the wrapper for this? Check if the error occurs and, if so, produce an empty bin as output? 

Another one for very small contigs:
```
2025-09-26 15:25:03 vgcnbwc-worker-c120m405-0021.novalocal SemiBin[1158340] INFO Binning for short_read
2025-09-26 15:26:36 vgcnbwc-worker-c120m405-0021.novalocal SemiBin[1158340] INFO Did not detect GPU, using CPU.
2025-09-26 15:26:37 vgcnbwc-worker-c120m405-0021.novalocal SemiBin[1158340] INFO Generating training data...
2025-09-26 15:26:54 vgcnbwc-worker-c120m405-0021.novalocal SemiBin[1158340] INFO Calculating coverage for every sample.
2025-09-26 15:26:56 vgcnbwc-worker-c120m405-0021.novalocal SemiBin[1158340] INFO Processed: 50contig_reads_2_name_50-contigs.bam
2025-09-26 15:26:57 vgcnbwc-worker-c120m405-0021.novalocal SemiBin[1158340] INFO Start binning.
2025-09-26 15:27:47 vgcnbwc-worker-c120m405-0021.novalocal SemiBin[1158340] INFO Number of bins prior to reclustering: 0
2025-09-26 15:27:52 vgcnbwc-worker-c120m405-0021.novalocal SemiBin[1158340] INFO Running naive ORF finder

Error: Sequence file /data/jwd07/main/088/687/88687613/tmp/tmpl59myacm/orfs.faa is empty or misformatted

Error: Running hmmsearch fail

```

A fallback for both would be good.

SantaMcCloud:
@paulzierep here is the fallback now if one of the 2 errors happen!

@bgruening you also should have a look over it since i dont know how to controll the exit code for Galaxy. Currently now evreything will be true and no error will get marked for Galaxy that the tool did failed. Do you have any idea how to have the control over it so only these  errors are not marked as error but evreyother is?

paulzierep:
> @paulzierep here is the fallback now if one of the 2 errors happen!
> 
> @bgruening you also should have a look over it since i dont know how to controll the exit code for Galaxy. Currently now evreything will be true and no error will get marked for Galaxy that the tool did failed. Do you have any idea how to have the control over it so only these errors are not marked as error but evreyother is?

I am not I understand the second comment, but there are a bunch of errors we should look into: https://github.com/galaxyproject/tools-iuc/actions/runs/19618557561?pr=7347

SantaMcCloud:
> > @paulzierep here is the fallback now if one of the 2 errors happen!
> > @bgruening you also should have a look over it since i dont know how to controll the exit code for Galaxy. Currently now evreything will be true and no error will get marked for Galaxy that the tool did failed. Do you have any idea how to have the control over it so only these errors are not marked as error but evreyother is?
> 
> I am not I understand the second comment, but there are a bunch of errors we should look into: https://github.com/galaxyproject/tools-iuc/actions/runs/19618557561?pr=7347

I wanted to restart this at some point but first o wanted to fix the logic problem. This is the second comment. The Problem here is that all the commands are joined via ands which means that when any of this command will fail the whole run will be failed and there will be a failed dataset/collection in galaxy aswell. This means even the 2 errors which i filter out the result will be a failed dataset/collection in Galaxy. Therefore to i had to manually overwrite, via an or, the error Code of Semibin such that the whole run will be completed anf that galaxy has empty files as output for it. This means since i have to manually overwrite it that ALL other erros never will be chatch by galaxy since i dont know if i can overwrite the whole runs exit code in an if Block. Because of this i did ask Björn if he maybe has an idea to not break the whole logic im Galaxy because of the overwrite.

SaimMomin12:
@paulzierep can you help reviewing here?

PR #7363, opened 2025-10-15
Add memory flag to metaSPAdes for better RAM handeling

SantaMcCloud:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/spades/macros.xml
+++ b/tools/spades/macros.xml
@@ -1,6 +1,6 @@
 <macros>
     <token name="@TOOL_VERSION@">4.2.0</token>
-    <token name="@VERSION_SUFFIX@">0</token>
+    <token name="@VERSION_SUFFIX@">1</token>
     <xml name="requirements">
         <requirements>
             <requirement type="package" version="@TOOL_VERSION@">spades</requirement>


tools/spades/metaspades.xml
```diff
@@ -49,6 +50,7 @@ metaspades.py
             <option value="--iontorrent">Iontorrent: required when assembling IonTorrent data (--iontorrent)</option>
         </expand>
         <expand macro="optional_output"/>
+        <param argument="--memory" type="integer" value="350" min="64" max="350" label="Set RAM limitation" help="When using a big input it is better to set a RAM limit to reduce the chance that metaSPAdes will get an error based on memory usage"/>
```
SaimMomin12, line 53: This is not a good idea to expose memory parameters to users. Instead this needs to be set up as a rule in TPV configs by admins. You can refer to some examples here https://github.com/usegalaxy-eu/infrastructure-playbook/blob/ee168ba226433d32f033663eaffb4e0a979fe694/files/galaxy/tpv/tools.yml

SantaMcCloud: This it a tool Limit such that certain hasmaps or so will be reduce to handle the the tool better.

@paulzierep can give you a better overview about it.

bgruening:
```suggestion
```

tools/spades/metaspades.xml
```diff
@@ -23,6 +23,7 @@
 ## run
 metaspades.py
     -o 'output'
+    --memory ${memory}
```
bgruening: please use `\$GALAXY_MEMORY_MB` or `\$GALAXY_MEMORY_MB_PER_SLOT`

tools/spades/metaspades.xml
bernt-matthias: I'm with @SaimMomin12. Also it is already set here https://github.com/galaxyproject/tools-iuc/blob/9fc4d01c03ad19302b6fd4b16264f6ae31729139/tools/spades/macros.xml#L117

paulzierep: Ah I did not check weather the wrapper already got this flag. I this case we can drop this PR, question would be why the tool then still fails due to memory, maybe we should reduce it a bit to make sure its below the Galaxy provided memory. 
Another question is whether this option changes the assembly logic, e.g. automatically use less kmers. If so we should reconsider the logic here, since that would mean, that TPV adjusts memory -> other -m paramter -> different assembly -> not reproducible. Will check with the devs.

paulzierep: FYI: https://github.com/ablab/spades/issues/1535

SantaMcCloud: If we set could we use a math fomrula there so we can calculate the memory for based on the Galaxy_var and try to set a Limit so that only the max of the set TSV memory can be used for the -m flag? Possible also based on the calation only give the tool 90/95% of it?

bgruening: https://github.com/galaxyproject/tools-iuc/blob/ff9564c2e6924fab084b6e02a3a196650041636a/tools/metabuli/metabuli.xml#L18

SaimMomin12: > Ah I did not check weather the wrapper already got this flag. I this case we can drop this PR, question would be why the tool then still fails due to memory, maybe we should reduce it a bit to make sure its below the Galaxy provided memory. Another question is whether this option changes the assembly logic, e.g. automatically use less kmers. If so we should reconsider the logic here, since that would mean, that TPV adjusts memory -> other -m paramter -> different assembly -> not reproducible. Will check with the devs.

So according to the upstream devs, changing the value of `-m` does not have any effect on the results. https://github.com/ablab/spades/issues/1535#issuecomment-3410938962

SantaMcCloud: > https://github.com/galaxyproject/tools-iuc/blob/ff9564c2e6924fab084b6e02a3a196650041636a/tools/metabuli/metabuli.xml#L18

could you use a min function on this will never be more memory then galaxy will provide?

bernt-matthias: taking the min is certainly possible, question is what you want to achieve: 

- just limit the memory (because spades may use a bit more than the value given with `-m`)
- allow user/workflow input of memory usage prediction (that could be picked up by TPV)

paulzierep: > https://github.com/galaxyproject/tools-iuc/blob/ff9564c2e6924fab084b6e02a3a196650041636a/tools/metabuli/metabuli.xml#L18

We should do that, maybe go for -10 to be sure and check if that works for some of our failed jobs.

SantaMcCloud: > taking the min is certainly possible, question is what you want to achieve:
> 
> * just limit the memory (because spades may use a bit more than the value given with `-m`)
> * allow user/workflow input of memory usage prediction (that could be picked up by TPV)

my idea was to have a hard Limit so it will never be more then the set memory from TPV. Even if my idea is bad it is still a learning for me thats why i did ask the question :)

paulzierep: > allow user/workflow input of memory usage prediction (that could be picked up by TPV)

Thats another question entirely, but I would say like in this case, we cannot directly allow it, but maybe with some encoding or as file metadata ... still no idea tbh :)

bernt-matthias: > We should do that, maybe go for -10 to be sure and check if that works for some of our failed jobs.

-10GB seems quite a lot. Could we ask upstream for a reasonable value.

SantaMcCloud: I think at a point it was the default option to use only 90% but i can ask them if they have a good idea for this. But we can also have different ratio for it like only remove 1 GB from it which will act as a buffer

SantaMcCloud: https://github.com/ablab/spades/issues/1536

bgruening: Provide me please with a patch, a single line that I need to add to the current wrapper on EU, and we can test it in real-life.

PR #7366, opened 2025-10-17
Fix sanitizer from SemiBin

SantaMcCloud:
FOR CONTRIBUTOR:

### DIFF
--- a/tools/semibin/macros.xml
+++ b/tools/semibin/macros.xml
@@ -101,56 +101,26 @@
         <param name="min_len" type="integer" min="0" value="0" label="Minimal length for contigs to be kept"/>
     </xml>
     <token name="@SINGLE_FASTA_FILES@"><![CDATA[
-#if $input_fasta.ext.endswith(".gz")
-ln -s '$input_fasta' 'contigs.fasta.gz' &&
-#else if $input_fasta.ext.endswith(".bz2")
-ln -s '$input_fasta' 'contigs.fasta.bz2' &&
-#else
-ln -s '$input_fasta' 'contigs.fasta' &&
-#end if
+ln -s '$input_fasta' 'contigs.$input_fasta.ext' &&
     ]]></token>
     <token name="@FASTA_FILES@"><![CDATA[
 #if $mode.select == 'single' or $mode.select == 'co'
-    #if $mode.input_fasta.ext.endswith(".gz")
-ln -s '$mode.input_fasta' 'contigs.fasta.gz' &&
-    #else if $mode.input_fasta.ext.endswith(".bz2")
-ln -s '$mode.input_fasta' 'contigs.fasta.bz2' &&
-    #else
-ln -s '$mode.input_fasta' 'contigs.fasta' &&
-    #end if
+ln -s '$input_fasta' 'contigs.$input_fasta.ext' &&
 #else
     #if $mode.multi_fasta.select == 'concatenated'
-        #if $mode.multi_fasta.input_fasta.ext.endswith(".gz")
-ln -s '$mode.multi_fasta.input_fasta' 'contigs.fasta.gz' &&
-        #else if $mode.multi_fasta.input_fasta.ext.endswith(".bz2")
-ln -s '$mode.multi_fasta.input_fasta' 'contigs.fasta.bz2' &&
-        #else
-ln -s '$mode.multi_fasta.input_fasta' 'contigs.fasta' &&
-        #end if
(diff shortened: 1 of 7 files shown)


SantaMcCloud:
@bernt-matthias

SantaMcCloud:
i had to change nameing of the softlink a bit to remove the file extension which was in the name. Now `'.'` should be fine!

tools/semibin/macros.xml
```diff
@@ -124,11 +124,12 @@ ln -s '$mode.multi_fasta.input_fasta' 'contigs.fasta' &&
 #set $separator = $mode.multi_fasta.separator
     #else
         #for $e in $mode.multi_fasta.input_fasta
-            #set $identifier = re.sub('[^\s\w\-\\.]', '_', str($e.element_identifier))
+            #set $identifier = re.sub('[^\s\w\-]', '_', str($e.element_identifier))
+            #set $final_identifier = '_'.join($identifier.split('_')[:-1])
```
bernt-matthias: This is not a good idea. If people have collection element identifiers containing any sanitized character or an underscore this will remove part of the identifier. Users should remove file extensions during collection creation.

I guess you had some problems in the test due to the file extensions. Then its better to adapt test assumptions.

SantaMcCloud: The problem in thr test was a code error because of the new naming of the file. The code tried to run to find something with `S1` but since the name of the file was `S1_fasta` it could not find it. So therefor i dont know how to fix beside of deleteing the test which should not be the point.

Do you might have any idea for this because i didnt thought about your point. I can also add a check to see if there is any file extension left of so cut it out?

bernt-matthias: I guess S1 is somewhere hardcoded in the test data files. This then need to be changed.

SantaMcCloud: After some work it should be fine now. I had to add extra files just for the one test case where the naming of the file was a problem.

SantaMcCloud: You should be able to merge it now. The fix is also implemented in my other PR!

tools/semibin/macros.xml
```diff
@@ -125,11 +125,10 @@ ln -s '$mode.multi_fasta.input_fasta' 'contigs.fasta' &&
...
         #for $e in $mode.multi_fasta.input_fasta
             #set $identifier = re.sub('[^\s\w\-]', '_', str($e.element_identifier))
-            #set $final_identifier = '_'.join($identifier.split('_')[:-1])
             #if $e.ext.endswith(".gz")
-gunzip -c '$e' > '${final_identifier}.fasta' &&
+gunzip -c '$e' > '${identifier}.fasta' &&
             #else
-ln -s '$e' '${final_identifier}.fasta' &&
+ln -s '$e' '${identifier}.fasta' &&
             #end if
```
bernt-matthias:
```suggestion
            ln -s '$e' '${identifier}.${e.ext}' &&
```

semibin should be able to unzip itself: https://github.com/BigDataBiology/SemiBin/blob/886913b48814cc0cd426ee46ec20d22cee875bf5/SemiBin/fasta.py#L23

also we should allow `fasta.bz2` in the formats of the input parameter (macros input-fasta-single and input-fasta-multiple). But this probably also requires changes in other tools?

SantaMcCloud: Will change it later today :)

SantaMcCloud: @bernt-matthias sorry that it did take a bit longer i had to adapte the logic for the changes. Now Semibin will use `.gz` and `.bz2` files and will open them in the function so therer is no unzipping befor needed.

I also add and changes some test to cover this! I hope these changes should be good enough now?

tools/semibin/macros.xml
```diff
@@ -102,38 +102,56 @@
     </xml>
     <token name="@SINGLE_FASTA_FILES@"><![CDATA[
 #if $input_fasta.ext.endswith(".gz")
-gunzip -c '$input_fasta' > 'contigs.fasta' &&
+ln -s '$input_fasta' 'contigs.fasta.gz' &&
```
bernt-matthias: You may just drop the whole if-else block and just 

```
ln -s '$input_fasta' 'contigs.$input_fasta.ext'
```

And analogously in all other places where we us `ln`

SantaMcCloud: should be done now! I also change the usage of the files with this to make it consistent! :)

PR #7407, opened 2025-10-29
Add Samtools Consensus

Smkingsize:
FOR CONTRIBUTOR:
https://www.htslib.org/doc/samtools-consensus.html

### DIFF
--- a/tool_collections/samtools/bam_to_sam/bam_to_sam.xml
+++ b/tool_collections/samtools/bam_to_sam/bam_to_sam.xml
@@ -1,4 +1,4 @@
-<tool id="bam_to_sam" name="BAM-to-SAM" version="2.0.6" profile="@PROFILE@">
+<tool id="bam_to_sam" name="BAM-to-SAM" version="2.0.7" profile="@PROFILE@">
     <description>convert BAM to SAM</description>
 
     <macros>
--- a/tool_collections/samtools/sam_to_bam/sam_to_bam.xml
+++ b/tool_collections/samtools/sam_to_bam/sam_to_bam.xml
@@ -1,4 +1,4 @@
-<tool id="sam_to_bam" name="SAM-to-BAM" version="2.1.4" profile="@PROFILE@">
+<tool id="sam_to_bam" name="SAM-to-BAM" version="2.1.5" profile="@PROFILE@">
     <description>convert SAM to BAM</description>
 
     <macros>
--- a/tool_collections/samtools/samtools_bedcov/samtools_bedcov.xml
+++ b/tool_collections/samtools/samtools_bedcov/samtools_bedcov.xml
@@ -1,4 +1,4 @@
-<tool id="samtools_bedcov" name="Samtools bedcov" version="2.0.6" profile="@PROFILE@">
+<tool id="samtools_bedcov" name="Samtools bedcov" version="2.0.7" profile="@PROFILE@">
     <description>calculate read depth for a set of genomic intervals</description>
     <macros>
         <import>macros.xml</import>
--- a/tool_collections/samtools/samtools_calmd/samtools_calmd.xml
+++ b/tool_collections/samtools/samtools_calmd/samtools_calmd.xml
@@ -1,4 +1,4 @@
-<tool id="samtools_calmd" name="Samtools calmd" version="2.0.6" profile="@PROFILE@">
+<tool id="samtools_calmd" name="Samtools calmd" version="2.0.7" profile="@PROFILE@">
     <description>recalculate MD/NM tags</description>
     <macros>
         <import>macros.xml</import>
--- a/tool_collections/samtools/samtools_flagstat/samtools_flagstat.xml
+++ b/tool_collections/samtools/samtools_flagstat/samtools_flagstat.xml
@@ -1,4 +1,4 @@
-<tool id="samtools_flagstat" name="Samtools flagstat" version="2.0.7" profile="@PROFILE@">
+<tool id="samtools_flagstat" name="Samtools flagstat" version="2.0.8" profile="@PROFILE@">
     <description>tabulate descriptive stats for BAM datset</description>
 
     <macros>
--- a/tool_collections/samtools/samtools_idxstats/samtools_idxstats.xml
+++ b/tool_collections/samtools/samtools_idxstats/samtools_idxstats.xml
@@ -1,4 +1,4 @@
-<tool id="samtools_idxstats" name="Samtools idxstats" version="2.0.7" profile="@PROFILE@">
+<tool id="samtools_idxstats" name="Samtools idxstats" version="2.0.8" profile="@PROFILE@">
     <description>reports stats of the BAM index file</description>
     <macros>
         <import>macros.xml</import>
--- a/tool_collections/samtools/samtools_mpileup/samtools_mpileup.xml
+++ b/tool_collections/samtools/samtools_mpileup/samtools_mpileup.xml
@@ -1,4 +1,4 @@
-<tool id="samtools_mpileup" name="Samtools mpileup" version="2.1.9" profile="@PROFILE@">
+<tool id="samtools_mpileup" name="Samtools mpileup" version="2.2.0" profile="@PROFILE@">
     <description>multi-way pileup of variants</description>
     <macros>
         <import>macros.xml</import>
--- a/tool_collections/samtools/samtools_reheader/samtools_reheader.xml
+++ b/tool_collections/samtools/samtools_reheader/samtools_reheader.xml
@@ -1,4 +1,4 @@
-<tool id="samtools_reheader" name="Samtools reheader" version="2.0.5" profile="@PROFILE@">
+<tool id="samtools_reheader" name="Samtools reheader" version="2.0.6" profile="@PROFILE@">
     <description>copy SAM/BAM header between datasets</description>
     <macros>
         <import>macros.xml</import>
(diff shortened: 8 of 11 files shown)


tool_collections/samtools/samtools_consensus/.shed.yml
```diff
@@ -0,0 +1,10 @@
+categories:
+- SAM
+description: Generate consensus from a SAM, BAM or CRAM file.
+long_description: |
+  This tool generates consensus from a SAM, BAM or CRAM file based on the contents of the alignment records. The consensus is written either as FASTA, FASTQ, or a pileup oriented format. This is selected using the -f FORMAT option.
+name: samtools_conssensus
+owner: devteam
```
RZ9082:
```suggestion
owner: iuc
```

tool_collections/samtools/samtools_consensus/samtools_consensus.xml
```diff
@@ -0,0 +1,611 @@
...
+        <!-- Basic options -->
+        <param name="input" type="data" format="sam,bam,cram" label="BAM file"/>
+        <param argument="--format" type="select" label="Output file type" help="Produce format FMT, with 'fastq', 'fasta' and 'pileup' as permitted options.">
+            <option value="fasta" selected="true">FASTA</option>
+            <option value="fastq">FASTQ</option>
+            <option value="pileup">Pileup</option>
+        </param>
+        <!-- Preprocessing options (reads) -->
+        <section name="preprocessing_options" title="Preprocessing Options" expanded="no">
+            <param name="minimum_MQ" argument="--min-MQ" type="integer" value="0" label="Minimum mapping quality" help="Filters out reads with a mapping quality below specified value"/>
```
bernt-matthias: Are there reasonable min/max values for numeric partameters?

tool_collections/samtools/samtools_consensus/samtools_consensus.xml
```diff
@@ -0,0 +1,611 @@
...
+            <param name="minimum_BQ" argument="--min-BQ" type="integer" value="0" label="Minimum base quality" help="Filters out bases with a base quality below specified value"/>
+            <param argument="--rf" type="text" label="Include reads by FLAG" value="0"  help="Only include reads with at least one FLAG bit set"/>
```
bernt-matthias: Is it possible to reuse https://github.com/galaxyproject/tools-iuc/blob/f28bd9adb58adbb4c0b52cbd8dfe334fe939700e/tool_collections/samtools/samtools_view/samtools_view.xml#L327

tool_collections/samtools/samtools_consensus/samtools_consensus.xml
```diff
@@ -0,0 +1,611 @@
...
+            </when>
+            <!-- bayesian 1.16 mode options -->
+            <when value="bayesian_116">
+                <expand macro="bayesian_settings_macro"/>
+            </when>
+        </conditional>
+        <!-- Gloabl Settings -->
+        <section name="global_settings" title="Global Settings" expanded="no">
+            <param name="minimum_depth" argument="--min-depth" type="integer" value="1" label="Minimum depth" help="The minimum depth required to make a call. Failing this depth check will produce consensus 'N', or absent if it is an insertion. Note this check is performed after filtering by flags and mapping/base quality"/>
+            <conditional name="region_cond">
```
bernt-matthias: Can we reuse https://github.com/galaxyproject/tools-iuc/blob/f28bd9adb58adbb4c0b52cbd8dfe334fe939700e/tool_collections/samtools/samtools_view/samtools_view.xml#L277

tool_collections/samtools/samtools_consensus/samtools_consensus.xml
```diff
@@ -0,0 +1,611 @@
...
+            <param name="show_insertions" argument="--show-ins" label="Show insertions" type="boolean" truevalue="yes" falsevalue="no" checked="true"  help="Whether to show insertions in the consensus"/>
+            <param argument="--ambig" type="boolean" truevalue="--ambig" falsevalue="" checked="false" label="IUPAC ambiguity codes" help="Enables IUPAC ambiguity codes in the consensus output. Without this the output will be limited to A, C, G, T, N and *"/>
+            <param name="mark_insertions" argument="--mark-ins" type="boolean" truevalue="--mark-ins" falsevalue="" checked="false" label="Mark insertions" help="Mark insertions by adding an underscore before every inserted base, plus a corresponding character in the quality for fastq format. When used in conjunction with '-a' and '--show-del', this permits an easy derivation of the consensus to reference coordinate mapping"/>
+        </section>
+    </inputs>
+    <!-- Outputs -->
+    <outputs>
+        <data name="output_file" format="fasta" label="${tool.name} on ${on_string}: consensus">
```
bernt-matthias: We do not need a label if there is only one output

tool_collections/samtools/samtools_consensus/samtools_consensus.xml
Smkingsize: not as it is since my tool does not support region from bed file. And I like my implementation more since it doesn't require the user to use the format **[REG:START-END]** but assembles it automatically! 
```
  <repeat name="reg_repeat" title="region" min="1">
      <param name="reg" type="text" optional="false" format="fasta" label="Region"/>
      <param name="start" type="integer" format="fasta" optional="true" label="Start"/>
      <param name="end" type="integer" format="fasta" optional="true" label="End"/>
  </repeat>
```

tool_collections/samtools/samtools_consensus/samtools_consensus.xml
Smkingsize: yes! should I edit the `macros.xml` and the `samtools_view.xml` accordingly?

tool_collections/samtools/samtools_consensus/samtools_consensus.xml
Smkingsize: I will look into it!

tool_collections/samtools/samtools_consensus/samtools_consensus.xml
bernt-matthias: Do we need a repeat here, ie can you really specify multiple regions on the command line?

In anyway the conditional can be removed by making the repeat (or if not needed: the parameter) optional.

I would maybe use the same parameter as here https://github.com/galaxyproject/tools-iuc/blob/f28bd9adb58adbb4c0b52cbd8dfe334fe939700e/tool_collections/samtools/samtools_view/samtools_view.xml#L285 via a macro. Maybe just add a validator over there?

Smkingsize: yes we can specify multiple regions in case the SAM file has multiple references so more than one consensus would be generated! 1 for each ref mapping.

should I write the `regions` param with a validator as a macro and use it in both tools then?

tool_collections/samtools/samtools_consensus/samtools_consensus.xml
bernt-matthias: Yes, please

tool_collections/samtools/samtools_consensus/samtools_consensus.xml
bernt-matthias: Yes, please

Smkingsize: so after deeper look. consensus don't allow multiple region specification! 
and the logic of the region selection of consensus and view are not the same. (for example samtools view accepts '*' to output unmmaped reads while samtools don't have that option)

Smkingsize:
linting is failing with some tools. example. 
`.. ERROR (ShedVersion): bam_to_sam: version 2.0.6 is less or equal than version of the latest installable revision 2.0.6`
how can I fix this? (should I increment the version token in macros?) `token name="@VERSION_SUFFIX@">0</token>`

@bernt-matthias

tool_collections/samtools/macros.xml
```diff
@@ -217,6 +229,11 @@
         </conditional>
     </xml>
 
+    <!-- add region macro -->
+    
+    <!-- add exclusive filter macro -->
+
+
```
bernt-matthias:
```suggestion
```

tool_collections/samtools/samtools_consensus/samtools_consensus.xml
```diff
@@ -0,0 +1,606 @@
+<tool id="samtools_consensus" name="Samtools consensus" version="0.1.0+galaxy0" python_template_version="3.5" profile="@PROFILE@">
```
bernt-matthias: version should be @TOOL_VERSION@+galaxy@VERSION_SUFFIX@

bernt-matthias: python template version can be removed

bernt-matthias (commented):
You can ignore the warning about the versions. It just warns us that a few tools of the suite are not bumped. Which is fine here. We can overwrite this when merging.

bernt-matthias:
Unfortunately you also need to manually bump all tools that are not using the `TOOL_VERSION` macro token: e.g. https://github.com/galaxyproject/tools-iuc/blob/3d0c8087a4329e615218f1228a0ae30f5e9fb3fa/tool_collections/samtools/bam_to_sam/bam_to_sam.xml#L1 (since they use `PROFILE`).

Since they have a version > TOOL_VERSION the suggestion is to increase only the minor version (hoping that samtools will catch up at some point ;) )

bernt-matthias (approved):

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/19296541104

mvdbeek:

https://github.com/galaxyproject/tools-iuc/actions/runs/19296541104

bernt-matthias:
Guess I will need to push samtools_consensus manually  and then we retrigger deployment

bernt-matthias:
Hrm. The tool actually never was tested, because it does not load: `Error reading tool from path: /home/runner/work/tools-iuc/tools-iuc/tool_collections/samtools/samtools_consensus/samtools_consensus.xml`

bernt-matthias:
@Smkingsize I started to fix bugs here: https://github.com/galaxyproject/tools-iuc/pull/7454, but I do not have the time to finish it atm. Can you do so?

